PR #244, opened 2013-11-09
Make autogen.sh use bash

JasonGross:
Otherwise, on my Ubuntu machine, it complains about SIGINT being a bad trap.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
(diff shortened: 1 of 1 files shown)


Zimmi48:
For wider compatibility, `#!/bin/bash` should be replaced with `#!/usr/bin/env bash` (in case the `bash` command is installed somewhere else). See also the discussion at https://github.com/coq/coq/pull/456.
Thanks!

PR #631, opened 2014-12-03
Lex modalities and the fracture theorem

mikeshulman:
Long-promised.  (-:  Some notes:

1. Also includes some general theory about pullbacks, in `Pullback.v`, and a notation `f^* g` for the pullback of `g` along `f`.
2. At first I tried making lex-ness a module type to be assumed on top of a modality module-type, as for accessibility.  But then I realized that since lex-ness is a "universe-local property" (unlike accessibility!) we can just make it a typeclass and allow ordinary universe polymorphism to kick in.
3. Given this, I was strongly tempted _not_ to include separate files `Lex` and `Topological`, putting the relevant theory into `Modality` or `Accessible` depending on whether it needs accessibility.  (This would even be possible for `Fracture`.)  This would have the advantage of avoiding the proliferation of `_Theory` modules and (presumably) solving the multiple-inheritance problem described in the comments in `Lex.v`.  I have so far resisted the temptation, but we'll see how annoyed I get in the future with the current setup.  (Would using `Include` instead of `Export`, if we trusted it to be anomaly-free, also solve these problems?)


mikeshulman:
Rebased again.  (Among other things, I've been apparently using the notation `^*` for both pullbacks of maps and inversion of pointed homotopies.  Now I've put them in separate scopes, at least.)

JasonGross:
Travis is unhappy.  What's the error message? (My phone's browser doesn't handle travis logs on low quality wifi.)

mikeshulman:
I can't find an error message; it just ends on line 10000 while printing out `./theories/Pullback.timing`.

mikeshulman:
I found some references on the internet to travis cutting off all logs after 10000 lines.  Can we change our build scripts to spew less so that we can see the actual error?

JasonGross:
It cuts off after line 10000.  You click the "download log" button towards the top to get the full log.

mikeshulman:
Looks like"trouble in init_rank" again.

JasonGross:
Ugggghhh, `dot` is looping again.

mikeshulman:
It would sure be nice if Travis would add a comment at the end saying "this build log has been truncated after 10000 lines; click Download Log at the top to obtain the complete log".

JasonGross:
I'm sure they have a bug tracker / feature request form somewhere.

mikeshulman:
This PR is coming up on 2 months old; it would be really nice to get it merged.

JasonGross:
I will hopefully have a chance to look at `dot` again in the next week or two.

mikeshulman:
Rebased again (`mapinO_pr1` was duplicated between this PR and a recent one).

JasonGross:
Oh, hey, travis seems happy.  Do we need ASPOE, or has this been sitting around long enough without objections that we should just merge it?

andrejbauer:
If I squint a little, I can be a second pair.

spitters:
I am still motivated to actually read it, but I don't think that needs to
block merging.
Please go ahead and I'll send patches in necessary. Sorry, that this toke
so long.

mikeshulman:
Thanks!

Issue #636, opened 2014-12-08
Make a directory for theories of specific types


mikeshulman:
I think we should have a directory to contain theories about _specific_ types (by contrast with `Types/` which is about the generic type formers).  For instance, in `Idempotent.v` there is some theory of cantor space, and also of the operation `BAut`.  Once we free the HITs #573, we could put `Circle` and `Spheres` etc. there too.  But I'm not sure how to name this directory, since `Types/` is taken.  One thought is `Spaces`, but possibly that's a bad idea.

JasonGross:
We now have `Spaces`

PR #658, opened 2014-12-29
Move BAut, Cantor, and BAut(Cantor) to Spaces/

mikeshulman:
Partial progress towards resolving #636.  Unless I messed up somewhere, this commit only moves existing code around.


mikeshulman:
APOE?

PR #660, opened 2014-12-29
split off Basics/Decidable.v, with a few new lemmas in it

mikeshulman:


theories/Basics/Decidable.v
```diff
@@ -0,0 +1,128 @@
...
+Require Import Overture PathGroupoids Contractible Equivalences Trunc.
+Local Open Scope equiv_scope.
+Local Open Scope trunc_scope.
+Local Open Scope path_scope.
+
+(** * Decidability *)
+
+(** ** Definitions *)
+
+(* NB: This has to come after our definition of [not] (which is in [Overture]), so that it refers to our [not] rather than the one in [Coq.Logic]. *)
```
andrejbauer: Is this NB still relevant? It got copied from `Overture`.

mikeshulman: I thought it would be nice to keep it, to remind the reader that our `not` is different from the one in the stdlib.

Issue #657, opened 2014-12-29
Fixing `compose` and catching up with trunk


JasonGross:
As per [bug 3881](https://coq.inria.fr/bugs/show_bug.cgi?id=3881), we managed to catch Coq in a state where it used a different order of evar resolution than in 8.4.  With https://github.com/coq/coq/commit/96906c2a1ba9426271a4048bfa8b3991db51c192, Coq switched back to how it was doing things in 8.4, resulting in many cases of `isequiv_compose` being uninferrable.  If we β-expand `o` to make it
```coq
Notation compose := (fun g f x => g (f x)).
Notation "g 'o' f" := (compose g f) (at level 40, left associativity) : function_scope.
```
I suspect this will be enough to fix it.  I am open to other suggestions, though.  As an example alternative, simply fixing the proofs will involve many diffs that look something like
```diff
diff --git a/theories/Basics/Equivalences.v b/theories/Basics/Equivalences.v
index 5a15081..4fc64d8 100644
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -148,9 +148,14 @@ Global Instance symmetric_equiv : Symmetric Equiv | 0 := @equiv_inverse.
 (** If [g \o f] and [f] are equivalences, so is [g].  This is not an Instance because it would require Coq to guess [f]. *)
 Definition cancelR_isequiv {A B C} (f : A -> B) {g : B -> C}
   `{IsEquiv A B f} `{IsEquiv A C (g o f)}
-  : IsEquiv g
+  : IsEquiv g.
-  := isequiv_homotopic (compose (compose g f) f^-1)
-       (fun b => ap g (eisretr f b)).
+Proof.
+  refine (isequiv_homotopic
+            ((g o f) o f^-1)
+            (fun b => ap g (eisretr f b))).
+  refine (@isequiv_compose _ _ f^-1 _ _ (g o f) _).
+  exact _.
+Defined.

 Definition cancelR_equiv {A B C} (f : A -> B) {g : B -> C}
   `{IsEquiv A B f} `{IsEquiv A C (g o f)}
```

mikeshulman:
Note: see discussion at #663 for what still needs fixing.

mikeshulman:
I have a question about the above modified proof of `cancelR_isequiv`:

```coq
refine (isequiv_homotopic
          ((g o f) o f^-1)
          (fun b => ap g (eisretr f b))).
[...]

PR #663, opened 2015-01-05
Use [refine] rather than [apply] in more places

JasonGross:
See #657 for discussion.  Mainly, this is a way to aggregate the changes
and allow discussion of the name `reapply` (for a combination of
`rapply` and `eapply`), so that the commit that bumps trunk will be
small.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -645,12 +645,6 @@ universe.  Thus, applying `symmetry` to an `Equiv` introduces a
 strictly larger universe.  A solution is to `apply equiv_inverse`
 instead.  Similarly, use `equiv_compose'` instead of `transitivity`.
 
-Given `P : B -> Type` and `f : A -> B`, writing `P o f` introduces a
-universe parameter strictly larger than the codomain of `P` (since it
-has to be passed to the function `compose`).  A solution is to write
-`fun a => P (f a)` instead.  (This may no longer be true with
-`compose` a notation rather than a function.)
-
 Typeclass inference doesn't always find the simplest solution, and may
 insert unnecessary calls to instances that introduce additional
 universes.  One solution is to alter the proofs of those instances as
@@ -898,10 +892,14 @@ instead.
 ### Notations ###
 
 The operation `compose`, notation `g o f`, is simply a notation for
-`fun x => g (f x)` rather than a defined constant.  This means that,
-for instance, you can't partially apply it and write `compose g` for
-`fun f => g o f`.  We could define `compose := (fun g f x => g (f x))`
-instead of `compose g f := (fun x => g (f x))` to allow this, but we
+`fun x => g (f x)` rather than a defined constant.  We define `compose
+:= (fun g f x => g (f x))` so that typeclass inference can pick up
+`isequiv_compose` instances.  This has the unfortunate side-effect
+that `simpl`/`cbn` is enough to "unfold" `compose`, and there's no way
+to prevent this.  We could additionally define `g o f := (fun x => g
+(f x))` to change this, but this would result in identically looking
+goals which are really different.  We consider it poor style to use
+`compose` as a partially applied constant, such as `compose g`; we
 take the point of view that `fun f => g o f` is more readable anyway.
 
 ### Unfolding definitions ###
@@ -986,6 +984,20 @@ where they are defined.
   automatically.  If you need a version with more fields than yet
   exists, feel free to add it.)
 
+- `rapply`, `erapply`: Defined in `coq/theories/Program/Tactics` and
+  `Basics/Overture` respectively, these tactics are more well-behaved
+  variants of `apply` for theorems with fewer than 16 arguments.  (It
+  is trivial to extend it to *n* arguments for any finite fixed *n*.)
+  The unification algorithm used by `apply` is different and often
+  less powerful than the one used by `refine`, though it is
+  occasionally better at pattern matching.  If `apply` fails with a
+  unification error you think it shouldn't have, try `rapply` or
+  `erapply`.
+
+  The difference between `rapply` and `erapply` is that `rapply` only
+  accepts lemmas with no holes (and will do typeclass inference
+  early), while `erapply` accepts lemmas with holes (such as `ap f`,
+  i.e., `@ap _ _ f _ _`), and does typeclass inference late.
 
 ## Contributing to the library ##
 
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -117,7 +117,7 @@ Notation "x .2" := (pr2 x) (at level 3, format "x '.2'") : fibration_scope.
 
 (** Composition of functions. *)
 
-Notation compose g f := (fun x => g (f x)).
+Notation compose := (fun g' f' x => g' (f' x)). (* use primed names because https://coq.inria.fr/bugs/show_bug.cgi?id=3892 *)
 
 (** We put the following notation in a scope because leaving it unscoped causes it to override identical notations in other scopes.  It's convenient to use the same notation for, e.g., function composition, morphism composition in a category, and functor composition, and let Coq automatically infer which one we mean by scopes.  We can't do this if this notation isn't scoped.  Unfortunately, Coq doesn't have a built-in [function_scope] like [type_scope]; [type_scope] is automatically opened wherever Coq is expecting a [Sort], and it would be nice if [function_scope] were automatically opened whenever Coq expects a thing of type [forall _, _] or [_ -> _].  To work around this, we open [function_scope] globally. *)
 Notation "g 'o' f" := (compose g f) (at level 40, left associativity) : function_scope.
@@ -601,6 +601,14 @@ Global Existing Instance ispointed_type.
 (** Homotopy fibers are homotopical inverse images of points.  *)
 
 Definition hfiber {A B : Type} (f : A -> B) (y : B) := { x : A & f x = y }.
+
+(** *** More tactics *)
+
+(** We want access to useful standard library tactics, such as [rapply]. *)
+Require Export Coq.Program.Tactics.
+
+(** [erapply lem] is like [apply lem] (rather, [rapply lem]), but it allows holes in [lem] *)
+Tactic Notation "erapply" open_constr(term) := rapply term.
 
 (** Ssreflect tactics, adapted by Robbert Krebbers *)
 Ltac done :=
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -290,7 +290,7 @@ Section Reflective_Subuniverse.
   Section Functor.
 
     (** In this section, we see that [O] is a functor. *)
-    
+
     Definition O_functor {A B : Type} (f : A -> B) : O A -> O B
       := O_rec (to O B o f).
 
@@ -325,10 +325,10 @@ Section Reflective_Subuniverse.
     Proof.
       intros x.
       transitivity (O_functor (f o pi1) x).
-      - symmetry; apply O_functor_compose.
+      - symmetry; erapply O_functor_compose.
       - transitivity (O_functor (g o pi2) x).
         * apply O_functor_homotopy, comm.
-        * apply O_functor_compose.
+        * erapply O_functor_compose.
     Defined.
 
     (** Functoriality on identities *)
@@ -380,7 +380,7 @@ Section Reflective_Subuniverse.
         + intros y; apply eissect.
         + apply O_functor_idmap.
     Defined.
-      
+
     Definition equiv_O_functor {A B : Type} (f : A <~> B)
     : O A <~> O B
     := BuildEquiv _ _ (O_functor f) _.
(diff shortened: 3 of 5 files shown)


mikeshulman:
I think your markdown is a bit confused.  But `reapply` seems like a terrible name to me, if that's the discussion.  `erapply` would be a little better...

JasonGross:
Markdown fixed (it was actually coqdoc).

JasonGross:
I'll go make it `erapply` soon (as soon as I finish porting the chain categories).

mikeshulman:
Why do we need `erapply` instead of `rapply` in places where we used to have just `apply`?

JasonGross:
Because sometimes `lem` in `apply lem` has maximally inserted implicit arguments that aren't inferrable until the type of the goal is known, and hence `constr:(lem)` fails because it can't fully infer the types.   And `apply` is special.

For example, `apply (ap f)` works, while `constr:(ap f)` complains that it can't figure out `x` and `y`.

JasonGross:
Rename done

mikeshulman:
Rrgh.  There ought to be a better way to deal with that than jumping all the way to an `e`tactic.  E.g. there ought to be a variant of `constr:` that says "don't try to infer even maximally-inserted implicit arguments in this term".  But I guess we're stuck with what we have.

mikeshulman:
How about updating the section of `STYLE.md` that discusses `compose`, and also maybe adding to it a mention of `rapply`, `erapply`, and the brokenness of `apply`?

JasonGross:
Well, `rapply` is already kind-of like `eapply`, because `refine` lets you leave evars over.  `uconstr` might do what we want, but I don't have a good sense of when we should use `uconstr` vs `open_constr` (except, `uconstr` if you want to have holes take on multiple different evars, like as in a pattern).

mikeshulman:
I thought there was an `erefine` that allowed evars but `refine` didn't.

JasonGross:
If there was an `erefine`, it was a long time ago.  `refine` allows evars even in 8.4.

mikeshulman:
Probably I was just confused, because `refine` generally does what I want and doesn't confuse me by sticking weird evars in the goals the way tactics like `eapply` tend to.  (-:

JasonGross:
Well, `refine` leaves you with dependent subgoals ;-)

mikeshulman:
But evars are different from dependent subgoals.  Maybe they're not very different internally, but they present differently to the user, because I have to / get to prove the dependent subgoals, whereas an evar like `?1` I don't have any control over or understanding of.

JasonGross:
This used to be true of evars, but not anymore:
```coq
Goal exists n, n = 0.
Proof.
  eexists.
  Unshelve.
  all:cycle 1.
  (* 2 subgoals, subgoal 1 (ID 6)

  ============================
   nat

subgoal 2 (ID 7) is:
 ?n = 0
(dependent evars: ?X6 open,)
 *)
```

mikeshulman:
Is this operation of "(un)shelving" documented anywhere?

JasonGross:
It's documented [in the refman source](https://github.com/coq/coq/blob/9c24cecec3a7381cd924c56ca50c77a49750e2e5/doc/refman/RefMan-tac.tex#L4835) and [in the changelog](https://github.com/coq/coq/blob/47b8109321446ebf153807fe8a26151c7c0b003a/CHANGES#L137).  But these will not help you find it, probably, unless you look at the diffs of these every time we bump the Coq version.  The way I know about these things is that I subscribe to coq-commits@.

JasonGross:
Also, note: you can now refer to evars by name.  You can do things like `let x := type of ?n in ...` and `unify ?n foobar`.

JasonGross:
@mikeshulman: I've updated `STYLE.md`.

mikeshulman:
> This has the unfortunate side-effect
> that `simpl`/`cbn` is enough to "unfold" `compose`, and there's no way
> to prevent this.

Does that mean that `simpl` will turn `g o f` into `fun x => g (f x)`?!

JasonGross:
Yes.

JasonGross:
And it gets worse for long chains of composition.

mikeshulman:
Drat.  I'm not sure that is acceptable.  Is there any prospect of getting typeclass inference to backtrack into evar ordering, so we wouldn't need to do this?

JasonGross:
I summon @mattam82 to answer that question.

JasonGross:
We can always revert the "make it a notation" patch series...

mikeshulman:
Yeah, that's on the table too.

Why does this definition of the notation `compose` cause `simpl` to unfold it even partially applied, but the previous one doesn't?

JasonGross:
The previous one does, but it was only noticeable in `f o g o h`.

mikeshulman:
I need more explanation than that, sorry.

JasonGross:
Maybe what I mean by "unfold" will clarify this.  In the new system, before `simpl`,
```coq
f o g o h
```
is, (assuming I'm guessing associativity right)
```coq
(fun g' f' x => g' (f' x)) f ((fun g' f' x => g' (f' x)) g h)
```
`cbn`/`simpl` will perform beta reduction, giving
```coq
fun x => f (g (h x))
```
which has no corresponding notation.

In the old notation,
```coq
f o g o h
```
is, (assuming I'm guessing associativity right)
```coq
(fun x => f ((fun x => g (h x)) x))
```
`cbn`/`simpl` will perform beta reduction, giving
```coq
fun x => f (g (h x))
```

In the two-function case, the new notation gives `(fun g' f' x => g' (f' x)) f g`, while the old notation gives `(fun x => f (g x))`; the former can be beta-reduced, while the latter cannot.

Futhermore, this ability-to-beta-reduce (more precisely, being able to distinguish between various terms that beta-reduce to the same term) is what gives typeclass resolution the ability to solve the `isequiv_compose` instances.

mattam82:
I guess you're asking for typeclass resolution to try a different order of
resolution upon failure when it has interdependent constraints? I fear this
will make failures extremelly slow. Another option which Enrico mentioned
is to keep resolving them in their order of introduction in the term (order
of creation by the refiner), which has the advantage of being relatively
predictable too. What was the example problem again?

2015-01-05 10:08 GMT+01:00 Jason Gross <notifications@github.com>:

> Maybe what I mean by "unfold" will clarify this. In the new system, before
> simpl,
>
> f o g o h
>
> is, (assuming I'm guessing associativity right)
>
> (fun g' f' x => g' (f' x)) f ((fun g' f' x => g' (f' x)) g h)
>
> cbn/simpl will perform beta reduction, giving
>
> fun x => f (g (h x))
>
> which has no corresponding notation.
>
> In the old notation,
>
> f o g o h
>
> is, (assuming I'm guessing associativity right)
>
> (fun x => f ((fun x => g (h x)) x))
>
> cbn/simpl will perform beta reduction, giving
>
> fun x => f (g (h x))
>
> In the two-function case, the new notation gives (fun g' f' x => g' (f'
> x)) f g, while the old notation gives (fun x => f (g x)); the former can
> be beta-reduced, while the latter cannot.
>
> Futhermore, this ability-to-beta-reduce (more precisely, being able to
> distinguish between various terms that beta-reduce to the same term) is
> what gives typeclass resolution the ability to solve the isequiv_compose
> instances.
>
> —
> <https://github.com/HoTT/HoTT/pull/663#issuecomment-68682962>.
>



-- 
-- Matthieu

spitters:
LGTM

mikeshulman:
Okay, I think I see the problem.

In how many places would we need to call `isequiv_compose` explicitly without this change?  If we change the order and implicitness of its arguments to be more sensible (which I've been intending to do anyway), it might not be so bad.

Another thought: what if we define the notation `g 'o' f` twice, first as `fun x => g (f x)` and then as `compose g f`?  Would that have the effect that _writing_ `g o f` would be get parsed as `compose g f` (i.e. `(fun g' f' x => g' (f' x)) g f`), so that Coq could pick up the typeclass instances, but Coq would still be willing to _display_ `fun x => g (f x)` as `g o f`?

JasonGross:
> In how many places would we need to call `isequiv_compose` explicitly without this change?

I probably gave up at around a dozen last time I tried.

> If we change the order and implicitness of its arguments to be more sensible (which I've been intending to do anyway), it might not be so bad.

Perhaps.  Try it and see?  (I'll be travelling for the next month, and will not be on my computer much.)

> what if we define the notation `g 'o' f` twice, first as `fun x => g (f x)` and then as `compose g f`?

Yes, that will work, though not for things like `f o g o h`, which, under `simpl`, will still become `fun x => f (g (h x))`.  This was my "We could additionally define `g o f := (fun x => g (f x))` to change this, but this would result in identically looking goals which are really different." comment in `STYLE.md`.  I'm ambivalent about whether we should do this.  (Of course, we can define a separate notation, for each of, say, up to 15 functions, or something, and say that this pattern can be easily extended.)

JasonGross:
@mattam82, this is my [bug report turned feature request 3881](https://coq.inria.fr/bugs/show_bug.cgi?id=3881).

mikeshulman:
> We could additionally define `g o f := (fun x => g (f x))` to change this, but this would result in identically looking goals which are really different.

Ah, okay, now I understand that comment!  (-:

Maybe I should be more bothered by this than I am, but I don't find it particularly troublesome, because the identical-looking goals are still *judgmentally* equal, and in fact they are even equal modulo only beta-reduction (no unfolding of definitions needed).  Is there any situation in which that would matter?  It does matter for typeclass resolution, as we see, but is there any tactic the _user_ would be using to solve a goal that is unwilling to do beta-reduction?  For instance, `rewrite` is evidently willing to do beta-reduction, even though it isn't always willing to unfold definitions.  It would be pretty terrible if there were a tactic that would stop working after the insertion of a `simpl` that looks like a no-op.

JasonGross:
`rewrite` is often unwilling to do beta *expansion*.  For example, if you have `f o (g o h) o i`, and you have `g o h = k`, and you want to rewrite with this, you can do this before `simpl`, but not after.  (This will not look like a no-op, but a re-association.)  I would not be surprised if there are rewrites that stop working after this kind of simpl.  At least, before keyed unification.  I suspect even with keyed unification, too.

mikeshulman:
*throws up hands in disgust*

mikeshulman:
This sort of thing is the reason I still say that proof assistants are not yet ready for use by mainstream mathematicians.  Bah.

JasonGross:
Beta expansion is hard, and slow, because it's heuristic, but, yes, *throws up hands in disgust* is approximately how I feel.  (On the other hand, the `rewriteoA` tactics let us side-step this particular problem, I think.)

I'm inclined to go back to the having-a-separate-constant-for-compose, possibly seeing if we can get keyed unification to make it so we don't have to adjust the arguments to `compose` so much.

mikeshulman:
I really disliked having to `Local Arguments compose / .` and `simpl rewrite` all the time.  Moreover, it's one more of those little-ish things that accumulate to make a system incomprehensible to newcomers.  I'm more inclined to try giving `isequiv_compose` explicitly where needed.  At least when Coq fails to notice that a composite function is an equivalence, it's obvious how to give the missing proof.

mikeshulman:
(Also, I've been moving more towards a style that builds composite equivalences with `equiv_compose` directly, so that `isequiv_compose` isn't needed much.)

JasonGross:
Note: We will not only have to give `isequiv_compose` explicitly, but also give the two function arguments to `isequiv_compose` explicitly.

Is this newcomers to HoTT and Coq from math, or newcomers to HoTT from knowing Coq?  If the former, `rewriteoA` might be a good solution?

mikeshulman:
I'm much more concerned with newcomers from math.

When you talk about `rewriteoA` as a solution, what combination of other choices are you implying?

JasonGross:
Using `rewriteoA` should be able to replace `Arguments compose /` and `simpl rewrite`, should we decide to put the constant back.  It should also be able to handle composition both pre- and post- simpl, because it eliminates the need to do beta-expansion.

mikeshulman:
I don't think very many of the cases where I needed `Arguments compose /` and `simpl rewrite` were because of associativity.  I remember that after you defined `rewriteoA` I went looking for places where I could use it, but didn't find any.

JasonGross:
Another few arguments against the beta-reduced form: `refine (_ o f)` allows you to fill in a dependent function in the first slot.  The type-error messages from `refine (f o g)` are confusing and take some getting used to; they say things like "`g ?1` has type foo while it's expected to have type bar", or "cannot apply `f` to `g ?1`, `f` has type foo, expected a product".

mikeshulman:
I've never used `refine (f o g)`; I always `intro` the argument.  Wouldn't at least the first one be true as long as `compose` is a notation rather than a constant, regardless of whether it's beta-reduced or not?

JasonGross:
Yes.  These are, perhaps, arguments for making it a constant.

JasonGross:
I would like to decide on something, and merge it, so we can catch up to trunk again.  @mikeshulman, do you have preferences?  I would like to either revert back to having `compose` be a constant, or, perhaps, define it as it is in this PR, possibly with an additional notation for display.

mikeshulman:
Can you wait a few more days?  I'm busy preparing a talk to give next week, but hopefully soon I'll have time to experiment with explicit `isequiv_compose`.

mikeshulman:
Although, hm, I suppose this current PR is small enough and non-impacting enough that we could just merge it for now and then make a new PR for whatever fix for `compose` we decide on.  If that sounds good to you, feel free to merge it (I think we already had APOE).

JasonGross:
I will merge this, after rebasing and making sure travis is happy, and leave actually finding a sane solution for `compose` to #657.

(I have a sneaking suspicion that an alpha/beta release of Coq 8.5 will be announced next week at CoqPL at POPL.)

mikeshulman:
Sounds good.  An alpha/beta release is exciting.

PR #664, opened 2015-01-05
Also install hoqdep on `make install`

### LINKED ISSUE #637: `hoqdep` is not installed
`hoqdep` exists, but `make` does not install it.  Is this intentional, or an oversight?

(Not having `hoqdep` available makes it difficult to write intelligent makefiles for developments based over HoTT.  One can pass the HoTT library directory to `coqdep` explicitly, but this location will be system-dependent, and I can’t find a system-independent way to specify it.)

JasonGross:
This fixes #637


mikeshulman:
LGTM

Issue #654, opened 2014-12-27
Basic group theory


mikeshulman:
For something I'm currently trying to formalize, I need some very basic group theory (specifically, [Lagrange's theorem](https://en.wikipedia.org/wiki/Lagrange%27s_theorem_%28group_theory%29)).  I could just prove the special case of it that I need, but as time goes on we'll undoubtedly need more group theory.  Is there any existing library including group theory (with a compatible license) that we could steal from?  Or is the HoTT approach to things like hsets, subsets, images, and quotients so different from any other that that wouldn't save any appreciable work?

vladimirias:
Basic algebraic hierarchy is formalized in UniMath. There are the concepts of a group and of a subgroup. There is a concept of a finite type (finset) and the function that assigns to a finset its cardinality in nat. Therefore it is very easy to formulate the Lagrange theorem. There are various theorems about the cardinalities of dependent sums of families of finite types that should make the proof of this theorem easy.

mikeshulman:
I have Lagrange proved now [here](https://github.com/mikeshulman/HoTT/blob/groups/theories/Groups.v#L292), using a similar library of theorems about cardinalities of finite sets motivated by the one in UniMath (which I recall you demonstrating 5 years ago (!) at CMU).  I used the definition of "group" from #661 for simplicity, but it should be easy to adapt to any definition of group.

JasonGross:
> Given cumulativity, is there really any point to having a separate `Unit` in each universe?

Universe variance isn't judgmental, at the moment.  So we don't get `IsTrunc@{i} -2 Unit` from `IsTrunc@{Set} -2 Unit`, and this causes many, many problems with theorems that expect to be able to put arbitrary below the universe of `IsTrunc`.

`DecidablehProp` would be fine, I guess.  (An instance of `Decidable` would need to be proven anew for each definition, by recursion.)

We can always define `sum_of_bool : forall b : bool, b = true + b = false`

mikeshulman:
What do you mean by "put arbitrary below"?

I don't think we'd need to prove `Decidable` once per each definition; we could just make each definition recursive with values in `DecidablehProp`, supplying `DecidableUnit` or `DecidableEmpty` as appropriate (which I'd be tempted to try calling `True` and `False`...).  We could even make a coercion from `DecidablehProp` to `Bool` so that we could leave the `dec` out of `if...then...else`.

[...]

PR #665, opened 2015-01-05
Nat Properties

JasonGross:
Define `<`, `<=`, `>`, `>=`, classify the path space of `nat`, and prove that `nat` is an hSet.

Also make a module for making `is_true` a coercion.  This is one way to start off some definitions of #654.

### DIFF
--- a/theories/Types/Nat.v
+++ b/theories/Types/Nat.v
@@ -1,7 +1,74 @@
 (* -*- mode: coq; mode: visual-line -*- *)
 (** * Theorems about the natural numbers *)
 
+Require Export Coq.Init.Peano.
 Require Import HoTT.Basics.
+Require Import HoTT.Types.Bool.
+
+(** We reopen these scopes so they take precedence over nat_scope; otherwise, now that we have [Coq.Init.Peano], we'd get [* : nat -> nat -> nat] rather than [* : Type -> Type -> Type]. *)
+Global Open Scope type_scope.
+Global Open Scope core_scope.
+
+Local Open Scope equiv_scope.
 
 Scheme nat_ind := Induction for nat Sort Type.
 Scheme nat_rec := Minimality for nat Sort Type.
+
+(** Much of the layout of this file is adapted from ssreflect *)
+
+(** ** Equality *)
+(** *** Boolean equality and its properties *)
+
+Fixpoint code_nat (m n : nat) {struct m} :=
+  match m, n with
+    | 0, 0 => true
+    | m'.+1, n'.+1 => code_nat m' n'
+    | _, _ => false
+  end.
+
+Infix "=n" := code_nat (at level 70, no associativity) : nat_scope.
+
+Fixpoint idcode_nat {n} : (n =n n) = true :=
+  match n as n return (n =n n) = true with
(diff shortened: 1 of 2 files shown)


JasonGross:
Note: I have two branches that [add some more `nat` properties](https://github.com/JasonGross/HoTT/compare/HoTT:master...JasonGross:nat-more-props) and [implement chain categories `[n]` on top of that](https://github.com/JasonGross/HoTT/compare/HoTT:master...JasonGross:chain-category).  The extra properties should, I hope, be largely independent of whether we go via `is_true` or if we jump directly to the `DecidablehProp` route.

mikeshulman:
Looks good.  Although `(n =n m) = true` and `path_b_true__of__is_true` make me want even more to at least try `DecidablehProp`.

JasonGross:
The `(n =n m) = true` is primarily because the definition of `hProp` depends on all of `Types/`, so we don't have `hProp` available here.

theories/Types/Nat.v
```diff
@@ -1,7 +1,74 @@
...
 
 Scheme nat_ind := Induction for nat Sort Type.
 Scheme nat_rec := Minimality for nat Sort Type.
+
+(** Much of the layout of this file is adapted from ssreflect *)
+
+(** ** Equality *)
+(** *** Boolean equality and its properties *)
+
+Fixpoint eqn (m n : nat) {struct m} :=
```
mikeshulman: `eqn` seems a potentially confusing name for this, especially in global scope.  How about something like `code_nat` (since it's essentially the codes fibration for the encode-decode)?

theories/Types/Nat.v
```diff
@@ -1,7 +1,74 @@
...
+  match m, n with
+    | 0, 0 => true
+    | m'.+1, n'.+1 => eqn m' n'
+    | _, _ => false
+  end.
+
+Infix "=n" := eqn (at level 70, no associativity) : nat_scope.
+
+Fixpoint eqn_n {n} : (n =n n) = true :=
```
mikeshulman: This likewise I would be inclined to call something like `idcode_nat`.

theories/Types/Nat.v
```diff
@@ -1,7 +1,74 @@
...
+
+Global Instance decidable_paths_nat : DecidablePaths nat
+  := fun n m => @decidable_equiv _ _ (@path_nat n m) _ _.
+
+Corollary hset_nat : IsHSet nat.
+Proof.
+  exact _.
+Defined.
+
+(** ** Integer ordering, and its interaction with the other operations. *)
```
mikeshulman: Should "integer" be "natural number" here?

mikeshulman:
Modulo those comments, LGTM.

spitters:
LGTM. I agree with @mikeshulman s suggestions.

theories/Types/Nat.v
JasonGross: Yes, I was copying from ssr without thinking.  Fixed.

JasonGross:
Comments fixed.  I will merge when travis is happy, unless there are objections.

mikeshulman:
Go for it.

PR #666, opened 2015-01-05
Update sigma notations

JasonGross:
Now we won't see any more `{ x : _ | ... }` in the output; the type will
always appear.


mikeshulman:
That's nice!  How did you know that this change would have that effect?

JasonGross:
Because https://github.com/coq/coq/commit/86a1787cf9ae815ee8b1f6fd7f39ac615da031a4 told me so ;-)

PR #667, opened 2015-01-06
Add a separate file of theorems about [nat]

JasonGross:
We need to depend on [TruncType], so we can't live in [Types/].

Note: This is enough to build chain categories and the category of simplicial (h)sets and of semisimplicial (h)sets.

### DIFF
--- a/Makefile_targets.mk
+++ b/Makefile_targets.mk
@@ -53,6 +53,7 @@ CORE_VFILES = \
 	$(srcdir)/theories/Extensions.v \
 	$(srcdir)/theories/HSet.v \
 	$(srcdir)/theories/UnivalenceImpliesFunext.v \
+	$(srcdir)/theories/Spaces/Nat.v \
 	$(srcdir)/theories/Tactics.v \
 	$(srcdir)/theories/Tactics/RewriteModuloAssociativity.v \
 	$(srcdir)/theories/Tactics/BinderApply.v \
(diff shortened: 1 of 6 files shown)


mikeshulman:
Looks fine modulo the fact that I want to use `DecidablehProp` instead of `is_true`.

JasonGross:
`DecidablehProp` can't be defined until after `Types/`, because `hProp` is defined in `TruncType`.  I can move `code_nat` and everything below it from `Types/` into `Spaces/`.  If we don't want to do that, I'm not sure what to do.

In any case, the only things that will have to be changed in `Spaces/Nat` to get things working with `DecidablehProp` are `cast` and `cast'`.

JasonGross:
Do you want to have this merged, and then change it to `DecidablehProp` after, or do you want the change (whatever it's going to be) to be made first?

mikeshulman:
We can go ahead and merge it first.

I've been considering where to define `DecidablehProp`.  I think one option would be to move the definition of the type `TruncType` (and `hProp`) into `Basics/Trunc` (but keep all the theorems about it, which need `Types`, in `TruncType.v`).  Another would be, as you say, to essentially merge `Types/Nat` into `Spaces/Nat` &mdash; which raises an interesting question: is `nat` a "basic type former" that deserves placement in `Types` at all, or is it just a "particular type" that we study in `Spaces`?  Where do we draw the line?

PR #669, opened 2015-01-06
Miscellaneous sigma lemmas

JasonGross:
Needed for classifying when a full subprecategory is actually a category.

### DIFF
--- a/theories/Types/Sigma.v
+++ b/theories/Types/Sigma.v
@@ -367,14 +367,18 @@ Proof.
                                             (fun x y => ((g (f^-1 x))^-1 ((eisretr f x)^ # y)))) _ _);
   intros [x y].
   - refine (path_sigma' _ (eisretr f x) _); simpl.
-    rewrite (eisretr (g (f^-1 x))).
-    apply transport_pV.
+    abstract (
+        rewrite (eisretr (g (f^-1 x)));
+        apply transport_pV
+      ).
   - refine (path_sigma' _ (eissect f x) _); simpl.
     refine ((ap_transport (eissect f x) (fun x' => (g x') ^-1)
                           (transport Q (eisretr f (f x)) ^ (g x y)))^ @ _).
-    rewrite transport_compose, eisadj, transport_pV.
-    apply eissect.
-Qed.
+    abstract (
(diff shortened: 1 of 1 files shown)


mikeshulman:
We do already have `path_sigma_hprop`.

theories/Types/Sigma.v
```diff
@@ -358,23 +358,43 @@ Defined.
...
-Global Instance isequiv_functor_sigma `{P : A -> Type} `{Q : B -> Type}
-         `{IsEquiv A B f} `{forall a, @IsEquiv (P a) (Q (f a)) (g a)}
-: IsEquiv (functor_sigma f g) | 1000.
-Proof.
-  refine (isequiv_adjointify (functor_sigma f g)
-                             (functor_sigma (f^-1)
-                                            (fun x y => ((g (f^-1 x))^-1 ((eisretr f x)^ # y)))) _ _);
-  intros [x y].
-  - refine (path_sigma' _ (eisretr f x) _); simpl.
+(** We define the section and retraction separately so they can be opaque, while the equivalence is transparent. *)
```
mikeshulman: Another, less heavy-looking, possibility (which I had already done in an unpushed branch) is to wrap the rewrites in `abstract`.

JasonGross:
Updated to use `path_sigma_hprop`.

theories/Types/Sigma.v
JasonGross: Would you prefer that?  I was under the assumption that people would frown on me replacing periods with semicolons in core, but I'm happy to do that instead.

spitters: I'd prefer `abstract` too.

spitters:
LGTM

theories/Types/Sigma.v
JasonGross: Done

mikeshulman:
I was thinking of just abstracting the `rewrite; apply` that finishes each subproof; abstractin the whole things are indeed more semicolons than I would prefer.  Do you have a reason to want the whole subproofs to be opaque?

JasonGross:
> I was thinking of just abstracting the `rewrite; apply` that finishes each subproof

Done.

PR #670, opened 2015-01-06
Port the category of simplicial sets from catdb

JasonGross:
This is on top of #667 and #669.

### DIFF
--- a/theories/categories/ChainCategory.v
+++ b/theories/categories/ChainCategory.v
@@ -1,5 +1,6 @@
 (** * The category ω of (ℕ, ≤), and the chain categories [[n]] *)
 Require Import Category.Core Category.Subcategory.Full.
+Require Import Category.Sigma.Univalent.
 Require Import Category.Morphisms Category.Univalent Category.Strict.
 Require Import HoTT.Basics.Trunc HoTT.Types.Nat HoTT.Types.Bool HoTT.TruncType HoTT.Spaces.Nat.
 Import BoolSortCoercion.
@@ -62,5 +63,8 @@ Module Export Univalent.
     { intro m; apply leq_antisym; apply m. }
   Defined.
 
-  (** TODO: Derive [IsCategory (chain n)] from a lemma about sigma categories. *)
+  Definition iscategory_chain {n} : IsCategory [n].
+  Proof.
+    exact _.
+  Defined.
 End Univalent.
--- a/theories/categories/ChainCategory.v
+++ b/theories/categories/ChainCategory.v
@@ -1,0 +1,66 @@
+(** * The category ω of (ℕ, ≤), and the chain categories [[n]] *)
+Require Import Category.Core Category.Subcategory.Full.
+Require Import Category.Morphisms Category.Univalent Category.Strict.
+Require Import HoTT.Basics.Trunc HoTT.Types.Nat HoTT.Types.Bool HoTT.TruncType HoTT.Spaces.Nat.
+Import BoolSortCoercion.
+
+Set Universe Polymorphism.
+Set Implicit Arguments.
+Generalizable All Variables.
+Set Asymmetric Patterns.
+
+Module Export Core.
+  Definition omega : PreCategory
+    := @Build_PreCategory
+         nat
+         leq
+         leq_refl
+         (fun x y z p q => leq_trans _ _ _ q p)
+         (fun _ _ _ _ _ _ _ => path_ishprop _ _)
+         (fun _ _ _ => path_ishprop _ _)
+         (fun _ _ _ => path_ishprop _ _)
+         _.
+
+  Definition chain (n : nat) : PreCategory
+    := { m : omega | m <= n }%category.
+
+  Module Export ChainCategoryCoreNotations.
+    Notation "[ n ]" := (chain n) : category_scope.
+  End ChainCategoryCoreNotations.
+End Core.
+
+Module Export Notations.
+  Include ChainCategoryCoreNotations.
+End Notations.
+
+Module Utf8.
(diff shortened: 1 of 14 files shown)


mikeshulman:
Nice!  One terminological suggestion: you might want to mention the notion of [pseudomonic functor](http://ncatlab.org/nlab/show/pseudomonic+functor) somewhere.  It seems that part of what you're proving is that univalence lifts along suitable such functors, is that right?

Can you explain what a "chain category" is (and perhaps add such an explanation in comments)?  I don't quite follow it from the code.

JasonGross:
Added documentation for pseudomonic functors and about the meaning of "chain" categories.

mikeshulman:
Ok, so `ChainCategory.v` is just defining a few particular categories that happen to be chains.  And

```coq
Definition simplex_category := @Build_PreCategory nat Functor ...
```

says that the objects of `simplex_category` are natural numbers and the morphism are functors, obtained by silently regarding those natural numbers as categories via `chain`?  That wasn't clear to me at first.  Is `chain` declared as a coercion somewhere that makes this work?

theories/categories/ChainCategory.v
```diff
@@ -0,0 +1,90 @@
...
+         (fun _ _ _ => path_ishprop _ _)
+         (fun _ _ _ => path_ishprop _ _)
+         _.
+
+  (** *** [[n]], a linear order on a finite set with [n + 1] elements *)
+  (** Using [n + 1] elements allows us to agree with the common
+      definition of an [n]-simplex, where a 0-simplex is a point, and
+      a 1-simplex has two end-points, etc. *)
+  Definition chain (n : nat) : PreCategory
+    := { m : omega | m <= n }%category.
```
JasonGross: @spitters [commented](https://github.com/JasonGross/HoTT/commit/f66703e0e868e0872b9acd17df99e3358a32bc88#commitcomment-9174957) (transposed here so that it's not lost if I rebase):
> Would it be natural to generalize this definition to decidable total(?) orders?

All total orders on a finite set are isomorphic, right?  It might be, but I'd rather leave it like this for now.  I can add a comment about generalizing it, if you think it worth it?

theories/categories/SemiSimplicialSets.v
```diff
@@ -0,0 +1,44 @@
...
+    Context `{Funext}.
+    (** Quoting David Spivak:
+
+        Consider the subcategory of [Δ] with the same objects (wide)
+        but only injective morphisms.  If we call that [Γ] (which is
+        nonstandard), then semi-simplicial sets (also a non-standard
+        term) are [Fun(Γᵒᵖ, Set)]. Define the obvious inclusion [Γ ->
+        Δ], which we will use to make simplicial sets without having
+        to worry about "degneracies". *)
+
```
JasonGross: @spitters [commented](https://github.com/JasonGross/HoTT/commit/ebf2699a725b2e1f8cff7159c01fec5b7d8dc280#commitcomment-9174996) (transposed here in case I rebase):
> semi-simplicial sets seems standard (?)

*shrug*  Was it recently made standard?  I'm simply quoting David, here.  I can add a `[sic]`, if you think that's better?

JasonGross:
@mikeshulman, I've added a comment about the simplex category

theories/categories/SemiSimplicialSets.v
mikeshulman: It's becoming pretty standard.  Older topologists tend to object that "semi-simplicial set" used to mean what is now called "simplicial set", but that hasn't been true for quite some time, so it doesn't carry much weight with the younger generation, and this meaning of "semi-simplicial" is consistent with other uses of "semi" (semigroup, semicategory).

mikeshulman:
Thanks!  Sorry for being dense.  This LGTM.

spitters:
LGTM. Feel free to merge when Travis is happy.

theories/categories/SemiSimplicialSets.v
spitters: It's in the nlab, hence standard :-)
http://ncatlab.org/nlab/show/semi-simplicial+set

theories/categories/ChainCategory.v
spitters: Do we need to the set to be finite? It looks like the blurb you quoted from
wikipedia was general.
It might be good to record it in the documentation, but let's leave the
generalization in Coq for a later time.

JasonGross:
Updated with comments about your suggestions, @spitters, and rebased against master.

theories/categories/SemiSimplicialSets.v
peterlefanulumsdaine: On Tue, Jan 6, 2015 at 9:26 PM, Jason Gross <notifications@github.com>
wrote:

> @spitters <https://github.com/spitters> commented
> <https://github.com/JasonGross/HoTT/commit/ebf2699a725b2e1f8cff7159c01fec5b7d8dc280#commitcomment-9174996>
> (transposed here in case I rebase):
>
> semi-simplicial sets seems standard (?)
>
> *shrug* Was it recently made standard? I'm simply quoting David, here. I
> can add a [sic], if you think that's better?
>
There is a useful discussion of the naming conventions for these in [this
mathoverflow comment thread](http://mathoverflow.net/a/57655/2273), with if
not quite references then at least hints at them.

Rough summary: historically, “semi-simplicial” has meant quite a few
things, but it seems to be getting reasonably established nowadays with
this meaning.

–p.

theories/categories/ChainCategory.v
mikeshulman: Well, really all you need is an arbitrary poset...

Issue #671, opened 2015-01-06
Projecting truncatedness?


JasonGross:
I know that the following two theorems are true:
```coq
forall A P, Contr A -> Contr { x : A | P x } -> forall x, Contr (P x)
forall A P, IsHProp A -> IsHProp { x : A | P x } -> forall x, IsHProp (P x)
forall A P, IsHSet A -> IsHSet { x : A | P x } -> forall x, IsHSet (P x)
```
Does this generalize?  Is there a proof of this somewhere in the library already?

mikeshulman:
It generalizes to any reflective subuniverse, but it wasn't in the library yet: #672

PR #672, opened 2015-01-06
Reflective subuniverses are closed under fibers

mikeshulman:


JasonGross:
Nice!  Thanks!

I would not have discovered this if it were there, though.  Do you want to add convenience lemmas for `SearchAbout` somewhere that specialize this to `IsTrunc` (and perhaps also `Contr`, `IsHProp`, `IsHSet`)?

Also, mention somewhere in the commit or description that this closes #671?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -678,6 +678,32 @@ Section Reflective_Subuniverse.
...
+          refine (ap f (O_rec_beta pr1 x) @ _).
+          exact (x.2).
+      - intros [a p]; refine (path_sigma' _ _ _).
+        + exact (O_rec_beta pr1 (a;p)). 
+        + refine (ap (transport _ _) (O_indpaths_beta _ _ _ _) @ _); simpl.
+          refine (transport_paths_Fl _ _ @ _).
+          apply concat_V_pp.
+    Defined.
+
+    Definition inO_unsigma {A : Type} (B : A -> Type)
```
JasonGross: Can this be made `Hint Immediate`, or does that break things?

mikeshulman:
Yeah, this is a difficult problem.  Surely we don't want to add renamed copies of *all* the theorems about reflective subuniverses and modalities that specialize them to `IsTrunc`.  What about a comment in `Trunc` or a remark in `STYLE` suggesting that if you want a lemma about `IsTrunc` or `Trunc` and can't find it, you should try generalizing it to reflective subuniverses or modalities and looking for that?

JasonGross:
I think I would be satisfied with a comment in `STYLE.md` (perhaps also `Trunc`?), also mentioning my freshly created [feature request / bug 3904](https://coq.inria.fr/bugs/show_bug.cgi?id=3904).

mikeshulman:
Added.

JasonGross:
LGTM modulo travis, and assuming you check that the `Hint Immediate` hasn't slowed anything down.  (If you want to let travis do the timing, once travis finishes, you can search for `Time     | File Name` in the log; the last one was 4m40s total, with a minute for Adjoint/Pointwise, 21s for hit/V, and under 15s for everything else.)

mikeshulman:
No change in timing on my machine.

spitters:
Nice. LGTM.

PR #674, opened 2015-01-06
Centers of BAut(X), BAut(Bool), and an incoherent idempotent

mikeshulman:
Here we characterize the center and 2-center of BAut(X), and use those to construct an example of a quasi-idempotent that cannot be extended to a coherent one (though, by the general splitting construction, its underlying pre-idempotent can).

### DIFF
--- a/theories/Spaces/BAut.v
+++ b/theories/Spaces/BAut.v
@@ -1,5 +1,6 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 Require Import HoTT.Basics HoTT.Types.
+Require Import Constant Factorization UnivalenceImpliesFunext.
 Require Import Modalities.Modality hit.Truncations hit.Connectedness.
 Import TrM.
 
@@ -8,8 +9,22 @@ Open Scope equiv_scope.
 
 (** * BAut(X) *)
 
+(** ** Basics *)
+
 (** [BAut X] is the type of types that are merely equivalent to [X]. *)
 Definition BAut (X : Type) := { Z : Type & merely (Z = X) }.
+
+(** Equivalently, [BAut X] is the (-1)-image of the classifying map [1 -> Type] of [X]. *)
+
+Definition equiv_baut_image_unit X
+: BAut X <~> image (Tr -1) (unit_name X).
+Proof.
+  unfold BAut, image; simpl.
+  refine (equiv_functor_sigma' (equiv_idmap Type) _); intros Z; simpl.
+  apply equiv_O_functor; unfold hfiber.
+  refine (equiv_compose' (equiv_inverse (equiv_contr_sigma _)) _).
+  apply equiv_path_inverse.
+Defined.
 
 (** It is canonically pointed by [X] itself. *)
 Global Instance ispointed_baut X : IsPointed (BAut X)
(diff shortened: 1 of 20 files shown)


theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+:= merely_path_is0connected (BAut X) Z Z'.
+
+(** The following tactic, which applies when trying to prove an hprop, replaces all assumed elements of [BAut X] by [X] itself. *)
+Ltac baut_reduce :=
+  progress repeat
+    match goal with
+      | [ Z : BAut ?X |- _ ]
+        => let Zispoint := fresh "Zispoint" in
+           assert (Zispoint := merely_path_baut (point (BAut X)) Z);
+           strip_truncations;
```
JasonGross: Why `strip_truncations` here?  (And does it deserve a comment?)

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+  intros g; simpl.
+  refine (equiv_compose' (equiv_moveR_transport_V _ _ _ _) _).
+  refine (equiv_compose' (equiv_path_inverse _ _) _).
+  apply equiv_concat_l, transport2.
+  symmetry; apply (eissect (equiv_path X X)).
+Defined.
+
+(** This implies that if [X] is a set, then the center of [BAut X] is the set of automorphisms of [X] that commute with every other automorphism (i.e. the center, in the usual sense, of the group of automorphisms of [X]). *)
+
+Definition center_baut `{Univalence} X `{IsHSet X} :
```
JasonGross: Style Nit: `:` on the beginning of the next line

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+Defined.
+
+(** Similarly, if [X] is a 1-type, we can characterize the 2-center of [BAut X]. *)
+
+(** Coq is too eager about unfolding some things appearing in this proof. *)
+Section Center2BAut.
+Opaque equiv_path_equiv.
+Opaque equiv_path2_universe.
+
+Definition center2_baut `{Univalence} X `{IsTrunc 1 X} :
```
JasonGross: Style nit: `:` on the beginning of the following line

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+  { f : forall x:X, x=x & forall (g:X<~>X) (x:X), ap g (f x) = f (g x) }
+  <~> (forall Z:BAut X, (idpath Z) = (idpath Z)).
+Proof.
```
JasonGross: Any reason to not indent the `Center2BAut` section?

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+  baut_reduce.
+  exact (ap (path_sigma_hprop _ _) path_universe_1
+            @ path_sigma_hprop_1 _).
+Defined.
+
+(** Similarly, if [X] is a 1-type, we can characterize the 2-center of [BAut X]. *)
+
+(** Coq is too eager about unfolding some things appearing in this proof. *)
+Section Center2BAut.
+Opaque equiv_path_equiv.
```
JasonGross: I think `Local Arguments ... : simpl never` is preferable to `Opaque` when it works, because `Opaque` breaks things like `autounfold`, and also creates a weird disparity between the kernel's view (that two things are convertible) and `Ltac`'s view (that things are opaque).  See also [`Local Strategy`](https://coq.inria.fr/distrib/current/refman/Reference-Manual008.html#hevea_command188), which allows you to control the order in which things are unfolded by, e.g., unification.

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+Proof.
+  refine (equiv_compose'
+            (equiv_functor_forall'
+               (P := fun Z => idpath Z.1 = idpath Z.1)
+               (equiv_idmap (BAut X))
+               (fun Z => equiv_compose' (equiv_concat_lr _ _)
+                                        (equiv_ap (equiv_path_sigma_hprop Z Z)
+                                                  (idpath Z.1) (idpath Z.1)))) _).
+  1:symmetry; apply path_sigma_hprop_1.
+  1:apply path_sigma_hprop_1.
```
JasonGross: Why are these `1:` (which is technically a no-op)?  If they solve the first goals, you should wrap each of these two lines in a pair of curly braces, or at least change `apply path_sigma_hprop_1.`s to `solve [ apply path_sigma_hprop_1  ].`s, so that Coq checks that the relevant goal has been solved.

theories/Spaces/BAut.v
```diff
@@ -25,3 +40,230 @@ Proof.
...
+  simpl.
+  change (equiv_idmap X == equiv_idmap X) in f.
+  refine (equiv_concat_lr _ _).
+  - refine (_ @ (path2_universe_postcompose_idmap f g)^).
+    abstract (rewrite !whiskerR_pp, !concat_pp_p; reflexivity).
+  - refine (path2_universe_precompose_idmap f g @ _).
+    abstract (rewrite !whiskerL_pp, !concat_pp_p; reflexivity).
+Defined.
+
+(** Once again we compute it on the identity.  In this case it seems to be unavoidable to do some [simpl]ing, making this proof somewhat slower. *)
```
JasonGross: Is `cbn` faster than `simpl`?  Is there any way to wait to `simpl` until you are only `simpl`ing the `equiv_fun` or something?

JasonGross:
Could you include a timing diff for this PR on one of the commits, since you mention some of the proofs being slow?

theories/Spaces/BAut/Bool.v
```diff
@@ -0,0 +1,494 @@
...
+
+(** ** Equivalence types *)
+
+(** As soon as an element of [BAut Bool] is inhabited, it is (purely) equivalent to [Bool].  (Of course, every element of [BAut Bool] is *merely* inhabited, since [Bool] is.)  In fact, it is equivalent in two canonical ways.
+
+First we define the function that will be the equivalence. *)
+Definition inhab_baut_bool_from_bool (t : Bool)
+           (Z : BAut Bool) (z : Z)
+: Bool -> Z
+  := fun b => if t then
```
JasonGross: Should we define `xorb` and do `if xorb t b then negb_baut_bool Z z else z`?

theories/Spaces/BAut/Bool.v
```diff
@@ -0,0 +1,494 @@
...
+  - refine (isequiv_homotopic idmap _); intros []; reflexivity.
+Defined.
+
+Definition equiv_inhab_baut_bool_bool (t : Bool)
+           (Z : BAut Bool) (z : Z)
+: Bool <~> Z
+  := BuildEquiv _ _ (inhab_baut_bool_from_bool t Z z) _.
+
+Definition path_baut_bool_inhab (Z : BAut Bool) (z : Z)
+: (point (BAut Bool)) = Z.
```
JasonGross: Nit: Why parens around `point (BAut Bool)`?

theories/Spaces/BAut/Bool.v
```diff
@@ -0,0 +1,494 @@
...
+Import TrM.
+Require Import Spaces.BAut.
+
+Open Scope path_scope.
+Open Scope equiv_scope.
+
+(** * BAut(Bool) *)
+
+Section AssumeUnivalence.
+Context `{Univalence}.
```
JasonGross: Style nit-ish: I have a mild preference for indenting sections, even when they span the entire file; following this convention allows one to tell at a glance when definitions might depend on extra hypotheses or assumptions (if it's not indented, it's stand-alone).

theories/Spaces/BAut/Bool/IncoherentIdempotent.v
```diff
@@ -0,0 +1,56 @@
...
+  Let s := retract_sect (splitting_preidem_retractof_qidem (preidem_idmap (BAut (BAut Bool)))).
+  Let r := retract_retr (splitting_preidem_retractof_qidem (preidem_idmap (BAut (BAut Bool)))).
+  Let issect := retract_issect (splitting_preidem_retractof_qidem (preidem_idmap (BAut (BAut Bool)))) : r o s == idmap.
+
+  (** Since the space of splittings of the identity pre-idempotent is contractible, nontriviality of this 2-central element implies that not every quasi-idempotence witness of the identity is recoverable from its own splitting. *)
+  Definition splitting_preidem_notequiv_qidem_baut_baut_bool
+  : ~ (s o r == idmap).
+  Proof.
+    intros oops.
+    pose (p := oops nontrivial_qidem_baut_baut_bool); clearbody p.
```
JasonGross: N.B.  You won't need to `clearbody p`, if you use `assert` instead of `pose`.

theories/Types/Bool.v
```diff
@@ -67,12 +69,26 @@ Section BoolDecidable.
...
        | false => true_ne_false
      end.
 
+(** And conversely, if two elements of [Bool] are unequal, they must be related by [negb]. *)
+Definition negb_ne {b1 b2 : Bool}
+: (b1 <> b2) -> (b1 = negb b2).
+Proof.
+  destruct b1, b2.
+  - intros oops; case (oops idpath).
+  - intros _; reflexivity.
```
JasonGross: N.B. I believe `reflexivity` works without needing to `intros` first.

theories/Types/Bool.v
```diff
@@ -149,4 +165,30 @@ Section EquivBoolEquiv.
...
+
+  (** But, obviously, not both. *)
+  Definition idmap_bool_ne_negb : equiv_idmap Bool <> equiv_negb.
+  Proof.
+    intros oops.
+    exact (true_ne_false (ap10_equiv oops true)).
+  Defined.
+
+  (** In particular, every pair of automorphisms of [Bool] commute with each other. *)
+  Definition abelian_aut_bool `{Funext} (e1 e2 : Bool <~> Bool)
```
JasonGross: N.B. This does not need `Funext`:
```coq
Ltac case_path b :=
  let b' := fresh in
  pose b as b';
    generalize (idpath : b' = b);
    case b;
    subst b';
    let H := fresh in
    intro H.

Definition abelian_aut_bool (e1 e2 : Bool <~> Bool)
: e1 o e2 == e2 o e1.
Proof.
  intro b.
  generalize (eval_bool_isequiv e1).
  generalize (eval_bool_isequiv e2).
  case_path (e1 true).
  all:case_path (e2 true).
  all:simpl; intros.
  all:destruct b.
  all:repeat match goal with
               | _ => reflexivity
               | [ H : _ = ?f ?x |- context[?f ?x] ] => case H
               | [ H : ?f ?x = _ |- context[?f ?x] ] => apply symmetry in H; case H
             end.
Defined.
```

Perhaps we should include both proofs?

theories/Types/Forall.v
```diff
@@ -198,6 +198,19 @@ Proof.
   - intros f g; apply (trunc_equiv _ (apD10 ^-1)).
 Defined.
 
+(** TODO: This is also in branch lex, so watch out for merging issues. *)
```
JasonGross: This comment will, presumably, be removed by whichever branch is merged second?

theories/Types/Sum.v
```diff
@@ -220,6 +220,19 @@ Defined.
 Global Instance hset_sum `{HA : IsHSet A, HB : IsHSet B} : IsHSet (A + B) | 100
   := @trunc_sum -2 A HA B HB.
 
+(** Sums don't preserve hprops in general, but they do for disjoint sums. *)
+
+Definition ishprop_sum A B `{IsHProp A} `{IsHProp B}
```
JasonGross: Perhaps this should be an instance?  (I think typeclass search will automatically pick up `A -> B -> Empty` if it's in the context.)

theories/Types/Sum.v
```diff
@@ -220,6 +220,19 @@ Defined.
...
+: (A -> B -> Empty) -> IsHProp (A + B).
```
JasonGross: This theorem can be generalized to `IsTrunc n` for all `n`, right?  Or is it already true for `n >= 0` without the extra assumption?

theories/Types/Universe.v
```diff
@@ -162,6 +223,160 @@ Definition transport_path_universe_Vp `{Funext}
...
+  = path_universe_compose g f1
+    @ whiskerL (path_universe g) (path2_universe p)
+    @ (path_universe_compose g f2)^.
+>>
+
+but I haven't managed to prove them yet.  Fortunately, for existing applications what we actually need is the following rather different-looking version that applies only when [f1] and [f2] are identities. *)
+
+(** Coq is too eager about unfolding [equiv_path_equiv] in the following proofs, so we tell it not to.  We go into a section in order to limit the scope of the [Opaque] command. *)
+Section OpaquePathEquiv.
+Opaque equiv_path_equiv.
```
JasonGross: Same comment about `simpl never` as above.

theories/Types/Universe.v
```diff
@@ -162,6 +223,160 @@ Definition transport_path_universe_Vp `{Funext}
...
+
+Definition path2_universe_postcompose_idmap `{Funext}
```
JasonGross: Should this section be indented?

JasonGross:
Done commenting.  (Responses to all of the comments will likely result in an ack.)

theories/Spaces/BAut.v
mikeshulman: To turn `merely_path_baut` into an actual path.  But it should really just `Trunc_ind` on that particular path; I'll change it.

theories/Spaces/BAut.v
mikeshulman: I tried `simpl never` and it doesn't work with the current proof scripts.  Possibly it would work with minor modifications, but I don't have time to figure that out right now.

theories/Spaces/BAut.v
mikeshulman: Yes, `cbn` is faster, thanks.

theories/Spaces/BAut/Bool.v
mikeshulman: I don't think it's worth it here, but if someone else wants to define `xorb` I won't complain.

theories/Spaces/BAut/Bool.v
mikeshulman: Readability.

theories/Spaces/BAut/Bool.v
mikeshulman: Yeah, that's fair.  I'm sometimes lazy about it because I don't like to highlight a whole file and tell PG to re-indent it, because sometimes PG gets the indentation wrong (in my opinion), so it's time-consuming to re-indent a whole file.

theories/Types/Forall.v
mikeshulman: Whichever is merged first, I'll rebase the other.

theories/Types/Sum.v
mikeshulman: I didn't think it was the sort of thing that would tend to be in the context, rather the one that'd be left over as a goal.  But it probably doesn't do any harm to make it an instance.

theories/Types/Universe.v
mikeshulman: Same reply as above.  (-:

theories/Spaces/BAut.v
mikeshulman: Sorry, will do curly braces.

theories/Spaces/BAut.v
JasonGross: My guess is that replacing `simpl` with `cbn` would be sufficient (unless the relevant lines are not the `simpl` lines).

theories/Types/Bool.v
mikeshulman: Thanks for pointing that out.  I've found an intermediate proof, not using funext, which I prefer to both existing ones; we can just separate out the homotopy component of `equiv_bool_aut_bool` and apply it directly.

theories/Spaces/BAut.v
mikeshulman: It is; thanks.

mikeshulman:
I think I've addressed all your comments.  Thanks for them all!  The last commit has a timing diff.

JasonGross:
Note: where you do `unfold f, g; clear f g`, `subst f g` should do the same thing.  Feel free to make this change, or not, as you please.

JasonGross:
LGTM, with or without my one "Note".

mikeshulman:
Rebased against master.

mikeshulman:
Rebased.

PR #676, opened 2015-01-09
Pointed types and the loop-suspension adjunction

mikeshulman:
If anyone can manage to prove that the loop-suspension adjunction isomorphism given in the book is natural with respect to pointed maps, I'd love to see it!  I had to resort to a different definition using a unit and counit.

I am kind of pleased with the tactic `pointed_reduce`.  I had always expected that one of the more annoying things about HoTT for a classical homotopy theorist would be that basepoints can't be preserved strictly, so we have to carry around paths relating them everywhere.  But if we set things up correctly, then usually most of those paths can be path-inducted-away when actually proving things.

### DIFF
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -33,7 +33,7 @@ Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000 :
 (** Also, the total space of any based path space is contractible.  We define the [contr] fields as separate definitions, so that we can give them [simpl nomatch] annotations. *)
 
 Definition path_basedpaths {X : Type} {x y : X} (p : x = y)
-: (exist (fun y => x = y) x 1) = (y;p).
+: (x;1) = (y;p) :> {z:X & x=z}.
 Proof.
   destruct p; reflexivity.
 Defined.
@@ -41,18 +41,20 @@ Defined.
 Arguments path_basedpaths {X x y} p : simpl nomatch.
 
 Global Instance contr_basedpaths {X : Type} (x : X) : Contr {y : X & x = y} | 100.
+Proof.
   exists (x ; 1).
   intros [y p]; apply path_basedpaths.
 Defined.
 
 Definition path_basedpaths' {X : Type} {x y : X} (p : y = x)
-: (exist (fun y => y = x) x 1) = (y;p).
+: (x;1) = (y;p) :> {z:X & z=x}.
 Proof.
   destruct p; reflexivity.
 Defined.
 
 Global Instance contr_basedpaths' {X : Type} (x : X) : Contr {y : X & y = x} | 100.
-  exists (existT (fun y => y = x) x 1).
+Proof.
+  refine (BuildContr _ (x;1) _).
   intros [y p]; apply path_basedpaths'.
 Defined.
 
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -24,20 +24,68 @@ Defined.
 
 (** It follows that any space of paths in a contractible space is contractible. *)
 (** Because [Contr] is a notation, and [Contr_internal] is the record, we need to iota expand to fool Coq's typeclass machinery into accepting supposedly "mismatched" contexts. *)
+
 Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000 := let c := {|
   center := (contr x)^ @ contr y;
   contr := path2_contr ((contr x)^ @ contr y)
 |} in c.
 
-(** Also, the total space of any based path space is contractible. *)
+(** Also, the total space of any based path space is contractible.  We define the [contr] fields as separate definitions, so that we can give them [simpl nomatch] annotations. *)
+
+Definition path_basedpaths {X : Type} {x y : X} (p : x = y)
+: (exist (fun y => x = y) x 1) = (y;p).
+Proof.
+  destruct p; reflexivity.
+Defined.
+
+Arguments path_basedpaths {X x y} p : simpl nomatch.
+
 Global Instance contr_basedpaths {X : Type} (x : X) : Contr {y : X & x = y} | 100.
   exists (x ; 1).
-  intros [y []]; reflexivity.
+  intros [y p]; apply path_basedpaths.
+Defined.
+
+Definition path_basedpaths' {X : Type} {x y : X} (p : y = x)
+: (exist (fun y => y = x) x 1) = (y;p).
+Proof.
+  destruct p; reflexivity.
 Defined.
 
 Global Instance contr_basedpaths' {X : Type} (x : X) : Contr {y : X & y = x} | 100.
   exists (existT (fun y => y = x) x 1).
-  intros [y []]; reflexivity.
+  intros [y p]; apply path_basedpaths'.
+Defined.
+
+Arguments path_basedpaths' {X x y} p : simpl nomatch.
+
+(** Some useful computation laws for based path spaces *)
+
+Definition ap_pr1_path_contr_basedpaths {X : Type}
+           {x y z : X} (p : x = y) (q : x = z)
+: ap pr1 (path_contr ((y;p):{y':X & x = y'}) (z;q)) = p^ @ q.
+Proof.
+  destruct p,q; reflexivity.
+Defined.
+
+Definition ap_pr1_path_contr_basedpaths' {X : Type}
+           {x y z : X} (p : y = x) (q : z = x)
+: ap pr1 (path_contr ((y;p):{y':X & y' = x}) (z;q)) = p @ q^.
+Proof.
+  destruct p,q; reflexivity.
+Defined.
+
+Definition ap_pr1_path_basedpaths {X : Type}
+           {x y : X} (p : x = y)
+: ap pr1 (path_basedpaths p) = p.
+Proof.
+  destruct p; reflexivity.
+Defined.
+
+Definition ap_pr1_path_basedpaths' {X : Type}
+           {x y : X} (p : y = x)
+: ap pr1 (path_basedpaths' p) = p^.
+Proof.
+  destruct p; reflexivity.
 Defined.
 
 (** If the domain is contractible, the function is propositionally constant. *)
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -713,7 +713,7 @@ Ltac remember_as term name eqname :=
   pose (eqname := idpath : term = name);
   clearbody eqname name.
 
-Tactic Notation "remember" constr(term) "as" ident(name) "eqn" ident(eqname) :=
+Tactic Notation "remember" constr(term) "as" ident(name) "eqn:" ident(eqname) :=
   remember_as term name eqname.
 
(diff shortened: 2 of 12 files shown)


theories/Basics/Contractible.v
```diff
@@ -24,20 +24,68 @@ Defined.
...
 Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000 := let c := {|
   center := (contr x)^ @ contr y;
   contr := path2_contr ((contr x)^ @ contr y)
 |} in c.
 
-(** Also, the total space of any based path space is contractible. *)
+(** Also, the total space of any based path space is contractible.  We define the [contr] fields as separate definitions, so that we can give them [simpl nomatch] annotations. *)
+
+Definition path_basedpaths {X : Type} {x y : X} (p : x = y)
+: (exist (fun y => x = y) x 1) = (y;p).
```
JasonGross: This might look better as `(x; 1) = (y; p) :> { y : _ | x = y }`.

theories/Basics/Contractible.v
```diff
@@ -24,20 +24,68 @@ Defined.
...
+Arguments path_basedpaths {X x y} p : simpl nomatch.
+
 Global Instance contr_basedpaths {X : Type} (x : X) : Contr {y : X & x = y} | 100.
   exists (x ; 1).
-  intros [y []]; reflexivity.
+  intros [y p]; apply path_basedpaths.
+Defined.
+
+Definition path_basedpaths' {X : Type} {x y : X} (p : y = x)
+: (exist (fun y => y = x) x 1) = (y;p).
```
JasonGross: Same comment about `_ = _ :> _` as above.

theories/Basics/Contractible.v
```diff
@@ -24,20 +24,68 @@ Defined.
...
+Proof.
+  destruct p; reflexivity.
 Defined.
 
 Global Instance contr_basedpaths' {X : Type} (x : X) : Contr {y : X & y = x} | 100.
```
JasonGross: Do you want to add `Proof.` below this line, while you're updating this proof?

theories/Basics/Contractible.v
```diff
@@ -24,20 +24,68 @@ Defined.
...
   exists (existT (fun y => y = x) x 1).
```
JasonGross: Also consider `exists (x; 1)`, if you agree that reads better.

theories/Basics/Overture.v
```diff
@@ -698,3 +706,21 @@ Tactic Notation "transparent" "assert" "(" ident(name) ":" constr(type) ")" :=
...
 Tactic Notation "transparent" "eassert" "(" ident(name) ":" open_constr(type) ")" := transparent assert (name : type).
 Tactic Notation "transparent" "eassert" "(" ident(name) ":" open_constr(type) ")" "by" tactic3(tac) := transparent assert (name : type) by tac.
+
+(** A version of Coq's [remember] that uses our equality. *)
+Ltac remember_as term name eqname :=
+  set (name := term) in *;
+  pose (eqname := idpath : term = name);
+  clearbody eqname name.
+
+Tactic Notation "remember" constr(term) "as" ident(name) "eqn" ident(eqname) :=
```
JasonGross: Do we want to match the `eqn:` notation? (am I correctly remembering that Coq uses `eqn:`?)

theories/Basics/Contractible.v
mikeshulman: Curiously, that doesn't work on this one.

theories/Basics/Overture.v
mikeshulman: Yes, it does.  I guess I mistakenly thought that `Tactic Notation` couldn't handle the colon correctly.

theories/Basics/Contractible.v
JasonGross: Fascinating.  Does it work with `eexists`?  I presume it works with `refine`?

theories/Basics/Pointed.v
```diff
@@ -26,15 +28,223 @@ Global Instance ispointed_sigma `{IsPointed A} `{IsPointed (B (point A))}
...
+Definition pmap_idmap (A : pType): A ->* A
+  := Build_pMap A A idmap 1.
+
+Definition pmap_compose {A B C : pType}
+           (g : B ->* C) (f : A ->* B)
+: A ->* C
+  := Build_pMap A C (g o f)
+                (ap g (point_eq f) @ point_eq g).
+
+Infix "o*" := pmap_compose (at level 40).
```
JasonGross: An alternative: if you do
```coq
Delimit Scope pmap_scope with pmap.
Bind Scope pmap_scope with pMap.
Infix "o" := pmap_compose : pmap_scope.
```
You should be able to use `o` nearly seamlessly to mean both function composition and pointed-map composition.

theories/Basics/Pointed.v
```diff
@@ -26,15 +28,223 @@ Global Instance ispointed_sigma `{IsPointed A} `{IsPointed (B (point A))}
...
   := match n with
-       | O => existT IsPointed A (@point A H)
-       | S p => iteratedLoopSpace p (point = point)
+       | O => Build_pType A (@point A H)
+       | S p => iterated_loops p (point A = point A)
      end.
+
+(** ** Pointed functions *)
+
+Record pMap (A B : pType) :=
```
JasonGross: Do we want to try to have dependent pointed maps?  Presumably as another dependent type that `pMap` coerces into, because I don't see any way to give `forall` a type that doesn't involve `forall` in a bottomless way.

theories/Basics/Pointed.v
```diff
@@ -26,15 +28,223 @@ Global Instance ispointed_sigma `{IsPointed A} `{IsPointed (B (point A))}
...
+Ltac pointed_reduce :=
+  unfold pointed_fun, pointed_htpy; cbn;
+  repeat match goal with
+           | [ X : pType |- _ ] => destruct X as [X ?]
+           | [ phi : pMap ?X ?Y |- _ ] => destruct phi as [phi ?]
+           | [ alpha : pHomotopy ?f ?g |- _ ] => destruct alpha as [alpha ?]
+         end;
+  cbn in *; unfold point in *;
+  path_induction; cbn;
+  try rewrite !concat_p1;
```
JasonGross: Note: This is `rewrite ?concat_p1`.  You can even change them, and make it `rewrite ?concat_p1, ?concat_1p`.

theories/Basics/Pointed.v
```diff
@@ -26,15 +28,223 @@ Global Instance ispointed_sigma `{IsPointed A} `{IsPointed (B (point A))}
...
+           (h : C ->* D) (g : B ->* C) (f : A ->* B)
+: (h o* g) o* f ==* h o* (g o* f).
+Proof.
+  pointed_reduce.
+  refine (Build_pHomotopy _ _); cbn.
+  - intros ?; reflexivity.
+  - reflexivity.
+Defined.
+
+Definition pmap_compose_preid {A B : pType} (f : A ->* B)
```
JasonGross: Why `compose_pre` instead of `precompose_`?

theories/Basics/Pointed.v
```diff
@@ -26,15 +28,223 @@ Global Instance ispointed_sigma `{IsPointed A} `{IsPointed (B (point A))}
...
+: f o* pmap_idmap A ==* f.
+Proof.
+  pointed_reduce.
```
JasonGross: How much do we care about the fact that `pointed_reduce` (great tactic, by the way!) sometimes uses `rewrite`, and we want to use it in transparent proofs?

JasonGross:
I think `eqn` should be `eqn:`, but, other than that, LGTM, with or without my comments.

theories/Basics/Contractible.v
mikeshulman: No, `eexists` doesn't work either, but `refine (BuildContr _ (x;1) _)` does.  My best guess is that it has to do with the fact that `fun y => x = y` is equivalently `paths x`, whereas `fun y => y = x` can't be written without a lambda.

theories/Basics/Pointed.v
mikeshulman: It's worth thinking about, but I don't think it's worth trying to add to this PR.

theories/Basics/Pointed.v
mikeshulman: Hmm, I *thought* when I was writing this code that I would want to use it in transparent proofs, but it appears that we don't yet actually need any of the proofs involving it to be transparent.  Should we make them all opaque for now and deal with the problem if and when it turns out they have to be transparent?

JasonGross: I think either way, it should get a `TODO` comment.  I have no strong feelings on whether the proofs should be opaque or transparent, for the time being.  (I also tend to think in dependent `match` statements, though, and don't see `rewrite` as something super-terrible for transparent proofs...)

spitters:
LGTM.

theories/Basics/Pointed.v
mikeshulman: I just tried it, but it fails on the first line of the proof of `loop_susp_adjoint`, and I can't figure out how to fix it.

mikeshulman:
Comments mostly addressed, planning to merge when Travis finishes.

One question for advice: I would like `A ->* B` to also be a *pointed* type (its point is `const (point B)`).  One way I can think of to do that is to first define `pMap` as a record, and then some enhancement of it like `ppMap` an inhabitant of `pType`, with `A ->* B` as a notation for the latter.  Another would be to use the typeclass `IsPointed` rather than its sigma-type `pType` more systematically everywhere, so that it would be enough to define an instance `IsPointed (pMap A B)`.  Opinions?

theories/Basics/Pointed.v
JasonGross: Well, `%pmap` "fixes" it, but suspect that's not what you want to do there?  (Or maybe it is?)  The problem is that Coq won't pick scopes based on types on-the-fly.

mikeshulman: I couldn't get `%pmap` to fix it.  Do you have code that works?

JasonGross: [Here](https://github.com/JasonGross/HoTT/compare/mikeshulman:pointed...pr/676-pmap-scope)

mikeshulman:
Life would be so much easier if coercions could take typeclass arguments, e.g. if we could use a `Type` where a `pType` is expected and Coq would know to search the instance database for an `IsPointed`.

theories/Basics/Pointed.v
mikeshulman: Ah, I was fooled because the error was on the *second* argument to `equiv_adjointify`, so I didn't realize the *first* argument also needed a `%pmap`.

I'm not sure whether I like this or not, so maybe I won't for now, but I'll add a comment about it as a possibility for the future.

JasonGross: Coq does some things from rright to left.  If you annotate the second argument with `%pmap`, you will then get a nearly identical complaint about the first argument.  (Moving the error further to the left is making progress.)  I notice this most commonly with "unbound identifier" errors, when I rename a variable, and then let Coq pick out all the places I need to fix it.

mikeshulman: Ah, ok.

PR #683, opened 2015-01-13
Decidable props

mikeshulman:
See discussion at #654.  There are some unresolved questions and other possibilities to explore, but let's not let the perfect be the enemy of the good.  This adds `DHProp` (decidable props that are hprops without funext) to the previously linked implementation.


theories/DProp.v
```diff
@@ -0,0 +1,284 @@
...
+Proof.
+  intros P Q.
+  refine (trunc_equiv' _ (n := -1) (equiv_path_dprop P Q)).
+Defined.
+
+Global Instance isequiv_dprop_to_bool `{Univalence}
+: IsEquiv dprop_to_bool.
+Proof.
+  refine (isequiv_adjointify dprop_to_bool
+            (fun b:Bool => if b then True else False) _ _).
```
JasonGross: Why not `bool_to_dhprop`, and let coercions and eta-expansion take care of the types?

theories/DProp.v
```diff
@@ -0,0 +1,284 @@
...
+Infix "&&" := dand : dprop_scope.
+Infix "&&" := dhand : dhprop_scope.
+Infix "||" := dor : dprop_scope.
+Infix "||" := dhor : dhprop_scope.
+Infix "->" := dimpl : dprop_scope.
+Notation "!! P" := (dneg P) (at level 75, right associativity)
+                   : dprop_scope.
+
+Delimit Scope dprop_scope with dprop.
+Delimit Scope dhprop_scope with dhprop.
```
JasonGross: You should also `Bind Scope dprop_scope with DProp` and `Bind Scope dhprop_scope with DHProp`.  This way, `dprop_scope` is automatically opened for function arguments expecting a `DProp`, and `dhprop_scope` is automatically opened for function arguments expecting a `DHProp`.

theories/DProp.v
```diff
@@ -0,0 +1,284 @@
...
+: IsFalse (P && Q).
+Proof.
+  intros [p q].
+  exact (dprop_isfalse q).
+Defined.
+
+Global Instance dhand_true_true {P Q : DHProp} `{IsTrue P} `{IsTrue Q}
+: IsTrue (P && Q)%dhprop.
+Proof.
+  (** Why does Coq need the argument [P] here? *)
```
JasonGross: My understanding: when you say ` it needs to fill in type classes for `?1 : IsTrue ?2`, and also separately needs to unify `IsTrue ?2` with `P`.  Apparently, it does typeclass resolution first (since `Typeclasses Strict Resolution` is `Local` to `Overture`), and decides that `?1` should be `H0`, and `?2` should therefore be `Q`.  It then tries to unify `IsTrue Q` with `P`, and fails.  If you `Set Typeclasses Strict Resolution`, then it doesn't know how to unify `IsTrue ?2` with `P`.  (Apparently unification is unwilling to unfold `IsTrue`?  I'm not really sure what's going on here.  @mattam82, any idea what's going on if you `Set Typeclasses Strict Resolution` at the top, and then do `refine (_, _)`, and then `refine (_ : IsTrue _)`---why does it fail, rather than unfolding `IsTrue` and unifying the evar with `P`?)

theories/DProp.v
```diff
@@ -0,0 +1,284 @@
...
+Global Instance dor_true_r {P Q} `{IsTrue Q}
+: IsTrue (P || Q).
+Proof.
+  exact (tr (inr P dprop_istrue)).
+Defined.
+
+Global Instance dor_false_false {P Q} `{IsFalse P} `{IsFalse Q}
+: IsFalse (P || Q).
+Proof.
+  intros pq. strip_truncations. destruct pq as [p|q].
```
JasonGross: Note: Alternative that should work: `intros [p|q]; strip_truncations`.  Up to you which one you use.

theories/DProp.v
```diff
@@ -0,0 +1,284 @@
...
+: IsTrue (P || Q)%dhprop.
+Proof.
+  exact (tr (inr P dprop_istrue)).
+Defined.
+
+Global Instance dhor_false_false {P Q : DHProp} `{IsFalse P} `{IsFalse Q}
+: IsFalse (P || Q)%dhprop.
+Proof.
+  intros pq. strip_truncations. destruct pq as [p|q].
+  - (** Why does Coq need the argument [P] here? *)
```
JasonGross: See above.

theories/Types/Arrow.v
```diff
@@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D)
 (** What remains is really identical to that in [Forall].  *)
 
 End AssumeFunext.
+
+(** ** Decidability *)
+
+(** This doesn't require funext *)
+Global Instance decidable_arrow {A B : Type}
```
JasonGross: There should be a `decidable_forall`, right?

theories/DProp.v
JasonGross: My understanding: when you say `dprop_istrue` it needs to fill in type classes for `?1 : IsTrue ?2`, and also separately needs to unify `IsTrue ?2` with `P`.  Apparently, it does typeclass resolution first (since `Typeclasses Strict Resolution` is `Local` to `Overture`), and decides that `?1` should be `H0`, and `?2` should therefore be `Q`.  It then tries to unify `IsTrue Q` with `P`, and fails.  If you `Set Typeclasses Strict Resolution`, then it doesn't know how to unify `IsTrue ?2` with `P`.  (Apparently unification is unwilling to unfold `IsTrue`?  I'm not really sure what's going on here.  @mattam82, any idea what's going on if you `Set Typeclasses Strict Resolution` at the top, and then do `refine (_, _)`, and then `refine (_ : IsTrue _)`---why does it fail, rather than unfolding `IsTrue` and unifying the evar with `P`?)

theories/DProp.v
mikeshulman: Because I wrote that before I wrote `bool_to_dhprop`.  Fixed.

theories/DProp.v
mikeshulman: No, you can't destruct `P || Q` until you've stripped the truncation from it.

theories/Types/Arrow.v
mikeshulman: Only if the domain is a finite type.

spitters:
I don't fully appreciate what is going on with funext. It is not needed in the usual developments on decidable types. Could you explain? Did I miss a discussion?

mikeshulman:
We need funext if we want the dprop `A -> B` to be the actual function-space `A -> B` and also to be an hprop.

spitters:
I understand now, I was thinking about DecidablePaths.

LGTM

theories/Types/Arrow.v
andrejbauer: Only if the domain is an exhaustible space in the teminology of @martinescardo. Which spaces are exhaustible depends on whether we have excluded middle, or continuity principles, etc. But this is not the place to worry about such nice generalizations.

martinescardo: A lot of infinite types are exhaustible without assuming excluded middle 
or continuity principles. The easiest to explain, and perhaps the 
simplest, is the set of decreasing binary sequences (equivalent to that 
of increasing binary sequences, and also equivalent to the type of 
binary sequences that have at most one 1). This is one manifestation of 
the ordinal omega + 1. Another is N+1, which is not exhaustible unless 
LPO holds. Classically they are equivalent. Constructively, their 
equivalence is equivalent to LPO. I have implemented such examples of 
exhaustible infinite types, with their proofs of exhaustibility, in 
Agda. There is also a published paper.

JasonGross:
I would prefer a better comment than `(** Why does Coq need the argument [P] here? *)`, based on the explanation I gave, but this is minor, so, LGTM.

theories/DProp.v
mikeshulman: Do you have any explanation of why the explicit argument isn't needed in the proof of `dand_true_true`?

theories/Types/Arrow.v
andrejbauer: It would be cool to give 𝜔+1 as an example just to point out that `decidable_forall` isn't restricted to finite types.

martinescardo: On 15/01/15 23:42, Andrej Bauer wrote:
>> @@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D)
>>   (** What remains is really identical to that in [Forall].  *)
>>
>>   End AssumeFunext.
>> +
>> +(** ** Decidability *)
>> +
>> +(** This doesn't require funext *)
>> +Global Instance decidable_arrow {A B : Type}
>
> It would be cool to give 𝜔+1 as an example just to point out that `decidable_forall` isn't restricted to finite types.

If you want to do this, you can start from Vincent Siles' translation to 
Coq of part of my Agda code.

    http://www.cs.bham.ac.uk/~mhe/papers/omniscient-original/InfOmni.v

Martin

theories/DProp.v
JasonGross: Using `Set Debug Unification` suggests that in the case where it succeeds, it first tries to unify `dprop_type P` (from the goal) with `dprop_type ?P` (the type of `IsTrue ?P`), and says `?P` must equal `P`.  Then typeclass resolution fires, and it succeeds on `IsTrue P`.  In the case where it fails, it first tries to unify `@trunctype_type (trunc_S minus_two) (dhprop_hprop P)` with `dprop_type ?P`, and does not succeed; hence when typeclass resolution fires, it tries to solve `IsTrue ?P`, and instantiates `?P` with `Q`, which does not result in a type that unifies with the goal.

martinescardo:
Just for the record, regarding terminology: in the Bishop style mathematics community, usually

(1) A proposition P is called decidable if P or not P holds.

(2) a set is called discrete if it has decidable equality, and 

(3) a subset is called detachable if membership of a point of the superset in the subset is decidable.

But outside that community, people seem to use the terminology decidable for the three concepts, which is very confusing (as demonstrated above).

theories/Types/Arrow.v
mikeshulman: Is there something we can call the set of decreasing binary sequences, when viewing it as an ordinal, other than "𝜔+1"?  To me "𝜔+1" sounds like it should mean N+1.

mikeshulman:
I've fixed the comment based on your additional explanation, @JasonGross -- thanks.  Will now merge when Travis finishes.

@martinescardo, thanks for the terminological note.  I like "detachable" for (3) as it seems unambiguous.  Unfortunately "discrete" has many other meanings, so I'd rather avoid it, but "decidable equality" or "decidable paths" (what we're currently using) for (2) seem unambiguous.  I suppose we could say "decidably inhabited" for (1), which @spitters might have been suggesting above, but it would be a bit wordy.

theories/Types/Arrow.v
martinescardo: Many others and I call this set \mathbb{N}_\infty.

Rather than N+1, this is a sort of one-point compactification of N.
(Or one-point exhaustibilification.)

M.


On 18/01/15 04:56, Mike Shulman wrote:
> In theories/Types/Arrow.v
> <https://github.com/HoTT/HoTT/pull/683#discussion-diff-23130898>:
> 
>> @@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D)
>>  (** What remains is really identical to that in [Forall].  *)
>>  
>>  End AssumeFunext.
>> +
>> +(** ** Decidability *)
>> +
>> +(** This doesn't require funext *)
>> +Global Instance decidable_arrow {A B : Type}
> 
> Is there something we can call the set of decreasing binary sequences,
> when viewing it as an ordinal, other than "𝜔+1"? To me "𝜔+1" sounds
> like it should mean N+1.
> 
> —
> <https://github.com/HoTT/HoTT/pull/683/files#r23130898>.
> 

-- 
Martin Escardo
http://www.cs.bham.ac.uk/~mhe

mikeshulman: Yes, N_oo is a great name, but here you were calling it a version of 𝜔+1, which I guess is also true (is it something like the plump successor of the thin 𝜔?) but more ambiguous.

martinescardo: On 18/01/15 19:53, Mike Shulman wrote:
> In theories/Types/Arrow.v
> <https://github.com/HoTT/HoTT/pull/683#discussion-diff-23136294>:
> 
>> @@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D)
>>  (** What remains is really identical to that in [Forall].  *)
>>  
>>  End AssumeFunext.
>> +
>> +(** ** Decidability *)
>> +
>> +(** This doesn't require funext *)
>> +Global Instance decidable_arrow {A B : Type}
> 
> Yes, N_oo is a great name, but here you were calling it a version of
> 𝜔+1, which I guess is also true (is it something like the plump
> successor of the thin 𝜔?) but more ambiguous.

Classically, both constructions produce the same object (LPO is enough
to make them equal). Constructively, they are potentially different
(e.g. different if we have ¬WLPO, which is a weak continuity axiom).
N_oo is bigger than N+1, in the sense that has it embedded. The
complement of N+1 in N_oo is always empty (regardless of constructivity
assumptions). There is much more to say. E.g. N_oo is the final
coalgebra of 1+(-), but N+1 is not necessarily so. N_oo\{oo} is not
necessarily N.

martinescardo: On 18/01/15 19:53, Mike Shulman wrote:
>> @@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D)
>>   (** What remains is really identical to that in [Forall].  *)
>>
>>   End AssumeFunext.
>> +
>> +(** ** Decidability *)
>> +
>> +(** This doesn't require funext *)
>> +Global Instance decidable_arrow {A B : Type}
>
> Yes, N_oo is a great name, but here you were calling it a version of 𝜔+1, which I guess is also true (is it something like the plump successor of the thin 𝜔?) but more ambiguous.

Actually, yes, it is the plump successor (the ordinal of all initial 
segment of) N, isn't it?

mikeshulman: Well, now that I look it up, Taylor defines the plump successor of a plump ordinal to be the set of all of its initial segments that are themselves plump ordinals.  N is not itself plump, but we could apply the same definition, however it probably wouldn't give us much since not many initial segments of N are plump either.  So it's a little less clear.

BTW, when you reply to a github email with your email client, the github web interface apparently manages to format it readably, but the resulting email I get from github looks like this:

> On 18/01/15 19:53, Mike Shulman wrote: > @@ -198,3 +198,19 @@ Definition equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D) > (** What remains is really identical to that in [Forall]. *) > > End AssumeFunext. > + > +(** ** Decidability *) > + > +(** This doesn't require funext *) > +Global Instance decidable_arrow {A B : Type} Yes, N_oo is a great name, but here you were calling it a version of 𝜔+1, which I guess is also true (is it something like the plump successor of the thin 𝜔?) but more ambiguous. Actually, yes, it is the plump successor (the ordinal of all initial segment of) N, isn't it?

Deleting the quoted-reply text before replying might make it more readable.

spitters: Something strange is happening to your replies which makes them hard to
read.
One possibility is to use the github web interface.

> On 18/01/15 19:53, Mike Shulman wrote: > @@ -198,3 +198,19 @@ Definition
> equiv_functor_arrow' `(f : B <~> A) `(g : C <~> D) > (** What remains is
> really identical to that in [Forall]. *) > > End AssumeFunext. > + > +(**
> ** Decidability *) > + > +(** This doesn't require funext *) > +Global
> Instance decidable_arrow {A B : Type} Yes, N_oo is a great name, but here
> you were calling it a version of 𝜔+1, which I guess is also true (is it
> something like the plump successor of the thin 𝜔?) but more ambiguous.
> Actually, yes, it is the plump successor (the ordinal of all initial
> segment of) N, isn't it?
>

martinescardo: Also, Noo is the set of detachable initial segments of N, not all 
initial segments.

martinescardo: Ok, I will use the web interface, Mike and Bas.

mikeshulman: Well, maybe it doesn't have a very natural ordinal-theoretic description then.

Is there any way to define precisely what we mean by "a version of 𝜔+1" that would make it true that Noo is such a thing?  It ought to include both N+1 (the thin 𝜔+1) and the plump 𝜔+1 as "extremal" examples, with Noo somewhere in between.  I suppose one such definition would be "an ordinal that becomes equal to 𝜔+1 if we assume classical logic"...

martinescardo: On 19/01/15 17:13, Mike Shulman wrote:

> [0] Well, maybe it doesn't have a very natural ordinal-theoretic
> description then.

Let me come back to this below.

> [1] Is there any way to define precisely what we mean by "a version
> of 𝜔+1" that would make it true that Noo is such a thing?

I regard [3] below as a poor answer, but as a good starting one (and
the one I indicated in my discussion above).

> [2] It ought to include both N+1 (the thin 𝜔+1) and the plump 𝜔+1 as
> "extremal" examples, with Noo somewhere in between.

Agreed. (Maybe!)

> [3] I suppose one such definition would be "an ordinal that becomes
> equal to 𝜔+1 if we assume classical logic"...

This is what people often do, for situations like this, but I am not
happy about this, and I suppose you are not either.

Now suppose you consider, like Cantor did, the notion of ordinal in an
ambient poset. Cantor considered subsets of the real line that are
ordinals under the inherited order.

Now, you can embed (discrete) N into (Euclidean) R, even
topologically, in two fundamentally different ways. In one, the image
is just "N", namely, 0,1,2,3,... . In the other, the image is 0, 1/2,
3/4, 5/8, ... . The closure of the first is itself. Classically, the
closure of the second just adds 1. Constructively, the topological
closure is (equivalent to) Noo.

Now, there is something interesting which I don't know whether people
have noticed before in connection with constructive mathematics
(probably not). Although it is possible to find an order on (discrete)
N that makes it isomorphic to e.g. epsilon_0 (that's why we call
epsilon_0 a countable ordinal), there is no discrete subspace of R
which is order isomorphic to epsilon_0 in the inherited order - but
there are non-discrete subspaces with this property.  Actually, this
happens rather early, well before epsilon_0. Even for omega * omega we
have this phenomenon: it doesn't occur as a discrete subspace of R
under the inherited order, even though it is (classically) countable.

This doesn't answer your question, but points out that things that
look the same classically can be very different constructively. The
ordinal epsilon_0 inside R (or inside 2^N for that matter) is very
different from the epsilon_0 logicians consider, for example (which is
discrete).

However, N+1 and Noo are somewhat canonical (initial and final
coalgebras, constructively).

It would be interesting to learn systematically about the many
constructive lifes of the same classical ordinal. As we have seen, Noo
is exhaustible (or "compact"), but N+1 only has decidable equality (or
is "discrete).

There is more to say (some of which is implemented in Agda, and some
of which is in my paper notebooks only).

martinescardo: I updated the above comment.

martinescardo: I take back some of the above assertions. We can embed discrete omega * omega in R, for example.

mikeshulman: The comments on a github commit diff is not the best place to have this conversation.  What about a blog post or the mailing list?

martinescardo: Agreed. A private conversation would be better. A blog post would be premature, and I recently I prefer to avoid the mailing list. (But, just for the record, I entered this conversation because Andrej started it, and then I followed it because others did.)

PR #685, opened 2015-01-17
Bump Coq trunk

JasonGross:
This gives us [Set Typeclasses Modulo Eta] and [Set Typeclasses Dependency Order], which, together, might be able to deal with #657 in a better way.  (Though perhaps not; see [bug 3925](https://coq.inria.fr/bugs/show_bug.cgi?id=3925) and [bug 3926](https://coq.inria.fr/bugs/show_bug.cgi?id=3926)).


mikeshulman:
What do those two `Set`s do?

JasonGross:
The first one says "eta-expand when necessary" (i.e., don't fail on trying to find an instance for `IsEquiv (fun x => f^-1 x)` when there's no `isequiv_compose` and `isequiv_idmap` lying arouns).  The second one fixes the order in which typeclass goals are solved, so that solving `IsEquiv f^-1` doesn't fail if the evar for `IsEquiv f` that was created later on in the type.

mikeshulman:
I can't parse your second sentence.

mikeshulman:
I can't parse your second sentence.

JasonGross:
Let me try again; I was trying to explain the implementation, which is
complicated. Here is the observable effect:  Sometimes `refine` will leave
over goals which can be solved by `exact _`.  The second `Set` is an
attempt (hopefully successful) to fix that.

mikeshulman:
Ah, excellent!  Maybe that will allow us to remove some `exact _`s throughout the library.  But do we have to wait for 3925 and 3926 before actually setting those options?

JasonGross:
`Set Typeclasses Dependency Order` should be fine to set.  `Set Typeclasses Modulo Eta`, on the other hand, tends to cause loops of `isequiv_compose` with the first component as `isequiv_idmap`, and the second component as `isequiv_compose` with the first component as `isequiv_idmap`, and the second component as ...

mattam82:
One way to avoid looping would be to give the instance for isequiv_compose explicitely using a Hint Extern I suppose, forcing the goal to contain an actual composition.

mikeshulman:
`Global Set Typeclasses Dependency Order` appears to cause a coredump or stack overflow on `ap_morphism_inverse_path_isomorphic` in `categories/Category/Morphism`.

JasonGross:
That sounds like a bug.  Do you want to minimize it for @mattam82?

mattam82:
I suggest you look at the resolution trace, looping due to the instances
you have is the most likely reason.

2015-01-19 23:39 GMT+05:30 Jason Gross <notifications@github.com>:

> That sounds like a bug. Do you want to minimize it for @mattam82
> <https://github.com/mattam82>?
>
> —
> <https://github.com/HoTT/HoTT/pull/685#issuecomment-70536523>.
>



-- 
-- Matthieu

mikeshulman:
Does "resolution trace" mean `Set Debug Unification`?  That gives some output that I don't know how to read with lots of `_UNBOUND_REL`s, but it's not very long so it doesn't look like an infinite loop.  `Set Typeclasses Debug` doesn't produce any extra output.  (Why isn't it `Set Unification Debug` or `Set Debug Typeclasses` for consistency?)

mikeshulman:
<https://coq.inria.fr/bugs/show_bug.cgi?id=3943>

PR #686, opened 2015-01-18
Finite sets

mikeshulman:


theories/Spaces/Finite.v
```diff
@@ -0,0 +1,861 @@
...
+Local Open Scope path_scope.
+Local Open Scope equiv_scope.
+
+Local Open Scope nat_scope.
+
+(** * Finite sets *)
+
+(** ** Canonical finite sets *)
+
+(** A *finite set* is a type that is merely equivalent to the canonical finite set determined by some natural number.  There are many equivalent ways to define the canonical finite sets, such as [{ k : nat & k < n}]; we instead choose a recursive one. *)
```
andrejbauer: So these are the "strongly" finite sets. A weaker form of finite sets are the Kuratowski finite sets, which are images of the `Fin n`.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,861 @@
...
+(** Like canonical finite sets, finite sets have decidable equality. *)
+Global Instance decidablepaths_finite `{Funext} X `{Finite X}
+: DecidablePaths X.
+Proof.
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (decidablepaths_equiv _ e^-1 _).
+Defined.
+
+(** However, contrary to what you might expect, we cannot assert that "every finite set is decidable"!  That would be claiming a *uniform* way to select an element from every nonempty finite set, which contradicts univalence.  The most we can say is that any finite hprop is decidable.  Question: can this be proven without funext? *)
```
andrejbauer: Is every finite set merely decidable?

andrejbauer:
Very nice, but what are you up to?

mikeshulman:
I need finite sets in order to prove Lagrange's theorem.

theories/Spaces/Finite.v
mikeshulman: Yes, I've been taught to call those "finitely indexed" (or, as you say, Kuratowski-finite). <http://ncatlab.org/nlab/show/finite+set#Constructivist>

theories/Spaces/Finite.v
mikeshulman: Yes, I think so.  Every finite set is either empty or merely inhabited, and any merely inhabited type is merely decidable (since any inhabited type is decidable).  Should I add that?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,861 @@
...
+
+Definition fin_transpose_last_two (n : nat)
+: Fin n.+2 <~> Fin n.+2
+  := (equiv_compose'
+        (equiv_inverse (equiv_sum_assoc _ _ _))
+     (equiv_compose'
+        (equiv_functor_sum' (equiv_idmap _)
+                            (equiv_sum_symm _ _))
+        (equiv_sum_assoc _ _ _))).
+
```
spitters: It's a pity that these functions are not more readable.
On the other hand, these definitions have the advantage that you simulateneously prove that it is an equivalence.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,861 @@
...
+  transitivity (@fcard Y (finite_equiv X e _)).
+  - reflexivity.
+  - exact (ap (@fcard Y) (path_ishprop _ _)).
+Defined.
+
+Definition fcard_equiv' {X Y} (e : X <~> Y)
+           `{Finite X} `{Finite Y}
+: fcard X = fcard Y
+  := fcard_equiv e.
+
```
spitters: Shouldn't we have a lemma saying that `X <~>Y -> Finite X <~> Finite Y`  ?

mikeshulman: `finite_equiv'` says `(X <~> Y) -> (Finite X -> Finite Y)`.  Is there any purpose to improving the latter implication to an equivalence.

theories/Types/Sum.v
```diff
@@ -113,13 +165,167 @@ Definition transport_sum {A : Type} {P Q : A -> Type} {a a' : A} (p : a = a')
...
+
 (** ** Functorial action *)
 
 Definition functor_sum {A A' B B' : Type} (f : A -> A') (g : B -> B')
 : A + B -> A' + B'
   := fun z => match z with inl z' => inl (f z') | inr z' => inr (g z') end.
 
-(** ** Equivalences *)
+(** ** Unfunctorial action *)
+
```
spitters: The name is not entirely clear to me. Some more information might be helpful.

theories/Spaces/Finite.v
mikeshulman: Yes, that's an interesting point.  `equiv_compose` and `equiv_inverse` could be written out as `o` and `^-1` and then typeclass inference ought to pick up that they preserve equivalences, but `equiv_functor_sum` and `equiv_sum_assoc` and `equiv_sum_symm` would have to be written out as match statements and I don't think Coq could guess that they are equivalences.

If the consensus is that it would be better to define the functions more readably and then have to prove more by hand that they are equivalences, I'm happy to go along with that.

theories/Types/Sum.v
mikeshulman: Haha, fair enough.

theories/Spaces/Finite.v
spitters: I can see the benefits of both definitions. Please do is you see fit.

theories/Spaces/Finite.v
spitters: I don't know how useful it is, but it's a lemma that I expected in the flow of the development.

theories/Spaces/Finite.v
JasonGross: You can also add a notation `o` for `equiv_compose` in `equiv_scope`, and then put these in `equiv_scope`.

mikeshulman: Yes, I've been waffling for a while over a notation for `equiv_compose'`, which gets used a lot and is not very readable.  Another option is to define a special notation just for it, like `oE` or something, that's easy to identify visually as not quite the same as `o`.

theories/Basics/Decidable.v
```diff
@@ -17,6 +17,25 @@ Class DecidablePaths (A : Type) :=
...
 Global Existing Instance dec_paths.
 
+(** ** Decidable hprops *)
+
+(** Contractible types are decidable and have decidable equality. *)
+
+Global Instance decidable_contr X `{Contr X} : Decidable X
+  := inl (center X).
+
+Global Instance decidablepaths_contr X `{Contr X} : DecidablePaths X
```
JasonGross: Am I missing something, or can this be generalized to `IsHProp X -> DecidablePaths X`?  (And then have this be picked up automatically by typeclass resolution.)

mikeshulman: Yes, good point, of course it can.

theories/Spaces/Finite.v
JasonGross: I would be in favor of either `o` or `oE`.  (If it's relevant, `o` also means composition of morphisms, and it seems to rarely be ambiguous.  I have no objection to using `o` for all of the composition.)

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+: fin_transpose_last_two n (inl (inr tt)) = (inr tt)
+  := 1.
+
+Definition fin_transpose_last_two_rest (n : nat) (k : Fin n)
+: fin_transpose_last_two n (inl (inl k)) = (inl (inl k))
+  := 1.
+
+(** *** Swap the last element with [k]. *)
+
+Fixpoint fin_transpose_last_with (n : nat) (k : Fin n.+1)
```
JasonGross: Should these be prefixed with `equiv_`?

Also, why not `{n : nat}`?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+Definition fin_transpose_last_with_last (n : nat) (k : Fin n.+1)
+: fin_transpose_last_with n k (inr tt) = k.
+Proof.
+  destruct k as [k|].
+  - induction n as [|n IH]; simpl.
+    + elim k.
+    + destruct k as [k|].
+      * simpl. rewrite IH; reflexivity.
+      * simpl. apply ap, ap, path_contr.
+  - (** For some reason [simpl]/[cbn] won't simplify this until [n] is destructed. *)
```
JasonGross: `Fixpoint`s are not reduced unless their arguments are constructors.  (Otherwise, you'd run the risk of an infinite loop.)  I haven't looked at the goal, but does this answer your question?  (If not, paste the goal here?)

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+
+(** Here is the curried version that we will prove to be an equivalence. *)
+Definition fin_equiv' (n m : nat)
+: ((Fin m.+1) * (Fin n <~> Fin m)) -> (Fin n.+1 <~> Fin m.+1)
+  := fun ke => fin_equiv n m (fst ke) (snd ke).
+
+(** We construct its inverse and the two homotopies first as versions using homotopies without funext (similar to [ExtendableAlong]), then apply funext at the end. *)
+Definition fin_equiv_hfiber (n m : nat) (e : Fin n.+1 <~> Fin m.+1)
+: { kf : (Fin m.+1) * (Fin n <~> Fin m) & fin_equiv' n m kf == e }.
+Proof.
```
JasonGross: This is a long proof.  Any insight into what makes it so long?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+    + refine (fin_equiv_inj_fst n m k l e f p).
+    + apply path_equiv, path_arrow.
+      refine (fin_equiv_inj_snd n m k l e f p).
+Qed.
+
+Definition equiv_fin_equiv `{Funext} (n m : nat)
+: ((Fin m.+1) * (Fin n <~> Fin m)) <~> (Fin n.+1 <~> Fin m.+1)
+  := BuildEquiv _ _ (fin_equiv' n m) _.
+
+(** In particular, this implies that if two canonical finite sets are equivalent, then their cardinalities are equal. *)
```
JasonGross: Should we include a corollary about this meaning that `Fin` is injective?

mikeshulman: Hmm, well, we don't generally use the word "injective" except between hsets, and `Fin` is not an embedding.

theories/Spaces/Finite.v
mikeshulman: Not really.  (-:  If you can make it shorter, I'd be very happy...

theories/Spaces/Finite.v
mikeshulman: Hm.  What is an example of a fixpoint that would result in an infinite loop if it were reduced on a non-constructor?  I guess I thought the guardedness requirement that recursive calls only occur inside matches was sufficient to avoid infinite loops.

JasonGross:
```coq
Fixpoint id' (x : nat) : nat
:= match x with
     | 0 => 0
     | S x' => S (id' x')
   end.

Goal forall x, id' x = x.
  intro x.
  unfold id'.
```
Say you were allowed to reduce under fixpoints.  Then we can reduce
```
1 subgoals, subgoal 1 (ID 5)

  x : nat
  ============================
   (fix id' (x0 : nat) : nat :=
      match x0 with
      | 0 => 0
      | S x' => S (id' x')
      end) x = x

(dependent evars:)
```
to
```coq
   match x with
    | 0 => 0
    | S x' => S (id' x')
   end = x
```
to
```coq
   match x with
    | 0 => 0
    | S x' => S (match x' with
                   | 0 => 0
                   | S x'' => S (id' x'')
                 end)
   end = x
```
to ... (oops, infinite loop)

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+(** However, contrary to what you might expect, we cannot assert that "every finite set is decidable"!  That would be claiming a *uniform* way to select an element from every nonempty finite set, which contradicts univalence.  The most we can say is that any finite set is *merely* decidable. *)
+Definition merely_decidable_finite X `{Finite X}
+: merely (Decidable X).
+Proof.
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  apply tr; refine (decidable_equiv _ e^-1 _).
+Defined.
+
+(** We can also prove that any finite hprop is decidable.  Question: can this be proven without funext? *)
```
JasonGross: We should be able to, at least, strip off the `merely` without funext: we can `destruct n`, and then say that we must show `X -> Empty` in the case where `n = 0`, and `X` in the case where `n = S n'`.  Since `Empty` and `X` are both hprops, we can strip truncations.  Is this enough?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+Proof.
+  (** If we assume [Funext], then typeclass inference produces this automatically, since [X] has decidable equality and (hence) is a set, so [x=y] is a decidable hprop.  But we can also deduce it without funext, since [Finite] is an hprop even without funext. *)
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (finite_equiv _ (ap e)^-1 _).
+  apply finite_decidable_hprop; exact _.
+Defined.
+
+(** Finite sets are also closed under successors. *)
+Global Instance finite_succ X `{Finite X} : Finite (X + Unit).
```
JasonGross: Does `finite_sum` depend on this?  If not, want to move this down and that up, so this is done automatically?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+    + exact (tr (Empty_ind P)).
+    + specialize (IH (P o inl) (f o inl)).
+      assert (e := f (inr tt)).
+      strip_truncations.
+      exact (tr (sum_ind P IH (Unit_ind e))).
+Defined.
+
+(** ** Constructions on finite sets *)
+
+(** Finite sets are closed under sums, products, function spaces, and equivalence spaces.  There are multiple choices we could make regarding how to prove these facts.  Since we know what the cardinalities ought to be in all cases (since we know how to add, multiply, exponentiate, and take factorials of natural numbers), we could specify those off the bat, and then reduce to the case of canonical finite sets.  However, it's more amusing to instead prove finiteness of these constructions by "finite-set induction", and then *deduce* that their cardinalities are given by the corresponding operations on natural numbers (because they satisfy the same recurrences). *)
```
JasonGross: Cute!

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+  strip_truncations.
+  refine (finite_equiv _ (functor_sum idmap e^-1) _).
+  generalize (fcard Y); intros n.
+  induction n as [|n IH].
+  - refine (finite_equiv _ (sum_empty_r X)^-1 _).
+  - refine (finite_equiv _ (equiv_sum_assoc X _ Unit) _).
+Defined.
+
+(** Note that the cardinality function [fcard] actually computes.  The same will be true of all the other proofs in this section, though we don't always verify it. *)
+Goal fcard (Fin 3 + Fin 4) = 7.
```
JasonGross: Note: Another way to do this is `Check idpath : fcard (Fin 3 + Fin 4) = 7.`

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,877 @@
...
+    simpl.
+    refine (_ @ nat_plus_comm _ _).
+    refine (ap011 plus _ _).
+    + apply IH.
+    + apply fcard_equiv', prod_unit_l.
+  Defined.
+
+(** *** Function types *)
+
+Global Instance finite_arrow `{Funext} X Y `{Finite X} `{Finite Y}
```
JasonGross: This generalizes to `finite_forall`, right?

JasonGross:
I am boarding my flight to Zurich soon, but I shall try to finish reviewing this in the next few days.

theories/Spaces/Finite.v
peterlefanulumsdaine: I thought `Check` was deprecated outside specific test files, because of being noisy during compilation — am I misremembering?

theories/Spaces/Finite.v
mikeshulman: Oh, ok.  I forgot that you can reduce the branches of a match even if the discriminee isn't a constructor.

theories/Spaces/Finite.v
mikeshulman: Yep, that works.  Thanks!

theories/Spaces/Finite.v
mikeshulman: Yes, `finite_sum` does depend on `finite_succ`.

theories/Spaces/Finite.v
mikeshulman: I believe it is discouraged for that reason.  We're already using it elsewhere to verify that definitions have the right number of universe parameters, but it's probably better to avoid it when possible.

theories/Spaces/Finite.v
mikeshulman: Yes, it should, and its cardinality should be a `finmult` analogous to `finplus`.

theories/Spaces/Finite.v
mikeshulman: I think I found that Coq often needed the argument `n`.  As for prefixes, I've been leaving off the `equiv_` prefix sometimes when something is defined directly as an equivalence; since we have `equiv_fun` as a coercion, anyone looking for a simple function can use the equivalence by itself.  We need the prefixed `equiv_` for operations whose non-equivalence version already has a name, or isn't always an equivalence, but when something is defined directly as an equivalence it doesn't seem as necessary.

theories/Spaces/Finite.v
mikeshulman: Let's make that a separate issue and PR: #687

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+
+Local Open Scope nat_scope.
+
+(** * Finite sets *)
+
+(** ** Canonical finite sets *)
+
+(** A *finite set* is a type that is merely equivalent to the canonical finite set determined by some natural number.  There are many equivalent ways to define the canonical finite sets, such as [{ k : nat & k < n}]; we instead choose a recursive one. *)
+
+Fixpoint Fin (n : nat) : Type
```
JasonGross: Why `Type` instead of `Type`?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+  := match n with
+       | 0 => Empty
+       | S n => Fin n + Unit
```
JasonGross: I guess you don't have a special case for `| 1 => Unit` so that induction is easier?

JasonGross:
Do you want to (either in this PR or another one) replace [`CardinalityRepresentative` in NatCategory.v](https://github.com/HoTT/HoTT/blob/master/theories/categories/NatCategory.v#L13) with `Fin`?

theories/Spaces/Finite.v
mikeshulman: What.

theories/Spaces/Finite.v
mikeshulman: Well, actually it didn't occur to me to make such a special case, but I think yes, induction is easier this way.

theories/Spaces/Finite.v
JasonGross: Oops, I meant to say "Why `Type` instead of `Type1`?", but it seems my hands got ahead of (behind?) me.

mikeshulman: Well, since `Empty` and `Unit` already live in `Type1`, there's no danger of `Fin` ending up in `Set`.

JasonGross: *nods*  I guess when I wrote `CardinalityRepresentative`, we still had `Empty` and `Unit` in `Set`.

theories/Spaces/Finite.v
JasonGross: Still, `ap Fin` is an equivalence, right?  Is there a name for this property, and do we want to include the proof?

mikeshulman: No, `ap Fin` is not an equivalence (that would be the same as `Fin` being an embedding).  For instance, `2 = 2` is contractible, whereas `Fin 2 = Fin 2` is not.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+: DecidablePaths X.
+Proof.
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (decidablepaths_equiv _ e^-1 _).
+Defined.
+
+(** However, contrary to what you might expect, we cannot assert that "every finite set is decidable"!  That would be claiming a *uniform* way to select an element from every nonempty finite set, which contradicts univalence.  The most we can say is that any finite set is *merely* decidable. *)
+Definition merely_decidable_finite X `{Finite X}
+: merely (Decidable X).
```
JasonGross: We can prove something stronger that this, I think: Every finite set is either empty or merely inhabited.  (That is, we can shove the merely underneath the disjunction.)  Equivalently, I think, we can show `Decidable (merely X)`, and we can show that, in general, `Decidable (merely X) -> merely (Decidable X)`.

mikeshulman: Yes, `Decidable (merely X)` follows from `finite_merely` and `decidable_finite_hprop`.

JasonGross: Do you want to do that factoring for this proof?

mikeshulman:
Ah, yes, unifying those would be nice.  But you made a different choice about special-casing 1, so simply dropping `Fin` in for `CardinalityRepresentative` doesn't work.  I also just tried making the other choice for `Fin`, and it looks like it would require a lot of work, e.g. the definition of `fin_transpose_last_two` already requires changing.  I don't think I have time to resolve it either way, but feel free to have a go yourself!

theories/Spaces/Finite.v
mikeshulman: Done.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+  (** If we assume [Funext], then typeclass inference produces this automatically, since [X] has decidable equality and (hence) is a set, so [x=y] is a decidable hprop.  But we can also deduce it without funext, since [Finite] is an hprop even without funext. *)
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (finite_equiv _ (ap e)^-1 _).
+  apply finite_decidable_hprop; exact _.
+Defined.
+
+(** Finite sets are also closed under successors. *)
+
+Locate "*".
```
JasonGross: This should not be here, I presume?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+Defined.
+
+Definition fcard_succ X `{Finite X}
+: fcard (X + Unit) = (fcard X).+1
+  := 1.
+
+
+(** ** Induction over finite sets *)
+
+(** Most concrete applications of this don't actually require univalence, but the general version does.  For this reason the general statement is less useful (and less used in the sequel) than it might be. *)
```
JasonGross: What is "the sequel"?

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+       `{Finite X} `{forall x, Finite (Y x)}
+: Finite (forall x:X, Y x).
+Proof.
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (finite_equiv' _
+            (equiv_functor_forall' (P := fun x => Y (e^-1 x)) e _) _); try exact _.
+  { intros x; refine (equiv_transport _ _ _ (eissect e x)). }
+  set (Y' := Y o e^-1); change (Finite (forall x, Y' x)).
+  assert (forall x, Finite (Y' x)) by exact _; clearbody Y'; clear e.
```
JasonGross: Another option, in case you weren't aware: `pose proof (_ : forall x, Finite (Y' x))`

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+Global Instance finite_sigma {X} (Y : X -> Type)
+       `{Finite X} `{forall x, Finite (Y x)}
+: Finite { x:X & Y x }.
+Proof.
+  assert (e := merely_equiv_fin X).
+  strip_truncations.
+  refine (finite_equiv' _
+            (equiv_functor_sigma (equiv_inverse e)
+                                 (fun x (y:Y (e^-1 x)) => y)) _).
+  (** For some reason, [set (Y' := Y o e^-1)] doesn't change the goal. *)
```
JasonGross: Ideas: `o` is behaving weirdly, the instance you're picking up for `IsEquiv e` doesn't match?  What happens when you `Set Printing All` and `set` the printing all term?

theories/Spaces/Finite.v
mikeshulman: Yes, I noticed that too and got rid of it in the latest commit.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,900 @@
...
+  induction n as [|n IH].
+  - exact _.
+  - refine (finite_equiv _ (equiv_sum_ind Y') _).
+    apply finite_prod.
+    + apply IH; exact _.
+    + refine (finite_equiv _ (@Unit_ind (fun u => Y' (inr u))) _).
+      refine (isequiv_unit_ind (Y' o inr)).
+Defined.
+
+Definition fcard_arrow `{Funext} X Y `{Finite X} `{Finite Y}
```
JasonGross: Do you want to define `finmult` and prove `fcard_forall`, and relate `nat_exp` to `finmult`?

mikeshulman: Already done.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,904 @@
...
+Definition finplus_const X `{Finite X} n
+: finplus (fun x:X => n) = fcard X * n.
+Proof.
+  transitivity (fcard (X * Fin n)).
+  - exact (fcard_equiv' (equiv_sigma_prod0 X (Fin n))).
+  - exact (fcard_prod X (Fin n)).
+Defined.
+
+(** Closure under sigmas and paths also implies closure under hfibers. *)
+Global Instance finite_hfiber {X Y} (f : X -> Y) (y : Y)
```
JasonGross: Any reason to make this an `Instance` rather than a `Definition`, given that typeclass resolution can already infer it?

theories/Spaces/Finite.v
mikeshulman: "The sequel" means "what follows".

theories/Spaces/Finite.v
JasonGross: Oops, sorry about noise, then.

theories/Spaces/Finite.v
mikeshulman: Ohh, I bet it's because `compose` is still beta-expanded.

theories/Spaces/Finite.v
mikeshulman: Is one way or the other likely to make typeclass inference faster?

JasonGross:
> But you made a different choice about special-casing 1, so simply dropping `Fin` in for `CardinalityRepresentative` doesn't work.

Have you tried this?  I suspect it won't require much work to do the drop in, other than replacing `tt` with `inl tt` or `inr tt`.  (The comment is slightly misleading, thinking back on it, and applies more to `nat_category` than to `CardinalityRepresentative`, where it really matters that we choose `indiscrete_category 1` rather than `discrete_category 1`.)

If you don't have time, though, perhaps just add a `TODO` comment in that file saying that we should replace `CardinalityRepresentative` with `Fin`?

mikeshulman:
Yes, I got `Unable to unify "(Unit -> morphism D s d)%type" with "forall c : NatCategory.Core.nat_category 1, morphism D (!s c) (!d c)".` on line 115 of `Comma/InducedFunctors.v`.

JasonGross:
I suspect I used `unit_name`, and this will have to be replaced with `fun _ : Empty + Unit => ...` or whatever.

JasonGross:
Erm, ok, actually looking at it, it's just a "replace `exists (fun _ : Unit => m);` with `exists (fun _ : Fin 1 => m);`".  I suspect all other changes necessary will be similarly simple.

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,904 @@
...
+
+Definition detachable_finite_subset {X} `{Finite X}
+           (P : X -> Type) `{forall x, IsHProp (P x)}
+           {Pf : Finite ({ x:X & P x })}
+: forall x, Decidable (P x).
+Proof.
+  intros x.
+  refine (decidable_equiv _ (hfiber_fibration x P)^-1 _).
+  refine (detachable_image_finite pr1 x).
+  - assumption.                 (** Why doesn't Coq find this? *)
```
JasonGross: Does `Local Set Typeclasses Dependency Order` fix this?  (Does `exact _` work in place of `assumption` here?)

theories/Spaces/Finite.v
```diff
@@ -0,0 +1,904 @@
...
+  - apply mapinO_pr1; exact _.  (** Why doesn't Coq find this? *)
```
JasonGross: Are you asking, "why doesn't `exact _` pick up the `mapinO_pr1`?", or "why doesn't `apply` do typeclass resolution?"?

theories/Types/Sum.v
```diff
@@ -147,6 +355,74 @@ Definition equiv_functor_sum_r {A A' B : Type} (f : A <~> A')
 : A + B <~> A' + B
   := equiv_functor_sum (f := f) (g := idmap).
 
+(** ** Unfunctoriality on equivalences *)
+
+Definition isequiv_unfunctor_sum_l {A A' B B' : Type}
```
JasonGross: Why `Definition` and not `Instance`?

theories/Spaces/Finite.v
JasonGross: Ah, right.  Oops.  I forgot that `hSet` is not an hset.

theories/Spaces/Finite.v
JasonGross: Do you want to test that and adapt the comment to hat you find?

theories/Spaces/Finite.v
JasonGross: Not really, I think?  I don't think it matters much, unless you rely on the syntactic form of the inferred instance (not just its value up to judgmental equality), which would be terrible practice.

JasonGross:
I've finished reviewing. LGTM modulo responses to my comments.  (Mild preference for replacing the various comments asking questions with their answers.)

mikeshulman:
The next problem in replacing CardinalityRepresentative by Fin is lots of `tt`s in `categories/UniversalProperties` that need to be elements of `Fin 1`.  After that there is 

```
The term
 "Build_IsInitialMorphism Y G (F Y) ((A:AdjunctionUnit F G).1 Y)
    ((A:AdjunctionUnit F G).2 Y)" has type
 "IsInitialMorphism
    {|
    CommaCategory.a := Finite.unit_to_fin1 tt;
    CommaCategory.b := F Y;
    CommaCategory.f := A.1 Y |}" while it is expected to have type
 "IsInitialMorphism initial_morphism__of__adjunction"
```

on line 36 of `categories/Adjoint/UniversalMorphisms/Core`.  And I'm going to stop now because, as I said, I don't have time for it; feel free to do it yourself.

theories/Spaces/Finite.v
mikeshulman: No, and no.

theories/Spaces/Finite.v
mikeshulman: The former.  (Although if you know why `apply` doesn't do typeclass resolution, I'd be interested to hear.)

mikeshulman:
Ok, I think I've fixed/responded to everything.

JasonGross:
LGTM modulo travis.  (Travis was failing with what looked like an error to download a package from cabal due to transient http errors; I've kicked it.)

mikeshulman:
Thanks!  APOE?

By the way, in case you or anyone does feel inclined to rectify the `Fin`/`CardinalityRepresentative` gap, there would certainly be something to be said for changing the other side and having `Fin 1 := Unit` definitionally, and that could even be extended to argue for `Fin 2 := Bool` definitionally.

spitters:
For posterity. There is discussion here. @Barbichu has another ssr library:
http://coq-club.inria.narkive.com/Smup7DeP/announcement-finite-set-library-with-comprehension

PR #688, opened 2015-01-24
Add some tactics for globally-nameless proving style

JasonGross:


mikeshulman:
That's a very long proof of `False -> True`.  (-:  What is the real motivation?

JasonGross:
It's actually a proof of `False -> True -> False -> True`.  I just provided the middle `True` hypothesis explicitly.  It's partially to make sure that the tactics don't stop working, and partially to demonstrate how they're all supposed to be called.  The long type is to show that `hyp` can backtrack across hypotheses.

mikeshulman:
I mean, what's the real motivation for these tactics?

JasonGross:
I am working on redefining `Fin` so that `Fin 1` is judgmentally `Unit` and `Fin 2` is judgmentally `Bool`.  I want to use these tactics in some of my proofs.  (Speaking of which, do you happen to have a saner proof [than this](https://github.com/JasonGross/HoTT/blob/fin-defn/theories/Types/Sum.v#L447) that `Contr A -> Contr A' -> (A + B <~> A' + B') -> (B <~> B')`?  (I suppose it could be generalized to showing that this is true whenever `A` and `A'` are finite and have the same number of elements, maybe?  But I'm using it to prove `nat_eq_fin_equiv` earlier.)

mikeshulman:
That's great that you're working on that!  I'll have a look at your lemma.  But I was looking more for a simple example of a proof that's made easier/better by these tactics.  Your sample doesn't seem to me to be an improvement over `intros x y z; exact y.`.

JasonGross:
Here is an example from the code I'm writing:
```coq
(repeat let H := hyp in destruct H; try reflexivity); reflexivity
```
In the `match`-syle, this would be
```coq
repeat match goal with
         | [ H : _ |- _ ] => destruct H; try reflexivity
       end;
reflexivity
```

JasonGross:
Here's another one:
```coq
  repeat first [ reflexivity
               | let IHn := hyp in rewrite IHn
               | let IHn := hyp in let k := hyp in rewrite (IHn k)
               | let IHn := hyp in let l := hyp in rewrite (IHn _ l)
               | progress cbn
               | intro
               | rewrite red_equiv_fin_sum_fin_lift'
               | rewrite unfold_fin_transpose_last_with
               | let H := hyp in destruct (H idpath : False)
               | let H := hyp in enforce (H : Unit); destruct H
               | let H := hyp in enforce (H : Bool); destruct H
               | let H := hyp in enforce (H : _ + _); destruct H ].
```
and the `match` version:
```coq
  repeat match goal with
           | _ => reflexivity
           | [ IHn : _ |- _ ] => rewrite IHn
           | [ IHn : _, k : _ |- _ ] => rewrite (IHn k)
           | [ IHn : _, l : _ |- _ ] => rewrite (IHn _ l)
           | _ => progress cbn
           | _ => intro
           | _ => rewrite red_equiv_fin_sum_fin_lift'
           | _ => rewrite unfold_fin_transpose_last_with
           | [ H : _ |- _ ] => destruct (H idpath : False)
           | [ H : Unit |- _ ] => destruct H
           | [ H : Bool |- _ ] => destruct H
           | [ H : _ + _ |- _ ] => destruct H
         end.
```

mikeshulman:
So they are just a sort of syntactic sugar for a certain kind of `match goal` statement?

JasonGross:
Yes.

mikeshulman:
Okay.  Maybe you could include a more illustrative example then (and maybe also avoid using `Coq.Logic.True`)?

mikeshulman:
As for your unfunctor lemma, [here](https://github.com/mikeshulman/HoTT/compare/indecomposable) is how I would prove it, generalizing from contractible to "indecomposable" `A`, which includes all 0-connected types.  Is that saner?

Alizter:
@JasonGross Did you ever get Fin and Unit and Bool to be judgementally equal?

JasonGross:
I don't recall, sorry

mikeshulman:
Apparently this never got merged, but I don't know why.  Do you still want it?

JasonGross:
It never got merged because I never got around to implementing the requested change:
> Okay. Maybe you could include a more illustrative example then (and maybe also avoid using Coq.Logic.True)?

I'm not currently actively developing things here, so don't currently need the tactics, but I still think it might be neat to have them available.

mikeshulman:
Well, if you feel like resolving the conflicts, I'd be okay with merging it even without another example.

Alizter:
@JasonGross There seems to be a missing `end` in Nameless.v

theories/Tactics/Nameless.v
```diff
@@ -0,0 +1,25 @@
+(** * Building blocks for a globally nameless style of tactic reasoning *)
+
+(** ** [hyp] returns any hypothesis, with subsequent failures backtracking through all hypotheses *)
+Ltac hyp := multimatch goal with H : _ |- _ => constr:H end.
```
Alizter:
```suggestion
Ltac hyp := multimatch goal with H : _ |- _ => constr:(H) end.
```

Alizter:
My suggested change should fix this now.

PR #690, opened 2015-01-27
Actually install all of the symlinks to stdlib

JasonGross:
Previously, we had separate logic in Makefile.am and configure.ac for
dealing with symlinks to the standard library.  Since no one was testing
the install target regularly, they had gotten out of sync.  This was
causing hoqide to fail when the library was installed.

Now, the configure script makes a list of symlinks for the makefile to
copy over in the install target.  This should fix hoqide.  Additionally,
the install target should now work fine on (Windows) configurations
where symlinks don't work.

Finally, we now test each stdlib directory for existence before
symlinking to it.  This prevents us from installing dangling symlinks to
the system, which wouldn't serve any purpose.

Also, test the install target on travis.

I don't actually have an install with hoqide though, so someone with hoqide (@spitters?) should test this before it's merged.  The motivation for this PR is to fix the opam install (also untested), which @spitters is working on.
I propose that if this passes travis, and @spitters or someone else verifies that it fixes a `make install`ed hoqide, and there are no objections to the (idea of the) change, then it should be merged (independently of if anyone wants to actually read the autogoo changes I've made.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
 # `make-pretty-timed` is not an indication that the build succeeded.
 # We also use `make strict-test` at the beginning so that we error
 # quickly if someone commited .v files but neglected to mention them.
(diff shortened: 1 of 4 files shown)


mikeshulman:
Sounds good to me.

spitters:
opam reinstall coq:\hott.0.9.0

hoqide   [works! ]
Require Import HoTT
Also works!

I'll post further opam instructions on the wiki in due time.

Two more small things (that really don't belong here).
- Are we compatible with 8.5beta ?

PR #693, opened 2015-01-28
use Existing Class to get rid of some '_internal's

mikeshulman:
With `Existing Class In` we can get rid of the need for most of the `_internal` fields for reflective subuniverses and modalities.  Unfortunately, I wasn't able to get rid of `O_ind_internal` and `O_ind_beta_internal` for modalities, since we need to have them already in order to prove that modalities are reflective subuniverses so that we can refer to the correct typeclass `In`.  (Suggestions are welcome...)


JasonGross:
LGTM.  I should have noticed this when code reviewing modalities originally.

mikeshulman:
Is `Existing Class` a relatively new command?  Or did I just not notice it until recently?

JasonGross:
It's in 8.4.  Might even be in 8.3 and earlier.

JasonGross:
It's listed in [the vernacular index for 8.4](https://coq.inria.fr/distrib/8.4/refman/command-index.html).

PR #695, opened 2015-01-29
Also install .glob files

JasonGross:
This allows hyperlinking in pide to work correctly.

I've replaced the list of `*_VOFILES` with the single `ALL_VOFILES`; they expand to the same thing, and this one is less typing.


mikeshulman:
LGTM

PR #696, opened 2015-01-29
Bump Coq version to 8.5-β₁

JasonGross:
We should probably just track actual versions from now on, until and
unless we decide that we need some killer new feature in trunk.

(Anyone want to make bets on whether it'll be [feature
3927](https://coq.inria.fr/bugs/show_bug.cgi?id=3927), which would bring
Coq closer to Andromeda, Bruno's native HITs, or something else?)

(I hope nothing breaks; I'm relying on Travis to test this for me.)


vladimirias:
Speaking of features. It would be great if we started thinking about adding to Coq a feature of “user justified cast”.

At the first step it would be just “Cast Axiom” - a version of an axiom that says that a cast of an object X (with parameters) to a type T (with parameters) is permitted. Cast axiom will have a name and will have to be referred by name when used.

Later we would think of adding “Cast Theorems” - cast axioms that have been justified according to some justification principles that have been confirmed to be consistent and/or not affect computation.

Vladimir.

spitters:
This looks like CoqMT, or perhaps deduction modulo, doesn't it?

spitters:
Did you intend to move to the 8.5 branch? It looks like we are still on trunk.

JasonGross:
They've been merging 8.5 into trunk regularly.  I bumped us to the
"v8.5beta1" tag (or whatever tag looked roughly like that).

mikeshulman:
And a submodule doesn't point to a branch anyway but only to a particular commit, right?

JasonGross:
> And a submodule doesn't point to a branch anyway but only to a particular
commit, right?

Right.

PR #702, opened 2015-02-01
The surreal numbers

mikeshulman:
Although Coq doesn't support inductive-inductive types natively, I think I found a way of tricking it into defining the higher inductive-inductive surreals with a sort-of judgmental computation law on the point-constructor.  (I say "sort of" because the proof of cut-ness turns out to be a different axiom than the one one that you might hope it to be, but the options are judgmentally what you expect.)  If anyone can see a flaw in this definition (i.e. a way to escape from the trick and prove something that shouldn't be possible), please point it out.

I am somewhat astounded that Theorem 11.6.7 from the book turned out to be basically correct.  (-:O  And in retrospect, I'm not surprised that the definition of surreal addition was too simplistic and needed an enhanced double induction similar to 11.6.7; see book PR [762](https://github.com/HoTT/book/pull/762).

This file is kind of slow, but I already used all the universe annotations and abstraction barriers I could think of, just to get it to compile without running out of memory.  Suggestions are welcome.  (There are also a couple of failures of typeclass inference that I don't understand; maybe they are bugs?)

### DIFF
--- a/theories/Spaces/No.v
+++ b/theories/Spaces/No.v
@@ -22,7 +22,7 @@ Module Export Surreals.
 
   (** *** Games first *)
 
-  (** Since Coq doesn't support inductive-inductive types natively, we have to hack a bit.  Inspired by Conway, we define [Game]s to be constructed by the point-constructor of [No] but without the hypothesis on inequality of options.  Then we define the inequalities as a mutual inductive family over [Game], and put an inductive predicate on [Game] characterizing those that are Numbers.  We then proceed to add axioms for the path-constructors of [No].
+  (** Since Coq doesn't support inductive-inductive types natively, we have to hack a bit.  Inspired by Conway, we define [Game]s to be constructed by the point-constructor of [No] but without the hypothesis on inequality of options.  Then we define the inequalities as a mutual inductive family over [Game], and put an inductive predicate on [Game] characterizing those that are Numbers.  (This is roughly a standard technique described by Fredrik Nordvall Forsberg for reducing induction-induction to parametrized induction.)  We then proceed to add axioms for the path-constructors of [No].
 
   It should be emphasized that this is *not* currently a correct higher inductive-inductive definition of games; these "games" are only being used inside this module as a trick to produce [No] in a way that computes on the point-constructor.  It should be possible to make a higher inductive-inductive definition of games, but this is not it. *)
 
@@ -206,9 +206,19 @@ Module Export Surreals.
                      (dcut _ _ xL xR xcut fxL fxR fxcut)
                      (dcut _ _ yL yR ycut fyL fyR fycut)).
 
-    (** Technically, we induct over the inductive predicate witnessing Numberhood of games. *)
-    Local Fixpoint No_ind_internal (x : Game) (xno : is_surreal x)
-          {struct xno}
+    (** As usual for HITs implemented with [Private Inductive], we will define [No_ind] inside this module using a fixpoint over [No], thereby obtaining a judgmental computation rule for the point-constructor [No_cut], and then assert the other computation rules as axioms.  In this case, the relevant other rules are the preservation of inequalities.
+
+However, it turns out that in defining [No_cut] we already need to know that it preserves inequalities.  Since this is eventually an axiom anyway, we could just assert it with [admit] in the proof.  However, if we did this then the [admit] would not be *judgmentally* equal to the axiom [No_ind_lt] that we assert afterwards.  Instead, we make use of the fact that [admit] is essentially by definition [match proof_admitted with end] for a global axiom [proof_admitted : Empty], so that if we use the same [admit] both inside the definition of [No_ind] and in asserting [No_ind_lt] as an axiom, they will be the same term judgmentally.
+
+Finally, for conceptual isolation, and so as not to depend on the particular implementation of [admit], we introduce here local copies of [Empty] and [proof_admitted]. *)
+    Local Inductive No_Empty_for_admitted := .
+    Axiom No_Empty_admitted : No_Empty_for_admitted.
+
+    (** Technically, we induct over the inductive predicate witnessing Numberhood of games.  We prove the "induction step" separately to improve performance, possibly by preventing bare [fix]s from appearing upon simplification. *)
+    Local Definition No_ind_internal_step
+          (No_ind_internal : forall (x : Game) (xno : is_surreal x),
+                               A (Build_No x xno))
+          (x : Game) (xno : is_surreal x)
     : A (Build_No x xno).
     Proof.
       revert ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
@@ -218,8 +228,15 @@ Module Export Surreals.
                    (fun r => Build_No (xR r) (Rno r)) xcut _ _ _).
       - intros l; exact (No_ind_internal (xL l) (Lno l)).
       - intros r; exact (No_ind_internal (xR r) (Rno r)).
-      (** The following goal ought to be proven with [No_ind_lt], below.  Unfortunately, we can't state that axiom until *after* we've defined this function!  So instead we make it a different axiom.  This unfortunately means that the computation rule of [No_ind] is not as judgmental as we would like; see below. *)
-      - admit.
+      - intros; exact (match No_Empty_admitted with end).
+    Defined.
+
+    Local Fixpoint No_ind_internal (x : Game) (xno : is_surreal x)
+          {struct xno}
+    : A (Build_No x xno).
+    Proof.
+      destruct xno.
+      exact (No_ind_internal_step No_ind_internal _ _).
     Defined.
 
     Definition No_ind (x : No) : A x.
@@ -228,11 +245,13 @@ Module Export Surreals.
       exact (No_ind_internal x xno).
     Defined.
 
-    Axiom No_ind_le : forall (x y : No) (p : x <= y),
-                        dle x y p (No_ind x) (No_ind y).
+    Definition No_ind_le (x y : No) (p : x <= y)
+    : dle x y p (No_ind x) (No_ind y)
+      := match No_Empty_admitted with end.
 
-    Axiom No_ind_lt : forall (x y : No) (p : x < y),
-                        dlt x y p (No_ind x) (No_ind y).
+    Definition No_ind_lt (x y : No) (p : x < y)
+    : dlt x y p (No_ind x) (No_ind y)
+      := match No_Empty_admitted with end.
 
     (** Sometimes it's convenient to have all three parts of [No_ind] in one package, so that we only have to verify the hypotheses once. *)
     Definition No_ind_package
@@ -247,17 +266,15 @@ Module Export Surreals.
       (forall (p : x < y), dlt x y p (No_ind x) (No_ind y))
       := (No_ind_le x y , No_ind_lt x y).
 
-  (** Our definition ensures that [No_ind] computes on cuts to a cut, but it doesn't compute to quite what you would expect, since we had to assert the axioms [No_ind_lt] and [No_ind_le] *after* defining it.  So the options are what you think they should be, but the proof of cut-ness isn't, at least not judgmentally.  The following lemma observes that propositionally, at least, this doesn't matter. *)
-  Definition No_ind_cut `{Funext}
-             (L R : Type@{i}) (xL : L -> No) (xR : R -> No)
-             (xcut : forall (l:L) (r:R), (xL l) < (xR r))
-  : No_ind {{ xL | xR // xcut }}
-    = dcut L R xL xR xcut
-           (fun l => No_ind (xL l)) (fun r => No_ind (xR r))
-           (fun l r => No_ind_lt (xL l) (xR r) (xcut l r)).
-    Proof.
-      cbn; apply ap. apply path_ishprop.
-    Qed.
+    (** We verify that our definition computes judgmentally. *)
+    Definition No_ind_cut `{Funext}
+               (L R : Type@{i}) (xL : L -> No) (xR : R -> No)
+               (xcut : forall (l:L) (r:R), (xL l) < (xR r))
+    : No_ind {{ xL | xR // xcut }}
+      = dcut L R xL xR xcut
+             (fun l => No_ind (xL l)) (fun r => No_ind (xR r))
+             (fun l r => No_ind_lt (xL l) (xR r) (xcut l r))
+      := 1.
 
   End NoInd.
 
@@ -298,6 +315,7 @@ Defined.
 (** *** The non-dependent recursion principle *)
 
 Section NoRec.
+  Context `{Funext}.
 
   Context  (A : Type)
            (dle : A -> A -> Type) `{is_mere_relation A dle}
@@ -379,17 +397,6 @@ Section NoRec.
       (forall (x y : No) (p : x <= y), dle (f x) (f y)) *
       (forall (x y : No) (p : x < y), dlt (f x) (f y)) }
     := ( No_rec ; (No_rec_le , No_rec_lt) ).
-
-  Definition No_rec_cut `{Funext}
-             (L R : Type@{i}) (xL : L -> No) (xR : R -> No)
-             (xcut : forall (l:L) (r:R), (xL l) < (xR r))
-  : No_rec {{ xL | xR // xcut }}
-    = dcut L R xL xR xcut
-           (fun l => No_rec (xL l)) (fun r => No_rec (xR r))
-           (fun l r => No_rec_lt (xL l) (xR r) (xcut l r)).
-    Proof.
-      cbn; apply ap. apply path_ishprop.
-    Qed.
 
 End NoRec.
 
@@ -578,7 +585,7 @@ Section NoCodes.
                    [[x_le_z x_lt_z] ?]
                    p q;
             cbn; intros [p1 p2] [q1 q2];
-            rewrite transport_sigma'; cbn;
+            rewrite transport_sigma'; (* cbn; *)
             refine (path_sigma' _
                       (path_prod' (path_hprop (equiv_iff_hprop p1 q1))
                                   (path_hprop (equiv_iff_hprop p2 q2)))
@@ -646,6 +653,7 @@ Section NoCodes.
 
   End Inner.
 
+  (** We instruct [simpl]/[cbn] not to unfold [inner].  We will do the "unfolding" ourselves by rewriting along [inner_cut_le] and [inner_cut_lt], so as to keep better control over the resulting terms (and particularly their size).  *)
   Arguments inner : simpl never.
 
   Definition No_codes_package
@@ -895,8 +903,6 @@ Section Addition.
 
   Section Inner.
 
-Set Printing Universes.
-
     Context {L R : Type@{i} } (xL : L -> No@{i}) (xR : R -> No@{i})
             (xcut : forall (l : L) (r : R), xL l < xR r).
 
@@ -1003,8 +1009,8 @@ Set Printing Universes.
     Proof.
       (** Now we tell Coq that we want the equality to be definitional, and let it figure out what the proof of cut-ness has to be. *)
       eexists.
-      (** TODO: The following line takes over 1/3 of the compilation time of this entire file.  It would be nice to speed it up. *)
-      reflexivity.
+      (** Adding [rel_hnf] here speeds things up considerably, possibly because it puts the terms in a form where the evar can be instantiated without unfolding or reduction, preventing backtracking across the evar instantiation. *)
+      rel_hnf. reflexivity.
     Qed.
 
   End Inner.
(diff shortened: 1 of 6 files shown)


theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+    Local Definition inner_le y z p : A'le y z p (inner y) (inner z)
+      := fst (inner_package.2) y z p.
+
+    Local Definition inner_lt y z p : A'lt y z p (inner y) (inner z)
+      := snd (inner_package.2) y z p.
+
+  End Inner.
+
+  Arguments inner : simpl never.
+
```
spitters: This could use some documentation

spitters:
Could you explain a bit more what you are doing? 
Fredrik Nordvall Forsberg shows how to reduce induction induction to parametrized induction. (In fact, the method is older and also works for some instances of induction recursion.) This seems to be what you are doing. This whole construction is done in a `Module` using private types. We then get the right equality outside the module as usual.
Does that seem correct?

JasonGross:
Taking a look at timing, here is the tail-end of the expense list:
```
0.206 secs (0.208u,0.s) Chars 27536 - 27877 [Definition~inner_cut_le~(L'~R'...]
0.21 secs (0.072u,0.136s) Chars 137 - 168 [Require~Import~hit.Truncations...]
0.213 secs (0.212u,0.s) Chars 48915 - 48965 [all:rewrite~No_rec_cut;~[~refl...]
0.265 secs (0.264u,0.s) Chars 42928 - 42932 [Qed.]
0.269 secs (0.272u,0.s) Chars 22721 - 22725 [Qed.]
0.292 secs (0.292u,0.s) Chars 34405 - 34586 [refine~(No_ind_le_lt~(fun~_~=>...]
0.367 secs (0.364u,0.004s) Chars 25352 - 25808 [abstract~(intros~y~z~((x_le_y,...]
0.425 secs (0.424u,0.s) Chars 25817 - 26599 [abstract~(cbn;~intros~L'~R'~yL...]
0.458 secs (0.46u,0.s) Chars 48974 - 48987 [End~Addition.]
0.462 secs (0.464u,0.s) Chars 24133 - 24223 [refine~(No_ind_package~A'~A'le...]
0.735 secs (0.736u,0.s) Chars 30209 - 31044 [abstract~(intros~L~R~xL~xR~xcu...]
0.828 secs (0.808u,0.02s) Chars 29700 - 30202 [abstract~(intros~(x_le,~(x_lt,...]
1.822 secs (1.816u,0.004s) Chars 33238 - 33864 [Definition~le'_cut~(L~R~:~Type...]
2.42 secs (2.42u,0.s) Chars 48968 - 48972 [Qed.]
3.083 secs (3.076u,0.004s) Chars 18576 - 18590 [cbn;~apply~ap.]
3.514 secs (3.508u,0.008s) Chars 33868 - 34296 [Definition~lt'_cut~(L~R~:~Type...]
13.492 secs (13.488u,0.004s) Chars 42911 - 42923 [reflexivity.]
```
The lines that stand out to me are:
```
0.458 secs (0.46u,0.s) Chars 48974 - 48987 [End~Addition.]
1.822 secs (1.816u,0.004s) Chars 33238 - 33864 [Definition~le'_cut~(L~R~:~Type...]
2.42 secs (2.42u,0.s) Chars 48968 - 48972 [Qed.]
3.083 secs (3.076u,0.004s) Chars 18576 - 18590 [cbn;~apply~ap.]
3.514 secs (3.508u,0.008s) Chars 33868 - 34296 [Definition~lt'_cut~(L~R~:~Type...]
13.492 secs (13.488u,0.004s) Chars 42911 - 42923 [reflexivity.]
```
I shall look into them.

mikeshulman:
@JasonGross thanks.  I think all of those lines except for `End Addition` and `Qed` involve Coq having to evaluate a `No_ind` sufficiently to verify that it's equal to something else or has a specific form.

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+    : A (Build_No x xno).
+    Proof.
+      revert ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
+      destruct xno as [L R xL xR Lno Rno xcut].
+      intros ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
+      refine (dcut L R (fun l => Build_No (xL l) (Lno l))
+                   (fun r => Build_No (xR r) (Rno r)) xcut _ _ _).
+      - intros l; exact (No_ind_internal (xL l) (Lno l)).
+      - intros r; exact (No_ind_internal (xR r) (Rno r)).
+      (** The following goal ought to be proven with [No_ind_lt], below.  Unfortunately, we can't state that axiom until *after* we've defined this function!  So instead we make it a different axiom.  This unfortunately means that the computation rule of [No_ind] is not as judgmental as we would like; see below. *)
```
JasonGross: Here is a trick that might let you do this: rather than just constructing the function as a fixpoint, see if you can define a type of all functions that compute the right thing (e.g., the type of initial objects in whatever algebra-diagram, or however fixpoints are defined mathematically), and pass in the axiom as an argument applying to anything of the right type.  For example, I think this is most of the way to being able to pass the axiom as an argument, though it doesn't say enough about the recursive case (and to actually do that, you may need to switch to well-founded recursion):
```coq
    Definition No_ind_internal_step
               (No_ind_internal : forall (x : Game) (xno : is_surreal x), A (Build_No x xno))
               (x : Game) (xno : is_surreal x)
               (final_case : let (L, R, xL, xR, Lno, Rno, xcut) := xno in
                             forall (l : L) (r : R),
                               dlt {| game_of := xL l; isno_game_of := Lno l |}
                                   {| game_of := xR r; isno_game_of := Rno r |} (xcut l r)
                                   (No_ind_internal (xL l) (Lno l)) (No_ind_internal (xR r) (Rno r)))
    : A (Build_No x xno).
    Proof.
      revert ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
      destruct xno as [L R xL xR Lno Rno xcut].
      intros ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
      refine (dcut L R (fun l => Build_No (xL l) (Lno l))
                   (fun r => Build_No (xR r) (Rno r)) xcut _ _ _).
      - intros l; exact (No_ind_internal (xL l) (Lno l)).
      - intros r; exact (No_ind_internal (xR r) (Rno r)).
      - apply final_case.
    Defined.

    Definition No_ind_internal_step'
               (No_ind_internal
                : forall (x : Game) (xno : is_surreal x),
                    { ret : A (Build_No x xno)
                    | exists rec fin, (@No_ind_internal_step rec x xno fin = ret) })
               (x : Game) (xno : is_surreal x)
               (final_case : let No_ind_internal' x xno := @No_ind_internal_step (No_ind_internal x xno).2.1 x xno (No_ind_internal
x xno).2.2.1 in
                             let (L, R, xL, xR, Lno, Rno, xcut) := xno in
                             forall (l : L) (r : R),
                               dlt {| game_of := xL l; isno_game_of := Lno l |}
                                   {| game_of := xR r; isno_game_of := Rno r |} (xcut l r)
                                   (No_ind_internal' (xL l) (Lno l)) (No_ind_internal' (xR r) (Rno r)))
    : { ret : A (Build_No x xno)
      | exists rec fin, (@No_ind_internal_step rec x xno fin = ret) }.
    Proof.
      destruct xno.
      refine (@No_ind_internal_step (fun x xno => (No_ind_internal x xno).1) _ _ _; _).
      { intros l r.
        specialize (final_case l r); cbn in final_case.
        edestruct No_ind_internal as [rec [rec' [fin H]]].
        edestruct No_ind_internal as [rec1 [rec1' [fin1 H1]]].
        cbn in *.
        destruct H, H1.
        exact final_case. }
      { match goal with
          | [ |- exists rec fin, @No_ind_internal_step rec ?x ?xno fin = @No_ind_internal_step ?rec' ?x ?xno ?fin' ]
            => exists rec';
              exists fin';
              reflexivity
        end. }
    Defined.
    Local Fixpoint No_ind_internal (x : Game) (xno : is_surreal x)
          {struct xno}
    : { ret : A (Build_No x xno)
      | exists rec fin, (@No_ind_internal_step rec x xno fin = ret) (* * (ret = rec x xno) *) }.
    Proof.
      destruct xno.
      refine (@No_ind_internal_step' No_ind_internal _ _ _).
      cbn.
```

It's not clear whether it's worth it to jump through all these hoops...

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+      let zR := sum_ind (fun _ => No)
+                        (fun r => (xR_plus r).1 {{ yL | yR // ycut }})
+                        (fun r => (plus_inner.1 (yR r)).1)
+                : R'' -> No in
+      { zcut : forall (l:L'') (r:R''), zL l < zR r &
+        (plus_inner.1 {{ yL | yR // ycut }}).1 = {{ zL | zR // zcut }} }.
+    Proof.
+      (** Now we tell Coq that we want the equality to be definitional, and let it figure out what the proof of cut-ness has to be. *)
+      eexists.
+      (** TODO: The following line takes over 1/3 of the compilation time of this entire file.  It would be nice to speed it up. *)
```
JasonGross: Here is a fix that drops it down from 13.5 s to 0.16 s:  Add this tactic to `Basics/Overture`:
```coq
(** [expand], when given a goal of the form [R x y] for any relation [R], puts [x] and [y] in head normal form *)
Ltac expand :=
  idtac;
  match goal with
    | [ |- ?R ?x ?y ] => let x' := (eval hnf in x) in
                         let y' := (eval hnf in y) in
                         change (R x' y')
  end.
```
and then use `expand; reflexivity` instead of `reflexivity`.

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+    := ( No_rec ; (No_rec_le , No_rec_lt) ).
+
+  Definition No_rec_cut `{Funext}
+             (L R : Type@{i}) (xL : L -> No) (xR : R -> No)
+             (xcut : forall (l:L) (r:R), (xL l) < (xR r))
+  : No_rec {{ xL | xR // xcut }}
+    = dcut L R xL xR xcut
+           (fun l => No_rec (xL l)) (fun r => No_rec (xR r))
+           (fun l r => No_rec_lt (xL l) (xR r) (xcut l r)).
+    Proof.
```
JasonGross: This `Proof.` (and the subsequent proof code and `Qed`) seems to be indented one too many levels.

JasonGross:
Here is a change that cuts the file down from 33.4 s to 16.4 s:
```diff
diff --git a/theories/Spaces/No.v b/theories/Spaces/No.v
index 5d4043c..0a794da 100644
--- a/theories/Spaces/No.v
+++ b/theories/Spaces/No.v
@@ -207,8 +207,9 @@ Module Export Surreals.
                      (dcut _ _ yL yR ycut fyL fyR fycut)).

     (** Technically, we induct over the inductive predicate witnessing Numberhood of games. *)
-    Local Fixpoint No_ind_internal (x : Game) (xno : is_surreal x)
-          {struct xno}
+    Local Definition No_ind_internal_step
+          (No_ind_internal : forall (x : Game) (xno : is_surreal x), A (Build_No x xno))
+          (x : Game) (xno : is_surreal x)
     : A (Build_No x xno).
     Proof.
       revert ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
@@ -222,6 +223,14 @@ Module Export Surreals.
       - admit.
     Defined.

+    Local Fixpoint No_ind_internal (x : Game) (xno : is_surreal x)
+          {struct xno}
+    : A (Build_No x xno).
+    Proof.
+      destruct xno.
+      exact (No_ind_internal_step No_ind_internal _ _).
+    Defined.
+
     Definition No_ind (x : No) : A x.
     Proof.
       destruct x as [x xno].
```
(I think it makes `cbn` behave a bit better, because it means that there will ~never be a bare `fix` floating around.)
The tail of the timing profile after this change is:
```
0.209 secs (0.072u,0.132s) Chars 137 - 168 [Require~Import~hit.Truncations...]
0.221 secs (0.22u,0.s) Chars 18874 - 18888 [cbn;~apply~ap.]
0.228 secs (0.228u,0.s) Chars 49203 - 49253 [all:rewrite~No_rec_cut;~[~refl...]
0.233 secs (0.232u,0.s) Chars 34693 - 34874 [refine~(No_ind_le_lt~(fun~_~=>...]
0.259 secs (0.26u,0.s) Chars 43216 - 43220 [Qed.]
0.267 secs (0.268u,0.s) Chars 33526 - 34152 [Definition~le'_cut~(L~R~:~Type...]
0.361 secs (0.36u,0.s) Chars 25650 - 26106 [abstract~(intros~y~z~((x_le_y,...]
0.424 secs (0.424u,0.s) Chars 26115 - 26897 [abstract~(cbn;~intros~L'~R'~yL...]
0.425 secs (0.424u,0.s) Chars 49262 - 49275 [End~Addition.]
0.466 secs (0.464u,0.s) Chars 24431 - 24521 [refine~(No_ind_package~A'~A'le...]
0.509 secs (0.512u,0.s) Chars 34156 - 34584 [Definition~lt'_cut~(L~R~:~Type...]
0.671 secs (0.672u,0.s) Chars 30497 - 31332 [abstract~(intros~L~R~xL~xR~xcu...]
0.688 secs (0.688u,0.s) Chars 43199 - 43211 [reflexivity.]
0.733 secs (0.724u,0.008s) Chars 29988 - 30490 [abstract~(intros~(x_le,~(x_lt,...]
2.335 secs (2.336u,0.s) Chars 49256 - 49260 [Qed.]
4.061 secs (4.06u,0.s) Chars 49114 - 49134 [all:try~reflexivity.]
```

JasonGross:
Adding the `expand; reflexivity` on top of this change brings it down to 15.5 s.

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+                (fun l => plus_outer.1 (xL l))
+                (fun r => plus_outer.1 (xR r))
+                (fun l r => snd plus_outer.2 (xL l) (xR r) (xcut l r))
+                yL yR ycut) as [zcut p].
+    eexists.
+    refine (_ @ p @ _); clear p.
+    - unfold plus, plus_outer, No_rec_package; cbn [pr1].
+      rewrite No_rec_cut.
+      + reflexivity.
+      + (** TODO: Why doesn't Coq find this?  (Also below.) *)
```
JasonGross: Because the `Funext` argument is not sufficiently implicit at this point (there are non-implicit arguments before it).  If you had done `rewrite (@No_rec_cut _ _ _ _ _ _ _ _ _ _ _).`, then it would find it automatically.  If you reorganize your sections/definitions so that `Funext` comes first in `No_rec_cut`, then Coq will also find it automatically.  (This is at least part of why it's good practice to put all the typeclass arguments as early as possible.  Should this go in the style file?)

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+End NoCodes.
+
+(** ** Addition *)
+
+Section Addition.
+  Context `{Univalence}.
+
+  Section Inner.
+
+Set Printing Universes.
```
JasonGross: Did you actually mean to keep this in here?

JasonGross:
Replacing `all:try reflexivity` (2.7 s in interactive mode for me, after both changes above) with `all:[> reflexivity | | reflexivity | ]` (0.006 s in interactive mode) cuts the time from 15.5 s total to 11.5 s total (in batch mode).

theories/Spaces/No.v
spitters: The name expand is too generic for this tactic. Is it possible to make a general argument hnf_in_all_args ?

theories/Spaces/No.v
spitters: Yes, that should go in the style file.

theories/Spaces/No.v
JasonGross: It's specifically meant for goals of the form `_ = _` (you could in fact change `?R ?x ?y` to `?x = ?y` and I'd be fine).  I'd prefer a relatively shortish name for this, as it seems like a common operation.  Maybe `path_hnf` or `rel_hnf` or something?

JasonGross:
Here is a hack, which depends on the implementation of `admit`, that makes `No_ind_cut` definitional: replace `admit.` in `No_ind_internal` with `intros; admit.`, and replace `Axiom No_ind_lt : ...` with
```coq
    Definition No_ind_lt : forall (x y : No) (p : x < y),
                             dlt x y p (No_ind x) (No_ind y).
    Proof.
      intros; admit.
    Defined.
```
(Note that `admit` is easy to (re)implement without OCaml.  It is:
```coq
Axiom proof_admitted : Empty.
Ltac admit := exact (match proof_admitted with end).
```
So if the implementation of `admit` changes, we can reimplement the variant of `admit` we need that gives us this property.)

theories/Spaces/No.v
spitters: `rel_hnf` sounds fine.

spitters:
I don't really like the admit hack, but would agree with it, if properly documented.

JasonGross:
Using the `admit` trick and defining `plus_cut` as `plus_inner_cut` drops the time down to 8.5 s.

theories/Spaces/No.v
```diff
@@ -0,0 +1,1149 @@
...
+          | intros y z x_le_y y_lt_z;
+            exact (inner_lt xL_let xR_let x_lt_le y z y_lt_z x_le_y)
+          | intros y z x_lt_y y_le_z;
+            exact (snd (inner_le xL_let xR_let x_lt_le y z y_le_z) x_lt_y) ]).
+    - abstract (
+          intros [x_le [x_lt ?]] [x_le' [x_lt' ?]] p q; cbn in p, q;
+          refine (path_sigma' _ _ _);
+          [ apply path_arrow; intros y; apply path_hprop, equiv_iff_hprop;
+            [ exact (fst (q y)) | exact (fst (p y)) ]
+          | rewrite transport_sigma'; cbn;
```
JasonGross: Remove the `cbn` in this line; it accounts for about 0.4 s, a bit over half the time spent in this `abstract`.  (75% of the remaining time is spent in `apply path_ishprop`)

JasonGross: And you can bring that 0.25 s down to 0.05 s by replacing `apply path_ishprop` with `apply @path_ishprop; repeat (intros || apply @trunc_prod || apply @trunc_arrow || apply @trunc_forall); exact _.` (unclear if this is worth it).  Alternatively, you can replace `Global Instance trunc_forall ...` with `Definition trunc_forall ...` and `Hint Extern 1 (IsTrunc _ (forall a, _)) => apply @trunc_forall; intros : typeclass_instances.` to get the same speed-up.  (Yes, the trouble really is with typeclass resolution not being aggressive enough about `intros`.  @mattam82, is there a "suggested" way to correct for this?)

JasonGross:
Travis doesn't like you.  It seems to be hanging on, uh, `XSLTPROC proviola-xml/HoTTBookExercises.xml`?  But that doesn't make any sense, because that's the last one it runs, and then it's supposed to `make timing-html`...  Or maybe travis is barfing on concurency problems?
Can you see if you can `make proviola` locally?  (You'll need python 2.6 and pip, the python package manager, at which point you can run `etc/ci/install_proviola.sh` to get lxml, pexpect-u, and a local copy of proviola.)

theories/Spaces/No.v
mikeshulman: Can you say anything about why `rel_hnf` speeds things up so dramatically, and what led you to try it?

mikeshulman:
I like the `admit` hack.  Here is an idea for a variant: what if we do it with a reimplemented copy of `admit` using a `Local Inductive` version of `Empty`?  That way it's more conceptually isolated, in addition to not depending on the implementation of `admit`.

theories/Spaces/No.v
JasonGross: My guess: It puts the terms in a form where syntactic unification (no unfolding, no reduction) is sufficient to instantiate the evar.  It thus prevents backtracking across the evar instantiation, and probably gives Coq a bit of help by having it not try unification paths that don't end up working out.

Originally, I was trying things like `exact_no_check idpath` and `vm_compute`, to see if I could get it to not do unification twice.  But those left over the evar, so I started looking in to seeing how hard it would be to instantiate the evar manually with tactics.  So I started running`hnf` in both sides, and applying `ap`, etc.

JasonGross:
I'm not sure why you want a separate type for a local version of `Empty`, rather than just a fresh inhabitant of it, but I won't object.  You should end your axiom with `_admitted` or something else on the search blacklist, so that it doesn't show up in `SearchAbout` (see the refman on the search blacklist).

mikeshulman:
Well, it already makes me uneasy to have a global inhabitant of `Empty` floating around, so I was trying to avoid adding another one.  (-:

theories/Spaces/No.v
mikeshulman: My inclination is that the additional obfuscation of this isn't worth 0.2 seconds.

mikeshulman:
Thanks very much for all the suggestions!  I've implemented them all.  I never would have guessed it was possible to get this file all the way down to 9s.

Will now look into proviola.

mikeshulman:
Proviola makes fine on my computer.  And Travis now seems happy too.

spitters:
@mikeshulman It looks like you did not answer this.
>>>
Could you explain a bit more what you are doing?
Fredrik Nordvall Forsberg shows how to reduce induction induction to parametrized induction. (In fact, the method is older and also works for some instances of induction recursion.) This seems to be what you are doing. This whole construction is done in a Module using private types. We then get the right equality outside the module as usual.
Does that seem correct?
>>>

mikeshulman:
@spitters yes, that seems right.  I added a reference to Forsberg in the comment at the top of the file.

spitters:
When I merged Travis was happy. Now it is not:
https://travis-ci.org/HoTT/HoTT/jobs/49252158

File "./theories/Basics/Overture.v", line 134, characters 4-9:
Error: This expression should be a name.

This shouldn't happen, should it?

JasonGross:
It can happen if you merge something that is not fully rebased, and there has since been a merge that changed definitions.

JasonGross:
But you're looking at the wrong one.  That one's in the allowed-to-fail category, and is failing because it's using the ppa version of trunk.  The actual error is at https://travis-ci.org/HoTT/HoTT/jobs/49252155,

etc/DepsToDot.hs:11:8:
    Could not find module `Data.Text.Lazy'
    Perhaps you meant Data.STRef.Lazy (from base)
    Use -v to see a list of the files searched for.

It's a transient error from something failing to download, I think.  I'll kick the build.

spitters:
Thanks

Issue #691, opened 2015-01-27
Equivalence induction


JasonGross:
Given an equivalence `e : A <~> B`, we should be able to write a tactic `equiv_induction e` that behaves as well as `induction p` for `p : A = B`, and leaves a side-condition that the type of the goal respects equivalences.  This side condition should be automatically solvable in most cases, since most of the type-formers respect equivalence.

mikeshulman:
Yes, I also had the thought once that we could define a typeclass for "respects equivalences" to automatically infer those conditions.

andrejbauer:
This was one of the very early things I tried to do in this library. Peter Aczel suggested it do me. The tricky bit is when you get to dependent types. Is this supposed to cover those?

JasonGross:
Yes.  It is hard but doable in 8.4 with typeclasses (but not as a tactic without typeclasses).  It is much more straightforward in 8.4 with dependent subgoals.

mikeshulman:
I think when I tried, I got stuck at dependent types.

spitters:
Andrej, do you still have your experiments available? I think it was a nice
start.

andrejbauer:
> Andrej, do you still have your experiments available? I think it was a nice start.

They would be severly outdated by now, there wasn't all that much to them, but they ought to be in a prehistoric version of this repository, something under `experimantal`, called `DefinableEquivalence` or some such. It's not worth digging, I am sure.

PR #704, opened 2015-02-02
Add a reversed version of [rapply]

JasonGross:
This is necessary for getting #691 to work.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -984,20 +984,26 @@ where they are defined.
   automatically.  If you need a version with more fields than yet
   exists, feel free to add it.)
 
-- `rapply`, `erapply`: Defined in `coq/theories/Program/Tactics` and
-  `Basics/Overture` respectively, these tactics are more well-behaved
-  variants of `apply` for theorems with fewer than 16 arguments.  (It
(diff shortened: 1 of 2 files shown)


mikeshulman:
LGTM

spitters:
Do we need both `rapply` and `rapply'` ? I guess so.
The name is not very descriptive, but I don't know a better one.

Otherwise, LGTM

JasonGross:
Alternative names: `rev_rapply`, `rapply_rev`.  The issue is that when you `rapply` a lemma ending in an equivalence, you need to decide whether you want to apply the function, or the equivalence.  If you make the wrong choice, you can make typeclass resolution loop.

spitters:
I don't think the other names are any better they suggest applying A <- B

Feel free to merge.

mikeshulman:
>  The issue is that when you rapply a lemma ending in an equivalence, you need to decide whether you want to apply the function, or the equivalence. If you make the wrong choice, you can make typeclass resolution loop.

Maybe a comment about that in the source code or the style file would be useful.

JasonGross:
Added some more documentation, and `erapply'`.  Does it still look good to both of you?

spitters:
LGTM

Issue #687, opened 2015-01-19
A notation for equiv_compose


mikeshulman:
`equiv_compose'` is used a lot and merits a notation, perhaps `oE`, or simply `o` if we aren't worried about ambiguity.

JasonGross:
If no one else has a preference, I will express a very slight preference for ovoverloading `o` further.

peterlefanulumsdaine:
How does the overloaded version work in practice — does Coq frequently need the scope disambiguated by hand, or not?  If not, then I also vote for overloading `o`, for human-readability.  If it does, though, I would see a consistent `oE` as slightly nicer than an unpredictable sprinkling of scope delimiters.

(Sorry by the way that I am mostly commenting without contributing directly at the moment!  I’m working on a couple of developments over the library, so am keeping roughly up-to-date on it, but haven’t felt fully enough on top of the current state to be confident of contributing myself.  Very much appreciating all the recent additions, though!)

mikeshulman:
I think we'd have to try it and see.

(Do you think your developments-on-top might eventually be contributed back to the library?)

peterlefanulumsdaine:
(Continuing the parenthetical: at least in part, yes, depending on how they shape up.)

JasonGross:
To give a more systematic answer: overloaded notations tend to be
automatically resolvable when they appear as arguments to functions that
accept only the relevant type (so, equiv_compose could be inferred when it
shows up as an argument to a function that only accepts equivalences).
`function_scope` is an exception; Coq has no facility to bind it
automatically to arguments expecting lambdas.  So it's moderately likely
that if we want to mix function composition with equivalence composition in
the same file at top level, we'll need to manually play with scopes.
(Files not defining equivalences at top level probably won't need manual
scope munging.)

mikeshulman:
I'd kind of rather not take on this change until my open PRs get merged (#631, #674, #686) -- otherwise they'll need to be rebased again.

JasonGross:
I think we can consider this closed, though feel free to reopen if anyone thinks there should be more discussion over the notation.

PR #705, opened 2015-02-02
Add some notations for equivalences (for composition, idmap, and inverse)

JasonGross:
Also, correct equiv_scope.  The new rule for [equiv_scope] is that it
should match [function_scope], [morphism_scope], [type_scope]:
[equiv_scope] is for things that build [Equiv]s (just like [type_scope]
is for things that build [Type]s).

This permits and necessitates removing many [Local Open Scope
equiv_scope]; things that expect [Equiv] now get the notations
automatically, and since [<~>] is in [type_scope] and [^-1] is in
[function_scope], we don't need [equiv_scope] for those.  We also must
not have [equiv_scope] on top of [function_scope], or else we'll pick up
the wrong [o].

### DIFF
--- a/contrib/Freudenthal.v
+++ b/contrib/Freudenthal.v
@@ -8,7 +8,7 @@ Require Import hit.Connectedness hit.Suspension hit.Truncations.
 Import TrM.
 Local Open Scope trunc_scope.
 Local Open Scope path_scope.
-Local Open Scope equiv_scope.
+
 Generalizable Variables X A B f g n.
 
 (* ** Connectedness of the suspension *)
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -26,7 +26,6 @@
 Require Import HoTT Coq.Init.Peano.
 
 Local Open Scope path_scope.
-Local Open Scope equiv_scope.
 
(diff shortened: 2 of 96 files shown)


spitters:
LGTM

mikeshulman:
Thanks!  I notice you haven't yet actually used the notation `o` in place of `equiv_compose'` anywhere.  It would be nice to see how well that works, e.g. as discussed at #687 whether we'll need lots of `%equiv`s to use it.  Are you planning to do that as a separate PR?

JasonGross:
Yes, that's coming in a separate pull request.  See https://github.com/JasonGross/HoTT/compare/more-equiv-notation.

There is some pain caused by the fact that [we cannot `Bind Scope function_scope with Funclass`](https://coq.inria.fr/bugs/show_bug.cgi?id=3080), and so I've added a bunch of `Arguments` commands to make things nicer.  But, in general, I think it is not too bad.  (I would be open to adding `oE` or something, though.)

JasonGross:
Rebased against master, and removed `Open Scope equiv_scope` from `No.v`

mikeshulman:
Ok, we can discuss `o` vs `oE` at the other PR.

PR #708, opened 2015-02-03
More equiv notations

JasonGross:
This is on top of #705.

### DIFF
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -6,8 +6,6 @@ Import TrM.
 Require Import Spaces.Nat.
 
 Local Open Scope path_scope.
-
-
 Local Open Scope nat_scope.
 
 (** * Finite sets *)
@@ -37,7 +35,7 @@ Defined.
 
 Global Instance contr_fin1 : Contr (Fin 1).
 Proof.
-  refine (contr_equiv' Unit (equiv_inverse (sum_empty_l Unit))).
+  refine (contr_equiv' Unit (sum_empty_l Unit)^-1).
 Defined.
 
 Definition fin_empty (n : nat) (f : Fin n -> Empty) : n = 0.
@@ -54,12 +52,9 @@ Defined.
 
 Definition fin_transpose_last_two (n : nat)
 : Fin n.+2 <~> Fin n.+2
-  := (equiv_compose'
-        (equiv_inverse (equiv_sum_assoc _ _ _))
-     (equiv_compose'
-        (equiv_functor_sum' (equiv_idmap _)
-                            (equiv_sum_symm _ _))
-        (equiv_sum_assoc _ _ _))).
+  := (((equiv_sum_assoc _ _ _)^-1)
+        o (1 + (equiv_sum_symm _ _))
+        o (equiv_sum_assoc _ _ _))%equiv.
 
 Arguments fin_transpose_last_two : simpl nomatch.
 
@@ -84,12 +79,10 @@ Proof.
   - destruct n as [|n IH].
     + elim k.
     + destruct k as [k|].
-      * refine (equiv_compose'
-                  (fin_transpose_last_two n)
-                  (equiv_compose' _ (fin_transpose_last_two n))).
-        refine (equiv_functor_sum'
-                  (fin_transpose_last_with n (inl k))
-                  (equiv_idmap Unit)).
+      * refine ((fin_transpose_last_two n)
+                  o _
+                  o (fin_transpose_last_two n))%equiv.
+        refine ((fin_transpose_last_with n (inl k)) + 1)%equiv.
       * apply fin_transpose_last_two.
   - exact (equiv_idmap _).
 Defined.
@@ -176,9 +169,8 @@ Qed.
 Definition fin_equiv (n m : nat)
            (k : Fin m.+1) (e : Fin n <~> Fin m)
 : Fin n.+1 <~> Fin m.+1
-  := (equiv_compose'
-        (fin_transpose_last_with m k)
-        (equiv_functor_sum' e (equiv_idmap Unit))).
+  := ((fin_transpose_last_with m k)
+        o (e + 1))%equiv.
 
 (** Here is the curried version that we will prove to be an equivalence. *)
 Definition fin_equiv' (n m : nat)
@@ -196,9 +188,7 @@ Proof.
   exists y.
   destruct y as [y|[]].
   + refine (equiv_unfunctor_sum_l
-              (equiv_compose'
-                 (fin_transpose_last_with m (inl y))
-                 e)
+              (fin_transpose_last_with m (inl y) o e)
               _ _ ; _).
     { intros a. ev_equiv.
       assert (q : inl y <> e (inl a))
@@ -319,7 +309,7 @@ Proof.
   refine (trunc_equiv' _ (issig_finite X)).
   apply ishprop_sigma_disjoint; intros n m Hn Hm.
   strip_truncations.
-  refine (nat_eq_fin_equiv n m (equiv_compose' Hm (equiv_inverse Hn))).
+  refine (nat_eq_fin_equiv n m (Hm o Hn^-1)).
 Defined.
 
 (** ** Preservation of finiteness by equivalences *)
@@ -342,7 +332,7 @@ Corollary finite_equiv_equiv X Y
 : (X <~> Y) -> (Finite X <~> Finite Y).
 Proof.
   intros ?; apply equiv_iff_hprop; apply finite_equiv';
-    [ assumption | symmetry; assumption ]. 
+    [ assumption | symmetry; assumption ].
 Defined.
 
 Definition fcard_equiv {X Y} (e : X -> Y) `{IsEquiv X Y e}
@@ -422,7 +412,7 @@ Proof.
   refine (Build_Finite _ (fcard X).+1 _).
   pose proof (merely_equiv_fin X).
   strip_truncations; apply tr.
-  refine (equiv_functor_sum' _ (equiv_idmap _)); assumption.
+  refine (_ + 1)%equiv; assumption.
 Defined.
 
 Definition fcard_succ X `{Finite X}
@@ -540,7 +530,7 @@ Proof.
   refine (_ @ nat_plus_comm _ _).
   assert (e := merely_equiv_fin Y).
   strip_truncations.
-  refine (fcard_equiv' (equiv_functor_sum' (equiv_idmap X) e) @ _).
+  refine (fcard_equiv' (1 + e) @ _).
   refine (_ @ ap (fun y => (y + fcard X)) (fcard_equiv e^-1)).
   generalize (fcard Y); intros n.
   induction n as [|n IH].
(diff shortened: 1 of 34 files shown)


spitters:
Nice.

mikeshulman:
Thanks for doing this!  I think it will be a big improvement.  But currently there are too many `%equiv`s and `%function`s and `%path`s for me.  If this is the best we can do with `o`, then I vote for `oE` (and similarly for `^-1` etc.

JasonGross:
@mikeshulman: updated to use `oE`.  I would rather leave `^-1` and `1`.  There are still a few cases where we need explicit scopes:
- When doing anything under binders or inside a pair when in `equiv_scope`; [`Bind Scope` is a bit broken](https://coq.inria.fr/bugs/show_bug.cgi?id=3990)
- When doing `refine (...^-1)` or `exact 1` or `refine (_ + _)` at top-level (this accounts for about 10 `%equiv`s total, which I think is acceptable; the rest are, I think, from the above bullet)

mikeshulman:
I'm not so much concerned about the total *number* of `%equiv`s as about their potential for confusing the newcomer trying to write code.  I'm not really happy about having *any* explicit scope markers...

What would you think about trying to make `^-1` mean *only* `equiv_inverse`, with the underlying function obtained via the `equiv_fun` coercion?  We would probably need to have some wrappers like `eisretr_internal`/`eisretr`, since the fields in `IsEquiv` have to refer to `equiv_inv` directly whereas we'd like the type of `eisretr` to involve `^-1`.

mikeshulman:
Just to be sure I understand the problem: when we say something like `refine (...^-1)`, the issue is that Coq *parses* the term before it pays attention to the type that it's expecting the term to have, so it doesn't know to interpret `^-1` in `equiv_scope`?

Here's a crazy idea: what if we just made `1` a notation for `reflexivity`, and declared `Equiv` to be `Reflexive`?

JasonGross:
Given `f : A -> B` and `H : IsEquiv f`, how are you going to get `f^-1` to coerce `f` into an equivalence.  I think there should be a way to do it with canonical structures; we need to somehow tell Coq that it should try to unify `f` with `equiv_fun _`.  We could do something like this:
```coq
Canonical Structure default_Equiv {A B} := fun f H => @BuildEquiv A B f H.
Notation "e ^-1" := (@equiv_inverse _ _ (@BuildEquiv _ _ e (_ : IsEquiv (equiv_fun _)))) (at level 3, format "e '^-1'") :
equiv_scope.
Definition foo := (fun (f : Type -> Type) (_ : IsEquiv f) => f^-1)%equiv.
Set Printing All.
Print foo.
(* foo =
fun (f : forall _ : Type, Type) (H : @IsEquiv Type Type f) =>
@equiv_inverse Type Type
  (BuildEquiv Type Type f
     (H
      :@IsEquiv Type Type
         (@equiv_fun Type Type (@default_Equiv Type Type f H))))
     : forall (f : forall _ : Type, Type) (_ : @IsEquiv Type Type f),
       Equiv Type Type *)
```
But then we have this strange form, which might cause issues with pattern matching and `simpl`/`cbn`.  We can at least add
```coq
Notation "e ^-1" := (@equiv_inverse _ _ (@BuildEquiv _ _ e _)) (at level 3, format "e '^-1'") : equiv_scope.
```
above, so that things still print as `e^-1` after `cbn`.

Perhaps @beta-ziliani or @spitters can suggest an improvement that doesn't leave us with this extra type annotation/cast?

JasonGross:
> Just to be sure I understand the problem: when we say something like `refine (...^-1)`, the issue is that Coq *parses* the term before it pays attention to the type that it's expecting the term to have, so it doesn't know to interpret `^-1` in `equiv_scope`?

You're thinking wishfully.  Coq does parsing before typechecking, always.  There is no type information around when Coq needs to decide what scope to interpret things in.  The way you fake it is that you say "when you see the constant `foo`, you should interpret its third argument in scope `bar`", and then you tell Coq to do this automatically depending on the types of the arguments, and, viola!, you now have the appearance of notations specific to types, when you're not at top-level, and the types are sufficiently non-dependent (e.g., if `id := fun (T : Type) (x : T) => x`, then `@id Type some_notation` will not be enough to have Coq parse `some_notation` in `type_scope`, because its type is computed, not fixed as an argument to the function at function-definition time----this is why Coq can't figure out good scopes for `x` and `y` in `x = y :> A` even if it knows `A`).

You only ever get a single fixed stack of scopes at top-level.  Any top-level notation is interpreted in the first scope that it exists in.

JasonGross:
> Here's a crazy idea: what if we just made `1` a notation for `reflexivity`, and declared `Equiv` to be `Reflexive`?

I was thinking of suggesting the same thing!  (And `Equiv` already has a `Reflexive` instance.)  Shall I do that?  Or is it important that `1` insert `idpath` exactly (like you wanted `@` to insert `concat` exactly)?

Also, note that this means we can no longer do `match p with 1 => 1 end`, only `match p with idpath => 1 end`.  Is this a problem?

mikeshulman:
Oh, I skipped a step from my head: I was actually thinking of `^-1` being something that takes `f : A -> B` and `H : IsEquiv f` (which we can get from `f : A <~> B` by coercion and instance) and produces a `B <~> A`.

>  Coq does parsing before typechecking, always.

That's what I meant to say.

> Or is it important that `1` insert `idpath` exactly (like you wanted `@` to insert `concat` exactly)?  Also, note that this means we can no longer do `match p with 1 => 1 end`.

Hmm, maybe those mean this is a bad idea.  I'd like to hear more opinions here too, anyone else want to chime in?

JasonGross:
Oh, I see, you just want
```coq
Notation "e ^-1" := (@equiv_inverse _ _ (@BuildEquiv _ _ e _)) (at level 3, format "e '^-1'") : equiv_scope.
```
That works.


> Hmm, maybe those mean this is a bad idea.

To be clear, I'm in favor of
```coq
Notation "1" := (reflexivity _).
```

JasonGross:
Doing
```coq
Notation "e ^-1" := (@equiv_inverse _ _ (@BuildEquiv _ _ e _)) (at level 3, format "e '^-1'") : equiv_scope.
```
breaks all of the rewrites of `eissect` and `eisretr` because they can't find `equiv_inv` in the goal, anymore.  This can be fixed by calling `simpl` before doing the rewrite, but this will make the `^-1` notation not show up unless we keep the original one around for display purposes, which seems potentially confusing (possibly more so than needing to learn the rule "use `refine ()%equiv` when putting `^-1` on the outside").

I'd prefer to have this merged first, because I think it's better than `equiv_compose'` everywhere, and then decide what to do about the remaining notation issues separately.

mikeshulman:
Yes, that's why I said

> We would probably need to have some wrappers like `eisretr_internal`/`eisretr`, since the fields in `IsEquiv` have to refer to `equiv_inv` directly whereas we'd like the type of `eisretr` to involve `^-1`.

JasonGross:
I do not think I'll have time to play with wrappers in the near future.  Do you want to look into it?

mikeshulman:
I might give it a brief try.  I'd also like to hear from @peterlefanulumsdaine and/or @andrejbauer before we make a change like this.

mikeshulman:
Okay, I think I'm against `Notation "1" := (reflexivity _)` not just for the reasons already stated, but because if someone writes `idpath` somewhere (which already happens), then it won't ever be displayed as `1`.

mikeshulman:
Ok, I tried wrappers, but they break if you *do* call `simpl` (which sometimes you need to do for other reasons), since then `equiv_inverse` is no longer present to be matched against the type of the wrapper.

theories/TruncType.v
```diff
@@ -42,7 +42,7 @@ Definition equiv_path_trunctype {n : trunc_index} (A B : TruncType n)
 Proof.
   equiv_via (A = B :> Type).
   - apply equiv_path_universe.
-  - exact (equiv_inverse (BuildEquiv _ _ (@ap _ _ (@trunctype_type n) A B) _)).
+  - exact ((BuildEquiv _ _ (@ap _ _ (@trunctype_type n) A B) _)^-1)%equiv.
```
mikeshulman: Curiously, this `%equiv` doesn't seem to be necessary.  Any ideas why?

mikeshulman:
I guess I'm reluctantly okay with this as-is, especially if bug 3990 has any chance of being fixed.  (Reluctant because of the remaning scope markers, of course -- overall it is a great improvement.)  I think I am still against `->` as a notation for `equiv_functor_arrow'`, though, and ambivalent about `+` and `*`.  And I would still like to hear from @peterlefanulumsdaine and/or @andrejbauer.

andrejbauer:
I am not against scope markers. The Ocaml people adopted them and they seem to work nicely. In Ocaml the scope markers are actually module names, so there is one less thing to remember (namely the notion of a scope name) because people can actually guess what `Float.(sin x + cos y)` means.

JasonGross:
@mikeshulman: I've removed the `->` notation for `equiv_functor_arrow'`.

theories/Spaces/Finite.v
```diff
@@ -52,14 +50,11 @@ Defined.
 
 (** *** Swap the last two elements. *)
 
-Definition fin_transpose_last_two (n : nat)
+Timeout 1 Definition fin_transpose_last_two (n : nat)
```
mikeshulman: What is the `Timeout` for?

mikeshulman:
What about notation like `+E` and `*E` for the `equiv_functor` operations?  I just feel, looking at the code, that all the `+ 1`s look too much like arithmetic.

theories/Spaces/Finite.v
JasonGross: Oops.  I forgot to remove it from when I was diagnosing [bug 3997](https://coq.inria.fr/bugs/show_bug.cgi?id=3997).

theories/TruncType.v
JasonGross: Perhaps `exact` automatically puts things in bound scopes, even though `refine` doesn't?  @mattam82, do you know the answer to this?

JasonGross:
Now using `*E` and `+E`.

theories/Utf8.v
```diff
@@ -7,6 +7,7 @@ Notation Type₀ := Type0.
 Notation pr₁ := pr1.
 Notation pr₂ := pr2.
 Local Open Scope fibration_scope.
+Notation "f → g" := (f -> g)%equiv : equiv_scope.
```
mikeshulman: Should this also be commented out?

mikeshulman:
LGTM now!  Thanks for putting up with me.

theories/Utf8.v
JasonGross: Yes, thanks, fixed by latest commit.

JasonGross:
I've updated the UTF-8 notations to change `∘` to `∘ᴱ`, comment out `→`, add `*ᴱ`, `×ᴱ`, and `+ᴱ`, and also add `×` for `Datatypes.prod`.  @mikeshulman, are you fine with all of these notations?

I think I still need APOE?

spitters:
LGTM. Feel free to merge when travis is happy.

mikeshulman:
Yes, the new unicode notations look excellent.

theories/TruncType.v
JasonGross: The answer, as per [enhancement 4034](https://coq.inria.fr/bugs/show_bug.cgi?id=4034): `exact` automatically makes use of bound scopes, while `refine` does not.

mikeshulman: Fascinating.  Well, then we can remove the `%equiv` here.  Should the style file mention this?

JasonGross: I've removed them.  The style file currently does not mention scopes at all, except in the context of `Local Open Scope` commands and that they should go at the top.  If you want to write a section on scopes and notations, and include this tidbit, feel free.  I'd prefer not to have it tied to this PR, though.

mikeshulman: Yes, fine, go ahead and merge this.

PR #711, opened 2015-02-04
Retry `cabal update` 10 times

JasonGross:
`cabal update` sometimes errors with
  Downloading the latest package list from hackage.haskell.org
  cabal: <socket: 3>: resource vanished
Now we retry multiple times.

### DIFF
--- a/etc/ci/install_coq_dot_deps.sh
+++ b/etc/ci/install_coq_dot_deps.sh
@@ -5,5 +5,12 @@ set -x
 
 sudo apt-get update -qq
 sudo apt-get install -q ghc cabal-install libgraphviz4 graphviz
-cabal update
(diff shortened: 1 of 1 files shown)


spitters:
Thanks! LGTM. However,
` # substitute your command here`
Should not be there.

JasonGross:
Fixed

andrejbauer:
Graphviz is implemented in Haskell?!

andrejbauer:
Nope, those seem to be Haskell bindings for graphviz, yes? Why do we need those?

JasonGross:
Yes.  They're used in [this file I took from mathclasses](https://github.com/HoTT/HoTT/blob/master/etc/DepsToDot.hs).

PR #712, opened 2015-02-04
Adding TOC to STYLE

spitters:


JasonGross:
LGTM

mikeshulman:
Could Travis auto-update this for us when the style file is changed?

JasonGross:
Probably, but it's bad practice to mix autogenerated files with manually generated files.  Would it be sufficient to have travis error and print out a replacement TOC for PRs that update the file?

mikeshulman:
What is bad about it in this instance?  We could add a comment to the markdown saying that the toc is autogenerated so that people don't try to edit it by hand.

mikeshulman:
(Ideally, the autogenerated toc could be in another file that is included, but does markdown have include capabilities?)

JasonGross:
You want it updated a la the erratum of the HoTT book?

spitters:
The first line already states that it is generated.
This solution was suggested on stackexchange, but I am open to better ones.

**Table of Contents** *generated with [DocToc](
http://doctoc.herokuapp.com/)*

mikeshulman:
Yes, like the errata.  If we do that, I would make the comment say more specifically that the toc will be auto-updated on every commit, rather than merely that it *was* autogenerated at one point.

PR #719, opened 2015-02-12
Add 1-Groupoid structure of equivalences

JasonGross:
Should we consider replacing [IsEquiv] with a definition that has more
judgmental properties, and doesn't require [Funext] for proving all of
the groupoid structure?  (Is there such a definition?)

### DIFF
--- a/Makefile_targets.mk
+++ b/Makefile_targets.mk
@@ -46,6 +46,7 @@ CORE_VFILES = \
 	$(srcdir)/theories/Types/Universe.v \
 	$(srcdir)/theories/Types/Nat.v \
 	$(srcdir)/theories/Types.v \
+	$(srcdir)/theories/EquivGroupoids.v \
 	$(srcdir)/theories/Fibrations.v \
 	$(srcdir)/theories/Conjugation.v \
 	$(srcdir)/theories/HProp.v \
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -171,7 +171,7 @@ Definition inv_V {A : Type} {x y : A} (p : x = y) :
   match p with idpath => 1 end.
 
 
-(* *** Theorems for moving things around in equations. *)
+(** *** Theorems for moving things around in equations. *)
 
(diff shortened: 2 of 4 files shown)


spitters:
It would be nice to package this up in a tactic/ rewrite database.

Nit: Without univalence this is only a pregroupoid, isn't it.

mikeshulman:
Nice!  LGTM.

> Is there such a definition?

I am very doubtful.

JasonGross:
> I am very doubtful.

I asked a question on the mailing list a while back about making more properties judgmental, and @peterlefanulumsdaine replied with a way to get judgmental involutivity of symmetry.  That the inverse of the identity is the identity should also hold strictly for this definition.  Looking more at it, if we add a weak-eta rule for inductive types to Coq, then the left and right units for paths are strict, and the left and right units for this definition of equivalence also become strict.  It's not clear to me if/how we can strictify associativity for paths, but if we manage that, then it also becomes strict for this definition of equivalences.  I don't think there's any hope for the left and right inverse laws, but this would still be much more than we have now.  Should we consider switching?  (Separate issue?)

JasonGross:
> Nit: Without univalence this is only a pregroupoid, isn't it.

Yes.

JasonGross:
> Nit: Without univalence this is only a pregroupoid, isn't it.

Yes.  Should I rename the file?

JasonGross:
> It would be nice to package this up in a tactic/ rewrite database.

I will add a comment about pulling the relevant code from PathGroupoids.  I don't want to write the tactic without using it anywhere, because then it is likely to break.

JasonGross:
> It would be nice to package this up in a tactic/ rewrite database.

Comment added

> Nit: Without univalence this is only a pregroupoid, isn't it.

I've changed the title of the file, but will only change the filename if asked.

mikeshulman:
I don't think it's worth renaming the file.  In fact it's not even technically a pregroupoid but a pre-oo-groupoid...

mikeshulman:
Peter's definition goes up a universe level, so I don't think we should use it.

JasonGross:
APOE?

PR #720, opened 2015-02-12
Equivalence induction

### LINKED ISSUE #691: Equivalence induction

JasonGross:
I've defined `equiv_induction`.

Unfortunately, it currently requires `Funext` in essentially all cases.  It should be possible to make it more fine-grained, so that we don't need `Funext` to deal with functions in one direction or the other, and only need it for equalities.

This closes #691.

### DIFF
--- a/theories/Tactics/EquivalenceInduction.v
+++ b/theories/Tactics/EquivalenceInduction.v
@@ -5,25 +5,29 @@ Require Import HoTT.Basics HoTT.Types.
 
 Local Open Scope equiv_scope.
 
-Class RespectsEquivalenceL A (P : forall B, A <~> B -> Type)
-  := respects_equivalenceL : { e' : forall B (e : A <~> B), P A (equiv_idmap A) <~> P B e | Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
-Class RespectsEquivalenceR A (P : forall B, B <~> A -> Type)
-  := respects_equivalenceR : { e' : forall B (e : B <~> A), P A (equiv_idmap A) <~> P B e | Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
+Class RespectsEquivalenceL A (P : forall B, (A <~> B) -> Type)
+  := respects_equivalenceL : { e' : forall B (e : A <~> B), P A (equiv_idmap A) <~> P B e & Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
+Class RespectsEquivalenceR A (P : forall B, (B <~> A) -> Type)
+  := respects_equivalenceR : { e' : forall B (e : B <~> A), P A (equiv_idmap A) <~> P B e & Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
 
+Global Arguments RespectsEquivalenceL : clear implicits.
+Global Arguments RespectsEquivalenceR : clear implicits.
+
+(** When doing equivalence induction, typeclass inference will either fully solve the respectfulness side-conditions, or not make any progress.  We would like to progress as far as we can on the side-conditions, so that we leave the user with as little to prove as possible.  To do this, we create a "database", implemented using typeclasses, to look up the refinement lemma, keyed on the head of the term we want to respect equivalence. *)
 Class respects_equivalence_db {KT VT} (Key : KT) {lem : VT} := make_respects_equivalence_db : Unit.
 Definition get_lem' {KT VT} Key {lem} `{@respects_equivalence_db KT VT Key lem} : VT := lem.
-Notation get_lem key := $(let res := constr:(get_lem' key) in let res' := (eval unfold get_lem' in res) in exact res')$.
+Notation get_lem key := $(let res := constr:(get_lem' key) in let res' := (eval unfold get_lem' in res) in exact res')$ (only parsing).
 
 Section const.
   Context {A : Type} {T : Type}.
   Global Instance const_respects_equivalenceL
-  : @RespectsEquivalenceL A (fun _ _ => T).
+  : RespectsEquivalenceL A (fun _ _ => T).
   Proof.
     refine (fun _ _ => equiv_idmap T; fun _ => _).
     exact idpath.
   Defined.
   Global Instance const_respects_equivalenceR
-  : @RespectsEquivalenceR A (fun _ _ => T).
+  : RespectsEquivalenceR A (fun _ _ => T).
   Proof.
     refine (fun _ _ => equiv_idmap _; fun _ => _).
     exact idpath.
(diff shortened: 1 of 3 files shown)


spitters:
Nice!

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
+(** * Equivalence induction *)
+Require Import HoTT.Basics HoTT.Types.
+
+(** We define typeclasses and tactics for doing equivalence induction. *)
+
+Local Open Scope equiv_scope.
+
+Class RespectsEquivalenceL A (P : forall B, A <~> B -> Type)
+  := respects_equivalenceL : { e' : forall B (e : A <~> B), P A (equiv_idmap A) <~> P B e | Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
```
mikeshulman: I know we've redefined them to mean the same thing, but we generally use the notation `{ x:A & ... }` rather than `{ x:A | ... }` for sigma-types; it took me a minute or two to figure out what the `|` was doing here.  Also, why define the class to have one field that belongs to a sigma-type rather than two fields?

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
...
+Class RespectsEquivalenceR A (P : forall B, B <~> A -> Type)
+  := respects_equivalenceR : { e' : forall B (e : B <~> A), P A (equiv_idmap A) <~> P B e | Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
+
+Class respects_equivalence_db {KT VT} (Key : KT) {lem : VT} := make_respects_equivalence_db : Unit.
```
mikeshulman: How about some comments explaining what this is doing?

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
...
+               (fun e' => _ = equiv_functor_prod' e' _) (respects_equivalenceR.2 _)
+               (transport
+                  (fun e' => _ = equiv_functor_prod' _ e') (respects_equivalenceR.2 _)
+                  idpath)).
+  Defined.
+
+  Global Instance: @respects_equivalence_db _ _ (@prod) (@prod_respects_equivalenceL) := tt.
+End prod.
+
+Local Ltac t' :=
```
mikeshulman: Not a very descriptive name...

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
+(** * Equivalence induction *)
+Require Import HoTT.Basics HoTT.Types.
+
+(** We define typeclasses and tactics for doing equivalence induction. *)
+
+Local Open Scope equiv_scope.
+
+Class RespectsEquivalenceL A (P : forall B, A <~> B -> Type)
```
mikeshulman: I prefer parentheses in something like `A <~> B -> C`; leaving them out requires the reader to guess or know that `<~>` binds tighter than `->`, and I don't know any obvious or intuitive reason why that should be the case.  (There are a lot of these in this file.)

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
...
+  := respects_equivalenceR : { e' : forall B (e : B <~> A), P A (equiv_idmap A) <~> P B e | Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
+
+Class respects_equivalence_db {KT VT} (Key : KT) {lem : VT} := make_respects_equivalence_db : Unit.
+Definition get_lem' {KT VT} Key {lem} `{@respects_equivalence_db KT VT Key lem} : VT := lem.
+Notation get_lem key := $(let res := constr:(get_lem' key) in let res' := (eval unfold get_lem' in res) in exact res')$.
+
+Section const.
+  Context {A : Type} {T : Type}.
+  Global Instance const_respects_equivalenceL
+  : @RespectsEquivalenceL A (fun _ _ => T).
```
mikeshulman: It seems that `RespectsEquivalenceL` is used almost everywhere with a `@` in front of it.  Given that, it seems to me that it would make more sense to adjust the implicitness of its arguments?

theories/Tactics/EquivalenceInduction.v
```diff
@@ -0,0 +1,352 @@
...
+End ap.
+
+Ltac step_respects_equiv :=
+  idtac;
+  match goal with
+    | _ => progress intros
+    | _ => assumption
+    | _ => progress unfold respects_equivalenceL
+    | _ => progress cbn
+    | _ => exact _
```
mikeshulman: I gather that what's going on here is that sometimes typeclass inference needs some help.  Can you explain more about what sort of help it needs and how you are giving it?

mikeshulman:
This is very nice!

theories/Tactics/EquivalenceInduction.v
JasonGross: Done.

theories/Tactics/EquivalenceInduction.v
JasonGross: Added comments; let me know if some things merit more explanation.

theories/Tactics/EquivalenceInduction.v
JasonGross: Done.

theories/Tactics/EquivalenceInduction.v
JasonGross: Renamed to `t_step`, and added a comment about what the tactic is meant to do (though not the ad-hoc way it manages to do it).

theories/Tactics/EquivalenceInduction.v
JasonGross: Added comments to this tactic.  Let me know if some bits merit more explanation.

JasonGross:
@mikeshulman, I think I've addressed all of your concerns.

mikeshulman:
Can you say why one field that has a sigma-type rather than two fields in a record?

theories/Tactics/EquivalenceInduction.v
JasonGross: Updated to use `&`.

JasonGross:
I think there are three reasons:
1. I was too lazy to think of good names for the fields.
2. In the dependent cases, where one equivalence-respectfulness proof will show up in the body of another goal, it might be the case that using sigma types allows us to reuse the respectfulness lemmas of sigma types, rather than writing new ones for this type.
3. I expect it to be more useful to see the type of the fields than the type of the record; sigma types have more informative notations than record type names.  (Thus the user can run `hnf` to see what is left to do in the side conditions.)

mikeshulman:
Okay, that makes sense; maybe it would be good to comment on that too.

mikeshulman:
I don't completely understand everything that this is doing yet, but it looks great and I look forward to experimenting with using it.  LGTM.

JasonGross:
Comment added.

JasonGross:
I hope it is useful :-)  I am still hoping to find a way to drop the `Funext` dependency in most cases, though I won't have much time to devote to it in the near future.

@spitters, shall I consider your "Nice!" an ack?

PR #721, opened 2015-02-12
Add relational definition of equivalence

JasonGross:
I wanted to see how many judmental laws @peterlefanulumsdaine's
definition of equivalence supported.   It turns out it doesn't support
as many as I had hoped.  Are there other definitions that support other
laws judgmentally?


mikeshulman:
Nice, thanks!  LGTM.

I don't know about other definitions.

PR #723, opened 2015-02-22
Better `make clean` target

JasonGross:
With a bit of help from stackoverflow, it's possible to not run coqdep when we run `make clean`.


spitters:
I trust you, but this seems quite complicated.

mikeshulman:
It more or less makes sense to me, and seems reasonable.

JasonGross:
Travis is happy, so I will merge.

andrejbauer:
Link to stackoverflow for posterity?

JasonGross:
Links are in a comment in `configure.ac`:
```
# create a snippet to get a Makefile conditional through automake,
# from http://stackoverflow.com/a/8643550/377022 and
# http://stackoverflow.com/a/28652045/377022
```

PR #724, opened 2015-02-22
Added contravariant versions of paths in Sigma types.

kristinas:


theories/Types/Sigma.v
```diff
@@ -68,6 +68,23 @@ Definition path_sigma {A : Type} (P : A -> Type) (u v : sigT P)
            (p : u.1 = v.1) (q : p # u.2 = v.2)
 : u = v
   := path_sigma_uncurried P u v (p;q).
+  
+(* Added a contravariant instance of path_sigma_uncurried *)
+Definition path_sigma_uncurried' {A : Type} (P : A -> Type) (u v : sigT P)
+  (pq : {p : u.1 = v.1 & u.2 = p^ # v.2})
+  : u = v
+  := match pq with
```
JasonGross: Rather than matching on `pq` and passing in `p` and `q`, it would be conciser to use `pq.1` and `pq.2` everywhere you need them.

kristinas: I just copied and pasted the definition for path_sigma_uncurried, I see that it was changed in the newest version of HoTT (my repo was out of date).

theories/Types/Sigma.v
```diff
@@ -68,6 +68,23 @@ Definition path_sigma {A : Type} (P : A -> Type) (u v : sigT P)
...
+       | existT p q =>
+         match u, v return (forall p0, (u.2 = p0^ # v.2) -> (u=v)) with
```
JasonGross: I don't think you actually need to match on `u` nor `v`, either, given that we have primitive projections on.  You should just be able to match on `pq.1` and `pq.2`.  Alternatively, why not define this in terms of `path_sigma_uncurried`, as, I think, `(path_sigma_uncurried v u (pq.1)^ (pq.2)^)^`?

theories/Types/Sigma.v
```diff
@@ -68,6 +68,23 @@ Definition path_sigma {A : Type} (P : A -> Type) (u v : sigT P)
            (p : u.1 = v.1) (q : p # u.2 = v.2)
 : u = v
   := path_sigma_uncurried P u v (p;q).
+  
+(* Added a contravariant instance of path_sigma_uncurried *)
```
JasonGross: I'd prefer
```coq
(** A contravariant instance of [path_sigma_uncurried] *)
```
The verb "Added" sounds out of place.

theories/Types/Sigma.v
```diff
@@ -190,6 +207,22 @@ Defined.
 Definition equiv_path_sigma `(P : A -> Type) (u v : sigT P)
 : {p : u.1 = v.1 &  p # u.2 = v.2} <~> (u = v)
   := BuildEquiv _ _ (path_sigma_uncurried P u v) _.
+  
+(* Added a contravariant version of isequiv_path_sigma' *)
```
JasonGross: Similarly,
```coq
(** A contravariant version of [isequiv_path_sigma'] *)
```

theories/Types/Sigma.v
JasonGross: Er, I mean, `(path_sigma_uncurried v u ((pq.1)^; (pq.2)^))^`.

theories/Types/Sigma.v
```diff
@@ -190,6 +207,22 @@ Defined.
...
+Instance isequiv_path_sigma' {A : Type} `{P : A -> Type} {u v : sigT P}
+  : IsEquiv (path_sigma_uncurried' P u v) | 0.
+  apply (isequiv_adjointify (path_sigma_uncurried' P u v) (fun r => match r with idpath => (1; 1) end)).
```
JasonGross: If we define `path_sigma_uncurried'` in terms of `path_sigma_uncurried`, will Coq pick this up automatically?  If not, do we have the right `IsEquiv` instances to build this up by composing functions?

theories/Types/Sigma.v
kristinas: So am I supposed to make these changes or is it up to the developers?

mikeshulman: By virtue of submitting a pull request, you are one of the developers.  (-:

theories/Types/Sigma.v
andrejbauer: Can we come up with a better name? The apostrophe is not very descriptive.

theories/Types/Sigma.v
mikeshulman: More to the point: the standard procedure seems to be that the submitter of a pull request either makes the changes suggested by others, or explains why they would rather not.  To make changes, just add more commits to the same branch (`master` in this case) and push them to your repo fork.

theories/Types/Sigma.v
mikeshulman: Yes, and the prime is especially unfortunate because I think in some other cases it is used to denote the same thing that here is denoted by `_uncurried`.  (As an aside, I would love to have a shorter but unambiguous way to denote `_uncurried`.)  But I can't think of a better name for this right now either; a suffix like `_V` would suggest to me that it's talking about the inverses of paths in a sigma-type, rather than paths obtained by transporting backwards.

kristinas: How about C for contravariant?
Anyways, I didn't know Coq had eta for pairs. Seems like a new thing to me.

mikeshulman:
Unfortunately, "C" could also stand for "covariant"...

andrejbauer:
How about `_contra`? And what if there are two places where contravariance could be applied?

kristinas:
So should I change it to _contra and be done with it?

mikeshulman:
Sure, I think that's good enough for now.  If anyone thinks of a better name later we can change it.

PR #726, opened 2015-02-24
put surreal cut notation in a module to avoid clobbering nested sigma-types

mikeshulman:


JasonGross:
Have you checked that this fixes it?  I don't think it does, because `Reserved Notation` doesn't respect module boundaries.

mikeshulman:
Kristina's example

```coq
Require Import HoTT.
Check ({ l : Unit & { n : Unit & Unit }}).
```

now works for me.  I added it to `Tests.v` so Travis will check it too.

JasonGross:
While you're at it, do you want to `Bind Scope surreal_scope with No` right after you define `No`?

JasonGross:
I'm confused.  Either it should be enough to put it in `surreal_scope` (and you shouldn't need to put it in a module), or putting it in a module shouldn't be enough to fix things...

JasonGross:
LGTM, though I'd like to know why this works.

mikeshulman:
Let me know if you figure it out.

JasonGross:
Also, are you planning on a `Bind Scope`?

mikeshulman:
Sure, I'll do it later tonight.

mikeshulman:
scope bound.

mikeshulman:
bump

JasonGross:
We are just waiting for APOE?  (And I may or may not eventually investigate why this fixes things?)

mikeshulman:
I think so.

PR #728, opened 2015-02-24
Fix hoqide on Windows

JasonGross:
Do this by checking for the existence of `cygpath`, and Windowsifying
the paths if it exists.  This breaks machines with a `cygpath` in
`PATH` which don't support Windows paths, but such machines are rare,
I believe.

### DIFF
--- a/hoq-config.in
+++ b/hoq-config.in
@@ -38,6 +38,7 @@ function readlink_f() {
 # script, then use that one, otherwise use the global one. This trick
 # allows hoqc to work "in place" on the source files.
 mydir="$(dirname "$(readlink_f "${BASH_SOURCE[0]}")")"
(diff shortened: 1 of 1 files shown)


mikeshulman:
Sure.

spitters:
Do we really need to support obscure OS-es? :-)

I trust you to do the right things regarding windows.

JasonGross:
This is the change I needed to test Kristina's anomaly.  :-)

I will merge when Travis is happy.

andrejbauer:
I believe Windows might still be the most numerous OS out there. If we can have a contributor who would take care of the Windows side of things, that would be great.

JasonGross:
I use windows and am happy to fix problems as they arise, but don't use
hoqide much.

PR #733, opened 2015-03-05
Prove saturatedness - Grothendieck of F : C -> Set

JasonGross:
An answer to Mike's suspicion on the mailing list.

It seems to be much, much, much easier to construct an equivalence `(s = d) ≃ (s ≅ d)` and prove it's carrier function equal to `idtoiso`, than it is to directly construct a function `(s ≅ d) → (s = d)` and prove that it's a section and retraction of `idtoiso`.  Any ideas if this is an instance of a more general property?  (Or ideas for why this is the case?)

If anyone want's to try their hand at pseduofunctor "path" algebra, there are [a few `admit`s left in this branch that does this construction for the Grothendieck construction for a pseudofunctor to cat](https://github.com/HoTT/HoTT/compare/master...JasonGross:grothendieck-cat-univalent#diff-91e920f8e0ee250520376601b86f019eR39).

### DIFF
--- a/Makefile_targets.mk
+++ b/Makefile_targets.mk
@@ -330,6 +330,9 @@ CATEGORY_VFILES = \
 	\
 	$(srcdir)/theories/categories/Grothendieck.v \
 	$(srcdir)/theories/categories/Grothendieck/ToSet.v \
+	$(srcdir)/theories/categories/Grothendieck/ToSet/Core.v \
+	$(srcdir)/theories/categories/Grothendieck/ToSet/Morphisms.v \
+	$(srcdir)/theories/categories/Grothendieck/ToSet/Univalent.v \
 	$(srcdir)/theories/categories/Grothendieck/PseudofunctorToCat.v \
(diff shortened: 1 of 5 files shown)


mikeshulman:
Nice!!  LGTM.

We do have lots of other proofs by e.g. chaining equivalences that are much easier that way than directly constructing two functions and their inverses.  It's also actually pretty common in informal category theory to "prove" that some map is an equivalence by constructing a chain of known equivalences from its domain to its codomain and saying "we leave it to the reader to check that this map is in fact the desired one".  Although I do this myself as well, I've never been entirely sure whether or not these checks-left-to-the-reader should be considered easy or just tedious.  I guess here is some evidence that at least sometimes they are easy.

The only other example of this specifically that I can think of in our library so far was my original proof that reflective subuniverses preserve products up to equivalence, which you then massaged around into the other sort of proof.

andrejbauer:
Many things in category theory are easy *and* tedious.

PR #737, opened 2015-03-20
Adding information to the book.

spitters:


contrib/HoTTBook.v
```diff
@@ -1835,12 +1843,14 @@ Definition Book_9_8_2 := @HoTT.categories.Structure.IdentityPrinciple.structure_
 
 (* ================================================== lem:sets_exact *)
 (** Lemma 10.1.8 *)
-
+Definition Book_10_1_8 :=@hit.quotient.sets_exact.
```
JasonGross: Nit: add a space after `:=`

contrib/HoTTBook.v
```diff
@@ -1810,12 +1810,20 @@ Definition Book_9_8_2 := @HoTT.categories.Structure.IdentityPrinciple.structure_
 
 (* ================================================== thm:mono *)
 (** Lemma 10.1.1 *)
-
+(* No complete equivalence yet, but see:
```
JasonGross: Do we want these to be coqdoc comments?

spitters: I put them in for future reference. It would be nice to actually prove the
theorem, but I don't have the time now.

contrib/HoTTBook.v
```diff
@@ -1810,12 +1810,20 @@ Definition Book_9_8_2 := @HoTT.categories.Structure.IdentityPrinciple.structure_
...
+HSet.isembedding_isinj_hset
+HSet.isinj_ismono
+HSet.ismono_isinj
+HSet.ismono_isequiv *)
 
 
 (* ================================================== epis-surj *)
 (** Lemma 10.1.4 *)
-
+(* The proof are slightly different, but see:
```
mikeshulman: I think if the _statement_ is the same then we should link it as normal.

andrejbauer: I'd even link a partial solution as long as the comments make it clear that's what it is. We can give it a name such as `Book_10_1_4_partial`.

spitters: I've move the partial lemmas to Definitions as you suggested.

andrejbauer:
LGTM

PR #738, opened 2015-03-20
Move homotopy coequalizers into their own file

mikeshulman:
And define pushouts in terms of them.  That way we could in theory apply the existing flattening lemma to pushouts directly.


spitters:
LGTM

mikeshulman:
APOE?

PR #740, opened 2015-03-26
W types

kristinas:
Added a file on W types

### DIFF
--- a/Makefile_targets.mk
+++ b/Makefile_targets.mk
@@ -42,6 +42,7 @@ CORE_VFILES = \
 	$(srcdir)/theories/Types/Empty.v \
 	$(srcdir)/theories/Types/Bool.v \
 	$(srcdir)/theories/Types/Sum.v \
+  $(srcdir)/theories/Types/Wtype.v \
 	$(srcdir)/theories/Types/Equiv.v \
 	$(srcdir)/theories/Types/Universe.v \
 	$(srcdir)/theories/Types/Nat.v \
--- a/theories/Types/Wtype.v
+++ b/theories/Types/Wtype.v
@@ -79,7 +79,7 @@ Global Instance trunc_wtype `{Funext} `{B : A -> Type}
 : IsTrunc n.+1 (W A B) | 100.
 Proof.
   generalize dependent A; intros A B ac; intros z; induction z; intro y.
-  destruct y; refine (trunc_equiv _ (equiv_path_wtype _ _)).
+  destruct y as [a0 w0]; refine (trunc_equiv _ (equiv_path_wtype _ _)).
   apply (trunc_equiv' {p : a = a0 & forall b, w b = w0 (p # b)}).
   { transitivity {p : a = a0 & transport (fun a => B a -> W A B) p w = w0}.
     { apply equiv_functor_sigma_id; intro p; induction p.
--- a/theories/Types/Wtype.v
+++ b/theories/Types/Wtype.v
@@ -12,12 +12,14 @@ Inductive W (A : Type) (B : A -> Type) : Type :=
 
 (** ** Paths *)
 
+
 Definition path_wtype {A B} (z z' : W A B)
            (pq : match z, z' with
                    | sup a1 t1, sup a2 t2 =>
-                       (existT (fun a => B a -> W A B) a1 t1) = (a2;t2)
+                       (a1;t1) = (a2;t2) :> {a : _ & B a -> W A B}
                  end)
 : z = z'.
+Proof.
   destruct z, z'; apply (path_sigma_uncurried _ _ _)^-1 in pq.
   destruct pq as [p q]; cbn in p; destruct p; cbn in q; destruct q.
   exact idpath.
--- a/theories/Types/Wtype.v
+++ b/theories/Types/Wtype.v
@@ -8,30 +8,26 @@ Local Open Scope path_scope.
 Generalizable Variables X A B C f g n.
 
 Inductive W (A : Type) (B : A -> Type) : Type :=
-  sup : forall a, (B a -> W A B) -> W A B.
+  sup { w_label : A ; w_arg : B w_label -> W A B }.
+
+Arguments w_label {A B} _.
+Arguments w_arg {A B} _ _.
 
 (** ** Paths *)
 
-Definition path_wtype {A B} (z z' : W A B)
-           (pq : match z, z' with
-                   | sup a1 t1, sup a2 t2 =>
-                       (a1;t1) = (a2;t2) :> {a : _ & B a -> W A B}
-                 end)
-: z = z'.
+Definition path_wtype {A B} (u v : W A B)
+  (pq : (w_label u;w_arg u) = (w_label v;w_arg v) :> {a : _ & B a -> W A B})
+: u = v.
 Proof.
-  destruct z, z'; apply (path_sigma_uncurried _ _ _)^-1 in pq.
+  destruct u,v; apply (path_sigma_uncurried _ _ _)^-1 in pq.
   destruct pq as [p q]; cbn in p; destruct p; cbn in q; destruct q.
   exact idpath.
 Defined.
 
-Definition path_wtype_inv {A B} {z z' : W A B}
-           (pq : z = z')
-: match z, z' with
-    | sup a1 t1, sup a2 t2 =>
-         (existT (fun a => B a -> W A B) a1 t1) = (a2;t2)
-  end
+Definition path_wtype_inv {A B} {u v : W A B} (pq : u = v)
+: (w_label u;w_arg u) = (w_label v;w_arg v) :> {a : _ & B a -> W A B}
   := match pq with
-       | 1 => match z with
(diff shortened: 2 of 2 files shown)


theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+                       (@eisretr_path_wtype A B z z')
+                       (@eissect_path_wtype A B z z')).
+Defined.
+
+Definition equiv_path_wtype {A B} (z z' : W A B)
+  := BuildEquiv _ _ _ (@isequiv_path_wtype A B z z').
+
+(** ** W-types preserve truncation *)
+
+Global Instance trunc_sigma `{Funext} `{B : A -> Type}
```
JasonGross: I assume `trunc_sigma` is a name you forgot to update?

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+         `{IsTrunc n.+1 A}
+: IsTrunc n.+1 (W A B) | 100.
+Proof.
+  generalize dependent A; intros A B ac; intros z; induction z; intro y.
+  destruct y; refine (trunc_equiv _ (equiv_path_wtype _ _)).
+  apply (trunc_equiv' {p : a = a0 & forall b, w b = w0 (p # b)}).
+    transitivity {p : a = a0 & transport (fun a => B a -> W A B) p w = w0}.
```
JasonGross: Use bullets and braces to mark off different goals, rather than just indentation.

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+  sup : forall a, (B a -> W A B) -> W A B.
+
+(** ** Paths *)
+
+Definition path_wtype {A B} (z z' : W A B)
+           (pq : match z, z' with
+                   | sup a1 t1, sup a2 t2 =>
+                       (existT (fun a => B a -> W A B) a1 t1) = (a2;t2)
+                 end)
+: z = z'.
```
spitters: I would have indented this line and added `Proof.`

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+  destruct z, z'; apply (path_sigma_uncurried _ _ _)^-1 in pq.
+  destruct pq as [p q]; cbn in p; destruct p; cbn in q; destruct q.
+  exact idpath.
+Defined.
+
+Definition path_wtype_inv {A B} {z z' : W A B}
+           (pq : z = z')
+: match z, z' with
+    | sup a1 t1, sup a2 t2 =>
+         (existT (fun a => B a -> W A B) a1 t1) = (a2;t2)
```
JasonGross: Note: You can write `(a1; t1) = (a2; t2) :> { a : A | B a -> W A B }` here, if you think it looks nicer.

JasonGross:
You need to add `Wtype.v` to `Makefile_targets.mk` (this is what Travis is complaining about).

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+                       (@path_wtype_inv _ _ z z')
+                       (@eisretr_path_wtype A B z z')
+                       (@eissect_path_wtype A B z z')).
+Defined.
+
+Definition equiv_path_wtype {A B} (z z' : W A B)
+  := BuildEquiv _ _ _ (@isequiv_path_wtype A B z z').
+
+(** ** W-types preserve truncation *)
+
```
spitters: I believe this is a numbered thm in the book. It would be nice to add it to `Book.v`.

spitters:
LGTM.

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+(** * Theorems about W-types (well-founded trees) *)
+
+Require Import HoTT.Basics.
+Require Import Types.Forall Types.Sigma Types.Paths Types.Unit.
+Local Open Scope path_scope.
+
+Generalizable Variables X A B C f g n.
+
+Inductive W (A : Type) (B : A -> Type) : Type :=
+  sup : forall a, (B a -> W A B) -> W A B.
```
JasonGross: Do you want to give names to the projections `W A B -> A` and `forall w : W A B, B (<first projection w) -> W A B`?  If so, you can write
```coq
Inductive W (A : Type) (B : A -> Type) : Type :=
  sup { a_name : a ; b_name : B a_name -> W A B }.
```
presumably with better names for `a_name` and `b_name`.  Then your code below won't need so many `match`es.

spitters:
It would be nice to add some examples. Nat and the Brouwer ordinals for instance.

JasonGross:
Where do the examples belong?  In `Spaces/Wtype/Examples.v`?

spitters:
I'd be inclined to put them in the same file. It should be only two lines.

JasonGross:
But we may eventually want, e.g., a tactic that builds the W-type corresponding to an arbitrary inductive type, and another one that automatically proves the equivalence, and more examples.  It might be nice to plan for adding these things, by giving the examples their own file (where tactics and helper lemmas and such could be added).  @mikeshulman, @andrejbauer, what do you think?

Regardless, I don't think this should block merging; if this file is added to the list in `Makefile_targets.mk`, I will be willing to merge this, whether or not my other suggestions are implemented; things can always be improved later.

spitters:
I'd propose to put them here and move them when we want to include more
examples.

theories/Types/Wtype.v
kristinas: That gives me a parsing error at the very first lemma. I am not sure why but I do not want to spend a whole lot of time on this.

theories/Types/Wtype.v
```diff
@@ -0,0 +1,88 @@
...
+
+(** ** W-types preserve truncation *)
+
+Global Instance trunc_wtype `{Funext} `{B : A -> Type}
+         `{IsTrunc n.+1 A}
+: IsTrunc n.+1 (W A B) | 100.
+Proof.
+  generalize dependent A; intros A B ac; intros z; induction z; intro y.
+  destruct y; refine (trunc_equiv _ (equiv_path_wtype _ _)).
+  apply (trunc_equiv' {p : a = a0 & forall b, w b = w0 (p # b)}).
```
JasonGross: You did not introduce the name `w` nor `w0` anywhere.  Use `destruct y as ...` so you don't need to use variables you didn't name.

theories/Types/Wtype.v
JasonGross: Sorry, the problem is that I used `a_name : a` instead of `a_name : A`.  If you make that fix, the file compiles up to the comment I just added below, where you use `w` without introducing it.

kristinas: That is not the problem, of course, I used a_name : A. Parsing error at the first definition. 
If you want to change something, feel free, I will now commit the final version.

JasonGross: What is the syntax error?  I get no such error when I make the change; see https://github.com/kristinas/HoTT/compare/master...JasonGross:pr/740.

kristinas: I get the error after I try to get rid of the match in the first definition.

JasonGross: I've updated [the link](https://github.com/kristinas/HoTT/compare/master...JasonGross:pr/740) to use the projections, too.  Is that what you're looking for?

kristinas: I see; the problem was that I tried to also get rid of the argument name pq in the definition and then it gives a parsing error.

mikeshulman:
I don't have strong feelings about where the examples go.  If we don't have much to say about them yet, I think it would be fine to put them here.

andrejbauer:
It doesn't matter where examples go right now. If and when we want more, we can move the examples. It's not like we worry about backward compatibility.

theories/Types/Wtype.v
mikeshulman: It's Exercise 7.3, although as stated that includes the additional unnecessary assumption that B is a family of n-types.  We should probably remove it.

Makefile_targets.mk
```diff
@@ -42,6 +42,7 @@ CORE_VFILES = \
 	$(srcdir)/theories/Types/Empty.v \
 	$(srcdir)/theories/Types/Bool.v \
 	$(srcdir)/theories/Types/Sum.v \
+  $(srcdir)/theories/Types/Wtype.v \
```
mikeshulman: I hate to sound overly picky, but all the lines here ought to have the same indentation.  (-:O

mikeshulman:
Also, this file should be exported by `theories/Types.v`.

Makefile_targets.mk
JasonGross: I've done this with eeb45760436449efb7b4709bbb1405142e214f00.  The "mistake" was that the rest of the lines are indented with tabs, while this line was indented with two spaces; some editors display tabs as two spaces, some as 4 spaces, and GitHub (or chrome) apparently displays them as 7 spaces (or 8, maybe?).

PR #741, opened 2015-03-29
More about coequalizers and the circle

mikeshulman:
* Functoriality and double recursion/induction for coequalizers
* Put `Int` in its own file
* Define the circle as a coequalizer and prove that it is 1-truncated and 0-connected
* The action of `S1` on a type family is iteration of an equivalence

### DIFF
--- a/theories/hit/Coeq.v
+++ b/theories/hit/Coeq.v
@@ -44,3 +44,131 @@ Proof.
   refine ((apD_const (@Coeq_ind B A f g (fun _ => P) coeq' _) (cp b))^ @ _).
   refine (Coeq_ind_beta_cp (fun _ => P) _ _ _).
 Defined.
+
+(** ** A double recursion principle *)
+
+Section CoeqRec2.
+  Context `{Funext}
+          {B A : Type } {f g : B -> A} {B' A' : Type} {f' g' : B' -> A'}
+          (P : Type) (coeq' : A -> A' -> P)
+          (cpl : forall b a', coeq' (f b) a' = coeq' (g b) a')
+          (cpr : forall a b', coeq' a (f' b') = coeq' a (g' b'))
+          (cplr : forall b b', cpl b (f' b') @ cpr (g b) b'
+                               = cpr (f b) b' @ cpl b (g' b')).
+
+  Definition Coeq_rec2
+  : Coeq f g -> Coeq f' g' -> P.
+  Proof.
+    refine (Coeq_rec _ _ _).
+    - intros a.
+      refine (Coeq_rec _ _ _).
+      + intros a'.
+        exact (coeq' a a').
+      + intros b'; cbn.
+        apply cpr.
+    - intros b.
+      apply path_arrow; intros a.
+      revert a; refine (Coeq_ind _ _ _).
+      + intros a'. cbn.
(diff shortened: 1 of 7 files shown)


JasonGross:
A question (possibly independent of this PR): do we want to reinstate the `Z` of the Coq standard library in place of our `Int` (and get nice notation)?  Or do we want to keep the replacement stdlib small, and use our own definition of `Int`?

mikeshulman:
Is that the binary integers?

spitters:
The stdlib integers are [here](https://coq.inria.fr/library/Coq.Numbers.Integer.Abstract.ZAxioms.html).
Since the module system is a bit baroque, I think we should be careful to drag all of this in.


LGTM

theories/hit/Coeq.v
```diff
@@ -44,3 +46,279 @@ Proof.
...
+    all:rewrite (ap_compose (_ o h) k^-1).
+    all:rewrite <- !(ap_pp k^-1), !concat_pp_p.
+    1:rewrite (concat_Ap (fun b => (p b)^) (eissect h b)^).
+    2:rewrite (concat_Ap (fun b => (q b)^) (eissect h b)^).
+    all:rewrite concat_p_Vp, concat_p_pp.
+    all:rewrite <- (ap_compose (k o _) k^-1), (ap_compose _ (k^-1 o k)).
+    all:rewrite (concat_A1p (eissect k) _).
+    all:rewrite ap_V, concat_pV_p; reflexivity.
+  Close Scope long_path_scope.
+Qed.
```
JasonGross: Do we really want `(functor_coeq h k p q)^-1` to be opaque?

theories/hit/Coeq.v
```diff
@@ -44,3 +46,279 @@ Proof.
...
+        refine (transport_forall_constant _ _ _ @ _).
+        apply cpl.
+      + intros b'; cbn.
+        refine (transport_paths_FlFr_D (cp b') _ @ _).
+        rewrite Coeq_ind_beta_cp.
+        (** Now begins the long haul. *)
+        Open Scope long_path_scope.
+        rewrite ap_pp.
+        repeat rewrite concat_p_pp.
+        (* Our first order of business is to get rid of the [Coeq_ind]s, which only occur in the following incarnation. *)
```
JasonGross: Any particular reason that the first and last of these comments will show up in coqdoc's output, but the rest won't?

theories/hit/Coeq.v
```diff
@@ -44,3 +46,279 @@ Proof.
...
+
+  (** TODO: [Coeq_rec2_beta_cplr] *)
+
+End CoeqRec2.
+
+(** ** A double induction principle *)
+
+Section CoeqInd2.
+  Context `{Funext}
+          {B A : Type } {f g : B -> A} {B' A' : Type} {f' g' : B' -> A'}
```
JasonGross: Nit: no space in `Type }`

JasonGross:
LGTM, feel free to merge with or without fixing my suggestions

mikeshulman:
Suggestions all fixed.  Thanks!

mikeshulman:
Travis is happy, merging.

PR #742, opened 2015-03-31
quotients by free actions of Int

mikeshulman:
The quotient of a set by a free action of the integers can be defined as an untruncated homotopy coequalizer, but turns out to be a set automatically.  It ought to be true more generally that the homotopy quotient of a set by a free action of any group is again a set, but defining "homotopy quotient" in generality is tricky; for the integers we can do it by coequalizing the action of 1 with the identity.

### DIFF
--- a/theories/hit/FreeIntQuotient.v
+++ b/theories/hit/FreeIntQuotient.v
@@ -1,0 +1,101 @@
+(* -*- mode: coq; mode: visual-line -*- *)
+Require Import HoTT.Basics HoTT.Types.
+Require Import TruncType HProp HSet.
+Require Import Spaces.Int.
+Require Import hit.Coeq hit.Circle hit.Flattening hit.Connectedness hit.Truncations.
+Import TrM.
+
+Local Open Scope path_scope.
+
+(** * Quotients by free actions of [Int] *)
+
+(** We will show that if [Int] acts freely on a set, then the set-quotient of that action can be defined without a 0-truncation, giving it a universal property for mapping into all types. *)
+
+Section FreeIntAction.
+
+  Context `{Univalence}.
+  Context (R : Type) `{IsHSet R}.
+
+  (** A free action by [Int] is the same as a single autoequivalence [f] (the action of [1]) whose iterates are all pointwise distinct. *)
+  Context (f : R <~> R)
+          (f_free : forall (r : R) (n m : Int),
+                      (iter_int f n r = iter_int f m r) -> (n = m)).
+
+  (** We can then define the quotient to be the coequalizer of [f] and the identity map.  This gives it the desired universal property for all types; it remains to show that this definition gives a set. *)
+  Let RmodZ := Coeq f idmap.
+
+  (** Together, [R] and [f] define a fibration over [S1].  By the flattening lemma, its total space is equivalent to the quotient. *)
+  Global Instance isset_RmodZ : IsHSet RmodZ.
+  Proof.
+    refine (trunc_equiv'
+              { z : S1 & S1_rec Type R (path_universe f) z}
+              (_ oE (@equiv_flattening _ Unit Unit idmap idmap
+                                  (fun _ => R) (fun _ => f))^-1
+              oE _));
+    try exact _.
+    - refine (equiv_adjointify _ _ _ _).
+      + refine (Wtil_rec _ _ _).
+        * intros u r; exact (coeq r).
+        * intros u r; cbn. exact ((cp r)^).
+      + refine (Coeq_rec _ _ _).
+        * exact (cct tt).
+        * intros r; exact ((ppt tt r)^).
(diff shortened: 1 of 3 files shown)


theories/hit/FreeIntQuotient.v
```diff
@@ -0,0 +1,101 @@
...
+  (** Together, [R] and [f] define a fibration over [S1].  By the flattening lemma, its total space is equivalent to the quotient. *)
+  Global Instance isset_RmodZ : IsHSet RmodZ.
+  Proof.
+    refine (trunc_equiv'
+              { z : S1 & S1_rec Type R (path_universe f) z}
+              (_ oE (@equiv_flattening _ Unit Unit idmap idmap
+                                  (fun _ => R) (fun _ => f))^-1
+              oE _));
+    try exact _.
+    - refine (equiv_adjointify _ _ _ _).
```
spitters: Wouldn't `rapply` look better here?

spitters:
Nice! LGTM

theories/hit/FreeIntQuotient.v
mikeshulman: `rapply equiv_adjointify` fails with `Cannot infer the implicit parameter A of equiv_adjointify`, while `rapply @equiv_adjointify` apparently tries to construct an equivalence with an unknown domain and codomain `Wtil Unit Unit idmap idmap (unit_name R) (unit_name f) <~> RmodZ`, coerce it to a function, and then apply it to an unknown argument.  But `rapply' @equiv_adjointify` works.  Is this anything we can fix about `rapply`?

JasonGross: The difference between `rapply` and `rapply'` is whether they prefer applications with more arguments or with fewer arguments.  I don't see anything we can do about that; `rapply` with `equiv_adjointify` is always going to try the coercion to a function before anything else, because it prefers applications with more arguments.  If you want to get rid of the `Cannot infer...` message without using `@`, use `erapply` or `erapply'` instead, which will fill in the implicits with evars.

theories/hit/FreeIntQuotient.v
```diff
@@ -0,0 +1,101 @@
...
+
+  Context `{Univalence}.
+  Context (R : Type) `{IsHSet R}.
+
+  (** A free action by [Int] is the same as a single autoequivalence [f] (the action of [1]) whose iterates are all pointwise distinct. *)
+  Context (f : R <~> R)
+          (f_free : forall (r : R) (n m : Int),
+                      (iter_int f n r = iter_int f m r) -> (n = m)).
+
+  (** We can then define the quotient to be the coequalizer of [f] and the identity map.  This gives it the desired universal property for all types; it remains to show that this definition gives a set. *)
```
JasonGross: Do we want to prove that this lines up with the universal property of the quotient-via-hits / have Coq verify the equivalence in some way?

JasonGross:
LGTM

theories/hit/FreeIntQuotient.v
mikeshulman: Why does the unprimed version prefer applications with more arguments, rather than vice versa?  My gut would have been that applications with fewer arguments would be more likely to be the one intended by a user.

theories/hit/FreeIntQuotient.v
mikeshulman: Yes, that would be nice to do, but I don't have time to do it right now.  I'll add it as a TODO.

theories/hit/FreeIntQuotient.v
JasonGross: Because the unprimed version is copied verbatim from Coq's standard library, and the primed version was a variant of it.  I'd be fine with flipping the meanings if we defined them both in our theories files, and perhaps removing it from the file it's in in the replacement stdlib, or just not exporting that file.  I'd be slightly less fine with flipping the meanings while keeping the definition in the replacement stdlib.

mikeshulman: Do you think that people coming from vanilla Coq might get confused if we
switched the meanings?

JasonGross: I suspect very few people intentionally rely on the ordering, and most people black-box it as "like `apply`, but with different unification behavior", and won't be bothered if we fiddle with the unification some more.

mikeshulman: Ok.  I'm still a little uneasy about making such a change without knowing
_why_ the stdlib did it in the order they did.

JasonGross: I think @mattam82 wrote the original tactic, and might be able to answer that?

spitters: I seem to recall it came from our corn library written in collaboration
with pierre letouzey.
I don't think the order of the arguments was an issue at the time.
It would be good to discuss on coqdev whether this should be changed for
Coq too.

PR #743, opened 2015-03-31
reflectors preserve coproducts and coequalizers

mikeshulman:
where "preserve" means "map them to colimits in the subuniverse, i.e reflected colimits".  The proof for coequalizers was really hard; I kind of hope I am missing something obvious that would make it easier.

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -424,7 +424,8 @@ Section Reflective_Subuniverse.
       rewrite !concat_pV_p.
       apply whiskerL. apply inverse2.
       apply ap_compose.
-    Defined.
+      Close Scope long_path_scope.
+    Qed.
 
     (** Preservation of equivalences *)
     Global Instance isequiv_O_functor {A B : Type} (f : A -> B) `{IsEquiv _ _ f}
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -2,6 +2,8 @@
 Require Import HoTT.Basics HoTT.Types.
 Require Import UnivalenceImpliesFunext EquivalenceVarieties Extensions Fibrations HProp.
 Require Import HoTT.Tactics.
+Require Import hit.Coeq.
+Require Import Tactics.RewriteModuloAssociativity.
 
 Local Open Scope path_scope.
 
@@ -294,16 +296,21 @@ Section Reflective_Subuniverse.
     Definition O_functor {A B : Type} (f : A -> B) : O A -> O B
       := O_rec (to O B o f).
 
+    (** Naturality of [to O] *)
+    Definition to_O_natural {A B : Type} (f : A -> B)
+    : (O_functor f) o (to O A) == (to O B) o f
+    := (O_rec_beta _).
+
     (** Functoriality on composition *)
     Definition O_functor_compose {A B C : Type} (f : A -> B) (g : B -> C)
     : (O_functor (g o f)) == (O_functor g) o (O_functor f).
     Proof.
       apply O_indpaths; intros x.
-      refine (O_rec_beta (to O C o g o f) x @ _).
+      refine (to_O_natural (g o f) x @ _).
       transitivity (O_functor g (to O B (f x))).
-      - symmetry. exact (O_rec_beta (to O C o g) (f x)).
+      - symmetry. exact (to_O_natural g (f x)).
       - apply ap; symmetry.
-        exact (O_rec_beta (to O B o f) x).
+        exact (to_O_natural f x).
     Defined.
 
     (** Functoriality on homotopies (2-functoriality) *)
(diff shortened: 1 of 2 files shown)


spitters:
LGTM

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -808,6 +843,160 @@ Section Reflective_Subuniverse.
...
+          rewrite !ap_pp, !concat_p_pp.
+          unfold O_functor_homotopy; rewrite O_indpaths_beta.
+          rewrite !ap_pp, !concat_p_pp.
+          pose (p := O_functor_compose_compose
+                       g coeq coeq_to (to O B b)).
+          apply moveL_pV in p.
+          rewrite concat_pp_p in p.
+          apply moveR_Vp in p.
+          rewrite@A <- p; clear p.
+          (** for the next one, [rewrite@A] spins forever *)
```
JasonGross: Sounds like a bug in `rewrite@A`.  Any chance you can figure out what part it spins forever in?  (Perhaps with `Info 1 rewrite@A ...`?)

mikeshulman: `Info 1 rewrite@A` also spins forever, and doesn't produce any output when I break it.  Manually copy-pasting and stepping through suggests the problem is in `left_associate_concat_in`, which is a bit too much magic for me to understand.

mikeshulman: More specifically, here's some code that spins forever (where "forever" means "until I get bored and break it"):

```
  let lem' := preconcat_any @to_O_natural_compose in
  pose proof lem' as H.
  left_associate_concat_in H.
```

JasonGross: I think you can get info to print something by doing, e.g., `Info 1 try (timeout 2 <sometactic>)`.  The `timeout` will make it not spin forever, and the `try` will cause the whole command to succeed, so `Info` should print things.  I'll try to look into it when I get a chance, but I'm not sure when that will be.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -391,6 +404,28 @@ Section Reflective_Subuniverse.
...
+      Open Scope long_path_scope.
+      rewrite !O_indpaths_beta, !ap_pp, !ap_V, !concat_p_pp.
+      refine (whiskerL _ (apD _ (to_O_natural f a)^)^ @ _).
+      rewrite O_indpaths_beta.
+      rewrite transport_paths_FlFr, !concat_p_pp.
+      rewrite !ap_V, inv_V.
+      rewrite !concat_pV_p.
+      apply whiskerL. apply inverse2.
+      apply ap_compose.
+    Defined.
```
JasonGross: Is `Defined` with `rewrite`s a mistake?

mikeshulman: Yes, thanks.

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: That still doesn't print anything here.

JasonGross: That's unfortunate.  Assuming it is left_associate_concat_in that's being slow: there's a `@?c` that should be `@ ?c`, if the github display is right.  Any chance adding a space there fixes it?

JasonGross: Otherwise, you could stick `let dummy := constr:($(idtac H ":" T)$) in` after `let T := type of H in`, and seeing what `try (timeout 2 ...)` prints out.  That will tell us if it's spinning on some unification problem (we get a few lines, and then it spins without printing anything) or because it keeps doing the same thing over and over (we get a continuous stream of lines until the time runs out).  I don't have access to a computer right now, and am about to go to sleep, but I might get a chance to take a look at it this weekend, if these suggestions don't pan out.

mikeshulman: Adding a space doesn't change anything.  `let dummy` results in `Cannot infer an internal placeholder of type "Type"` when compiling `RewriteModuloAssociativity`.

I opened a dedicated issue for this: #744.

JasonGross: Oh, oops, make it `constr:($(idtac H ":" T; exact tt)$)`.  I'm out of practice with tactics in terms and forgot that the tactic actually has to build something.

mikeshulman: Ok, that gets us somewhere.  We have a continuous stream of lines which start like
```coq
(transport
   (fun _ : dst = to O C ((fun x : A => g0 (f x)) a) =>
    (fun _ : dst = to O C ((fun x : A => g0 (f x)) a) =>
     (fun _ : dst = to O C ((fun x : A => g0 (f x)) a) =>
```
so it does appear to be doing the same thing over and over (whatever that is).

mikeshulman:
I don't think this PR needs to wait on debugging `rewrite@A`, if it is otherwise ok.

PR #746, opened 2015-04-07
If a reflective subuniverse preserves fibers, it is a modality (and lex)

mikeshulman:
Also including the 3x3 lemma for fibers (`hfiber_functor_hfiber`).

### DIFF
--- a/theories/Modalities/Lex.v
+++ b/theories/Modalities/Lex.v
@@ -1,6 +1,6 @@
 (* -*- mode: coq; mode: visual-line -*- *)
 Require Import HoTT.Basics HoTT.Types.
-Require Import Fibrations Extensions Pullback NullHomotopy.
+Require Import EquivalenceVarieties Fibrations Extensions Pullback NullHomotopy.
 Require Import Modality Accessible.
 Require Import HoTT.Tactics.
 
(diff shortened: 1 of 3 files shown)


spitters:
LGTM

theories/Modalities/Lex.v
```diff
@@ -151,6 +181,79 @@ Module Lex_Modalities_Theory (Os : Modalities).
...
+    rewrite concat_pV_p.
+    subst p.
+    rewrite concat_pp_V.
+    rewrite concat_pp_p; apply moveR_Vp.
+    rewrite <- !(ap_compose (to O A) (to O A)^-1).
+    rapply @concat_A1p.
+    Close Scope long_path_scope.
+  Qed.
+
+(** As with [ReflectiveSubuniverses_to_Modalities], Coq won't actually accept this as a module of type [SigmaClosed Os] because the above proof introduces way too many universe parameters.  It might be possible to bring it down to become acceptable with heavy use of universe annotations, but this is probably not worthwhile bothering about because the only point of [SigmaClosed] is to be passed to [ReflectiveSubuniverses_to_Modalities], which in turn can't actually instantiate [Modalities] for more fundamental reasons.  *)
```
JasonGross: Are the "strictly between" universes?  That is, if we do `Definition inO_sigma'` (proof as above), and then do `Definition inO_sigma := @inO_sigma'@{...}`, can we get the right universes?

mikeshulman: Thanks for the suggestion.  Actually on closer inspection it turns out to be also _collapsing_ a couple of universes.  That's a little more worrisome and I feel more obligated to see that it can be prevented.

PR #747, opened 2015-04-07
Fix [rewrite@A]

### LINKED ISSUE #744: `rewrite@A` helper tactic diverges on to_O_natural_compose
As noticed at #743, the following code appears to spin forever:
```coq
Require Import HoTT.
Module Foo (Os : ReflectiveSubuniverses).
  Module Import Os_Theory := ReflectiveSubuniverses_Theory Os.
  Goal Unit.
    let lem' := preconcat_any @to_O_natural_compose in
    pose proof lem' as H.
    left_associate_concat_in H.
  Abort.
End Foo.
```

JasonGross:
When the hypothesis contained `a @ (b @ c)` where the implicit arguments to `concat` did not match the implicit arguments inferred for `@concat _ _ _ _ a (@concat _ _ _ _ b c)` (according to `pattern`),
`left_associate_concat_in` looped, repeatedly finding `a @ (b @ c)`, and failing to associate it to the left, introducing constant transports that did nothing.

This is now fixed, by picking up and passing the implicit arguments, explicitly.

The comment in Modalities/ReflectiveSubuniverse.v has been updated; a naive use of `rewrite@A` causes the subsequent rewrite to fail.  I haven't tracked it down, but I suspect @mikeshulman wouldn't have much trouble updating the file.

This fixes #744; the test is now added to Tests.v.

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -921,7 +921,7 @@ Section Reflective_Subuniverse.
           rewrite concat_pp_p in p.
           apply moveR_Vp in p.
           rewrite@A <- p; clear p.
-          (** for the next one, [rewrite@A] spins forever *)
+          (** TODO: use [rewrite@A] here *)
           apply moveL_pV; rewrite !concat_pp_p; apply moveR_Vp.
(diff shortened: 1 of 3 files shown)


mikeshulman:
LGTM, thanks!

PR #750, opened 2015-04-09
Add a local admit tactic to contrib/Frudenthal.v

JasonGross:
In 8.5beta2, the built-in [admit] which uses [proof_admitted] is
replaced by one aliased to [shelve]; we bring it back so we can use it
in transparent proofs, in contrib/Freudenthal.v.

### DIFF
--- a/contrib/Freudenthal.v
+++ b/contrib/Freudenthal.v
@@ -11,6 +11,8 @@ Local Open Scope path_scope.
 
 Generalizable Variables X A B f g n.
 
+(** [proof_admitted] is used to implement the admit tactic, as in Coq 8.5 beta 1. *)
+Local Axiom proof_admitted : False.
 Local Ltac admit := case proof_admitted.
 
 (* ** Connectedness of the suspension *)
--- a/contrib/Freudenthal.v
+++ b/contrib/Freudenthal.v
@@ -11,6 +11,8 @@ Local Open Scope path_scope.
 
 Generalizable Variables X A B f g n.
 
+Local Ltac admit := case proof_admitted.
+
 (* ** Connectedness of the suspension *)
 
 Global Instance isconn_susp {n : trunc_index} {X : Type} `{H : IsConnected n X}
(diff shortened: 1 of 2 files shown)


mikeshulman:
In what way does `shelve` act as `admit`?  I thought shelved goals had eventually to be unshelved and proven.

JasonGross:
There were two [long](https://sympa.inria.fr/sympa/arc/coqdev/2015-03/msg00005.html) [threads](https://sympa.inria.fr/sympa/arc/coqdev/2015-02/msg00051.html) on coqdev about `admit` and `proof_admitted`.  I will not try to summarize them here, but someone else should feel free to.

The standard library no longer supports transparent proofs with `admit` in any form; the error message is:
```
Error: Attempt to save a proof with given up goals. If this is really what
you want to do, use Admitted in place of Qed. (in proof FST_Codes_cross)
```

mikeshulman:
That seems highly unfortunate to me.

Anyway, this change LGTM.

JasonGross:
Feel free to contribute to the coq-dev discussion, or start a new one.  The main issues were that people do not like inconsistent axioms in the standard library, and, I think, Enrico is not a big fan of global side effects of tactics (which are essentially impossible to implement in asynchronous mode).  As it's easy enough for a project that wants it to add back in an `admit` tactic and an axiom proving `False` to implement it, I guess they didn't want to add back the 8.4 version of `admit` for transparent mode only?  I think it's kind-of unfortunate, too.

mikeshulman:
Ok; to be fair, it made me uneasy too when I learned that there was an inconsistent axiom `proof_admitted` floating around everywhere that we just remain consistent by not using.  In what sense is it a global side effect, though?  And what does the whole thing have to do with transparent/opaque?

JasonGross:
Transparent things can't be asynchronous, as future things need to know about their bodies.  Opaque things can be asynchronous.

Coq 8.4 `admit` added a `thm_admitted` name to the global environment every time you used `admit`.

JasonGross:
Oh, speaking of which, do we want to remove `proof_admitted` from our standard library, now that `admit` no longer uses it?

spitters:
Yes, it would be good to remove it from our stdlib.
Maybe leave a comment on how to put it back when we need it in the future?

LGTM. Feel free to merge after you made the hange to the std lib.

PR #752, opened 2015-04-17
Make Lex_Reflective_Subuniverses actually instantiate SigmaClosed

mikeshulman:
There was one too-big universe coming from an application of `transitivity` to `Equiv` elsewhere, which I got rid of.  The collapsing of universes was due to repleteness not being able to move between universes, which seems like a serious bug in its own right, so I fixed it (though I had to introduce an extra "max" universe everywhere).  And in order to track those down I pretty much had to get rid of all the redundant universes anyway, so now it actually works.

### DIFF
--- a/theories/hit/Truncations.v
+++ b/theories/hit/Truncations.v
@@ -72,7 +72,10 @@ Module Truncation_Modalities <: Modalities.
   Definition to (n : Modality@{u u'}) A := @tr n A.
 
   Definition inO_equiv_inO (n : Modality@{u u'})
-             (A B : Type@{i}) Atr f feq
+             (A : Type@{i}) (B : Type@{j}) Atr f feq
+  : let gei := ((fun x => x) : Type@{i} -> Type@{k}) in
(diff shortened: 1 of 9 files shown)


andrejbauer:
LGTM

theories/hit/Truncations.v
```diff
@@ -72,7 +72,10 @@ Module Truncation_Modalities <: Modalities.
   Definition to (n : Modality@{u u'}) A := @tr n A.
 
   Definition inO_equiv_inO (n : Modality@{u u'})
-             (A B : Type@{i}) Atr f feq
+             (A : Type@{i}) (B : Type@{j}) Atr f feq
+  : let gei := ((fun x => x) : Type@{i} -> Type@{k}) in
```
JasonGross: Note: If you want, you can do `(gei := ...)` before the `:`, rather than `let gei := ... in ...` after the `:`.  It doesn't make much of a difference, other than if you want to pick one for a consistent style.

JasonGross:
Great!

PR #753, opened 2015-04-17
A naive coreflective subuniverse is uniquely determined by an hprop

mikeshulman:
A sort of no-go theorem about what happens if we try to dualize modalities naively.


andrejbauer:
Fine by me once Travis blesses it.

Issue #754, opened 2015-05-08
Universe inconsistency with equiv_induction?


normalchaos:
I have encountered a rather weird bug today when using equiv_induction tactic. Coq claims universe inconsistency, but the code looks innocuous to me.

Require Import HoTT.
Definition nat_discr {n: nat}: 0 <> S n.
  intro p.
  equiv_induction (@equiv_path_nat 0 (S n)).

In nested Ltac calls to "equiv_induction" and
"refine ((fun g H B e => respects_equivalenceL.1 B e g) _ _)" , last call failed. Error:
In environment n : nat; g : ?T; H : ?T0; B : ?T1; e : ?T2
The term " fun (P : Type0)  (_ : 0 =n n.+1 <~> P) => P -> Empty" has type
 "forall P : Type0, 0 =n n.+1 <~> P -> Type" while it is expected to have type
 "forall B0 : Type, ?T3 <~> B0 -> Type" (universe inconsistency).

JasonGross:
It does look rather innocuous, doesn't it.  The bane of `Type0` (of `Set`) strikes again.  @mikeshulman, can you provide input on the right route to take here?  If move `Type1` to `Coq.Init.Datatypes`, and put `nat` in `Type1`, then I get a complaint that `ExtendableAlong` needs 5 universe arguments (and I'm not sure what impact this has on other things, like the rest of modalities, or performance, so I didn't touch it).  If I give `nat` a global `Universe` that is above `Set`, then I get complaints about mismatched module and signature universes (because the signature doesn't say things about the relation of universes to this universe of `nat`'s).  I'm going to see if I can make `equiv_induction` robust in the face of `Set` (maybe with a sprinkling of `uconstr`s), but we might want to move `nat` out of `Set` like we moved `unit`, too.

JasonGross:
[...]

PR #755, opened 2015-05-09
Add some definitions for lifting equivalences

JasonGross:
I wrote them out when trying to work around #754.  I didn't end up needing them, but I figure they might be useful eventually.


spitters:
LGTM

PR #756, opened 2015-05-09
Make [equiv_induction] a bit more robust

JasonGross:
This makes `equiv_induction` work even when typeclass resolution fully solves the side-condition.

Also add a test case from #754.

### DIFF
--- a/theories/Tactics/EquivalenceInduction.v
+++ b/theories/Tactics/EquivalenceInduction.v
@@ -5,9 +5,9 @@ Require Import HoTT.Basics HoTT.Types.
 
 Local Open Scope equiv_scope.
 
-Class RespectsEquivalenceL A (P : forall B, (A <~> B) -> Type)
+Class RespectsEquivalenceL (A : Type@{i}) (P : forall (B : Type@{j}), (A <~> B) -> Type)
   := respects_equivalenceL : { e' : forall B (e : A <~> B), P A (equiv_idmap A) <~> P B e & Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
-Class RespectsEquivalenceR A (P : forall B, (B <~> A) -> Type)
+Class RespectsEquivalenceR (A : Type@{i}) (P : forall (B : Type@{j}), (B <~> A) -> Type)
   := respects_equivalenceR : { e' : forall B (e : B <~> A), P A (equiv_idmap A) <~> P B e & Funext -> equiv_idmap _ = e' A (equiv_idmap _) }.
 (** We use a sigma type rather than a record for two reasons:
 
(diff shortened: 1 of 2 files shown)


spitters:
LGTM

theories/Tactics/EquivalenceInduction.v
```diff
@@ -350,8 +350,9 @@ Ltac equiv_induction p :=
     move p' at top;
       generalize dependent y;
       let P := match goal with |- forall y p, @?P y p => constr:P end in
-      refine ((fun g H B e => (@respects_equivalenceL _ P H).1 B e g) _ _);
-        [ | repeat step_respects_equiv ].
+      (* We use [(fun x => x) _] to block automatic typeclass resolution in the hole for the equivalence respectful proof. *)
```
andrejbauer: I despair every time I see a comment like this ("we trick Coq by standing on our head while holding a glass of beer in the other"). It's a clear sign something is wrong with the programming language that is being used.

mikeshulman: Agreed.

JasonGross: Would
```coq
first [ refine ((fun g H B e => (@respects_equivalenceL _ P H).1 B e g) _ _);
        [ intros | repeat step_respects_equiv ]
      | refine ((fun g H B e => (@respects_equivalenceL _ P H).1 B e g) _ _);
        [ intros ] ].
```
be better?

andrejbauer: No. There is a conceptual problem here that you cannot solve by holding beer in the other hand.

JasonGross: Here is the problem I am trying to solve: We need to pick up the equivalence-respectful proofs from a user-extensible database.  Furthermore, unlike typeclass resolution, we want to make use of partial solutions, rather than undoing our work when we can't fully solve the side-condition.  So we stuff the lemmas in a separate typeclass database keyed off the head constant, and apply those repeatedly while we can.  Because the user might want to key off something more complicated than just the head constant, or might want a backtracking search, we also let normal typeclass resolution fully solve the side-condition, when it can.  The issue with this is that we have no way of saying "if typeclass resolution didn't solve the side condition, then apply this tactic, otherwise do nothing".  (Well, actually, we can say something like that with
```coq
refine ((fun g H B e => (@respects_equivalenceL _ P H).1 B e g) _ _);
[ intros | repeat step_respects_equiv.. ]
```
but this is a bit less robust, because it doesn't error if the `refine` leaves over three holes, or four holes (which it should never do, so I guess it's unclear if it's worth being robust against that).)

So instead, we block typeclass resolution from happening too early, and then apply it manually inside of `step_respects_equiv`.

What is the conceptual issue, and what would you advocate?

andrejbauer: The conceptual issue is this: someone implemented various mechanisms (typeclass resolution, basic tactics, unification, etc) and these are hard-wired. The implementor of all these things tried to make the mechanisms flexible and extensible by the user (by introducing hints databases, type class priorities, various other flags that you can turn on or off). The end result is a complicated machine with many knobs and buttons. You are an expert in tuning the knobs just so that magic happens. But the problems you're solving are not naturally expressed in terms of "turn this knob here and push that button there to get the machine to understand you". Instead, these problems are best solved by having at hand suitable *control mechanisms* (in the sense of programming languages) that will correctly direct the search procedures. Since the control mechanisms are baked in, you're doomed to turning knobs and pushing buttons.

mikeshulman: So the point (which I agree with) is that Coq/Ltac is broken.  But it's not something that we can do any better in the current Coq library, right?

andrejbauer: Right, the current Coq is what it is. Mind you, I am not saying it's a bad tool. Quite the contrary, it's definitely a very good tool with lots and lots of very smart engineering which has been proved to work well in thousands of applications. I am just thinking about the (not so short term) future here.

PR #759, opened 2015-06-09
Move Pointed.v out of Basics and merge it with PType.v

mikeshulman:
There's no reason for it to be in `Basics`, and moving it out gets rid of the arbitrary-seeming division between `Pointed` and `PType`.


theories/Pullback.v
```diff
@@ -63,7 +63,7 @@ Definition pullback_along {A B C} (f : B -> A) (g : C -> A)
 : Pullback f g -> B
   := pr1.
 
-Notation "f ^*" := (pullback_along f) : function_scope.
+Notation "f ^*" := (pullback_along f) (at level 20) : function_scope.
```
JasonGross: Rather than doing this, do you want to make a `Basics/Notations.v` file and put `Reserved Notation "f ^*" (at level 20).` in there, and have `Overture` `Require Export Basics.Notations`?  This will make it less likely that we have notation clashes between files.

JasonGross:
Assuming the change to PType was a copy/paste, LGTM, modulo a response to my comment.

theories/Pullback.v
mikeshulman: Oh, I suppose that would be the correct thing to do...

mikeshulman:
Yes, it was copy-paste.  Now created Basics/Notations.

JasonGross:
Thanks! LGTM

PR #760, opened 2015-06-09
an enumeration of a finite set must repeat itself

mikeshulman:


theories/Spaces/Nat.v
```diff
@@ -111,3 +111,10 @@ Proof.
   apply path_nat.
   apply leq_antisymd; assumption.
 Defined.
+
+Definition not_nltn n : ~ (n < n).
```
mikeshulman: If anyone wants to tell me the name and proof of this from ssreflect, I'm happy to use that.

JasonGross: It's called `Ssreflect.ssrnat.ltnn`:
```coq
Lemma leqnn n : n <= n.                 Proof. by elim: n. Qed.

(* ... *)

Lemma leqNgt m n : (m <= n) = ~~ (n < m).
Proof. by elim: m n => [|m IHm] [|n] //; exact: IHm n. Qed.

Lemma ltnNge m n : (m < n) = ~~ (n <= m).
Proof. by rewrite leqNgt. Qed.

Lemma ltnn n : n < n = false.
Proof. by rewrite ltnNge leqnn. Qed.
```

mikeshulman: Hmm, one of those proofs uses ssreflect tactics, so someone else would have to translate them for me.

JasonGross: The proof of `leqNgt` says: "induct on `m` and destruct `n`; one case is solved by the induction hypothesis applied to `n`; the rest is trivial"

mikeshulman: So how ought these statements to be modified given that we are using decidable hprops instead of booleans?  Just turn the `=`s into `<->`s?

PR #765, opened 2015-06-12
Infinity-groups defined in terms of their classifying spaces

mikeshulman:


theories/Algebra/ooGroup.v
```diff
@@ -0,0 +1,261 @@
...
+: loops X <~> group_loops X.
+Proof.
+  unfold loops, group_type. simpl.
+  exact (equiv_path_sigma_hprop (point X ; tr 1) (point X ; tr 1)).
+Defined.
+
+(** ** Homomorphisms *)
+
+Definition ooGroupHom (G H : ooGroup) :=
+  pMap (B G) (B H).
```
JasonGross: Style nit: Sometimes you do
```coq
Definition foo :=
  bar
```
and sometimes you do
```coq
Definition foo
  := bar
```
Do you want to uniformize them?  (I think you suggested the latter in the style guide?)

mikeshulman: Oops.  Fixed this one, and another one in `ooAction`.  Let me know if you see more.

JasonGross:
Travis says:
```
HOQC theories/Algebra/ooGroup
File "./theories/Algebra/ooGroup.v", line 31, characters 2-109:
Error: Cannot infer the implicit parameter IsPointed of point whose type is
"IsPointed X" (no type class instance found) in environment:
X : pType
make: *** [theories/Algebra/ooGroup.vo] Error 1
```

mikeshulman:
Okay, I guess that changed with the submodule bump.

I'm having trouble seeing this as other than a bug in Coq.  I can do

```coq
pose (x0 := point X).
pose (existT (fun x:X => merely (x = point X)) x0 (tr 1)).
```

and that works, but

```coq
pose (existT (fun x:X => merely (x = point X)) (point X) (tr 1)).
```

gives the error.  I can't think of any valid reason for that to fail, can you?

mikeshulman:
This also works:

```coq
pose (P := fun x:X => (x = point X)).
pose (existT P (point X) 1).
```

I wonder if there is a universe problem?

JasonGross:
This plausibly seems like a bug.  What's the log if you do `Set Typeclasses Debug.  Set Printing Universes. try pose (existT (fun x:X => merely (x = point X)) (point X) (tr 1)).`?  (I don't have the newest version of Coq installed yet.)

mikeshulman:
I don't get anything.

mikeshulman:
I'm also trying the bug minimizer, but having some issues; apparently the assignment of universes is fragile enough that sticking things in a single file changes the number of universe parameters on some things, requiring some tweaking in order to get to the actual error.  Not an ideal situation...

mikeshulman:
And some of that fragile behavior is different for coqc and coqtop...

JasonGross:
> And some of that fragile behavior is different for coqc and coqtop...

You should be able to fix that bit by passing `-top Top` (if `coqtop` is the behavior you want) or `-top $filename` (if `coqc` is the behavior you want).  (Or, uh `-arg " -top" -arg "Top"` or something.)

mikeshulman:
Oh, is that bug not fixed yet?

Have you considered having the bug-finder do that automatically?

JasonGross:
I'm not sure; I haven't been super on top of watching the bug tracker.  @mattam82, does the filename still impact the universe minimization algorithm?

Actually, come to think of it, the latest version of the bug minimizer should do this automatically.  Can you check (a) if the `coq-prog-args` in your file has a `-top` argument, and (b) if you have the latest version?

mikeshulman:
(a) no and (b) yes.

mikeshulman:
Is there a way to instruct the bug finder to leave its temporary files around?  Splitting to statements is changing the number of universe parameters somewhere, but without having the temp file I can't see where to tell which definition it's happening to.

JasonGross:
> Have you considered having the bug-finder do that automatically?

Actually fixed, now, in the latest version.

> Is there a way to instruct the bug finder to leave its temporary files around?

If you pass a third argument, as in `/path/to/find-bug.py in_file.v out_file.v temp_file.v`, then it should leave the most recent temporary file lying around in `temp_file.v` when it exits.

That splitting to statements is doing this is worrying; it really shouldn't change anything other than the spacing.  I've made it so that if you pass `-vv`, you should get the invocation of `coqc` that gets called, in case that's relevant.

awodey:
Mike, this looks like a substantial piece of work -- can you give a brief summary of what it does?

mikeshulman:
@JasonGross I ran the bug-finder and got "failed to split to statements and preserve the error".  Then I ran the exact command line that -vv told me it had run, with the temp file replaced by the filename I had passed as the third argument, and got the correct error instead.

mikeshulman:
@awodey Sure.  The idea is to get a workable definition of "∞-group" (what a classical homotopy theorist would call a "grouplike A∞-space").  The classical definitions using operads or Segal spaces involve infinitely much data, which we don't know how to handle in HoTT.  But instead, we can invoke the theorem (which is a theorem in classical homotopy theory, and also in any ∞-topos) that every ∞-group is the loop space of some pointed connected object, and use it instead as a definition: we *define* an ∞-group to be a pointed connected type (its classifying space or delooping).  Then we make subsidiary definitions to allow us to treat such an object in the way we would expect, e.g. an ∞-group homomorphism is a pointed map between classifying spaces.

Of special note is that we declare a coercion from ∞-groups to types which takes a pointed connected type not to its underlying type, but to its loop space.  Thus, if `G : ooGroup`, then `g : G` means that `g` is an element of the ∞-group that `G` is intended to denote, which is the *loop space* of the pointed connected type that is technically the data of which `G : ooGroup` consists.  This makes it easier to really think of `G` as "really being" an ∞-group rather than its classifying space.

This is also convenient because elements of ∞-groups *are*, definitionally, loops in some type.  Thus, the ∞-group operations like multiplication, inverse, associativity, higher associativity, etc. are simply special cases of the corresponding operations for paths.

Beyond this, there's not much except basic theorems about the functoriality of ∞-group homomorphisms and so on.  And a definition of the "automorphism ∞-group" of a type, whose classifying space is the type of all types merely equivalent to it.

Oh, and a slightly sneaky tactic, which piggybacks on one that I wrote for pointed types, that often enables us to pretend that a pointed map of classifying spaces is *strictly* pointed, i.e. preserves the basepoint judgmentally.  In classical homotopy theory, one generally works with strictly pointed maps, since it is a colossal pain to have to carry around paths witnessing preservation of basepoints everywhere.  In HoTT we have to include those witness paths in the definition of a pointed map, but with some judicious application of path induction we can often make them vanish when doing proofs; the tactic automates that.

mikeshulman:
And now that I've written all that, maybe I'll put it in the comments in the file.

JasonGross:
> I ran the bug-finder and got "failed to split to statements and preserve the error". Then I ran the exact command line that -vv told me it had run, with the temp file replaced by the filename I had passed as the third argument, and got the correct error instead.

I've updated -vv so that it displays the error that the command gives (assuming I did it right).  But I think I see what's happening, here.  `coqc` ignores the `-top` argument (you can see this by putting `Axiom a : Type.  Locate a.` into a .v file and running `coqc` on it).  So the number of universes changes, because the temporary file has a different name.  (If you copy over the remaining temporary file to the name the bug-minimizer gives you, does that break things?)  Here's a work-around: use the following script which pretends that it's coqc (i.e., accepts `foo.v` rather than just `foo`), but actually accepts `-top`:
```bash
#!/bin/bash
function strip_v() {
    for i in "$@"
    do
        j="${i%.*}"
        if [ "$j.v" = "$i" ]; then
            echo "-load-vernac-source" # or -compile
            echo "$j"
        else
            echo "$i"
        fi

    done
}

cd "$(dirname "$0")"

exec ./hoqtop $(strip_v "$@")
```
and pass the path to this script as `--coqc`.

mikeshulman:
> If you copy over the remaining temporary file to the name the bug-minimizer gives you, does that break things?

Yep.  That's annoying that `coqc` accepts the `-top` argument but then silently ignores it.

Now trying your fake hoqc.

JasonGross:
I've updated `find-bug.py`; you should now be able to get this behavior by passing `--coqtop ../hoqtop --coqc-is-coqtop` (and either don't pass a `--coqc` argument, or pass `--coqc ../hoqtop`).

mikeshulman:
@JasonGross now it fails on `get_coqc_version`.

JasonGross:
Eww, try passing `--coq-is-coqtop` without passing any `--coqc` argument, and open an issue that I need to make `get_coqc_version` robust to being passed coqtop?  (If the first suggestion doesn't work, try modifying the python file and having `get_coqc_version` just `return ''`

mikeshulman:
I didn't pass any `--coqc`.  Will try modifying the file when I get back to this later.

mikeshulman:
If `get_coqc_version` returns `''`, it complains about `get_coqc_help`.  If that also returns `''`, then it dies on `deduplicate_trailing_dir_bindings`.

JasonGross:
Should now be fixed.  Try it again?  And let me know what the error message that it fails with if it fails again?

mikeshulman:
Moving to one-on-one email to stop spamming everyone else.

spitters:
Nice! When Travis is happy please feel free to merge.

mikeshulman:
Coq [bug report](https://coq.inria.fr/bugs/show_bug.cgi?id=4256) submitted, and workaround added to this PR.  Do we have APOE than @spitters ?

theories/Algebra/ooGroup.v
```diff
@@ -0,0 +1,281 @@
...
+  rewrite <- p.
+  unfold grouphom_fun.
+  rewrite !loops_functor_group.
+  apply ap.
+  apply loops_2functor; assumption.
+Defined.
+
+(** *** Homomorphic properties *)
+
+(** The following tactic often allows us to "pretend" that phi preserves basepoints strictly.  This is basically a simple extension of [pointed_reduce], which see. *)
```
JasonGross: You seem to be missing some words around "which see"

mikeshulman: Hmm, I meant it as an English translation of "q.v.".

JasonGross:
LGTM, modulo my documentation comment, and Travis.

theories/Algebra/ooGroup.v
JasonGross: I'd put something like "(see $filename.v)", I think?

PR #768, opened 2015-06-24
Add a _CoqProject file

JasonGross:
This almost replaces the .dir-locals.el.  Additionally, coqide has
special support for _CoqProject, and this is the standard way of
listing .v files in a project.


mikeshulman:
What do you mean by "almost replaces"?

JasonGross:
I mean that if we only needed to pass special arguments to the system coqtop, then we could replace `.dir-locals.el` with `_CoqProject`.  But there's no way to set the coqtop binary in _CoqProject in a way that ProofGeneral and CoqIDE understand.  If we want, we can copy over the flags from `hoqtop` to `_CoqProject` (`-no-native-compiler -coqlib "@hottdir@/coq" -R "@hottdir@/theories" HoTT -indices-matter` if we make a `_CoqProject.in` and require re-running `autogen.sh` every time a file is added, or hard-coding it to `-no-native-compiler -coqlib coq -R theories HoTT -indices-matter`), but it will then use the system coqtop, rather than the pinned one.

mikeshulman:
So ProofGeneral also reads `_CoqProject`?  It's weird if you can set the `coqc` there but not the `coqtop`.

JasonGross:
New enough ProofGenerals read `_CoqProject`.  They ignore settings of coqc, coqtop, etc.  `coq_makefile` respects settings of coqc, coqtop, etc, but setting coqtop for `coq_makefile` doesn't do much.

mikeshulman:
Ok, so by "almost replaces" you meant that _if_ ProofGeneral paid more attention to `_CoqProject`, then we _would_ be able to use it to replace `.dirlocals.el`, but as things stand, there isn't actually any sense in which it replaces any part of `.dirlocals.el`.

JasonGross:
Yes, that's correct.

mikeshulman:
np.  Do you want to modify `STYLE.md` to tell people to add new files to `_CoqProject` instead of `Makefile_targets.mk`?

JasonGross:
> modify `STYLE.md`

Done

mikeshulman:
In case anyone else gets bitten by this: after pulling from master with this PR merged, you have to re-run `configure` before `make`.

PR #771, opened 2015-07-07
Simplify some proofs in Spaces/BAut/Bool

mikeshulman:
A couple of these proofs were really unnecessarily complicated.  I'm not sure what I was thinking.  The shorter proofs speed the file up by almost 1/3.

Amusingly, I discovered this while _informalizing_ these proofs for a paper I'm writing.  There's probably a lesson in that.


andrejbauer:
Hmm, I forgot about the two pairs of eyes rule, but this seems completely unproblematic to me.

PR #776, opened 2015-07-15
define TwoSphere recursion

jdoughertyii:
I changed the type of S2_ind to match Chapter 6 of the HoTT Book, and changed S2_ind_beta_loop to S2_ind_beta_surf. I added some lemmas used in the definition of S2_rec_beta_surf to PathGroupoids.v.

I also constructed an equivalence S2 <~> Sphere 2 ([here](https://github.com/jdoughertyii/HoTT/blob/spheres_with_S2/theories/hit/Spheres.v#L78)), but it's not included in this pull request because it takes a full minute and a half to compile.  I don't know if that's something that would be worth trying to improve and submit.

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -963,6 +963,27 @@ Definition concat2_1p {A : Type} {x y : A} {p q : x = y} (h : p = q) :
   :=
   match h with idpath => 1 end.
 
+Definition cancel2L {A : Type} {x y z : A} {p p' : x = y} {q q' : y = z} 
+           (g : p = p') (h k : q = q')
+: (g @@ h = g @@ k) -> (h = k).
+Proof.
+  intro r. induction g, p, q.
+  refine ((whiskerL_1p h)^ @ _). refine (_ @ (whiskerL_1p k)). 
+  refine (whiskerR _ _). refine (whiskerL _ _).
+  apply r.
+Defined.
+
+Definition cancel2R {A : Type} {x y z : A} {p p' : x = y} {q q' : y = z} 
+           (g h : p = p') (k : q = q')
+: (g @@ k = h @@ k) -> (g = h).
+Proof.
+  intro r. induction k, p, q.
+  refine ((whiskerR_p1 g)^ @ _). refine (_ @ (whiskerR_p1 h)).
+  refine (whiskerR _ _). refine (whiskerL _ _).
+  apply r.
+Defined.
+
+
 (** Whiskering and composition *)
 
 (* The naming scheme for these is a little unclear; should [pp] refer to concatenation of the 2-paths being whiskered or of the paths we are whiskering by? *)
@@ -1089,6 +1110,16 @@ Definition apD02 {A : Type} {B : A -> Type} {x y : A} {p q : x = y}
   (f : forall x, B x) (r : p = q)
   : apD f p = transport2 B r (f x) @ apD f q
   := match r with idpath => (concat_1p _)^ end.
+
+Definition apD02_const {A B : Type} (f : A -> B) {x y : A} {p q : x = y} (r : p = q)
+: apD02 f r = (apD_const f p) 
+              @ (transport2_const r (f x) @@ ap02 f r)
+              @ (concat_p_pp _ _ _)^
+              @ (whiskerL (transport2 _ r (f x)) (apD_const f q)^)
+:=
+  match r with idpath =>
+    match p with idpath => 1
+    end end.
 
 (* And now for a lemma whose statement is much longer than its proof. *)
 Definition apD02_pp {A} (B : A -> Type) (f : forall x:A, B x) {x y : A}
--- a/theories/hit/TwoSphere.v
+++ b/theories/hit/TwoSphere.v
@@ -17,30 +17,55 @@ Private Inductive S2 : Type0 :=
 
 Axiom surf : idpath base = idpath base.
 
-Definition S2_ind (P : S2 -> Type) (b : P base) (l : transport2 P surf b = idpath b)
+Definition S2_ind (P : S2 -> Type) (b : P base) (s : idpath b = transport2 P surf b)
   : forall (x:S2), P x
-  := fun x => match x with base => fun _ => b end l.
+  := fun x => match x with base => fun _ => b end s.
 
-Axiom S2_ind_beta_loop
-  : forall (P : S2 -> Type) (b : P base) (l : transport2 P surf b = idpath b),
-      apD02 (S2_ind P b l) surf = l^ @ (concat_p1 _)^.
+Axiom S2_ind_beta_surf
+  : forall (P : S2 -> Type) (b : P base) (s : idpath b = transport2 P surf b),
+      apD02 (S2_ind P b s) surf = s @ (concat_p1 _)^.
 
 End TwoSphere.
 
 (* ** The non-dependent eliminator *)
 
-Definition S2_rec (P : Type) (b : P) (l : idpath b = idpath b)
+Definition S2_rec (P : Type) (b : P) (s : idpath b = idpath b)
   : S2 -> P
-  := S2_ind (fun _ => P) b ((concat_p1 _)^ @ (transport2_const surf b)^ @ l).
+  := S2_ind (fun _ => P) b (s @ (transport2_const surf b) @ (concat_p1 _)).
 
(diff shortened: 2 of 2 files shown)


mikeshulman:
Nice!  I think it would be definitely worth trying to improve the other equivalence and include it in the library.

theories/Basics/PathGroupoids.v
```diff
@@ -410,6 +410,13 @@ Definition concat_Ap {A B : Type} {f g : A -> B} (p : forall x, f x = g x) {x y
     | idpath => concat_1p _ @ ((concat_p1 _) ^)
   end.
 
+Definition ap_homotopic {A B : Type} {x y : A} {p : x = y}
```
mikeshulman: This is a rearrangement of `concat_Ap`; does it need another name?

theories/Basics/PathGroupoids.v
```diff
@@ -863,8 +870,47 @@ Definition inverse2 {A : Type} {x y : A} {p q : x = y} (h : p = q)
...
+: inverse2 (s @@ t) 
+  = 
+  (inv_pp p q) @ ((inverse2 t) @@ (inverse2 s)) @ (inv_pp p' q')^.
+Proof.
+    by path_induction.
+Defined.
+
 (** Some higher coherences *)
 
+Lemma ap_concat {A B : Type} {x' y' z' : B}
```
mikeshulman: I feel like there should be a better name for this, but I don't know what.

theories/Basics/PathGroupoids.v
```diff
@@ -879,6 +925,19 @@ Proof.
   destruct p; reflexivity.
 Defined.
 
+Lemma concatppp_pp_1 {A : Type} {x y z : A}
```
mikeshulman: What led you to the name for this lemma?

theories/Basics/PathGroupoids.v
```diff
@@ -891,6 +950,63 @@ Proof.
   destruct r, p; reflexivity.
 Defined.
 
+Definition concat2_V2p {A : Type} {x y : A} {p q : x = y} (r : p = q)
+: (inverse2 r @@ r) = concat_Vp p @ (concat_Vp q)^.
+Proof.
+  by path_induction.
+Defined.
+
+Definition concat2_p2p_p2 {A : Type} {x y z w : A}
```
mikeshulman: Actually I don't understand the names of most of these.  Where are you inserting `2`s and where not?

theories/Basics/PathGroupoids.v
```diff
@@ -891,6 +950,63 @@ Proof.
...
+           {p p' p'' : x = y} {q q' : y = z} (s : p = p') (t : p' = p'')
+           (r : q = q')
+: (s @ t) @@ r
+  =
+  (s @@ r) @ (t @@ 1).
+Proof.
+    by path_induction.
+Defined.
+
+Definition concat2_pp_p_r {A : Type} {x y z : A}
```
mikeshulman: What does `_r` mean?

theories/hit/TwoSphere.v
```diff
@@ -17,30 +17,55 @@ Private Inductive S2 : Type0 :=
...
-...
->>
-*)
+  apply (cancel2L (transport2_const surf b)).
+  apply (cancelL (apD_const (S2_rec P b s) (idpath base))).
+  apply (cancelR _ _ (concat_pp_p _ (transport_const _ b) _)).
+  apply (cancelR _ _ (whiskerL (transport2 _ surf b) (apD_const _ _)^)).
+  refine ((apD02_const (S2_rec P b s) surf)^ @ _).
+  refine ((S2_ind_beta_surf _ _ _) @ _).
+  hott_simpl.
```
mikeshulman: I think if we set `Arguments concat_pp_p : simpl nomatch.` (which we should do anyway), then this line and the next can be replaced by `cbn. hott_simpl. apply moveL_Mp.` making the proof a bit faster.

theories/hit/TwoSphere.v
```diff
@@ -17,30 +17,55 @@ Private Inductive S2 : Type0 :=
...
+  apply (cancelR _ _ (concat_pp_p _ (transport_const _ b) _)).
+  apply (cancelR _ _ (whiskerL (transport2 _ surf b) (apD_const _ _)^)).
+  refine ((apD02_const (S2_rec P b s) surf)^ @ _).
+  refine ((S2_ind_beta_surf _ _ _) @ _).
+  hott_simpl. 
+  apply moveL_pM. hott_simpl. apply moveR_pV, moveL_Mp.
+  path_via (concat_pp_p (transport2 (fun _ : S2 => P) surf b) 
+     (transport_const 1 b) 1).
+  refine (_ @ (concatppp_pp_1 _ _)^). hott_simpl.
+  refine (_ @ (apD concat_p1 (transport2_const _ _))). simpl.
```
mikeshulman: My previous suggestion seems to require passing the arguments explicitly here `(transport2_const surf b)`

theories/Basics/PathGroupoids.v
jdoughertyii: Perhaps not.  `concat_Ap` seems inconvenient to me, because when I want to apply it I have `ap f p` and want to replace it with something.  But applying `concat_pV` and then `concat_Ap` would also work.

theories/hit/TwoSphere.v
```diff
@@ -17,30 +17,55 @@ Private Inductive S2 : Type0 :=
...
+  refine (@dpath_path_FlFr _ _ (fun p : b = b => p @ 1) (fun p : b = b => p) _ 
+                           _ _ _ _ _).
+  refine ((concat_1p _) @ _). hott_simpl. apply moveL_pM.
+  refine ((concat_pV _) @ _). apply moveL_pM. 
+  refine ((concat_1p _) @ _). apply moveL_Mp.
+  refine ((inv_V _)^ @ _). apply inverse2.
+  refine ((inv_pp _ _) @ _). apply moveR_pV.
+  refine ((inv_V _) @ _). apply moveL_Mp.
+  path_via ((transport2_const surf b @@ s)^ @ (1 @@ s)).
+  f_ap. refine (_ @ (concat2_1p s)^). refine (whiskerL_1p_1 _)^.
```
mikeshulman: This line should probably be in `{ ... }` like the later ones.

theories/Basics/PathGroupoids.v
mikeshulman: Did you actually use this lemma anywhere in this PR?

jdoughertyii: Hmm, no.  It looks like the version of `PathGroupoids.v` in this PR has all the stuff from the branch with the S2 <~> Sphere 2 equivalence.  I must have gotten them mixed up in the rebase shuffle.

theories/Basics/PathGroupoids.v
jdoughertyii: Since it's about `concat_pp_p`, I prefixed it with `concatppp_*` (I thought I saw this underscore-removal elsewhere, but now I can only find examples like `ap_pp_concat_pV` that don't remove them).  The `_pp_1` is supposed to indicate that it's `concat_pp_p` and that the first two arguments are arbitrary paths and the third is `idpath`.  Maybe `concat_pp_p_pp_1` or `concat_pp_p_pp1` would be better.

This, `apD02_const`, `cancel2L`, and `concat2_V` are the only additions to `PathGroupoids.v` needed in this PR, so if we settle on a name for this I'll update the PR to remove all of the other additions.

mikeshulman: My vote would probably be for `concat_pp_p_pp1`, although at this level we do seem to be stretching our naming conventions to the limit.

mikeshulman: One might suggest `concat_pp_p__pp1` to visually separate the name of the lemma from its arguments, but I don't know if that would be more ugly than it's worth.

jdoughertyii:
I cut the knot on the naming thing.  I found a shorter definition for `S2_rec_beta_surf` that's also faster (bringing the compile time of `S1_rec_beta_surf` from 4.12s to 0.81s) and only adds `apD02_const` and `cancel2*` to `PathGroupoids.v`

jdoughertyii:
Shoot. Didn't mean to close it.

mikeshulman:
Sorry for the delay.  LGTM!  APOE?

PR #778, opened 2015-07-21
all modal generators inhabited -> all hprops are modal

mikeshulman:


andrejbauer:
I am not really following the math here, but LGTM as far as Coq is concerned.

PR #782, opened 2015-08-06
S2 <~> Sphere 2

jdoughertyii:
Construct an equivalence S2 <~> Sphere 2. Rename beta principles for suspensions to match other HITs.  The compile time for Spheres.v is up to 3.3s, but this is as fast as I could get it.


mikeshulman:
LGTM too once Travis agrees.  This is an impressive-looking calculation!

PR #787, opened 2015-12-09
rewriting with reflexivity

mikeshulman:
Here's a little tactic that takes a lemma whose definition is `idpath` and behaves like `rewrite`, except that it doesn't insert any transport lemmas like `Overture.internal_paths_rew_r`.  In other words, it does a `change`, but leverages the pattern-matching and substitution engine of `rewrite` to decide what to `change` into.  I don't have any applications in mind for it, but the potential usefulness of such a tactic came up on the Lean mailing list, so I thought I would try implementing it in Coq.

I also moved `eval_in` and friends into their own file, since I could never remember where to look for them when they were in `BinderApply.v`.


JasonGross:
Assuming the move is really just cut and paste (I didn't check too carefully), lgtm.

As an aside, we'll have to replace $(...)$ with ltac:(...) when we move to 8.5beta4 or whatever.

Issue #784, opened 2015-10-02
Universe Polymorphism Issue


td202:
### Context

Using the HoTT library, it is difficult to construct `IsHSet` instances for composite types involving `nat`. There seem to be a couple of contributing factors:
- `nat` is at type `Set` whereas other basic types (e.g. `Unit` and `Empty`) are at `Type1` (which is a higher universe).
- the universe polymorphism is insufficiently polymorphic.

Consider:

``` Coq
Lemma hset_nat_plus_Unit : IsHSet (nat + Unit).
  refine (@hset_sum _ _ _ _). (* Fails *)
```

This code fails because `hset_sum` expects the type of `nat` to be at least `Type1` (effectively).
This suggests the following alternative:

``` Coq
Lemma hset_nat_plus_Unit : IsHSet ((nat:Type1) + Unit).
  refine (@hset_sum _ _ _ _).
  (* Goal: ISHSet nat *)
  exact hset_nat. (* Fails *)
```

The problem here seems to be that `hset_nat` considers `nat` at type `Set`, but what is needed is `nat` at type `Type1`.  I would expect the universe polymorphism to support this.
### Example

Here is an example that illustrates the issue:

``` Coq
Definition Type1 := Eval hnf in let gt := (Set : Type@{i}) in Type@{i}.
Inductive Unit : Type1 := tt : Unit.
Inductive A : Type := a : A | b : A.
Definition P (A : Type) := Unit.
Definition foo : P A := tt.
Lemma bar : P (A : Type1).
  exact foo. (* Fails with universe inconsistency *)
```

In HoTT, or in Coq (8.5beta2) with the definitions of `P` and `foo` polymorphic, this fails.  In Coq without the definitions being polymorphic, it works.  (Maybe this is more of a Coq issue than a HoTT issue?)

mikeshulman:
Probably related to #754, #774.

mattam82:
Hi, your example should not fail as P@{Top.1} (A : Type1) and P@{Set} A (note that A : Set) both reduce to Unit. I did in 8.5beta2 but doesn't in the current 8.5 branch, this was a bug. In your original problem, you expect IsHSet@{i} to be coercible to IsHSet@{j} with i < j. The universe system does not support this currently and you have to use an explicit coercion there. We are working on extending cumulativity to inductive so that this is indeed accepted.

spitters:
Coq trunk now has cumulative inductive types.
[...]

PR #788, opened 2016-01-07
Proof of [IsHSet nat] added to Nat.v

marcbezem:
Proved by the encode/decode argument from Section 2.13

### DIFF
--- a/theories/Types/Nat.v
+++ b/theories/Types/Nat.v
@@ -3,6 +3,8 @@
 
 Require Export Coq.Init.Peano.
 Require Import HoTT.Basics.
+Require Import HoTT.Types.Empty.
+Require Import HoTT.Types.Unit.
 Require Import HoTT.Types.Bool.
 
 (** We reopen these scopes so they take precedence over nat_scope; otherwise, now that we have [Coq.Init.Peano], we'd get [* : nat -> nat -> nat] rather than [* : Type -> Type -> Type]. *)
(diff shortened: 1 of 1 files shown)


marcbezem:
I see only now that there is already a hset_nat  in theories/Spaces/Nat.v. I have several questions now:
- Should I just close the pull request?
- Should the name conflict not be detected? Or maybe there is no conflict?
- How to avoid such duplicates in the future? It seems impossible to overlook the whole library.
I found it by coincidence, reading issue #784. Then a "grep -r hset_nat ." found it.
- BTW, "grep -r hprop_Empty ." succeeds, "grep -r hprop_Unit ." fails.

Thanks in advance (doing the proof was instructive anyway).

andrejbauer:
I think this pull request could be closed, but not before we discuss how we got to have both `types/Nat.v` and `Spaces/Nat.v`. Any idea?

mikeshulman:
Probably first we had `types/Nat.v` before `Spaces/` existed, and then once `Spaces/` was created I started putting stuff in `Spaces/Nat.v` because I thought it was a better fit (since `nat` is a particular type rather than a general type former), but never got around to moving everything that was already in `types/Nat.v`.  Unless there is some dependency reason that some facts about nat need to be in `types/`?

marcbezem:
I would like to have one file Nat.v, but have no strong opinion whether it should be under Types or Spaces. Mike's argument about type formers would also apply to Empty, Unit and Bool, wouldn't it? Another view that could be defended is that they are all, like nat, nullary type formers.

For discovering duplicates, it would be nice to have some function searching the library. In hoqide,
if one selects "nat" in either Nat.v, and then clicks "Queries/Search" one gets dependent_choice, nat_rec, nat_rect, S and 0, with all there types. It seems to search Coq.Init.Datatypes.nat. But I'm aware that "Tell me all you know about nat" will return quite a lot. However, the Proviola-generated index *has* an entry "hset_nat [in HoTT.Spaces.Nat]", so in principle I could have avoided the duplication.

mikeshulman:
Right, nat is of course a nullary type former.  But that argument could also be applied to most things in `Spaces/`, so we have to draw the line somewhere.  For Empty and Unit, at least, I think there are dependency arguments: `Types/` is lower down the dependency ladder, and Empty and Unit are used in lots of other places in the library.  The situation with Bool is admittedly less clear; I could probably see an argument for moving Bool into `Spaces/`.

marcbezem:
What exactly does "Spaces" express? Empty and Unit are hprop's, Bool is not. That could be an argument to move Bool and nat to Spaces. Browsing Spaces, I already found a Bool.v there:

 HoTT/theories/Spaces/BAut/Bool.v

Again, I would prefer one file Bool.v as a plausible place to look for results on Bool.

mikeshulman:
Baut/bool is not about bool, it's about baut(bool).

Spaces is about 'particular spaces' rather than 'type formers'. As we've
seen this is not a totally clear boundary, but I think of it as types is
part of the foundations whereas spaces is the mathematics we do in that
foundations.

marcbezem:
I think the issue has been sufficiently discussed so that this pull request can now be closed.

mikeshulman:
Another pull request unifying the two nat.v's would be welcome.

marcbezem:
Sure, I can do that. It may even be the case that Types/nat.v duplicates stuff in the Coq-library, and is actually obsolete. Adding hprop_Unit to Unit.v is also on my list. (May take some time, very busy right now.)

mikeshulman:
We didn't copy very much from the Coq library into the HoTT library, but you can look to see.  I think hprop_Unit should be inferrable automatically from contr_Unit and trunc_succ.

marcbezem:
It is good to discuss this before I do something. Here are two examples:

Types/Nat.v has Definition nat_plus_comm (n m : nat) : n + m = m + n.
https://coq.inria.fr/library/Coq.Arith.Plus.html has Lemma plus_comm : forall n m, n + m = m + n.

Both https://coq.inria.fr/library/Coq.Arith.Factorial.html and Types/Nat.v have
Fixpoint factorial (n : nat) : nat
:= match n with
| 0 => 1
| S n => S n * factorial n
end.

There might be subtle differences (Qed vs Defined, not here).
I have not yet tested/changed anything. 
We should first make a decision on policy. This could be: reuse as much of the Coq library as is possible/convenient. If so, Types/Nat.v becomes obsolete, as far as I can see.

andrejbauer:
This issue is closed. Let's move the discussion to a separate issue, see #790

PR #792, opened 2016-01-16
Types/Nat.v integrated in Spaces/Nat.v

marcbezem:
Types/Nat.v integrated in Spaces/Nat.v, with the necessary changes in some other files.


marcbezem:
Interesting that it fails: make contrib (after make clean) went OK. Are there other dependencies on Types/Nat.v that should be updated?
A minor thing that annoys me is that my local history (the first three commits, which cancel out) pollutes the history of the HoTT library. What I did was

    git revert --no-commit HEAD~2..HEAD

to get rid of the first two commits. Is there a better way?

mikeshulman:
You need to edit `_CoqProject` so that `make` doesn't try to compile the nonexistent `Types/Nat.v`.

marcbezem:
`_CoqProject`was hard to find, but showed up with `ls -a`. Check with `make` went OK.

mikeshulman:
LGTM.

spitters:
The commit sequence looks confusing. @JasonGross will know by heart how to cherrypick to commits, but it's worth reading the git documentation for that.

mikeshulman:
One simple fix for a confusing commit sequence is to `git reset --soft master` and then `git commit` and `git push -f`.  This squashes all the commits down to one new one.

However, I don't think we need to enforce that contributors clean up their commits -- a clean commit sequence is nice, but less important than being as open and encouraging as possible to newcomers.  So if another pair of eyes is fine with the changes themselves, I think we can go ahead and merge.

PR #798, opened 2016-03-21
A bit of messing around with spectra

mikeshulman:
I'm not planning to do any more of this in Coq right now, but I thought I might as well commit it so that anyone else who wants to build on it can.

### DIFF
--- a/theories/Pointed.v
+++ b/theories/Pointed.v
@@ -323,6 +323,14 @@ Proof.
   apply moveR_equiv_V; symmetry; apply point_eq.
 Defined.
 
+Definition pequiv_compose {A B C : pType}
+           (f : A <~>* B) (g : B <~>* C)
+: A <~>* C
+  := (Build_pEquiv A C (g o* f) isequiv_compose).
+
+Notation "g o*E f" := (pequiv_compose f g) (at level 40, left associativity) : pointed_scope.
+
+
 (** ** Equivalences *)
 
 Definition issig_ptype : { X : Type & X } <~> pType.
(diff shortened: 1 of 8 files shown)


andrejbauer:
Other than the fact that I generally distrust Coq's coinductive types, I am fine with this.

theories/Pointed.v
```diff
@@ -323,6 +323,14 @@ Proof.
   apply moveR_equiv_V; symmetry; apply point_eq.
 Defined.
 
+Definition pequiv_compose {A B C : pType}
+           (f : A <~>* B) (g : B <~>* C)
+: A <~>* C
+  := (Build_pEquiv A C (g o* f) isequiv_compose).
+
+Notation "g o*E f" := (pequiv_compose f g) (at level 40, left associativity) : pointed_scope.
```
JasonGross: I would suggest reserving this notation in notations.v

theories/Spectra/Spectrum.v
```diff
@@ -0,0 +1,46 @@
...
+
+Record Spectrum :=
+  { to_prespectrum :> Prespectrum
+    ; to_is_spectrum : IsSpectrum to_prespectrum }.
+
+Existing Instance to_is_spectrum.
+
+(** ** Truncations of spectra *)
+
+Fixpoint inc (n : nat) (k : trunc_index) : trunc_index
```
JasonGross: I think this belongs in a more general file.

JasonGross:
LGTM, modulo my comments

mikeshulman:
Thanks, fixed!

mikeshulman:
(I don't trust Coq's coinductive types either, but I thought it would be fun to see what spectra would be like in that language.)

andrejbauer:
LGTM, @JasonGross will you accept it?

theories/Pointed.v
JasonGross: Since you've now reserved the notation in `Notations.v`, you can drop the `(at level 40, left associativity)` here and instead `Import HoTT.Basics.Notations` at the top.

JasonGross:
I've updated my comment on the notation.  Modulo that, LGTM.

mikeshulman:
Ok, fixed.  (HoTT.Basics was already Imported.)

JasonGross:
Thanks!  LGTM if travis is happy.

PR #799, opened 2016-04-07
proved lemma and definitions

georgydunaev:


mikeshulman:
LGTM, although probably eventually the theory of retracts should be moved out of `Idempotents.v`.

PR #808, opened 2016-06-09
Bugfix #4527 Add universe annotations

mattam82:
To be compatible with the new semantics of explicit universe
declarations in lemma statements (Coq pr #177, forward compatible).


spitters:
Thanks. LGTM.

Issue #168, opened 2013-08-07
Enable Travis-CI (Continuous Integration) for HoTT


JasonGross:
Someone who has admin bits to the repo should go to https://travis-ci.org/profile/HoTT and log in with github and enable the (free) Travis-CI service, so that we get to see whether or not pull requests build.  (Warning: It sends email to whichever email is signed up to manage the repo whenever builds pass/fail.  I think there's a way to change this, though I haven't looked into it yet.  I can, if that's desired.)

PR #809, opened 2016-06-09
Fix universe annotation that was silently ignored

mattam82:
And now produces the right length-mismatch error
with Coq's PR #168 (backward compatible change).


mikeshulman:
LGTM.

Does anyone know if there's a standard way in github markdown to link to an issue in another repository?  When you type `#168` it auto-links it to issue 168 in this repository, which is confusing.

JasonGross:
`coq/coq#168` gives coq/coq#168 ,  `https://github.com/coq/coq/pull/168` gives https://github.com/coq/coq/pull/168

PR #813, opened 2016-06-27
Put Empty and Unit in Type0

SkySkimmer:
The only complicated fixing was in Modalities which needed changes in the universe annotations.
Universe minimization to Set is globally disabled so that we get lemmas like `Contr@{i} Unit` for all `i`. If Coq had cumulative inductive types like in http://advent-project.eu/pubs/cit-ictac15.pdf it might not be needed.
It is locally re-enabled in Idempotents to avoid a blowup in the number of universes (IsIdempotent goes from 4 to 51 universes), and for `Book_3_14` in the exercises to avoid some error.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -517,7 +517,8 @@ Section Book_3_14.
       apply hprop_allpath.
       intros x' y'.
       etransitivity; [ symmetry; apply (p x x y' x') | ].
-     (* Without this it somehow proves [H'] using the wrong universe for hprop_Empty and fails when we do [Defined]. *)
+     (* Without this it somehow proves [H'] using the wrong universe for hprop_Empty and fails when we do [Defined].
+        See Coq #4862. *)
       set (path := path_ishprop x x).
       assert (H' : idpath = path) by apply path_ishprop.
       destruct H'.
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -505,8 +505,6 @@ Section Book_3_14.
   Context `{Funext}.
   Hypothesis LEM : forall A : Type, IsHProp A -> A + ~A.
 
-  Local Set Universe Minimization ToSet.
-
   Definition Book_3_14
   : forall A (P : ~~A -> Type),
     (forall a, P (fun na => na a))
@@ -519,7 +517,9 @@ Section Book_3_14.
       apply hprop_allpath.
       intros x' y'.
       etransitivity; [ symmetry; apply (p x x y' x') | ].
-      assert (H' : idpath = path_ishprop x x) by apply path_ishprop.
+     (* Without this it somehow proves [H'] using the wrong universe for hprop_Empty and fails when we do [Defined]. *)
+      set (path := path_ishprop x x).
+      assert (H' : idpath = path) by apply path_ishprop.
       destruct H'.
       reflexivity.
     - destruct (LEM (P nna) _) as [pnna|npnna]; trivial.
@@ -529,8 +529,6 @@ Section Book_3_14.
       apply npnna.
       exact (transport P (path_ishprop _ _) (base a)).
   Defined.
-
-  Local Unset Universe Minimization ToSet.
 
   Lemma Book_3_14_equiv A : merely A <~> ~~A.
   Proof.
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -505,6 +505,8 @@ Section Book_3_14.
   Context `{Funext}.
   Hypothesis LEM : forall A : Type, IsHProp A -> A + ~A.
 
+  Local Set Universe Minimization ToSet.
+
   Definition Book_3_14
   : forall A (P : ~~A -> Type),
     (forall a, P (fun na => na a))
@@ -527,6 +529,8 @@ Section Book_3_14.
       apply npnna.
       exact (transport P (path_ishprop _ _) (base a)).
   Defined.
+
+  Local Unset Universe Minimization ToSet.
 
   Lemma Book_3_14_equiv A : merely A <~> ~~A.
   Proof.
--- a/theories/Extensions.v
+++ b/theories/Extensions.v
@@ -97,7 +97,7 @@ Section Extensions.
            (n : nat) {A : Type@{i}} {B : Type@{j}}
            (f : A -> B) (C : B -> Type@{k}) : Type@{l}
     := match n with
-         | 0 => Unit@{l}
+         | 0 => Unit
          | S n => (forall (g : forall a, C (f a)),
                      ExtensionAlong@{i j k l l} f C g) *
                   forall (h k : forall b, C b),
@@ -326,7 +326,7 @@ Section Extensions.
   Definition ooExtendableAlong@{i j k l}
              {A : Type@{i}} {B : Type@{j}}
              (f : A -> B) (C : B -> Type@{k}) : Type@{l}
-    := forall n, ExtendableAlong@{i j k l} n f C.
+    := forall n : nat, ExtendableAlong@{i j k l} n f C.
   (** Universe parameters are the same as for [ExtendableAlong]. *)
   (** We would like to say [Check], but because of bug #4517, https://coq.inria.fr/bugs/show_bug.cgi?id=4517, we can't. *)
   Definition check_ooExtendableAlong@{a b c r} : True@{Set}.
--- a/theories/Idempotents.v
+++ b/theories/Idempotents.v
@@ -5,6 +5,7 @@ Require Import hit.Truncations.
 
 Local Open Scope path_scope.
 
+Local Set Universe Minimization ToSet.
 
 (** * Idempotents and their splittings *)
 
@@ -428,7 +429,7 @@ Section Splitting.
 
   (** The splitting will be the sequential limit of the sequence [... -> X -> X -> X]. *)
   Definition split_idem : Type
-    := { a : nat -> X & forall n, f (a n.+1) = a n }.
+    := { a : nat -> X & forall n : nat, f (a n.+1) = a n }.
 
   Definition split_idem_pr1 : split_idem -> (nat -> X)
     := pr1.
@@ -457,7 +458,7 @@ Section Splitting.
 
   Definition path_split_idem {a a' : split_idem}
     (p : a.1 == a'.1)
-    (q : forall n, a.2 n @ p n = ap f (p n.+1) @ a'.2 n)
+    (q : forall n : nat, a.2 n @ p n = ap f (p n.+1) @ a'.2 n)
   : a = a'.
   Proof.
     simple refine (path_sigma' _ _ _).
--- a/theories/Idempotents.v
+++ b/theories/Idempotents.v
@@ -429,7 +429,7 @@ Section Splitting.
 
   (** The splitting will be the sequential limit of the sequence [... -> X -> X -> X]. *)
   Definition split_idem : Type
-    := { a : nat -> X & forall n : nat, f (a n.+1) = a n }.
+    := { a : nat -> X & forall n, f (a n.+1) = a n }.
 
   Definition split_idem_pr1 : split_idem -> (nat -> X)
     := pr1.
@@ -458,7 +458,7 @@ Section Splitting.
 
   Definition path_split_idem {a a' : split_idem}
     (p : a.1 == a'.1)
-    (q : forall n : nat, a.2 n @ p n = ap f (p n.+1) @ a'.2 n)
+    (q : forall n, a.2 n @ p n = ap f (p n.+1) @ a'.2 n)
   : a = a'.
   Proof.
     simple refine (path_sigma' _ _ _).
--- a/theories/Modalities/Accessible.v
+++ b/theories/Modalities/Accessible.v
@@ -125,7 +125,7 @@ Defined.
 
 Definition ooextendable_isnull_fibers {A B} (f : A -> B) (C : B -> Type)
 : (forall b, ooExtendableAlong (@const (hfiber f b) Unit tt)
-                               (fun _ : Unit => C b))
+                               (fun _ => C b))
   -> ooExtendableAlong f C
 := fun null n => extendable_isnull_fibers n f C null.
 
--- a/theories/Modalities/Accessible.v
+++ b/theories/Modalities/Accessible.v
@@ -96,7 +96,7 @@ Record NullGenerators :=
 Coercion ngen_type : NullGenerators >-> Funclass.
 
 Definition null_to_local_generators : NullGenerators@{a1} -> LocalGenerators@{a2}
-  := fun S => Build_LocalGenerators (ngen_indices S) (ngen_type S) (fun _ => Unit@{a2}) (fun _ _ => tt).
+  := fun S => Build_LocalGenerators (ngen_indices S) (ngen_type S) (fun _ => Unit) (fun _ _ => tt).
 
 (** As with [IsLocal], the real version of this notation will be defined in [Nullification]. *)
 Module Import IsNull_Internal.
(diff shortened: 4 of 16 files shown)


spitters:
LGTM, if Travis is happy.

mikeshulman:
I'm confused.  I thought the point was to put Empty and Unit in the lowest universe, i.e. Set.  So why do they start acquiring universe annotations?

mikeshulman:
Also I'm not real happy with hacky fixes when we don't understand what's going on.  Let's try to figure out what the problems are everywhere; if they are Coq bugs we should submit them, in any case we should at least understand what we're working around.

SkySkimmer:
I made a less hacky fix for the exercise error. That one might be an underlying Coq bug.

Some of the new annotations are universes which used to be minimized to Set. For instance if you do
`Definition foo : 0 = 0 := idpath. About foo.` you get something like `foo@{Top.434} : @paths@{Top.434} nat O O`.

We need to disable minimization to get that kind of lemma though.
For instance, `Types.Unit.equiv_contr_unit` without minimization gets type `forall A : Type@{i}, Contr@{i} A -> Equiv@{i Set} A Unit`. However in `UnivalenceImpliesFunext.Univalence_implies_WeakFunext` we use it to prove `Equiv@{i i} (P x) Unit` for arbitrary `i`.
So it's either disable minimization or annotate equiv_contr_unit and a bunch of other lemmas.

Then since Empty/Unit aren't annotated any more they can't be used to determine which universe should be used so we get some more extra.
For instance under `A : Type@{i}`, `(@const A Unit@{i} tt)` becomes `@const@{i i} A Unit tt` in Modality.v, because that definition requires that universes not escape. Without annotations we get `const@{Top.13362 i}` and the new universe will stay forever.
Meanwhile definitions which weren't annotated just went through producing extra universes.

JasonGross:
@mikeshulman:
> So why do they start acquiring universe annotations?

To give an much briefer answer: It's not `Unit` and `Empty` that acquire extra universe annotations, but everything which uses them in multiple places.

@SkySkimmer I'd like to see a timing diff for this change (`PREV_COMMIT=<sha1 of the version of master that you're basing this on> ./etc/coq-scripts/timing/make-pretty-timed-diff-tip.sh` and then look in `time-of-build-both.log`).  In particular, I want to see if the extra universes in things we don't annotate (e.g., `V.v`) make things slower.)

contrib/HoTTBookExercises.v
```diff
@@ -517,7 +517,9 @@ Section Book_3_14.
       apply hprop_allpath.
       intros x' y'.
       etransitivity; [ symmetry; apply (p x x y' x') | ].
-      assert (H' : idpath = path_ishprop x x) by apply path_ishprop.
+     (* Without this it somehow proves [H'] using the wrong universe for hprop_Empty and fails when we do [Defined]. *)
```
JasonGross: This sounds like a Coq bug.  Would you be willing to run (on a version of HoTT without this change, but with all the other ones) [`/path/to/find-bug.py contrib/HoTTBookExercises.v bug_failed_defined_01.v -R theories HoTT --coqc ./hoqc --coqtop ./hoqtop`](https://github.com/JasonGross/coq-tools)?

SkySkimmer:
@JasonGross 
>@SkySkimmer I'd like to see a timing diff for this change (PREV_COMMIT=<sha1 of the version of master that you're basing this on> ./etc/coq-scripts/timing/make-pretty-timed-diff-tip.sh and then look in time-of-build-both.log). In particular, I want to see if the extra universes in things we don't annotate (e.g., V.v) make things slower.)

In progress.

>This sounds like a Coq bug. Would you be willing to run (on a version of HoTT without this change, but with all the other ones) /path/to/find-bug.py contrib/HoTTBookExercises.v bug_failed_defined_01.v -R theories HoTT --coqc ./hoqc --coqtop ./hoqtop?

What is bug_failed_defined_01.v?

theories/Extensions.v
```diff
@@ -326,7 +326,7 @@ Section Extensions.
   Definition ooExtendableAlong@{i j k l}
              {A : Type@{i}} {B : Type@{j}}
              (f : A -> B) (C : B -> Type@{k}) : Type@{l}
-    := forall n, ExtendableAlong@{i j k l} n f C.
+    := forall n : nat, ExtendableAlong@{i j k l} n f C.
```
JasonGross: Why is this annotation required?  Is it because otherwise we get a fresh universe `m` and an annotation `forall n : (nat : Type@{m}), ...`?  If that's the case, it seems like a bug in the setting (@mattam82); even when not minimizing most universes to `Set`, if a universe variable is not introduced by the user, and it's not passed to any identifier (only appears as a kind of phantom type annotation on a fully resolved type), then it's fine to minimize it to `Sed`.

theories/Idempotents.v
```diff
@@ -428,7 +429,7 @@ Section Splitting.
 
   (** The splitting will be the sequential limit of the sequence [... -> X -> X -> X]. *)
   Definition split_idem : Type
-    := { a : nat -> X & forall n, f (a n.+1) = a n }.
+    := { a : nat -> X & forall n : nat, f (a n.+1) = a n }.
```
JasonGross: This is required even given the `Local Set`ting at the top?

theories/Modalities/Accessible.v
```diff
@@ -256,6 +256,7 @@ Module Accessible_Modalities_from_ReflectiveSubuniverses
     Definition inO_iff_isnull@{u a i} (O : Modality@{u a}) (X : Type@{i})
     : iff@{i i i} (In@{u a i} O X) (IsNull@{a i} (acc_gen O) X).
     Proof.
+      pose proof (@conn_map_to_O@{u a a a i a a a}).
```
JasonGross: Why do we need this?

theories/Modalities/Identity.v
```diff
@@ -84,13 +84,19 @@ Module Accessible_Identity <: Accessible_Modalities Identity_Modalities.
...
 
-  Definition inO_iff_isnull
+  Definition inO_iff_isnull@{u a i}
   : forall (O : Modality@{u a}) (X : Type@{i}),
       iff@{i i i}
         (In@{u a i} O X)
-        (IsNull@{u a i} (acc_gen O) X)
-    := fun O X => (fun _ => Empty_ind _ , fun _ => tt).
+        (IsNull@{u a i} (acc_gen@{u a} O) X).
+  Proof.
```
JasonGross: Did you say that you got an error if you gave this inline, or just that the universes were wrong?  Is there some set of universe annotations we can give (e.g., to `pair`) such that this works out nicely without the `Proof`?

JasonGross:
> What is bug_failed_defined_01.v?

The name of the new file that the minimized test case will go into, once the script finishes running.

SkySkimmer:
@JasonGross timing diff: https://gist.github.com/SkySkimmer/c989e2eb2c33f3c13945cab719fd47cd

JasonGross:
Cool, thanks!

SkySkimmer:
@JasonGross find-bug gets to this https://gist.github.com/SkySkimmer/be9fefed4a714c582d2bfd7d85c90ddc then it runs into changes in the error which I don't know how to adapt the regex to get past.

theories/Extensions.v
SkySkimmer: > Is it because otherwise we get a fresh universe m and an annotation forall n : (nat : Type@{m}), ...?
Pretty much.

theories/Idempotents.v
SkySkimmer: No it isn't.

theories/Modalities/Accessible.v
SkySkimmer: Without it it makes a fresh universe instead of the `i` (twice because conn_map_to_O gets used twice).

mikeshulman:
> It's not Unit and Empty that acquire extra universe annotations, but everything which uses them in multiple places.

That would make sense to me if it were true, but I actually see half a dozen places where Unit and Empty themselves have annotations.  That's what I don't understand.

theories/Extensions.v
mikeshulman: Yikes!

SkySkimmer:
@mikeshulman 
>That would make sense to me if it were true, but I actually see half a dozen places where Unit and Empty themselves have annotations. That's what I don't understand.

Could you specify a line? I'm not sure where you're looking.

theories/Extensions.v
```diff
@@ -97,7 +97,7 @@ Section Extensions.
            (n : nat) {A : Type@{i}} {B : Type@{j}}
            (f : A -> B) (C : B -> Type@{k}) : Type@{l}
     := match n with
-         | 0 => Unit@{l}
```
mikeshulman: Here Unit has an annotation.

theories/Modalities/Accessible.v
```diff
@@ -96,7 +96,7 @@ Record NullGenerators :=
 Coercion ngen_type : NullGenerators >-> Funclass.
 
 Definition null_to_local_generators : NullGenerators@{a1} -> LocalGenerators@{a2}
-  := fun S => Build_LocalGenerators (ngen_indices S) (ngen_type S) (fun _ => Unit@{a2}) (fun _ _ => tt).
```
mikeshulman: Here Unit has an annotation.

theories/Modalities/Identity.v
```diff
@@ -23,7 +23,7 @@ Module Identity_Modalities <: Modalities.
 
   Definition In : forall (O : Modality@{u a}),
                              Type@{i} -> Type@{i}
-    := fun O X => Unit@{i}.
```
mikeshulman: Here Unit has an annotation.  Etc.

theories/Extensions.v
SkySkimmer: That's the current (master) version of the file.

SkySkimmer:
Changed a couple things.

@mikeshulman All those are where I removed a type annotation on Unit/Empty.

JasonGross:
@SkySkimmer When I run a fresh instance of `find-bug.py` on https://gist.github.com/SkySkimmer/be9fefed4a714c582d2bfd7d85c90ddc (telling it to input that file, and output into another file), it seems to not have much trouble?

JasonGross:
@SkySkimmer reported as [bug #4862](https://coq.inria.fr/bugs/show_bug.cgi?id=4862)

theories/Modalities/Accessible.v
```diff
@@ -125,7 +125,7 @@ Defined.
 
 Definition ooextendable_isnull_fibers {A B} (f : A -> B) (C : B -> Type)
 : (forall b, ooExtendableAlong (@const (hfiber f b) Unit tt)
-                               (fun _ => C b))
+                               (fun _ : Unit => C b))
```
mikeshulman: Why is this necessary?

theories/Modalities/Notnot.v
```diff
@@ -22,43 +22,42 @@ Module Notnot_Easy_Modalities <: EasyModalities.
...
     intros f z nBz.
     pose (unNotnot O).          (** Access the [Funext] hypothesis *)
-    (** The goal is [Empty@{j}], whereas [z] has codomain [Empty@{i}].  Thus, simply applying [z] would collapse the universe parameters undesirably, so we first alter the goal to be [Empty@{i}]. *)
-    cut (Empty@{i}); [ intros [] | ].
     apply z; intros a.
-    (** Now the goal is [Empty@{i}], whereas [f] has codomain [Empty@{j}]. *)
-    cut (Empty@{j}); [ intros [] | ].
-    exact (f a (transport (fun x => not@{j j j} (B x))
+    pose proof (hprop_Empty@{i}).
+    exact (f a (transport (fun x => not@{j j} (B x))
```
mikeshulman: It's nice that we can avoid this hack now.

mikeshulman:
`*`crawls into a hole`*` My brain must *really* have been turned off...

theories/Modalities/Identity.v
```diff
@@ -84,13 +84,15 @@ Module Accessible_Identity <: Accessible_Modalities Identity_Modalities.
...
 
-  Definition inO_iff_isnull
+  Definition inO_iff_isnull@{u a i}
   : forall (O : Modality@{u a}) (X : Type@{i}),
       iff@{i i i}
         (In@{u a i} O X)
-        (IsNull@{u a i} (acc_gen O) X)
-    := fun O X => (fun _ => Empty_ind _ , fun _ => tt).
+        (IsNull@{u a i} (acc_gen@{u a} O) X)
+  := fun O X => @pair _ (_ -> Unit)
```
mikeshulman: What's going on here?

theories/Modalities/Open.v
```diff
@@ -135,6 +135,8 @@ Module Accessible_OpenModalities <: Accessible_Modalities OpenModalities.
       refine (isequiv_ooextendable (fun _ => X) (@const (unOp O) Unit tt) ext).
   Defined.
 
+  Definition inO_iff_isnull@{u a i} := inO_iff_isnull'@{u a i i i i i i i i i i i i i i}.
+
```
mikeshulman: It would be better if this hack could be avoided, but if not, then I'd like to at least have a comment in the file about what's going on.

mikeshulman:
Ok, I'd like to know what's going on in `Idempotents` still, but aside from that I'm pretty happy.  We should include some remarks in `STYLE` about the sort of hacks that are required, like including extra type annotations `: nat` (unless that can be fixed) and `pose proof`ing lemmas so that typeclass inference doesn't find them more than once (it would be nice if typeclass inference could be made smarter about that too).

theories/Modalities/Accessible.v
SkySkimmer: Because it isn't.

theories/Modalities/Identity.v
SkySkimmer: The new annotation on `acc_gen` isn't necessary, but it produces an unbound universe without the `(_ -> Unit)` bit.

theories/Modalities/Open.v
SkySkimmer: I reduced it some and made a comment, but couldn't avoid it completely.

contrib/HoTTBookExercises.v
JasonGross: @SkySkimmer , could you add a reference to [bug #4862](https://coq.inria.fr/bugs/show_bug.cgi?id=4862) here?

theories/Types/Unit.v
```diff
@@ -66,12 +66,12 @@ Global Instance isequiv_unit_rec `{Funext} (A : Type)
 (* For various reasons, it is typically more convenient to define functions out of the unit as constant maps, rather than [Unit_ind]. *)
 Notation unit_name x := (fun (_ : Unit) => x).
 
-Global Instance isequiv_unit_name `{Funext} (A : Type)
-: IsEquiv (fun (a:A) => unit_name a).
+Global Instance isequiv_unit_name `{Funext} (A : Type@{i})
```
JasonGross: I'm surprised you don't need `isequiv_unit_name@{i j}` here.  It's probably good practice to insert the `@{i j}` anyway.

JasonGross:
Modulo adding a reference to the bug I mentioned in the comment, adding `@{i j}` to the instance, and adding a note to `STYLE` on adding explicit quantification of `nat` and `unit`, LGTM.

JasonGross:
I reported the phantom binder issue as [bug #4868](https://coq.inria.fr/bugs/show_bug.cgi?id=4868).  I will add a mention of it to the style guide.

@mikeshulman after inspecting a few definitions, I think what's going on in modalities is that previously, there were a bunch of places where we had `Unit@{i}` and `Unit@{j}` in the same definition, and since `i` and `j` were not explicitly given by the user, Coq collapsed both to `Type1`.  Now, when it creates universes for `Unit` to live in, the universes are constrained only by `Set <= i` and `Set <= j`, and Coq does not identify `i` and `j`; see [bug #4870](https://coq.inria.fr/bugs/show_bug.cgi?id=4870).

I'm going to merge this, based on @spitters' and my acks.  Thanks @SkySkimmer !

PR #818, opened 2016-06-30
Put more types in Type0

SkySkimmer:
Time diff: https://gist.github.com/SkySkimmer/55d3835550db6722137af3940a181bd7
This time everything just went through.

### DIFF
--- a/theories/Classes/interfaces/canonical_names.v
+++ b/theories/Classes/interfaces/canonical_names.v
@@ -25,9 +25,6 @@ apply Trunc_ind.
 - exact E.
 Qed.
 
-Monomorphic Universe Ubool.
-Definition bool := (HoTT.Types.Bool.Bool@{Ubool}).
-
 Notation " g ∘ f " := (Compose g f)%mc
   (at level 40, left associativity).
 Notation "(∘)" := Compose (only parsing) : mc_scope.
@@ -337,19 +334,19 @@ Notation "(?=)" := compare (only parsing) : mc_scope.
 Notation "( x ?=)" := (compare x) (only parsing) : mc_scope.
 Notation "(?= y )" := (fun x => x ?= y) (only parsing) : mc_scope.
 
-Class Eqb A := eqb : A -> A -> bool.
+Class Eqb A := eqb : A -> A -> Bool.
 Infix "=?" := eqb (at level 70, no associativity) : mc_scope.
 Notation "(=?)" := eqb (only parsing) : mc_scope.
 Notation "( x =?)" := (eqb x) (only parsing) : mc_scope.
 Notation "(=? y )" := (fun x => x =? y) (only parsing) : mc_scope.
 
-Class Ltb A := ltb : A -> A -> bool.
+Class Ltb A := ltb : A -> A -> Bool.
 Infix "<?" := ltb (at level 70, no associativity) : mc_scope.
 Notation "(<?)" := ltb (only parsing) : mc_scope.
 Notation "( x <?)" := (ltb x) (only parsing) : mc_scope.
 Notation "(<? y )" := (fun x => x <? y) (only parsing) : mc_scope.
 
-Class Leb A := leb : A -> A -> bool.
+Class Leb A := leb : A -> A -> Bool.
 Infix "<=?" := leb (at level 70, no associativity) : mc_scope.
 Notation "(<=?)" := leb (only parsing) : mc_scope.
 Notation "( x <=?)" := (leb x) (only parsing) : mc_scope.
--- a/theories/Classes/tactics/ring_pol.v
+++ b/theories/Classes/tactics/ring_pol.v
@@ -6,6 +6,7 @@ Require Import
   HoTT.Classes.theory.rings.
 
 Import Quoting.
+Local Set Universe Minimization ToSet.
 
 Section content.
 Local Existing Instance almost_ring_semiring.
@@ -32,7 +33,7 @@ Context `{Trichotomy@{Set Set Set} V Vlt}.
      + forall w in P, w <= v
      + forall w in Q, w <  v *)
 
-Fixpoint Peqb P Q : bool :=
+Fixpoint Peqb P Q : Bool :=
   match P, Q with
   | Pconst c, Pconst d => c =? d
   | PX P1 v P2, PX Q1 w Q2 =>
@@ -59,7 +60,7 @@ Fixpoint eval (vs : Vars V) (P : Pol) : R :=
     (eval vs P) * (vs v) + (eval vs Q)
   end.
 
-Lemma andb_true@{} : forall a b : bool, andb a b = true -> a = true /\ b = true.
+Lemma andb_true : forall a b : Bool, andb a b = true -> a = true /\ b = true.
 Proof.
 intros [|] [|];simpl;auto.
 Qed.
--- a/theories/Classes/theory/additional_operations.v
+++ b/theories/Classes/theory/additional_operations.v
@@ -6,7 +6,7 @@ Instance decide_eqb `{DecidablePaths A} : Eqb A
   := fun a b => if decide_rel paths a b then true else false.
 
 Lemma decide_eqb_ok@{i} {A:Type@{i} } `{DecidablePaths A} :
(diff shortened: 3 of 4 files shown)


spitters:
LGTM, but I don't understand why things get slower.

JasonGross:
@ppedrot @mattam82 are universes (universe hashconsing? something else?) the reason that `Bool.v` got 15%-20% slower when we turned off `Universe Minimization ToSet`?  (and `IncoherentIdempotent` 50% slower?)  I'm curious how many more universes we use now in each of the top three files:
```
3m54.23s | Total                                                     | 3m48.72s || +0m05.51s
0m25.18s | Spaces/BAut/Bool                                          | 0m21.49s || +0m03.69s
0m06.02s | Spaces/BAut/Bool/IncoherentIdempotent                     | 0m03.96s || +0m02.05s
0m12.16s | hit/V                                                     | 0m10.88s || +0m01.27s
```

@spitters, the reason is that unsetting minimization to set can cause an exponential blow-up in universe variables; every variable that was previously minimized to set now gets its own variable.

I'm not sure it's worth merging the changes to `Bool` given the timing diff.  @SkySkimmer if you want to create a separate PR for everything here except `Bool`, and run a timing diff on that, I'd be happy to have that merged while we wait for feedback from @ppedrot and @mattam82 on the cause of slowdown here.

SkySkimmer:
Checking the number of universes generated with `Check Type.` at the end of the file:
Spaces/Baut/Bool:
before 98870
after 123283 (+25%)

Spaces/BAut/Bool/IncoherentIdempotent:
before 17068
after 26030 (+50%)

SkySkimmer:
Changing Type1 to Type1le for Bool doesn't slow things down (it even speeds up a bit).
See https://gist.github.com/SkySkimmer/108259061efe3b718388120b26ea82f8
This one is changing Type1 to Type@{i}: https://gist.github.com/SkySkimmer/a46a4359ac80882e17479cbbbf2a23b4
(both compared to current master 938f60873ddec8b35ee9b8ffdcdfdcc5fa35d293)

SkySkimmer:
@JasonGross what do you think of this change?

JasonGross:
You should add a comment explaining `Type1le`, much as the other `Type*`s are explained.  Given that, LGTM.

mikeshulman:
Wait, what?  Defining an alias that looks exactly like `Type1` produces fewer universes?

SkySkimmer:
Yes.
Why? Who knows.

mattam82:
Writing Type makes the i rigid, writing Typele makes it flexible, so it can unify with other universes. If we made Type generally flexible would mean you get less general typings than what you were used to in 8.4.

mattam82:
The i in Type@{i} that is

mikeshulman:
Okay.  Can we call this something other that `Type1le`?  That name makes it look like a version of `Type1` when actually it is a version of `Type`.  The universe `Type2` has two universe parameters `0 < i < j`, while the universe `Type2le` has two universe parameters `0 < i <= j`, so that pattern would imply that `Type1le` would mean the same thing as `Type1`, namely one universe parameter with `0 < i`.  How about calling this `Type0`?

SkySkimmer:
Sure.
- should Empty, Unit, etc use the new Type0 or stay with Set?
- in Tests.v there are a few lines like `Fail Check Type0 : Type0.` which now don't Fail. What should happend there?

mikeshulman:
Hmm, what do other people think?  Should we keep `Type0` as meaning literally `Set`, or should we make it conform to the meaning of `Type1` and `Type2` (i.e. `TypeN` means any universe with n universes strictly below it)?

JasonGross:
I think `Type0` should mean `Set`, i.e., something with 0 universe variables associated with it.  Given that we've `Unset Universe Minimization ToSet`, we can probably say that `TypeN` should mean a `Type` with `N` universe variables associated with it, in the chain `0 < u0 < u1 < ... < uN`, and `TypeNle` should mean a `Type` with `N` universe variables associated with it, in the chain `0 <= u0 <= u1 <= ... <= uN`.  What do you think of this schema?

spitters:
That sounds sensible. We should probably try it out to see how it works in
practice.

mikeshulman:
What would you call `0 < u0 <= u1 <= ... <= uN` then?  (That's what `TypeNle` currently means.)

JasonGross:
@mikeshulman I'd call it unnecessary?

mikeshulman:
Why?

JasonGross:
The reason we had `0 < ` in the first place was to prevent things from being collapsed to `Set`, which would cause problems later.  With `Unset Universe Minimization ToSet`, this is no longer necessary, so I don't see any design reason to use it.

mikeshulman:
Ok, well then someone would have to fix the modalities code to work with that definition.  It might be easy, but it doesn't work out of the box, and I don't have time to fix it myself.

SkySkimmer:
I was going to look at this again so I ran the timing and got https://gist.github.com/SkySkimmer/7d0da71155be7f1dfaf3a50e7dc64fd0
```

5m00.32s | Total                                                     | 5m33.52s || -0m33.20s
0m42.07s | Spaces/BAut/Bool                                          | 0m50.32s || -0m08.25s
0m21.63s | Categories/Category/Sigma/Univalent                       | 0m24.02s || -0m02.39s
0m11.69s | Algebra/ooGroup                                           | 0m14.13s || -0m02.44s
0m08.02s | Spaces/Finite                                             | 0m10.33s || -0m02.31s
0m08.02s | Spaces/BAut/Rigid                                         | 0m09.30s || -0m01.28s
0m07.28s | Spaces/BAut                                               | 0m09.05s || -0m01.77s
0m06.60s | HIT/Colimits/Flattening                                   | 0m08.06s || -0m01.46s
0m05.72s | Spaces/No                                                 | 0m06.84s || -0m01.12s
0m11.86s | Comodalities/CoreflectiveSubuniverse | 0m12.16s || -0m00.30s
(< 1s absolute changes)
```
Can someone else try it to check if I made a mistake? Or maybe travis does it automatically?

mikeshulman:
I get

```
4m34.91s | Total                                                     | 4m33.06s || +0m01.84s
0m37.21s | Spaces/BAut/Bool                                          | 0m35.41s || +0m01.80s
0m18.88s | Categories/Category/Sigma/Univalent                       | 0m19.05s || -0m00.17s
0m11.54s | Algebra/ooGroup                                           | 0m11.67s || -0m00.13s
0m11.00s | Comodalities/CoreflectiveSubuniverse                      | 0m10.56s || +0m00.43s
0m09.17s | contrib/Freudenthal                                       | 0m08.71s || +0m00.45s
0m08.76s | Modalities/ReflectiveSubuniverse                          | 0m08.61s || +0m00.15s
0m07.60s | Spaces/Finite                                             | 0m07.61s || -0m00.01s
0m07.57s | Pointed                                                   | 0m08.04s || -0m00.46s
```

spitters:
What's the status of this, and how does it relate to Cumulative inductive types.

https://github.com/HoTT/HoTT/issues/886

SkySkimmer:
I feel like with cumulative types it should be possible to put minimization to Set back on (for probable performance gains) while keeping bool and company in Set. I haven't tried though.
Also I expect we shouldn't need to prove every lemma on `nat` at arbitrary levels (https://github.com/HoTT/HoTT/blob/master/theories/Classes/implementations/peano_naturals.v#L13) (IIRC it was needed to make some of the lemmas in https://github.com/HoTT/HoTT/blob/master/theories/Classes/theory/naturals.v go through).

spitters:
It would be great if you could try wether this actually works.

from my phone

spitters:
Is my impression correct that we should pull this 1 char change?

SkySkimmer:
You should rerun travis just in case something relevant changed in the last 9 months.

spitters:
File "./theories/Classes/interfaces/canonical_names.v", line 29, characters 20-40:
Error: Universe instance should have length 0

SkySkimmer:
Right, hottclasses has a monomorphised bool.
Should be fixed.

spitters:
OK. Pulling this one. Travis is hanging on grphs.

PR #820, opened 2016-07-13
Updating HoTTBook.v (through section 6.12)

mikeshulman:
`HoTTBook.v` was way behind in coding up what has actually been implemented.  I started updating it some, but ran out of steam for now towards the end of chapter 6.

### DIFF
--- a/theories/Types/Universe.v
+++ b/theories/Types/Universe.v
@@ -2,7 +2,7 @@
 (** * Theorems about the universe, including the Univalence Axiom. *)
 
 Require Import HoTT.Basics.
-Require Import Types.Sigma Types.Forall Types.Arrow Types.Paths Types.Equiv.
+Require Import Types.Sigma Types.Forall Types.Arrow Types.Paths Types.Equiv Types.Bool.
 
 Local Open Scope path_scope.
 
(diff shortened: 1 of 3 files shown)


spitters:
Looks like @JasonGross name change broke this pull request.

mikeshulman:
bah

JasonGross:
You should be able to fix it with
```bash
sed -i 's|hit\.|HIT.|g' contrib/HoTTBook.v
sed -i 's|Require HoTT.categories.|Require HoTT.Categories.|g' contrib/HoTTBook.v
```
and merging/rebasing.

mikeshulman:
Thanks; done.

mikeshulman:
bump

theories/Types/Universe.v
```diff
@@ -466,4 +466,15 @@ Proof.
   refine (trunc_equiv _ path_universe_uncurried).
 Defined.
 
+(** We can also say easily that the universe is not a set. *)
+Definition not_hset_Type `{Univalence} : ~ (IsHSet Type).
```
JasonGross: Can this be generalized to "If a universe contains an `n`-Type, then that universe is not an `n`-Type?"  (Is that true?)

JasonGross:
I'm going to trust that you lined up the theorems to the book correctly, and say LGTM

theories/Types/Universe.v
mikeshulman: Not easily, certainly.  As discussed on the hott mailing list a while back, there do exist n-types X such that BAut(X) is a k-type for k<n+1.  In particular, there are 1-types X (of the form BG for suitably chosen groups G) such that BAut(X) is contractible.  That doesn't necessarily mean the conjecture is false, but any proof of it would have to use some closure properties of a universe in addition to the fact that it contains an n-type and is univalent.

PR #822, opened 2016-07-13
theories/{hit => HIT,categories => Categories}

### LINKED ISSUE #612: Rename `hit` and `categories` directories
We seem to be trending to using capitals for directories (this should probably be recorded in the style file), so we should probably rename `hit` to `HIT` and `categories` to `Categories`.

JasonGross:
This closes #612

The library compiles like this.  Nearly all changes were made by the sequence:
```bash
git mv theories/{hit,HIT}
git mv theories/{categories,Categories}
git mv theories/{categories,Categories}.v
git grep --name-only '/hit' | xargs sed -i 's|/hit|/HIT|g'
git grep --name-only 'hit/' | xargs sed -i 's|hit/|HIT/|g'
git grep --name-only 'hit\.' | xargs sed -i 's|hit\.|HIT.|g'
git grep --name-only '/categories\|categories/' | xargs sed -i s'|/categories|/Categories|g'
git grep --name-only '/categories\|categories/' | xargs sed -i s'|categories/|Categories/|g'
git grep --name-only '\.categories\.' | xargs sed s'|\.categories\.|.Categories.|g' -i
```


spitters:
Indeed, looks like search and replace. LGTM

PR #824, opened 2016-08-02
Fix typo in Build_PreCategory wrapper

SkySkimmer:


mikeshulman:
Does the fact that the library compiled before this fix mean that the wrapper is never actually used?

JasonGross:
No; the new wrapper is alpha-convertible with the old wrapper, it's just more intuitively named (i.e., it's no longer misleadingly named).  Both are convertible with
```coq
Definition Build_PreCategory
            a b c d associativity left_identity right_identity
    := @Build_PreCategory'
         a b c d
         associativity
         (fun _ _ _ _ _ _ _ => symmetry _ _ (associativity _ _ _ _ _ _ _))
         left_identity
         right_identity
         (fun _ => left_identity _ _ _).
```

mikeshulman:
oh, duh.

PR #825, opened 2016-08-13
equiv_functor_forall_covariant and its composition.

mikeshulman:
This is a port of `equiv-Π'` from HoTT-Agda, and a composition lemma about it that isn't in HoTT-Agda yet.  The composition lemma may be the first place where it matters that we defined isequiv_compose without adjointifying, so that both homotopies remain the explicitly given ones.

### DIFF
--- a/theories/Types/Forall.v
+++ b/theories/Types/Forall.v
@@ -7,7 +7,7 @@ Require Import Types.Paths.
 Local Open Scope path_scope.
 
 
-Generalizable Variables A B f g e n.
+Generalizable Variables A B C f g e n.
 
 Section AssumeFunext.
 Context `{Funext}.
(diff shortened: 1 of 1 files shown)


mikeshulman:
Specifically, if I modify the definition of `isequiv_compose` by inserting an unnecessary `idpath` in the first homotopy and a corresponding `concat_p1` in the triangle law, then before this commit the whole library still compiles, but the composition lemma does not.

mikeshulman:
(I discovered this by trying to prove the composition lemma in HoTT-Agda, banging my head against a wall for a while, and finally realizing that their version of `isequiv_compose` *was* proved by adjointifying, so that the homotopies were not what I expected them to be.)

JasonGross:
LGTM

PR #826, opened 2016-08-15
Bump coq-HoTT, to pull in a few bugfixes.

peterlefanulumsdaine:
Coq 8.5 has had a few bugfixes since we last bumped the submodule — in particular, https://github.com/coq/coq/pull/228, which fixes a build breakage over recent versions of OCaml (discussed further [here](https://github.com/ocaml/opam-repository/issues/6849)).


JasonGross:
I think we should track only released versions of Coq, unless we have good reason not to.  I don't think we should bump the submissive again until either 8.5pl3 or 8.6

peterlefanulumsdaine:
@JasonGross  We are still at 8.5pl1; would bumping to 8.5pl2 be acceptable, or was there a reason for skipping it that I’ve missed?

(8.5pl2 suffices for the specific bugfix I wanted.  That bug is that on 8.5pl1, the build is broken with recent versions of OCaml; double-linking of modules was previously just a warning, but is an error in OCaml ≥4.0.3, so a double-linked module that already existed in Coq now causes a breakage.)

I’m opening a PR for 8.5pl2 now, just to get Travis checking the build.  Is it better to continue this discussion here, or move it there?

peterlefanulumsdaine:
OK, I’m closing this PR now, as superseded by #827 .

JasonGross:
Ah, I hadn't realized we were still on pl1.  Bumping to pl2 seems good.  Better if we can stay compatible with both pl1 and pl2.  (Let's nice further discussion there.)

PR #827, opened 2016-08-15
Bumped coq-HoTT submodule to 8.5pl2, pulling in a few bugfixes.

peterlefanulumsdaine:
Replacement for https://github.com/HoTT/HoTT/pull/826; see discussion there.


peterlefanulumsdaine:
(Sorry for noise of closing/reopening; just hit the wrong button there.)

JasonGross:
Do we want to continue targeting a single pl, or would it be reasonable to aim for compatibility with both pl1 and pl2?  (I can make Travis check both versions in addition to the submodule, if desired.)

In any case, assuming this is the 8.5pl2 tag and assuming that it builds, LGTM.

mikeshulman:
If it's easy to make Travis check both versions, then we might as well.  That way we'll at least *know* if and when we break pl1 compatibility, and we can decide then whether it's worth trying to maintain.

JasonGross:
Should be pretty easy.  We do it [here](https://github.com/mit-plv/fiat-crypto/blob/master/.travis.yml), for example.

peterlefanulumsdaine:
Agreed with Mike — maintaining dual compatibility seems worth doing at least as long as it remains easy.

PR #829, opened 2016-08-29
Add simple_rapply and use tactics

SimonBoulier:
Hi,
Here is just a proposal.
As I find `simple refine` much more useful than `refine`, I propose to add a `simple rapply` tactic which use it (we could also change rapply directly).
Then we need `simple erapply` to allow holes in terms, I propose to call it `use` (seen in some code of Jason, I can't remember where) as it is very convenient to have a short name for a so useful tactic.
Feel free to comment.


mikeshulman:
Can you give an example where you would prefer the `simple` version?

(Would you want to use a Tactic Notation to make it `simple rapply` instead of `simple_rapply`?)

SimonBoulier:
On 29/08/2016 20:43, Mike Shulman wrote:
>
> Can you give an example where you would prefer the |simple| version?
>
A typical example is `use isequiv_adjointify.`. I want than Coq asks me
the inverse function before proving the section and retraction.

> (Would you want to use a Tactic Notation to make it |simple rapply|
> instead of |simple_rapply|?)
>
It doesn't matter because I only use `use`.

JasonGross:
I think it would be better to change all of the `open_constr` to `uconstr` (this will better match the behavior of `eapply`), and rename `use` to `simple erapply`.

JasonGross:
I'm not a big fan of short names, but will defer to others on the point of naming if there are strong preferences.

mikeshulman:
I would also prefer `simple erapply`.

Actually my initial reaction is that I would want this `simple` case to be the default behavior of `refine` and `rapply`.  In what situation do you want the "non-simple" version?  My general experience with "shelved" goals is that they cause problems.

JasonGross:
If you want to write code that is compatible with both Coq 8.4 and Coq 8.5, then either the default of `refine` must be to shelve dependent subgoals, or you must give up on using `refine` entirely.  Many of Coq's design decisions about defaults are for backwards compatibility, which is useful for very large projects.

More philosophically, when you use `refine` with simple types, where all of the dependently typed subgoals have types like `nat`, you want them shelved.  (Similarly if you do something like `refine (_ @ _)` (i.e., `etransitivity`), you don't want to be given a goal to construct the intermediate term, generally; you instead want to construct the intermediate by giving the paths.)

So here we have the question: do we reverse the conventions of standard Coq, or do we pick longer names for more common tactics?

mikeshulman:
what about something like `srefine` and `srapply` -- reasonably short, but still indicates that it's a variant of `refine` (unlike `use`)?

JasonGross:
Fine by me.  So then we'd have
```coq
Tactic Notation "srefine" uconstr(term) := simple refine term.
Ltac srapply p := ...
Tactic Notation "serapply" uconstr(term) := srapply term.
```

SimonBoulier:
It's fine to me.

Another naive question: do we really need `erapply` and `serapply`?
Couldn't we just overload the names `rapply` and `srapply` with a tactic notation?

Moreover, as far as I understand:
  - in `eapply`, `eauto`, `econstructor`, ... the 'e' means than the tactic can leave some metavariables
  - whereas in `erapply` it only means that it won't miserably fail with a term whose first arguments are implicits.

mikeshulman:
> Couldn't we just overload the names rapply and srapply with a tactic notation?

If I understand correctly, the `Tactic Notation` actually behaves differently because it specifies a particular way of interpreting the argument (as a `uconstr`, whatever that is), which for some reason an `Ltac` is unable to do.  But maybe @JasonGross  can correct me, and also address the issue of the meaning of `e`?

JasonGross:
There's a naming question here: does `rapply` have a hidden `e`?  `e` stands for "leaves evars".  `rapply` leaves evars.  There's nothing like `apply` which uses `refine`; there are only things like `eapply` that use `refine`, and there are things kind-of-vaguely-like an `eapply` that's trying to be `apply`.  I'd say that `srapply` and `rapply` aren't terribly useful, and the nice variants are the ones that use `uconstr`.

A `uconstr` is a raw syntax tree (untyped constr); implicits are not yet resolved.  The default tactic interpretation scope is `constr`; to get this without tactic notation, you'd need to write `rapply uconstr:(term)`.

SimonBoulier:
It's ok for me.

JasonGross:
LGTM

PR #831, opened 2016-10-10
V8.6

mattam82:
This PR shows the changes necessary to compile with the current 8.6 branch, soon to be beta1, once we resolve the following issues:
- symmetry shadowing (I patched by using apply symmetry everywhere)
- Arguments foo /. fails currently, but I think we'll backtrack on making this an error
- One real bug in typeclasses.

In the end only the next-to-last patch should apply, which is about universe annotations consistent with the fixed minimization which is now insensitive to redundant constraints in the graph (see https://github.com/coq/coq/pull/315) for the patched Coq that goes with this.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -22,6 +22,9 @@ Global Unset Strict Universe Declaration.
 (** This command makes it so that when we say something like [IsHSet nat] we get [IsHSet@{i} nat] instead of [IsHSet@{Set} nat]. *)
 Global Unset Universe Minimization ToSet.
 
+(** Change in introduction patterns not adding an implicit [] *)
+Global Unset Bracketing Last Introduction Pattern.
+
 Definition relation (A : Type) := A -> A -> Type.
 
 Class Reflexive {A} (R : relation A) :=
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -508,6 +508,8 @@ Fixpoint IsTrunc_internal (n : trunc_index) (A : Type) : Type :=
     | n'.+1 => forall (x y : A), IsTrunc_internal n' (x = y)
   end.
 
+Existing Class IsTrunc_internal.
+
 Arguments IsTrunc_internal n A : simpl nomatch.
 
 Class IsTrunc (n : trunc_index) (A : Type) : Type :=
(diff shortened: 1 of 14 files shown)


mattam82:
Hmm, I realize the first two patches are not clean, I'll have to rebase those before merging too.

mattam82:
Apart from the symmetry weirdness (which I avoided in my patch using apply symmetry, (@ppedrot any idea?), the Arguments and the typeclass bug fixes should be in 8.6 by tomorrow.

ppedrot:
The symmetry issue should have been fixed by 9bdd8aff. If not, then we should do something about it.

mattam82:
Rebased for latest 8.6 and cleaned-up the commits for things that have been fixed now.
Remains: ltac:(), IsTrunc_internal has to be declared a class if you want to do resolution on it (actually made the proof search faster by indexing the hints related to IsTrunc(_internal)) and the new universe annotations due to the fixed minimization.

mattam82:
There is also one bugfix due to the fixed keyed unification strategy (a simpl has to be done instead of an useless hnf in the script).

JasonGross:
Thanks!

I don't suppose we can get code that builds with both 8.5 and 8.6?

mattam82:
You have to try, I think the ltac:() is retrocompatible, maybe we just have
to do a bit of massaging for the universes (be more explicit).

JasonGross:
Can you split out the retro-compatible changes into a separate PR?  I'm definitely happy to merge the parenthetical changes now, and, if hnf -> simpl is because of fixed point refolding, I'm happy to merge that too.  I'm weakly in favor of waiting until 8.6 stable is released before we make a tag for "last 8.5-compatible version" and make HoTT incompatible with 8.5.

theories/Basics/Overture.v
```diff
@@ -505,6 +508,8 @@ Fixpoint IsTrunc_internal (n : trunc_index) (A : Type) : Type :=
     | n'.+1 => forall (x y : A), IsTrunc_internal n' (x = y)
   end.
 
+Existing Class IsTrunc_internal.
```
JasonGross: You do this twice.  Is that necessary?

JasonGross:
Closed by #863

PR #837, opened 2016-11-11
Symmetry for Coeq

spiceghello:
Proof that Coeq f g <~> Coeq g f.

### DIFF
--- a/theories/HIT/Coeq.v
+++ b/theories/HIT/Coeq.v
@@ -340,3 +340,21 @@ Section CoeqInd2.
   Qed.
 
 End CoeqInd2.
+
+(** ** Symmetry *)
+
+Definition Coeq_sym_map {B A} (f g : B -> A) : Coeq f g -> Coeq g f :=
+  Coeq_rec (Coeq g f) coeq (fun b : B => (cp b)^).
+
+Lemma sect_Coeq_sym_map {B A} {f g : B -> A} : Sect (Coeq_sym_map g f) (Coeq_sym_map f g).
+Proof.
+  unfold Sect. srapply @Coeq_ind.
+  - simpl. auto.
+  - intro b. simpl.
+    rewrite transport_paths_FFlr, Coeq_rec_beta_cp, ap_V, Coeq_rec_beta_cp; hott_simpl.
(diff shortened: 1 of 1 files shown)


theories/HIT/Coeq.v
```diff
@@ -340,3 +340,21 @@ Section CoeqInd2.
...
+
+Definition Coeq_sym_map {B A} (f g : B -> A) : Coeq f g -> Coeq g f :=
+  Coeq_rec (Coeq g f) coeq (fun b : B => (cp b)^).
+
+Lemma sect_Coeq_sym_map {B A} {f g : B -> A} : Sect (Coeq_sym_map g f) (Coeq_sym_map f g).
+Proof.
+  unfold Sect. srapply @Coeq_ind.
+  - simpl. auto.
+  - intro b. simpl.
+    rewrite transport_paths_FFlr, Coeq_rec_beta_cp, ap_V, Coeq_rec_beta_cp; hott_simpl.
```
JasonGross: This should probably either be wrapped in `abstract`, or made into an explicit path via use of `refine`.

theories/HIT/Coeq.v
```diff
@@ -340,3 +340,21 @@ Section CoeqInd2.
...
+
+(** ** Symmetry *)
+
+Definition Coeq_sym_map {B A} (f g : B -> A) : Coeq f g -> Coeq g f :=
+  Coeq_rec (Coeq g f) coeq (fun b : B => (cp b)^).
+
+Lemma sect_Coeq_sym_map {B A} {f g : B -> A} : Sect (Coeq_sym_map g f) (Coeq_sym_map f g).
+Proof.
+  unfold Sect. srapply @Coeq_ind.
+  - simpl. auto.
```
JasonGross: What does `auto` do here?  If it's just proving `x = x`, please use `reflexivity` instead, since this is a transparent proof.

mikeshulman:
LGTM modulo Jason's suggestions.  Thanks!

spiceghello:
Thank you for the feedback. I chose `abstract` over `refine` since I find it is more clear.

PR #838, opened 2016-11-21
Symmetry for pushout, pushouts preserve equivalences

spiceghello:
Proof that `pushout f g <~> pushout g f`, and results directly from functoriality of Coeq.
It might (?) be useful, in the spirit of forgetting that pushouts are defined via coequalizers, to have the whole Functoriality part of .Coeq redone for .Pushout, but so far this is the only result I really wanted to have without going back to Coeq.

### DIFF
--- a/theories/HIT/Pushout.v
+++ b/theories/HIT/Pushout.v
@@ -73,6 +73,49 @@ Proof.
   refine (pushout_ind_beta_pp (fun _ => P) _ _ _).
 Defined.
 
+(** ** Symmetry *)
+
+Definition pushout_sym_map {A B C f g} : pushout f g -> pushout g f :=
+  pushout_rec (pushout g f) (push o (equiv_sum_symm B C)) (fun a : A => (pp a)^).
+
+Lemma sect_pushout_sym_map {A B C f g} : Sect (@pushout_sym_map A C B g f) (@pushout_sym_map A B C f g).
+Proof.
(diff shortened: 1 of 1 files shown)


mikeshulman:
LGTM.  I would have said that the whole point of defining pushouts in terms of coequalizers was *not* to have to re-do work like this!

theories/HIT/Pushout.v
```diff
@@ -73,6 +73,49 @@ Proof.
...
+  - intros [ | ]. all:reflexivity.
+  - intro a.
+    abstract (rewrite transport_paths_FFlr, pushout_rec_beta_pp, ap_V, pushout_rec_beta_pp; hott_simpl).
+Defined.
+
+Lemma pushout_sym {A B C} {f : A -> B} {g : A -> C} : pushout f g <~> pushout g f.
+Proof.
+  exact (equiv_adjointify pushout_sym_map pushout_sym_map sect_pushout_sym_map sect_pushout_sym_map).
+Defined.
+
```
spitters: I would prefer 
```
:= (equiv_adjointify pushout_sym_map pushout_sym_map sect_pushout_sym_map sect_pushout_sym_map)
```

here. There seems to be no point in opening an interactive proof when you just give the proof term.

spitters:
LGTM modulo the minor suggestion.

PR #841, opened 2016-11-29
Basic 8.6 compatibility.

ejgallego:
- adapt to new ltac:(constr) syntax.
- `Global Unset Bracketing Last Introduction Pattern` set.
- @mattam82's tweaks for `Hint Extern` and class resolution.
- @mattam82's scoping backport in `Init.Notations`

Tested with:

```
The Coq Proof Assistant, version 8.5pl3 (November 2016)
compiled on Nov 29 2016 17:34:9 with OCaml 4.03.0
```

### DIFF
--- a/coq/theories/Init/Notations.v
+++ b/coq/theories/Init/Notations.v
@@ -76,9 +76,15 @@ Reserved Notation "{ x : A  & P }" (at level 0, x at level 99).
 Reserved Notation "{ x : A  & P  & Q }" (at level 0, x at level 99).
 
 Delimit Scope type_scope with type.
+Delimit Scope function_scope with function.
 Delimit Scope core_scope with core.
 
+(* 8.6 only :S *)
+(* Bind Scope type_scope with Sortclass. *)
+(* Bind Scope function_scope with Funclass. *)
+
 Open Scope core_scope.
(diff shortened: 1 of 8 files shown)


JasonGross:
Can you make sure the typeclasses change has no timing impact?  (`./etc/coq-scripts/timing/make-pretty-timed-diff-tip.sh`)  other than that, and modulo Travis, LGTM

ejgallego:
Tests running.

mattam:
I think you meant to reference @mattam82

ejgallego:
@JasonGross I see a 3 seconds increase overall, so that seems like noise; running again just in case.

ejgallego:
I think I ran the script correctly (file staged, but not commited), this is what I see:
```
3m08.46s | Total                                                     | 3m08.71s || -0m00.24s
```

JasonGross:
Oh, that's right, I gave you the wrong script for an already committed change.  Yeah, if that's what you see when you run it with the change staged but not committed, that's good, and I'll say LGTM modulo travis.  (The actual one I meant to suggest was `./etc/coq-scripts/timing/make-pretty-timed-diff.sh`, which uses `git checkout HEAD^` rather than `git stash`.)

Issue #839, opened 2016-11-29
8.6


spitters:
@ejgallego has ported hott to v8.6 on [jshott](https://x80.org/rhino-hott/). I am hoping he will tell us here which obstacles he met, so that we can do the same.

ejgallego:
My current tree is at https://github.com/ejgallego/HoTT/tree/compat-8.6

Note that the port it is not yet complete, but I think the diff is reasonably small. There are a couple of strange things with notations.

Wondering if we should merge this with #804 ?

mattam82:
Huh, isn't m'y PR enough? It's very small now.

mattam82:
PR #831

ejgallego:
Oh indeed I had missed it, I have too many versions on my HD :S

I did as Jason suggested and split the changes in 8.5 compatible (that are in a separate PR) and rebased @mattam82 's tree at: https://github.com/ejgallego/HoTT/tree/mz-8.6

It should be possible to fix the universe problems as highlighted by Matthieu.

mattam82:
LGTM

spitters:
It looks good to me too. However, we do not have a concrete PR now, do we?

mattam82:
I guess @ejgallego can make one now?

ejgallego:
I hesitate as this PR breaks compatibility with 8.5. As Matthieu pointed out, explicit universe annotations would be needed.

spitters:
With 8.6 out, I think we should now create a 8.5 compatibility branch and have the main branch follow 8.6. Right?

ejgallego:
IMVHO it really depends how robust you want to be vs "universe minimization" @JasonGross ?

Adding such annotations would increase compatibility and maybe remove the need for an 8.5 branch, however I ignore if the HoTT developers would like to have such annotations in the code.

JasonGross:
I agree with @spitters, though, I'm not sure we want to keep maintaining an 8.5 branch.  I'm thinking an 8.5 tag rather than a branch.  @spitters (and @andrejbauer, @mikeshulman and others), thoughts?

I'd say, yeah, @ejgallego, make a PR which bumps the submodule to 8.6 released and makes breaking changes as needed.  If you can make universe annotations, that would be nice, but don't worry about it if it's hard.  (In either case, we'd need to do some fiddling with the standard library, which I don't know how to do in an 8.5+8.6 compatible way.)

mikeshulman:
[...]

Issue #804, opened 2016-05-19
Compiling HoTT with recent coq


SkySkimmer:
I want to experiment with @mattam82's Inductive-Inductive definitions (https://github.com/mattam82/coq/tree/IR) and HoTT, but HoTT doesn't compile and immediately complains:

```
COQTOP coq/theories/Init/Notations
File "./coq/theories/Init/Notations.v", line 88, characters 0-26:
Error: File not found on loadpath : eauto.cmxs
Loadpath: ./coq/theories:.:~/HoTT/coq-HoTT/ide/utils:~/HoTT/coq-HoTT/ide/MacOS:~/HoTT/coq-HoTT/ide:/usr/share/coq:./coq/plugins/omega:./coq/plugins/rtauto:./coq/plugins/firstorder:./coq/plugins/cc:./coq/plugins/extraction:./coq/plugins/micromega:./coq/plugins/romega:./coq/plugins/quote:./coq/plugins/xml:./coq/plugins/btauto:./coq/plugins/nsatz:./coq/plugins/fourier:./coq/plugins/derive:./coq/plugins/decl_mode:./coq/plugins/syntax:./coq/plugins/funind:./coq/plugins/setoid_ring:./coq/plugins:./coq/toploop:./coq/stm:./coq/dev
Makefile:904: recipe for target 'coq/theories/Init/Notations.vo' failed
make: *** [coq/theories/Init/Notations.vo] Error 1
```

The error is the same with coq trunk. IR appears to have been branched after https://github.com/mattam82/coq/commit/a581331f26d96d1a037128ae83bebd5e6c27f665 (which might be related to the problem since it moves eauto around).

andrejbauer:
Right, it's likely that `Notations.v` is not compatible with that branch of Coq. One possibility is that you create an `IR` branch of HoTT and just do whatever is needed. The files in `coq/theories` exist only because without them Coq is unahppy. So, if everyone is happy without `eauto` you can just delete it.

spitters:
Since this also shows up in trunk. It would be good to have a more informed
solution.
I am not sure whether we use eauto, so perhap you can check this and if
not, comment it out.

SkySkimmer:
eauto appears a bunch of times in theories/categories/, a few times in contrib/ and a small number of times elsewhere:
```
contrib/Freudenthal.v
82:  { apply @trunc_sigma; try typeclasses eauto.
83:    { apply @trunc_forall; try typeclasses eauto; intro.
[...]

PR #846, opened 2016-12-16
Matthieu's fixes to enable compilation on 8.6

ejgallego:
Last set of remaining changes, closes: #839 #804 #831

Builds on my machine™


theories/Modalities/Lex.v
```diff
@@ -212,14 +212,15 @@ Module Lex_Reflective_Subuniverses
...
+    refine (contr_equiv' _ (hfiber_hfiber_compose_map@{k k i k k k k k} _ g x)).
     apply fcontr_isequiv.
     unfold hfiber_compose_map.
     transparent assert (h : (Equiv@{k k} (hfiber@{k i} (@pr1 A B) (g x))
                                          (hfiber@{k i} g (g x)))).
     { refine (_ oE equiv_to_O@{u a k k} O _).
       - refine (_ oE BuildEquiv _ _
-                  (O_functor_hfiber@{u a k i k k i k k (* <- this k is irrelevant *) k k k k}
+                  (O_functor_hfiber@{u a k i k k i k k k}
+                                   (* MS: Annotation can be dropped now *)
```
spitters: So, shall we remove the annotation then?

theories/Modalities/Nullification.v
```diff
@@ -78,7 +78,8 @@ Module Nullification_Modalities <: Modalities.
     refine (Localize_ind@{a i j k}
              (null_to_local_generators@{a a} (unNul O)) A B g _); intros i.
     apply (ooextendable_over_unit@{a i j a k}); intros c.
-    refine (ooextendable_postcompose@{a a j j k j j j k j k k}
+    refine (ooextendable_postcompose@{a a j j k k k k k k}
+                                    (* MS: Can be dropped *)
```
spitters: Same here.

spitters:
LGTM, modulo the two optional changes.
We should wait for travis to finish on v8.6 though.

ejgallego:
Ok both changes addressed.

spitters:
Thanks! OK for me. When Travis is happy.

spitters:
There is:
https://github.com/SkySkimmer/HoTTClasses/
by @SkySkimmer which depends on induction recursion, which may not have been upgraded to 8.6.
@mattam82, any insights?

ejgallego:
Dunno what's making this fail, I have no clue, the git id of the submodule seems correct.

JasonGross:
The issue:
```
Your version of OCaml is 3.12.1.
You need OCaml 4.01 or later.
Configuration script failed
```

Try adding `dist: trusty` above the `python:` line:
```
language: python
# python, for proviola

dist: trusty

python:
 - "2.7"
```

ejgallego:
Now ocamlfind is missing, this is handled in the script repos, right?

JasonGross:
@ejgallego https://github.com/HoTT/HoTT/blob/master/etc/ci/install_coq_deps.sh#L10

ejgallego:
Yeah I figured it out, thanks! trying again...

JasonGross:
*sighhhhhh* libgraphviz has the wrong version on trusty :(

JasonGross:
I'll probably look into this on Monday or something.  Getting a version of graphviz that works with Haskell and actually supports building our library dependency graphs is a tremendous pain.

ejgallego:
Don't worry @JasonGross , OCaml 4.01 in trusty indeed seems broken and cannot build Coq. I warned the developers many times, maybe some clarification is needed in the Coq documentation, @maximedenes ? 

From the log :

```
OCAMLC    ltac/g_eqdecide.ml
OCAMLOPT  ltac/g_eqdecide.ml
/tmp/camlasm882a17.s: Assembler messages:
/tmp/camlasm882a17.s:161: Error: file number less than one
/tmp/camlasm882a17.s:266: Error: file number less than one
/tmp/camlasm882a17.s:283: Error: file number less than one
/tmp/camlasm882a17.s:425: Error: file number less than one
/tmp/camlasm882a17.s:441: Error: file number less than one
File "ltac/g_eqdecide.ml4", line 1:
Error: Assembler error, input left in file /tmp/camlasm882a17.s
make[1]: *** [ltac/g_eqdecide.cmx] Error 2
make[1]: Leaving directory `/home/travis/build/HoTT/HoTT/coq-HoTT'
```

I'm afraid HoTT will have to switch to building OCaml with OPAM, or you could alternatively use the ocaml packages at my PPA:
https://launchpad.net/~e+launchpad/+archive/ubuntu/coq-85?field.series_filter=trusty

JasonGross:
> Error: file number less than one

This is easy to fix, just pass `READABLE_ML4=1` to `make`.  Or, maybe, use `camlp5` rather than `camlp4`?  I'm not sure about this one.  We cannot use opam; that takes too long to build Coq.  We could use a ppa, either yours or https://launchpad.net/~jgross-h/+archive/ubuntu/many-coq-versions

ejgallego:
Note the build is using `camlp5` already:
https://travis-ci.org/HoTT/HoTT/jobs/184685243

I may try the `READABLE_ML4=1` fix, but I don't remember it being so easy, I had to backport the patch from Debian which in turn is a backport from 4.02.0 which is in turn broken too. The minimum sane version for Coq is 4.02.3 these days, and it was not me who said that ;)

mattam82:
About IR, yes I will rebase it on 8.6 and maybe do a point release
actually, the branch still needs quite a lot of development before it's
ready for prime time.

ejgallego:
It seems the travis script do need some more work to adapt to 8.6; I won't be able to have a look at it in the immediate future thou :S

spitters:
@JasonGross @ejgallego 
To understand what the issue is. 
Do we want to use docker in travis to bump ubuntu to 16.04, or would that be too slow?

ejgallego:
@spitters I think it is possible to avoid that using Jason's trick, however there are other parts of the script that need some 8.6 work; it seems we are building the wrong Coq make target now.

JasonGross:
@ejgallego The target we need to build for Coq no longer exists, because @letouzey removed it in https://github.com/coq/coq/commit/301ec42f1b38623df8f5de2cfb69da6836bd6e01 (see [bug #5272](https://coq.inria.fr/bugs/show_bug.cgi?id=5272)).

@spitters @ejgallego I'm currently trying to get a graphviz package working on trusty; I'll update you when I've managed to switch HoTT to trusty.

JasonGross:
@ejgallego I've updated my ppa with a working version of graphviz, so we can now build on trusty.  Any thoughts on the best way to resolve the missing Coq installation target?

ejgallego:
I see three choices here:
- reinstate the old installation target upstream.
- pass a flag to HoTT configure so it uses the local coq build (this is the way I usually built HoTT as I never do system-wide installs)
- add prebuilt Coq packages to your ppa.

ejgallego:
@JasonGross I did a manual fix for this, what do you think?

The good side is that the Coq build should be faster now. Globally, I think it would make sense just to build coq with `-local`, update the path, and simplify a bit the script.

WRT Ocaml I would follow the solution used in Coq's `.travis.yml` with the difference of using the system compiler + opam for `findlib camlp5`, this should also help to simplify the dpgraph stuff.

etc/ci/install_coq.sh
```diff
@@ -39,10 +39,12 @@ then
 fi
 echo '$ ./configure '"$@"
 ./configure "$@"
-echo '$ make coqlight'
-make coqlight READABLE_ML4=1
-echo '$ sudo make install-coqlight install-devfiles'
-sudo make install-coqlight install-devfiles
+echo '$ make states READABLE_ML4=1 tools'
+make states tools READABLE_ML4=1
```
JasonGross: Why `states tools` rather than `coqlight`?

.travis.yml
```diff
@@ -20,7 +20,6 @@ env:
    - UPDATE_HTML=""       WITH_AUTORECONF=""     UPDATE_QUICK_DOC=""     UPDATE_AUTOGEN=""     UPDATE_DEP_GRAPHS=""    BUILD_COQ="yes"  UPDATE_OPAM=""       FORCE_COQ_VERSION=""
    - UPDATE_HTML=""       WITH_AUTORECONF="yes"  UPDATE_QUICK_DOC=""     UPDATE_AUTOGEN=""     UPDATE_DEP_GRAPHS=""    BUILD_COQ=""     UPDATE_OPAM=""       FORCE_COQ_VERSION=""
    - UPDATE_HTML=""       WITH_AUTORECONF="yes"  UPDATE_QUICK_DOC=""     UPDATE_AUTOGEN=""     UPDATE_DEP_GRAPHS=""    BUILD_COQ="yes"  UPDATE_OPAM=""       FORCE_COQ_VERSION=""
-   - UPDATE_HTML=""       WITH_AUTORECONF="yes"  UPDATE_QUICK_DOC=""     UPDATE_AUTOGEN=""     UPDATE_DEP_GRAPHS=""    BUILD_COQ="yes"  UPDATE_OPAM=""       FORCE_COQ_VERSION="v8.5"
```
JasonGross: Why remove this rather than just changing `v8.5` to `v8.6`?

.travis.yml
```diff
@@ -31,8 +30,8 @@ env:
 matrix:
   fast_finish: true
   allow_failures:
+    - env: UPDATE_HTML=""       WITH_AUTORECONF="yes"  UPDATE_QUICK_DOC=""     UPDATE_AUTOGEN=""     UPDATE_DEP_GRAPHS="yes" BUILD_COQ="yes"  UPDATE_OPAM=""       FORCE_COQ_VERSION=""
```
JasonGross: I don't think `UPDATE_DEP_GRAPHS` should be an allowed failure; I want to know when things break this.

JasonGross:
>  I did a manual fix for this, what do you think?

Seems reasonable.

> Globally, I think it would make sense just to build coq with `-local`, update the path, and simplify a bit the script.

I'd also be okay with that.

etc/ci/install_coq.sh
ejgallego: Why `coqlight` rather than `states tools` ? :D Jokes asides, `coqlight` doesn't build anything useful for HoTT, so why to do use it? Also, both targets are widely used by developers unlike `coqlight` that may disappear at any time.

.travis.yml
ejgallego: What is the point of forcing `v8.6`? This is the default build now.

.travis.yml
ejgallego: We need to update `dpdgraph` then, it is doable. I dunno however why it is embedded instead of a submodule?

ejgallego: Unfortunately I won't have time to the update in the next weeks, so feel free to take over.

.travis.yml
JasonGross: No, the default build should be the `V8.6` tag (i.e., 8.6 released); this says to also try building with the tip of the `v8.6` (or `v8.5`) branch.

ejgallego: IMHO no point in testing against `V8.6`, it has too many bugs; maybe `V8.6.1` would indeed make sense. But if you want I can enable it of course.

ejgallego: But I would rather do the default against the branch, and the single one against the release tag.

.travis.yml
JasonGross: Oh, does the bundled `dpdgraph` not compile with 8.6?

The reason it's bundled is that when we started using it, the only available versions built [with Coq 6.3](http://www-sop.inria.fr/lemme/coq-graphs/) and [with Coq 8.3 with a not-totally-working-version for Coq 8.4](https://anne.pacalet.fr/dev/dpdgraph/).  Now that there's [a github repo](https://github.com/Karmaki/coq-dpdgraph), we could just make that a submodule.  I might take a stab at it when I get a chance.

.travis.yml
JasonGross: There is no `V8.6.1` tag, and the latest released version is still `V8.6`.

> But I would rather do the default against the branch, and the single one against the release tag.

I strongly disagree; sometimes the Coq build breaks, and we should not block merging HoTT PRs just because the tip of the v8.6 branch does not build.  However, it's nice to know when we've broken something relative to the tip of v8.6, which is why I want to have it, and make it an allowed failure.

ejgallego: Ok, fixed!

ejgallego:
So I think we are good but for the dpdgraph integration.

ejgallego:
> I'd also be okay with that.

Cool, I propose further updates to happen in #851 if you agree.

JasonGross:
Closing as superseded by #851.

ejgallego:
Umm, I am not sure it makes "development" sense to close this PR; #851 is all about Travis, however this one is mainly about integrating Matthieu's changes so HoTT runs in 8.6. I'd rather have this one merged (with the necessary fixes to dpdgraph if you think they cannot be delayed), and talk about Travis in #851

JasonGross:
Oh.  I think my current preference is not merging the 8.6 changes until we have Travis + dpdgraph working, but I guess there's the additional changes you want to make to travis in #851?

ejgallego:
Yes, that was the idea thanks!

ejgallego:
This seems to be bitrotting, couldn't we make incremental progress here and leave the graphviz fix for later?

JasonGross:
@ejgallego Can you use the commit at the tip of https://github.com/JasonGross/HoTT/tree/dpdgraph-submodule to fix dpdgraph?

ejgallego:
@JasonGross I'll try thanks.

JasonGross:
Closed by #863

PR #848, opened 2016-12-24
Use functor notations [F _0 x] and [F _1 m] more

JasonGross:
This paves the way for making `Functor C D` unify with `Functor C' D'` so long as the fields unify.  See #849.


mikeshulman:
LGTM

PR #849, opened 2016-12-24
Make functor types unify if fields are the same

JasonGross:
If we ever get [feature #5291](https://coq.inria.fr/bugs/show_bug.cgi?id=5291), a judgmentally singleton type, then this change will bring us closer to allowing `1 → Cᵒᵖ` and `(1 → C)ᵒᵖ` to unify judgmentally, I believe.  (I think we'll have to do the same thing for natural transformations, though.)

See also [feature #5293](https://coq.inria.fr/bugs/show_bug.cgi?id=5293).

This is on top of #848.

I'm not sure whether or not we should merge this; the indirection adds 36 seconds to build time.  My guess is that it's because there are places where we're not taking advantage of primitive projections... @mattam82, would you be able to confirm this?

```
3m37.15s | Total                                                     | 3m00.75s || +0m36.39s
0m23.97s | Categories/Adjoint/Pointwise                              | 0m05.76s || +0m18.21s
0m09.19s | Categories/Adjoint/Functorial/Laws                        | 0m04.24s || +0m04.94s
0m04.91s | Categories/LaxComma/CoreLaws                              | 0m02.20s || +0m02.71s
0m05.24s | Categories/ExponentialLaws/Law2/Law                       | 0m04.16s || +0m01.08s
0m04.16s | Categories/ExponentialLaws/Law4/Law                       | 0m02.86s || +0m01.30s
```

### DIFF
--- a/theories/Categories/Adjoint/Composition/LawsTactic.v
+++ b/theories/Categories/Adjoint/Composition/LawsTactic.v
@@ -24,17 +24,17 @@ Ltac law_t :=
   rewrite !transport_forall_constant;
   repeat
     match goal with
-      | [ |- context[transport (fun y => ?f (@object_of ?C ?D y ?x))] ]
+      | [ |- context[transport (fun y : Functor ?C ?D => ?f (y _0 ?x)%object)] ]
         => rewrite (fun a b => @transport_compose _ _ a b (fun y' => f (y' x)) (@object_of C D))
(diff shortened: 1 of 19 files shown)


mikeshulman:
What's the status of this PR -- are you still working in this direction?  If not, maybe we can close it?  I don't really want to add 30 seconds to build time.

JasonGross:
Let's revisit this if we ever get judgment eta for the unit type; this is only useful for making progress on Kan extensions when combined with that.

PR #794, opened 2016-01-24
Make HoTT compile with 8.5beta3

JasonGross:
We probably don't want to merge this yet, given that 8.5beta3 incurs a 3x slowdown over 8.5beta2:
```
After    | File Name                                                 | Before   || Change
--------------------------------------------------------------------------------------------
9m17.26s | Total                                                     | 3m35.47s || +5m41.79s
--------------------------------------------------------------------------------------------
1m01.59s | categories/Category/Sigma/Univalent                       | 0m11.02s || +0m50.57s
0m45.68s | hit/FreeIntQuotient                                       | 0m01.93s || +0m43.75s
0m38.37s | Spaces/Finite                                             | 0m05.67s || +0m32.69s
0m40.97s | hit/V                                                     | 0m11.04s || +0m29.92s
0m35.95s | contrib/Freudenthal                                       | 0m06.62s || +0m29.33s
0m34.28s | Spaces/No                                                 | 0m05.20s || +0m29.08s
0m29.99s | categories/LaxComma/CoreLaws                              | 0m02.98s || +0m27.00s
0m23.35s | categories/Category/Paths                                 | 0m02.38s || +0m20.97s
0m13.54s | hit/Truncations                                           | 0m01.10s || +0m12.43s
0m20.43s | Spaces/BAut/Bool                                          | 0m11.50s || +0m08.92s
0m06.82s | categories/Functor/Composition/Functorial/Attributes      | 0m00.85s || +0m05.97s
0m06.84s | Spaces/BAut                                               | 0m02.41s || +0m04.42s
0m20.16s | categories/Adjoint/Pointwise                              | 0m24.01s || -0m03.85s
0m10.32s | Algebra/ooGroup                                           | 0m06.72s || +0m03.60s
0m08.80s | Modalities/Lex                                            | 0m05.38s || +0m03.42s
0m04.79s | contrib/HoTTBook                                          | 0m01.22s || +0m03.57s
0m03.68s | hit/quotient                                              | 0m00.55s || +0m03.12s
0m11.83s | Modalities/ReflectiveSubuniverse                          | 0m08.96s || +0m02.86s
0m07.41s | Pointed                                                   | 0m05.02s || +0m02.39s
0m03.71s | hit/Connectedness                                         | 0m01.39s || +0m02.32s
0m02.35s | categories/GroupoidCategory/Dual                          | 0m00.19s || +0m02.16s
0m06.83s | Idempotents                                               | 0m05.77s || +0m01.06s
[...]

PR #851, opened 2017-02-04
[travis] Ocaml.4.02.3 + Coq 8.7

ejgallego:
We update Travis building on #846, but following a bit more the recent Coq's Travis file. In particular, we include a non-broken Ocaml version which allows us to test Coq 8.7.

This seems ready to merge but could use some more refinements, in particular wrt the matrix and packages.

Also, there are good news, HoTT works well with current Coq trunk. I understand current Coq's policy is to treat compatibility breakages very seriously, thus I suggest the `trunk` build remains enabled.

Possible TODOs:

- [ ] Build coq with `-local` at `install`.
- [ ] Cache coq build.
- [ ] Remove `sudo` use.
- [ ] Improve matrix for packages.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = v8.6
+	branch = trunk
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
 	url = https://github.com/JasonGross/coq-scripts.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -8,3 +8,4 @@
 [submodule "etc/coq-dpdgraph"]
 	path = etc/coq-dpdgraph
 	url = https://github.com/Karmaki/coq-dpdgraph.git
+	branch = coq-trunk
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = v8.7
+	branch = master
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
 	url = https://github.com/JasonGross/coq-scripts.git
 [submodule "etc/coq-dpdgraph"]
 	path = etc/coq-dpdgraph
 	url = https://github.com/Karmaki/coq-dpdgraph.git
-	branch = coq-v8.7
+	branch = coq-trunk
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = master
+	branch = v8.7
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
 	url = https://github.com/JasonGross/coq-scripts.git
 [submodule "etc/coq-dpdgraph"]
 	path = etc/coq-dpdgraph
 	url = https://github.com/Karmaki/coq-dpdgraph.git
-	branch = coq-trunk
+	branch = coq-v8.7
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = trunk
+	branch = master
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
 	url = https://github.com/JasonGross/coq-scripts.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,6 @@ matrix:
   allow_failures:
     - env: VALIDATE="yes"
     - env: FORCE_COQ_VERSION="v8.6"  BUILD_COQ="yes" WITH_AUTORECONF="yes"
-    - env: FORCE_COQ_VERSION="trunk" BUILD_COQ="yes" WITH_AUTORECONF="yes"
 
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,9 +47,9 @@ env:
    - UPDATE_QUICK_DOC="yes"  WITH_AUTORECONF="yes" UPDATE_AUTOGEN="yes" UPDATE_OPAM="yes"
    - UPDATE_HTML="yes"       WITH_AUTORECONF="yes"
    - VALIDATE="yes"
-   - FORCE_COQ_VERSION=""      BUILD_COQ="yes" WITH_AUTORECONF="yes"
-   - FORCE_COQ_VERSION="v8.6"  BUILD_COQ="yes" WITH_AUTORECONF="yes"
-   - FORCE_COQ_VERSION="trunk" BUILD_COQ="yes" WITH_AUTORECONF="yes"
+   - FORCE_COQ_VERSION=""       BUILD_COQ="yes" WITH_AUTORECONF="yes"
+   - FORCE_COQ_VERSION="v8.7"   BUILD_COQ="yes" WITH_AUTORECONF="yes"
+   - FORCE_COQ_VERSION="master" BUILD_COQ="yes" WITH_AUTORECONF="yes"
 
  global:
 # encrypted version of deploy key for pushing
@@ -60,7 +60,7 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: VALIDATE="yes"
-    - env: FORCE_COQ_VERSION="v8.6"  BUILD_COQ="yes" WITH_AUTORECONF="yes"
+
 
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: VALIDATE="yes"
-    - FORCE_COQ_VERSION="master" BUILD_COQ="yes" WITH_AUTORECONF="yes"
+    - FORCE_COQ_VERSION="v8.7"   BUILD_COQ="yes" WITH_AUTORECONF="yes"
 
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,6 +60,7 @@ matrix:
   fast_finish: true
   allow_failures:
     - env: VALIDATE="yes"
+    - FORCE_COQ_VERSION="master" BUILD_COQ="yes" WITH_AUTORECONF="yes"
 
 
 
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # don't fall back on git if you interrupt or kill this script
 trap "exit 1" SIGINT SIGTERM
--- a/coq/theories/Init/Notations.v
+++ b/coq/theories/Init/Notations.v
@@ -1,6 +1,6 @@
 (************************************************************************)
 (*  v      *   The Coq Proof Assistant  /  The Coq Development Team     *)
-(* <O___,, *   INRIA - CNRS - LIX - LRI - PPS - Copyright 1999-2012     *)
+(* <O___,, *   INRIA - CNRS - LIX - LRI - PPS - Copyright 1999-2016     *)
 (*   \VV/  **************************************************************)
 (*    //   *      This file is distributed under the terms of the       *)
 (*         *       GNU Lesser General Public License Version 2.1        *)
@@ -59,7 +59,7 @@ Reserved Notation "( x , y , .. , z )" (at level 0).
 
 (** Notation "{ x }" is reserved and has a special status as component
     of other notations such as "{ A } + { B }" and "A + { B }" (which
-    are at the same level than "x + y");
+    are at the same level as "x + y");
     "{ x }" is at level 0 to factor with "{ x : A | P }" *)
 
 Reserved Notation "{ x }" (at level 0, x at level 99).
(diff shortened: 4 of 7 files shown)


.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = trunk
+	branch = V8.6
```
JasonGross: I think this should be `v8.6`, unless `branch` can also refer to tags?  However, the submodule itself should be checked out to the `V8.6` tag, which I'll trust that you've done.

.travis.yml
```diff
@@ -1,26 +1,52 @@
-language: python
-# python, for proviola
-
 dist: trusty
+sudo: required
```
JasonGross: Do we need sudo?  (Is it faster with sudo?)

.travis.yml
```diff
@@ -1,26 +1,52 @@
...
+addons:
+  apt:
+    sources:
+    - avsm
+# new version of graphviz (>= 2.38) to work around trouble in init_rank
+    - sourceline: 'ppa:jgross-h/graphviz'
+    packages:
+    - opam
+    - aspcud
+# This should be conditional on the matrix
```
JasonGross: You should be able to do this by doing
```travis
matrix:
  - env: <first env>
    packages:
      - <packages for first>
  - env: <second env>
...
```
Are you willing to try that and see if it works?  (On the other hand, I'm not sure this will play nicely with caching apt...)

JasonGross:
I'm still unsure about trunk and v8.6 being not allowed to fail; it's not so much incompatibilities that I'm worried about, as it is someone accidentally pushing something that breaks the build of Coq.  I've seen this happen a lot, and I don't think we should block PRs being merged on it.  I'm willing defer to others, though. @andrejbauer , @mikeshulman , @spitters do you have opinions on this?

I'd also like to fix dpdgraph before we merge this.  This is currently blocked on my two [pull](https://github.com/Karmaki/coq-dpdgraph/pull/12) [requests](https://github.com/Karmaki/coq-dpdgraph/pull/13) being merged into upstream coq-dpdgraph.  In particular, without the second of these, the tool will stack-overflow on anything involving primitive projections.

.travis.yml
ejgallego: It seems faster in our tests, and the current scripts use `sudo` in many places; I guess it could be eventually be removed.

.travis.yml
ejgallego: I'll try, the solution that works for me is using `matrix: ... include:` as done in https://github.com/ejgallego/coq/tree/travis-2

.gitmodules
ejgallego: The checkout is at `V8.6` indeed; I thought in git there were not differences between tags and branches, no problem in changing it back.

ejgallego:
> I'm still unsure about trunk and v8.6 being not allowed to fail; it's not so much incompatibilities that I'm worried about, as it is someone accidentally pushing something that breaks the build of Coq.

I am not sure I follow what you mean, by definition if someone pushes that, it won't pass CI. By the way all changes in Coq are going to be tested vs HoTT now (see my travis tree) Coq is going to be released every 6 months and the idea is that all important developments should _always_ compile with trunk, or either they may not be supported upstream.

> I'd also like to fix dpdgraph before we merge this.

Sounds reasonable, however maybe we are accumulating too much stuff here? IMVHO the dpd stuff could progress independently from the rest.

mikeshulman:
My inclination would also be not to require builds to succeed against the current development tip of Coq, if that's what you mean -- the latter seems too unstable to block our own PRs on until there is at least a release candidate.

ejgallego:
> the latter seems too unstable

Ok, I will switch back the failure setting. However note that the general idea in the Coq community is to make `trunk` stable enough by all the major contribs, otherwise there are problems with lack of testing and a 6-month release schedule may be, in fact, unattainable. 

It is being discusses whether Coq's PR will blocked on breaking HoTT etc... but regardless of that IMHO it would make great feedback if you could report on breakage.

.gitmodules
ejgallego: Ok done.

.travis.yml
ejgallego: Hi @JasonGross , this doesn't work as you wrote, I get a syntax error: https://travis-ci.org/ejgallego/coq/builds/198574788

andrejbauer:
Coq people are very brave to put HoTT around their necks.

When HoTT fails to compile with Coq trunk, the fault may be either with HoTT or with Coq trunk, or both. I do not want to make *any* promise that HoTT will fix its part of the problems quickly enough to allow Coq to progress with its intended speed.

That is, Coq people may wish to test against HoTT, and that is probably useful for them, but they should not expect us to fix our problems quickly. We're small fish, they're big fish. They should just swim ahead, and we'll cope.

andrejbauer:
Regarding which version of Coq to use: we should use a reasonably new version of Coq, but we should *not* insist that the latest version of Coq is the one that we must be using.

maximedenes:
I think you would be the first ones to benefit from a tighter integration of HoTT in our (Coq's) testing infrastructure. For example, we are studying the integration of a complex PR (https://github.com/coq/coq/pull/379) that fixes some bugs in unification. I already spent a week (full time) understanding the details of those changes and the fixes required on the developments we test, including HoTT.

Having us submit such fixes to you is probably a much more lightweight workflow (for you!) than waiting for the next Coq release and having to investigate all those problems on your own. Especially if you are small fish :)

spitters:
@maximedenes I think it would be great if we could start following 8.7.
With @SkySkimmer and @FFaissole we are building on @mattam82 's experimental IR branch, which I understand will be merged into 8.7. It would be great if we would have one less experiment to worry about.

At the same time, @SkySkimmer 's work has has already been a good testbed for IR.

JasonGross:
@ejgallego Can you rebase this on master?  It would be nice to see the actual diff between 8.6 and 8.7 for HoTT.

ejgallego:
The right branch to look as for the development is https://github.com/ejgallego/HoTT/tree/mz-8.7 , I've rebased it, the diff is pretty minimal, but however the rebased build is currently broken :(

I've put the rebased repos at https://github.com/ejgallego/HoTT/tree/mz-8.7-rebased , the error is:
```
File "./theories/Spaces/BAut.v", line 101, characters 2-185:
Error:
Ltac call to "refine (uconstr)" failed.
In environment
H : Univalence
X : Type
P : Type -> Type
H0 : forall Z : BAut X, IsHSet (P Z)
The term "((equiv_sigT_ind (fun Zp : {x : _ & ?T0 x} => P Zp.1))^-1)%equiv"
has type
 "(forall xy : {x : _ & ?T0 x}, P xy.1) <~>
  (forall (x : Type) (y : ?T0 x), P (x; y).1)"
while it is expected to have type
 "(forall x : {x : _ & ?T0 x}, P x.1) <~> ?B0" (cannot satisfy constraint
"P (Z; p).1" == "?T"; cannot instantiate "?T" because 
"p" is not in its scope: available arguments are "H" 
"X" "P" "H0" "Z").
```
Likely an EConstr thing.

ejgallego:
Well in fact the rebase of this branch is identical to the mz-8.7-rebased one, so we are done I guess.

ejgallego:
The issue of making HoTT compile with 8.7 should go to a different PR I'd guess.

ejgallego:
Aha, I had messed up my merge, now

JasonGross:
If you want to make a separate PR(s) with 2182b19 and f3d5a91, I'd be happy to merge those right away.  I think we shouldn't merge changes that break the 8.6 build until 8.7 comes out, though.  (Also, shouldn't you be updating `.travis.yml` to actually build trunk rather than 8.6 in this PR?)

ejgallego:
Ok, done in #865.

JasonGross:
Thanks!

ejgallego:
@JasonGross I am wondering about whether to remove `BUILD_COQ`, `WITH_AUTORECONF`, and `UPDATE_AUTOGEN`. What are your thoughts? Do you think the overhead is worth the utility? [as for instance, auto-deploying a `.tar.bz2` with the correct files on tag/release?]

ejgallego:
Also, it is worth to have a submodule for Coq as opposed to just do a shallow clone of the correct upstream tag?

ejgallego:
In fact, I would maybe make the Coq version mandatory in the script, (with a set default). For better or worse, distros are gonna lag behind wrt Coq 8.6 schedule.

JasonGross:
I think we should get input from @mikeshulman, @andrejbauer, and @spitters on how we want users to install things.  I'm personally indifferent to whether or not we require `autoreconf` (allowing users to not have to install that is the motivation for testing a build without it), and indifferent to how we manage the dependency on a particular version of Coq.  I seem to recall some of the others wanting to support users who don't know much about command line tools, though?

ejgallego:
Maybe you could just deploy the proper tarball with the autofoo stuff at each build, instead of having to use `autogen_and_push.sh`. Github releases allows that automatically but only for releases that are tagged.

spitters:
@JasonGross I don't fully understand how @ejgallego 's proposal would effect users not comfortable with the command line. However, generally, I think we are doing well given that we now depend on stable Coq and have opam.

mikeshulman:
Could someone summarize exactly what the issues and options being considered are?  I got somewhat lost in the conversation.

ejgallego:
IMHO it should have minimal impact, the main purpose is to remove a large mass of HoTT custom scripts that may not be worth the maintenance effort anymore; I guess that's under @JasonGross hat.

JasonGross:
@mikeshulman As I understand it, the options are:
1. Keep the scripts as they currently are.
  - Benefits: not much immediate work.
  - Downsides: It's not clear the autofoo stuff actually works (see #868), and it may be a maintenance burden.
2. Drop the autoconf-and-push travis stuff
  - Benefits: we stop using code we weren't really testing, less code
  - Downsides: we will no longer support end-users of the HoTT library who don't install auto(re)conf on their local machine
3. Drop the submodule, include particular versions of Coq in the build scripts
  - Benefits: we'll be making less use of git submodules (@ejgallego, any others you see?)
  - Downsides: we'll likely lose the ability to peg versions of Coq that are neither released nor the tip of a branch

We can either do 1, or 2, or 3, or 2 & 3.  Does this all seem right?

mikeshulman:
I like the sound of 2 because I like simplicity, and having code hanging around that isn't tested and that the core developers don't use themselves seems like a bad idea.  I don't know how many users this would affect or how hard it would make it for them.  How hard is it to install autoreconf on a Windows or Mac?  Is it likely to be burdensome for anyone who is already installing other things we require like git and make?  Are they just extra boxes to check in Cygwin, for instance?  And if I understand correctly, this would only affect people who actually want to contribute to the library, since people who just want to use it can use the opam package instead?

Regarding 3, how would we as the developers record which version of Coq the library uses?  By editing the build scripts or some conffile from which they grab the Coq version to checkout?  That sounds like basically reimplementing submodules-lite on our own.  I think most of the problems with submodules arise when you have the same people coding on the submodule and the supermodule; here I think we're actually using them for their intended purpose, and it seems like a good idea to be able to peg arbitrary commits of Coq if we want to track Coq development closer than major releases.

So I guess my inclination is to do 2 only, but I don't feel particularly strongly.

JasonGross:
> How hard is it to install autoreconf on a Windows or Mac?

On Windows, it is just another box to tick in cygwin.  On Mac, it seems to be as simple as
```
sudo port install automake autoconf libtool
```
or
```
brew install automake
```

I'll plan to make a PR to do 2.

Re 3, I guess another benefit is that if we use a file rather than a submodule, we don't need `git submodule sync` when checking out different commits or branches that change the version of Coq.  I imagine the best way to do it would be to have either a file that records the version/tag, or a line in configure that sets a variable.

ejgallego:
For the autoconf stuff you could always use tags and github releases to produce a tarball including the proper files.

SkySkimmer:
This is broken after coq/coq@db601b06b, you need to remove the extra `Declare ML Module`s.

ejgallego:
> This is broken after coq/coq@db601b0, you need to remove the extra Declare ML Modules.

@SkySkimmer thanks for pointing it out, indeed it is strange that @maximedenes merged #794 when it was explicitly stated that HoTT needed a patch, I guess he didn't realize.

I won't have time to fix this now, feel free to force-push or submit a PR.

ejgallego:
@JasonGross what's the status of this, should we merge some partial bits? Coq 8.7 will be released in a couple of days.

JasonGross:
Blocked (kind-of) on #883.  I haven't made time to get the branches/tags of HoTT on good order for that.

JasonGross:
I've pushed a commit that updates travis to use master, rather than the outdated trunk.  I think this should be merged as soon as travis is happy.

JasonGross:
@ejgallego Coq fails to build from source with
```
Warning: plugins/micromega/micromega.ml and the code generated by plugins/micromega/MExtraction.v differ !
make[1]: *** No rule to make target `grammar/compat5.cmo'. Stop.
make[1]: Leaving directory `/home/travis/build/HoTT/HoTT/coq-HoTT'
```

This is with V8.7.0, v8.7, and master.  Maybe something is up with the OCaml version?  Can you look into fixing this?

maximedenes:
> Warning: plugins/micromega/micromega.ml and the code generated by plugins/micromega/MExtraction.v differ !

This is a warning, not an error.

> make[1]: *** No rule to make target `grammar/compat5.cmo'. Stop.

This `compat` file no longer exists. Did you `git clean`?

JasonGross:
Oh, right, we were manually telling Coq to make `grammar/compat5.cmo` because the `coqlight` and/or `install-coqlight` targets were removed.

JasonGross:
Because the standard library takes too long to build and I don't want to wait for it for every travis build, just to throw it away.

ejgallego:
Thanks @JasonGross

PR #852, opened 2017-02-14
Remove a nested comment to make proviola happy

SimonBoulier:
As you can see here: http://hott.github.io/HoTT/coqdoc-html/HoTT.HIT.Suspension.html
The nested comment just before [loop_susp_unit](http://hott.github.io/HoTT/proviola-html/HoTT.HIT.Suspension.html#loop_susp_unit) is badly handled and make the whole file corrupted.

As this comment is redundant, I propose to remove it.

### DIFF
--- a/theories/HIT/Suspension.v
+++ b/theories/HIT/Suspension.v
@@ -252,7 +252,6 @@ Module Book_Loop_Susp_Adjunction.
       reflexivity.
     - cbn.
       Fail reflexivity.
-      (** ...but this goal is pretty intractable. *)
   Abort.
 >>
    *)


JasonGross:
I would rather move the comment text down two lines and unnest it.  (Better, fix proviola, but I don't want to block integrating colimits on that.  Any chance you could make a reproducing test case and report it as a bug on https://github.com/JasonGross/proviola-source ?)

SimonBoulier:
It's already written on line 234, do you really prefer to repeat it?

JasonGross:
Oops, missed that.  Thanks!

PR #853, opened 2017-03-10
The join of propositions is their disjunction

mikeshulman:


spitters:
LGTM. Feel free to merge after answering @JasonGross question. I don't recall such a convention.

spitters:
LGTM. Feel free to merge after answering @JasonGross question. I don't recall such a convention.

mikeshulman:
I don't think there is such an explicit convention, but possibly `equiv_join_hor` would be a little more in line with other names, so I'll change it.

mikeshulman:
Hmm, something weird just happened: somehow the commits that appeared on the master branch are different from the ones I committed in the PR.  Did you "merge" this as a rebase?

JasonGross:
Yes, I did.  Did I mix up the conventions?

mikeshulman:
I don't know whether we established that as a convention, but if we did, I'd like to change it.  If the commits that end up in master are different from the ones that I originally commit, then even after I pull the master branch from upstream, my local branch that I created for the PR still shows as "not fully merged into master", which makes me uneasy about deleting it.

JasonGross:
Ah, yeah, that makes sense.  I wonder if we can disable rebasing in GitHub options.  Anyway, you can rebase your local branch on master, and then git will think it's fully merged.

mikeshulman:
Disabled.  Thanks!

PR #858, opened 2017-03-16
Constructing the natural numbers using propresizing and univalence

mikeshulman:
This is a "dead-end" development, i.e. I don't expect anyone would use this Nat instead of the usual one; the point is rather semantic, that an "elementary (oo,1)-topos" automatically has a NNO.  For that reason I'm a bit worried about including it in the library given how slow it is, but at least some abstraction speeds it up so it's only the _second_ slowest file.  (-:  Suggestions for further speedup would be welcome.

```
3m48.69s | Total                                                     | 3m27.96s || +0m20.73s
0m16.26s | PropResizing/Nat                                          |   N/A    || +0m16.26s
0m18.54s | Spaces/BAut/Bool                                          | 0m18.41s || +0m00.12s
0m11.06s | Algebra/ooGroup                                           | 0m11.08s || -0m00.01s
0m09.84s | HIT/V                                                     | 0m09.87s || -0m00.02s
...
```


JasonGross:
I've made a 2x speedup in https://github.com/mikeshulman/HoTT/pull/1

SkySkimmer:
Since the accessibility predicate is HProp I used well-founded induction to skip the entire NRec section, after which PropResizing/Nat.v is only the 39th slowest file (including the @JasonGross changes): https://github.com/SkySkimmer/HoTT/blob/propresizing/theories/PropResizing/Nat.v
Using an inductive definition for <= also avoids defining addition.

Is this somehow cheating?

mikeshulman:
Thanks @JasonGross !  I'll look at it.

Yes, @SkySkimmer , that's cheating.  (-:  The point is that in an elementary (oo,1)-topos whose _only_ HITs (including ordinary ITs) are non-recursive, we can still construct an NNO.  The inductive Acc predicate is recursive.

mikeshulman:
I took all of your suggestions, @JasonGross, and added some comments for the casual reader.

SkySkimmer:
Non cheating speedup at mikeshulman#2.
It's all universes and removing the use of `merely` (which I think makes new universes every time it's used).
PropResizing/Nat.v goes from 09.67s to 02.48s, becomes 24th slowest in the repository.

JasonGross:
Neat!

I wonder if the new universe algorithm in 8.6 would have a similar effect.  I'm also curious if @mattam82 has any thoughts

> which I think makes new universes every time it's used

Any idea ways going on here?

SkySkimmer:
I think the new algorithm is for checking that the universe graph is acyclic from coq/coq@551a03d, so while it would speed up things it can't remove the 3518 universes and 3199 constraints on contr_NRec.

(To see the number of universes do `Check contr_NRec@{Set}.`, if it works there's only 1 otherwise it will complain with `Error: Universe instance should have length 3518`. To see the number of constraints do `About constr_NRec.` with print universes enabled and count the constraint lines.)

>Any idea what's going on here?

This is speculation:
`ishprop_merely` (notably used in `functor_merely`) has more universes than it needs. If it gets used multiple times at the same universes (and same type) for `merely` the extra ones seem to get unified, but different appearances of merely don't get the same universes as the hidden `P : Type@{j}` may need to be different (and so the `i |= j < i, A : Type@{i}` as well).
The use of propositional resizing in `merely_rec` may help keep them from getting unified, but I don't think it's necessary (mostly because I've had universe blowup from extra universes on some lemma before and I didn't use resizing).

mikeshulman:
Nice, thanks!

I do wish that there were a way to hide all the universe annotations, since they make it significantly harder for someone to actually read the code.

SkySkimmer:
Update: universes are collapsed as much as possible.
The constraints on `contr_NRec` are now readable:
```
contr_NRec@{s u p large x nr} : Contr@{nr} NRec@{s u p large x nr}
(* s u p large x nr
|= s < u < p
   p <= nr
   x <= nr
   nr < large
    *)
```
where
```
vertex g, edge g x y : Type_s
Graph : Type_u
N : Type_p
X : Type_x
```
Morally,
`u = s+1` and `p = u+1 = s+2`
`nr = max(p,x)`
and `large` is the max of `p` and the universes of each type we use `N_propind` on.

mikeshulman:
Can anyone figure out why Travis is unhappy?  I can never read its logs any more.

JasonGross:
I think it seems to think that `dot` isn't installed, for reasons that I haven't managed to track down.  Any thoughts on what's going on?  Or how to make the logs more readable for you?

mikeshulman:
No idea about dot.  The logs are just too long for me to find the relevant information; maybe we could dial down the verbosity some (e.g. do we need all the output of building Coq?)

SkySkimmer:
Travis does a bunch of builds with different options, and only one succeeds, the one with
`UPDATE_HTML=""       WITH_AUTORECONF="yes"  UPDATE_QUICK_DOC="yes"  UPDATE_AUTOGEN="yes"  UPDATE_DEP_GRAPHS=""    BUILD_COQ="yes"  UPDATE_OPAM="yes"       FORCE_COQ_VERSION=""`
`UPDATE_QUICK_DOC="yes"  UPDATE_AUTOGEN="yes" UPDATE_OPAM="yes"` are special to it.

mikeshulman:
Resolved conflicts to restore mergeability.

mikeshulman:
And I don't know what happened, but Travis is now happy.  Any acks for merging?

PR #859, opened 2017-03-16
Give trunc_{contr,hprop,hset} only one universe

JasonGross:
This is part of an attempt to make the hpropresizing branch faster.  I'm not sure I picked the right universe to remove `theories/Modalities/Closed.v`; @mikeshulman, can you take a look at that change?

### DIFF
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -58,14 +58,14 @@ Defined.
 
 (** In particular, a contractible type, hprop, or hset is truncated at all higher levels.  We don't allow these to be used as idmaps, since there would be no point to it. *)
 
-Definition trunc_contr {n} {A} `{Contr A} : IsTrunc n.+1 A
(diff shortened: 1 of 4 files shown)


spitters:
I don't see a problem with this.

JasonGross:
> But why not eliminate the unnecessary universe already in `trunc_leq`?

Because it shows up in the type of `trunc_leq`, whereas these extra universes show up only in the bodies and not in the type:
```coq
trunc_leq@{Top.109 Top.111} :
forall (m n : trunc_index) (_ : trunc_index_leq@{Top.109} m n)
  (A : Type@{Top.111}) (_ : IsTrunc@{Top.111} m A), 
IsTrunc@{Top.111} n A
(* Top.109 Top.111 |=  *)
```

We could make `trunc_index_leq` live in `Type0` rather than `Type`; perhaps that's better?

mikeshulman:
Looks good!

PR #857, opened 2017-03-16
IsTruncMap n.+1 f <-> IsTruncMap n (ap f)

SkySkimmer:

PR #861, opened 2017-04-03
Lemma 7.6.2 [IsTruncMap n.+1 f <-> IsTruncMap n (ap f)] to HoTTBook

SkySkimmer:
I only noticed that it was a lemma in the book after #857 was merged.


mikeshulman:
LGTM

PR #863, opened 2017-04-18
Update version to Coq 8.6

JasonGross:
I finally got around to fixing the dependency graphs on travis for 8.6.

This closes #841 and #846.

I'd suggest that after this is merged, we push a `v8.5` branch that's the tip master right before this is merged.


spitters:
Thanks. LGTM.

mikeshulman:
thanks!

spitters:
@JasonGross 
Given Tadeusz question on Coq-club, is it possible that we forgot to update INSTALL.md and opam?

spitters:
Maybe @ejgallego or @matejkosik know more about this. I understand we are on 8.6 stable.
https://github.com/coq/opam-coq-archive/tree/master/extra-dev/packages/coq-hott/coq-hott.8.6.dev

SkySkimmer:
INSTALL.md is definitely out of date:
>The HoTT library requires a custom version of Coq with three modifications: the command-line option -indices-matter, universe polymorphism, and private types. The first one changes the interpretation of equality to one that conforms to the homotopy-theoretic interpretation; the second one is needed to avoid certain universe inconsistency errors (which are not really there, but Coq **8.4** cannot see that); and the third one is a modification which allows us to implement a poor man's version of higher-inductive types.

matejkosik:
@spitters I do not know the proper precedure (to install HoTT).

I've created an OPAM package (for benchmarking HoTT) but I just followed (more or less) what @ejgallego was doing.

spitters:
@matejkosik thanks for your work. We depend on
https://github.com/coq/coq/tree/0d1438851ba3a0b9f76847abc42f3bf8ad26c4cb
(that is coq 8.6), so I think you can safely update the opam repo to the stable 8.6 version.

JasonGross:
@matejkosik I think the issue is that @ejgallego 's patch for 8.6 compatibility has been merged into our master branch, so it no longer applies, and thus gives an error.  So you should just be able to drop the `emilio.diff` patch and have it work (but I haven't tested this).

matejkosik:
That is (I hope) a good news. I am going to update the package and test it. I'll let you know the result.

matejkosik:
The `coq-hott.8.6.dev` is updated and (at the moment) [installable](https://ci.inria.fr/coq/job/opam-install-coq-hott/121/console).

The package is still more fragile than necessary because of this [additional package](https://github.com/coq/opam-coq-archive/blob/master/extra-dev/packages/coq-hott/coq-hott.8.6.dev/files/no_clone.diff). The patch modifies the original source-code so that:
- submodules are not cloned
- the available Coq is used instead.

Would HoTT-team be willing to provide some official way how one could (at one's own risk) use the Coq that is directly available in the PATH instead of letting HoTT to clone and compile a new one? We could then drop [the patch](https://github.com/coq/opam-coq-archive/blob/master/extra-dev/packages/coq-hott/coq-hott.8.6.dev/files/no_clone.diff) and use the official mechanism instead.

That's what we do with [coq-unimath](https://github.com/coq/opam-coq-archive/blob/master/extra-dev/packages/coq-unimath/coq-unimath.8.6.dev/opam#L9) too.

spitters:
@matejkosik Thanks! 
I am inclined to think that will work. However, I'll let @JasonGross confirm, as I may be overlooking something.

JasonGross:
@matejkosik Is it really necessary to avoid cloning the submodules?  Certainly, it takes a bit of extra time and bandwidth, and if you want, I could allow `./autogen.sh` to take a `--no-submodules` argument or something, but it seems not terribly important?  Either way, that patch doesn't actually touch the build mechanism of HoTT (it just changes what gets cloned), so you can safely drop it, and HoTT will still pick up the version of Coq in `PATH`.

matejkosik:
@JasonGross From the two options:
- submodules are cloned
- submodules are not cloned

the second one is (for us) strictly better because:
- we would like to measure HoTT compilation time and if cloning submodules is indivisible part of HoTT compilation, the results we get are noisier
- we would like to install HoTT as reliably as possible (if cloning of submodules is indivisible part of HoTT installation, they contribute to unreliability of HoTT installation --- networks and services we rely on are never 100% reliable (e.g. reachability of github).

Is it a wrong assumption that by dropping cloning of submodules, nothing is lost (only gained)?

tadeuszlitak:
Thanks for fixing the issue quickly. It seems, however, it is not possible to use hoqide with the present OPAM setup:

$ hoqide
/Users/litak/ocamlbrew/ocaml-4.02.3/.opam/system/bin/hoqide: line 49: /Users/litak/ocamlbrew/ocaml-4.02.3/.opam/system/bin/coqide: No such file or directory
/Users/litak/ocamlbrew/ocaml-4.02.3/.opam/system/bin/hoqide: line 49: exec: /Users/litak/ocamlbrew/ocaml-4.02.3/.opam/system/bin/coqide: cannot execute: No such file or directory

$ opam install coqide
The following dependencies couldn't be met:
  - coqide -> coq (= 8.4.5 | = 8.4.6 | = 8.5.0 | = 8.5.1 | = 8.5.2 | = 8.5.3 | = 8.6)
  - coq-hott -> coq = 8.6.dev
Your request can't be satisfied:
  - Conflicting version constraints for coq

No solution found, exiting

PR #865, opened 2017-05-13
Tweaks for 8.7, 8.6 compatible part

ejgallego:
These two commits should be 8.6 compatible but are needed for 8.7 due to some bugfixes in Coq.


spitters:
Thanks.

PR #866, opened 2017-05-14
Improve a comment in Lex.v

JasonGross:
Coqdoc has a special meaning for `%`, so we use unicode instead, so that coqdoc doesn't chop off the comment.  (Also change http to https.)


mikeshulman:
Is there no ASCII way to include a `%` in a coqdoc comment?  That seems odd.

JasonGross:
Oh, it looks like we could use `%%` to encode `%` according to https://coq.inria.fr/refman/Reference-Manual017.html.  But then copy-pasting the url from source won't work, and I don't know any non-unicode way to make it work to copy-paste both from source and from coqdoc.

mikeshulman:
Well, I suppose this is ok then.

PR #867, opened 2017-05-18
A few IsTruncMap proofs

SkySkimmer:


spitters:
Looks good. Thanks.

theories/Fibrations.v
```diff
@@ -173,8 +173,26 @@ Section UnstableOctahedral.
     exact (equiv_contr_sigma (fun (w:{b:B & f a = b}) => g w.1 = c)).
   Defined.
 
+  Global Instance istruncmap_compose `{!IsTruncMap n g} `{!IsTruncMap n f}
```
JasonGross: What does ! do?

SkySkimmer: Without it I have to use
``` coq
`{IsTruncMap n _ _ g}
```

The non ! version can also unnecessarily duplicate typeclass arguments, but I can't remember an example offhand. It came up with mathclasses style operational type classes IIRC, so maybe @spitters knows.

Because of this I feel like it's more robust if a bit less powerful, so I use it when I don't need the extra power.

spitters: The ! modifier switches the way a binder is parsed back to the regular interpretation of Coq. In particular, it uses the implicit arguments mechanism if available, as shown in the example.
https://coq.inria.fr/refman/Reference-Manual023.html

JasonGross: Neat!  I learned a new thing today

mikeshulman: Nice to know.  This feels to me like one more in the category of "we did the syntax wrong the first time, but now we don't want to break existing code, so we introduce a new kind of syntax that does what we should have done in the first place".  (-:O

mattam82: The two modes had their uses, but I guess the automatic generalization of missing arguments in the non-! case is indeed less popular now, people seem to rather build hiearchies by packing classes instead of parameterizing by classes.

mattam82: (contrary to Haskell where only the later is available)

PR #869, opened 2017-05-30
Add hoqchk and a validate target

JasonGross:
This will allow the devs (and us) to test coqchk.


spitters:
LGTM

JasonGross:
Rebased, and now on top of #874, so that we don't conflict with that when that one gets merged.

JasonGross:
Rebased again, without #874 (maybe this can get merged while we're waiting to resolve discussion on #874?)

mikeshulman:
I don't really know what this is doing, but I trust you. (-:

JasonGross:
Coq has a separate binary `coqchk` that is an independent validator on the `.vo` files.  It is useful, for example, if someone malicious crafted a bad `.vo` file by hand that claims to contain an axiom-free proof of `False`.  In normal execution, `coqc` trusts that `.vo` files are good; `coqchk` validates this assumption.

It turns out that `coqchk` is not very-well tested, and I'd like the devs to fix a few bugs in `coqchk` and then have HoTT testing coqchk.  That's what this target is for.  (As far as the code goes, I mostly copied it from `coq_makefile`-generated makefiles, or from our other targets.)

PR #870, opened 2017-05-30
Faster travis builds (install, don't build, Coq)

JasonGross:


spitters:
OK. So, you have set up your own ppa.
LGTM, but Travis is not happy

JasonGross:
I believe Travis will be happy now

JasonGross:
Travis is happy; can I get a second ack?

PR #871, opened 2017-05-31
Two more characterizations of lex modalities

mikeshulman:
Turns out the pullback lemma that "probably is not equivalent to lex-ness" is actually equivalent to lex-ness.


spitters:
LGTM. I forgot our policy on unfolds. Coq probably(?) does not need them.
Do we want to keep them to make the proofs readable?

mikeshulman:
I generally like to keep proofs readable.

PR #872, opened 2017-06-01
Silence a number of warnings from Coq

JasonGross:
The first commit contains what I expect are the uncontroversial changes.  The second commit involves modifying the warnings string in various places; we could instead suppress all warnings about overridden notations from the makefile.


spitters:
The first commit is OK. I am ambivalent about the second one. It would be nicer to actually get this fixed. However, we may as well mute the warnings for the time being.

PR #873, opened 2017-06-01
Factor all notation levels through a common file

JasonGross:
This allows us to ensure that notation levels don't conflict (some of them did), and fail fast if they do.  I had to adjust some notations a bit, e.g., I replaced `{{ xL | xR // xcut }}` with `{ { xL | xR // xcut } }` so that it doesn't conflict with `{x : Type@{i}}`, and adjusted the level of `xR` so that it doesn't conflict with the `//` notation for lax comma categories.

For some unknown reason, a number of files now need `Local Open Scope nat_scope` to not get confused about `nat_scope` vs `trunc_scope`.

### DIFF
--- a/theories/Spaces/No.v
+++ b/theories/Spaces/No.v
@@ -1098,8 +1098,7 @@ Section Addition.
       refine (Conway_theorem0_ii_r _ _ _ _ _ (inl r)) ).
   Defined.
 
-  (** Oddly, without the universe annotations here, Coq turns all these [No]s into [No@{Set}}. *)
-  Definition plus (x y : No@{i}) : No@{i}
+  Definition plus (x y : No) : No
     := (plus_outer.1 x).1 y.
 
   Infix "+" := plus : surreal_scope.
--- a/theories/Spaces/No.v
+++ b/theories/Spaces/No.v
@@ -4,6 +4,7 @@ Require Import UnivalenceImpliesFunext TruncType HSet.
 Require Import HIT.Truncations.
 Import TrM.
 
+Local Open Scope nat_scope.
 Local Open Scope path_scope.
 
 (** * The surreal numbers *)
@@ -86,7 +87,7 @@ Module Export Surreals.
                 (isno _ _ _ _ (isno_game_of o xL)
                       (isno_game_of o xR) xcut).
 
-  Local Notation "{{ xL | xR // xcut }}" := (No_cut xL xR xcut) : surreal_scope.
+  Local Notation "{ { xL | xR // xcut } }" := (No_cut xL xR xcut) : surreal_scope.
 
   Axiom path_No : forall (x y : No), (x <= y) -> (y <= x) -> (x = y).
   Arguments path_No {x y} _ _.
@@ -283,7 +284,7 @@ End Surreals.
 
 (** We put this in a module so that it doesn't prevent other people from using notations with double `}}`, e.g. nested sigma-types.  Apparently just putting it in a closed scope is not good enough for that.  Anyone else who wants to use this notation can import this module. *)
 Module Import Surreal_Cut_Notation.
-  Notation "{{ xL | xR // xcut }}" := (No_cut xL xR xcut) : surreal_scope.
+  Notation "{ { xL | xR // xcut } }" := (No_cut xL xR xcut) : surreal_scope.
 End Surreal_Cut_Notation.
 
 (** ** A few surreal numbers *)
@@ -1097,7 +1098,7 @@ Section Addition.
       refine (Conway_theorem0_ii_r _ _ _ _ _ (inl r)) ).
   Defined.
 
-  (** Oddly, without the universe annotations here, Coq turns all these [No]s into [No@{Set}]. *)
+  (** Oddly, without the universe annotations here, Coq turns all these [No]s into [No@{Set}}. *)
   Definition plus (x y : No@{i}) : No@{i}
     := (plus_outer.1 x).1 y.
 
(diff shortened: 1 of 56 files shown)


spitters:
LGTM

theories/Spaces/No.v
```diff
@@ -1097,7 +1098,7 @@ Section Addition.
       refine (Conway_theorem0_ii_r _ _ _ _ _ (inl r)) ).
   Defined.
 
-  (** Oddly, without the universe annotations here, Coq turns all these [No]s into [No@{Set}]. *)
+  (** Oddly, without the universe annotations here, Coq turns all these [No]s into [No@{Set}}. *)
```
mikeshulman: This one I think should be `}]`.

SkySkimmer: Actually I think that comment hasn't been true since 23be0bee4ddc926e0fc7cb7737c5eb22085ec4b1 did `Global Unset Universe Minimization ToSet.`

mikeshulman:
I notice that you didn't change any of the actual uses of the surreal cut notation.  Would than ever be necessary, or will this only affect the display of that notation?

theories/Spaces/No.v
JasonGross: Comment removed

JasonGross:
> Would than ever be necessary, or will this only affect the display of that notation?

This does not affect the display of the notation.  It affects the parsing rules; the notation can now be parsed as `{ { xL | xR // xcut } }`, but it can also be parsed as `{{ xL | xR // xcut }}`, and it prints as `{{ xL | xR // xcut }}`.  If we add some other notation involving `{{` or `}}` rather than `{ {` or `} }`, then we'll have to change the use of the surreal cut notation (but we shouldn't add such notations, and should instead do the same transformation I did here).

JasonGross:
Rebased and updated with the suggestion about the comment.

mikeshulman:
Then I don't understand what this change does and why it is necessary.  If we can still write surreal cuts as `}}`, then it seems there would still be just as much potential ambiguity when conflicting with other notations that have double brackets, and allowing `} }` for surreal cuts seems like it would *prevent* disambiguating `{A : Type@{i}}` by writing it as `{ A : Type@{i} }`.

JasonGross:
> Then I don't understand what this change does and why it is necessary.

The issue is at the lexing phase, not the parsing phase.  If we have a notation involving `}}`, then Coq turns `}}` into a token, and thereafter lexes it as `}}`.  If Coq lexes `{A : Type@{i}}` as `{` `A` `:` `Type` `@` `{` `i` `}}`, then it can't recognize the universe notation.  However, if it lexes it as `{` `A` `:` `Type` `@` `{` `i` `}` `}`, then the parsing phase can do the disambiguation.  (Indeed, if we add a notation `{ foo : bar } }`, then it's quite possible that Coq will confuse it with `{A : Type@{i}}` (though I think it might have special parsing rules in binding locations that prevent that).

mikeshulman:
Ah, I see.  So this change just enables Coq to be smarter about doing what we mean, rather than painting itself into a corner.

JasonGross:
Yes

SkySkimmer:
This broke HoTTClasses, see SkySkimmer/HoTTClasses#2. Fix commit at https://github.com/SkySkimmer/HoTTClasses/commit/bcda9057c26270351475dbda867a536bd653282b.
Breaks are as follows:
# The confusing one
Notation `[ ]` for `nil` does not parse:
``` coq
Require Import HoTT.Basics.Overture.

Open Scope list_scope.
Notation "[ ]" := nil : list_scope.
Check [ ].
(*
Error:
Syntax error: [constr:operconstr level 200] expected after '[' (in [constr:operconstr]).
*)
```
(Without space works though)
I don't know what's going on there.

# The other ones
- `x // y` (division in constructive fields) in HoTTClasses is `(at level 35, right associativity)` but in HoTT is `(at level 40, left associativity)`. The level change is not important AFAIK. Having `x // y // z` mean `(x // y) // z` seems saner than the opposite, I don't know why mathclasses does it the other way.
Changing it to match HoTT only broke `- // x = // [redacted]` which had to be changed to `- (// x) = // [redacted]` so it's OK.
- `x ** y` (exponential) in HoTTClasses is `(at level 30, right associativity)` but in HoTT is `(at level 40, no associativity)`. Here the level change matters since multiplication's `*` is also at level 40 and writing `x * y ** z` happens a few times.
I changed the HoTTClasses notation to `x ^^ y` which worked.
- matching natural numbers with `0` interprets it as a `trunc_index` where before it didn't for some obscure reason. Changing to `0%nat` works.
- in one place `change ((a + (b + c)).+1 = (a + b + c).+1).` didn't detect `nat_scope` properly and interpreted as `trunc_S`, I just changed to non-notation `S`.

mikeshulman:
Is there a reason for our choice of level for `**`?  The only place I can think of that we're using `**` is in `Spaces/BAut/Bool` and I don't think the level would matter very much there.

mikeshulman:
Back on the surreal cut notation: with this change, is it still necessary to have it isolated in a module, or would it now be enough to have it in a scope?

JasonGross:
> Is there a reason for our choice of level for `**`?

Feel free to change it.

> Back on the surreal cut notation: with this change, is it still necessary to have it isolated in a module, or would it now be enough to have it in a scope?

Should be enough to have it in scope, and not isolate it.

mikeshulman:
Okay, see #888 and #889.

JasonGross:
> The confusing one

Can you figure out which reserved notation breaks this?  I suspect it's the `[ x ]` notation, but I don't have access to a computer right now.

I wonder if we now open trunc_scope in a place we didn't, previously...

mikeshulman:
In principle, we should probably discuss this somewhere other than at a closed PR.

PR #868, opened 2017-05-23
Fix travis testing of no-autoconf building

SkySkimmer:
dh-autoreconf is a debhelper (dh) package, what HoTT uses is autoconf which is preinstalled on travis.

Latest build on master at the time of this PR with `WITH_AUTORECONF=""`: https://travis-ci.org/HoTT/HoTT/jobs/234410119
```
$ ./autogen.sh && ./configure && if [ -z "$UPDATE_QUICK_DOC" ]; then make strict-test && ./etc/coq-scripts/timing/make-pretty-timed.sh V=1 -j2 && make && ./etc/ci/test-install-target.sh && cat ./time-of-build-pretty.log && echo; fi && ./etc/ci/after_success.sh

autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I etc
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force
autoreconf: configure.ac: not using Autoheader
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:12: installing 'etc/install-sh'
configure.ac:12: installing 'etc/missing'
autoreconf: Leaving directory `.'
```

First job to have `WITH_AUTORECONF=""`: https://travis-ci.org/HoTT/HoTT/jobs/10635031
```
$ ./autogen.sh && ./configure && make
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
autoreconf: running: aclocal --force -I etc
autoreconf: configure.ac: tracing
autoreconf: configure.ac: not using Libtool
autoreconf: running: /usr/bin/autoconf --force
autoreconf: configure.ac: not using Autoheader
autoreconf: running: automake --add-missing --copy --force-missing
configure.ac:14: installing `etc/install-sh'
configure.ac:14: installing `etc/missing'
autoreconf: Leaving directory `.'
```

Expected: `Warning: autoreconf failed.  Falling back on git`.

PR #874, opened 2017-06-01
Drop support for systems without autoconf

JasonGross:
Closes #868

See https://github.com/HoTT/HoTT/pull/851#issuecomment-305111736 and
surrounding comments.

While we're at it, also remove update_opam, since we no longer use a
custom version of Coq, and coq.hott no longer exists on opam.


spitters:
OK when Travis is happy.

andrejbauer:
I don't know about this. Autoconf was never supposed to be something that exists on the user side. It's a tool for developers to create portable installation scripts. This made sense before git existsed and people downloaded bundles of software that then installed itself.

How often does the configure script get changed nowadays? If not very often, we could perhaps include it in the repo?

mikeshulman:
My instinct would be that anyone cloning the git repo should be considered a "developer".  We should provide other ways for "users" to install the package, such as opam or a tarball including `configure`.

mikeshulman:
What's the status of this?

Alizter:
I'm in favour of merging this once the conflicts have been resolved.

PR #877, opened 2017-06-06
Factor out part of accessible lex modalities

mikeshulman:
It turns out that the main part of the characterization of accessible lex modalities in terms of modal-ness of the universe doesn't depend on accessibility and is an equivalent characterization of lex-ness for arbitrary modalities, namely that families of modal types indexed by a connected type are constant.

### DIFF
--- a/theories/Modalities/Lex.v
+++ b/theories/Modalities/Lex.v
@@ -226,6 +226,44 @@ Module Lex_Modalities_Theory (Os : Modalities).
     apply (equiv_isequiv (prod_unit_l B)).
   Defined.
 
+  (** 10. Families of modal types indexed by connected types are constant. *)
+  Definition modal_over_connected_isconst_lex (O : Modality) `{Lex O}
+             (A : Type) `{IsConnected O A} (P : A -> Type) `{forall x, In O (P x)}
+    : {Q : Type & In O Q * forall x, P x <~> Q}.
+  Proof.
+    exists (O {x:A & P x}); split; [ exact _ | intros x].
+    refine (BuildEquiv _ _ (fun p => to O _ (x ; p)) _).
+    refine (isequiv_conn_map_ino O _).
+    revert x.
+    apply conn_map_fiber.
+    refine (cancelL_conn_map O _ (fun z:{x:A & O {x : A & P x}} => z.2) _ _).
+    intros z.
+    refine (isconnected_equiv' O A _ _).
+    unfold hfiber.
+    refine (equiv_adjointify (fun x => ((x ; z) ; 1))
+                             (fun y => y.1.1) _ _). 
+    - intros [[x y] []]; reflexivity.
+    - intros x; reflexivity.
+  Defined.
+
+  (** And conversely. *)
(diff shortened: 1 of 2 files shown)


theories/Modalities/Modality.v
```diff
@@ -991,6 +991,15 @@ Section ConnectedMaps.
       + refine (to_O_natural _ _ _).
   Defined.
 
+  (** As a consequence, connected maps between modal types are equivalences. *)
+  Definition isequiv_conn_map_ino {A B : Type} (f : A -> B)
+         `{In O A} `{In O B} `{IsConnMap O _ _ f}
+    : IsEquiv f.
+  Proof.
+    refine (isequiv_commsq' f (O_functor O f) (to O A) (to O B)
```
JasonGross: Why `refine`, rather than just `Definition ... := ...`?

theories/Modalities/Lex.v
```diff
@@ -347,26 +385,12 @@ Module Accessible_Lex_Modalities_Theory
...
-      + refine (ispullback_connmap_mapino_commsq O
-                 ((fun w:sigT P => 1)
-                  : const tt o to O (sigT P) == const tt o pr1)).
-        (** All the necessary hypotheses are found by typeclass magic! *)
-      + refine (@isequiv_compose _ _
-                  (equiv_compose
-                    (equiv_prod_symm (O (sigT P)) (acc_gen O i))
-                    (equiv_pullback_unit_prod (O (sigT P)) (acc_gen O i)))
-                  _ _ (equiv_sigma_prod0 _ _)^-1 _).
+      (* The case [n=0] is basically just one of the above characterizations of lex-ness. *)
```
JasonGross: Any reason for this not to be a coqdoc comment?

theories/Modalities/Modality.v
mikeshulman: um, I didn't realize when I started the proof that it would be a one liner?  Will change.

theories/Modalities/Lex.v
mikeshulman: no, will fix.

theories/Modalities/Lex.v
```diff
@@ -226,6 +226,44 @@ Module Lex_Modalities_Theory (Os : Modalities).
...
+
+  (** And conversely. *)
+  Definition lex_from_modal_over_connected_isconst (O : Modality)
+             (H : forall (A : Type) (P : A -> Type),
+                 (IsConnected O A) -> (forall x, In O (P x)) ->
+                 {Q : Type & In O Q * forall x, P x <~> Q})
+    : Lex O.
+  Proof.
+    intros A x y ?.
+    apply isconnected_from_elim_to_O.
```
spitters: You might want to introduce H here, as you use the name later in the script.

spitters:
LGTM. I'll leave it to you to deal with the comment or not. It's OK to pull.

theories/Modalities/Lex.v
mikeshulman: H is named as a parameter of the definition.

PR #879, opened 2017-06-16
Rigid types and automorphisms of the universe

mikeshulman:
The part of https://homotopytypetheory.org/2017/01/26/parametricity-automorphisms-of-the-universe-and-excluded-middle/ having to do with automorphisms of the universe, plus a clever argument due to @peterlefanulumsdaine showing that we can swap `X*A` with `X*B` when `A` and `B` are rigid and connected while `X` is truncated.

### DIFF
--- a/theories/ExcludedMiddle.v
+++ b/theories/ExcludedMiddle.v
@@ -7,6 +7,8 @@ Existing Class ExcludedMiddle.
 
 Axiom LEM : forall `{ExcludedMiddle} (P : Type), IsHProp P -> P + ~P.
 
+Definition LEM_type := forall (P : Type), IsHProp P -> P + ~P.
+
 (** ** LEM means that all propositions are decidable *)
 
 Global Instance decidable_lem `{ExcludedMiddle} (P : Type) `{IsHProp P} : Decidable P
@@ -26,7 +28,7 @@ Defined.
 
 (** This direction requires Funext. *)
 Definition DNE_to_LEM `{Funext} : 
-  DNE -> forall (P : Type), IsHProp P -> P + ~P.
+  DNE -> LEM_type.
 Proof.
   intros dn P hp.
   refine (dn (P + ~P) _ _).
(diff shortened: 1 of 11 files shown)


mikeshulman:
Anyone willing to ack this?  We would like to get it merged so we can cite the HoTT library in the final version of https://arxiv.org/abs/1701.05617

mikeshulman:
Thanks!  I named `LEM_type` by analogy to `Funext_type` in `Basics/FunextVarieties`.  I don't think we have a standard convention on this yet, but we could declare one.  What do you think?

theories/ExcludedMiddle.v
```diff
@@ -0,0 +1,67 @@
...
+(** ** LEM means that all propositions are decidable *)
+
+Global Instance decidable_lem `{ExcludedMiddle} (P : Type) `{IsHProp P} : Decidable P
+  := LEM P _.
+
+(** ** Double-negation elimination *)
+
+Definition DNE := forall P, IsHProp P -> ~~P -> P.
+
+Definition LEM_to_DNE : ExcludedMiddle -> DNE.
```
mikeshulman: Because an argument of type `ExcludedMiddle` can be picked up by typeclass inference.

theories/Spaces/Universe.v
```diff
@@ -0,0 +1,219 @@
...
+  - intros q.
+    apply equiv_to_empty in q.
+    apply path_universe_uncurried in q.
+    rewrite <- eu in q.
+    apply ((ap f)^-1) in q.
+    exact (snd (equiv_hprop_idprod A P a) q).
+Defined.
+
+(** If you can derive a constructive taboo from an automorphism of the universe such that [g X <> X], then you get [X]-many beers. *)
+
```
spitters: Not sure about this inside joke here.

theories/Types/Universe.v
```diff
@@ -145,6 +147,28 @@ Proof.
...
+  : equiv_path (A * B) (A * C) (ap (prod A) (path_universe f))
+    = equiv_functor_prod_l f.
+Proof.
+  revert f. equiv_intro (equiv_path B C) f.
+  rewrite (eissect (equiv_path B C) f : path_universe (equiv_path B C f) = f).
+  destruct f.
+  apply path_equiv, path_arrow; intros x; reflexivity.
+Defined.
+
+(** [ap (fun Z => Z * A)] behaves like [equiv_functor_prod_l]. *)
```
spitters: should be _r in the comment

spitters:
LGTM. Feel free to merge after fixing the typo, and perhaps providing a reference to the beer joke.

JasonGross:
> I don't think we have a standard convention on this yet, but we could declare one. What do you think?

Yeah, I think the standard should perhaps be that for anything commonly taken as an axiom (univalence, funext, LEM, DNE, Pierce's law, AC), the type should be given by adding the `_type postfix.  This would require changing `DNE` to `DNE_type` here, and maybe adding a line to the style file here or in another PR.

mikeshulman:
Ok, made all the suggested fixes.  Will merge once Travis is happy.

PR #885, opened 2017-07-01
Universal property of the image of a function.

SimonBoulier:
This is not ready to merge yet (although it is a very simple PR).

This PR propose a few properties about the image of a function and the universal property of it.

There are at least two problems:

- I don't know well were to put the lemma `emb_hom_hprop` (I shall rename it by the way)

- But more importantly, I am stuck with a universe inconsistency line 120. Can anyone help me?

### DIFF
--- a/theories/Image.v
+++ b/theories/Image.v
@@ -1,0 +1,133 @@
+Require Import HoTT.Basics HoTT.Types HoTT.HIT.Truncations HoTT.Factorization.
+
+(** Universal property of the image as stated in:
+    "The join construction - Egbert Rijke - arXiv:1701.07538". *)
+
+
+
+(* Auxiliary lemma, should be but in another place? *)
+
+(** If f' is an embedding, the type of morphisms *)
+(** from f to f' is a mere proposition.          *)
+(**         f                                    *)
+(**   A ---------> B                             *)
+(**                ^                             *)
+(**                |                             *)
+(**                | f'                          *)
+(**                |                             *)
+(**                A'                            *)
+
+Instance emb_hom_hprop {Fs: Funext} {A A' B} (f : A -> B) (f' : A' -> B)
+           (Hf' : IsEmbedding f')
+  : IsHProp {g : A -> A' & f' o g == f}.
+Proof.
+  eapply trunc_equiv'. symmetry.
+  refine (equiv_sigT_coind (fun _ => A') (fun x y => f' y = f x)). cbn.
+  eapply trunc_forall.
+Defined.
+
+
+(** The image of a function f if { b : B & merely (hfiber f)}.
+    It is defined in the file HIT.Truncations, and before in Modalities.Modality
+    (for all truncation modalities). We only give here shorter names. *)
+Section Image.
+  Context {A B} (f : A -> B).
+
+  (* {b : B & merely (hfiber f b)} *)
+  Definition Im : Type := himage f.
+
+  Definition Im_q : A -> Im
+    := factor1 (himage f).
+
+  Definition Im_i : Im -> B
+    := factor2 (himage f).
+
+  Definition Im_Q : Im_i o Im_q == f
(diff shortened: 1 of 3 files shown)


mikeshulman:
For the first question, I would generalize it to a lemma about any reflective subuniverse and put it in `ReflectiveSubuniverse`.  In fact, shouldn't this result actually be a general theorem about factorization systems?

SkySkimmer:
Needs to be added to _CoqProject.

SimonBoulier:
I generalized the auxiliary lemma to ReflexiveSubuniverse.
Unfortunately it breaks the typeclass resolution (I have to manually apply it whereas it was automatically inferred before.)
Is there any way to avoid this?

mikeshulman:
This is a known problem with back-and-forth inferences between `IsTrunc n` and `InO (Tr n)`: only one direction can be automatically inferred in general, and we chose the other one.  Would it help to write a tactic that proves any goal of the form `IsTrunc n` by changing it to `InO (Tr n)` and then applying typeclass resolution?  At least that way the user wouldn't have to look up the name of the relevant lemma about `InO`.

JasonGross:
> This is a known problem with back-and-forth inferences

We might actually be able to fix this with `Hint Cut ` which I believe can tell the typeclasses resolution algorithm what a useless loop looks like, and have it end resolution there.

mikeshulman:
@JasonGross That would be cool!  Do you have time to try it?  I don't know the syntax of `Hint Cut`.  The instances in question are in `theories/HIT/Truncations.v`.

JasonGross:
@mikeshulman 
>  Do you have time to try it?

Apparently the answer is "not at the present moment".

SimonBoulier:
I solved the universe problem by defining an intermediate lemma.
Not very beautiful but does the job ...
Do you think that this PR is worth to be merged?

mikeshulman:
What would you think about proving it in the generality of a factorization system?  That seems to me the natural place to have it.

SimonBoulier:
@mikeshulman I'm not sure, but after several tries, it seems to me that this universal property works only for factorization systems in which the right maps are monomorphisms.
So I think it is good to keep only for the image.

SimonBoulier:
Feel free to say if you think that it is too specific for library.
(I feel it is not but I may be wrong.)

spitters:
LGTM

mikeshulman:
The diagonal fill-in property for any factorization system says that given a left-map `e:A->B` and a right-map `m:C->D`, the map from `B->C` to the type of commutative squares `exists f, exists g, m o f = g o e` that takes `h:B->C` to `h o e` and `m o h` with the identity homotopy in the square is an equivalence.  This map lies over `B->D`, and if we take its fiber over a particular map `g:B->D` then the resulting equivalence has domain `exists h:B->C, m o h = g` and codomain (using the identification of the fibers of a fibration) `exists f:A->C, m o f = g o e`.  If we take `g o e` to be the image factorization of some given map, this gives the universal property.

SimonBoulier:
Let correct me if I misunderstood you but you prove that `is_image(e,g) -> UP(e,g)` while what is difficult is the other direction. Do I missed something?

theories/Image.v
```diff
@@ -0,0 +1,114 @@
+Require Import HoTT.Basics HoTT.Types HoTT.HIT.Truncations HoTT.Factorization.
+
+(** Universal property of the image as stated in:
+    "The join construction - Egbert Rijke - arXiv:1701.07538". *)
+
+
+(** The image of a function f if { b : B & merely (hfiber f)}.
```
mattam82: typo if -> is

mikeshulman:
Well, any two things with the same universal property should be equivalent, right?  So if if the image has the universal property, anything else with the universal property should also be an image.

SimonBoulier:
I got it! Thanks.
I will formalize it.

PR #887, opened 2017-07-07
Use .v.d suffix for Coq dependency files.

SkySkimmer:
This way they can be ignored by file managers / emacs autocompletion/ etc without ignoring D language files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,7 +53,6 @@ coq/ide
 *.a
 *.cmxa
 # generated dependency files:
-*.d
 TAGS
 # ignore compiled python files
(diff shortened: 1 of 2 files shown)


spitters:
I don't have strong feelings about this. However, the use of the D language seems quite special, and .d is the standard abbreviation in Coq.

SkySkimmer:
>.d is the standard abbreviation in Coq.

What do you mean? Coq itself and coq_makefile use `.v.d`.

spitters:
Sorry, I've been using HoTT for too long :-)

LGTM

PR #890, opened 2017-07-10
Split surreals into multiple files in a directory (no content changes)

mikeshulman:
This is just a reorganization, in preparation for adding new content later.


spitters:
LGTM

JasonGross:
I'm not sure about the move of `No.v` to `No/No.v`; what's the reasoning behind that?  (I think that the absolute names are nicer without this move.)

mikeshulman:
@JasonGross what would you call it?  `No/Surreals.v`?  `No/Definition.v`?

JasonGross:
I would leave it as `Spaces/No.v`, and still move the other things into the `No/` directory.

mikeshulman:
Hmm, if there were a `Spaces/No.v` in addition to a `No/` directory, I would expect the former to simply import and re-export everything in the subdirectory.  To me a subdirectory is a grouping feature, and the basic definitions and encode-decode, etc. are all part of the "group" that lives inside the subdirectory.

JasonGross:
Ah, how about following the pattern of `Categories/`: name the file
`No/Core.v` (or `No/No.v` as you have it might even be a bit better), and
make `Spaces/No.v` with
```coq
Require HoTT.Spaces.No.No.
Require HoTT.Spaces.No.Addition.
Require HoTT.Spaces.No.Negation.

Include Spaces.No.No.
```

mikeshulman:
Why wouldn't it be

```
Require Export HoTT.Spaces.No.No.
Require Export HoTT.Spaces.No.Addition.
Require Export HoTT.Spaces.No.Negation.
```

?  I.e. why not export them, and why `Include` only the first one?

JasonGross:
Because `Include` makes the identifiers be accessible under `Spaces.No.*`, in case we ever have overlap with another `Module No`

JasonGross:
Re `Import ` vs `Export `, that's up to you, depending on what names you want to be accessible unqualified if you `Import No`

mikeshulman:
Ok, here you go.

theories/Spaces/No.v
```diff
@@ -0,0 +1,4 @@
+Require HoTT.Spaces.No.Core.
+Import HoTT.Spaces.No.Core.
```
JasonGross: Include, not Import?

mikeshulman:
Oops, fixed.

mikeshulman:
Is this ok now?

spitters:
Still good for me.

Issue #757, opened 2015-05-13
Including a simpler proof that NaiveFunext -> Funext?


JasonGross:
I'm designing a week-long class on homotopy type theory for Mathcamp, and in trying to simplify the material, noticed that the essential insight to the proofs of the following two types is exactly the same:
```coq
dec_to_UIP : forall {A} {x : A},
  (forall y, (x = y) + (x <> y)) -> (forall y (p q : x = y), p = q)

UIP_bump : forall {A} {x : A},
  (forall y, x = y) -> (forall y (p q : x = y), p = q)
```
This is, in some sense, because the proof that `IsTrunc n A -> IsTrunc n.+1 A` (which is another way of phrasing `UIP_bump`, I believe) is proved by instantiating an equality decider with the function that always returns "yes, they're equal".

When looking through the HoTT book for exercises, I noticed 2.16, that `NaiveFunext -> Funext`.  And the adjustment to the function given by `NaiveFunext` is exactly analogous to the adjustment in the proofs of the above two theorems (I guessed that this might be the case, and checked by reducing the term given by `NaiveFunext_implies_Funext`).  Inspired by this, I wrote a relatively short, fairly direct proof that `NaiveFunext -> Funext`:

```coq
Definition NaiveFunext_implies_Funext : NaiveFunext -> Funext_type.
Proof.
  intros path_forall A P f g; hnf in path_forall.
  refine (isequiv_adjointify _ _ _ _).
  { exact (fun H => (path_forall _ _ f f (fun _ => idpath))^ @ path_forall _ _ f g H). }
  { intro H.
    set (H' := fun x => (g x; H x)).
    change H with (fun x => (H' x).2).
    change g with (fun x => (H' x).1).
    clearbody H'; clear H g.
    assert (H'' : forall x, (f x; 1) = H' x).
    { intro x.
      change (H' x) with ((H' x).1; (H' x).2).
      destruct (H' x).2; reflexivity. }
    apply path_forall in H''.
    destruct H''; simpl.
    rewrite concat_Vp; reflexivity. }
  { intro H.
    destruct H.
    unfold apD10; simpl.
    rewrite concat_Vp.
    reflexivity. }
Defined.
```

[...]

PR #892, opened 2017-08-11
Some weaker forms of univalence that suffice to imply it

mikeshulman:


mikeshulman:
I didn't find it.  (But I did neglect to notice that `Univalence_type` was
already defined in `UnivalenceImpliesFunext`.  I removed the duplication.)

JasonGross:
Oh, I was thinking of https://github.com/HoTT/HoTT/issues/757#issuecomment-102128029 .  Anyway, still LGTM

PR #894, opened 2017-09-08
Import HoTTClasses as HoTT.Classes.

SkySkimmer:
This is without the files which use inductive-inductive types (ie the partiality monad and Cauchy and Dedekind reals).


mikeshulman:
Awesome!

My only comment is that `Classes/theory/jections` seems like it should be part of the main repository, somewhere around `HIT/Truncations`, `HIT/epi`, `HIT/quotient`, etc.

SkySkimmer:
Actually `surjective_factor` is a corollary of `conn_map_elim` (see 3910d5e87) so `jections` do not depend on HITs any more.

mikeshulman:
Well, it's certainly nicer to get it as a corollary, but you're still technically using HITs since truncations are only a modality in the presence of HITs.  `TrM` is the truncation modality that's constructed in `HoTT.HIT.Truncations`, which you're not importing directly into `jections` any more but you must be still picking it up by some other chain of imports.  So I think I would still be inclined to put this somewhere in the main `theories/HIT` directory.

(Actually, where is this file actually used elsewhere in HoTTClasses?  It looks like this last commit changed the type of the lemma surjective_factor_pr from an equality to a homotopy, but didn't need to change any other files, so apparently that lemma is not used anywhere else?)

SkySkimmer:
I see. I only looked at the `Print Assumptions surjective_factor_pr.` which says
``` coq
Axioms:
isequiv_apD10 : Funext -> forall (A : Type) (P : A -> Type) (f g : forall x : A, P x), IsEquiv apD10
Funext : Type0
```
`Truncations` is imported in `canonical_names`.

`surjective_factor` is used for Cauchy reals when defining multiplication and multiplicative inverse. That part of HoTTClasses is staying in its own repository since it depends on inductive-inductive types. See SkySkimmer/HoTTClasses@10641cf.

`surjective_factor` could go to its own file in HIT/ like how unique choice is done, WDYT?

mikeshulman:
Sure, that sounds fine.  It could also go in `HIT/epi.v`, since categorically the statement is that every surjection is a *regular* epi (the quotient of its kernel pair).  It might be nice to rename it more descriptively along those lines too.

spitters:
I've reviewed this code already when @SkySkimmer was writing it, so I could pull this PR, after @SkySkimmer implements @mikeshulman s suggestion.
However, perhaps @JasonGross want to chime in on this ?

SkySkimmer:
Moved it to HIT/, but I don't know about epis so renaming is beyond me.

mikeshulman:
Thanks!  Probably it should also be added to `HoTT.v`.

The travis error is `Package libgraphviz4 is not available, but is referred to by another package.`  Help someone?

spitters:
Maybe we should install https://packages.ubuntu.com/trusty/graphviz (no "4")
Why are we still using trusty (14.04 ) ? Maybe it's time to move to 16.04 LTS ?
@JasonGross  ?

PR #896, opened 2017-09-26
elaborate on example for prop. resizing for truncation

rmatthes:
Add an example that shows the benefit of propositional resizing for propositional truncation.

This is just for illustration - only two local definitions given.

I wanted to see what happens w/o resizing in this context, and my experiments are seen in a new sister file, see [my commit in another branch](https://github.com/rmatthes/HoTT/commit/6c38787d73fe0e21093753d0dc169dbe4f9081ce). Maybe that other part was obvious to you.

### DIFF
--- a/theories/PropResizing/Truncations.v
+++ b/theories/PropResizing/Truncations.v
@@ -47,24 +47,10 @@ Section AssumePropResizing.
 
 End AssumePropResizing.
 
-(* Here, we show what goes wrong without propositional resizing. We do not use what we imported with Require Import PropResizing.PropResizing. *)
-
-(* a more explicit definition of being a prop *)
-Local Definition IsHProp_expl (A : Type@{i}) : Type@{i} :=
-  forall x y : A, Contr_internal (x = y).
-
-Local Definition ishprop_expl_merely `{Funext} (A : Type@{i})
-  : IsHProp_expl@{i} (merely@{i j} A).
-Proof.
-  exact (@trunc_forall H Type (fun P : Type => IsHProp_expl P -> (A -> P) -> P)
-  (-1)
-  (fun P : Type =>
-   @trunc_forall H (IsHProp_expl P) (fun _ : IsHProp_expl P => (A -> P) -> P)
-     (-1) (fun p : IsHProp_expl P => @trunc_arrow H (A -> P) P (-1) p))).
-Defined.
+(* Here, we show what goes wrong without propositional resizing. *)
 
 (* the naive definition *)
-Local Definition merely_rec_naive {A : Type@{i}} {P : Type@{j}} {p:IsHProp_expl@{j} P} (f : A -> P)
+Local Definition merely_rec_naive {A : Type@{i}} {P : Type@{j}} {p:IsHProp@{j} P} (f : A -> P)
   : merely@{i j} A -> P
   := fun ma => ma P p f.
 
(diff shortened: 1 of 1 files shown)


mikeshulman:
This is a nice idea.  I think it would be clearer if it also included the example that fails without propresizing, since the one that succeeds looks so "obvious" (especially without universe annotations) that it's not clear that anything is happening.  But we don't want to spend a lot of time on it either; what about a few more local definitions after `AssumePropResizing` ends that define the naive version of `merely_rec` and verify that the corresponding `_equalargs` (which I would probably name something like `_sameuniv`, since it's not the arguments that are the same but rather their universes) indeed `Fail`s?

rmatthes:
@mikeshulman : Thanks for the encouragement. Please check if the current state of the PR is acceptable. (It is now meant to be pulled, before it was meant to be discussed.)

theories/PropResizing/Truncations.v
```diff
@@ -34,10 +34,58 @@ Section AssumePropResizing.
...
+    : typeofid (merely A) := functor_merely f.
+
+(* a more informative version using universe levels *)
+  Local Definition functor_merely_sameuniv `{Funext} {A B: Type@{i}} (f : A -> B)
+    : merely@{j k} A -> merely@{j k} B:= functor_merely f.    (* requires i <= j & k < j *)
+
 
 End AssumePropResizing.
+
+(* Here, we show what goes wrong without propositional resizing. We do not use what we imported with Require Import PropResizing.PropResizing. *)
```
mikeshulman: What do you mean by "we do not use what we imported"?  It's not importing something that makes the propositional resizing axiom available, it's assuming the axiom `PropResizing` as we did in the above section and as we are not doing here.

theories/PropResizing/Truncations.v
```diff
@@ -34,10 +34,58 @@ Section AssumePropResizing.
...
+
+(* a more explicit definition of being a prop *)
```
mikeshulman: I don't see what purpose this serves; can't we just use the ordinary `IsHProp`?

theories/PropResizing/Truncations.v
```diff
@@ -34,10 +34,58 @@ Section AssumePropResizing.
...
+
+(* the too weak definition *)
+Local Definition functor_merely_weak `{Funext} {A B : Type@{k}} (f : A -> B)
+  : merely@{j k} A -> merely@{k l} B. (* evidently, this requires k<j and l<k *)
+Proof.
+  srefine (merely_rec_naive (trm o f)).
+  apply ishprop_expl_merely.
+Defined.
+
+(* impossible due to universe inconsistency:
```
mikeshulman: I would uncomment this but put a `Fail` before it, so that its impossibility is actually verified by Coq.

theories/PropResizing/Truncations.v
```diff
@@ -34,10 +34,58 @@ Section AssumePropResizing.
...
-  Local Definition typeofid (A: Type) := A -> A.
+(* show what is gained by propositional resizing, without mentioning universe levels *)
+  Local Definition typeofid (A : Type) := A -> A.
    
-  Local Definition functor_merely_equalargs `{Funext} {A : Type} (f : typeofid A)
-    : typeofid(merely A) := functor_merely f.
+  Local Definition functor_merely_sameargs `{Funext} {A : Type} (f : typeofid A)
+    : typeofid (merely A) := functor_merely f.
+
+(* a more informative version using universe levels *)
```
mikeshulman: Do we need both the "with universe levels" and "without universe levels" versions?

rmatthes:
I tried to take into account the comments - thanks. Some of my Coq comments were overly cautious and are not necessary for the HoTT user. I would still like to maintain that already the version with the same arguments fails while the innocent-looking variant without type indices does work, and then show the annotated version with same universe of the arguments. I now agree that the use of the axiom would be sufficiently marked in the types so that I do not fear hidden uses of the axiom in the use of the typeclass mechanism for `ishprop_merely`.

mikeshulman:
Thanks!  LGTM.  Anyone else?

spitters:
Would any of this change with universe polymorphic inductive definitions, or is that a red herring?

mikeshulman:
Red herring.

SkySkimmer:
It wouldn't change.

Let `A : Type0` to simplify things. `merely@{j} A := forall P : Type@{j}, IsHProp P -> (A -> P) -> P` lives in `Type@{j+1}` so a term of type `merely@{j} A` cannot be applied to `merely@{j} A`. Cumulative inductive types would help only if the problem appeared at the `IsHProp` argument.

PR #898, opened 2017-10-17
Backward compatible fix w.r.t. changes in primitives projections

ppedrot:
This patch makes HoTT compile after the changes in primitive projections introduced in PR coq/coq#984. It should be backward compatible with Coq 8.7.0.

### DIFF
--- a/theories/Categories/ChainCategory.v
+++ b/theories/Categories/ChainCategory.v
@@ -81,7 +81,7 @@ Module Export Univalent.
   Proof.
     intros s d.
     refine (isequiv_iff_hprop _ _).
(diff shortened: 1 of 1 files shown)


mikeshulman:
Looks fine to me.  Does this have to do with an actual use of canonical structures in HoTT, or is it a side effect of the change for that purpose?

spitters:
I guess this breaks HoTT on pre-8.7.0 ?
Otherwise, LGTM.

SkySkimmer:
Before the Coq change the refine solved the goal so the added tactics don't break on previous Coq. See travis for confirmation.

Issue #895, opened 2017-09-12
Program tactic and equations


spitters:
It would be nice to have Program for HoTT, e.g. for use in HoTT-Classes.
I keep on forgetting what we have done already.

There was a discussion here:
https://github.com/HoTT/HoTT/commit/6a99db1c31fe267fdef7be755fa169fb6a87e3cf

However, it would be really nice to combine this with the [Equations plugin](https://github.com/mattam82/Coq-Equations/commits/master) by @mattam82 and @cmangin.

spitters:
https://mattam82.github.io/Coq-Equations/

mattam82:
Note that Equations currently cannot be used directly with the HoTT library as it compiles against the standard library but can surely be made independent.

spitters:
Do you or @cmangin (or @SkySkimmer for that matter) have any plans of doing that in the future. HoTT seems to be an ideal application for Equations.

mattam82:
I’ve tried a little bit and it’s mostly an engineering issue, so any help
would be appreciated!
Le jeu. 16 nov. 2017 à 08:20, Bas Spitters <notifications@github.com> a
écrit :

spitters:
Recording an email discussion with @mattam82 

 the main issue is binding of the constants used by Equations,
and probably a bit of fidling with the current Coq proofs the theories/ 
library which might use tactics that are broken with an equality in Type
(though unlikely now). The binding is mostly straightforward, for general
constants everything is done in src/equations_common.ml

let coq_unit = lazy (init_reference ["Coq";"Init";"Datatypes"] "unit")
let coq_tt = lazy (init_reference ["Coq";"Init";"Datatypes"] "tt")
let coq_Empty = lazy (init_reference ["Coq";"Init";"Datatypes"] "Empty_set")
...
https://github.com/mattam82/Coq-Equations/blob/master/src/equations_common.ml

However, this are currently unavailable in:
https://github.com/HoTT/HoTT/blob/master/coq/theories/Init/Datatypes.v
They are here:
https://github.com/HoTT/HoTT/blob/master/theories/Basics/Overture.v

An issue is *maintainability*. Should we have a  _binding_ in a .v file like Agda's PRAGMA BUILTIN or Equations Logic command?

I wonder how this is treated by @SkySkimmer in his sProp work.

SkySkimmer:
What does this have to do with SProp?

spitters:
I meant: you are also introducing a new class of propositions.
[...]

PR #901, opened 2017-10-22
contrib/HoTTBookExercises: Solutions to several exercises

siddharthist:
The full solution to 1.8 is very tedious, I've only completed part of it myself. 

I'm not sure if my solution properly reflects what 2.1 was asking, especially whether or not I was supposed to use functional extensionality.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -155,13 +155,10 @@ Definition add : nat -> nat -> nat :=
 Definition mult : nat -> nat -> nat  :=
   rec_nat' (nat -> nat) (fun m => 0) (fun n g m => add m (g m)).
 
-Definition exp' : nat -> nat -> nat  :=
-  rec_nat' (nat -> nat) (fun m => (S 0)) (fun n g m => mult m (g m)).
-
-Definition flip {A B C : Type} (f : A -> B -> C) : (B -> A -> C) :=
-  fun b a => f a b.
-
-Definition exp := flip exp'.
+(* rec_nat' gives back a function with the wrong argument order, so we flip the
+   order of the arguments p and q *)
+Definition exp : nat -> nat -> nat  :=
+  fun p q => (rec_nat' (nat -> nat) (fun m => (S 0)) (fun n g m => mult m (g m))) q p.
 
 Example add_example: add 32 17 = 49. reflexivity. Defined.
 Example mult_example: mult 20 5 = 100. reflexivity. Defined.
@@ -246,24 +243,24 @@ Definition Book_2_1_concatenation3 :
   exact x_eq_y.
 Defined.
 
-Local Notation "p ⬛1 q" := (Book_2_1_concatenation1 p q) (at level 10).
-Local Notation "p ⬛2 q" := (Book_2_1_concatenation2 p q) (at level 10).
-Local Notation "p ⬛3 q" := (Book_2_1_concatenation3 p q) (at level 10).
+Local Notation "p *1 q" := (Book_2_1_concatenation1 p q) (at level 10).
+Local Notation "p *2 q" := (Book_2_1_concatenation2 p q) (at level 10).
+Local Notation "p *3 q" := (Book_2_1_concatenation3 p q) (at level 10).
 
 Section Book_2_1_Proofs_Are_Equal.
   Context {A : Type} {x y z : A}.
   Variable (p : x = y) (q : y = z).
-  Definition Book_2_1_concatenation1_eq_Book_2_1_concatenation2 : p ⬛1 q = p ⬛2 q.
+  Definition Book_2_1_concatenation1_eq_Book_2_1_concatenation2 : p *1 q = p *2 q.
     induction p, q.
     reflexivity.
   Defined.
 
-  Definition Book_2_1_concatenation2_eq_Book_2_1_concatenation3 : p ⬛2 q =  p ⬛3 q.
+  Definition Book_2_1_concatenation2_eq_Book_2_1_concatenation3 : p *2 q =  p *3 q.
     induction p, q.
     reflexivity.
   Defined.
 
-  Definition Book_2_1_concatenation1_eq_Book_2_1_concatenation3 : p ⬛1 q = p ⬛3 q.
+  Definition Book_2_1_concatenation1_eq_Book_2_1_concatenation3 : p *1 q = p *3 q.
     induction p, q.
     reflexivity.
   Defined.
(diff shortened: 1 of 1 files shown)


siddharthist:
Is anybody interested in reviewing these? I've been updating this with more and more solutions, but I don't want to make this too big if there's no hope of it getting merged.

mikeshulman:
Sorry, I do want to review this but haven't had time yet.

siddharthist:
@mikeshulman No worries, just wanted to make sure I wasn't wasting effort :) would you prefer I add solutions as I write them up, or leave this in the state it's in until you have time to review it?

contrib/HoTTBookExercises.v
```diff
@@ -143,7 +143,31 @@ End Book_1_3_sig.
...
+Definition mult : nat -> nat -> nat  :=
+  rec_nat' (nat -> nat) (fun m => 0) (fun n g m => add m (g m)).
+
+Definition exp' : nat -> nat -> nat  :=
+  rec_nat' (nat -> nat) (fun m => (S 0)) (fun n g m => mult m (g m)).
+
+Definition flip {A B C : Type} (f : A -> B -> C) : (B -> A -> C) :=
+  fun b a => f a b.
 
+Definition exp := flip exp'.
```
mikeshulman: I would probably write `Definition exp (n m : nat) : nat := rec_nat' ... m n` to avoid the need for `flip`.

contrib/HoTTBookExercises.v
```diff
@@ -199,17 +223,80 @@ Definition Book_1_10 := ack.
...
+Defined.
 
+Definition Book_2_1_concatenation3 :
+    forall {A : Type} {x y z : A}, x = y -> y = z -> x = z.
+  intros A x y z x_eq_y y_eq_z.
+  induction y_eq_z.
+  exact x_eq_y.
+Defined.
+
+Local Notation "p ⬛1 q" := (Book_2_1_concatenation1 p q) (at level 10).
```
mikeshulman: Could you do this without unicode?  Unicode is avoided in the core library, which this isn't really part of, but it also seems not really necessary here.

contrib/HoTTBookExercises.v
```diff
@@ -219,12 +306,59 @@ Definition Book_1_10 := ack.
...
+  fun q => p @ q.
+
+Definition concat_right {A : Type} {x y : A} (z : A) (p : x = y)
+    : (x = z) -> (y = z) :=
+  fun q => (inverse p) @ q.
+
+(* Again, by induction on p, it suffices to assume that x ≡ y and p ≡ refl, so
+   the above equations concatenate identity paths, which are units under 
+   concatenation. *)
+Definition Book_2_6 {A : Type} {x y z : A} (p : x = y)
```
mikeshulman: This exercise is just `HoTT.Types.Paths.isequiv_concat_l`.  I'm not sure we have an official policy regarding when exercises are already in the main library.  I can see an argument that having explicit simple proofs here is helpful for people reading the book, but it may also be helpful to point users of the library to where the result can actually be found.  Maybe just add a comment with the pointer?

contrib/HoTTBookExercises.v
```diff
@@ -239,17 +373,62 @@ Definition Book_1_10 := ack.
...
+Definition coprod_ump1 {A B X} : (A + B -> X) -> (A -> X) * (B -> X) :=
+  fun f => (f o inl, f o inr).
+
+(* To create a function on the direct sum from functions on the summands, work
+   by cases *)
+Check prod_rect.
+Definition coprod_ump2 {A B X} : (A -> X) * (B -> X) -> (A + B -> X) :=
+  prod_rect (fun _ => A + B -> X) (fun f g => sum_rect (fun _ => X) f g).
+
+Definition Book_2_9 {A B X} `{Funext} : (A -> X) * (B -> X) <~> (A + B -> X).
```
mikeshulman: `HoTT.Types.Sum.equiv_sum_ind`.

contrib/HoTTBookExercises.v
```diff
@@ -239,17 +373,62 @@ Definition Book_1_10 := ack.
...
+  apply (equiv_adjointify coprod_ump2 coprod_ump1).
+  - intros f.
+    apply path_forall.
+    intros [a | b]; reflexivity.
+  - intros [f g].
+    reflexivity.
+Defined.
 
 (* ================================================== ex:sigma-assoc *)
 (** Exercise 2.10 *)
```
mikeshulman: `HoTT.Types.Sigma.equiv_sigma_assoc`

contrib/HoTTBookExercises.v
```diff
@@ -239,17 +373,62 @@ Definition Book_1_10 := ack.
...
+  Defined.
+End TwoTen.
 
 (* ================================================== ex:pullback *)
 (** Exercise 2.11 *)
 
-
+(* This library actually defines pullbacks *)
+Definition Book_2_11 {A B C : Type} (f : A -> C) (g : B -> C)
+  : IsPullback (pullback_commsq f g).
```
mikeshulman: This is not actually what the exercise says!  The library definition of `IsPullback` says essentially that a square "is a pullback" if it is equivalent to "the" pullback, which of course is tautologically true of "the" pullback; whereas the exercise is to prove that "the" pullback in fact has the universal property of a pullback.

siddharthist: Thanks for pointing that out :) I was a little confused here (obviously), and should have spent more time considering what exactly the library code said.

siddharthist:
I've now made the changes requested, except I haven't completed the proof of the UMP of the pullback. If you think partial solutions are still helpful, we could include it, but I'm also happy to remove it until I have more time to think about and complete it.

I think you're right that it's helpful to have simple, straightforward proofs from the book for the exercises, even when they are given in the library. I agree that we should reference the "official" locations, to help people transition from learning from the book to making contributions!

Thanks for reviewing, it helps me make sure I'm on the right track as well 😄

mikeshulman:
I think I'd prefer not to have partial solutions (but if anyone else disagrees, please speak up).

siddharthist:
@mikeshulman Removed!

mikeshulman:
Thanks, LGTM.  APOE?

siddharthist:
Sorry, what do you mean by "APOE"?

Thanks again for working with me through the review! I'll have solutions for some chapter 3 exercises soon.

mikeshulman:
LGTM = Looks Good To Me
APOE = Another Pair Of Eyes -- two "pairs of eyes" are required to approve a pull request before merging it.

spitters:
Thanks. LGTM.

BTW, I don't know how difficult it is to get this to work, but Equations is more mature now.
Using pattern matching would probably give one nicer proofs.
https://github.com/HoTT/HoTT/issues/895

PR #903, opened 2017-10-27
ProofGeneral variable for Coq

tymmym:
Changing `proof-prog-name` has no effect for me. Maybe we should use `coq-prog-name`?

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -85,7 +85,7 @@ minimal fuss, you should try your luck by following these instructions:
 
 5. If you use ProofGeneral (PG), do not forget to change the name of the Coq
    program called by PG. For instance you can edit the name of the executable
-   called by PG by typing `M-x customize-variable`, then `proof-prog-name` which
+   called by PG by typing `M-x customize-variable`, then `coq-prog-name` which
    displays a customization utility. Another option is to type `M-x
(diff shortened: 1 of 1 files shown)


mikeshulman:
Yes, it looks like `coq-prog-name` is used by coq PG to set `proof-prog-name`.  I'm not sure why this is, but I agree it's `coq-prog-name` we should be telling the user to set.

However, we also have a `.dir-locals.el` file that's supposed to automatically set `coq-prog-name` correctly.  Does it not work for you?

tymmym:
I assume it works fine for browsing existing theories. But in my case after installation I created a file and tried `Require Import HoTT`.

mikeshulman:
Creating a new file in `theories/` it works fine for me; a dir-locals file should also apply to new files created in a directory.

tymmym:
I created a file outside HoTT repository.

mikeshulman:
Oh, ok.  Was setting `coq-prog-name` sufficient to make it work?

mikeshulman:
It might be good to also mention in these instructions that this should only be necessary if you're outside the repo directory.

tymmym:
Yes, it was sufficient.

mikeshulman:
Thanks!  LGTM.

PR #904, opened 2017-10-30
Add notations (x ; y ; z) to ( x ; y ; z ; t ; u ; v )

SimonBoulier:
This PR propose to add notations for nested pairs, a feature I often wanted.

Unfortunately we cannot define a recursive notation like this one:
` Notation "( x ; .. ; y ; z )" := (existT _ x .. (existT _ y z) .. ) : fibration_scope. `
because of bug https://github.com/coq/coq/issues/6032.

I removed a few now unnecessary parentheses to test.
You can find them with `grep -R "(.*; (.*; .*))" *` (there are some false positive) ; there are not so much in the library.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -144,7 +144,12 @@ Open Scope core_scope.
 Definition const {A B} (b : B) := fun x : A => b.
 
 (** We define notation for dependent pairs because it is too annoying to write and see [existT P x y] all the time.  However, we put it in its own scope, because sometimes it is necessary to give the particular dependent type, so we'd like to be able to turn off this notation selectively. *)
+(* We define notations for nested pairs as we can't define a recursive notations; see https://github.com/coq/coq/issues/6032 (should be solved in Coq 8.8). *)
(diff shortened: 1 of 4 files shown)


mikeshulman:
What is different about this recursive notation from other recursive notations that invokes the bug?

spitters:
I'm OK with using this as a workaround.

SimonBoulier:
@mikeshulman I don't use a recursive notation but repeat the pattern for n=3 to 6 ...

mikeshulman:
@SimonBoulier right, I got that -- I'm asking why you need to do that workaround?  Recursive notations are a standard Coq feature used in other places; why don't they work here?

SimonBoulier:
Not so standard apparently...
So, it is because of the mentioned bug: you can define them but then, when you try to use them it fails.
As far as I understand, it is because it associates on the left instead of the right

mikeshulman:
Oh!  I was misled by their presence in the reference manual...

mikeshulman:
LGTM once Travis is happy.

PR #907, opened 2017-11-20
Explicitate Existing Instance for Axioms (coq/coq#6183).

SkySkimmer:
This should be backwards compatible (with previous Coq ignoring the unknown option).

### DIFF
--- a/theories/Modalities/Lex.v
+++ b/theories/Modalities/Lex.v
@@ -289,6 +289,7 @@ Module Type Preserves_Fibers (Os : ReflectiveSubuniverses).
   Parameter isequiv_O_functor_hfiber :
      forall (O : ReflectiveSubuniverse) {A B} (f : A -> B) (b : B),
        IsEquiv (O_functor_hfiber O f b).
+  Existing Instance isequiv_O_functor_hfiber.
 
 End Preserves_Fibers.
 
--- a/theories/Classes/implementations/assume_rationals.v
+++ b/theories/Classes/implementations/assume_rationals.v
@@ -18,3 +18,7 @@ Parameters (Q : Type@{UQ}) (Qap : Apart@{UQ UQ} Q)
 (* I don't even want to know why this is necessary. *)
 Parameter Qenum : Enumerable Q.
 Notation "Q+" := (Qpos Q).
+
+Existing Instances Qap Qplus Qmult Qzero Qone Qneg Qrecip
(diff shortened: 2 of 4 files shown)


SkySkimmer:
Not sure if that PR is going to be merged. If not the Existing Instance calls are useful to make things clearer but not the Local Set call.

theories/Modalities/Lex.v
```diff
@@ -289,6 +289,7 @@ Module Type Preserves_Fibers (Os : ReflectiveSubuniverses).
   Parameter isequiv_O_functor_hfiber :
      forall (O : ReflectiveSubuniverse) {A B} (f : A -> B) (b : B),
        IsEquiv (O_functor_hfiber O f b).
+  Existing Instance isequiv_O_functor_hfiber.
```
SkySkimmer: This could instead go in any functors with arguments of this module type, but this change is closer to what Coq does automatically.

mikeshulman:
What does the new option do?

SkySkimmer:
It reverts to <=8.7 behaviour of declaring instances from every axiom.

mikeshulman:
Sorry, then I should have asked "what does Coq now do without the new option"?

SkySkimmer:
It doesn't declare instances for axioms, just like axioms from doing `Admitted` (except for axioms from Variable/Context and aliases outside sections).

ie
```coq
(* Coq 8.7 *)
Class Foo := {}.
Lemma foo : Foo. Admitted. Fail Definition bar : Foo := _.
Axiom foo' : Foo. (* or Variable foo' : Foo. *) Definition bar : Foo := _. (* foo' *)
```
```coq
(* Coq PR version without option *)
Class Foo := {}.
Lemma foo : Foo. Admitted. Fail Definition bar : Foo := _.
Axiom foo' : Foo. Fail Definition bar : Foo := _.
Variable foo'' : Foo. Definition bar : Foo := _. (* foo'' *)
```

mikeshulman:
I don't think I realized that Coq used to automatically declare all axioms as instances.  Wouldn't it be better to stick to the new behavior and just add explicit instance declarations for all the axioms we want as instances?

SkySkimmer:
>Wouldn't it be better to stick to the new behavior and just add explicit instance declarations for all the axioms we want as instances?

Sure why not. I was only putting the option in assume_rationals since it's just a big axiom block.

mikeshulman:
Thanks.  LGTM.

PR #908, opened 2017-11-22
contrib/exercises: solution to npaths

siddharthist:
Another solution 😄


siddharthist:
@JasonGross Thanks for the review! There are a few other exercises with multiple parts that follow the `Book_X_Y_*` naming convention where `*` is some specific part of the exercise, so I've done that here.

PR #910, opened 2017-11-26
contrib/exercises: ¬ (forall A, ∥ A ∥ → A)

siddharthist:
Another solution (with the right naming scheme this time ;). Any suggestions for simplifying the proof are more than welcome.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -629,7 +629,74 @@ Defined.
 (* ================================================== ex:not-brck-A-impl-A *)
 (** Exercise 3.11 *)
 
+(* This theorem extracts the main idea leading to the contradiction constructed
+   in the proof of Theorem 3.2.2, that univalence implies that all functions are
+   natural with respect to equivalences.
 
+   The terms are complicated, but it pretty much follows the proof in the book,
+   step by step.
+ *)
+Lemma univalence_func_natural_equiv `{Univalence}
+  : forall (C : Type -> Type) (all_contr : forall A, Contr (C A -> C A))
+      (g : forall A, C A -> A) {A : Type}  (e : A <~> A),
+    e o (g A) = (g A).
+  intros C all_contr g A e.
+  apply path_forall.
+
+  intros x.
+  pose (p := path_universe_uncurried e).
+
+  (* The propositional computation rule for univalence of section 2.10 *)
+  refine (concat (happly (transport_idmap_path_universe_uncurried e)^ (g A x)) _).
+
+  (* To obtain the situation of 2.9.4, we rewrite x using
+
+       x = transport (fun A : Type => C A) p^ x
+
+     This equality holds because (C A) -> (C A) is contractible, so
+
+       transport (fun A : Type => C A) p^ = idmap
+
+     In both Theorem 3.2.2 and the following result, the hypothesis
+     Contr ((C A) -> (C A)) will follow from the contractibility of (C A).
+   *)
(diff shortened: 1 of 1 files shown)


contrib/HoTTBookExercises.v
```diff
@@ -629,7 +629,74 @@ Defined.
 (* ================================================== ex:not-brck-A-impl-A *)
 (** Exercise 3.11 *)
 
+(* This theorem extracts the main idea leading to the contradiction constructed
```
JasonGross: Please start comments with `(**`, so that they will show up in [the wiki](http://hott.github.io/HoTT/coqdoc-html/HoTTBookExercises.html).

contrib/HoTTBookExercises.v
```diff
@@ -629,7 +629,74 @@ Defined.
...
+   natural with respect to equivalences.
 
+   The terms are complicated, but it pretty much follows the proof in the book,
+   step by step.
+ *)
+Lemma univalence_func_natural_equiv `{Univalence}
+  : forall (C : Type -> Type) (all_contr : forall A, Contr (C A -> C A))
+      (g : forall A, C A -> A) {A : Type}  (e : A <~> A),
+    e o (g A) = (g A).
+  intros C all_contr g A e.
```
JasonGross: Is there any reason to not begin with `Proof.`?

contrib/HoTTBookExercises.v
```diff
@@ -629,7 +629,74 @@ Defined.
...
+  (* To obtain the situation of 2.9.4, we rewrite x using
+
+       x = transport (fun A : Type => C A) p^ x
+
+     This equality holds because (C A) -> (C A) is contractible, so
+
+       transport (fun A : Type => C A) p^ = idmap
+
+     In both Theorem 3.2.2 and the following result, the hypothesis
+     Contr ((C A) -> (C A)) will follow from the contractibility of (C A).
```
JasonGross: If you want this to appear pretty in coqdoc (which I'd suggest), use square brackets to enclose terms as in `[Contr ((C A) -> (C A))]`

contrib/HoTTBookExercises.v
```diff
@@ -629,7 +629,74 @@ Defined.
...
+  (* The propositional computation rule for univalence of section 2.10 *)
+  refine (concat (happly (transport_idmap_path_universe_uncurried e)^ (g A x)) _).
+
+  (* To obtain the situation of 2.9.4, we rewrite x using
+
+       x = transport (fun A : Type => C A) p^ x
+
+     This equality holds because (C A) -> (C A) is contractible, so
+
+       transport (fun A : Type => C A) p^ = idmap
```
JasonGross: If you want this to appear pretty in coqdoc, you can either use square brackets or use `<<<` as in
```coq
<<<
         transport (fun A : Type => C A) p^ = idmap
>>>
```

mikeshulman:
Also LGTM.

Regarding placement, I would say that the solution to an exercise should be in the main library if there's a chance that someone might want to use that result when doing something else; no one should ever have to `Import HoTTBookExercises`.  For purely negative results like this one and #923, that seems somewhat unlikely, so I lean towards not cluttering up the main library with them; but if anyone wanted to make a case that they do belong in the main library, I could certainly be convinced.

mikeshulman:
@siddharthist are you planning to make the stylistic changes that @JasonGross suggested?

siddharthist:
Done, thanks for the reminder and review @mikeshulman!

PR #906, opened 2017-11-19
Coq now prunes phantom universes.

SkySkimmer:
(after coq/coq#1033)
This should be backwards compatible, which is why there are all these ugly `ltac:(first [exact foo@{bla}|exact foo@{bli}])` things.

PR #911, opened 2017-12-27
Fix compilation in the check-poly-effects branch.

ppedrot:
This patch fixes the compilation of HoTT with the soon-to-be master branch of Coq. Indeed, PR coq/coq#6483 suppresses spurious universes generated by monomorphic effects in polymorphic proofs. Unluckily, it is not backward-compatible, although if desired I can use the same kind of hacks as in PR #906 to work around this.

### DIFF
--- a/theories/Classes/theory/premetric.v
+++ b/theories/Classes/theory/premetric.v
@@ -95,7 +95,8 @@ apply le_equiv_lt in E. destruct E as [E|E].
   rewrite E'. apply rounded_plus. trivial.
 Qed.
 
-Definition rounded_le@{i j} := @rounded_le'@{j i Ularge j}.
+(* Coq pre 8.8 produces phantom universes, see coq/coq#6483 **)
+Definition rounded_le@{i j} := ltac:(first [exact @rounded_le'@{j i Ularge}|exact @rounded_le'@{j i Ularge j}]).
 Arguments rounded_le {A _ _} e u v _ d _.
 
 Section close_prod.
@@ -160,7 +161,8 @@ intros e u v. split.
   rewrite E1;split;apply rounded_plus,E2.
 Qed.
 
-Definition close_prod_rounded@{j} := @close_prod_rounded'@{j j j j j}.
+(* Coq pre 8.8 produces phantom universes, see coq/coq#6483 **)
+Definition close_prod_rounded@{j} := ltac:(first [exact @close_prod_rounded'@{j j j j j}|exact @close_prod_rounded'@{j j}]).
 Arguments close_prod_rounded {_ _} _ _ _.
 Global Existing Instance close_prod_rounded.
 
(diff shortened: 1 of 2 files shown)


spitters:
Thanks!
I think, in general, we'd like to stay backwards compatible for a while. So, ugly temporary hacks might be better.

ppedrot:
So, I easily fixed the difference in quantification length by using @SkySkimmer's tricks, but I don't know how to handle properly the change in applied universes. Essentially, I need a way to perform `x@{i}` in 8.7 and `x@{j}` in 8.8. Is there any similar (potentially hackish) trick that one could use to discriminate in Ltac between those two versions?

theories/Modalities/Accessible.v
```diff
@@ -267,15 +267,15 @@ Module Accessible_Modalities_from_ReflectiveSubuniverses
         refine (cancelL_ooextendable@{a a a i i i i i i}
                   (fun _ => X) (Acc.acc_gen O i)
                   (to O (lgen_codomain (Acc.acc_gen O) i)) _ _).
-        + apply ooextendable_isnull_fibers@{a a i i a a i a a a}; intros x.
+        + apply ooextendable_isnull_fibers@{a a i i a a i a a i}; intros x.
```
SkySkimmer: Using `i` regardless of Coq version might just work.
Otherwise you need to lift the universe constraints for the module typechecking to put them before the definition, so that the failure is early. Since 8.7 doesn't have inline constraints that means using sections, as:
```coq
Section Sec.
Universes u a i.
Constraint a < u.
Constraint a <= i.

Definition inO_iff_isnull@{}  ...
... first [...] ...

End Sec.
```

ppedrot:
It looks like @SkySkimmer was right, the current PR now compiles with both versions.

theories/Classes/theory/premetric.v
```diff
@@ -95,7 +95,8 @@ apply le_equiv_lt in E. destruct E as [E|E].
   rewrite E'. apply rounded_plus. trivial.
 Qed.
 
-Definition rounded_le@{i j} := @rounded_le'@{j i Ularge j}.
+(* Coq pre 8.8 produces phantom universes, see coq/coq#6483 **)
```
SkySkimmer: Are they really phantom (= not appearing in the term)?

ppedrot: I didn't check in detail, but I'd think they are, they're spuriously generated by internal monomorphic side-effects.

spitters:
Travis was unhappy about coq-dpdgraph. I restarted it. If it keeps being unhappy, perhaps @JasonGross can have a look.

spitters:
Thanks. I think we can pull when Travis is happy.

ppedrot:
dpdgraph is broken because of unrelated changes, namely the removal of the API.

spitters:
What would be the best way to proceed? Should we disable pdpgraph in travis?
Is there a separate Coq issue for that?

ppedrot:
@spitters It looks people are discussing the dpdgraph issue on coq/coq#6493. Anyways, the current patch has nothing to do with that issue and ought to be merged ASAP for coq/coq#6483 to be in turn merged into Coq.

spitters:
Thanks! The pdpgraph issue is annoying is it will be harder to check travis :-(

Zimmi48:
The fix for dpd-graph was merged: https://github.com/Karmaki/coq-dpdgraph/pull/37

PR #919, opened 2018-02-26
Binary naturals

abooij:
I have implemented a variant on binary naturals, that I learned from Martín Escardó and Cory Knapp, given directly as one inductively defined type.
```coq
Inductive binnat : Type :=
| bzero   :           binnat  (* zero *)
| double1 : binnat -> binnat  (* 2n+1 *)
| double2 : binnat -> binnat. (* 2n+2 *)
```
This allows doing computations with larger naturals than the unary presentation.

I have placed the code in `Classes/implementations`. Feedback is very welcome as I don't consider myself a Coq expert at all.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -394,6 +394,7 @@ theories/Classes/orders/nat_int.v
 theories/Classes/orders/integers.v
 theories/Classes/implementations/assume_rationals.v
 theories/Classes/implementations/peano_naturals.v
+theories/Classes/implementations/binary_naturals.v
 theories/Classes/implementations/natpair_integers.v
 theories/Classes/implementations/field_of_fractions.v
 theories/Classes/implementations/list.v
--- a/theories/Basics/Decidable.v
+++ b/theories/Basics/Decidable.v
@@ -12,6 +12,21 @@ Class Decidable (A : Type) :=
   dec : A + (~ A).
 Arguments dec A {_}.
 
+(** The [decide_type] and [decide] tactic allow to automatically prove
+decidable claims using previously written decision procedures that
+compute. *)
+Ltac decide_type A :=
+  let K := (eval hnf in (dec A)) in
+  match K with
+  | inl ?Z => exact Z
+  | inr ?Z => exact Z
+  end.
+
+Ltac decide :=
+  match goal with
+  | [|- ?A] => decide_type A
+  end.
+
 Class DecidablePaths (A : Type) :=
   dec_paths : forall (x y : A), Decidable (x = y).
 Global Existing Instance dec_paths.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -428,6 +428,15 @@ Ltac equiv_intro E x :=
       refine (equiv_ind E Q _); intros x
   end.
 
+(** The same, but for several variables. *)
+
+Tactic Notation "equiv_intros" constr(E) ident(x)
+  := equiv_intro E x.
+Tactic Notation "equiv_intros" constr(E) ident(x) ident(y)
+  := equiv_intro E x; equiv_intro E y.
+Tactic Notation "equiv_intros" constr(E) ident(x) ident(y) ident(z)
+  := equiv_intro E x; equiv_intro E y; equiv_intro E z.
+
 (** [equiv_composeR'], a flipped version of [equiv_compose'], is (like [concatR]) most often useful partially applied, to give the “first half” of an equivalence one is constructing and leave the rest as a subgoal. One could similarly define [equiv_composeR] as a flip of [equiv_compose], but it doesn’t seem so useful since it doesn’t leave the remaining equivalence as a subgoal. *)
 Definition equiv_composeR' {A B C} (f : A <~> B) (g : B <~> C)
   := equiv_compose' g f.
--- a/theories/Classes/implementations/binary_naturals.v
+++ b/theories/Classes/implementations/binary_naturals.v
@@ -15,6 +15,9 @@ Require Import
         HoTT.Classes.orders.semirings
         HoTT.Classes.theory.apartness.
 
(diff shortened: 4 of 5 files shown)


spitters:
Thanks. This looks nice.
I have the impression that you may be able to obtain the semiring axioms from univalence, as we already have them for the unaries (I hope).

SkySkimmer:
It should be doable using stuff from Classes/isomorphisms/rings.v

Gaëtan Gilbert

abooij:
@spitters: Indeed I proved the semiring axioms from the fact that the unary naturals satisfy them. See e.g. `binnat_add_assoc`, `binnat_distr_l`, `binnat_full` and `binnat_toR_unique`. However, some things I proved directly since they are computationally relevant, such as the semiring operations, `binnat_dec : DecidablePaths binnat`, and `binnat_trichotomy : Trichotomy (lt : Lt binnat)` (even though the latter are propositions).

mikeshulman:
Interesting!  If we had a computing univalence like in cubical TT, of course, we could use univalence everywhere.  How did you decide which hprops are "computationally relevant" and which are not?

mikeshulman:
Also, a question to someone else: is Travis just permanently useless now?  Can we fix it somehow so that it actually tells us whether a PR compiles or not, regardless of whatever other dpdgraph nonsense is broken?

abooij:
(I forgot to add that I learned of this presentation of the naturals from Martín Escardó and Cory Knapp.)

@mikeshulman: Even if we could use univalence everywhere, this is not desirable. You don't want to compute `x+y` by first computing the unary representation of `x` and `y`, doing the unary addition, and then converting back to binary. Doing it in binary directly is faster. Similarly for e.g. deciding equalities.

I'm not sure how to characterize mathematically which things are computationally relevant. As a first approximation: everything that, when fully evaluated at closed terms, does _not_ result in a reflexivity path. Addition of binary naturals gives a new binary natural, which I might wish to print on the screen. So that's computationally relevant to me. Comparing two binary naturals might give < or = or >, so that is computationally relevant to me: there are different possible outcomes (although for fixed closed terms there is only one). But, for example, I won't print the path inhabiting 1+2=2+1 on the screen, because it is reflexivity.

But I'm not sure this answers your question.

mikeshulman:
Good point about univalence!  What about the other direction; would we want to define operations on unary naturals in terms of binary ones?

As a second approximation to computational irrelevance, instead of just "reflexivity" what about saying that there is only one canonical form for closed terms?

abooij:
Strictly speaking, one might not always want to go in the other direction either: for example, the successor operation is O(1) for unary, but O(log n) for binary. But I suppose that's not the worst of complexities :)

Your second approximation explains my choices in this PR, with the clarification that I only consider closed terms of non-function type. For example, inequalities like `ineq_bzero_double1 : bzero <> double1 n` are not relevant to me, even if for different closed terms `n` they give different terms, since we won't be able to further evaluate them to a non-function type.

mikeshulman:
I'm confused -- are you saying that we should evaluate function terms fully on closed terms before comparing them, i.e. we could have two different function terms that would evaluate to the same thing on all closed terms?  In particular that would mean that no negation is ever computationally relevant --- which seems right --- since we can't ever evaluate it fully as there are no closed terms of type `Empty` to get out.

abooij:
That sounds correct. Let me phrase my thoughts another way, to make sure.

I consider a construction `A` computationally relevant if the following situation is possible: after fully evaluating the proof of `A` at closed terms to obtain a term not of function type, it is possible to get different terms (canonical forms).

So negations are never computationally relevant, because we would have to evaluate to get a term of type **0**, which (hopefully) doesn't exist. And the equalities in the semiring axioms aren't computationally relevant, because they always result in the reflexivity path. But addition is relevant, because we may get `bzero` or `double1 bzero` or... And deciding paths is relevant, because we may get `inl idpath` or `inr (\ x -> ...)`. Ditto for the proof of trichotomy.

This definition isn't completely satisfactory either: what to think of proofs of product types? If both of the factors are negations, such as `forall n, (n <> n+1) x (n <> n+2)`, then my definition may still consider them computationally relevant: we can evaluate the term to obtain a non-function type `(n <> n+1) x (n <> n+2)` for some closed term `n`, with different canonical forms for different choices of `n`. However, both projections are computationally irrelevant, as they are negations. Or, in a different scenario, such as `A x (~ A)`, the second factor makes the first factor irrelevant: if `~A` then we can't have *any* terms of type `A`, so in particular we can't have two different ones. So what is the correct definition in general?

Luckily the latter scenarios don't arise in this PR.

mikeshulman:
It feels like we want to be defining some sort of "syntactic extensional equality" by induction on types, so that it could descend into products as well as evaluating functions.

spitters:
@mikeshulman About Travis, I understand the graph issue is now solved, so things should work again if we bump to the latest Coq. We should probably also do a new opam release them, but I was waiting for someone to confirm this.

mikeshulman:
@spitters That's good to hear.  Can someone bump Coq?  Can I do that easily?

spitters:
I think this was the last time it was done:
https://github.com/HoTT/HoTT/commit/5b6d2caeeb605e87c5b52193e091a0e497751663

mikeshulman:
So I guess what I need to do is check out the "latest Coq" in the `coq-HoTT` subdirectory, then add that submodule to the commit.  But what does "latest" mean?  Which branch or tag?  Sorry, I've forgotten what our policy is here...

spitters:
I think this is the latest version:
https://github.com/coq/coq/tree/V8.7.2
We should also update .travis.yml to V8.7.2 
(I also forgot the precise process for updating)

spitters:
@mikeshulman I think we can pull this PR, can't we?

mikeshulman:
Well, I haven't had a chance to actually look at the code yet, which I feel obligated to do before approving it.  (-:

Issue #921, opened 2018-03-01
Possible bug in Book.py ?


marcbezem:
When I run Book.py with fresh HoTT/book and HoTT/HoTT, something weird occurs:
in the definition of Book_2_2 it substitutes occurrences of "2_1" by "2_2", compromizing the file HoTTBookExercises.v. This file was fine before I ran Book.py. There are a couple of other errors that I don't understand, but probably this part of the diff is sufficient to diagnose the problem (it is of course possible that I did something wrong):

````
HoTT> diff contrib/HoTTBookExercises.v contrib/HoTTBookExercises.v.bak.1
274,276c274,276
<     (Book_2_2_concatenation1_eq_Book_2_2_concatenation2 p q) *1
<     (Book_2_2_concatenation2_eq_Book_2_2_concatenation3 p q) =
<     (Book_2_2_concatenation1_eq_Book_2_2_concatenation3 p q).
---
>     (Book_2_1_concatenation1_eq_Book_2_1_concatenation2 p q) *1
>     (Book_2_1_concatenation2_eq_Book_2_1_concatenation3 p q) =
>     (Book_2_1_concatenation1_eq_Book_2_1_concatenation3 p q).
````
Book.py is happy with the labels, but does something that Coq is not happy about.

**Comment by mikeshulman** [comment]
Exactly what command line are you running, so I can test it?

And what branch/version of HoTT/HoTT are you using?  In my copy of upstream/master all these lines already start with `Book_2_1`.

**Comment by marcbezem** [comment]
Tomorrow I'll run `git show-branch` on the desktop I worked on today. Yesterday, I deleted my very old fork (too far behind) and forked afresh, from both HoTT/HoTT and HoTT/book. Part of my work flow is:

````
cat ../book/*.aux | etc/Book.py contrib/HoTTBook.v
cat ../book/*.aux | etc/Book.py --exercises contrib/HoTTBookExercises.v
````
The first went fine, the second does the weird thing in Book_2_2 (plus in some others). That definition uses definitions from Section Book_2_1_Proofs_Are_Equal, which are OK, and should read 2_1.

**Comment by mikeshulman** [comment]
Oh, I misread your diff: I didn't realize that you diffed from the new version to the old version (which is backwards to me), so I thought the change had gone in the opposite direction.  Now I see the problem, it's this line in Book.py:

```python
content = re.sub('Book_[0-9_]*[0-9]', 'Book_{0}'.format(book), content)
```

[...]

PR #922, opened 2018-03-01
simpler proof of Exercise 6.9 not using univalence

marcbezem:
HoTTBookExercises.v has been validated with `make contrib', and this went well. However, there is issue #921. The warnings about Book_5_2 in HoTTBookExercises.v appear to stem from pedagogical considerations:

````
Query commands should not be inserted in scripts
prod_rect
     : forall (A B : Type) (P : A * B -> Type), (forall (fst : A) (snd : B), P (fst, snd)) -> forall p : A * B, P p
The command has indeed failed with message:
In environment
ez := true : Bool
es := fun _ : nat => idmap : nat -> Bool -> Bool
The term "1" has type "Book_5_2_i = Book_5_2_i" while it is expected to have type "Book_5_2_i = Book_5_2_ii"
(cannot unify "Book_5_2_i" and "Book_5_2_ii").
The command has indeed failed with message:
In environment
ez := 1 : nat
es := fun _ : nat => S : nat -> nat -> nat
The term "1" has type "Book_5_2'_i = Book_5_2'_i" while it is expected to have type "Book_5_2'_i = Book_5_2'_ii"
(cannot unify "Book_5_2'_i" and "Book_5_2'_ii").
````

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1145,10 +1145,6 @@ Section Book_6_9.
 
   Definition AllExistsOther(X : Type) := forall x:X, { y:X | y <> x }.
 
-  Lemma isHPropAllExOth `{Funext} (X: Type) : IsHProp(Contr (AllExistsOther X)).
-  Proof. apply (hprop_trunc -2).
-  Defined.
-
   Definition centerAllExOthBool : AllExistsOther Bool := 
     fun (b:Bool) => (negb b ; not_fixed_negb b).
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: contrib/HoTTBookExercises.v [hunk 1]
```diff
@@ -1140,6 +1140,47 @@ Section Book_6_9.
         | [ H : false = true |- _ ] => exact (match false_ne_true H with end)
       end.
   Qed.
+
+(** Simpler solution not using univalence **)
+
+  Definition AllExistsOther(X : Type) := forall x:X, { y:X | y <> x }.
+
+  Lemma isHPropAllExOth `{Funext} (X: Type) : IsHProp(Contr (AllExistsOther X)).
```
mikeshulman: I don't think you need this lemma, since it can be inferred automatically by typeclass inference; just replace the invocations of this lemma with `_`.

marcbezem: Thanks, even simpler. I'll do this tomorrow. Is Travis is giving false negatives?

(@SIben, @jonaprieto)

mikeshulman: Yeah, it sounded like someone needs to try bumping the Coq version and see if that fixes travis.

mikeshulman:
FWIW, if a new commit is pushed to a PR *after* someone says LGTM, I don't think they count as a POE any more until they re-approve it.

spitters:
Sorry, the changes seemed minor enough. I'll wait for another LGTM next time.

PR #923, opened 2018-03-02
Solution to Exercise 4.6 (in pure MLTT)

mikeshulman:


spitters:
LGTM.

However, https://github.com/coq/coq/issues/3848 is closed.
Is there another issue you want to report?

mikeshulman:
No, that comment was just copied along with the code from `types/Forall.v`, where no one had gotten around to removing it after the Coq bug was fixed.  I removed them both along with the workaround line.

spitters:
Thanks. LGTM

andrejbauer:
Why did Travis fail?

mikeshulman:
Someone needs to update Coq to fix the dpdgraph problem.  I haven't had a chance to try yet.

mikeshulman:
If someone is willing to pull this ASAP that would be great -- Martin Escardo wants to cite it in an arxiv preprint, and a pull request isn't a good stable URL to cite.

spitters:
You mean @martinescardo ? 
I think @abooij knows enough Coq to review this. :-)

mikeshulman:
For some reason Github wouldn't autocomplete Martin's handle for me, so I thought I didn't have it right.

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+Section Book_4_6_i.
+  Section OneUniverse.
+  Definition qinv (A B : Type)
+    := { f : A -> B & { g : B -> A & (Sect g f * Sect f g)%type } }.
+  Definition qinv_id A : qinv A A
+    := (fun x => x; (fun x => x ; (fun x => 1, fun x => 1))).
+  Definition qinv_path A B : (A = B) -> qinv A B
+    := fun p => match p with 1 => qinv_id _ end.
+  Definition QInv_Univalence_type := forall (A B : Type@{i}),
+      IsEquiv (qinv_path A B).
```
abooij: This states that idtoqinv is a (regular) equivalence. However, (i) says that idtoqinv itself is a quasi-inverse.

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+  Definition equiv_qinv_path (ua: QInv_Univalence_type) (A B : Type)
```
abooij: The variable name `ua` may confuse some readers, as it suggests to be an abbreviation for "univalence axiom", which this is not.

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+    intros g;
+    first [ change ((fun x => w'.1 ( w'.2.1 (g x))) = g)
+          | change ((fun x => w'.2.1 ( w'.1 (g x))) = g) ];
+    clearbody w'; clear H0 w;
+    rewrite <- (@eisretr _ _ (@qinv_path A B) (ua A B) w');
+    generalize ((@equiv_inv _ _ (qinv_path A B) (ua A B)) w');
+    intro p; clear w'; destruct p; reflexivity.
+  Defined.
+
+  (* Now the rest is literally copied from UnivalenceImpliesFunext. *)
```
JasonGross: Why copy rather than reference?

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+
+  Definition ft_equiv_functor_forall_id 
+        {A:Type} `{P : A -> Type} `{Q : A -> Type}
+        (g : forall a, P a <~> Q a)
+  : (forall a, P a) <~> (forall a, Q a)
+    := ft_equiv_functor_forall (equiv_idmap A) g.
+
+End EquivFunctorFunextType.
+
+(* Using the Kraus-Sattler space of loops rather than the version in the book, since it is simpler and avoids use of propositional truncation. *)
```
JasonGross: You may want `(**` here, so that it shows up in the wiki / coqdoc version.

JasonGross:
> Someone needs to update Coq to fix the dpdgraph problem

#925 should fix the dpdgraph problem, though we may still have a Haskell problem

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+  Definition QInv_Univalence_type := forall (A B : Type@{i}),
+      IsEquiv (qinv_path A B).
+  Definition equiv_qinv_path (ua: QInv_Univalence_type) (A B : Type)
+    : (A = B) <~> qinv A B
+    := BuildEquiv _ _ (qinv_path A B) (ua A B).
+
+  Definition qinv_equiv {A B} (f : A <~> B) : qinv A B
+    := (equiv_fun f ; (f^-1 ; (eisretr f , eissect f))).
+  Definition qinv_isequiv {A B} (f : A -> B) `{IsEquiv _ _ f} : qinv A B
+    := (f ; (f^-1 ; (eisretr f , eissect f))).
```
abooij: (Doesn't the latter get inferred from the former by coercing an `f : A <~> B` to `f : A -> B`?)

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+  Local Instance ft_isequiv_functor_forall 
+        {A B:Type} `{P : A -> Type} `{Q : B -> Type}
+          {f : B -> A} {g : forall b:B, P (f b) -> Q b}
+        `{IsEquiv B A f} `{forall b, @IsEquiv (P (f b)) (Q b) (g b)}
+    : IsEquiv (functor_forall f g) | 1000.
+  Proof.
+    simple refine (isequiv_adjointify (functor_forall f g)
+                                      (functor_forall (f^-1)
+                                                      (fun (x:A) (y:Q (f^-1 x)) => eisretr f x # (g (f^-1 x))^-1 y
+                                      )) _ _);
```
abooij: As a suggestion, to reduce whitespace usage:
```
    simple refine (isequiv_adjointify
                     (functor_forall f g)
                     (functor_forall
                        (f^-1)
                        (fun (x:A) (y:Q (f^-1 x)) => eisretr f x # (g (f^-1 x))^-1 y)
                     ) _ _);
```

contrib/HoTTBookExercises.v
```diff
@@ -873,6 +873,228 @@ End Book_4_5.
...
+          rewrite ap_transport;
+          rewrite eisretr;
+          apply apD
+        ).
+    - abstract (
+          apply ft_path_forall; intros a; unfold functor_forall;
+          rewrite eissect;
+          apply apD
+        ).
+  Defined.
```
abooij: As the proof term is largely opaque, shouldn't the entire term be made opaque?

contrib/HoTTBookExercises.v
mikeshulman: Well, I had to change the names of a few lemmas so that they would use the current assumption of qinv-univalence rather than a global assumption of ordinary univalence.

mikeshulman: I should probably delete "literally" -- I think that comment was written at an early stage when it was a literal copy.

contrib/HoTTBookExercises.v
JasonGross: No, if the term is made opaque, then we lose the inverse function.  Although the third and fourth arguments to `isequiv_adjointify` are (mostly) opaque (and the `intros h` could be moved inside the abstract), the first two arguments are transparent.

contrib/HoTTBookExercises.v
mikeshulman: The other way around, but yes.

mikeshulman:
Thanks for your suggestions @JasonGross  and @abooij , I think I've fixed them all.

martinescardo:
Could we have this merged so that I can cite it in a paper that I am about to submit to the arxiv?
Thanks.

martinescardo:
Thanks!

lbordowitz:
@martinescardo I was able to find this pull request in the citation [of a paper](http://www.cs.bham.ac.uk/~mhe/agda-new/UnivalenceFromScratch.html), using this format:

> Michael Shulman.  Solution to Exercise 4.6 (in pure MLTT), March
> 2018.  https://github.com/HoTT/HoTT/pull/923.

Referring to the pull request can also be less laborious for researchers, getting a relevant snippet of code instead of having to cull through a large library.

martinescardo:
Ha. This is what I had originally. But then @mikeshulman asked me not to cite the pull request but instead the exercises. :-)

martinescardo:
I am citing https://github.com/HoTT/HoTT/blob/master/contrib/HoTTBookExercises.v#L873 at his request, but I am not sure how robust "'#L873" is.

mikeshulman:
It's not, sorry.  The most robust link would be to a particular commit, like maybe https://github.com/HoTT/HoTT/commit/531bc5865089cb8b32a0c49d0f9bf220f811a292 .

martinescardo:
I submitted the paper with this reference now. Thanks for solving this exercise for the paper. :-)

mikeshulman:
Great!  You're welcome, this is a nice fact to have -- that qinv-univalence, at least for three nested universes, is inconsistent in pure MLTT.  I would like to know whether qinv-univalence is consistent for *two* nested universes.  I wonder whether it would be worth adding a note to Exercise 4.6 in the book that you can get extra credit by doing it without propositional truncation.

martinescardo:
What if you have a universe U whose elements are all sets, and one universe U' having everything that U has plus U itself, with Id_U'(U,X)=0 for all x:U? Then we should have qinv(f) equivalent to isEquiv(f), right? Doesn't the groupoid model validate that?

martinescardo:
For the record, the paper has now appeared as https://arxiv.org/abs/1803.02294 with the Agda code for the definition of the univalence type as an ancillary file.

martinescardo:
As I just said in the HoTT list, I am looking for a volunteer to produce a self-contained file with the quickest possible formulation of Voevodsky's version of univalence in Coq, to be cited in the eventual version 2 of the above paper. In the Agda version attached as an ancillary file to arxiv paper, I add at the end of the file the normal form of the univalence type in two forms (1) with the implicit arguments omitted, and (2) with them made explicit. It would be nice to have the same in a Coq version.

PR #924, opened 2018-03-05
No substitution of "Book_X_Y" at all

marcbezem:
Line 193 in Book.py broke HoTTBookExercises.v, for example, in Book_2_2 where references to Book_2_1 occur. I experimented with limiting substitution to contexts of the form `Definition Book_X_Y`. But then Exercise 2.3 was still broken, while it made sense to call the definition there Book_2_1_concatenation4. We probably don't need the substitution anymore, as the numbering in book is not likely to change. If this assumption proves to be false, the substitution could be reintroduced on an ad-hoc basis. 

If this patch is accepted, I will change the guidelines in the files HoTTBook{Exercises}.v accordingly.

### DIFF
--- a/etc/Book.py
+++ b/etc/Book.py
@@ -190,7 +190,9 @@ for label in sorted(entries.keys(),
         content = content[content.index('\n')+1:]
         # Update Book_X_Y_Z
         book = "_".join(map(str,entry['number']))
-        content = re.sub('Definition Book_[0-9_]*[0-9]', 'Definition Book_{0}'.format(book), content)
+        # content = re.sub('Book_[0-9_]*[0-9]', 'Book_{0}'.format(book), content) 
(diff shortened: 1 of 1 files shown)


mikeshulman:
I agree with this.

(Also, on principle, I think it's better for any change like this to have its own PR so that people can see what's happening, rather than being a commit added later to a PR that was originally about something else.  I didn't realize until I read this PR that the "Definition" version of this change had already been slipped into #922 after I approved it but before @spitters merged it.)

marcbezem:
It was unintended to change my previous pull request. How do I prevent further commits to be added to an open pull request?

mikeshulman:
Make each pull request from a separate branch.

JasonGross:
This seems all-right to me, though I have no strong opinion either way; I'll defer to @mikeshulman 's agreement and say feel free to merge this.

marcbezem:
Thanks. I will now update the guidelines in the files HoTTBook{Exercises}.v accordingly.

Issue #883, opened 2017-06-26
Branch organization for Coq versions


JasonGross:
How do we want to organize branches for different versions of Coq?  In particular, should `master` track trunk, or should it track the latest released version of Coq?  I see four possibilities:
1. `master` tracks Coq trunk; new development happens on `master`, version-specific branches are considered stale/locked
2. `master` tracks Coq trunk; new development happens on either `master` or version-specific branches, and version-specific branches are merged into master periodically
3. `master` tracks Coq trunk; new development happens on either `master` or version-specific branches, and master is merged into version-specific branches periodically
4. `master` tracks the latest released version of Coq, we have a `trunk` branch that tracks trunk, but on which no new development happens; someone (me?) maintains the `trunk` branch either by rebase+force-push or by merging `master` into it; when a new stable version of Coq is released, we branch off the old version from `master` (so, when 8.7 is released, we branch off `v8.6`), and merge `trunk` into `master` and update `master` to track the new version.

I'm partial to either 2 or 4; 2 if we want to support developments on trunk before it's released (e.g., experiments with universe cumulativity, induction-recursion, etc), or 4 if we want to support trunk only for Coq's ci-testing.

mikeshulman:
I don't understand how 3 would work; if `master` has moved on from Coq v8.6, then how could it be merged into a `v8.6` branch?

4 sounds complicated and error-prone.

What is the advantage of 2 over 1?  Having development happen on multiple branches seems also complicated and error-prone.

JasonGross:
3 does not work well, I expect.  It would rely on Coq not changing too much between versions.

4 only works if no-one does development on the trunk branch, and there's only one person who touches it.

2 gives more support for people who don't want to install a Coq trunk.  For example, we could have the "official" branch be `v8.6` until 8.7 is released.  I agree, though, that it is more complicated, and I'd be fine with 1.

mikeshulman:
I could live with 2 if others think it's better, but I'm slightly more inclined to the simpler 1.

andrejbauer:
We do not have manpower to support lots of versions of the library. We have a two-dimensional problem:

* support with respect to Coq versions
* support with respect to development

For development only, a common strategy is to have `stable` (for users) and `master` (for developers). I am not familiar with how people deal with dependence on another piece of software that changes often, but I suppose again it would make sense to have two branches, one fixed to the current official release of Coq, and another one tracking Coq development.

But how do we solve both dimensions at once? Perhaps like this:

[...]

PR #925, opened 2018-03-05
Update coq submodule

JasonGross:
As per #883, we are tracking Coq master with our master branch.  I've updated coq-dpdgraph to the latest commit on its coq-trunk branch, and updated coq-scripts to its latest commit (because why not?).  This will
hopefully fix the issues we've been having recently with graphs on Travis, though I haven't looked into them much.  (I'm a bit worried that one of the failures to build seems to be a result of cabal not being able to resolve dependencies for the packages we're looking for...)

### DIFF
--- a/etc/ci/install_coq_dot_deps.sh
+++ b/etc/ci/install_coq_dot_deps.sh
@@ -11,6 +11,4 @@ do
     n=$[$n+1]
     sleep 10
 done
-cabal install graphviz-2999.18.0.2 text || \
-    (cabal install graphviz-2999.18.0.2 &&
-            cabal install text) || exit $?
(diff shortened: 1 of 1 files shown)


spitters:
LGTM, in that I don't see anything strange and I trust @JasonGross in this.
I'm surprised we are still supporting all these old versions, but then again, why not.

JasonGross:
@spitters It's closer to "I don't want to update the code every time a new version comes out" than "I want to support old versions"

mikeshulman:
LGTM, thanks!  I hope it makes Travis happy.

JasonGross:
Updating the version of Haskell's graphviz we ask for seems to have fixed the issue with cabal.

Issue #927, opened 2018-03-24
Switching to 8.7


andrejbauer:
Is there a reason the HoTT library is becoming seriously stale and difficult to compile with new OCaml and new Coq? I have HoTT `V8.7` working on my machine. Why isn't it the official version of HoTT? Unless there's a good reason, we should switch ASAP.

spitters:
Hasn't this been done in:
https://github.com/HoTT/HoTT/pull/925

SkySkimmer:
The stable branch's coq submodule needs to be updated to 8.7.2 to compile with ocaml 4.06. Is that what you mean?
What is the official version of HoTT exactly? Where is that stated?

spitters:
coq-HoTT currently points here:
https://github.com/coq/coq/tree/15331729aaab16678c2f7e29dd391f72df53d76e

@JasonGross  updated this 20days ago, so I expect it to be new enough.

mikeshulman:
And I finally got it to compile on my machine by blowing away all my ubuntu ocaml packages and reinstalling them all with opam instead.

```
~/hott/coq$ ./hoqc --version
The Coq Proof Assistant, version 8.8+alpha (March 2018)
compiled on Mar 26 2018 9:53:25 with OCaml 4.06.1
```

So I'm also confused about what you mean, @andrejbauer .

PR #930, opened 2018-03-29
Update INSTALL.md

spitters:
Updating opam instructions


spitters:
Travis is unhappy about the graphs. This should be related to #927

spitters:
We should probably update to 8.8. However, I don't fully recall how to do that.
This seems to be the bump to 8.7.
https://github.com/HoTT/HoTT/commit/5b6d2caeeb605e87c5b52193e091a0e497751663

We should also update .travis.yml and opam.
@JasonGross am I overlooking something?

JasonGross:
#946 should hopefully fix the graphs issue

PR #931, opened 2018-04-13
Non-dependent funext implies dependent funext

mikeshulman:
Martin Escardo pointed out that this fact is in Foundations.  But somehow it never made it into this library, and I wasn't even aware that it was known.

I had to move `FunextVarieties` out of `Basics` so that it can import `EquivalenceVarieties`, but this had no ill effects, just requires importing it explicitly in a few places.

### DIFF
--- a/theories/FunextVarieties.v
+++ b/theories/FunextVarieties.v
@@ -8,40 +8,47 @@ Local Open Scope path_scope.
 
    This proof is originally due to Voevodsky; it has since been simplified by Peter Lumsdaine and Michael Shulman. *)
 
-(** Naive funext is the simple assertion that pointwise equal functions are equal. *)
+(** Naive funext is the simple assertion that pointwise equal functions are equal.  The domain and codomain could live in different universes; the third universe argument is essentially the max of [i] and [j] (and similarly for all subsequent axioms). *)
 
 Definition NaiveFunext :=
-  forall (A : Type) (P : A -> Type) (f g : forall x, P x),
+  forall (A : Type@{i}) (P : A -> Type@{j}) (f g : forall x, P x),
     (forall x, f x = g x) -> (f = g).
+Check NaiveFunext@{i j max}.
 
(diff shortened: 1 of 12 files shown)


theories/FunextVarieties.v
```diff
@@ -25,7 +31,7 @@ Definition Funext_type :=
   forall (A : Type) (P : A -> Type) f g, IsEquiv (@apD10 A P f g).
```
mattam82: As opposed to which version of Funext which involves an axiom?

SkySkimmer: In order to be able to abstract over funext at every universe level there is a monomorphic axiom `Funext : Type` together with an axiom that it implies the funext statement.

mikeshulman:
Right, the point is that this is true without univalence.

spitters:
LGTM

mikeshulman:
Since this hasn't been merged yet, I went ahead and added to it: by using this proof, we can eliminate the use of two univalent universes in the proof of funext -- one univalent universe suffices to have funext for types in that universe.  (This is essentially the proof in the book, though not factored in this way; I don't know why it never made it into this library before.)  This also simplifies the proof that qinv_univalence is inconsistent so that it requires only two (qinv-)univalent universes, which is clearly the minimum (one qinv-univalent universe is consistent, at least as long as it doesn't contain HITs, since it could contain only sets).

mikeshulman:
(This also all arose in further conversation with Martin.)

martinescardo:
I implemented this two days ago in Agda, http://www.cs.bham.ac.uk/~mhe/agda-new/UF-FunExt-from-Naive-FunExt.html, where one sees that we get, with non-cumulative universes,

∀ {U} {V} → naive-funext U (U ⊔ V) → naive-funext U U → funext U V

and in particular 

∀ {U} → naive-funext U U → funext U U

spitters:
I'm taking this as an ack from @mattam82 and @martinescardo

PR #932, opened 2018-04-18
Add `path_prod_V` (from Section 2.6)

co-dan:
A simple addition. I didn't find anything for `path_prod` and `inverse`, and I had to use this lemma in some other formalisation.


spitters:
Thanks. LGTM.

PR #936, opened 2018-05-10
Show that `path_sigma_hprop` and `path_trunctype` are functorial.

co-dan:
We've used these properties in another development, but they seemed general enough to me.
I don't think those are mentioned in the HoTT book necessarily, though.
I tried to implement it without rewrites, but failed to do so in a number of cases, don't know if that's important.

### DIFF
--- a/theories/TruncType.v
+++ b/theories/TruncType.v
@@ -56,6 +56,41 @@ Proof.
   exact _.
 Defined.
 
+(** [path_trunctype] is functorial *)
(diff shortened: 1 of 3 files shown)


co-dan:
Oops those Qed's were typos. I changed them to `Defined`. Do you want me to use `Qed`s in the lemmas with `rewrite`s in this pull request?

mikeshulman:
Sure, if you don't mind.

PR #938, opened 2018-05-18
Timing files should be "foo.v.timing" to match coq_makefile

SkySkimmer:
This should allow them to be picked up by the timing to html script in
Coq's benchmark.


JasonGross:
Travis is happy aside from dpdgraph being the wrong version for the Coq that we're using, which we should fix.

spitters:
dpd-graph is currently a submodule. However, it's currently also in opam.
@JasonGross should we move it to opam?

PR #939, opened 2018-05-18
Added some functors/natural transformations.

nmvdw:
I added the constant functor (new file).
I added the canonical functor for associativity of taking products of categories.
I added the pairing of natural transformations.

### DIFF
--- a/theories/Categories/Functor/Constant.v
+++ b/theories/Categories/Functor/Constant.v
@@ -1,0 +1,14 @@
+(** * Constant functor *)
+Require Import Categories.Category.
+Require Import Category.Core Functor.Core.
+
+Definition const_functor
+           {C D : PreCategory}
+           (X : D)
+  : Functor C D
+  := Build_Functor C
+                   D
+                   (fun _ => X)
+                   (fun _ _ _ => Category.identity X)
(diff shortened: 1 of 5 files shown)


spitters:
Thanks. 
However, you don't prove anything about these operations.
You must have had a reason for including them.
Ideally, this should be better connected to the rest of the library.

Of course, there is no reason for not pulling these additions now. They are welcome additions.

theories/Categories/NaturalTransformation/Prod.v
```diff
@@ -29,6 +29,19 @@ End prod.
 
 Local Infix "*" := prod : natural_transformation_scope.
 
+Definition pair
```
JasonGross: Please add a notation for this below, as is done in the functor prod file.

theories/Categories/Functor/Constant.v
```diff
@@ -0,0 +1,14 @@
+(** * Constant functor *)
+Require Import Categories.Category.
```
JasonGross: Please only `Require` the files needed, and not file that imports everything in the directory.

Alizter:
Since it looks like this PR is inactive I will close it. @nmvdw do let me know if you think otherwise.

PR #944, opened 2018-07-12
Implement Escardo's bounded search

abooij:
If we know that there exists a natural number for which a decidable predicate `P` holds, then we can find a natural number for which `P` holds, by a bounded search.

This theorem was also implemented by [Matthieu](https://mattam82.github.io/HoTT_Markov/) (using the HoTT library) and [myself](https://github.com/UniMath/UniMath/pull/787) (using the UniMath library) (although I have previously erroneously referred to this as an instance of Markov's principle, which it is not: MP converts negative information into positive information, whereas this theorem converts mere existence into witnesses.).

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -26,6 +26,7 @@ theories/Types/Wtype.v
 theories/Types/Equiv.v
 theories/Types/Universe.v
 theories/Types.v
+theories/BoundedSearch.v
 theories/EquivGroupoids.v
 theories/Fibrations.v
 theories/Conjugation.v
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -776,7 +776,7 @@ End Book_3_14.
 (* ================================================== ex:decidable-choice *)
 (** Exercise 3.19 *)
 
-
+Definition Book_3_19 := @HoTT.BoundedSearch.minimal_n.
 
 (* ================================================== ex:omit-contr2 *)
 (** Exercise 3.20 *)
--- a/theories/BoundedSearch.v
+++ b/theories/BoundedSearch.v
@@ -1,0 +1,113 @@
+Require Import Coq.Init.Peano.
+Require Import HoTT.Basics HoTT.HIT.Truncations HoTT.HProp HoTT.Types.Arrow HoTT.Types.Forall HoTT.Types.Sigma HoTT.Spaces.Nat HoTT.DProp.
+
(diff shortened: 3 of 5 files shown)


spitters:
LGTM. Although, I don't think that's due to Martin. Folklore, I think.

abooij:
I don't know what the earliest version of this is. For what it's worth, Martin has written about it on the [Univalent Foundations list in 2013](https://groups.google.com/forum/#!msg/univalent-foundations/SA0dzenV1G4/d5iIGdKKNxMJ).

mikeshulman:
Oh, wait, maybe that's something else; this is just Exercise 3.19 in the book, right?  That can't date much further back than 2013 in the specific form referring to propositional truncation in HoTT!  However I'm sure it has obvious precursors in other theories.

abooij:
@mikeshulman: Right, this is Exercise 3.19 - thanks for that. I suppose I should add a commit that adds a cross-reference to HoTTBookExercises.v.

The 2007 blog posts indeed concerns a different kind of search :)

abooij:
As a further reference: Mike [added](https://github.com/HoTT/book/commit/61f4e14c8d95b0eaa797a247afb1563dc1819761) this as a theorem to `basics.tex` of the HoTT book on March 12, 2013.

mikeshulman:
Martin's message to the UF list was dated March 13, 2013, but I would expect that it came first - timezone issue maybe.

spitters:
Travis failed with;
```
Could not locate requirements.txt. Override the install: key in your .travis.yml to install dependencies.
```
I don't know what that means, but I've restarted it.

mikeshulman:
I don't know what that means either.  Anyone with travis-fu want to help?

SkySkimmer:
I don't see that error, I only see some universe instance length issue and some dpdgraph issue.

mikeshulman:
Ah, I see, that message is not actually the failure error -- it's just some kind of warning.  To view the actual error you have to expand the last fold in the output.

mikeshulman:
I never know what's up with dpdgraph.  The universe instance error is on Coq v8.7; didn't we just decide to drop support for that in #946 ?  Why is Travis still testing it?

mikeshulman:
Huh.  I thought travis tested a PR based on the current configuration of the repo, rather than on the contents of the PR itself -- otherwise a PR could tell travis not to test it and vacuously "pass".   I must be confused about something.

PR #945, opened 2018-07-17
Fix for numeral notations

JasonGross:
This PR is required for compatibility with https://github.com/coq/coq/pull/8064.  Since it is not backwards compatible (and there's no easy way to make it backwards compatible), it should not be merged until https://github.com/coq/coq/pull/8064 is merged (and it should be merged immediately afterwards).

I still need to update .travis on this PR to test only with the master branch.

Before this PR is merged, we should make a tag for v8.8, and update travis (since we're no longer compatible with 8.7).


Zimmi48:
@JasonGross When the tag for v8.8 is made, you should open a PR on Coq v8.8 branch to make it test this tag instead of the branch.

JasonGross:
Note that this is ready for review, even if it cannot be merged yet.

JasonGross:
Rebased, and travis passes

Issue #934, opened 2018-05-04
Totality of relations as property or structure?


abooij:
Currently, in Classes, a relation is considered total if it comes equipped with a choice of disjuncts
https://github.com/HoTT/HoTT/blob/eeaffdba64ff6d9c7d38871532375c45cb3166a8/theories/Classes/interfaces/canonical_names.v#L245

Would it perhaps be better to use truncated disjunction (`hor`)? Speaking of which, would it be reasonable to add notation for `hor`, `hexists` and `merely` (defined in `theories/HIT/Truncations.v`)? (Although some editors, such as company-coq, prettify the "exists" tactic to the reverse-E existential quantifier, which may be confusing, because it generates an element of a Σ-type. And the `\/` notation from the Coq Init library gets prettified to `∨`, even though it is `+`.)

abooij:
I have [tried](https://github.com/abooij/HoTT/tree/total_hor) making this change from structure to property. Most proofs require very little change, but I have found two theorems (about the rationals) so far that require some redevelopment.

mikeshulman:
Then are you saying that the HoTT library defines `\/` to mean `+`?!?  That's clearly the wrong thing to do; where is that?

mikeshulman:
@abooij said "the `\/` notation from the Coq Init library" so I thought he was talking about vanilla Coq.

abooij:
https://github.com/HoTT/HoTT/blob/eeaffdba64ff6d9c7d38871532375c45cb3166a8/coq/theories/Init/Notations.v#L16

mikeshulman:
Sure, that's where the notaion is reserved -- where is it defined?

abooij:
Sorry, I linked the wrong code, please refresh this page to see my updated comment :)

mikeshulman:
Huh.  Well, we should get rid of that.  I suppose whoever wrote that was figuring that `\/` should mean _something_, and `hor` isn't defined yet in that file, but I think it's highly confusing to define `\/` to mean `+`.  It should mean `hor`; at least, that's how the book uses it.

mikeshulman:
Regarding the original question, are there any examples of relations that are constructively total-as-property but not total-as-structure?  I think the order on the rationals is total-as-structure, since it is decidable, whereas the order on the reals is not even total-as-property.

SkySkimmer:
IIRC I kept using `\/` in Classes because `+` is overloaded. The point is that we should have a notation for `+` which is less ambiguous than it.

[...]

PR #948, opened 2018-07-24
Make \/ mean hor, not sum.  Use |_| if sum is intended.

mikeshulman:
See discussion at the (largely unrelated) issue #934 .

### DIFF
--- a/coq/theories/Init/Datatypes.v
+++ b/coq/theories/Init/Datatypes.v
@@ -44,8 +44,8 @@ Notation "x + y" := (sum x y) : type_scope.
 Arguments inl {A B} _ , [A] B _.
 Arguments inr {A B} _ , A [B] _.
 
-Notation "A \/ B" := (A + B)%type (only parsing) : type_scope.
-Notation or := sum (only parsing).
+(* A notation for coproduct that's less overloaded than [+] *)
+Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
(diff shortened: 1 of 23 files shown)


coq/theories/Init/Datatypes.v
```diff
@@ -44,8 +44,8 @@ Notation "x + y" := (sum x y) : type_scope.
 Arguments inl {A B} _ , [A] B _.
 Arguments inr {A B} _ , A [B] _.
 
-Notation "A \/ B" := (A + B)%type (only parsing) : type_scope.
-Notation or := sum (only parsing).
+(* A notation for coproduct that's less overloaded than [+] *)
+Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
```
JasonGross: Why are we keeping this `only parsing`?  The `\/` notation was `only parsing` so that if you wrote `a + b`, it wouldn't print as `a \/ b`.

theories/Utf8.v
```diff
@@ -49,6 +49,8 @@ Notation "∥ A ∥" := (Trunc -1 A) (only parsing).
 Notation "∥ A ∥₋₁" := (Trunc -1 A).
 Notation "| a |₋₁" := (@tr -1 _ a) : trunc_scope.
 
+Notation "x ∨ y" := (hor x y) : type_scope.
+
```
JasonGross: Maybe also add `Notation "x ⊔ y" := (sum x y) : type_scope.`?

coq/theories/Init/Datatypes.v
mikeshulman: I prefer `a + b` to `a |_| b` myself, so I wouldn't want it to display as the latter.  I only included `|_|` as an option because in #934 there were requests for a less overloaded notation than `+`.

theories/Utf8.v
mikeshulman: Done.

abooij:
LGTM

spitters:
@abooij Since you are the second person to review this and travis is happy, it would be OK to pull.

PR #952, opened 2018-08-28
Set Coq version to 8.8.1

JasonGross:
This commit is intended to be the last commit that we guarantee builds
with Coq v8.8.  This commit (or the merge commit of this PR) will be the
target of the V8.8 tag, and after this commit is merged, we will drop
support for Coq 8.8 in preparation to merge the PR that allows
compatibility with numeral notations.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,11 +1,11 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = master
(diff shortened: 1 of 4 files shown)


JasonGross:
@spitters (or @andrejbauer or @peterlefanulumsdaine), could you review this (and #953 and #945)?  I think merging this and #953 and having #945 reviewed and ready to merge is all that's blocking the numeral notations PR.

PR #956, opened 2018-09-03
Don't allow validate to fail

JasonGross:
Now that the validate target actually works, we don't want to allow it to fail.


spitters:
LGTM

PR #961, opened 2018-10-01
Fix configure.ac

mattam82:
toploop doesn't exist but toplevel does. It worked for me.


spitters:
Travis:
checking coq version... 8.10+alpha (September 2018)
configure: error: coq-dpdgraph needs Coq version 8.8 or higher

mattam82:
I might be wrong somewhere, which version is HoTT's master supposed to follow?

spitters:
https://github.com/HoTT/HoTT/blob/master/INSTALL.md says:
One can install our version of Coq by:

   opam install coq.8.8.0

We will be compatible with Coq 8.9, so binary packages can be used once it is released. Until then, the development version of Coq must be installed. Either way, paths still need to be set manually.

JasonGross:
Please rebase on top of current master to get the fixes for travis

PR #966, opened 2018-10-26
Adapt to coq/coq#8820

jashug:
Adds compatibility with coq/coq#8820.
`EquivalenceInduction.v` had two premises named `HP`.
The changes in `InitialTerminalCategory/` are to avoid coq/coq#8819.

### DIFF
--- a/theories/Categories/InitialTerminalCategory/Functors.v
+++ b/theories/Categories/InitialTerminalCategory/Functors.v
@@ -98,7 +98,7 @@ Section unique.
     := fun P x => initial_category_ind P (F x).
 
   Global Instance trunc_terminal_category
-         `{@IsTerminalCategory one H0 H1}
+         `{@IsTerminalCategory one H1 H2}
   : Contr (Functor C one).
   Proof.
     refine (BuildContr _ (to_terminal C) _).
--- a/theories/Categories/InitialTerminalCategory/NaturalTransformations.v
+++ b/theories/Categories/InitialTerminalCategory/NaturalTransformations.v
@@ -43,7 +43,7 @@ Section NaturalTransformations.
     := trunc_from_initial F G.
 
   Definition to_terminal
-             `{@IsTerminalCategory one H0 H1} (F G : Functor C one)
+             `{@IsTerminalCategory one H1 H2} (F G : Functor C one)
   : NaturalTransformation F G
     := Build_NaturalTransformation
          F G
@@ -52,7 +52,7 @@ Section NaturalTransformations.
 
   Global Instance trunc_to_terminal
          `{Funext}
-         `{@IsTerminalCategory one H0 H1} (F G : Functor C one)
+         `{@IsTerminalCategory one H1 H2} (F G : Functor C one)
   : Contr (NaturalTransformation F G).
   Proof.
     refine (BuildContr _ (to_terminal F G) _).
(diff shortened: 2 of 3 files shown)


spitters:
Ideally, we should avoid referring to autogenerated names.

Travis could not find coqtop. Not sure what is going on. I've restarted it.

SkySkimmer:
Try purging the travis cache

Gaëtan Gilbert

mikeshulman:
I don't think we are referring to any autogenerated names.  It looks like the change to Categories is to work around a Coq bug that sometimes doesn't allow user-chosen names that *look* like autogenerated names, and the change to EquivalenceInduction is because a change to Coq isn't any more going to allow two user-chosen names that are the same (why was that ever allowed?).

LGTM once Travis is happy.

spitters:
@SkySkimmer I deleted the cache and restarted. Still failing. Can you have a look?

spitters:
@mikeshulman I meant the autogenerated ones here:
`{@IsTerminalCategory one H1 H2}

SkySkimmer:
Looks like travis runs into some issue installing from the @JasonGross ppa. I'll leave further debugging to him.

mikeshulman:
@spitters I don't understand: what's autogenerated there?  The parameters H1, H2 are explicitly user-given names (though their types are inferred by automatic typeclass generalization).  Am I reading the code wrong?

spitters:
@mikeshulman You are right. I was reading it too quickly.

JasonGross:
Looks like there's another package I need to add to the ppa, since I updated the version of OCaml to work with master; I'll try to fix it within the next day or two

JasonGross:
I have tried fixing this.  Can someone remind me why the HoTT travis build depends on opam, and whether it might be possible to eliminate this dependency?  I am failing at getting an opam package built on my ppa, because it depends on jsonm [which has an Ubuntu package which does not build](https://bugs.launchpad.net/ubuntu/+source/jsonm/+bug/1800410), and dose3 which depends on rpm which I have not yet managed to get to build on Trusty.

spitters:
I don't recall, but perhaps now that the stdlib issue has been resolved
things can be simplified.

from my phone

JasonGross:
Trying to drop opam stuff on travis at #967

JasonGross:
Please rebase on top of current master to get the fixes for travis

JasonGross:
@mikeshulman The issue is that, in a section where we have ```Context `{Funext}.```, we have (implicitly) `Context {H : Funext}.`, and then if we write ``` `{@IsTerminalCategory one H0 H1}```. Coq elaborates this to `{H0 : @IsTerminalCategory one H0 H1}`, which this update to Coq makes invalid.  So, indeed, we are relying on shadowing of autogenerated names.

jashug:
Thanks for tracking down the exact path of dependencies @JasonGross. I liked how @mikeshulman phrased it, that this arises from "a Coq bug that sometimes doesn't allow user-chosen names that look like autogenerated names".

Moving forward, I'm looking at changes to Coq to automatically avoid names like `H0` given by the user in `{@IsTerminalCategory one H0 H1}`, so that we could automatically have `{H2 : @IsTerminalCategory one H0 H1}`. However, the binder `{H : Funext}` given by `Context` is fixed early on, so writing `{@IsTerminalCategory one H H0}` will probably fail with a collision on `H` for the forseeable future.

mikeshulman:
I feel like the ideal solution would be to make it impossible for user-chosen names to ever conflict with autogenerated names; they should really have entirely different namespaces.  Maybe autogenerate names according to some pattern, or containing some special characters, that are disallowed in user-chosen names.  Probably that is infeasible however.

JasonGross:
Maybe Coq can wrap autogenerated names in angle brackets or something

JasonGross:
It looks like this PR is still waiting for APOE to actually ack it

PR #964, opened 2018-10-18
Update Coq prelude for numeral notations using Coqlib

vbgl:
This small patch will make this development compatible with Coq master, even after https://github.com/coq/coq/pull/8720 is merged.

PR #969, opened 2018-11-05
Adapt to local universes for opaque polymorphic definitions.

SkySkimmer:
This patch should be backwards compatible.

Once we drop compatibility this will allow us to remove a bunch of
annoying boilerplate universe manipulation.


abooij:
> Currently, things seem to be getting longer not shorter.
> However, I understand this is a preparation for an upcoming change.

If so, doesn't it make more sense to wait for the stuff that this PR depends on to complete?

SkySkimmer:
I can make a PR with non backward compatible https://github.com/SkySkimmer/HoTT/commit/2d4bf3814b1c2a72fb910f072e714c9c1e792b45 instead if you prefer.

spitters:
Travis is not happy.
What do you mean "Once we drop compatibility" ?

JasonGross:
I think my preference here is to merge the backwards compatible PR once Travis is happy, and have the non-backwards-compatible PR that makes things shorter ready at the same time, but not merge that until after 8.9 is released, as in #964

abooij:
(Nice commit hash! `abacdc`)

I'm a bit confused by this. Why is a commit being merged that doesn't improve the current code? The potential improvements in universe handling look nice, but why does something have to be merged right now, which only requires another PR later on? Couldn't this have waited until those improvements are actually possible? Someone reading this code right now could end up confused about the state of universes.

JasonGross:
The commit is being merged because the contract between Coq and projects on Coq's CI is that projects on the CI maintain compatibility with the tip of Coq master.  The code we had before this PR will be incompatible with Coq master once the PR is merged.  At the same time, I don't want to break compatibility with Coq 8.9 before 8.9 is even released.  So we resort to this kludge.

Once Coq 8.9 is released, I plan to tag a version of HoTT as "officially compatible with Coq 8.9".  Then we can drop support for 8.9 on the tip of master, and merge a clean-up PR that improves universe handling.

abooij:
Okay, I see, thanks for explaining!

PR #971, opened 2018-11-28
[build] Use coqc for batch compilation instead of coqtop.

ejgallego:
`coqtop` is more geared towards interactive use; use `coqc` instead in
prevision of changes upstream.


mikeshulman:
LGTM.  Is Travis still choking on dpd-graph?

ejgallego:
Thanks!

PR #974, opened 2018-12-27
Cantor's theorem, k-small categories and power objects

Alizter:
Attempted to prove Cantor's theorem in theories/Spaces/Card.v but couldn't finish.

I have tried to define strict inequality but it seems there are two choices. I have gone for the more constructive looking one but I am not sure if this matters much given Cantor's theorem has a nonconstructive proof.

Added definitions for what it means for a precategory to be k-small and started a proof of Freyd's preorder proposition. Which a full proof is detailed in the file contrib/SmallCategory.v but I think I will need some more cardinal arithmetic lemmas before I can finish it. Also defined thin categories, having k-small (co)limits, and k-small diagrams etc.

Defined the indexed power functor in contrib/Powers.v. It just takes the limit functor over the discrete category of a set.

---

I am submitting the pull request in case anybody want's to help. Also I might be inclined to make Cantor's theorem a theorem in HSet.v with a corollary in Card.v however this is up to you. I have no idea if there are any constructive variants of cantors theorem.

The notion of thin category / preorder category might be better suited on its own.

There aren't many examples of limits in the Category theory library (unlike the UniMath one). I have started Powers.v in order to reason about it's universal property C(A,B^l) = C(A, B)^l. There is still more to add.

The proof of Freyd's preorder proposition may have a constructive counter part but I have not been able to come up with one.

### DIFF
--- a/contrib/FixPoint.v
+++ b/contrib/FixPoint.v
@@ -1,0 +1,62 @@
+Require Import Basics.Overture.
+Require Import Types.Bool.
+Require Import HProp.
+
+Require Import Basics.Trunc. (* IsEmbedding *)
+Require Import HIT.Truncations. (* IsSurjection *)
+
+Import TrM.
+
+Lemma isSur2 {A B} {f : A -> B} {sur : IsSurjection f} :
+  forall y, exists x, f x = y.
+Proof.
+  intro.
+  destruct (sur y) as [c d].
+  set (c' := c).
+  apply (untrunc_istrunc) in c'.
+  + apply c'.
+Admitted.
+
+Section Lawvere.
+
+  Context {A B : Type}.
+
+
+(** 
+    Lawveres theorem
+    If there is an onto map e : A -> B^A then every f : B -> B has 
+    a fixed point.
+**)
+
+  Theorem Lawvere (e : A -> (A -> B)) {sur : IsSurjection e} 
+    : forall (f : B -> B), exists x, f x = x.
+  Proof.
+    intro f.
+    set (sur2 := isSur2 e).
+    destruct (isSur2 (fun x => f (e x x))) as [y p].
+    refine (e y y; _).
+    set (foo := f (e y y)).
+    rewrite p. unfold foo. reflexivity.
+  Admitted.
+
+End Lawvere.
+
+Section Cantor.
+  Context {A : Type}.
+
+  Lemma negb_not_fixed : not (exists x, negb x = x).
+  Proof.
+    intro.
+    destruct X as [x p].
+    revert p. 
+    apply (not_fixed_negb x).
+  Qed.
+
+  Corollary cantor (f : A -> (A -> Bool)) :  not (IsSurjection f).
+  Proof.
+    intro X.
+    pose ((@Lawvere _ _ f X negb)) as bar.
+    apply (negb_not_fixed bar).
+  Qed.
+
+End Cantor.
--- a/contrib/SmallCategory.v
+++ b/contrib/SmallCategory.v
@@ -77,18 +77,6 @@ Record ThinCat := {
   thin : IsThinCat thin_cat;
 }.
 
-Lemma exp_card_leq `{Univalence} {a b : Card} (e : leq_card a b) (c : Card) 
-  : leq_card (exp_card c a) (exp_card c b).
-  Proof.
-    revert e; strip_truncations.
-    intro; strip_truncations; apply tr.
-    induction e as [e_i e_isinj].
-    srefine (fun f => e_i o f; _).
-    intros f g h; apply path_forall.
-    intro x; apply e_isinj.
-    apply (ap10 h).
-  Defined.
-
 Section Freyd.
   Context `{Univalence}.
   Context `{ExcludedMiddle}.
--- a/contrib/SmallCategory.v
+++ b/contrib/SmallCategory.v
@@ -4,7 +4,7 @@ Require Import Functor.
 Require Import Limits.Core.
 Require Import Spaces.Card.
 Require Import HIT.Truncations.
-
+Require Import ExcludedMiddle.
 
 Set Universe Polymorphism.
 Set Implicit Arguments.
@@ -65,6 +65,7 @@ Record ThinCat := {
 Section freyd.
 
 Context `{Univalence}.
+Context `{ExcludedMiddle}.
 
 (**
   Prop 3.7.3 Category theory in context, Riehl
@@ -83,9 +84,9 @@ Context `{Univalence}.
  
  **)
 
-Proposition Freyd_preorder_limit (k : Card) (C : kSmallCategory k) 
-  {h : HaskLimits k C} 
-  : IsThin C.
+Proposition Freyd_preorder_limit 
+  (k : Card) (C : kSmallCategory k) {h : HaskLimits k C} : 
+    IsThin C.
 Proof.
   remember (CatCard C) as l eqn:p.
   intros X Y f g; srapply (BuildContr).
--- a/contrib/SmallCategory.v
+++ b/contrib/SmallCategory.v
@@ -1,0 +1,95 @@
+Require Import HoTT.Basics HoTT.Types.
+Require Import Category.Core Category.Strict.
+Require Import Functor.
+Require Import Limits.Core.
+Require Import Spaces.Card.
+Require Import HIT.Truncations.
+
+
+Set Universe Polymorphism.
+Set Implicit Arguments.
+Generalizable All Variables.
+Set Asymmetric Patterns.
+
+Section Small_Category.
+
+Context `{Univalence}.
+
+(* A small category is a strict category *)
+
+(* Cardinality of a small category *)
+Definition CatCard (C : StrictCategory) : Card := 
+  tr (BuildhSet (exists (X Y : C), C.(morphism) X Y)).
+
+(* A k-small category is a small category with object set 
+   cardinality less than k *)
+Definition IsSmall (k : Card) (C : StrictCategory): hProp :=
+  leq_card (CatCard C) k.
+
+Record kSmallCategory (k : Card) := {
+  C :> StrictCategory;
+  k_small : IsSmall k C
+}.
+
+Context {C : PreCategory}.
+Context {strc : IsStrictCategory C}.
+
+Definition DiagramIskSmall {D : PreCategory} (k : Card) (F : Functor C D) : hProp
+ := IsSmall k (Build_StrictCategory C strc).
+
+Record kSmallDiagram (k : Card) (D : PreCategory) := {
+  F :> Functor C D;
+  isks : DiagramIskSmall k F
+}.
+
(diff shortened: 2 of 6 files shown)


mikeshulman:
There's nothing nonconstructive about Cantor's theorem; it's just a [proof of negation](http://math.andrej.com/2010/03/29/proof-of-negation-and-proof-by-contradiction/).

mikeshulman:
Ah, well, if you state it as `a < 2^a` with 2 meaning the cardinality of `bool`, then the `a <= 2^a` part is nonconstructive.  But the interesting part of Cantor's theorem that `a <> 2^a` is still valid constructively even if 2 means the cardinality of `bool` (although one might argue that `a < Prop^a` is a more constructively relevant statement, and in that case there's no problem with constructivity even for the `<=` direction).

mikeshulman:
Freyd's theorem, on the other hand, is known to be false constructively: there are toposes containing small complete non-thin categories (notably the internal category of "modest objects" in the effective topos).

I submitted a PR to your branch that completes the proof of Cantor's theorem, both the constructive part and the non-constructive part.

andrejbauer:
Here is a [more relevant blog post](http://math.andrej.com/2007/04/08/on-a-proof-of-cantors-theorem/) of mine. Cantor's theorem is not about `2^A` but rather about `P(A)`, the powerset. These two are not the same, as the first one is the type of decidable predicates and the second one of all predicates (nontheless, Cantor's theorem holds in both forms). If anything is going to get formalized here, I would suggest first formalizing Lawvere's theorem (see blog post), from which we can then easily derive various forms of Cantor's theorem.

Also, please note that there is no reasonable constructive notion of cardinality. Assuming there is a non-constant function `card : P(unit) -> nat` already implies excluded middle. In addition, it is constructively consistent to assume that there is an injective function from `nat -> nat` to `nat`.

theories/Spaces/Card.v
```diff
@@ -4,6 +4,8 @@ Require Import HoTT.Basics HoTT.Types HoTT.HSet HoTT.TruncType.
 Require Import HoTT.Classes.interfaces.abstract_algebra.
 Require Import HoTT.HIT.Truncations.
 
+Require Import ExcludedMiddle.
+
```
andrejbauer: Let's not do this.

Alizter:
>I would suggest first formalizing Lawvere's theorem (see blog post), from which we can then easily derive various forms of Cantor's theorem.

I have done this in HSet.v.

mikeshulman:
I don't agree that there is no reasonable constructive notion of cardinality.  The one in the book makes perfect sense to me.  It's true that not every subset of a finite set is finite, constructively, but that doesn't mean we can't talk about their cardinalities, just that the finite cardinalities are not a downset in the poset of cardinalities.

mikeshulman:
If Lawvere's theorem doesn't require any hSet hypotheses (which seems reasonable to me), maybe we should put it somewhere else than hSet.  I can't think of anywhere else that it belongs though.  Maybe a dedicated file like `FixedPoint.v`?

Alizter:
I have moved Lawvere's theorem and £Cantor's theorems" to `contrib/FixPoint.v`. I have relaxed the conditions for Cantor as well making it a theorem about types in general.

The surjective/injective stuff in HSet.v is about general types. I think we should get rid of these and instead use the HoTT versions: (-1)-truncated/connected fibers. Maybe plop those in a file somewhere too?

It then seems the original cardinality versions  of Cantor will be "functorial" applications of `tr` to some inj/sur function of sets.

Finally in order to prove Freyd's theorem I will need to get the universal property of Powers written out. Then hopefully the proof follows by contradiction from there.

Overall I am trying to see how much locally presentable/accessable category theory can be done, with my distant view on an adjoint functor theorem in HoTT.

mikeshulman:
The notion of injectivity in HSet.v is mainly about hsets.  It so happens that the notion can be defined for general types, but it is not very interesting or useful for non-hsets, so it belongs in HSet.v.  If you're doing something that makes sense for general types, then by all means instead use `IsEmbedding` instead (defined in `Basics/Trunc.v` to mean (-1)-truncated fibers).

Surjectivity, by contrast, is literally defined (for all types) to be the same as (-1)-connectedness (`IsSurjection` in `HIT/Truncations.v`).

I am excited to see some actual category theory being developed over this library!

Alizter:
So the `issur` in HSet.v is defined for general types but this will not follow from `IsSurjective`. So as it stands its a bit useless in that generality (and misleading). I would recommend strengthening the conditions to make them hSet's instead.

In my latest commit in `FixPoint.v` you can see why this doesn't work. Which also means Lawvere might not be true for general types. If this is the case we can put all of FixPoint.v back into HSet.v...

Awaiting your thoughts.

mikeshulman:
Right, your `issur` is the wrong definition, because `exists` in Coq refers to an untruncated Sigma-type.  As I said, the correct definition of surjectivity is `IsSurjective`.  This is equivalent to `forall y:B, hexists {x:A & f x = y}`; see `BuildIsSurjection` in `HIT/Truncations.v`.  I am not sure what you mean by "strengthening the conditions" -- what conditions?

Alizter:
> Right, your `issur` is the wrong definition, because `exists` in Coq refers to an untruncated Sigma-type. As I said, the correct definition of surjectivity is `IsSurjective`. This is equivalent to `forall y:B, hexists {x:A & f x = y}`; see `BuildIsSurjection` in `HIT/Truncations.v`. I am not sure what you mean by "strengthening the conditions" -- what conditions?

By conditions I meant make them hSets but now reading over it I realise I have made an error. I will look at it another time.

Alizter:
In this commit I have chosen a definition of regular cardinal to formalise. How are set families (covering spaces) best handled in this library? At the moment writing `F : X -> hSet` for a hSet X gives you some messy terms.

How can we go about constructing `aleph_w` and show it isn't regular. Showing that `aleph_0` is regular is basically already done in the HoTT book. I haven't actually added a condition for the infinite cardinal requirement which may also need to be thought about.

mikeshulman:
I would write either `F : X -> hSet` or `(F : X -> Type) {fset : forall x, IsSet (F x)}`.  The latter is generally cleaner, as you noted, but sometimes the former is necessary if you want to use the homotopy type of `hSet`.

There isn't really a good constructive definition of any alephs above aleph0, as far as I know.  The problem is that aleph_(n+1) is defined to be the _least_ cardinal greater than aleph_n, but constructively the cardinals are not classically-well-ordered, so there may not be any such.  But you could define beth_(n+1) = P(beth_n) and show that beth_omega is not regular.

By the way, as far as I know it is a rather open problem how to deal with locally presentable / accessible categories in constructive mathematics.  Which doesn't mean you shouldn't work on it, it just means you'll either have to do some new mathematics (rather than just formalizing mathematics that's already known) or assume excluded middle (and perhaps choice).

mikeshulman:
The definition of regular cardinal that you chose is indeed probably the best one.  However, I believe that for most _practical_ purposes, in constructive mathematics a "regular cardinal" is not actually what you want.  What you want instead is a "regular family of sets": a small subtype of `hSet` that is closed under Sigma-types.  In classical mathematics (almost? I forget the details) any regular family of sets consists of the sets of cardinality less than some regular cardinal, but constructively there's no reason for that to be true, among other reasons because cardinals are no longer totally ordered.

contrib/FixPoint.v
```diff
@@ -0,0 +1,62 @@
...
+
+Lemma isSur2 {A B} {f : A -> B} {sur : IsSurjection f} :
+  forall y, exists x, f x = y.
+Proof.
+  intro.
+  destruct (sur y) as [c d].
+  set (c' := c).
+  apply (untrunc_istrunc) in c'.
+  + apply c'.
+Admitted.
```
mikeshulman: You're not going to be able to prove this; it's essentially the axiom of choice.

mikeshulman:
I also suggest that you use the notation `{x:A & P x}` rather than `exists x:A, P x`.  The latter is potentially confusing because it looks like a proposition, but actually it isn't (the proposition is `hexists (fun x:A => P x)` or `merely {x:A & P x}`.

Probably a note about this should be added to `STYLE.md`.

Alizter:
`Syntax error: '&' or '|' expected after [constr:operconstr level 200] (in [constr:operconstr]).`

Alizter:
Okay I have worked it out. For some reason `Require Import HoTT.Classes.interfaces.abstract_algebra.` is breaking the notation.

Alizter:
Some progress made on Freyd's theorem. I am however stuck on the universal property of the power object in Powers.v. The proof from Riehls book uses the fact that the covariant hom functor is continuous.

I have proven a lemma `bool_leq_at_least_two` about having at least two distinct elements of a set means it has a larger cardinality than Bool. There is another cardinality lemma I added in SmallCategory.v: `exp_card_leq` that `exp_card` preserves inequality. I have no clue how to go about proving this however.

TODO:
* Prove universal property of power object
* Prove `exp_card_leq` and move it into Card.v

contrib/SmallCategory.v
```diff
@@ -0,0 +1,120 @@
...
+
+  Context (C : PreCategory).
+  Context `{IsStrictCategory C}.
+
+  Context (k : Card).
+  Context `{IskSmall C k}.
+  Context {h : kComplete _ k C}.
+
+  Lemma exp_card_leq {a b : Card} (e : leq_card a b) (c : Card) :
+    leq_card (exp_card a c) (exp_card b c).
```
mikeshulman: Just strip truncations and expand the definitions: every function c->a can be composed with your injection a->b to give a function c->b.

mikeshulman: Try `revert e` before `strip_truncations`.  The problem (at least, the problem I had at first) is that `strip_truncations` can't strip `a` and `b` because `revert_opaque` isn't smart enough to revert `e` along with `a` and `b`.  That seems like a bug with the tactic that it would be nice to fix.

Alizter: I've been going at it for a few hours now and I can't seem to get past this last bit.
```
Lemma exp_card_leq `{Univalence} {a b : Card} (e : leq_card a b) (c : Card) :
    leq_card (exp_card c a) (exp_card c b).
  Proof.
    revert e; strip_truncations.
    intro; strip_truncations.
    apply tr; simpl in *.
    induction e as [e_i e_isinj].
    srefine (_; _); simpl.
    + refine (fun f => e_i o f).
    + unfold isinj in e_isinj.
```

mikeshulman: Well, of course you need function extensionality (`path_arrow`) to prove that two functions are equal.  Then use the opposite of function extensionality (`ap10`) on the hypothesis.

Alizter: Ah ok that has fallen into place now. I will try to clean up my proof now.

mikeshulman:
Is this ready to merge?  If so, I will look at it more carefully.  If not, please close the PR for now and open issues if you have specific requests for help.

PR #977, opened 2018-12-31
Hopf fibration

Alizter:
I have been following the Lean library but we will need some lemmas about connectivity in ReflectiveSubuniverse.v. I have added the statement but I couldn't prove it. I am also having some trouble with applying the flattening lemma.

I can't seem to progress in proving Lemma 8.5.3. I think it is because we don't have pushout support for the flattening lemma. (Which we should get round to doing soon).

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -414,3 +414,4 @@ theories/Classes/tests/ring_tac.v
 contrib/HoTTBook.v
 contrib/HoTTBookExercises.v
 contrib/Freudenthal.v
+contrib/Hopf.v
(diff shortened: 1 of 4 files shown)


mikeshulman:
I'm not sure what Travis's issue is.  Is `camlp5 macro files not supported anymore, please port searchdepend.ml4 to coqpp` the relevant error?

mikeshulman:
Closing this as discussed at #980.  If you have particular things you want to ask for help with, please open an Issue and be specific.  I will try to help as my time allows, and I hope that others may feel inspired to do so too.  (To repeat, I am very happy to have you contributing to the library -- I don't want to come off as unwelcoming!  I just want to maintain some organization in the repo -- we have too many old PRs and issues lying around as it is.)

Issue #978, opened 2019-01-02
How can we improve the library?


Alizter:
What features do we want/need to make doing homotopy theory easier? It is very easy to get overwhelmed with matches and transports making it tricky to prove much about higher dimensional paths.

If you look at the [lean library](https://github.com/leanprover/lean2/blob/master/hott/cubical/square.hlean) there are many lemmas about squares. At the moment we have a few lemmas, namely whiskerings, in PathGroupoid.v. These would be useful things to add to this library. It would make some of the proofs in my pull request Hopf.v much easier too. You need only look at the lean implementation to see the benefits.

This has been in my opinion, one of the main obstacle to doing homotopy theory in the coq library. The Agda and lean implementations can deal with these constructions more naturally, thus they are written down faster.

Another thing is HIT. If we had more tactics like induction with support for HIT that would make writing some proofs easier. Especially with things like case analysis. (But this is a lot to ask for and I am not the only one asking).

It would be nice to hear what other people want for this library. What they think might help if it were implemented. Just so we are all aware on what people want. :-)

mikeshulman:
The absence of lemmas about squares and cubes from this library is not a design choice, but an "early adopter problem" -- the basic structure of this library and its lemmas predates the [Licata-Brunerie paper](https://guillaumebrunerie.github.io/pdf/lb15cubicalsynth.pdf) that introduced the use of squares and cubes in Book HoTT.  That and the fact that no one has since been motivated to, and had the time to, implement them.  (The existing lemmas aren't too bad once you get used to them, but there are certainly applications for which squares and cubes are easier.)

A little while ago I started implementing some basic square and cube lemmas, but never polished them to the point of merging.  Since you seem to be interested in this, I went ahead and cleaned them up and made a PR #979.  If that seems like a good starting point, I think it would be great if you felt like adding to it with more lemmas and so on as in the agda and lean libraries.

mikeshulman:
Regarding induction tactics, I've never really felt the lack of them for HITs.  Can you give an example of what you mean?  In general Coq's induction tactic is basically just "apply thing_ind" with some extra fanciness that guesses the induction motive, and I doubt the latter would be very useful or possible for HITs; usually I think it's better to carefully choose the induction motive by hand.

Alizter:
When doing induction over a higher inductive type, I usually have to make an effort to turn transports into aps using `rewrite transport_paths_FlFr` or something. This almost always spews out silly terms like `bah @ 1` which I have to reduce later.

[...]

PR #979, opened 2019-01-03
some stubby cubical stuff

mikeshulman:
In partial response to #978, here are some lemmas about squares and cubes that I started working on a while ago.  They're very incomplete because I had a specific purpose in mind at the time rather than setting out to write a complete library for squares and cubes, but I did try to choose relatively predictable names for the lemmas and their arguments at least.

If this seems to be a worthwhile starting point, we could merge it and then others could work on improving and adding to it (i don't have the time to work on it further myself right now) -- that would probably be easier than trying to develop it it a branch or a fork and only merging it when it looks more complete.  It compiles and doesn't affect any existing code (so using it is completely optional).

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -26,6 +26,10 @@ theories/Types/Wtype.v
 theories/Types/Equiv.v
 theories/Types/Universe.v
 theories/Types.v
+theories/Cubical/DPath.v
+theories/Cubical/Square.v
+theories/Cubical/Cube.v
+theories/Cubical.v
 theories/BoundedSearch.v
 theories/EquivGroupoids.v
 theories/Fibrations.v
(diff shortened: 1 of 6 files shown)


Alizter:
I had a go redefining the definitions in Square.v using the interval type instead, but it seems to be more work. Perhaps if we wrote some clever tactics for dealing with it? 

A possible advantage of using interval with some clever tactics is that it will allow us to reason about higher cubes without writing them down explicitly.

I also tried labeling the points with a function `Bool -> Bool -> A` but this got tedious quickly. However the paths then become homotopies etc. which may be useful.

mikeshulman:
True cubical type theory uses an interval, but in Book HoTT the interval isn't useful for much; instead we generally define paths, squares, etc. as inductive families.

Alizter:
I am in favour of merging this at the moment. I think the dependent paths will be very useful.

mikeshulman:
Anyone else have a thought about merging this now (pro or con)?

mikeshulman:
Rebased.

Alizter:
@spitters We can merge this too if you are happy with it.

PR #980, opened 2019-01-06
Rezk completion

Alizter:
I have started a Rezk completion file. The goal is to prove theorem 9.9.5 from the HoTT book.

I have written some constructors down for the higher inductive type of objects of a Rezk completed category. The induction principle is tricker to write down however. I will also need eliminators for the path constructors.

There are also some suggestions of what the finished theorem will look like, but not much can be done without finishing the HIT.

I don't believe anybody has actually written down this version of the proof in a proof assistant yet.

### DIFF
--- a/contrib/Rezk.v
+++ b/contrib/Rezk.v
@@ -1,0 +1,69 @@
+Require Import Basics.Overture.
+Require Import Category.Core.
+Require Import Category.Univalent.
+Require Import Category.Morphisms.
+
+Require Import Functor.Core.
+Require Import Functor.Attributes.
+Require Import Yoneda.
+
+Set Universe Polymorphism.
+Set Implicit Arguments.
+Generalizable All Variables.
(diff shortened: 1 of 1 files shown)


mikeshulman:
I'm happy that you're working on formalizing all of this stuff.  However, I wonder what exactly is your goal with all these pull requests?  Do you want to get this work-in-progress merged into the main library?  I think it's more common to mostly finish a particular development in your own fork before submitting a pull request.

Alizter:
I submit pull requests of things that I would like to see in the library. It also makes it easier for other to contribute.

The reason I have made this one for example is so I can get help writing out the induction rules. After that I don't think the proof is too difficult.

My Hopf fibration pull request is mostly done, I just have a few proofs to finish and and cleaning up to do but it is almost ready for merging.

But if you think I am submitting too much, let me know, perhaps I am being a bit noisy.

PR #981, opened 2019-01-07
Fix implicit arguments to be compatible with coq/coq#9314

jashug:
The default choice of implicit arguments for Instances combined with the `Implicit Arguments` option is currently inconsistent with other commands like `Definition`. coq/coq#9314 changes this behavior, but https://github.com/HoTT/HoTT/blob/31c31c5c7673141979e274bc0afcc21417af9698/theories/Categories/Pseudofunctor/RewriteLaws.v#L44 is incompatible with the new default choice of implicit arguments for `iso_whisker_l`
Here we explicitly specify the implicit arguments for nearby instances as well to maintain compatibility.


mikeshulman:
LGTM.

PR #986, opened 2019-01-19
Changed definition of suspension to be in terms of pushout

Alizter:
Changed the definition of suspension to be defined in terms a pushout. It should lose none of it's functionality. This will also make it possible to use flattening lemma on pushouts.

I have also added universe parameters to coequalizers as they were omitted before.

### DIFF
--- a/theories/HIT/Suspension.v
+++ b/theories/HIT/Suspension.v
@@ -2,16 +2,16 @@
 
 (** * The suspension of a type *)
 
-Require Import HoTT.Basics HoTT.Types.
+Require Import Basics Types.
+Require Import HIT.Pushout.
 Require Import Pointed NullHomotopy.
-
 Local Open Scope path_scope.
 Local Open Scope pointed_scope.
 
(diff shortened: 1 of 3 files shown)


theories/HIT/Suspension.v
```diff
@@ -38,7 +38,38 @@ Axiom Susp_ind_beta_merid : forall {X : Type} (P : Susp X -> Type)
...
+
+Definition Susp_ind_beta_merid {X : Type}
+  (P : Susp X -> Type) (H_N : P North) (H_S : P South)
+  (H_merid : forall x:X, (merid x) # H_N = H_S) (x : X)
+  : apD (Susp_ind P H_N H_S H_merid) (merid x) = H_merid x.
+Proof.
+  serapply pushout_ind_beta_pp.
+Defined.
+
+(** But we want to allow the user to forget that we've defined the circle in that way. *)
```
mikeshulman: should be "...defined the suspension in that way."  (-:

PR #988, opened 2019-01-22
Bump versions to 8.9.0

JasonGross:
Now that 8.9.0 has been released, we can bump the versions of things to 8.9.

Once this is merged, I will tag V8.9, and I will make a PR to bump the version of Coq to current master, and then we can merge #964, #971, and maybe #961.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,7 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = master
+	branch = v8.9
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
 	url = https://github.com/JasonGross/coq-scripts.git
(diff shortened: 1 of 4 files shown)


mikeshulman:
LGTM

.gitmodules
```diff
@@ -1,7 +1,7 @@
 [submodule "coq-HoTT"]
 	path = coq-HoTT
 	url = https://github.com/coq/coq.git
-	branch = master
+	branch = v8.9
```
Alizter: Will this have to be changed to master again when coq master is 8.9?

JasonGross: This will be changed to master again when we start tracking master rather than 8.9 (which will happen very shortly after this PR is merged)

SkySkimmer:
8.9 isn't quite released yet due to build issues on osx though.

spitters:
LGTM. But let's wait for @SkySkimmer to answer

PR #989, opened 2019-01-24
Make trivial instances explicit

maximedenes:
This is in preparation for coq/coq#9274.

Should be backward compatible but that remains to be tested.

### DIFF
--- a/theories/Categories/ExponentialLaws/Law0.v
+++ b/theories/Categories/ExponentialLaws/Law0.v
@@ -28,7 +28,7 @@ Section law0.
   Variable C : PreCategory.
 
   Global Instance IsTerminalCategory_functors_from_initial
-  : IsTerminalCategory (0 -> C).
+  : IsTerminalCategory (0 -> C) := {}.
 
   (** There is only one functor to the terminal category [1]. *)
   Definition functor : Functor (0 -> C) 1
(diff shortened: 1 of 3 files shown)


Alizter:
What is going on here? Can you add some more explanation?

SkySkimmer:
When a class has no content, such as
~~~coq
Class IsTerminalCategory (C : PreCategory)
      `{Contr (object C)}
      `{forall s d, Contr (morphism C s d)} := {}.
~~~
then `Instance : IsTerminalCategory C.` will define it without opening a proof (assuming it manages to fill the implicit arguments of the type). To make it easier to predict when proofs are opened, we are planning to change this behaviour.

maximedenes:
The two errors I get seem unrelated:
```
camlp5 macro files not supported anymore, please port searchdepend.ml4 to coqpp
```
and
```
make[1]: *** No rule to make target `grammar/grammar.cma'.  Stop.
```

is HoTT's CI working properly?

JasonGross:
#988 fixes the CI.  Until that is merged, we expect dpdgraph to fail.  (If 8.9.0 is not going to be actually released soon, or if the tag is going to change, we could merge the bits of that PR that fix the CI without merging the version bump.)

JasonGross:
The issue with master is half Coq's fault; there is still no install target for coqlight (https://github.com/coq/coq/issues/5272), and I guess `grammar/grammar.cma` recently got replaced by something else on master?  (What did it get replaced by?)

PR #990, opened 2019-01-24
Bump the version of Coq we are using to current master

JasonGross:
This includes the changes from #964 and #961.  The changes from #964 are needed for our code to compile on master at all.  (The changes from #961 are not needed, but this seems like a good place to include the infrastructure updates that result from Coq changing its directory names across versions.)

This should only be merged if Travis fully passes.

Closes #961
Closes #964


spitters:
LGTM, if travis is happy

mikeshulman:
LGTM too.

Issue #991, opened 2019-01-25
Adding and removing hints in the core database implicitly is deprecated


Alizter:
>Warning: Adding and removing hints in the core database implicitly is
deprecated. Please specify a hint database.

I keep getting this warning when compiling the library after the push to 8.9. What should be done about these? They appear in Overture.v a lot.

JasonGross:
We can just add `: core` to all the `Hint ...` statements, probably.

Alizter:
If you do end up doing this it breaks a bunch of files. See #992.

PR #992, opened 2019-01-29
changed implicit hints to reference core

Alizter:
As talked about in #991.

However now we cannot build. First problem is in Sum.v and is not the last.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -4,6 +4,9 @@
 (** Import the file of reserved notations so we maintain consistent level notations throughout the library *)
 Require Export Basics.Notations.
 
+(** Change in introduction patterns not adding an implicit [] *)
+Global Unset Bracketing Last Introduction Pattern.
+
 (** ** Type classes *)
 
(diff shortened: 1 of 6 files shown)


JasonGross:
That seems suspicious.  If you remove all the changes and just stick `Hint Resolve Set : core.` at the top of `Overture.v`, does that break everything as well?  (Current hypothesis: the first `Hint Resolve` creates the hint database in a way that breaks things.)

Alizter:
@JasonGross `Set cannot be used as a hint.`

JasonGross:
Er, just pick anything that exists.  Maybe, `Inductive foo := bar. Hint Resolve bar : core.`

Alizter:
So I was getting:

```
File "./theories/Basics/Overture.v", line 8, characters 0-50:
Warning: Option Bracketing Last Introduction Pattern is deprecated
[deprecated-option,deprecated]
```

So I removed the culprit without thinking much. Apparently we need it to make Sum.v work but this is another issue.

This has nothing to do with hint databases so I think we can merge it. I will check the travis log just to see if we missed any however.

PR #994, opened 2019-01-31
Remove Bracketing Last Introduction Pattern, close #993

### LINKED ISSUE #993: Option Bracketing Last Introduction Pattern is deprecated
```
File "./theories/Basics/Overture.v", line 8, characters 0-50:
Warning: Option Bracketing Last Introduction Pattern is deprecated
[deprecated-option,deprecated]
```

as found out in #992 this is essential for Sum.v and a few other files to compile. What is the reason for this to still be used and what should we do about it?

mikeshulman:
I don't know if anyone is listening, but I think this is an argument against relying on fancy tactic stuff like this.  Not only does it make the code harder to step through and understand, but it renders it fragile to changes to the details of tactics, and when that happens it is hard to figure out how to fix it because it's hard to step through and understand.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -4,9 +4,6 @@
 (** Import the file of reserved notations so we maintain consistent level notations throughout the library *)
 Require Export Basics.Notations.
 
-(** Change in introduction patterns not adding an implicit [] *)
-Global Unset Bracketing Last Introduction Pattern.
-
 (** ** Type classes *)
 
 (** This command prevents Coq from trying to guess the values of existential variables while doing typeclass resolution.  If you don't know what that means, ignore it. *)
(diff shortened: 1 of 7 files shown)


Alizter:
I agree there is a problem with the use of `intro []`. I wonder if it might be worth going through the library and finding "poorly written" code. Then people would know what to fix. Ones I have found which are difficult to read are the old stuff like Fibrations.v, the colimits library, Modalities/Topological.v etc.

mikeshulman:
I don't have a problem with ordinary `intro []` which means to introduce something and immediately destruct it.  I do agree that Fibrations and colimits are hard to read, if anyone wants to clean them up that would be great.  I don't have a problem with Topological myself (unsurprisingly, since I wrote it), but if you have suggestions to improve it somehow that's great too.

Alizter:
Can we resolve and merge this?

spitters:
"This branch has conflicts that must be resolved"

mikeshulman:
Rebased against master.  (Some of the changes overlapped with #998.)

PR #995, opened 2019-01-31
Fix bracketing

Alizter:
This is essentially #994 but I would argue mine is more readable. I will close this one if Mike's is preferred.

### DIFF
--- a/theories/Categories/Category/Sigma/OnObjects.v
+++ b/theories/Categories/Category/Sigma/OnObjects.v
@@ -66,7 +66,8 @@ Section sigT_obj.
     /\ sigT_functor_obj_inv o sigT_functor_obj = 1.
   Proof.
     split; path_functor; trivial.
-    repeat (intros [] || intro || apply path_forall).
+    repeat ( apply path_forall || serapply sig_ind || intro ).
+    destruct x1 as [x1 x2], x2.
     reflexivity.
   Qed.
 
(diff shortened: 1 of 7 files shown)


theories/Categories/Category/Sigma/OnObjects.v
```diff
@@ -66,7 +66,8 @@ Section sigT_obj.
     /\ sigT_functor_obj_inv o sigT_functor_obj = 1.
   Proof.
     split; path_functor; trivial.
-    repeat (intros [] || intro || apply path_forall).
+    repeat ( apply path_forall || serapply sig_ind || intro ).
+    destruct x1 as [x1 x2], x2.
```
JasonGross: Please do not use autogenerated names

Alizter:
@mikeshulman What's wrong with adjointifying?

Alizter:
@mikeshulman Regarding bullets, I usually don't bother if its one line, but if its more than one line then it is worth showing.

Alizter:
@mikeshulman I am fine with using bullets everywhere.

Regarding equivalences: I don't see any reason not to use `equiv_adjointify`. I can't think of a case where you would need for an equivalence made with adjointify and one built by hand to be definitionall equal. Surely if we all used `equiv_adjointify` this would be the case anyhow? I think I am being a bit naive here however.

theories/Categories/Category/Sigma/OnObjects.v
Alizter: Ok I have rewritten this now.

mikeshulman:
See #825 .

Alizter:
@JasonGross Are the changes adequate?

PR #998, opened 2019-02-01
Use strict focusing (forced bullets / Default Goal Selector "!")

SkySkimmer:
Prompted by recent discussion around https://github.com/HoTT/HoTT/pull/995#issuecomment-459721744, we set Default Goal Selector to "!" (new to Coq 8.9).

This makes it an error to run a tactic when more than 1 goal is
focused unless using an explicit selector. ie

~~~coq
Goal True /\ True.
Proof.
  split.
  Fail exact I. (* this now fails *)
  1: exact I. (* this still works *)
Abort.

(* the expected way to write most proofs *)
Goal True /\ True.
Proof.
  split.
  - exact I.
  - exact I.
Abort.
~~~

`1:` is useful for instance when you don't want to bullet because
you're preparing your goals to run with `all:`, like in
`Univalent.equiv_iso_A'_eisretr_helper`.


contrib/Freudenthal.v
```diff
@@ -110,19 +110,21 @@ Proof.
...
-    apply isequiv_Trunc_functor, @isequiv_functor_sigma. refine _.
-    intros a. apply isequiv_cancelR. }
-  { hnf. apply Trunc_ind. intros ?; apply trunc_succ.
-    intros [x r]; simpl.
-    unfold functor_sigma; simpl.
-    symmetry.
-    exact (ap10 (wedge_incl_comp1 x0 x0 _ _ _ _ x) r). }
+  - intros x; generalize dependent n. intros [ | n'] imposs.
+    + destruct (imposs 1).
+    + intros ?. apply (@trunc_leq -1). * exact tt. * apply hprop_isequiv.
```
mikeshulman: I think it's more readable to always put bullets at the beginning of lines.  Sometimes in a situation like this where I don't want to use the extra vertical space I write `apply (@trunc_leq -1); [ exact tt | apply hprop_isequiv ].`.

I started marking all the places where this happened, but there are so many and you can probably find them just as easily yourself.

contrib/HoTTBookExercises.v
```diff
@@ -610,7 +611,7 @@ Proof.
   intros isHProp_A isProp_B nab.
   apply hprop_allpath. intros x y.
   destruct x as [a1|b1]; destruct y as [a2|b2]; apply path_sum; try apply path_ishprop.
-  exact (nab (a1,b2)). exact (nab (a2,b1)).
+  - exact (nab (a1,b2)). - exact (nab (a2,b1)).
```
mikeshulman: And here.

theories/EquivalenceVarieties.v
```diff
@@ -209,18 +209,18 @@ Proof.
...
 Proof.
   apply equiv_iff_hprop.
-  by apply isequiv_biinv.
-  intros ?.  split.
-  by exists (f^-1); apply eissect.
-  by exists (f^-1); apply eisretr.
+  - by apply isequiv_biinv.
+  - intros ?.  split.
+    + by exists (f^-1); apply eissect.
+           + by exists (f^-1); apply eisretr.
```
mikeshulman: The indentation is wrong here.

theories/HIT/Circle.v
```diff
@@ -142,12 +142,12 @@ Definition S1_encode_isequiv (x:S1) : IsEquiv (S1_encode x).
...
   (* Here we induct on [x:S1].  We just did the case when [x] is [base]. *)
-  refine (S1_ind (fun x => Sect (S1_decode x) (S1_encode x))
-    S1_encode_loopexp _ _).
-  (* What remains is easy since [Int] is known to be a set. *)
-  by apply path_forall; intros z; apply set_path2.
-  (* The other side is trivial by path induction. *)
-  intros []; reflexivity.
+  - refine (S1_ind (fun x => Sect (S1_decode x) (S1_encode x))
+                   S1_encode_loopexp _ _).
+      (* What remains is easy since [Int] is known to be a set. *)
```
mikeshulman: Indentation is wrong for these lines.

theories/HIT/Colimits/CommutativeSquares.v
```diff
@@ -70,22 +70,22 @@ Proof.
   set (p := (ap wB (ap (wB ^-1) (ap f' (eisretr wA a)))
             @ eisretr wB (f' a))).
   path_via ((eisretr wB _)^ @ p).
-  apply whiskerR.
+  - apply whiskerR.
     apply moveR_pM.
     path_via ((eisretr wB (f' (wA (wA ^-1 a))))^ @
-       ap (wB o wB ^-1) (wf ((wA ^-1) a))).
-      rewrite ap_V. rewrite <- eisadj.
+                                                   ap (wB o wB ^-1) (wf ((wA ^-1) a))).
```
mikeshulman: Indentation

Alizter:
@JasonGross What is it about curly braces that you prefer?

theories/EquivalenceVarieties.v
SkySkimmer: emacs auto-indent seems to get confused by the exists on the previous line (maybe because company-coq prettifies it?)
Will fix

mikeshulman:
One nice thing about curly braces is that you can easily jump from the start to the end with `C-M-f` and `C-M-b` (or whatever command your preferred editor uses instead).  Also curly braces can be cut and pasted into a different place in a proof without being sensitive to whether the same bullet symbol was already used elsewhere in the nesting.  So I also prefer curly braces for long sub-proofs, although for fairly short ones I find them too visually heavy relative to bullets.

theories/EquivalenceVarieties.v
mikeshulman: Yes, I've encountered that bug myself.

contrib/Freudenthal.v
SkySkimmer: >I think it's more readable to always put bullets at the beginning of lines.

Changed (should be all of them but ofc I may have missed some)

theories/Categories/UniversalProperties.v
```diff
@@ -85,13 +85,13 @@ Section UniversalMorphism.
...
-          (exists (@center _ UniversalProperty).1).
-          abstract (progress rewrite ?right_identity, ?left_identity;
-                    exact (@center _ UniversalProperty).2).
+        - apply CommaCategory.issig_morphism.
+        - apply contr_inhabited_hprop.
+          + abstract univ_hprop_t UniversalProperty.
+          + (exists tt).
+            (exists (@center _ UniversalProperty).1).
+            abstract (progress rewrite ?right_identity, ?left_identity;
+                      exact (@center _ UniversalProperty).2).
```
Alizter: Indent

SkySkimmer: It's correct AFAICT

Alizter: That's fine then.

Alizter:
LGTM

PR #1004, opened 2019-02-13
Added definition of homotopy group

Alizter:
Defined the homotopy group of a pointed type. It uses the Groups structures found in mathclasses. I also modified `iterated_loops` in `Pointed.v` so that I can have loops on the outside, allowing path lemmas to be used.

The constructors for mathclasses are very slow, I am not really sure why? Perhaps it is my use of `serapply`?

I have also started a `Homotopy` foldeer in `theories`, later on we can move more synthetic homotopy theory relevent things here. At least it is a start.

### DIFF
--- a/theories/Homotopy/HomotopyGroup.v
+++ b/theories/Homotopy/HomotopyGroup.v
@@ -46,7 +46,6 @@ Section HomotopyGroups.
   Proof.
     refine (tr _).
     destruct n, gt.
-    rewrite iterated_loops_unfold.
     exact idpath.
   Defined.
 
--- a/theories/Homotopy/HomotopyGroup.v
+++ b/theories/Homotopy/HomotopyGroup.v
@@ -11,7 +11,7 @@ Section HomotopyGroups.
 
   Context
    `{Univalence}
-    {n : nat}
+    (n : nat)
     (X : pType).
 
   Definition Pi := Tr 0 (iterated_loops n X).
--- a/theories/Homotopy/HomotopyGroup.v
+++ b/theories/Homotopy/HomotopyGroup.v
@@ -10,7 +10,6 @@ Require Import DProp.
 Section HomotopyGroups.
 
   Context
-   `{Univalence}
     (n : nat)
     (X : Type)
    `{IsPointed X}.
@@ -34,11 +33,8 @@ Section HomotopyGroups.
   Proof.
     intros x y z.
     strip_truncations.
-    unfold PiOp.
-    simpl.
-    serapply equiv_path_Tr.
-    refine (tr _).
-    simpl.
+    unfold PiOp; cbn.
+    apply ap.
     destruct n, gt.
     refine ((concat_pp_p _ _ _)^).
   Defined.
@@ -54,10 +50,8 @@ Section HomotopyGroups.
   Proof.
     intro x.
     strip_truncations.
-    unfold PiOp.
-    simpl.
-    serapply equiv_path_Tr.
-    refine (tr _).
+    unfold PiOp; cbn.
+    apply ap.
     destruct n, gt.
     apply concat_1p.
   Defined.
@@ -66,10 +60,8 @@ Section HomotopyGroups.
   Proof.
     intro x.
     strip_truncations.
-    unfold PiOp.
-    simpl.
-    serapply equiv_path_Tr.
-    refine (tr _).
+    unfold PiOp; cbn.
+    apply ap.
     destruct n, gt.
     apply concat_p1.
   Defined.
@@ -87,11 +79,8 @@ Section HomotopyGroups.
   Proof.
     intro x.
     strip_truncations.
-    unfold PiOp.
-    simpl.
-    serapply equiv_path_Tr.
-    simpl.
-    refine (tr _).
+    unfold PiOp; cbn.
+    apply ap.
     destruct n, gt.
     apply concat_Vp.
   Defined.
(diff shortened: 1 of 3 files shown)


theories/Homotopy/HomotopyGroup.v
```diff
@@ -0,0 +1,129 @@
...
+Require Import abstract_algebra.
+Require Import Spaces.Nat.
+Require Import DProp.
+(* In this file we define homotopy groups *)
+
+Section HomotopyGroups.
+
+  Context
+   `{Univalence}
+    {n : nat}
```
mikeshulman: Shouldn't the argument `n` be explicit?

mikeshulman:
Thanks!  I worry that `Homotopy` is too generic-sounding of a name for the folder, though.

theories/Homotopy/HomotopyGroup.v
Alizter: Yes, I missed that, hotfix on the way.

Alizter:
Every other formalisation has a folder named homotopy. I don't think its too hard to guess what its about (which is synthetic homotopy theory).

mikeshulman:
Okay `*`shrug`*`

theories/Homotopy/HomotopyGroup.v
```diff
@@ -0,0 +1,129 @@
...
+    simpl.
+    serapply equiv_path_Tr.
+    simpl.
+    refine (tr _).
+    destruct n, gt.
+    apply concat_pV.
+  Defined.
+
+  (* Why is this so slow? *)
+  Global Instance pi_is_Group : @Group Pi PiOp PiUnit PiInverse.
```
SkySkimmer: If you use apply it's not slow.
Or you could do 
~~~coq
  Global Instance pi_is_Group : @Group Pi PiOp PiUnit PiInverse.
  Proof.
    repeat split.
    - exact _.
    - exact PiOp_assoc.
    - exact PiOp_leftId.
    - exact PiOp_rightId.
    - exact PiOp_leftInv.
    - exact PiOp_rightInv.
  Qed.
~~~
Or define the intermediate instances ie f7b114b78b74b78863f7027b864225a98a4410dd

mikeshulman: Interesting.  Any idea *why* it's slow?

Alizter: In this situation, Qed or Defined?

SkySkimmer: When doing `refine (Build_Group __)` (imagine I put the correct number of `_` there) it's going to trigger typeclass search for a `Monoid Pi PiOp PiUnit` which takes some time to fail.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -0,0 +1,122 @@
...
+    unfold PiOp.
+    simpl.
+    serapply equiv_path_Tr.
+    simpl.
+    refine (tr _).
+    destruct n, gt.
+    apply concat_pV.
+  Defined.
+
+ Global Instance pi_is_Group : @Group Pi PiOp PiUnit PiInverse.
```
SkySkimmer: Incorrect indent (1 space, should be 2)

Alizter:
I just noticed that it is very rare that we have a pType. It is much more convenient to have an instance of `IsPointed` floating around.

mikeshulman:
I agree that it is generally better to have `IsPointed` hypotheses, just as we do with `IsTrunc` and other examples like that.  Only when we really need to for some reason (e.g. to talk about the identity types of `pType`) do we bundle it up.

Alizter:
At some point we should change `loops` to do the same too then. I find it very cumbersome having everything packaged up when working with it.

Alizter:
Ok loops has been changed too. I managed to compile, so coq doesn't have an issue coercing when needed. There may be some redundent Build_pTypes around however.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -0,0 +1,123 @@
...
+Require Import Pointed.
+Require Import abstract_algebra.
+Require Import Spaces.Nat.
+Require Import DProp.
+(* In this file we define homotopy groups *)
+
+Section HomotopyGroups.
+
+  Context
+   `{Univalence}
```
SkySkimmer: You don't need univalence for these proofs. It's used for `equiv_path_Tr` but `serapply equiv_path_Tr; refine (tr _).` in the proofs can be replaced by `apply ap.`, and that's the only way you use `equiv_path_Tr`.

theories/Homotopy/HomotopyGroup.v
SkySkimmer: >In this situation, Qed or Defined?

No strong opinion, but if you use Qed for the Group proof you should use it for the intermediate proofs too.

theories/Homotopy/HomotopyGroup.v
Alizter: Nice, that make much more sense. I've pushed the change.

theories/Homotopy/HomotopyGroup.v
Alizter: I'll stick to `Defined`.

mikeshulman:
Yeah, in general I prefer to stick with `Defined` until and unless evidence emerges that a `Qed` will improve performance.

Alizter:
I've tried to use instances a bit more. This should make the code much easier to maintain too.

mikeshulman:
Great!  I think this is probably ready to merge, APOE?

PR #1006, opened 2019-02-20
Fix unfoldings of private recursors.

SkySkimmer:
After coq/coq#9609 the kernel will reject matches on private
inductives outside their home module. This breaks proofs that generate
such terms by over-reducing.

Since HoTT is the primary user of Private (AFAIK) please speak up if
that's not a desired change.

Should be backwards compatible.

### DIFF
--- a/theories/Classes/implementations/natpair_integers.v
+++ b/theories/Classes/implementations/natpair_integers.v
@@ -846,7 +846,7 @@ Proof.
 pose proof Z_ring.
 first [apply from_full_pseudo_ring_order@{UN UN UN UN UN UN UN Ularge}|
        apply from_full_pseudo_ring_order]; try apply _.
-apply apartness.strong_binary_setoid_morphism_commutative.
+srapply apartness.strong_binary_setoid_morphism_commutative.
 Qed.
 
 Goal FullPseudoSemiRingOrder Zle Zlt.
--- a/theories/Spaces/Card.v
+++ b/theories/Spaces/Card.v
@@ -93,10 +93,20 @@ Section contents.
 
   (** *** Properties of exponentiation *)
   Lemma exp_zero_card (a : Card) : exp_card 0 a = 1.
-  Proof. simpl_ops. reduce. symmetry. apply equiv_empty_rec. Defined.
+  Proof.
+    revert a. apply Trunc_ind;[exact _|].
+    intros a. simpl. unfold one_card,one.
+    f_ap. apply path_hset.
+    symmetry; apply equiv_empty_rec.
+  Defined.
 
   Lemma exp_card_one (a : Card) : exp_card a 1 = 1.
-  Proof. simpl_ops. reduce. symmetry. apply equiv_unit_coind. Defined.
+  Proof.
+    revert a. refine (Trunc_ind _ _).
+    intros a;simpl. unfold one,one_card.
+    f_ap. apply path_hset.
+    symmetry. apply equiv_unit_coind.
+  Defined.
 
   Lemma exp_one_card (a : Card) : exp_card 1 a = a.
   Proof. reduce. symmetry. apply equiv_unit_rec. Defined.
--- a/theories/Spaces/No/Core.v
+++ b/theories/Spaces/No/Core.v
@@ -1016,7 +1016,8 @@ Section RaiseSort.
         {{ (fun l => No_raise (xL l)) | (fun r => No_raise (xR r)) // rxcut }} }.
   Proof.
     eexists.
-    reflexivity.
+    rel_hnf. (* required to avoid exposing the private inductive *)
+    exact idpath.
   Qed.
 
   Definition No_raise_le (x y : GenNo S)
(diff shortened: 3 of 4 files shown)


mikeshulman:
You mean anyone using a private inductive outside its home module will have to be careful not to reduce it too far in order to avoid "seeing" a match on it, even if that match was actually written *inside* the home module?  That seems really annoying and unintuitive, especially for newcomers who want their HITs to "just work".

Would it be possible instead to mark all the matches on private inductives (written inside their modules) as something like `simpl nomatch`, so that they will never be reduced to expose the illegal match unless that match itself can also be reduced?

SkySkimmer:
> That seems really annoying and unintuitive

That's because Private Inductive is basically a hack. Currently it provides 0 actual guarantees. For instance 
~~~coq
Module Import M.
  Inductive Box (A:Type) := box : A -> Box A.
  Arguments box {_} _.

  Definition box_functor {A B} (f:A -> B) (b:Box A) : Box B :=
    box match b with box x => f x end.
End M.

Definition unbox {A} : Box A -> A.
Proof.
  assert (U : sig (fun unbox => forall b : Box A, box_functor (fun x => x) b = box (unbox b)));
    [|exact (proj1_sig U)].
  eexists. intros b. reflexivity.
Defined.
~~~
and in general the kernel doesn't care about Private, so it has no real meaning.

So we have a choice between staying in the current situation and basically hoping nobody gets through the holes, or trying to fix it. The proposed patch to Coq is the only idea I have for a fix.

>Would it be possible instead to mark all the matches on private inductives (written inside their modules) as something like simpl nomatch, so that they will never be reduced to expose the illegal match unless that match itself can also be reduced?

I don't know.

mikeshulman:
Well, you said to speak up if that's not a desired change, so I was speaking up.  It's one thing for the tacticized proofs in `Spaces/Card.v` that are impossible for me to understand anyway and actually do visibly expose the private match, but I would not intuitively expect simply saying `reflexivity` to ever "expose a private match".

SkySkimmer:
I was too agressive in my reply, sorry about that. Mostly I want to be clear if you "undesire" it enough to accept the current state.
>I would not intuitively expect simply saying reflexivity to ever "expose a private match".

reflexivity used to instantiate some hole (and more generally unification) is probably the easiest way to expose a private match. That's why I went for it in the `unbox` example I posted ;)
Indeed the 2 `reflexivity` I touched are this same `eexists; reflexivity.` code pattern.

mikeshulman:
I'd be interested to hear what others think.

JasonGross:
I am okay with this change, but defer to @mikeshulman.

Does marking the eliminator with `Strategy 100` (which says "unfold me after you unfold other things (with markings less than 100)") help at all?

Or what about marking the eliminators `Opaque` and only making them `Transparent ` in the specific places where we want them to reduce?

@mikeshulman note that `simpl nomatch ` only applies to `cbn` and `simpl `; I believe it does not impact unfolding/refolding during unification.

JasonGross:
One thing I want to check @SkySkimmer : without these changes, the error message shows up in the tactic immediately following the changed tactic, and not just at `Defined `, right?  (This is not one of those things, like guardedness, that the kernel enforces but that tactics don't check, right?)

SkySkimmer:
No, it only shows up at Defined. I haven't tried to add checks to tactics so maybe it's possible, but currently it isn't done.

JasonGross:
Travis fails because the html build job shoots up from 30 minutes to over 50 minutes.  Is this a Travis glitch, or is there a perf impact of this PR?

mikeshulman:
Can someone provide instructions that we could paste into `STYLE.md` telling people what to do when they encounter this error message?  I notice most of your changes are just adding `rel_hnf`.  Does this always work?  Why?

JasonGross:
> No, it only shows up at Defined

I think this is a very bad state of affairs (though possibly better then the current one).  This will make it impossible for beginners to debug issues, and I expect it will make it a pain for even more advanced users to debug issues.

mikeshulman:
I wouldn't be surprised if the `rel_hnf` in `No/Core` has a performance impact; the surreals seem to be pretty fragile performance-wise to being over-reduced.

SkySkimmer:
Perf impact was minimal to slightly faster when I ran IIRC.

I have no clue why rel_hnf works, when I encountered the error in `No_raise_cut` I was totally lost so I looked at `plus_inner_cut` since there's a comment about it, and the proof there uses `rel_hnf` so I tried it and it went through. It seems to work well for the issues in No, but that's pretty specific (for instance those terms are usually matches on the empty_for_no axiom) so I don't know if it generalizes.

>I expect it will make it a pain for even more advanced users to debug issues.

It is a pain, although how much of that is me knowing nothing about No / Card I can't tell.

mikeshulman:
I agree that the error only showing up at `Defined` time is bad.  However, I'm getting a little more amenable to the change overall.  It's encouraging that changes were only necessary to three developments, and two of those are using funky tactics while the other is the surreals (which are pretty funky already in other ways).  That makes me think that probably the average user is unlikely to ever encounter it.

SkySkimmer:
So merge or no merge?

spitters:
This seems like a reasonable restriction, but I do not really have a strong opinion.

Alizter:
It seems Travis doesn't have enough build time however.

mikeshulman:
I will support merging it if (1) Travis can build it and (2) some explanation and guidance is added to `STYLE.md`.

SkySkimmer:
I qon't have time/motivation to figure this out further.

PR #1007, opened 2019-02-20
Universal algebra: preliminary definitions.

andreaslyn:
This is the first of a series of commits implementing parts of universal algebra in HoTT. The complete code can be found at https://github.com/andreaslyn/hott-classes. It is based on math classes, https://github.com/coq-community/math-classes.

This commits contains the preliminary definitions. Following commits will contain definitions and results about: algebra, homomorphism, isomorphism, subalgebra, product algebra, the three isomorphism theorems.

### DIFF
--- a/theories/Classes/interfaces/canonical_names.v
+++ b/theories/Classes/interfaces/canonical_names.v
@@ -1,5 +1,7 @@
 Require Export HoTT.Basics.Overture HoTT.Types.Bool HoTT.Basics.Decidable HoTT.Basics.Trunc HoTT.HIT.Truncations.
 
+Require Import HoTT.Types.Sigma HoTT.Types.Forall HoTT.Types.Record.
+
 Declare Scope mc_scope.
 Delimit Scope mc_scope with mc.
 Global Open Scope mc_scope.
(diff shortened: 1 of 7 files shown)


Alizter:
This is just standard list stuff?

andreaslyn:
> This is just standard list stuff?

Yes, and the changes to canonical_names.v.

Alizter:
If we can get some more group theory and some vector spaces/modules that would be very useful. These are the main things I feel mathclasses is lacking at the moment.

For example the lean library has the Spectral library which has a lot of standard algebra. We could use a lot of that especially when Eilenberg MacLane spaces are finished, allowing us to define cohomology etc.

Also I’d imagine formalising abelian categories wouldn’t be too difficult, just a matter of writing out definitions.

theories/Classes/interfaces/canonical_names.v
```diff
@@ -263,6 +263,33 @@ Class Equivalence `(R : relation A) : Type :=
...
+  (R : relation A) {n} `{!forall (x y : A), IsTrunc n (R x y)}
+  :  IsTrunc n (SigEquivalence R).
+Proof.
+  apply @trunc_sigma.
+  - apply trunc_forall.
+  - intros. apply @trunc_sigma; intros; apply trunc_forall.
+Defined.
+
+Lemma issig_equivalence {A:Type} (R : relation A)
+  : Equivalence R <~> SigEquivalence R.
```
mikeshulman: I'm just curious, why did you define it in this order rather than the other way, given that you have to apply `symmetric_equiv` in proving it *and* take its inverse `^-1` when using it in the next instance?

theories/Classes/interfaces/canonical_names.v
```diff
@@ -263,6 +263,33 @@ Class Equivalence `(R : relation A) : Type :=
     Equivalence_Symmetric :> Symmetric R ;
     Equivalence_Transitive :> Transitive R }.
 
+Definition SigEquivalence {A:Type} (R : relation A) : Type :=
+  {_ : Reflexive R | { _ : Symmetric R | Transitive R}}.
+
+Global Instance trunc_sig_equivalence `{Funext} {A : Type}
+  (R : relation A) {n} `{!forall (x y : A), IsTrunc n (R x y)}
```
mikeshulman: Not really about this PR, but: hott-classes doesn't already require an "equivalence relation" to be prop-valued?  Is that what we want?

I also wonder about using the word `Equivalence` for equivalence relations, it seems too easily confused with `Equiv`.  Maybe we could change it to `EqRel` or something?

SkySkimmer: >Not really about this PR, but: hott-classes doesn't already require an "equivalence relation" to be prop-valued? Is that what we want?

Generally hottclasses doesn't put the truncation hypotheses together with the other structure stuff. 

EqRel sounds like it's about equality.

mikeshulman: `EquivRel` then?

Where are the truncation hypotheses put in general?

theories/Classes/interfaces/canonical_names.v
andreaslyn: It just felt more natural while writing the type. But I agree, the other way around is better.

theories/Classes/interfaces/canonical_names.v
andreaslyn: > Not really about this PR, but: hott-classes doesn't already require an "equivalence relation" to be prop-valued? Is that what we want?
> 
> I also wonder about using the word `Equivalence` for equivalence relations, it seems too easily confused with `Equiv`. Maybe we could change it to `EqRel` or something?

HoTT-classes requires equivalence relation to be prop-valued. I made it more generic because, well, it is more generic.

I have been thinking about changing the name `Equivalence` to something like `EquivRel` for the same reason. The name `Equivalence` was already part of the library, so I felt like I should not change it.

andreaslyn: > `EquivRel` then?
> 
> Where are the truncation hypotheses put in general?

I am not sure whether this is the best place to put it. I felt like it could not be too bad to put it next to `Equivalence` because it closely related.

mikeshulman: > HoTT-classes requires equivalence relation to be prop-valued.

Really?  Where?  I don't see it.

andreaslyn: It is required that a congruence is a mere relation https://github.com/andreaslyn/hott-classes/blob/debfbd8254a1a5274afb1ecda791bfde3f4211da/interfaces/ua_congruence.v#L48.

mikeshulman: Are we using the term "hott-classes" for two different things?  I thought it referred to the port of mathclasses to hott that @SkySkimmer wrote (https://github.com/SkySkimmer/HoTTClasses) and which is now mostly part of this library, but @andreaslyn you seem to be using it to refer to your own new code instead?

andreaslyn: Yes, sorry. My code is also based on math-classes. But it is not a direct port because math-classes uses setoids which we do not need in set-level universal algebra in HoTT.

andreaslyn:
The name `Equivalence` was only used in two places, so I changed it to `EquivRel`. I have also changed the type of `issig`.

mikeshulman:
Thanks!  This looks fine to me.

PR #1009, opened 2019-02-23
Universal algebra: Definition of algebra

andreaslyn:
This is the second universal algebra pull request. The first one: https://github.com/HoTT/HoTT/pull/1007

### DIFF
--- a/theories/Classes/interfaces/ua_algebra.v
+++ b/theories/Classes/interfaces/ua_algebra.v
@@ -1,0 +1,197 @@
+(** This file defines [Algebra]. *)
+
+Require Export
+  Coq.Unicode.Utf8
+  HoTT.Classes.implementations.ne_list
+  HoTT.Classes.implementations.family_prod.
+
+Require Import
+  HoTT.Basics.Overture
+  HoTT.Types.Record
+  HoTT.Types.Sigma
+  HoTT.Types.Arrow
+  HoTT.Types.Forall
+  HoTT.Types.Universe
+  HoTT.HSet
+  HoTT.Classes.interfaces.abstract_algebra
+  HoTT.Classes.implementations.list.
+
+Import ne_list.notations.
+
+Declare Scope Algebra_scope.
+
+Delimit Scope Algebra_scope with Algebra.
+
+Open Scope Algebra_scope.
+
+Definition SymbolType_internal : Type → Type := ne_list.
+
+(** A [Signature] is used to characterise [Algebra]s. In particular
+    a signature specifies which operations (functions) an algebra for
+    the signature is expected to provide. A signature consists of
+    - A type of [Sort]s. An algebra for the signature must provide
+      a type for each [s : Sort].
+    - A type of function symbols [Symbol]. For each function symbol
+      [u : Symbol], an algebra for the signature must provide a
+      corresponding operation.
+    - The field [symbol_types σ u] indicates which type the operation
+      corresponding to [u] should have. *)
+
+Record Signature : Type := BuildSignature
+  { Sort : Type
+  ; Symbol : Type
+  ; symbol_types : Symbol → SymbolType_internal Sort }.
+
+(** We have this implicit coercion allowing us to use a signature
+    [σ:Signature] as a map [Symbol σ → SymbolType σ]
+    (see [SymbolType] below). *)
+
+Global Coercion symbol_types : Signature >-> Funclass.
+
+(** A single sorted [Signature] is a signature with [Sort = Unit]. *)
+
+Definition BuildSingleSortedSignature (sym : Type) (arities : sym → nat)
+  : Signature
+  := BuildSignature Unit sym (ne_list.replicate_Sn tt o arities).
+
+(** Let [σ:Signature]. For each symbol [u : Symbol σ], [σ u]
+    associates [u] to a [SymbolType σ]. This represents the required
+    type of the algebra operation corresponding to [u]. *)
+
+Definition SymbolType (σ : Signature) : Type := ne_list (Sort σ).
+
+(** For [s : SymbolType σ], [cod_symboltype σ] is the codomain of the
+    symbol type [s]. *)
+
+Definition cod_symboltype {σ} : SymbolType σ → Sort σ
+  := ne_list.last.
+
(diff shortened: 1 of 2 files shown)


spitters:
LGTM.
@SkySkimmer what do you think?

theories/Classes/interfaces/ua_algebra.v
```diff
@@ -0,0 +1,213 @@
...
+Definition arity_symboltype {σ} : SymbolType σ → nat
+  := length o dom_symboltype.
+
+(** An [Algebra] must provide a family of [Carriers σ] indexed by
+    [Sort σ]. These carriers are the "objects" (types) of the algebra. *)
+
+(* [Carriers] is a notation because it will be used for an implicit
+   coercion [Algebra >-> Funclass] below. *)
+
+Notation Carriers := (λ σ, Sort σ → Type).
```
SkySkimmer: Why not `Notation Carriers σ := (Sort σ -> Type).`? I don't think you need to generate the beta redexes.

theories/Classes/interfaces/ua_algebra.v
```diff
@@ -0,0 +1,213 @@
...
+(** An [Algebra σ] for a signature [σ] consists of a family [carriers :
+    Carriers σ] indexed by the sorts [s : Sort σ], and for each symbol
+    [u : Symbol σ], an operation of type [Operation carriers (σ u)],
+    where [σ u : SymbolType σ] is the symbol type of [u]. For each
+    sort [s : Sort σ], [carriers s] is required to be a set. *)
+
+Record Algebra {σ : Signature} : Type := BuildAlgebra
+  { carriers : Carriers σ
+  ; operations : ∀ (u : Symbol σ), Operation carriers (σ u)
+  ; hset_carriers_algebra : ∀ (s : Sort σ), IsHSet (carriers s) }.
```
SkySkimmer: Why do we require carriers to be sets?

theories/Classes/interfaces/ua_algebra.v
andreaslyn: You are right. Good catch.

theories/Classes/interfaces/ua_algebra.v
andreaslyn: It was more convenient for my work. But it is more generic to remove the set requirement from the algebra definition, and add `IsTruncAlgebra` and `IsHSetAlgebra` typeclasses. I am working on it.

andreaslyn:
@SkySkimmer @spitters what do think about it now?

theories/Classes/interfaces/ua_algebra.v
```diff
@@ -0,0 +1,213 @@
...
+Lemma issig_algebra (σ : Signature) : SigAlgebra σ <~> Algebra σ.
+Proof.
+  unfold SigAlgebra.
+  issig (@BuildAlgebra σ) (@carriers σ) (@operations σ).
+Defined.
+
+(** To find a path between two algebras [A B : Algebra σ] it suffices
+    to find a path between the carriers and the operations. *)
+
+Lemma path_algebra `{Funext} {σ : Signature} (A B : Algebra σ)
```
SkySkimmer: You shouldn't need funext for this now that the record doesn't contain `IsHSet`.
Also the proof may be cleaner done in direct style (destruct A, B then the equality proofs) rather than through the issig lemma.

theories/Classes/interfaces/ua_algebra.v
spitters: I'm not yet convinced that that would be right generalization.
Higher universal algebra would be something like:
https://www.irif.fr/_media/rencontres/pps2018/finster.pdf
But that's not what this formalization is about.

Of course, it might simplify the formalization of one can freely remove the set restriction.

mikeshulman: Yeah, simply removing the truncation axiom from (say) a group doesn't give you the *right* notion of "higher group".  But it does give you a notion of (say) "wild group" that might nevertheless be useful for some purposes, like [wild categories](https://ncatlab.org/homotopytypetheory/show/wild+category) are.

andreaslyn: I have just implemented set-level universal algebra, but it makes sense to have the right definition of algebra to allow for extending with higher universal algebra.

I have looked at https://www.irif.fr/_media/rencontres/pps2018/finster.pdf, but I am still not sure what the right definition is. Does it require a lot of changes to the current definition?

spitters: Yes, I think it will require a lot of work.
However, as Mike suggest wild structures may be interesting, so if there's an easy way to remove the set level restriction it may be worth it. Otherwise, leave it for future work.

spitters: @ericfinster may have a quick suggestion about the usefulness of "wild" universal algebra.

andreaslyn: Removing the set level restriction is mostly straight forward. Then I can add `IsHSetAlgebra` arguments where the set level restriction is assumed.

theories/Classes/interfaces/ua_algebra.v
andreaslyn: Thanks. The proof is much cleaner now.

theories/Classes/interfaces/ua_algebra.v
ericfinster: Yes, I agree that "wild" structures can be very interesting.  Wild monoids (what topologists call H-spaces) have played a historically important role in topology and there are many beautiful theorems about them.  To some extent, it is even more surprising and interesting when we find objects which admit a structure of some finite level of coherence but cannot be completely to fully coherent one.  And there are examples of this in the literature.  The important thing is rather to understand the difference and not confuse the two things.

mikeshulman: Note that a "wild group" is also known as a grouplike associative H-space.  So there are certainly uses for some wild algebraic structures at least!  Also in the special case of theories without axioms, a wild structure is trivially the same as a coherent one.

Fully coherent higher universal algebra would almost certainly require a totally different approach along the lines of Eric's work.

PR #1010, opened 2019-02-27
Universal algebra: Homomorphism and isomorphism.

andreaslyn:
This is the third universal algebra pull request. The two previous pull requests are:

- https://github.com/HoTT/HoTT/pull/1007
- https://github.com/HoTT/HoTT/pull/1009


andreaslyn:
The proof of `path_operations_equiv` is using `rewrite` and `transport_forall_hammer`, hence is left opaque. This makes for a short proof, but now that paths in algebra are not necessarily propositions, maybe it is better to avoid `transport_forall_hammer` and `rewrite` and make the proof transparent. What do you think?

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,391 @@
...
+    a homomorphism if for each function symbol [u : Symbol σ], it is
+    [OpPreserving (u^^A) (u^^B)] with respect to the algebra
+    operations [u^^A] and [u^^B] corresponding to [u]. *)
+
+  Definition IsHomomorphism : Type
+    := ∀ (u : Symbol σ), OpPreserving (u^^A) (u^^B).
+
+  Global Instance trunc_is_homomorphism `{Funext} {n : trunc_index}
+    `{!IsTruncAlgebra n.+1 B}
+    : IsTrunc n IsHomomorphism.
```
SkySkimmer: Do we usually make all the arguments implicit for this sort of lemma?

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,391 @@
...
+
+Lemma path_hset_homomorphism `{Funext} {σ} {A B : Algebra σ} `{!IsHSetAlgebra B}
+  (f g : Homomorphism A B) (p : def_hom f = def_hom g)
+  : f = g.
+Proof.
+  apply (path_homomorphism _ _ p). apply path_ishprop.
+Defined.
+
+(** [f : Homomorphism A B] is an isomorphism if for each [s : Sort σ],
+    [f s] is both an embedding and a surjection. *)
```
SkySkimmer: Why is an isomorphism not just a homomorphism which is also an equivalence?

theories/Classes/theory/ua_homomorphism.v
andreaslyn: There is an explicit `f : ∀ s, A s → B s` argument in the top of the section. Is there another argument that ought to be explicit here?

SkySkimmer: Was reading too fast

theories/Classes/theory/ua_homomorphism.v
andreaslyn: Before an isomorphism was a homomorphism which is both an injection and a surjection for each sort. Now that algebra carriers are not always sets I changed it to embedding and surjection for each sort, but it is clearly better to define it as a homomorphism which is an equivalence for each sort.

mikeshulman: I would tend to argue that the latter is a better definition even for sets... (-:

andreaslyn: I have changed it now. It simplifies some proofs a lot.

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,334 @@
...
+  Qed.
+End hom_compose.
+
+(** If [f : ∀ i, F i <~> G i] is a family of equivalences,
+    then by function extensionality composed with univalence there is
+    a path [F = G]. *)
+
+Definition path_equiv_family `{Univalence} {I : Type}
+  {F G : I → Type} (f : ∀ i, F i <~> G i)
+  : F = G
```
SkySkimmer: Not sure that this is worth its own definition.

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,334 @@
...
+  Defined.
+End is_homomorphism.
+
+(** Given algebras [A B : Algebra σ] a homomorphism [Homomorphism A B]
+    is a family of functions [f : ∀ (s : Sort σ), A s → B s] where
+    [IsHomomorphism f] holds. *)
+
+Record Homomorphism {σ} {A B : Algebra σ} : Type := BuildHomomorphism
+  { def_hom : ∀ (s : Sort σ), A s → B s
+  ; is_homomorphism_hom : IsHomomorphism def_hom }.
```
SkySkimmer: Why go with the bundled approach here vs staying with raw function arguments + ishomo instances?

andreaslyn: I feel like the bundled approach gives clearer interfaces. Is there an advantage to the alternative?

theories/Classes/theory/ua_homomorphism.v
andreaslyn: It is also referred to in the documentation. Would it be OK to make it a local notation instead?

theories/Classes/theory/ua_homomorphism.v
andreaslyn: What is the experience with `IsEquiv` and `Equiv`? Is it good to have both, or would one of them have been enough/better?

mikeshulman: It is good to have both.

andreaslyn: I am considering making `IsHomomorphism` a typeclass. `IsHomomorphism` is not always a proposition, so are there any problems related to this?

Does it make sense to keep the bundled `Homomorphism`, and implement two versions of some functions, one version with `Homomorphism` argument and another version with raw function + `IsHomomorphism` arguments?

mikeshulman: I wouldn't expect any significant problems; certainly in the case of most interest when the carriers are sets, it is a proposition.

We do have analogous copies of some functions that use `Equiv` and `IsEquiv`.  I don't think it's ideal, but it is sometimes convenient to have both.

spitters:
LGTM. @SkySkimmer what do you think?

andreaslyn:
I have changed `IsHomomorphism` to a type class and I have changes some `f : Homomorphism` arguments to `f : ∀ s, A s → B s` + `IsHomomorphism f`.

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,337 @@
...
+  : f = g.
+Proof.
+  apply (path_homomorphism _ _ p). apply path_ishprop.
+Defined.
+
+(** A homomorphism [f : ∀ s, A s → B s] is an isomorphism if for each
+    [s : Sort σ], [f s] is an equivalence. *)
+
+Class IsIsomorphism {σ} {A B : Algebra σ} (f : ∀ s, A s → B s) `{!IsHomomorphism f}
+  := isequiv_carriers_isomorphism : ∀ (s : Sort σ), IsEquiv (f s).
```
SkySkimmer: I would instead write this as `Class IsIso f := { iso_homo :> IsHomo f; iso_equiv :> forall s, IsEquiv (f s) }`

andreaslyn:
I have defined `IsIsomorphism` as suggested by @SkySkimmer. I have also removed the bundled `Homomorphism`.

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,309 @@
...
+       | [:s:] => λ α β, f s α = β
+       | s ::: y => λ α β, ∀ (x : A s), OpPreserving (α x) (β (f s x))
+       end.
+
+  Global Instance trunc_oppreserving `{Funext} {n : trunc_index}
+    `{!IsTruncAlgebra n.+1 B} {w : SymbolType σ}
+    (α : Operation A w) (β : Operation B w)
+    : IsTrunc n (OpPreserving α β).
+  Proof.
+    intros. induction w; exact _.
```
Alizter: Unneeded intros.

theories/Classes/theory/ua_homomorphism.v
```diff
@@ -0,0 +1,309 @@
...
+    operations [u^^A] and [u^^B] corresponding to [u]. *)
+
+  Class IsHomomorphism : Type
+    := oppreserving_hom : ∀ (u : Symbol σ), OpPreserving (u^^A) (u^^B).
+
+  Global Instance trunc_is_homomorphism `{Funext} {n : trunc_index}
+    `{!IsTruncAlgebra n.+1 B}
+    : IsTrunc n IsHomomorphism.
+  Proof.
+    intros. apply trunc_forall.
```
Alizter: Unneeded intros.

andreaslyn: Thanks.

andreaslyn:
@SkySkimmer what do you think?

PR #1013, opened 2019-03-08
Universal algebra: product algebra

andreaslyn:


theories/Classes/theory/ua_prod_algebra.v
```diff
@@ -0,0 +1,222 @@
...
+     such that [h i s = pr i s ∘ f s], where
+
+    <<
+      pr i = hom_projection_prod_algebra I A i
+    >>
+
+    is the ith projection homomorphism. *)
+
+  Lemma ump_prod_algebra `{!∀ i, IsHSetAlgebra (A i)}
+   : (∀ (i:I), {g : ∀ s, C s → A i s | IsHomomorphism g})
```
andreaslyn: This is a place where I prefer to have bundled homomorphism. Is there another way to fix this?

theories/Classes/theory/ua_prod_algebra.v
```diff
@@ -0,0 +1,222 @@
...
+    <~>
+    {f : ∀ s, C s → ProdAlgebra I A s | IsHomomorphism f}.
```
andreaslyn: This is a place where I prefer to have bundled homomorphism. Is there another way to fix this type?

mikeshulman: I think there are times when you want things to be bundled.  The commonest case is, as in this example, when you're comparing the homotopy types of the bundles.  The unbundled information is already carried by `is_homomorphism_ump_prod_algebra_mapin` and `is_homomorphism_prod_algebra_mapout`.

andreaslyn: Ok thanks.

The current `IsIsomorphism` definition https://github.com/HoTT/HoTT/blob/365cbed0291d3efdf41014c4e2eaf63dec8b3820/theories/Classes/theory/ua_homomorphism.v#L68 makes the bundled homomorphism less attractive because having both `f : Homomorphism A B` and `IsIsomorphism f` in context gives two proofs of `IsHomomorphism f` in context, which seems confusing. One can define a bundled `Isomorphism` to prevent it, but this feels like circumventing the issue rather than fixing it.

Since `IsIsomorphism` is supposed to extend `IsHomomorphism`, I would expect the definition of `IsIsomorphism` to be
```
Class IsIsomorphism {σ} {A B : Algebra σ} (f : ∀ s, A s → B s) `{!IsHomomorphism f}
  := isequiv_isomorphism : ∀ (s : Sort σ), IsEquiv (f s).
```
This fixes the issue. Is there something wrong with this approach?

mikeshulman: This is getting into deeper typeclass waters than I can swim in.

theories/Classes/theory/ua_prod_algebra.v
```diff
@@ -0,0 +1,197 @@
...
+      such that [h i = hom_compose (pr i) f], where
+
+      <<
+        pr i = hom_proj_prod_algebra I A i
+      >>
+
+      is the ith projection homomorphism. *)
+
+ Lemma ump_prod_algebra `{!∀ i, IsHSetAlgebra (A i)}
+   : (∀ (i:I), Homomorphism C (A i)) <~> Homomorphism C (ProdAlgebra I A).
```
andreaslyn: I have reintroduced the bundled homomorphism, which improves on this.

spitters:
LGTM.

andreaslyn:
Does someone want to review this?

spitters:
@mikeshulman @SkySkimmer  ?

theories/Classes/theory/ua_prod_algebra.v
SkySkimmer: >Since IsIsomorphism is supposed to extend IsHomomorphism, I would expect the definition of IsIsomorphism to be

For me extensions generally mean it's a field of the new class, ie
~~~coq
Class IsIsomorphism ... f := {
  ishomomorphism_isomorphism :> IsHomomorphism f;
  isequiv_isomorphism :> ∀ (s : Sort σ), IsEquiv (f s);
}.
~~~
I don't know if it makes things easier or harder for your use case.

andreaslyn: Thanks.

The current way (with `IsHomomorphism` as argument) does correspond to haskell type class inheritance, right?
```
class (IsHomomorphism f) => IsIsomorphism f where
  isequiv_isomorphism : ...
```
Having the base class `IsHomomorphism` as a field reduces the number of arguments to functions with `IsIsomorphism` arguments, but has complications, as explained above. Are there other benefits of having a base class as field?

andreaslyn: In Haskell `f` is required to be type valued, `f : ... -> *`, so maybe the analogy with Haskell type classes breaks a bit there.

PR #1015, opened 2019-03-16
Minor coq fixes

mattam82:
These patches add declarations to help Coq tactics know about the identity type used in HoTT and allow calling "induction" on booleans (useful for generic tactics), along with a minor fix for "autoapply"

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -221,8 +221,12 @@ Arguments paths_ind [A] a P f y p.
 Scheme paths_rec := Minimality for paths Sort Type.
 Arguments paths_rec [A] a P f y p.
 
+Register idpath as core.identity.refl.
+
 (* See comment above about the tactic [induction]. *)
 Definition paths_rect := paths_ind.
+
+Register paths_rect as core.identity.ind.
 
 Notation "x = y :> A" := (@paths A x y) : type_scope.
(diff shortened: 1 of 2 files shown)


mattam82:
The autoapply fix only works in master.

Alizter:
Why isn't travis happy?

mattam82:
Because it tries with a version of Coq I cannot determine which doesn't support the fixed syntax, I believe.

mattam82:
Ok,thanks!

mattam82:
Looks ready now

theories/Basics/Overture.v
```diff
@@ -257,12 +263,18 @@ Local Open Scope path_scope.
...
 
+Register inverse as core.identity.sym.
+
 (** Declaring this as [simpl nomatch] prevents the tactic [simpl] from expanding it out into [match] statements.  We only want [inverse] to simplify when applied to an identity path. *)
 Arguments inverse {A x y} p : simpl nomatch.
 
 Global Instance symmetric_paths {A} : Symmetric (@paths A) | 0 := @inverse A.
 Arguments symmetric_paths / .
 
+(** This allows [rewrite] to both in left-to-right and right-to left directions. *)
```
Alizter: Will this mean the rewrite tactic will be using path_ind?

Issue #963, opened 2018-10-11
Drop HoTT's copy of the Coq prelude.


ejgallego:
Coq's https://github.com/coq/coq/pull/186 has landed and while not perfect yet it should mean HoTT should be able to stop vendoring its own modified version of Coq's prelude.

Once this is done, this shall close #477

ejgallego:
@Alizter TTBOMK this is still not solved, tho we are not far from actually doing it, however I think some bugs in Coq will appear when the full conversion is done.

Alizter:
@ejgallego Thank you for letting me know, I have reopened the issue.

Alizter:
@ejgallego Sorry I missed your gitter reply. Is this the issue you were talking about. I would be happy to lend a hand with finishing this.

spitters:
https://github.com/HoTT/HoTT/issues/1424

PR #1016, opened 2019-03-16
Cumulative paths

mattam82:
Make the paths type cumulative, so that it can be used with Equations

Making paths cumulative means that the universe parameter is now treated as irrelevant:`@paths@{i} A x y <= @paths@{j} A x y` for any i and j (which must by typing be greater or equal to A's universe). See the [section on cumulative inductives](https://coq.inria.fr/refman/addendum/universe-polymorphism.html?highlight=irrelevant#cumulative-noncumulative) for more information.

This allows, e.g. to inject `@paths@{Set} nat x y` equalities in higher universes, and deconstruct equalities in the homogeneous cumulative sigma type used by Equations (where both components live in the same universe, and the sigma as well) into equalities for the components, e.g: `@paths{i} (Sigma@{i} x : nat, P x) p q` can be turned into `Sigma@{i} e : @paths@{i} nat p.1 q.1, @paths@{i} (transport P e p.2) q.2` without trouble. 

Trying out the same sigma-types for HoTT would be a good exercise: instead of "preemptively" making sigma types allow cumulativity by using different universes `i` and `j` for the components and putting the sigma in `max(i,j)`, it rather says that every universe is closed under sigmas, and cumulativity is orthogonal (the universe annotation of the sigma is irrelevant as well). This can result in less useless universes representing `max(i,j)` and still allow using cumulativity when necessary.

This requires two restrictiction of the more flexible constraints due to cumulativity of paths in Modalities and PropResizing, otherwise nothing changes.

### DIFF
--- a/theories/TruncType.v
+++ b/theories/TruncType.v
@@ -94,11 +94,11 @@ Qed.
 Definition path_hset {A B} := @path_trunctype 0 A B.
 Definition path_hprop {A B} := @path_trunctype -1 A B.
 
-Global Instance istrunc_trunctype {n : trunc_index}
-  : IsTrunc n.+1 (TruncType n) | 0.
+Global Instance istrunc_trunctype@{i j k | i < j, j < k} {n : trunc_index}
+  : IsTrunc@{j} n.+1 (TruncType@{i} n) | 0.
 Proof.
   intros A B.
-  refine (trunc_equiv _ (equiv_path_trunctype A B)).
+  refine (trunc_equiv _ (equiv_path_trunctype@{i j k} A B)).
   case n as [ | n'].
   - apply contr_equiv_contr_contr. (* The reason is different in this case. *)
   - apply istrunc_equiv.
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -214,7 +214,7 @@ Notation "g 'oD' f" := (composeD g f) : function_scope.
 (** The results in this file are used everywhere else, so we need to be extra careful about how we define and prove things.  We prefer hand-written terms, or at least tactics that allow us to retain clear control over the proof-term produced. *)
 
 (** We define our own identity type, rather than using the one in the Coq standard library, so as to have more control over transitivity, symmetry and inverse.  It seems impossible to change these for the standard eq/identity type (or its Type-valued version) because it breaks various other standard things.  Merely changing notations also doesn't seem to quite work. *)
-Inductive paths {A : Type} (a : A) : A -> Type :=
+Cumulative Inductive paths {A : Type} (a : A) : A -> Type :=
   idpath : paths a a.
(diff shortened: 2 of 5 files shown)


theories/TruncType.v
```diff
@@ -94,8 +94,8 @@ Qed.
 Definition path_hset {A B} := @path_trunctype 0 A B.
 Definition path_hprop {A B} := @path_trunctype -1 A B.
 
-Global Instance istrunc_trunctype {n : trunc_index}
-  : IsTrunc n.+1 (TruncType n) | 0.
+Global Instance istrunc_trunctype@{i j k | i < j, j < k} {n : trunc_index}
+  : IsTrunc@{j} n.+1 (TruncType@{i} n) | 0.
```
SkySkimmer: What's `k` for?

mattam82:
@amintimany this might interest you!

theories/TruncType.v
mattam82: The proof requires a larger universe

mattam82: We share l.u.b's of flexible universes during minimization, that's why it magically gets unified with k. This is done precisely to avoid generating many universes representing the same max(i,j).

Alizter:
Can you elaborate some more on using this with Equations. I wasn't aware of that project, and I had a quick look and it looks really cool! The only problem I see with Equations is that everybody will need to plugin for a proof written with it to work no?

mattam82:
Well, of course if you use a tool you need the tool. In principle a .vo produced using Equations definitions could actually be independent of the tool (it produces legit pure Coq definition) but that’s not the case today.

spitters:
Cool. LGTM. @andreaslyn and @ejgallego will also be interested.
What's currently keeping Equations for HoTT?
#963 and #895

mattam82:
My HoTT port doesn’t need anything else than this and a fix in Coq that should be there by 8.10. I don’t think it’s a good idea to port Program which relies on UIP intrinsically to handle pattern-matching. My port relies heavily on @ejgallego ‘s Coqlib change but does not need anything from Coq’s stdlib (apart from nat and non-dependent products, but that’s annecdotical and could be changed easily). Orthogonally, we could try improving support for tactics when bound to HoTT constants in Coq, like congruence, rewrite and the like. I think porting ssreflect tactics should be a breeze now, as well.

spitters:
@SkySkimmer do you have any more questions, if not feel free to pull.

SkySkimmer:
>@SkySkimmer do you have any more questions, if not feel free to pull.

I don't have pull powers on this repo though?

mattam82:
Thanks! I'll release an equations-hott package ASAP!

spitters:
@mattam82 Did you get around to working on this? I'd be curious to try.

PR #1020, opened 2019-03-31
Update INSTALL.md

semorrison:
`coq.dev` is gone from opam, so I suggest removing the first section of instructions.

```
$ opam install coq.dev
[ERROR] Package coq has no version dev.
```


mseri:
If you want the dev version you can do `opam install coq --dev`

mikeshulman:
This was fixed by #1025 .

PR #1023, opened 2019-04-02
Rename PreCategory/Category to Category/UnivalentCategory

Alizter:
I have wanted to do this for a while. There may be a lot of comments that I missed.


spitters:
You may want to put .crashcoqide in your .gitignore.

Is there concensus about this new naming?
Should the book/nlab be changed too?

Alizter:
@spitters I may have staged those files accidentally.

There was an email discussion on the HoTT group a while back but I don't remember any consensus being reached. [This preprint](https://arxiv.org/abs/1903.01152) makes me think that this way of naming is more correct.

If what is traditionally done as category theory can be done for precategories, surely these ought to be our categories?

A thing that the bicategories paper brings up is that there can be "different levels" of univalence.

And for the HoTT book I think it would be better to call them categories/univalent categories. From the get go it is more obvious which has what property.

spitters:
From those authors, only Benedikt was involved in writing the book. I don't want to randomly change terminology to conflict with what we used in the book. It should be part of a broader discussion, and perhaps a simultaneous change in the book. @mikeshulman @andrejbauer ?
Also it will make @JasonGross work more difficult to read.

Alizter:
Here is the [mailing list discussion](https://groups.google.com/forum/#!topic/homotopytypetheory/ZXfRp05oZeU).

spitters:
Correct me if I'm wrong, but Ulrik and Mike seem to disagree with the change in terminology.

Alizter:
@spitters I know that Mike doesn’t agree but I have made this pull request anyway. There seems to be a bit of a split in the whole community and I would like to understand why this is. This pull request can sit in limbo till it’s resolved.

spitters:
Thanks. It would saved me a bit of time if you had stated that upfront.

Thanks for your contributions in any case!

Alizter:
Sorry for not being clearer!

mikeshulman:
It is not true that "what is traditionally done as category theory can be done for precategories".  The most obvious example is that in traditional category theory, a fully faithful and essentially surjective functor is an equivalence.  And indeed, there can be different levels of "pre-ness", but the correct notion of "bicategory" (or any categorical structure) is the fully univalent one.  As suggested on the mailing list, to talk about different levels of pre-ness we can use prefixes on the adjective "flagged".

Note that this is the *universally accepted* terminology in traditional set-based higher category theory: there is *no one* who refers to a [Segal space](https://ncatlab.org/nlab/show/Segal+space) as an "(oo,1)-category" and a [complete Segal space](https://ncatlab.org/nlab/show/complete+Segal+space) as a "univalent (oo,1)-category".  And there are extremely good reasons for that, despite the fact that *some* of higher category theory can be developed for Segal spaces (or, more model-independently, flagged (oo,1)-categories) as well.

If you want to change "precategory" to "flagged category" I would support that.  But I do not support making pull requests until *after* a consensus has been reached.

Issue #1021, opened 2019-03-31
install instructions on macOS: "cannot find ocamlfind in your path"


semorrison:
Following the instructions at https://github.com/HoTT/HoTT/blob/master/INSTALL.md on macOS, when we get to the `etc/install_coq.sh` step in the HoTT directory, it fails with the following output:

```
$ git submodule sync
Synchronizing submodule url for 'coq-HoTT'
Synchronizing submodule url for 'etc/coq-dpdgraph'
Synchronizing submodule url for 'etc/coq-scripts'
$ git submodule update --init --recursive
~/scratch/HoTT/coq-HoTT ~/scratch/HoTT ~/scratch/HoTT/etc ~/scratch/HoTT
$ ./configure -local 
Error: cannot find 'ocamlfind' in your path!
```
Please adjust your path or use the -ocamlfind option of ./configure
Configuration script failed!

kalfsvag:
Just wanted to say that I have the exact same issue, trying to install on Linux Mint.

mikeshulman:
Did ocamlfind get installed somewhere on your system and it's just missing from the path?  Or did it not get installed at all?

mikeshulman:
Actually I can't square the commands you listed with the instructions at `INSTALL.md`.  I don't see anywhere in that file instructions to run `git submodule` or `configure -local`.

kalfsvag:
Everything from install_coq_deps.sh is installed.

Tried installing ocaml-findlib, and now the error instead is

kalfsvag:
Everything from install_coq_deps.sh is installed.

Tried installing ocaml-findlib, and now the error instead is 

> Your version of OCaml is 4.02.3.
> You need OCaml 4.05.0 or later.

When trying to update ocaml using apt-get, I only get the assertion that I have the latest version. . .

mikeshulman:
Try installing ocaml with opam instead: #929.

mseri:
@kalfsvag make sure to use opam2: https://opam.ocaml.org/doc/Install.html

kalfsvag:
I have to admit I am terribly confused right now.

I have updated to opam2 and used it to install ocaml. 
Running "ocamlfind -v" says that "The program 'ocamlfind' is currently not installed."

Running "opam install ocamlfind" gives me
"Package ocamlfind is already installed (current version is 1.8.0)."

Does this mean that "it's just missing from the path"? What does that mean, and how can it be fixed?

[...]

PR #1025, opened 2019-04-11
update INSTALL.md, close #929 and #1021

### LINKED ISSUE #929: Update INSTALL.md?
Currently `INSTALL.md` instructs Linux users to use their package manager to install ocaml, camlp5, etc.  Is this still the preferred approach?  I had issues with the recent bump of Coq until I blew away all my apt packages and installed everything with opam instead.

mikeshulman:
Since no one who actually uses OSX seems interested in updating the INSTALL file, I attempted to do it myself based on others comments.  Maybe someone who uses OSX can at least test the instructions as I've written them and see whether they work?

I wasn't sure where camlp5 should be installed; does it also come from opam or from brew?  And for the Linux instructions, should we alter `etc/install_coq_deps.sh` to not deal with ocaml/camlp5 and leave them to the user to do with opam, or should we make it call opam itself?  In fact, could we also create a version of `install_coq_deps.sh` for OSX?

I also got fed up with trying to maintain two separate "simplified" and "detailed" sets of instructions, which were highly redundant and at times conflicting, and merged them into one set of instructions.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,7 +3,7 @@
 Make sure you have all the prerequisites for compiling Coq. These are
 `OCaml`, `camlp5`, `git`, and `make`. If you want to build the HoTT
 version of the graphical user interface `coqide` you also need the
-`lablgtk3` and `lablgtksourceview2` libraries.  To get these:
+`lablgtk3` and `lablgtksourceview3` libraries.  To get these:
 
 ### Linux and OSX
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,214 +1,253 @@
-# INSTALLATION via OPAM
+## Obtain Prerequisites
 
-After installing Coq via opam [1]. One can install our version of Coq by:
-```
-   opam install coq.dev
-```
+Make sure you have all the prerequisites for compiling Coq. These are
+`OCaml`, `camlp5`, `git`, and `make`. If you want to build the HoTT
+version of the graphical user interface `coqide` you also need the
+`lablgtk3` and `lablgtksourceview2` libraries.  To get these:
 
-However, one still needs to install the library via git to contribute.
-Opam support on windows is experimental.
-[1]: https://coq.inria.fr/opam/www/using.html
+### Linux and OSX
 
-We are compatible with [Coq 8.10](https://github.com/coq/coq/releases/tag/V8.10.0),
-so binary packages can be used.  Paths still need to be set manually.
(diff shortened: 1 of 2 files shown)


Alizter:
We could ask travis to do a build for OSX, if we set it up like you have specified in the INSTALL file. That way we won't be at odds when needing to test for OSX as well. The downside is of course Travis may take longer...

McTano:
I just posted some notes in issue #1020 gathered from helping a half-dozen people install HoTT-Coq on Linux at the HoTT Summer School. Hopefully they're useful.

To respond to a question you raised in this thread, since opam is necessary in both the Mac OS and Linux installations, I suggest instructing users to install with opam wherever possible.

When I have time (hopefully this week or next), I'll have a better look at this PR and try out the installation on MacOS.

mikeshulman:
Thanks!  (I think you mean #929.)

Alizter:
I am in favour of merging this however I have a few suggestions:
 - If everybody else agrees, we should replace the coq dependencies with the correct ones. Notably gtk3 instead of gtk2. See #1115 
 - Add some information about updating the coq submodule. I couldn't work out how to do this myself when we moved to 8.10 so I ended up having to get rid of my local copy and start from scratch. My new install now uses a seperate coq clone where I can be sure about updating it.

mikeshulman:
Re: updating coq, what generally works for me when the coq pin has changed is

```
git pull
git submodule update
opam update
opam upgrade
etc/install_coq.sh
make
```

mikeshulman:
Rebased, reorganized, and updated to use gtk3 and with instructions for recompiling Coq.

Alizter:
LGTM, do you want APOE?

INSTALL.md
```diff
@@ -1,220 +1,176 @@
...
+## Obtain Prerequisites
 
-After installing Coq via opam [1]. One can install our version of Coq by:
-```
-   opam install coq.dev
-```
+Make sure you have all the prerequisites for compiling Coq. These are
+`OCaml`, `camlp5`, `git`, and `make`. If you want to build the HoTT
+version of the graphical user interface `coqide` you also need the
+`lablgtk3` and `lablgtksourceview2` libraries.  To get these:
```
McTano: This line still refers to the old `lablgtksourceview2` version.

mikeshulman: Fixed, thanks.

PR #1026, opened 2019-04-20
New lemmas and some cleanup.

andreaslyn:
This will be used to develop a univalent category of algebras and homomorphisms for a signature.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -399,6 +399,7 @@ theories/Classes/interfaces/canonical_names.v
 theories/Classes/interfaces/orders.v
 theories/Classes/interfaces/integers.v
 theories/Classes/interfaces/ua_algebra.v
+theories/Classes/interfaces/ua_setalgebra.v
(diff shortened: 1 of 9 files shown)


spitters:
LGTM.
I think we want to avoid simpl and use cbn instead.

mikeshulman:
LGTM too.  Is there a reason to use cbn in cases where simpl works?  I thought it was just that cbn usually works better.

Alizter:
@mikeshulman I think simpl does a few more things than cbn, and I guess this means cbn is more consistent. That's the impression I got anyway.

spitters:
"The cbn tactic is claimed to be a more principled, faster and more predictable replacement for simpl."
https://coq.inria.fr/refman/proof-engine/tactics.html

I thought simpl would be deprecated, but it appears that it's not as strong as that.

andreaslyn:
I am now using `cbn` instead of `simpl`. In my experience `cbn` can be significantly faster when there is a lot to reduce.

PR #1028, opened 2019-04-29
The first isomorphism theorem

andreaslyn:


theories/HSet.v
```diff
@@ -120,8 +120,7 @@ Proof.
   exact (ap pr1 q).
 Defined.
 
-Lemma isembedding_isinj_hset {A B : Type} `{IsHSet A} `{IsHSet B}
```
andreaslyn: Unused assumption `{IsHSet A}`.

Alizter:
@JasonGross For some reason dep graphs is causing travis to fail here.

JasonGross:
This is probably because we are building the depgraphs job with master and the current version of dpdgraphs is incompatible with master.

If in `.travis.yml` you change
```
   - UPDATE_DEP_GRAPHS="yes" WITH_AUTORECONF="yes"
```
to
```
   - UPDATE_DEP_GRAPHS="yes" WITH_AUTORECONF="yes" BUILD_COQ="yes"
```
does that fix things?

Alizter:
@andreaslyn I've sent a PR to your branch updating `.travis.yml` like Jason said.

PR #1030, opened 2019-05-07
Local Unset Elimination Schemes

andreaslyn:
I have spotted some places where inductive types are defined with the default elimination schemes to `Prop` and `Set`. We are not using `Prop` and `Set` in HoTT, so I have changed this, according to the style guideline https://github.com/HoTT/HoTT/blob/master/STYLE.md#induction-and-recursion-principles.


SkySkimmer:
We do use Set, it's just called Type0.

PR #1032, opened 2019-05-20
Declare argument scopes explicitly for `AdjunctionUnitCounit`

maximedenes:
HoTT was relying on a Coq bug which made some bound scopes visible even
if they were not imported in some scenarios. We avoid this by declaring explicitly some argument scopes.

This is in preparation for https://github.com/coq/coq/pull/10181, but
should be backward compatible (remains to be tested).


spitters:
What do you mean "remains to be tested"?
@andreaslyn mentioned that an earlier patch broke compatibility with 8.9, so maybe we should be a bit more careful.

maximedenes:
> What do you mean "remains to be tested"?
> @andreaslyn mentioned that an earlier patch broke compatibility with 8.9, so maybe we should be a bit more careful.

I meant to let CI check it, but it seems to have succeeded :)

andreaslyn:
LGTM

Issue #1031, opened 2019-05-16
Finite induction tactic


Alizter:
If we have a type `X` that is `Finite X`, I would like to have a tactic that when applied to a goal `X -> Foo`, and it generates n `Foo` goals where n is the `fcard` of the type.

This would be very useful, especially when working with `Fin n` since it is quite difficult to keep inducting on smaller ones.

Maybe calling this tactic `FiniteInduction` would be a start?

I don't know how to write Ltac tactics very well but I attempted the following. Note I have specialised the problem for constructing `Fin n -> Y`, since I can write a tactic that will generalise this to `X -> Y` for some `Finite X`.

```
Section Foo.

  Context
    {X Y : Type}
   `{Finite X}
    {n : nat}.

  Ltac FinInd :=
    match goal with
      | [ |- Fin ?a -> ?X] =>
          induction a;
          [ apply Empty_ind
          | revert a;
            srefine (nat_ind _ _ _); cbn;
            [ apply Empty_ind
            | intros ? ?h;
              apply sum_rect;
              [ refine (h o _); exact idmap
              | apply Unit_ind; refine _]]]
    end.


  Goal Fin 3 -> Y.
    FinInd.
```

Now this doesn't work, but in my example at the bottom the intended behavior is to generate 3 goals of `Y`. I think the problem arises because I am trying to use `nat` in ltac langauge. Surely I could perhaps get away with repeating until it doesn't progress or something.

How would I go about writing such a tactic?

Alizter:
OK so shortly posting this it came to me:

```
  Ltac FinInd :=
    repeat apply sum_rect;
    match goal with
      | [ |- Empty -> ?A] => apply Empty_ind
      | [ |- Unit -> ?A] => apply Unit_ind
    end.
```

It was much simpler than I had thought.

Alizter:
I am going to open this again because it doesn't work for everything:

```
Require Import Spaces.Finite.
Require Import Types.Empty.

  Ltac FinInd :=
    repeat apply sum_rect;
    match goal with
      | [ |- Empty -> ?A] => apply Empty_ind
      | [ |- Unit -> ?A] => apply Unit_ind
    end.

Section Foo.
Context {X : Type} {x : X}.

Goal Fin 3 -> X.
  FinInd; exact x.
Defined.

Goal Fin 100 -> X.
  FinInd; exact x.
Defined.

Goal Fin 3 -> Fin 3 -> X.
  FinInd. (* Breaks here :-( *)

[...]

PR #1033, opened 2019-05-21
added finite induction tactic

Alizter:
as discussed in #1031. This makes working with `Fin n` for small values of `n` a bit easier.

### DIFF
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -1035,18 +1035,13 @@ Section Enumeration.
 
 End Enumeration.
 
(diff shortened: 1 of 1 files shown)


theories/Spaces/Finite.v
```diff
@@ -1034,3 +1034,19 @@ Section Enumeration.
   Defined.
 
 End Enumeration.
+
+Section Tactics.
```
mikeshulman: The `Section` doesn't serve any purpose, does it?  If not, I would remove it.

PR #1039, opened 2019-05-31
Universal algebra: the second isomorphism theorem

andreaslyn:
There is also a specialization of the first isomorphism theorem and some cleanup.

### DIFF
--- a/theories/Classes/theory/ua_first_isomorphism.v
+++ b/theories/Classes/theory/ua_first_isomorphism.v
@@ -91,11 +91,15 @@ Section in_image_hom.
       by destruct p.
   Qed.
 
-  Global Instance is_closed_under_ops_in_image_hom
+  Lemma is_closed_under_ops_in_image_hom
     : IsClosedUnderOps B in_image_hom.
   Proof.
     intro u. eapply closed_under_op_in_image_hom, hom.
   Qed.
+
+  Global Instance is_subalgebra_predicate_in_image_hom
+    : IsSubalgebraPredicate B in_image_hom
+    := BuildIsSubalgebraPredicate is_closed_under_ops_in_image_hom.
 End in_image_hom.
 
 (** The folowing section proves the first isomorphism theorem,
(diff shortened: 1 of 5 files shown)


theories/Classes/theory/ua_first_isomorphism.v
```diff
@@ -234,3 +238,40 @@ Section first_isomorphism_surjection.
...
+
+  Definition hom_first_isomorphism_inj
+    : Homomorphism A (B & in_image_hom f)
+    := hom_compose
+          (hom_first_isomorphism f)
+          (hom_quotient (cong_ker f)).
+
+  Theorem isomorphic_first_isomorphism_inj
+    : A ≅ B & in_image_hom f.
+  Proof.
```
spitters: This and the following one can just be definitions.

spitters:
LGTM

theories/Classes/theory/ua_first_isomorphism.v
Alizter: Is there any difference in behaviour?

andreaslyn: I have changed it now.

> Is there any difference in behaviour?

Same behaviour, but the code becomes a bit shorter with definitions.

theories/Classes/theory/ua_first_isomorphism.v
```diff
@@ -91,11 +91,15 @@ Section in_image_hom.
       by destruct p.
   Qed.
 
-  Global Instance is_closed_under_ops_in_image_hom
+  Lemma is_closed_under_ops_in_image_hom
```
Alizter: Why has this been changed from being an instance?

andreaslyn: I changed `IsClosedUnderOps` to a definition and added the `IsSubalgebraPredicate` type class. Like this the subalgebra argument `P : ∀ s, A s → hProp ` can be changed to `P : ∀ s, A s → Type`, which I think is more consistent with the rest of the development. For instance, the quotient algebra argument `Φ : ∀ s, A s → A s → Type`.

andreaslyn: Do you prefer to keep `IsClosedUnderOps` a type class?

Alizter: I think `IsSubalgebraPredicate` is fine, I see why you have done it now.

PR #1047, opened 2019-07-02
Remove compilation warnings

andreaslyn:
This patch fixes all the compilation warnings.

**EDIT:** except from the warnings from coqdep `declared ML module ... has not been found!`

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -15,9 +15,6 @@ Local Unset Elimination Schemes.
 (** This command changes Coq's subterm selection to always use full conversion after finding a subterm whose head/key matches the key of the term we're looking for.  This applies to [rewrite] and higher-order unification in [apply]/[elim]/[destruct].  Again, if you don't know what that means, ignore it. *)
 Global Set Keyed Unification.
 
-(** This command makes it so that when defining an [Instance], if you give a term explicitly with [:=], then Coq must be able to fill in all the holes in that term.  This is the same as the behavior of [Definition].  (Without this command, if Coq is unable to fill in all the holes in a term given to an [Instance], it silently drops into interactive proof mode.  This is a source of hard-to-find bugs, and you can always obtain equivalent behavior by beginning an interactive proof with [refine].) *)
-Global Unset Refine Instance Mode.
-
 (** This command makes it so that you don't have to declare universes explicitly when mentioning them in the type.  (Without this command, if you want to say [Definition foo := Type@{i}.], you must instead say [Definition foo@{i} := Type@{i}.]. *)
 Global Unset Strict Universe Declaration.
 
@@ -31,7 +28,7 @@ Global Set Default Goal Selector "!".
 Global Set Printing Primitive Projection Parameters.
 
 (** Apply using the same opacity information as typeclass proof search. *)
-Ltac class_apply c := autoapply c using typeclass_instances.
+Ltac class_apply c := autoapply c with typeclass_instances.
 
 Definition relation (A : Type) := A -> A -> Type.
 
--- a/theories/Categories/Category/Morphisms.v
+++ b/theories/Categories/Category/Morphisms.v
@@ -28,7 +28,7 @@ Hint Rewrite @left_inverse @right_inverse : morphism.
 
 Class Isomorphic {C : PreCategory} s d :=
   {
-    morphism_isomorphic :> morphism C s d;
+    morphism_isomorphic : morphism C s d;
     isisomorphism_isomorphic :> IsIsomorphism morphism_isomorphic
   }.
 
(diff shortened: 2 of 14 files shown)


spitters:
LGTM

theories/Basics/Overture.v
```diff
@@ -15,9 +15,6 @@ Local Unset Elimination Schemes.
 (** This command changes Coq's subterm selection to always use full conversion after finding a subterm whose head/key matches the key of the term we're looking for.  This applies to [rewrite] and higher-order unification in [apply]/[elim]/[destruct].  Again, if you don't know what that means, ignore it. *)
 Global Set Keyed Unification.
 
-(** This command makes it so that when defining an [Instance], if you give a term explicitly with [:=], then Coq must be able to fill in all the holes in that term.  This is the same as the behavior of [Definition].  (Without this command, if Coq is unable to fill in all the holes in a term given to an [Instance], it silently drops into interactive proof mode.  This is a source of hard-to-find bugs, and you can always obtain equivalent behavior by beginning an interactive proof with [refine].) *)
```
Alizter: What is the reason for removing this?

SkySkimmer: The option was removed (with the resulting behaviour being the way hott likes it)

theories/Categories/Category/Morphisms.v
```diff
@@ -28,7 +28,7 @@ Hint Rewrite @left_inverse @right_inverse : morphism.
 
 Class Isomorphic {C : PreCategory} s d :=
   {
-    morphism_isomorphic :> morphism C s d;
+    morphism_isomorphic : morphism C s d;
```
Alizter: I am mostly okay with the other changes in coercions, but is there a reason for this one?

Alizter:
I redirected the errors from a fresh build, this is what I got:
```
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
*** Warning: in file coq/theories/Init/Prelude.v, declared ML module numeral_notation_plugin has not been found!
*** Warning: in file coq/theories/Init/Prelude.v, declared ML module numeral_notation_plugin has not been found!
*** Warning: in file coq/theories/Init/Notations.v, declared ML module ltac_plugin has not been found!
*** Warning: in file coq/theories/Init/Notations.v, declared ML module ltac_plugin has not been found!
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
File "./theories/Idempotents.v", line 886, characters 2-82:
Warning:
Diff failure: Input string is not lexable
Showing message without diff highlighting
Please report at http://coq.inria.fr/bugs/.
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
find: ‘./coq/toplevel’: No such file or directory
```

I understand we don't have a solution for the ML module ones. But notably we have:
```
File "./theories/Idempotents.v", line 886, characters 2-82:
Warning:
Diff failure: Input string is not lexable
Showing message without diff highlighting
Please report at http://coq.inria.fr/bugs/.
```
I don't understand what the coq/toplevel warnings are about.

theories/Categories/Category/Morphisms.v
andreaslyn: In type-classes the `:>` syntax only makes sense to use when the type to the right of `:>` is a type-class. And `morphism` is not a type-class, just a field of `PreCategory`.

andreaslyn:
I know a way to get rid of the ML module warnings, but it not not a good solution.

The problem is that `coqdep` seems to assume that ML modules are local to the project. So it looks for a `.mllib` or `.mlpack` file, but is unable to find such a file because these files are part of the coq source code.

One solution is to add empty "dummy" files `numeral_notation_plugin.mlpack` and `ltac_plugin.mlpack` to make `coqdep` shut up. This however seems very confusing. A better solution is to add a flag to `coqdep`, which we can use to tell it the location of the `.cmxs` ML libraries, and such that `coqdep` knows not to look for the `.mlpack` and `.mllib` files.

@Alizter Which commands did you use for the fresh build?

Alizter:
@andreaslyn I only rebuilt the library, by cleaning and building as usual.
```
make clean
sudo make install -j4
```

Alizter:
I had a look in Idempotents. It seems that error was intentional on line 855, as a test. The diff failure is a bug in coq however it seems. It seems to occur when trying to diff with universe annotations.

I don't know how to silence things that ought to fail during compilation.

andreaslyn:
> I don't know how to silence things that ought to fail during compilation.

This is what the `Fail` command is for. I am unable to reproduce that warning.

PR #1049, opened 2019-07-04
Avoid relying on `Export` bug

maximedenes:
See https://github.com/coq/coq/issues/10480

This is in preparation for a fix (https://github.com/coq/coq/pull/10476), but should be backward compatible (to be tested).


spitters:
LGTM

PR #1055, opened 2019-07-23
improved HomotopyGroup.v and added Pi1S1

Alizter:
I have rewritten HomotopyGroup.v and added some lemmas such as the distribution over products. I have added PiS1.v in /Homotopy/ too. In order to do this I have added a few lemmas to Pointed.v.

I am thinking about possibly breaking up Pointed.v into smaller files which will easier to navigate and work with. (They will still be indexed by Pointed.v).

Once the Torus PR is pulled, I will also add a computation of Pi1 of the Torus later. I currently also have a half finished file for PinSn lying around which I may finish soon.

### DIFF
--- a/theories/Homotopy/HomotopyGroup.v
+++ b/theories/Homotopy/HomotopyGroup.v
@@ -4,100 +4,136 @@ Require Import HIT.Truncations.
 Require Import Pointed.
 Require Import abstract_algebra.
 Require Import Spaces.Nat.
-Require Import DProp.
+Require Import UnivalenceAxiom.
+
+Import TrM.
+
 (* In this file we define homotopy groups *)
 
-Section HomotopyGroups.
+Definition Pi n (X : pType) := Tr 0 (iterated_loops n X).
+
+Section PiGroup.
 
   Context
     (n : nat)
-    (X : Type)
-   `{IsPointed X}.
+    (X : pType).
 
-  Definition Pi := Tr 0 (iterated_loops n X).
-
-  (* We only have a group structure when 0 < n *)
-  Context `{gt : 0 < n}.
+  Global Instance PiUnit : MonUnit (Pi n.+1 X).
+  Proof.
+    by apply tr.
+  Defined.
 
   (* We explicitly give the operation here *)
-  Global Instance PiOp : SgOp Pi.
+  Global Instance PiOp : SgOp (Pi n.+1 X).
   Proof.
-    intros a b.
-    strip_truncations.
-    refine (tr _).
-    destruct n, gt.
-    exact (concat a b).
+    intros a b; strip_truncations.
+    exact (tr (a @ b)).
   Defined.
 
   Global Instance PiOp_assoc : Associative PiOp.
   Proof.
-    intros x y z.
-    strip_truncations.
-    unfold PiOp; cbn.
-    apply ap.
-    destruct n, gt.
-    refine (concat_pp_p _ _ _)^.
-  Defined.
-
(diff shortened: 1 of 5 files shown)


Alizter:
@spitters @mikeshulman Looks good?

mikeshulman:
Haven't had a chance to look at it yet, will hopefully do this week.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -4,100 +4,136 @@ Require Import HIT.Truncations.
...
+  intros x y; strip_truncations; cbn.
+  apply ap, eckmann_hilton.
+Defined.
+
+Global Instance pi_is_AbGroup n X : AbGroup (Pi n.+2 X).
+Proof.
+  serapply Build_AbGroup.
+Defined.
+
+(* TODO:
```
andreaslyn: Did you forget to move this to Pointed.v?

theories/Homotopy/Pi1S1.v
```diff
@@ -0,0 +1,41 @@
...
+    unfold Pi; cbn.
+    equiv_via (base = base).
+    1: symmetry; exact (equiv_tr 0 (loops (S1, base))).
+    exact equiv_loopS1_int.
+  Defined.
+
+  Theorem Pi1S1' : Pi 1 (psphere 1) <~> Int.
+  Proof.
+    unfold psphere.
+    assert (X : (Sphere 1, North) = (S1, base)).
```
andreaslyn: Maybe this could be a separate lemma somewhere?

theories/Pointed.v
```diff
@@ -323,11 +375,23 @@ Proof.
   apply moveR_equiv_V; symmetry; apply point_eq.
 Defined.
 
+(* Notation "f ^-1" := (pequiv_inverse f). *)
```
andreaslyn: Maybe we should remove this comment.

theories/Homotopy/HomotopyGroup.v
Alizter: No, it needs to be here for now. I have another PR coming soon where it is moved.

theories/Homotopy/Pi1S1.v
Alizter: I would argue that this is really the correct result, and S1 is a lemma. But the loop space of the circle was proven, not the loop space of the suspension of 2. I am not sure it should go elsewhere.

theories/Pointed.v
Alizter: I have a PR coming up for pointed in which this will be included so it is harmless to leave for now.

theories/Homotopy/Pi1S1.v
andreaslyn: Sorry, I mean the `assert` could be a separate lemma.

Alizter: OK fair enough. I think it would be fine in Spheres.v

andreaslyn: Stated as an equivalence `(Sphere 1, North) <~> (S1, base)`. I guess that this results is more related to HIT/Sphere.v.

Alizter: Once it is a pointed equivalence it is only path_ptype away from being a path anyway.

andreaslyn: Yes, a pointed equivalence `(Sphere 1, North) <~>* (S1, base)`, of course.

Issue #1057, opened 2019-07-26
Windows


spitters:
I got a request on using HoTT with coqide on Windows.
@JasonGross are you aware of any progress ?
Alternatively, we can suggest them to use jscoq, but we would need to ask @ejgallego to update it to a recent version.

kristinas:
I sent an email to @ejgallego, hopefully he can get back to me.

JasonGross:
I haven't used CoqIDE on Windows in quite a while, so I'm not sure.

kristinas:
Ok, we are moving towards using JsHoTT so I suppose the issue is moot now. Thanks Bas and Jason!

Alizter:
The latest versions of windows 10 have the inbuilt linux kernel. It may be possible in the future to build everything on that. That would somewhat count as support for windows.

JasonGross:
Last I checked the Linux subsystem did not support non-terminal graphical apps.

mikeshulman:
Why not use Emacs+ProofGeneral?  Last I checked Emacs works just fine on Windows.

Alizter:
I think Kristina is looking for the easiest solution to teach usage of the library on. Emacs is generally, in my experience, a steep learning curve. I think it is a sensible idea to use CoqIDE for such a task. But it looks unlikely that will be working on windows anytime soon.

It would however be not too bad if Emacs+ProofGeneral was used with some effort put into how to use it. Perhaps a cheatsheet of frequently used keyboard commands, and how Emacs works? I don't doubt that once you learn Emacs it is a fine way of using the library (if not better).

mikeshulman:
Learning all the ins and outs of Emacs is a long process and never complete (I just learned a new trick last week).  But you don't need any of the fancy stuff just to edit files; you just type stuff.  (Unlike, say, vi.)  You don't even need to know any key commands; there are graphical menus.  I guess for ProofGeneral you probably need to know a couple key commands since you don't want to be going to the menus every time you want Coq to process a line, but that's true in CoqIDE as well; in fact I think many of CoqIDE's commands were copied from Emacs.

Using Coq in a browser is okay for the duration of a tutorial, I guess, but if the aim is to teach the students to be successful long-term users of Coq in their own right, it would be much better to set them up with an install on their own computers.

kristinas:
[...]

PR #1058, opened 2019-07-27
Fix handling of --datadir in autoconf and configure

MSoegtropIMC:
in Install.md the option --datadir to configure is described. But it does not work. The variable datadir is set but not used. This PR fixes this.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -11,7 +11,7 @@ if make_hoqchk
   bin_SCRIPTS += hoqchk
 endif
 
-hottdir=$(datarootdir)/hott
+hottdir=$(datadir)/hott
 EXTRA_DIST = coq theories etc LICENSE.txt CREDITS.txt INSTALL.txt README.markdown
 
 # The path to the Coq library in $(srcdir)
--- a/configure.ac
+++ b/configure.ac
@@ -172,7 +172,7 @@ AS_IF([test "x$COQMAKEFILE" = "xno"],
       [AC_MSG_ERROR([Could not find coq_makefile])])
 
 
-hottdir="$datarootdir/hott"
+hottdir="$datadir/hott"
 AC_SUBST([hottdir])
 
 AC_CHECK_PROGS(ETAGS,etags,[: skipping etags])
--- a/hoq-config.in
+++ b/hoq-config.in
@@ -4,6 +4,7 @@
 # options.  One day we might figure out how to create an honest Coq
 # toplevel instead.
 prefix="@prefix@"
+datarootdir="@datarootdir@"
 export COQC="@COQC@"
 export COQCHK="@COQCHK@"
 export COQDEP="@COQDEP@"


Alizter:
It seems travis isn't happy with changing that.

MSoegtropIMC:
> It seems travis isn't happy with changing that.

I need to check what this is - I tested it both with and without configure options locally and it did work. Maybe it is because travis uses slightly different build commands than given in Install.md (I followed that). I will try the same commands Travis uses locally.

MSoegtropIMC:
The additional redirection (hottdir -> datadir -> datarootdir -> prefix) is not handled by auto tools. It seems to handle only one level. Additional levels need to be added as variable assignment in configure input files like hoq-config.in. `prefix` was already there before and I added `datarootdir`. This should work now. The automake tool does this automatically for the make file.

MSoegtropIMC:
Thanks for the quick merge - I removed the patch from the Coq Windows build.

Btw.: I added HoTT to the official Coq Windows installers - it might make it into 8.10 but will definitely be in 8.10.1. I hope there are no objections against this from your side. I did this on request of someone giving lectures on HoTT. This includes batchfiles to replace your hoq___ scripts, which don't work on Windows without a bash. I mostly tested hoqide.bat which seems to work fine. Maybe someone who understands HoTT can have a look at this. An installer is available from:

[64 bit](https://gitlab.com/coq/coq/-/jobs/260656282/artifacts/browse/artifacts/)
[32 bit](https://gitlab.com/coq/coq/-/jobs/260618509/artifacts/browse/artifacts/)

This also means that your installation instructions on Windows will be outdated. Shall I provide a patch for your instructions - possibly including instructions on how to build it from scratch?

spitters:
That would be great. Does this also solve the coqide issue?
https://github.com/HoTT/HoTT/issues/1057

MSoegtropIMC:
> That would be great.

I take this as "you don't have objections" :-)

> Does this also solve the coqide issue?

The issue is not very descriptive, so it is hard to tell what the issue was. What I tested successfully is: start coqide via the hoqide.bat file I supply (in the bin folder) and run through the randomly chosen file `contrib/Freudenthal.v`. This successfully loads some HoTT libraries, so it seems to basically work. I would appreciate if someone does a few more tests if HoTT really works (I linked installers above - it only takes a minute to install) or sends me some files I should test run in CoqIDE.

The batch files btw read:
```
SET HERE=%~dp0/..
start "hoqide" "%HERE%/bin/coqide" -indices-matter -coqlib "%HERE%/lib/coq/user-contrib/hott/coq" -R "%HERE%/lib/coq/user-contrib/hott/theories" HoTT -Q "%HERE%/lib/coq/user-contrib/hott/contrib" "" %*
```
for coqide and for coqc:
```
SET HERE=%~dp0/..
"%HERE%/bin/coqc" -indices-matter -coqlib "%HERE%/lib/coq/user-contrib/hott/coq" -R "%HERE%/lib/coq/user-contrib/hott/theories" HoTT -Q "%HERE%/lib/coq/user-contrib/hott/contrib" "" %*
```
The installer has a fixed folder structure, so it is quite a bit easier than the corresponding shell scripts.

spitters:
The issue was: is there any progress on getting coqide to work. If so, it
could be used in the HoTT school.
https://hott.github.io/HoTT-2019/

From your description, it sounds like it now works. ( I have no way of
further testing it myself. )

MSoegtropIMC:
> If so, it could be used in the HoTT school.

Can you tell me something about the build complexity of the lecture projects:

1. trivial, that is mostly independent individual files or short dependency chains

2. simple, that is something which can be build with a batch file or make file using CMD shell

3. complex, that is using fancy shell stuff in make files

spitters:
@kristinas ?

MSoegtropIMC:
I did some more tests, especially on building. As it looks the CoqIDE "compile buffer" function does not work cause of the `-Q "%HERE%/lib/coq/user-contrib/hott/contrib" ""`. The `""` is not properly quoted when coqide calls coqc. So running a file within coqide works, but coqide cannot cannot call coqc to create .vo files.

There are a few solutions to this:

1.) remove the `-Q "%HERE%/lib/coq/user-contrib/hott/contrib" ""` - not sure if the access to contrib is required
2.) rename it to `-Q "%HERE%/lib/coq/user-contrib/hott/contrib" "contrib"` - anything which is not a space will do
3.) Don't use CoqIDE compile buffer, but call hoqc from the command line or a batch file
4.) Hack quoting it in CoqIDE yet once more
5.) Fix quoting in CoqIDE
6.) Add an alias for `""` in coqc

kristinas:
Hi Michael,

For the lecture, the build complexity would be trivial.

However, I was also asking about this for myself. I am a Windows user who prefers to use CoqIDE as opposed to Emacs/Proof General and would prefer not to get a Linux distribution just to be able to use HoTT Coq on my machine. It would be great if there would be a full-fledged Windows distribution for HoTT Coq that includes CoqIDE. The discussion above seems to suggest there is indeed a problem with CoqIDE on Windows, is this correct?

Thanks for your help!

MSoegtropIMC:
> The discussion above seems to suggest there is indeed a problem with CoqIDE on Windows, is this correct?

There is a minor problem. Compilation with "compile buffer" does not work when one maps the contrib folder to the logical root, that is the `-Q "%HERE%/lib/coq/user-contrib/hott/contrib" ""` option in the hoqide.bat file. There are various work arounds as described above. In case you don't need this mapping or can live with a renamed mapping, just edit the hoqide.bat file. As far as I can tell everything else works. Please remember to start CoqIDE via hoqide.bat from the bin folder.

Please test one of the installers I linked above and see if they work. If not please let me know asap.

We can also discuss what build methods you expect from the Windows Installer for your daily work. Mid term we want to bundle dune with it and use dune for building Coq projects. This would solve the shell issues one has with make. Until then I can provide a restricted make.

kristinas:
I don't know if I;m doing something wrong but it doesn't seem to work for me. I installed the 64 installer from the link provided and when I type

From HoTT Require Import Basics.

into CoqIDE I get the error

Cannot find a physical path bound to logical path matching suffix <> and prefix HoTT.

Alizter:
@kristinas What about just typing Require Import Basics.

kristinas:
That seems to work but what about files appearing in directories? For example, 

From Types Require Import Paths.

gives an error,

as does

Require Import Types.Paths.

Alizter:
I would be less inclined to use the From syntax. For Basics and Types it is best to just import the whole thing.

Require Import Basics.
Require Import Types.

For files appearing in directories, say HIT.Circle you can say Require Import HIT.Circle or even Require Import Circle.

Could you tell me what error Types.Paths is giving you?

kristinas:
I tried Require Import HIT.Circle and get the error

"Cannot find a physical path bound to logical path matching suffix HIT."

When I try Import Circle. I get 

"Unable to locate library Circle."

Which is unsurprising I guess.

Alizter:
Can you tell me what happens with the following?

Require Import HoTT.Basics.Overture.
Require Import HoTT.Basics.
Require Import HoTT.Types.
Require Import HoTT.Types.Paths.

kristinas:
"Cannot find a physical path bound to logical path matching suffix HoTT.Basics."

Alizter:
Yes you are running the wrong program :). You are running CoqIde which is using regular coq. You need to run Hoqide which starts a custom configuration of Coqide.

kristinas:
Oh, I thought since the installation includes the HoTT package that gets integrated.

Alizter:
Are you able to run Hoqide?

kristinas:
Yes, I have to start it from the Coq/bin folder, though, there isn't an associated icon like with CoqIDE. It does seem to work though! Thanks a lot.

Alizter:
If Hoqide can't find coq then you haven't done the configure step properly.
` ./configure COQBIN="`pwd`/coq-HoTT/bin"`
You shouldn't need to touch anything in Coq/bin

kristinas:
where should I run this?

Alizter:
Just in HoTT

kristinas:
windows doesn't recognize the ./ command

Alizter:
You will have to run configure as a bash script then

Alizter:
I don't actually know how you installed the library btw, I am only guessing. Which part of INSTALL.md did you follow?

kristinas:
I didn't follow any part of the INSTALL.md.  I ran the Windows installer from Michael, I thought that was supposed to take care of things.

Alizter:
In which case disregard what I have said. Are you sure Hoqide is working now? (Even where you are running it from).

kristinas:
It appears so. I will test it a bit more over the next few days to see if I don't run into issues but so fa haven't found any.

kristinas:
Thank you @MSoegtropIMC!

MSoegtropIMC:
@kristinas : I am happy to hear that it works! A few notes:

- The file <coqroot>/bin/hoqide.bat can be started just there with a double click. I didn't provide a shortcut because this is a generic Coq Installer, not a HoTT specific one (and also because it is a bit of work to do so). It shouldn't be hard to make a shortcut to the batch file manually for those who work on HoTT.

- There is still the issue with the contrib folder. Can you please share your thoughts if it would be ok to "mount" it as Contrib rather than ""? One would have to write `Require Contrib.Freudenthal` instead of `Require Freudenthal` then. Everything except for the contrib folder should be ok and compatible with Linux. The "Compile Buffer" function of CoqIDE would work then. I am not sure I can properly fix it (so that "" works) until the course starts.

andreaslyn:
@MSoegtropIMC I think it is good to name it `Contrib` rather than `""`.

MSoegtropIMC:
Just a note: the issue with the `-Q <path> ""` option and CoqIDE's compile buffer command also seems to exist on Linux. See the discussion at (https://github.com/coq/coq/issues/10590).

It was questioned there if using -Q with "" is recommended, so you might want to think about this also on Linux.

P.S.: This gets increasingly off topic. I am fine with hijacking my own PR for this discussion but in case you want this discussion to take place elsewhere, please let me know.

Alizter:
@MSoegtropIMC I think creating an issue will be fine. Just reference the PR so we know where we left off.

MSoegtropIMC:
Renaming the -Q logical path to Contrib doesn't work. For `Require Import Contrib.Freudenthal.` I get:
```
The file D:\bin\coq-810-64-hott-3\lib\coq\user-contrib\hott\contrib\Freudenthal.vo contains library
Freudenthal and not library Contrib.Freudenthal
```
Still I would use the installers from below links which have this change so that CoqIDE compile buffer works.
[64-bit](https://gitlab.com/coq/coq/-/jobs/261177798/artifacts/browse/artifacts/)
[32-bit](https://gitlab.com/coq/coq/-/jobs/261177800/artifacts/browse/artifacts/)
In case one needs one of the files in contrib, one can copy the .v files there and recompile them in CoqIDE.

Issue #1054, opened 2019-07-22
Replacing Require Imports using script


Alizter:
Would it be beneficial to replace "index files" such as Basics.v with the parts that are actually needed in a given file? This would streamline compilation since there is no reason why a file should wait for all of Basics to be compiled if it doesn't use it.

There is one way we could do this. Write a script that replaces each instance of an "index file" with its contents (kind of like C #include), then have another script remove unnecessary imports (which surely someone must have written for coq).

I am not saying that we should avoid using index files but rather clean them up now and then where we don't need them.

mikeshulman:
I find it annoying when I am working in the middle of a file and try to use a standard lemma from Basics or Types, only to find that whoever created the file didn't actually import all of Basics or Types, so I have to go edit the imports line and then wait for the whole file to re-load in PG.  And I don't see any real benefit to it in terms of compilation, since everything has to be compiled anyway, Basics/Types don't take very long to compile, and they don't change very often either.

Alizter:
What if these scripts were run temporarily before compilation? As in, the original files don't change but when they are built they are preprocessed. The reason I want something like this is because the index files cause a bottleneck when compiling on more than 4 processors. I suppose its just an idea to float for now.

mikeshulman:
What I'm talking about is not compilation but interactive editing.

Compiling Basics and Types takes less than 20 seconds on my laptop, which is not exactly an up to date machine.  Go get a cup of coffee or something while you wait.  (-:

BTW, of those 20 seconds, about half the time is spent on one line in `Types.Sum.equiv_unfunctor_sum_indecomposable_ll`.  Maybe a significant speedup could be obtained by optimizing that proof.

Alizter:
So I had a look at that proof and coq is mostly spending its time inferring types.

If we replace the destruct with the following:
```
destruct (equiv_unfunctor_sum k'
       (fun x : decompose_l (h o inl) + decompose_l (h o inr) =>
        match x as x0 return (is_inl (k' (inl x0))) with
        | inl x0 => x0.2
        | inr x0 => x0.2
        end)
       (fun x : decompose_r (h o inl) + decompose_r (h o inr) =>
        match x as x0 return (is_inr (k' (inr x0))) with
        | inl x0 => x0.2
        | inr x0 => x0.2
        end)) as [s t]; clear k k'.
  refine (t oE (_ +E 1) oE g^-1).
```
It is much faster. I will PR this soon.

Alizter:
[...]

PR #1060, opened 2019-07-29
optimised Types/Sum.v

Alizter:
this was highlighted in #1054 . This should save around 6s in compilation.


mikeshulman:
I figured it would probably be something like this.

PR #1061, opened 2019-07-29
split up Pointed

Alizter:
- Split up pointed into multiple files
- Moved Suspension from HITs to Homotopy
- Cleanup in spheres

Since Suspensions are no longer HITs with axioms, but rather defined as pushouts I see no reason to keep it in HITs. This will be the start of moving various "non-primitive" higher inductive types out of that folder. Eventually I want to touch Join and Pushout.


Alizter:
Whoops this needs to wait a bit.

PR #1062, opened 2019-07-30
organising _CoqProject

Alizter:
This was meant to be part of the Torus PR but since that one is making travis cry I will add this change separately whilst I work out what is wrong with the other.

I haven't bothered giving sections to subfolders but those are typically easier to navigate than the rest.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -2,6 +2,10 @@
 -Q contrib ""
 COQC = hoqc
 
+#
+#   Basics
+#
+
 theories/Basics/Notations.v
 theories/Basics/Utf8.v
 theories/Basics/Overture.v
@@ -12,6 +16,11 @@ theories/Basics/Equivalences.v
 theories/Basics/Trunc.v
 theories/Basics/Decidable.v
 theories/Basics.v
+
+#
+#   Types
+#
+
 theories/Types/Paths.v
 theories/Types/Unit.v
 theories/Types/Forall.v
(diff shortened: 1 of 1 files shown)


_CoqProject
```diff
@@ -12,6 +16,11 @@ theories/Basics/Equivalences.v
 theories/Basics/Trunc.v
 theories/Basics/Decidable.v
 theories/Basics.v
+
+#
+#   Types_CoqProject _CoqProject _CoqProject
```
mikeshulman: What?

Alizter: Hmm I have no idea how this got in here. Perhaps I ran a rogue script at some point.

Alizter:
@mikeshulman That's all fixed. The merge feature on github did something weird that is all.

mikeshulman:
LGTM

mattam82:
It was probably not meant to stop compiling and installing the contrib/ files was it? With this  `hoqc`/`hoqtop` now fail to find the `share/contrib` directory

Alizter:
@mattam82 I don't think this has anything to do with this change. I would delete share/contrib and share/theories and reinstall after compiling and see if that changes anything.

mattam82:
Looking at the changes of this PR, you can see that none of the files in contrib/ are built anymore (they’re commented out), so when calling hoqdep, or hoqc after installation, it uses an option -R .../contrib that makes it error fatally because the directory no longer exists.

SkySkimmer:
Why list files in _coqproject instead of doing a find in the makefile? (considering hott doesn't use coq_makefile)

mattam82:
For IDEs I think

SkySkimmer:
I don't think ides look at the list of files

Gaëtan Gilbert

JasonGross:
> Why list files in _coqproject instead of doing a find in the makefile? (considering hott doesn't use coq_makefile)

I am opposed to using `find` because it will pick up non-version-controlled files.  I'd be more okay with `git ls-files`, but we support building from a tar ball, not just git, and it seems slightly poor to have a different file selection mechanism depending on whether or not we use git.

SkySkimmer:
Why is picking up non vc files a problem?

Gaëtan Gilbert

mikeshulman:
Make should only be making files that are part of the library.  A user may have their own .v files in the directory that are not part of the library.

Alizter:
@mattam82 My apologies, you are correct. I got confused since in #1066 I uncommented the contrib lines in the _CoqProject file, instead of doing it here. I should have been more careful.

PR #1065, opened 2019-08-02
Torus is equivalent to product of circles and cubical library

Alizter:
This is my second attempt at getting #1053 to work. I have spent some time debugging it and I see that isequiv_cu_path in Cubical/Cube.v is the culprit. I couldn't reopen my previous PR since I force-pushed to this one.

I can't see any reason why proviola gets stuck however. If you comment out everything after and including intros [] near line 100 and stick an admitted there proviola is fine.


Alizter:
@JasonGross What do you think?

Alizter:
Should I leave the proof as admitted just to make travis happy?

mikeshulman:
We shouldn't keep admits around in the main branch of the library, if that's what you're asking.

JasonGross:
If you replace `intros [].` with `intro x; destruct x` or similar, does that fix the issue with proviola?

Alizter:
OK I can't get this to work anymore. I will close this PR and try again later.

PR #1066, opened 2019-08-03
Torus and cubical

Alizter:
OK 3rd attempt at #1053 and #1065. This time proviola is happy.


Alizter:
For reference, the reason I closed #1053 is because proviola (the script that does the html docs) couldn't parse one of my proofs in Cube.v properly and got stuck. #1065 was my failed attempt at getting it to work again. And this PR is it working now.

The only difference between this and #1053 is the proof of `isequiv_cu_path` which can be found in `Cubical/Cube.v`.

Alizter:
@mikeshulman @andrejbauer @spitters looks good?

Alizter:
@mikeshulman This did occur to me too. I think it would be better to have a folder called Torus in Spaces with both files. I will push something to fix this in a bit.

Alizter:
I have moved the Torus stuff into a Torus folder in Spaces as @mikeshulman suggested. While I was at it I quickly added the calculation of the fundamental group of the torus.

mikeshulman:
LGTM finally!

Alizter:
Uncommenting Cubical in HoTT.v breaks a load of stuff in HoTTBook. I will see why this is later. For now I am going to stop touching this PR so it can finally be merged.

Alizter:
@JasonGross Are you happy with this now?

JasonGross:
I don't have time to review right now, but I don't object to it.  If you can't get APOE, I'll try to review within the next couple of days.

Alizter:
@spitters @andrejbauer Can any of you review this?

Alizter:
Can we merge this now, since @andrejbauer already reviewed the previous #1053 which only differs by some refactoring to make proviola happy?

PR #1076, opened 2019-08-11
A bunch of lemmas about pushouts

mikeshulman:
Pushouts of equivalences are equivalence, pushouts are associative, pushouts have a universal property, and a reflective subuniverse reflects pushouts (into pushouts in the subuniverse, as with other colimits).  I had to break the bit about "set cones" out of Pushout.v to avoid cyclic dependencies with ReflectiveSubuniverse.

### DIFF
--- a/theories/HIT/Pushout.v
+++ b/theories/HIT/Pushout.v
@@ -91,6 +91,52 @@ Proof.
   refine (pushout_ind_beta_pp (fun _ => P) _ _ _ _).
 Defined.
 
+(** ** Universal property *)
+
+Definition pushout_unrec {A B C P} (f : A -> B) (g : A -> C)
+           (h : pushout f g -> P)
+  : {psh : (B -> P) * (C -> P) &
+           forall a, fst psh (f a) = snd psh (g a)}.
+Proof.
+  exists (h o pushl , h o pushr).
+  intros a; cbn.
+  exact (ap h (pp a)).
+Defined.
+
+Definition isequiv_pushout_rec `{Funext} {A B C} (f : A -> B) (g : A -> C) P
+  : IsEquiv (fun p : {psh : (B -> P) * (C -> P) &
+                            forall a, fst psh (f a) = snd psh (g a) }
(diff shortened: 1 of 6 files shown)


theories/HIT/Pushout.v
```diff
@@ -61,23 +73,69 @@ Arguments pushout_ind_beta_pp : simpl never.
...
 Proof.
   unfold pushout_rec.
   eapply (cancelL (transport_const (pp a) _)).
   refine ((apD_const (@pushout_ind A B C f g (fun _ => P) pushb pushc _) (pp a))^ @ _).
   refine (pushout_ind_beta_pp (fun _ => P) _ _ _ _).
 Defined.
 
+(** ** Universal property *)
+
+Definition pushout_unrec {A B C P} (f : A -> B) (g : A -> C)
```
Alizter: Would writing this in curried form change performance?

mikeshulman: I'm not sure what you mean by "curried form".  The *output* here is a Sigma-type because that's the type that's equivalent to `pushout f g -> P`.

Alizter: Ah ok, stupid question then.

theories/HIT/SetCone.v
```diff
@@ -0,0 +1,18 @@
+(* -*- mode: coq; mode: visual-line -*- *)
+Require Import HoTT.Basics.
+Require Import HoTT.Types.
+Require Import HSet TruncType.
+Require Import HIT.Pushout.
+Require Import HIT.Truncations.
+
+(** ** Cones of hsets *)
```
JasonGross:
```suggestion
(** * Cones of hsets *)
```

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1017,6 +1017,32 @@ Section Reflective_Subuniverse.
 
     End OCoeq.
 
+    (* ** Pushouts *)
```
JasonGross:
```suggestion
    (** ** Pushouts *)
```

PR #1077, opened 2019-08-13
Uncomment contrib in _CoqProject

JasonGross:
The contrib files were incorrectly commented out in #1062, preventing hoqc from functioning from an installed HoTT.

See also https://github.com/HoTT/HoTT/pull/1062#issuecomment-520593918


spitters:
LGTM

PR #1078, opened 2019-08-16
Generalized Blakers-Massey and Freudenthal

mikeshulman:
This proof of (G)BM is a bit different from the existing one: by using a pushout to define the left side of the codes fibration, it doesn't need to assume surjectivity of one map until the point of proving contractibility.  I suspect it should be possible to do away with surjectivity there too, but haven't managed to work it out yet.  Anyway, assuming surjectivity is enough for classical BM and hence Freudenthal, which are also included here.  ABFJ use a different trick to discharge the assumption of surjectivity that works for arbitrary modalities (though only for code-right), but I haven't tried to implement that yet.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -65,6 +65,11 @@ Notation "g 'oE' f" := (equiv_compose' g%equiv f%equiv) : equiv_scope.
 Global Instance transitive_equiv : Transitive Equiv | 0 :=
   fun _ _ _ f g => equiv_compose g f.
 
+(** This is sometimes useful to have a name for, so we can rewrite along it. *)
+Definition apply_equiv_compose {A B C} (f : A <~> B) (g : B <~> C) (a : A)
+  : (g oE f) a = g (f a)
+  := 1.
+
 
 (** Anything homotopic to an equivalence is an equivalence. *)
 Section IsEquivHomotopic.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -65,11 +65,11 @@ Notation "g 'oE' f" := (equiv_compose' g%equiv f%equiv) : equiv_scope.
 Global Instance transitive_equiv : Transitive Equiv | 0 :=
   fun _ _ _ f g => equiv_compose g f.
 
-(** This is sometimes useful to have a name for, so we can rewrite along it. *)
-Definition apply_equiv_compose {A B C} (f : A <~> B) (g : B <~> C) (a : A)
-  : (g oE f) a = g (f a)
-  := 1.
-
+Ltac change_apply_equiv_compose :=
+  match goal with
+  | [ |- context [ equiv_fun (?f oE ?g) ?x ] ] =>
+    change ((f oE g) x) with (f (g x))
+  end.
 
 (** Anything homotopic to an equivalence is an equivalence. *)
 Section IsEquivHomotopic.
--- a/theories/HIT/Pushout.v
+++ b/theories/HIT/Pushout.v
@@ -173,22 +173,6 @@ Section EquivPushout.
     + apply q.
   Defined.
 
-  Definition equiv_pushout_pushl (b : B)
-    : equiv_pushout (pushl b) = pushl (eB b)
-    := 1.
-
-  Definition equiv_pushout_pushr (c : C)
-    : equiv_pushout (pushr c) = pushr (eC c)
-    := 1.
-
-  Definition inverse_equiv_pushout_pushl (b : B')
-    : equiv_pushout^-1 (pushl b) = pushl (eB^-1 b)
-    := 1.
-
-  Definition inverse_equiv_pushout_pushr (c : C')
-    : equiv_pushout^-1 (pushr c) = pushr (eC^-1 c)
-    := 1.
-
   Lemma equiv_pushout_pp (a : A)
     : ap equiv_pushout (pp a)
       = ap pushl (p a) @ pp (eA a) @ ap pushr (q a)^.
--- a/theories/HIT/Pushout.v
+++ b/theories/HIT/Pushout.v
@@ -158,26 +158,49 @@ equiv_adjointify pushout_sym_map pushout_sym_map sect_pushout_sym_map sect_pusho
 
 (** Pushouts preserve equivalences. *)
 
-Lemma equiv_pushout {A B C f g A' B' C' f' g'}
-  (eA : A <~> A') (eB : B <~> B') (eC : C <~> C')
-  (p : eB o f == f' o eA) (q : eC o g == g' o eA)
-  : pushout f g <~> pushout f' g'.
-Proof.
-  refine (equiv_functor_coeq' eA (equiv_functor_sum' eB eC) _ _).
-  all:unfold pointwise_paths.
-  all:intro; simpl; apply ap.
-  + apply p.
-  + apply q.
-Defined.
+Section EquivPushout.
+  Context {A B C f g A' B' C' f' g'}
+          (eA : A <~> A') (eB : B <~> B') (eC : C <~> C')
+          (p : eB o f == f' o eA) (q : eC o g == g' o eA).
 
-Lemma equiv_pushout_pp {A B C f g A' B' C' f' g'}
-  {eA : A <~> A'} {eB : B <~> B'} {eC : C <~> C'}
-  {p : eB o f == f' o eA} {q : eC o g == g' o eA}
-  : forall a : A, ap (equiv_pushout eA eB eC p q) (pp a)
-    = ap push (ap inl (p a)) @ pp (eA a) @ ap push (ap inr (q a))^.
-Proof.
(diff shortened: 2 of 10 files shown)


theories/Basics/Equivalences.v
```diff
@@ -65,6 +65,11 @@ Notation "g 'oE' f" := (equiv_compose' g%equiv f%equiv) : equiv_scope.
 Global Instance transitive_equiv : Transitive Equiv | 0 :=
   fun _ _ _ f g => equiv_compose g f.
 
+(** This is sometimes useful to have a name for, so we can rewrite along it. *)
+Definition apply_equiv_compose {A B C} (f : A <~> B) (g : B <~> C) (a : A)
```
Alizter: Is this really necessary? Can the same effect not be achieved with `change`?

Alizter:
We have a problem wiith travis here. @JasonGross Do you know what is causing this?

theories/HIT/Pushout.v
```diff
@@ -158,26 +158,49 @@ equiv_adjointify pushout_sym_map pushout_sym_map sect_pushout_sym_map sect_pusho
...
+    : pushout f g <~> pushout f' g'.
+  Proof.
+    refine (equiv_functor_coeq' eA (equiv_functor_sum' eB eC) _ _).
+    all:unfold pointwise_paths.
+    all:intro; simpl; apply ap.
+    + apply p.
+    + apply q.
+  Defined.
+
+  Definition equiv_pushout_pushl (b : B)
```
Alizter: Same thing with `change` here too. Perhaps instead of getting rid of them, we can replace them with comments saying: "The following are definitionally equal, so use change like this` or something similar.

JasonGross:
Searching for "error" in the Travis log gives
```
File "./theories/Homotopy/BlakersMassey.v", line 2, characters 32-52:
Error: Unable to locate library HoTT.HIT.SpanPushout.
```
Looks like you forgot to `git add` that file?

mikeshulman:
Huh, that's weird.  I distinctly remember adding it.  Maybe it got lost somewhere in munging the commits.  Now I added it, thanks.

theories/Basics/Equivalences.v
mikeshulman: `change` would be much more work to use, since you have to type out the entire subterm and what you want to replace it with.  Check out l596 in `BlakersMassey.v`; if you can find a similarly concise way to achive the same result as `do 6 rewrite apply_equiv_compose` there, I'm all ears.

theories/Homotopy/BlakersMassey.v
```diff
@@ -0,0 +1,677 @@
...
+      destruct inh as [y0 q00].
+      exact (BuildContr _ (center_code p r) (contraction_code q00 (p;r))).
+    Defined.
+
+    (** This version is sufficient for the classical Blakers-Massey theorem, as we'll see below, since its leg-wise connectivity hypothesis implies the above surjectivity assumption.  ABFJ have a different method for eliminating the surjectivity assumption using a lemma about pushouts of monos also being pullbacks, though it seems to only work for coderight. *)
+
+  End GBM.
+End GenBlakersMassey.
+
+(** ** The classical Blakers-Massey Theorem *)
```
Alizter: It might be worth adding this into its own file.

theories/Homotopy/BlakersMassey.v
```diff
@@ -0,0 +1,677 @@
...
+
+Import TrM.
+Module Import BlakersMassey := GenBlakersMassey Truncation_RSUs.
+
+Definition blakers_massey `{Univalence} (m n : trunc_index)
```
Alizter: This can be an instance right?

theories/Homotopy/BlakersMassey.v
```diff
@@ -0,0 +1,677 @@
...
+           (x : X) (y : Y)
+  : IsConnMap (m +2+ n) (@sglue X Y Q x y).
+Proof.
+  intros r.
+  srefine (contr_code_inhab Q (m +2+ n) _ x
+                            (merely_isconnected n _) (spushr Q y) r).
+  intros x1 x3 y2 y4 q12 q32 q34; apply isconnected_join; exact _.
+Defined.
+
+(** ** The Freudenthal Suspension Theorem *)
```
Alizter: This might be better off in its own file too. It doesn't have to be this PR though.

theories/Homotopy/BlakersMassey.v
```diff
@@ -0,0 +1,677 @@
...
+
+Definition freudenthal `{Univalence} (n : trunc_index)
```
Alizter: This can be an instance too.

Alizter:
I'm happy with merging this now. I can help split up the bits I have commented later.

theories/Basics/Equivalences.v
JasonGross: You should be able to write a tactic that is `change_apply_equiv_compose` using `match` and have it just work.  It might be slightly faster and result in a nicer proof term; do either of those matter where you're doing this?

theories/Homotopy/BlakersMassey.v
mikeshulman: I guess so.

theories/Homotopy/BlakersMassey.v
mikeshulman: I thought about that, but I wasn't sure there would be any point, since it's such a trivial corollary.

theories/Basics/Equivalences.v
mikeshulman: No, they don't, at least not at present.  The line where it's used is not slow, and the proof is opaque anyway because of other rewrites.  However, if anyone else wants to write such a tactic, I wouldn't object.

mikeshulman: Oh, what the heck, I wrote the tactic.

mikeshulman:
After some consideration, I decided I'm convinced by the argument to move Freudenthal to its own file, so I updated this PR to do that.

theories/HIT/Pushout.v
mikeshulman: Actually, it seems that I didn't end up using these lemmas anyway; they were part of a different approach that I ended up abandoning.  So I've removed them.

Alizter:
Here is another instance you can add to Freudenthal.v, the connectivity of the loop_susp_unit map. I am not religious about the name I have given, perhaps something like conn_map_loop_susp_unit is better?

```coq
Global Instance conn_map_merid_concat (X : pType) `{IsConnected n.+1 X}
  : IsConnMap (n +2+ n) (fun x => merid x @ (merid (point X))^).
Proof.
  refine (conn_map_compose _ _ (equiv_concat_r (merid (point _))^ _)).
Defined.
```

mikeshulman:
Or should that maybe go in the appropriate `Pointed` file?

Alizter:
@mikeshulman I think you are right, I will add it to pSusp.v when I need it.

Alizter:
@spitters @andrejbauer @andreaslyn @JasonGross anyone else able to review?

theories/Homotopy/BlakersMassey.v
```diff
@@ -0,0 +1,653 @@
...
+  apply (ap equiv_fun) in ev.
+  apply ap10 in ev.
+  specialize (ev u).
+  cbn in ev.
+  exact ev.
+Defined.
+
+(** The second one is not supposed to make any sense, so don't try.  It's just a mess of path algebra that collapses under J; we'll see later where it gets used. *)
+
+Lemma unfrobnicate {A : Type} {x y : A} (p : x = y)
```
andreaslyn: Lemma `unfrobnicate` seems rather specialized for one particular use below. I think it is better to remove this lemma and replace `apply unfrobnicate` with `now destruct (glue q01)` below. This will make the code shorter and less messy IMO.

Alizter:
I have a PR in Mike's branch implementing Andreas' changes and adding the Freudenthal corollaries I have mentioned. 

https://github.com/mikeshulman/HoTT/pull/11

theories/Homotopy/BlakersMassey.v
mikeshulman: I didn't know (or had forgotten) you can `destruct` something that's not a variable.  I tried for a while to find some tactic that would manually "generalize" over `glue q01` so that I could destruct it, but couldn't figure it out.  But somehow `destruct` can do that by itself?

mikeshulman:
I implemented the suggestion myself.  (Some comments also had to be modified.)

Alizter:
One of the travis jobs got stuck installing, can somebody restart it?

theories/Homotopy/BlakersMassey.v
andreaslyn: Yes, destruct can often figure out how to do that.

andreaslyn:
I will try to close the PR and reopen it. Apparently, this should trigger a new build: https://stackoverflow.com/questions/17606874/trigger-a-travis-ci-rebuild-without-pushing-a-commit

Alizter:
All good we can merge.

PR #1079, opened 2019-08-21
Reorganised Pointed and added Smash and Wedge

Alizter:
I have split up the Pointed.v file into multiple files:

 - Core.v - contains all the basic pointed deifnitions
 - Loops.v - contains the definition of loop space and lemmas about them
 - pEquiv.v - contains lemmas about pointed equivalences
 - pFiber.v - contains lemmas about pointed fibers
 - pHomotopy.v - contains lemmas about pointed homotopies
 - pMap.v - contains lemmas about pointed maps
 - pSusp.v - contains lemmas about pointed suspensions
 - pTrunc.v - contains lemmas about pointed truncations

Notably, all the loop-susp adjunction stuff has been moved from HIT.Suspension into pSusp. Pointed.v is now an index file for all of the above.

I have also done some general clean up and compatibility changes.

This PR also contains the definition of the Wedge sum in Homotopy/Wedge.v and the definition of the Smash product in Homotopy/Smash.v with its basic induction principles.

The definition of strunc in Spectrum.v was changed to give the explicit definition. This saves time in compilation, by not having typeclasses guess, at the cost of some readability.

As always, let us discuss the merits and flaws of this PR, I am happy to change any things that people are not happy with.


theories/HIT/Suspension.v
```diff
@@ -4,41 +4,12 @@
...
-:= fun y => (match y return (_ -> P y)
-     with North => (fun _ => H_N) | South => (fun _ => H_S) end) H_merid.
-
-Axiom Susp_ind_beta_merid : forall {X : Type} (P : Susp X -> Type)
-  (H_N : P North) (H_S : P South)
-  (H_merid : forall x:X, (merid x) # H_N = H_S)
-  (x:X),
-apD (Susp_ind P H_N H_S H_merid) (merid x) = H_merid x.
-
-End Suspension. *)
```
mikeshulman: I like keeping around the commented-out direct definition.  For instance, it's what's given in the book.

theories/HIT/Suspension.v
```diff
@@ -64,7 +35,8 @@ Proof.
   serapply pushout_ind_beta_pp.
 Defined.
 
-(** We want to allow the user to forget that we've defined suspension in this way. *)
+(** We want to allow the user to forget that we've defined suspension
+    in this way. *)
```
mikeshulman: According to `STYLE.md`, comments should not contain hard newlines.

theories/Homotopy/Smash.v
```diff
@@ -0,0 +1,220 @@
...
+  Definition smash_ind_beta_gluer {P : Smash X Y -> Type}
+    {Psm : forall a b, P (sm a b)} {Pl : P auxl} {Pr : P auxr}
+    (Pgl : forall a, gluel a # Psm a pt = Pl)
+    (Pgr : forall b, gluer b # Psm pt b = Pr) (b : Y)
+    : apD (smash_ind Psm Pl Pr Pgl Pgr) (gluer b) = Pgr b
+    := pushout_ind_beta_pp _ _ _ _ _.
+
+  (* TODO: Move *)
+  Definition transport_inv {A} {P : A -> Type} {x y : A} {p : x = y}
+    {u : P x} {v : P y} (r : p # u = v) : p^ # v = u.
```
mikeshulman: I'm not convinced that this needs its own name.

theories/Homotopy/Smash.v
```diff
@@ -0,0 +1,220 @@
...
+  Proof.
+    apply moveR_transport_V.
+    symmetry; assumption.
+  Defined.
+
+  (* TODO: Move, rename *)
+  Definition transport_concat {A} {P : A -> Type} {a1 a2 a3 : A} {p1 : a1 = a2}
+    {p2 : a2 = a3} {b1} {b2} {b3} (r1 : p1 # b1 = b2) (r2 : p2 # b2 = b3)
+    : transport P (p1 @ p2) b1 = b3.
```
mikeshulman: I thought we already had this under the name `concatD`, but now I can't find it.  But now that we have the start of a `Cubical` library, wouldn't it be better to use `dpath`s instead?

theories/Pointed/Loops.v
```diff
@@ -0,0 +1,367 @@
...
+
+(** And likewise the connectedness.  *)
+(* Note: We give the definition explicitly since it was slow before. *)
+Global Instance isconnected_loops_functor `{Univalence} {n : trunc_index}
+  (A B : pType) (f : A ->* B) `{IsConnMap n.+1 _ _ f}
+  : IsConnMap n (loops_functor f) := fun (p : loops B) =>
+    isconnected_equiv' n _ (equiv_functor_sigma' 1
+    (fun q => equiv_moveR_Vp _ p _)) (isconnected_equiv' n _
+    (equiv_functor_sigma' 1 (fun q => equiv_moveR_pM _ _ _))
+    (isconnected_equiv' n _ (hfiber_ap _)^-1 (isconnected_paths _ _))).
```
mikeshulman: This is hard for me to read.  I would prefer using tactics (I hope it should still be possible to give each part explicitly that way, e.g. using `srefine`), but if we have to give a single long proof term I would prefer to format it in a way that displays the nesting of arguments, e.g. traditional lisp-style

```
(foo a b
  (bar c d
    (baz x y)
    (qux z w))
  (kmo f g))
```

theories/Spectra/Spectrum.v
```diff
@@ -35,10 +37,12 @@ Existing Instance to_is_spectrum.
...
-Defined.
+Definition strunc `{Univalence} (k : trunc_index) (E : Spectrum) : Spectrum
+  := Build_Spectrum
+      (Build_Prespectrum (fun n : nat => pTr (trunc_index_inc n k) (E n))
+        (fun n : nat => ptr_loops (trunc_index_inc n k) (E n.+1)
+          o*E ptr_pequiv (trunc_index_inc n k) (equiv_glue E n)))
+      (fun n : nat => pointed_isequiv (pTr (trunc_index_inc n k) (E n))
+        (loops (pTr (trunc_index_inc n k).+1 (E n.+1)))
+        (ptr_loops (trunc_index_inc n k) (E n.+1)
+          o*E ptr_pequiv (trunc_index_inc n k) (equiv_glue E n))).
```
mikeshulman: Would it speed things up equally to give the explicit terms at the appropriate place in the tactic script?  I always prefer constructing large terms like this using tactics since the reader can step through to see what pieces are satisfying what goals.

theories/HIT/Suspension.v
Alizter: That seems sound to me.

theories/HIT/Suspension.v
Alizter: I was not aware of this, thanks for letting me know.

theories/Homotopy/Smash.v
Alizter: We have concatenation of dependent paths in the cubical library. The reason Smash doesn't use it is because I wrote it before I wrote the cubical library :-)

theories/Homotopy/Smash.v
Alizter: I can't remember why I included this, I will check it out tonight,

theories/Pointed/Loops.v
Alizter: I am happy to spend some time writing this with tactics and find a compromise.

theories/Spectra/Spectrum.v
Alizter: I agree, will change.

Alizter: I managed to change it back to the original tactic based definition. I sped up the exact by giving isequiv_compose. It still hangs but much less.

theories/Pointed/Loops.v
Alizter: I couldn't write this out with tactics without being painfully slow. I have indented it like you suggested however.

Alizter:
I rebased quickly to make certain that things are working.

mikeshulman:
Thanks, LGTM!

At some point we should probably consider going through all the HITs and rephrasing their induction principles using DPaths.

Alizter:
@mikeshulman I think keeping both might be better. In my attempt at the 3x3 lemma, I have extra induction and computation rules for pushouts phrased in terms of dependent paths. I don't think it would be productive to get rid of the transport versions.

Alizter:
I have finished my formalisation of EM spaces, I will create a PR once this is merged.

andreaslyn:
Nice work!

Issue #1085, opened 2019-09-05
Organising truncations


Alizter:
Here are some suggestions on the organisation of truncations. This is somewhat related to #1067.

Firstly I would like to create a folder called Truncations in theories. This would contain the following files:
 - Core.v for the definition of Truncation and basic properties (i.e. what is currently in HIT.Truncations).
 - Truncations.v for various lemmas deriving from truncations
 - Connectedness.v for various lemmas about connectedness
 - TruncIndex.v for various lemmas about trunc indexes, their arithemetic etc.

I would also like to make some changes to `trunc_index_leq` by declaring it to be a class.

These would all be packaged up in an index file called Truncations.v, so whenever you want a lemma about truncations, Require Importing Truncations.v will cover it for you.

I am open to keeping the definition of truncation (what would be in Core.v) in HIT since it is a "raw" HIT as discussed in #1067.

A seperate file for the arithmetic of trunc indexes would be beneficial since it is not require by most things that import Basics.

Please tell me how you like this idea, and any suggestions you might have.

Alizter:
@mikeshulman The difference being, Core contains the definition and Truncations (or Truncation) contains lemmas about it. It could be that the later depends on TruncIndex for example so to save these all being stuffed into a big file I suggest splitting these up.

mikeshulman:
Hmm.  I don't like the name Truncations/Truncations, it is very uninformative.  Can you give a more descriptive name to those lemmas?

Alizter:
@mikeshulman Actually putting them in Core shouldn't be such an issue. Core can depend on TruncIndex stuff, and connectedness will depend on them both most likely.

Issue #661, opened 2014-12-31
On the definition of groups


mikeshulman:
A radical suggestion:

``` coq
Record Group :=
  { classifying_space : Type ;
    grpt : classifying_space ;
    isconn_classifying_space : IsConnected 0 classifying_space ;
    istrunc_classifying_space : IsTrunc 1 classifying_space }.

Definition group_set (G : Group) : Type
  := grpt G = grpt G.

Coercion group_set : Group >-> Sortclass.
```

spitters:
@andrejbauer In general, I agree with your dislike for particular implementations.
About the sayings. It's [you can't have your cake and eat it](http://en.wikipedia.org/wiki/You_can%27t_have_your_cake_and_eat_it) in English. We say: to save the goat and the cabbage. Here's the [xkcd solution to the problem](http://xkcd.com/1134/).

Alizter:
Has this been solved by #1122? And since we are implementing classifying spaces soon, we can also show that every group is isomorpic to `loops (B G)` which means we can use the advantages of Mike's suggestion whenever.

mikeshulman:
Yes, I guess we've decided to use hottclasses groups to fit them into the algebraic hierarchy.

PR #1086, opened 2019-09-05
Eilenberg-MacLane spaces

### LINKED ISSUE #984: Proving Lemma 8.5.5
I have been stuck on this for a few weeks now:

```
Section HSpace.
(*   Context `{Funext}. *)

  Class IsHSpace (space : Type) := {
    id : space;
    mu : space -> space -> space;
    left_id : forall a, mu id a = a;
    right_id : forall a, mu a id = a
  }.

  Context {A : Type}
         `{IsHSpace A}
         `{IsConnected 0 A}.

  Lemma mu_l_equiv' : forall (a : A), IsEquiv (mu a).
  Proof.
    intro.
  Admitted.

  Lemma mu_r_equiv' : forall (a : A), IsEquiv (fun x => mu x a).
  Proof.
    serapply (conn_map_elim -1).
  Admitted.
```

It is quite difficult to follow the book proof, as it is not so specific. Any pointers on how to prove this?

Alizter:
Here is a formalisation of Eilenberg-MacLane spaces:

We have the following new files:

**HSpace.v**
Here we define a type X being a H-Space, and show that for a 0-connected X, we have left and right multiplications inducing equivalences. This closes #984. I have an alternate branch using mathclasses stuff, but I feel like I haven't really gotten it to work well yet. So there is room for improvement here. We also define what is called a "coherent H-space" in Licata-Finster.

**ClassifyingSpace.v**
Here we define the HIT `ClassifyingSpace` which turns a group G into its classifying space. We also define a pointed version called `B G`. We have the standard induction and computation rules for these as well (helped greatly by the cubical library).

We prove:
 - Various lemmas about inverses and identities (standard group theory).
 - BG is 0-connected and 1-truncated.
 - G is equivalent to the loop space of BG
 - BG has a coherent H-space structure when G is abelian.

Note that any group can now be made into a loop space making the arguments for loop space groups in #661 work for any group.

**EMSpace.v**
We define the Eilenberg-MacLane spaces by induction on the naturals.

We prove:
 - K(G,n) is (n-1)-connected and n-truncated.
 - the loop space of K(G,n+1) is K(G,n).

Note we have an trunc lemma floating around, but I hope to clean this out when we implement changes discussed in #1085.

It is worth noting that I (and probably many others) thought that K(G,n+1) being the delooping of K(G,n) was a simple application of Freudenthal. This is not the case, and in `pSusp.v` we have a whole section proving that for X, a 0-connected, 1-truncated coherent H-Space, we have `pTr 1 (loops (psusp X)) <~>* X`. This isn't so easy to prove, and requires and encode decode argument. What isn't mentioned in Licata-Finster is that we are defining n-Type families, not just Type families which makes some lemmas a bit tricker to prove. I have called this lemma `licata_finster` since it is this one special case.

Awaiting your suggestions and feedback!

P.S. I have some comments in ClassifyingSpace.v about HSpaces near the top which are complete nonsense and I forgot to remove. I will do this later however.

### DIFF
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -17,6 +17,23 @@ Fixpoint trunc_index_add (m n : trunc_index) : trunc_index
 
 Notation "m +2+ n" := (trunc_index_add m n) : trunc_scope.
 
+Definition trunc_index_add_minus_two m : m +2+ -2 = m.
+Proof.
+  induction m.
+  1: reflexivity.
+  cbn; apply ap.
+  assumption.
+Defined.
+
+Definition trunc_index_add_succ m n : m +2+ n.+1 = (m +2+ n).+1.
+Proof.
+  revert m.
+  induction n; induction m.
+  1,3: reflexivity.
+  all: cbn; apply ap.
+  all: assumption.
+Defined.
+
 Fixpoint trunc_index_leq (m n : trunc_index) : Type0
   := match m, n with
        | -2, _ => Unit
@@ -28,6 +45,100 @@ Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
 
 Fixpoint trunc_index_inc (n : nat) (k : trunc_index) : trunc_index
   := match n with O => k | S m => (trunc_index_inc m k).+1 end.
+
+Definition trunc_index_pred : trunc_index -> trunc_index.
+Proof.
+  intros [|m].
+  1: exact -2.
+  exact m.
+Defined.
+
+(* TODO: Add to Notations.v *)
+Notation "n '.-1'" := (trunc_index_pred n) (at level 10) : trunc_scope.
+Notation "n '.-2'" := (n.-1.-1) (at level 10) : trunc_scope.
+
+Definition trunc_index_leq_minus_two {n} : n <= -2 -> n = -2.
+Proof.
+  destruct n.
+  1: reflexivity.
+  contradiction.
+Defined.
+
+Fixpoint trunc_index_leq_succ n m : n <= m -> n <= m.+1.
+Proof.
+  destruct n.
+  { apply idmap. }
+  intro p.
+  destruct m.
+  { destruct p. }
+  exact (trunc_index_leq_succ n m p).
+Defined.
+
+Fixpoint trunc_index_leq_refl' a : a <= a.
+Proof.
+  destruct a.
+  1: exact tt.
+  exact (trunc_index_leq_refl' a).
+Defined.
+
+Global Instance trunc_index_leq_refl : Reflexive trunc_index_leq
+  := trunc_index_leq_refl'.
+
+Fixpoint trunc_index_leq_concat {a b c} {struct c} : a <= b -> b <= c -> a <= c.
+Proof.
+  intros p q.
+  destruct a, b, c.
+  all: try (exact tt || assumption).
+  { revert p; intros []. }
+  exact (@trunc_index_leq_concat a b c p q).
+Defined.
+
+Global Instance trunc_index_leq_transitive : Transitive trunc_index_leq.
+Proof.
+  intros a b c p q.
+  apply (trunc_index_leq_concat p q).
+Defined.
+
+Fixpoint trunc_index_min (n m : trunc_index) : trunc_index.
+Proof.
+  destruct n.
+  1: exact -2.
+  destruct m.
+  1: exact -2.
+  exact (trunc_index_min n m).+1.
+Defined.
+
+Definition trunc_index_min_minus_two n : trunc_index_min n -2 = -2.
+Proof.
+  destruct n; reflexivity.
+Defined.
+
+Definition trunc_index_min_leq_left (n m : trunc_index)
+  : trunc_index_min n m <= n.
+Proof.
+  destruct n.
+  1: reflexivity.
+Admitted.
+
+Definition trunc_index_min_leq_right (n m : trunc_index)
+  : trunc_index_min n m <= m.
+Proof.
+Admitted.
+
+Definition trunc_index_min_swap n m
+  : trunc_index_min n m = trunc_index_min m n.
+Proof.
+  revert m.
+  induction n.
+  { intro.
+    symmetry.
+    apply trunc_index_min_minus_two. }
+  induction m.
+  1: reflexivity.
+  cbn.
+  apply ap.
+  apply IHn.
+Defined.
 
 (** ** Truncatedness proper. *)
 
@@ -50,6 +161,15 @@ Proof.
   induction n.
   + exact -2.
   + exact IHn.+1.
+Defined.
+
+Lemma nat_to_trunc_index_2_eq n
+  : nat_to_trunc_index_2 n.+2 = nat_to_trunc_index n.
+Proof.
+  induction n.
+  1: reflexivity.
+  cbn; apply ap.
+  assumption.
 Defined.
 
 (** This could be an [Instance] (with very high priority, so it doesn't get applied trivially).  However, we haven't given typeclass search any hints allowing it to solve goals like [m <= n], so it would only ever be used trivially.  *)
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -41,7 +41,11 @@ Fixpoint trunc_index_leq (m n : trunc_index) : Type0
        | m'.+1, n'.+1 => trunc_index_leq m' n'
      end.
 
+Existing Class trunc_index_leq.
+
 Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
+
+Global Instance minus_two_leq_minus_two : -2 <= -2 := tt.
 
 Fixpoint trunc_index_inc (n : nat) (k : trunc_index) : trunc_index
   := match n with O => k | S m => (trunc_index_inc m k).+1 end.
@@ -53,9 +57,8 @@ Proof.
   exact m.
 Defined.
 
-(* TODO: Add to Notations.v *)
-Notation "n '.-1'" := (trunc_index_pred n) (at level 10) : trunc_scope.
-Notation "n '.-2'" := (n.-1.-1) (at level 10) : trunc_scope.
+Notation "n '.-1'" := (trunc_index_pred n) : trunc_scope.
+Notation "n '.-2'" := (n.-1.-1) : trunc_scope.
 
 Definition trunc_index_leq_minus_two {n} : n <= -2 -> n = -2.
 Proof.
@@ -76,13 +79,13 @@ Defined.
 
 Fixpoint trunc_index_leq_refl' a : a <= a.
 Proof.
-  destruct a.
-  1: exact tt.
-  exact (trunc_index_leq_refl' a).
+  by (destruct a; cbn).
 Defined.
 
-Global Instance trunc_index_leq_refl : Reflexive trunc_index_leq
-  := trunc_index_leq_refl'.
+Global Instance trunc_index_leq_refl : Reflexive trunc_index_leq.
+Proof.
+  intro; apply trunc_index_leq_refl'.
+Defined.
 
 Fixpoint trunc_index_leq_concat {a b c} {struct c} : a <= b -> b <= c -> a <= c.
 Proof.
@@ -113,18 +116,6 @@ Proof.
   destruct n; reflexivity.
 Defined.
 
-Definition trunc_index_min_leq_left (n m : trunc_index)
-  : trunc_index_min n m <= n.
-Proof.
-  destruct n.
-  1: reflexivity.
-Admitted.
-
-Definition trunc_index_min_leq_right (n m : trunc_index)
-  : trunc_index_min n m <= m.
-Proof.
-Admitted.
-
 Definition trunc_index_min_swap n m
   : trunc_index_min n m = trunc_index_min m n.
 Proof.
@@ -138,6 +129,40 @@ Proof.
   cbn.
   apply ap.
   apply IHn.
+Defined.
+
+Definition trunc_index_min_path n m
+  : (trunc_index_min n m = n) + (trunc_index_min n m = m).
+Proof.
+  revert m.
+  induction n.
+  1: by intro; apply inl.
+  induction m.
+  1: by apply inr.
+  destruct (IHn m).
+  1: apply inl.
+  2: apply inr.
+  all: cbn; by apply ap.
+Defined.
+
+Definition trunc_index_min_leq_left (n m : trunc_index)
+  : trunc_index_min n m <= n.
+Proof.
+  revert n m.
+  refine (trunc_index_ind _ _ _); [ | intros n IHn ].
+  all: refine (trunc_index_ind _ _ _); [ | intros m IHm ].
+  all: try exact tt.
+  exact (IHn m).
+Defined.
+
+Definition trunc_index_min_leq_right (n m : trunc_index)
+  : trunc_index_min n m <= m.
+Proof.
+  revert n m.
+  refine (trunc_index_ind _ _ _); [ | intros n IHn ].
+  all: refine (trunc_index_ind _ _ _); [ | intros m IHm ].
+  all: try exact tt.
+  exact (IHn m).
 Defined.
 
 (** ** Truncatedness proper. *)
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -45,10 +45,18 @@ Existing Class trunc_index_leq.
 
 Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
 
-Global Instance minus_two_leq_minus_two : -2 <= -2 := tt.
+Global Instance trunc_index_leq_minus_two_n n : -2 <= n := tt.
+
+Global Instance trunc_index_leq_succ n : n <= n.+1.
+Proof.
+  by induction n as [|n IHn] using trunc_index_ind.
+Defined.
 
 Fixpoint trunc_index_inc (n : nat) (k : trunc_index) : trunc_index
-  := match n with O => k | S m => (trunc_index_inc m k).+1 end.
+  := match n with
+      | O => k
+      | S m => (trunc_index_inc m k).+1
+    end.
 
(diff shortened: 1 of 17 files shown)


mikeshulman:
More later, but just quickly:

> It is worth noting that I (and probably many others) thought that K(G,n+1) being the delooping of K(G,n) was a simple application of Freudenthal. This is not the case

But it is simple when n>1, right?  The hard part is only the case n=1?

Alizter:
@mikeshulman Yep this is correct, but 90% of the proof was spent trying to understand the n=1 case.

Alizter:
I should note that it is trivial to define Eilenberg-MacLane spectra for an abelian group G now. I would like to understand spectrification better in the future so we can define suspension spectra.

mikeshulman:
Even in the case n=1, Freudenthal gives you something: the map from K(G,1) to loops on K(G,2) is 0-connected, i.e. an isomorphism on pi0 and a surjection on pi1.  But I suppose probably knowing that doesn't make the rest of the proof any easier...

Alizter:
Another interesting thing is that the definition of K(G,n) doesn't require the group to be abelian. Only if you want to show the loop space is equivalent to the one below. I don't know if it makes sense to ask what the loop space of a K(G, n) is for a nonabelian G.

Like what is the loop space of `K(S_3,2)`? (symmetric group). Would it be the infinite dimensional real projective space? Since the abelianisation of S_3 would be Z/2Z. I am hesitant to really call these Eilenberg-MacLane spaces since they don't satisfy the "universal property".

mikeshulman:
The 2-truncation of the suspension of K(G,1), for G nonabelian, should be K(G_ab,2), where G_ab is the abelianization of G.  Hence your "K(G,n)" for G nonabelian and n>1 should reduce to the usual K(G_ab,n).  The application of Freudenthal to the case n=1 that I mentioned above doesn't depend on G being abelian, and the surjectivity on pi1 corresponds to G -> G_ab being surjective.  Injectivity is where you need G abelian.  It would be nice if we could show more generally that, say, if X is 0-connected then the map X -> loop susp X is abelianization on pi1; that would imply the result you need in the case when G is abelian.  Any chance your encode-decode would generalize to this case?

Alizter:
@mikeshulman I don't know if the encode-decode would generalise, but I have an idea how to go about proving it. I am bit tired, but I will sketch some thoughts: The James construction ought to say something about pi1, so I reckon we only need to analyse J1 or J2 and check that its homotopy groups abelianise. I will check this in more detail tomorrow.

Alizter:
I’m a bit out of my depth here, but I notice that you are asking if Pi1 takes the unit of the loop susp adjunction to the unit of the forget abelianize adjunction.

Is there some property of pi that we can use to show this?

mikeshulman:
Yes, in fact I now think that result should follow fairly directly from the simpler one about it being an equivalence when G is abelian.  Because then we should be able to show that the category of pointed (n-1)-connected n-types is equivalent to the category of abelian groups when n>1 and to the category of groups when n=1, with the loops functor from n=2 to n=1 corresponding to the inclusion and therefore its left adjoint susp corresponding to the left adjoint to that inclusion, i.e. abelianization.

theories/Basics/Trunc.v
```diff
@@ -29,6 +46,100 @@ Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
...
   := match n with O => k | S m => (trunc_index_inc m k).+1 end.
 
+Definition trunc_index_pred : trunc_index -> trunc_index.
+Proof.
+  intros [|m].
+  1: exact -2.
+  exact m.
+Defined.
+
+(* TODO: Add to Notations.v *)
```
mikeshulman: Maybe do this?

theories/Basics/Trunc.v
```diff
@@ -29,6 +46,100 @@ Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
...
+Proof.
+  destruct a.
+  1: exact tt.
+  exact (trunc_index_leq_refl' a).
+Defined.
+
+Global Instance trunc_index_leq_refl : Reflexive trunc_index_leq
+  := trunc_index_leq_refl'.
+
+Fixpoint trunc_index_leq_concat {a b c} {struct c} : a <= b -> b <= c -> a <= c.
```
mikeshulman: I would personally find it simpler to prove these with an `induction` inside the `Instance`, rather than a separate `Fixpoint`.  But if you do it the latter way, I would suggest calling this one `transitive'` analogously to your `refl'` above.

theories/HIT/Connectedness.v
```diff
@@ -103,6 +103,16 @@ Proof.
   - apply IHn, isconnected_pred; assumption.
 Defined.
 
+Definition isconnected_pred_trunc_index_add n m A `{H : IsConnected (n +2+ m) A}
```
mikeshulman: I feel like there must be a better name for this.

theories/Basics/Trunc.v
Alizter: Makes sense.

theories/HIT/Connectedness.v
Alizter: I would prefer making this an instance and giving it no name, but this breaks the typeclass search. I feel like there is an elegant solution to this, where coq can automatically infer that n+2+m is bigger than m.

theories/HIT/Truncations.v
```diff
@@ -293,4 +293,77 @@ Proof.
     strip_truncations. destruct c. reflexivity.
 Defined.
 
+Definition trunc_index_min_path n m
```
mikeshulman: Shouldn't this go in `Basics.Trunc`?

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,373 @@
+Require Import Basics.
+Require Import Types.
+Require Import Pointed.
+Require Import Cubical.
+Require Import abstract_algebra.
```
mikeshulman: I think it would be helpful to give a more complete path name here.  I started looking up the definition of `Group`, but it took some work to figure out that the file `abstract_algebra` is in `Classes/interfaces`.

theories/HIT/Truncations.v
Alizter: Unfortunately not, becuase we need functor_sum whcih is in Types. Perhaps there is a way though.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,373 @@
...
+End ClassifyingSpace.
+
+Global Instance ispointed_BG `{Group G} : IsPointed (ClassifyingSpace G).
+Proof.
+  exact (bbase G).
+Defined.
+
+Definition B (G : Type) `{Group G} := Build_pType (ClassifyingSpace G) _.
+
+(* TODO: Move to "Group" file? *)
```
mikeshulman: Meaning `Classes.theory.groups`?  Or somewhere else?  I wonder whether these would fit more naturally in a file about H-spaces.

Alizter: I am thinking somewhere else. I already think that mathclasses is a bit "all over the place" when it comes to organisation. So I don't wish to add to the pile, especially if it wasn't there to begin with.

I wish we could define all our classes as truncated H-stuff, but I am not sure if this is useful. So I think it is worth "duplicating" such results here.

theories/Homotopy/ClassifyingSpace.v
Alizter: I would argue that in the future, we ought to have something like "Require Import Group.". But for now I will make it clearer where exactly abstract_algebra is. Even though it is one of the most undescriptive names for its contents.

theories/HIT/Connectedness.v
mikeshulman: I don't think Coq is going to be able to find this automatically in general because it requires guessing the number `m`.  But it might work sometimes to make it a `Hint Immediate`?

But in any case, even an instance needs a name.

theories/HIT/Truncations.v
mikeshulman: The definition of `functor_sum` is one line.  Just do a manual destruct.

theories/Homotopy/ClassifyingSpace.v
mikeshulman: The problem with having it here is that people won't think of looking here for it, so they may end up re-proving it later on.

mikeshulman:
Overall this looks good.  I wonder how much more work it would be to construct classifying spaces (i.e. Rezk completions) of pregroupoids, and derive classifying spaces of groups as a special case.

theories/HIT/Truncations.v
Alizter: OK I will look into this.

Alizter:
@mikeshulman I think we run into the inheritance problem here: "What should a group be a special case of?". If it is a one object pregroupoid, then sure this approach would be nice. But since there are few other objects a group is a special case of, it seems painful to make an equivalence between one object pregroupoids, take the Rezk completion, the come back to whatever definition of group we are using.

mathclasses does a pretty good job of inheriting algebraic structure. But in mathematics we still make up new inheritances without really thinking about them as different objects.

Groups can be:
 - One object pregroupoids
 - Monoids with inverses
 - Loops with associativity
 - Truncated H-groups

I feel like these should all be equivalent by definition somehow. If this is the case then taking the Rezk completion would be fine. But currently we are forced to choose one, and hope not to have to use the others.

mikeshulman:
I don't think it would be very painful in this case.

FWIW, pregroupoids, monoids, and H-groups are all special cases of "wild categories" (while "loops" in the algebraic sense are kind of exotic and, I expect, unlikely to matter much for us).  So if there's some construction that makes sense in that generality, its specialization to groups could be almost definitionally the same through all of those paths.

theories/Basics/Trunc.v
Alizter: Ah, the reason I didn't use induction is because I couldn't. For some reason coq has trouble doing it, so I just created the hypothesis manually with fixpoint.

theories/HIT/Connectedness.v
Alizter: I don't know how to give hints to typeclass search (yet). Do you have any idea for a better name?

Alizter:
The reason I mention loopy structures is that S7 has a moufang loop structure. So I don't want to shove such structures in a corner and call them exotic if we don't have to.

Since mathclasses is very set based, I think it is reasonable to want to expand it a bit. Perhaps we can reorganise its contents and have a different structure to all the algebra. Say perhaps, deriving from wild categories as you say. Jason's category library uses records if I recall correctly, so it shouldn't be an issue to turn them into Classes.

I think the UniMath library does something similar to this, where they define "precategory" to mean what we call a wild category. I don't think any of their algebra stuff derives from this though.

Another possible advantage I can see is that if in the future, someone works out how to get infinite coherences going we can define oo-cats as special wild categories. This will also mean we can define things like An/Aoo and En/Eoo spaces this way too.

theories/Basics/Trunc.v
mikeshulman: Whoa, that's an error I've never seen before:

```
Warning: Cannot remove a. [cannot-remove-as-expected,tactics]
```

Can anyone offer some insight?  (@JasonGross, @mattam82 ?)

But applying the induction principle manually works:

```
Global Instance trunc_index_leq_transitive : Transitive trunc_index_leq.
Proof.
  refine (trunc_index_ind _ _ _); [ | intros a IHa].
  all:refine (trunc_index_ind _ _ _); [ | intros b IHb].
  all:refine (trunc_index_ind _ _ _); [ | intros c IHc]; intros p q.
  all: try (exact tt || assumption).
  - destruct p.
  - exact (IHa b c p q).
Defined.
```

theories/HIT/Connectedness.v
mikeshulman: I want to say something like `isconnected_leq`, but I suppose that would be naturally interpreted as a reference to `trunc_index_leq` instead.  Maybe `isconnected_subt` or `isconnected_minus`?

mikeshulman:
I suppose one could talk about "categories" that are so wild they aren't even associative ("magmoids"?).  But that seems less likely to be useful.  I'm not saying loops aren't good for anything, but it seems unlikely to me that any concrete *construction* we care about would happen at a level of generality including them and also want to be specialized to categories, monoids, etc.

theories/HIT/Connectedness.v
Alizter: I don't think the name is currently bad, just too long to be of practical use. Perhaps we should just leave a TODO here in case we have better ideas in the future.

I am warming up to isconnected_minus however, I will think about it.

Alizter:
I suppose one could take this argument to the extreme and say all our algebraic structures should come from a universal algebra framework (not just for sets however). But I am not sure how useful this will be or if we can even do it effectively. This is kind of like the discussion with colimits and choosing a fundamental one.

theories/HIT/Connectedness.v
mikeshulman: I do think it is bad; I wouldn't be able to guess or remember it.

theories/Basics/Trunc.v
```diff
@@ -29,6 +46,100 @@ Notation "m <= n" := (trunc_index_leq m n) : trunc_scope.
...
+  1: exact -2.
+  exact (trunc_index_min n m).+1.
+Defined.
+
+Definition trunc_index_min_minus_two n : trunc_index_min n -2 = -2.
+Proof.
+  destruct n; reflexivity.
+Defined.
+
+Definition trunc_index_min_leq_left (n m : trunc_index)
```
Alizter: This is very sloppy on my part. I have left these as admitted.

theories/Basics/Trunc.v
Alizter: I can't make the Reflexive a fixpoint, since I cannot specify the decreasing argument. This is why there are two definitions. If induction worked all would be fine.

theories/Basics/Trunc.v
mikeshulman: Oops, I should have caught that too.  Are you going to fix it I hope?

theories/Basics/Trunc.v
mikeshulman: I don't understand what you are saying.  Currently both of them are fixpoints.  I expect that reflexive could be made into a single proof just like I did above for transitive, applying the induction principle explicitly rather than relying on the `induction` tactic.

theories/Basics/Trunc.v
Alizter: Yes.

Alizter: They are a bit tricky to prove however...

theories/Basics/Trunc.v
SkySkimmer: You could use `fix` in the term instead of the command `Fixpoint`

theories/Basics/Trunc.v
mikeshulman:
```
  revert n m.
  refine (trunc_index_ind _ _ _); [ | intros n IHn ].
  all:refine (trunc_index_ind _ _ _); [ | intros m IHm ].
  all:try exact tt.
  exact (IHn m).
```

Alizter: Nice. I have added those in now.

theories/Basics/Trunc.v
Alizter: @SkySkimmer That makes sense.

Alizter:
So I have started an Algebra/Group file to start collecting results about groups. The current proofs of the left and right multiplication equivalences use the group structure. There is a proof that doesn't use this and infact can be found in HSpace.v.

In my mind if we ever write down Lagrange's theorem, this would be the suitable home.

mikeshulman:
> The current proofs of the left and right multiplication equivalences use the group structure. There is a proof that doesn't use this and in fact can be found in HSpace.v.

The latter is a proof of a different statement, that a *connected* H-space has these properties.

theories/Basics/Trunc.v
mikeshulman: Are you planning to make these two definitions parallel at least?

JasonGross: Do you have a standalone example of the induction warning message?  I've never seen that one before, and am intrigued.

JasonGross: This is because `induction` when the goal is in `Set` (which is what `Type0` is) uses `trunc_index_rec`, and `trunc_index_rec` is a `Minimality` scheme, not an `Induction` scheme.  I shall report a bug in Coq

JasonGross: https://github.com/coq/coq/issues/10766

JasonGross: Note that you _can_ do `induction a as [|a IHa] using trunc_index_ind.`

Alizter:
@mikeshulman Yes, of course. I seem to have been confused about this.

theories/HIT/Connectedness.v
Alizter: I've changed it to `isconnected_pred_add` since it is a version of `isconencted_pred` but with an `add`.

Alizter:
I want to also check that everything is living in the right universe. If `G` is in some universe `U` then `K(G,n)` ought to be in `U` too, for all `n` right?

I notice this might not be the case since I am using `Tr` instead of `Trunc`.

mikeshulman:
Yes, `K(G,n)` ought to be in the same universe as `G`.  What makes you think that `Tr` changes universe levels?  `Tr` and `Trunc` are definitionally equal.

Alizter:
@mikeshulman Maybe I got confused since `Tr` gets put in `Type2le`. I thought this was part of the problem when trying to debug V.v in #1088

mikeshulman:
I don't have any idea what the problem with #1088 is, but in this case I think you should stick with `Tr` and just check that the universe level doesn't go up (I'd be very surprised if it did).

Alizter:
@mikeshulman You are right, the level doesn't go up. The problem in #1088 has really confused me. But nevertheless we see that `K(G,n)` is living in the right place.

Alizter:
rebased

Alizter:
I will keep this PR open for now until I can get round #1088.

Alizter:
I've got some work to do here. I will create a new PR when I am done.

Issue #1067, opened 2019-08-03
Reorganising HITs


Alizter:
So I know there was a discussion a while back #573, about moving some files out of HITs to other places. The way I see it, there are only a few "real" HITs that have to be specified with axioms. These should stay in HITs, the rest are defined in terms of these so they should be allowed to dwell anywhere, as some already do.

I propose moving files such as Suspension and Pushout into the Homotopy folder. I also propose moving files such as Fibrations and Pullbacks into this folder. Files such as coequalizers however should stay in HITs.

Files such as Spheres and Circle should be moved into the Spaces folder.

I am already splitting up Pointed into multiple files and giving it its own folder. This would be a natural place for future files such as Wedge and Smash to live in. I want to develop constructions with pointed types properly.

I am however unsure about moving Quotients and friends into a Set folder. Though it could be argued this is the correct thing to do. I don't want to spend time doing these things if its not what people want.

What do you think about this proposal?

spitters:
That seems reasonable. Although I don't have a strong opinion on it.

There are a number of people building on the library, but not contributing to it. E.g.
https://github.com/HoTT/HoTT/wiki/Publications-based-on-the-HoTT-library
( And others )

These kind of big changes might break their work. Arguably, that's their responsibility.

Alizter:
@spitters Thanks for letting me know about those. I think creating a list of changes will help people when they wish to take into account these changes without them having to guess.

mikeshulman:
That sort of thing seems like a reasonable start.  I'm not sure about your proposed move of files into the Homotopy folder, though.  I think of the Homotopy folder as being about actual synthetic homotopy theory, whereas the files you mention are really just defining much more basic constructions on types.  Maybe they should be in a folder about "limits and colimits" or something.

Alizter:
@mikeshulman I think I will move results about homotopy groups into folders containing the spaces in question. Spheres and circles will be moved to their own folders in spaces containing various results about them.

The Homotopy folder will contain general things like the hopf fibrations etc.

I am currently trying to finish off the 3x3 lemma, after which I have the real and complex hopf fibrations ready to go. (They are currently waiting for a proof of the associativity of join (which needs 3x3 or patience)). I have a semifinished version of the Quarternionic hopf fibration which shouldn't be too difficult to clean up.

After which I will try and spend some time proving the smash-pointed map adjunction so we can get some coherence results. This will lead to the development of homology and the Gysin sequence, eventually leading to Pi4S3.

[...]

PR #1087, opened 2019-09-10
colimit cleanup

Alizter:
I have seperated up the files in the colimits library a bit more. Notably we now have a Diagrams and Colimits folder which will make room for a limits library in the future.

The Colimits library will also be the natural home for some of the HITs we have, as discussed in #1067.

I welcome all comments, suggestions and criticisms.

I am also considering defining colimits as a coequalizer rather than a private inductive type like we currently have. This will be akin to how I changed Suspension to use pushouts #986. This will allow us to keep very few private inductive types hidden away in HIT, and keep perhaps put other files, such as Pushout.v in Colimits. This also explains the new naming convention I have given for pushouts defined as colimits.

By limits library I mean defining the limit of a diagram as in [Homotopy limits in Coq](https://arxiv.org/abs/1304.0680v1). This will eventually let us prove the two pullbacks lemma, via universal properties. As discussed in #1012.


Alizter:
@mikeshulman @spitters @andreaslyn anyone able to review?

theories/Diagrams/#Grothendieck.v#
```diff
@@ -0,0 +1,20 @@
+Require Import Basics.
```
mikeshulman: Usually files with `#Names#` are Emacs backup files and shouldn't be committed.  Is that the case here?

Alizter: Hmm, I must have accidentally commited this. I will get rid of it.

Alizter: This was a half finished version of DDiagram

theories/Diagrams/MappingTelescope.v
```diff
@@ -0,0 +1,51 @@
...
+Require Import Types.
+Require Import Diagrams.Graph.
+Require Import Diagrams.Diagram.
+
+Local Open Scope nat_scope.
+Local Open Scope path_scope.
+
+(** * Mapping telescope *)
+
+(** A mapping telescope, or a sequence, is an infinite sequence of types [X(n)] with an arrow from [X(n)] to [X(n+1)]. *)
```
mikeshulman: I guess this terminology was inherited from the old colimits library, but it is wrong and should be fixed.  A mapping telescope is the *colimit* of a diagram of this sort; the diagram itself is just called a sequence, or an N-sequence or omega-sequence to be precise.

theories/Diagrams/CoeqDiagram.v
```diff
@@ -0,0 +1,23 @@
+Require Import Basics.
+Require Import Types.
+Require Import Diagrams.Graph.
+Require Import Diagrams.Diagram.
+
+(** The shape of a coequalizer diagram. *)
```
mikeshulman: If you're separating out diagrams for use in defining limits as well as colimits, then it would be better to call this a "parallel pair" than "the shape of a coequalizer", since it is also the shape of an equalizer.

theories/Diagrams/Cocone.v
```diff
@@ -0,0 +1,256 @@
+Require Import Basics.
+Require Import Types.
+Require Import Diagrams.Graph.
+Require Import Diagrams.Diagram.
+
+Local Open Scope path_scope.
+Generalizable All Variables.
+
+(** * Cocones *)
```
mikeshulman: It's unclear to me whether the theory of cocones belongs in `Diagrams` or in `Colimits`...

theories/Colimits/Colimit_Pushout_Flattening.v
```diff
@@ -0,0 +1,80 @@
...
+Require Import Diagrams.DDiagram.
+Require Import Diagrams.Span.
+Require Import Diagrams.Cocone.
+Require Import Colimits.Colimit.
+Require Import Colimits.Colimit_Pushout.
+Require Import Colimits.Colimit_Flattening.
+
+(** * Pushout case *)
+
+(** We instanciate the flattening lemma in the case of the pushout. *)
```
mikeshulman: Might as well fix the spelling here.

theories/Diagrams/MappingTelescope.v
Alizter: I think sequence should be the correct choice.

theories/Colimits/Colimit_Pushout.v
```diff
@@ -0,0 +1,208 @@
...
+Require Import Colimits.Colimit.
+
+(* We require this now, but will import later. *)
+Require HIT.Pushout.
+
+Local Open Scope path_scope.
+
+(** * Pushout as a colimit *)
+
+(** In this file, we define [PO] the pushout of two maps as the colimit of a particuliar diagram, and then show that it is equivalent to [pushout] the primitive pushout defined as an HIT. *)
```
mikeshulman: Might as well fix the spelling here (`particuliar`).

theories/Diagrams/CoeqDiagram.v
Alizter: Makes sense. I am currently writing the limits library (using the limits in type theory paper). I might leave this change till then.

theories/Diagrams/Cocone.v
Alizter: I have kept cocones in diagrams since I think of them as "a map between a diagram and a type".

mikeshulman:
This looks pretty good, thanks.

I'm generally unsatisfied with the situation of having two kinds of coequalizer and two kinds of pushout.  Defining colimits in terms of coequalizers would reduce the number of axioms, but wouldn't really address this issue.  I wonder whether it would be better to keep general colimits as a private type and instead define coequalizers and pushouts in terms of them?  Then there would only be one kind of "coequalizer" in the library.

theories/Diagrams/MappingTelescope.v
Alizter: The nlab is a bit confusing with regard to this terminology. If we call the diagram giving a mapping telescope a sequence, what do we call the diagram giving an inverse limit?

mikeshulman: Which nlab page are you looking at?

If we call it an N-sequence or omega-sequence, then the dual is an N^op-sequence or omega^op-sequence.  But sometimes "tower" is also used.

Alizter: if we can call such a diagram a tower, maybe we can call an "upside-down tower" a telescope.

The nlab page [https://ncatlab.org/nlab/show/mapping+cocone](https://ncatlab.org/nlab/show/mapping+cocone) doesn't have a limit equivalent, or a discussion of the diagram involved.

mikeshulman: https://ncatlab.org/nlab/show/mapping+telescope seems the appropriate page.  The name "telescope" is totally standard in algebraic topology for the *colimit*, not the diagram.  The diagram is an N-sequence or omega-sequence.

Once or twice I've seen the phrase "mapping microscope" for the limit of a tower, but I suspect it was more than half in jest.

Alizter:
@mikeshulman Basing everything off of colimit sounds interesting. I will give it a try later.

In my mind, coequalizers seem like they should be more fundamental since they are simpler. If we define all our other colimits in terms of colimit, I think there might be a performance issue. But I don't know since I haven't tried it yet.

theories/Diagrams/MappingTelescope.v
Alizter: Yes I meant that page, guess I didn't copy the right link. I am in favour of sequence and tower.

Alizter:
@mikeshulman For now, can we have two kinds of coequalizer and pushout? Since this isn't anything new. I don't really want to add new stuff in this PR since it's just supposed to be clean up. I will create an issue about it though.

theories/Colimits/Colimit_Pushout_Flattening.v
Alizter: @mikeshulman `instanciate` seems to be a word, but I have changed it to `show` since `instantiate` would mean there are type classes around. Weird Latin spelling.

theories/Diagrams/Cocone.v
Alizter: And since colimits are a special cocone I don't see why cocones should be in the colimits bit.

theories/Colimits/Colimit_Pushout_Flattening.v
mikeshulman: Whaddaya know, you learn something new every day.

mikeshulman:
Yeah sure.

Alizter:
@spitters @andrejbauer @andreaslyn Anyone else able to review?

theories/Colimits/Colimit_Coequalizer.v
```diff
@@ -0,0 +1,88 @@
...
+
+  (** ** Equivalence with [Coeq] *)
+
+  Context {f g : B -> A}.
+
+  Definition cocone_Coeq : Cocone (parallel_pair f g) (Coeq f g).
+  Proof.
+    serapply Build_Cocone.
+    + intros []; [exact (coeq o g)| exact coeq].
+    + intros i j phi x; destruct i, j, phi; simpl;
```
spitters: It's good style to use intro patterns, instead of intros ... destruct.

Alizter: This is the original code written for the colimits library which I didn't want to touch too much.

Alizter: I agree using intro patterns (sensibly) is easier to maintain.

spitters:
It's a big commit and I don't have time for a careful review anytime soon.
Generally, the commit looks nice and clean. I'm OK with pulling this one.

Alizter:
@mikeshulman Do you want to pull this now or wait a bit more?

PR #1089, opened 2019-09-17
Remove some unnecessary funext and univalence, esp. in Pointed

mikeshulman:
While reviewing #1086 I noticed that a fair number of funext and univalence assumptions in the Pointed library could be eliminated -- some completely, others by rephrasing theorems in terms of pointed equivalences rather than equalities of pointed types (which are generally more useful to have defined directly anyway -- it produces fewer univalence redexes).

Whichever of this or #1086 is merged second will probably require rebasing onto the other.

### DIFF
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -93,7 +93,7 @@ Coercion pfam_pr1 : pFam >-> Funclass.
 
 (* IsTrunc for a pointed type family *)
 Class IsTrunc_pFam n {A} (X : pFam A)
-  := Trunc_pFam_is_trunc : forall x, IsTrunc n (X.1 x).
+  := trunc_pfam_is_trunc : forall x, IsTrunc n (X.1 x).
 
 (* Pointed sigma *)
 Definition psigma {A : pType} (P : pFam A) : pType.
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -88,24 +88,23 @@ Coercion pointed_equiv_equiv {A B} (f : A <~>* B)
 
 (* Pointed type family *)
 Definition pFam (A : pType) := {P : A -> Type & P (point A)}.
+Definition pfam_pr1 {A : pType} (P : pFam A) : A -> Type := pr1 P.
+Coercion pfam_pr1 : pFam >-> Funclass.
 
 (* IsTrunc for a pointed type family *)
-Definition IsTrunc_pFam n {A} (X : pFam A) := forall x, IsTrunc n (X.1 x).
+Class IsTrunc_pFam n {A} (X : pFam A)
+  := Trunc_pFam_is_trunc : forall x, IsTrunc n (X.1 x).
 
 (* Pointed sigma *)
-Definition psigma : {A : pType & pFam A} -> pType.
+Definition psigma {A : pType} (P : pFam A) : pType.
 Proof.
-  intros [[A a] [P p]].
   exists {x : A & P x}.
-  exact (a; p).
+  exact (point A; P.2).
 Defined.
 
 (* Pointed pi types, note that the domain is not pointed *)
-Definition pforall : {A : Type & A -> pType} -> pType.
-Proof.
-  intros [A F].
-  exact (Build_pType (forall (a : A), pointed_type (F a)) (ispointed_type o F)).
-Defined.
+Definition pforall {A : Type} (F : A -> pType) : pType
+  := Build_pType (forall (a : A), pointed_type (F a)) (ispointed_type o F).
 
 (** The following tactic often allows us to "pretend" that pointed maps and homotopies preserve basepoints strictly.  We have carefully defined [pMap] and [pHomotopy] so that when destructed, their second components are paths with right endpoints free, to which we can apply Paulin-Morhing path-induction. *)
 Ltac pointed_reduce :=
--- a/theories/Pointed/Loops.v
+++ b/theories/Pointed/Loops.v
@@ -208,7 +208,7 @@ Proof.
   reflexivity.
 Defined.
 
-(* Iterated loops products preserve products *)
+(* Iterated loops of products are products of iterated loops *)
 Lemma iterated_loops_prod (X Y : pType) {n}
   : iterated_loops n (X * Y) <~>* (iterated_loops n X) * (iterated_loops n Y).
 Proof.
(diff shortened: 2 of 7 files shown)


theories/Pointed/Core.v
```diff
@@ -88,24 +88,23 @@ Coercion pointed_equiv_equiv {A B} (f : A <~>* B)
 
 (* Pointed type family *)
 Definition pFam (A : pType) := {P : A -> Type & P (point A)}.
+Definition pfam_pr1 {A : pType} (P : pFam A) : A -> Type := pr1 P.
+Coercion pfam_pr1 : pFam >-> Funclass.
 
 (* IsTrunc for a pointed type family *)
-Definition IsTrunc_pFam n {A} (X : pFam A) := forall x, IsTrunc n (X.1 x).
+Class IsTrunc_pFam n {A} (X : pFam A)
+  := Trunc_pFam_is_trunc : forall x, IsTrunc n (X.1 x).
```
Alizter: Maybe `Trunc` should be `trunc`?

theories/Pointed/Loops.v
```diff
@@ -204,7 +209,7 @@ Proof.
 Defined.
 
 (* Iterated loops products preserve products *)
```
Alizter: I must have written this when tired. Maybe
```
Iterated loops of products are products of iterated loops
```
is a better description.

theories/Pointed/Loops.v
```diff
@@ -214,107 +219,86 @@ Proof.
...
 (* psigma and loops 'commute' *)
-Lemma loops_psigma_commute `{Univalence}
-  : loops o psigma == psigma o pfam_loops.
+Lemma loops_psigma_commute (A : pType) (P : pFam A)
+  : loops (psigma P) <~>* psigma (loopsD P).
 Proof.
-  intros x.
-  apply path_ptype.
   srefine (Build_pEquiv _ _ (Build_pMap _ _ (_ : Equiv _ _) _) _).
   1: exact (equiv_path_sigma _ _ _)^-1%equiv.
```
Alizter: The `%equiv` is an artifact from when pEquiv and Equiv used the same inverse notation.

theories/Pointed/Loops.v
```diff
@@ -214,107 +219,86 @@ Proof.
...
-  refine (IHn _ _ _ @ (unfold_iterated_loops _ _)^).
-  intro; refine (@istrunc_paths _ _ (p (point A)) _ _).
+  refine (pequiv_inverse (unfold_iterated_loops' _ _) o*E _ o*E unfold_iterated_loops' _ _).
+  refine (IHn _ _ _ o*E _).
+  apply pequiv_iterated_loops_functor.
+  apply loops_psigma_commute.
 Defined.
 
 (* We declare this local notation to make it easier to write pointed types *)
 Local Notation "( X , x )" := (Build_pType X x).
```
Alizter: I wonder if we can make this a global notation. Would `( X ,, x )` be a good one, to stop from clashing with pair?

theories/Pointed/Loops.v
```diff
@@ -214,107 +219,86 @@ Proof.
...
-    {f : A -> A & IsEquiv f} (idmap; isequiv_idmap A).
-Proof.
-  symmetry.
-  apply path_ptype.
-  refine (Build_pEquiv _ _ (Build_pMap
-    ({f : A -> A & IsEquiv f}, (idmap; isequiv_idmap A))
-    (A <~> A, 1%equiv) (issig_equiv _ _) 1) _).
-Defined.
-
 (* We can convert between families of loops in a type and
```
Alizter: Line break can be removed

theories/Pointed/pEquiv.v
```diff
@@ -75,6 +75,11 @@ Defined.
 Definition path_ptype `{Univalence} {A B : pType} : (A <~>* B) -> A = B
   := equiv_path_ptype A B.
 
+Definition pequiv_path {A B : pType} : (A = B) -> (A <~>* B).
```
Alizter: Should this be defined using J, or would it be better to just show the transports are pointed? You can use `Build_pEquiv'` to just give an Equiv and a proof that it is pointed.

theories/Pointed/pMap.v
```diff
@@ -54,7 +54,7 @@ Proof.
   - reflexivity.
 Qed.
 
-(* A kind of ap for pHomotopies *)
+(* A kind of ap for pHomotopies, obtained by going through funext.  Usually unnecessary. *)
 Definition phomotopy_ap `{Funext} {A B C D : pType} {f g : A ->* B}
```
Alizter: Since we aren't using this anymore, I think it's entirely reasonable to completely remove.

theories/Pointed/Core.v
mikeshulman: I don't think it matters much what we call the field of a one-field typeclass.

mikeshulman: But sure.

theories/Pointed/Loops.v
mikeshulman: I think it's still necessary, because unadorned `^-1` refers to the inverse function rather than the inverse equivalence.

theories/Pointed/Loops.v
mikeshulman: Let's not worry about that as part of this PR.

theories/Pointed/pEquiv.v
mikeshulman: Sure, why not.

theories/Pointed/pEquiv.v
```diff
@@ -109,12 +114,6 @@ Proof.
   apply eisadj.
 Qed.
 
-(* Sometimes we wish to construct a pEquiv from an equiv and a proof
-  that it is pointed *)
-Definition Build_pEquiv' {A B : pType} (f : A <~> B)
```
Alizter: Why was this removed?

mikeshulman: I had to move it to earlier in the file in order to use it in `pequiv_path`.

Alizter: Ah OK I missed that.

Alizter:
I think we are good to merge. I am not sure how useful it will be for someone else to look this over now, since this is just janitorial work.

mikeshulman:
Let's give them a chance.

PR #1095, opened 2019-10-06
Cleaned up Pushouts, coequalizers etc. closes #1080

### LINKED ISSUE #1080: Rename pp
I think the name `pp` for the glue in a `pushout` is not good.  It's too short and ambiguous for a global name.  Moreover, it leads to lemmas like `equiv_pushout_pp`, which is potentially conflicty with our convention of `pp` in lemmas to refer to a concatenation of paths (e.g. `transport_pp`).  Maybe something like `pglue`?

Alizter:
Here are the main changes:
 - Renamed `pp` to `pglue`.
 - Renamed `joinpp` to `jglue`.
 - Renamed `cp` to `cglue`.
 - Changed `pushout` to `Pushout` and associated functions.
 - Changed `join` to `Join`
 - Changed `spushout` to `SPushout` (and `sglue` to `spglue`. Will probably rename `glue` in `Smash.v` to `sglue` in the future).

I am not finished yet however. I would like to propose the following changes too:
 - Move `Pushout.v` to `Colimits/`.
 - Move `Join.v` to `Homotopy/`.
 - Move `Susp.v` to `Homotopy/`.
 - Move `SpanPushout.v` to `Colimits/`.

This is in accordance with the discussion in #1067. I haven't carried these ought yet so we can discuss them.

What do you all think?


mikeshulman:
I think your other 4 changes sound reasonable too.

mikeshulman:
(oops)

Alizter:
And one more thing I have been a bit unsure on. Should I change the induction principles for coeq, pushout, suspension and everything that relies on this to use DPaths instead?

Otherwise I will add DPath induction principles seperately when we need them.

Alizter:
Also I've noticed that the old Freudenthal is still getting compiled. I will remove it from `_CoqProject` and put it in `contrib/old`.

mikeshulman:
We could probably delete the old (incomplete) Freudenthal completely; anyone who wants it can find it in the git history.

mikeshulman:
I don't know about DPath.  Let's discuss that in a separate issue/PR.

Alizter:
@mikeshulman I will ask Peter Lumsdaine since it belongs to him. For now I think it does no harm in old.

PR #1102, opened 2019-10-11
reorganised truncations

### LINKED ISSUE #1081: Require Import UnivalenceAxiom leaks Univalence
If I have a file A.v that has `Require Import UnivalenceAxiom.` in it, and I have another file B.v that Require Imports A, then we currently have Univalence in B also.

This has caused me a few headaches since things that require Univalence suddenly work without wanting them to.

There are two solutions:
 - Don't use UnivalenceAxiom anywhere in the library (but then why does it exist?)
 - Somehow stop UnivalenceAxiom from leaking Univalence

I would prefer the second option, but I have no idea how to do this.

### LINKED ISSUE #1085: Organising truncations

### LINKED ISSUE #1100: Change name of PropResizing/Truncations
I would like to change the name of PropResizing/Truncations to something else for two reasons:

1. It stops anyone ever importing it instead of HIT.Truncations without knowing.
2. It stops a naming conflict to that we don't have to write HIT. everytime.

The second one is more important since in #1085, I am planning to have an index file Truncations.v which will export the contents of a Truncations folder. This currently clashes with PropResizing.

Since this is only about propositional truncation, perhaps a name like `PropTrunc.v` might be in order.

Another suggestion is `ImpredicativeTrunc.v` or some variant thereof.

### LINKED ISSUE #1101: UnivalenceImpliesFunext being leaked
It was suggested in #1081 that there be checks in HoTT.v to make sure none of the axiom files are being leaked.

I added the following at the end of HoTT.v:
```coq
(** We check that UnivalenceAxiom, FunextAxiom, UnivalenceImpliesFunext, IntervalImpliesFunext and TruncImpliesFunext aren't being leaked. *)
Fail Check HoTT.UnivalenceAxiom.univalence_axiom.
Fail Check HoTT.FunextAxiom.funext_axiom.
Fail Check HoTT.UnivalenceImpliesFunext.Univalence_implies_Funext.
Fail Check HoTT.HIT.IntervalImpliesFunext.funext_from_interval.
Fail Check HoTT.TruncImpliesFunext.funext_from_trunc.
```
Only `Univalence_implies_Funext` is failing. I notice that we are using this everywhere (in at least 20 files). The problem seems to be that if one file imports it than anything that imports this file inherits UnivalenceImpliesFunext.

Now I see two solutions to this:
 - Somehow stop the leakage. I don't know how to do this.
 - Assume UnivalenceImpliesFunext in Basics (i.e. move it there) so that it always holds.

I can't imagine why you would have univalence but not funext so the second option is more appealing to me. I won't suggest it however since I could be shortsighted.

Alizter:
I have moved HIT/Truncations to Truncations/Core and HIT/Connectedness to Truncations/Connectedness as discussed in #1085. There is an index file Truncations which imports both of these. Therefore if you need connectedness or truncations you simply need to import `Truncations`.

I have also regorganised the _CoqProject and HoTT.v files.

This PR closes #1081, closes #1085, closes #1101, closes #1100.


Alizter:
This now also closes #1100.

Alizter:
@JasonGross @spitters @andreaslyn Anyone else able to review?

PR #1105, opened 2019-10-16
Cleaned up quotient

### LINKED ISSUE #1088: Clean up quotient
I would like to propose cleaning up quotient.v. This would involve renaming the file, renaming some of the lemmas to adhere to conventions. It may also be worth considering if we want to define quotients as a coequalizer? We can keep comments detailing the private inductive version too.

What are your thoughts? Do you have anything to add?

### LINKED ISSUE #1103: Quotient has the wrong universes
In #1088 I tried defining `quotient` as a truncated coequalizer, which broke the rest of the library, notably V.v. I originally thought this was a problem with V.v, but I noticed after omitting it from the _CoqProject, other things break too.

A careful analysis of quotient.v led me to note that given a type A:U and a relation A -> A -> V, the quotient was living in U. There is the following comment in quotient.v which Jason wrote in #374 after noting that quotients had the wrong universe in #372.

`
Note: If we wanted to be really accurate, we'd need to put [@quotient A R sr] in the max [U_{sup(i, j)}] of the universes of [A : U_i] and [R : A -> A -> U_j].  But this requires some hacky code, at the moment, and the only thing we gain is avoiding making use of an implicit hpropositional resizing "axiom".
`

I made the following changes (I have omitted comments for brevity):
```coq
  Section Domain.

    Universe i j.

    Context {A : Type@{i}} (R:relation@{i j} A) {sR: is_mere_relation _ R}.

    Private Inductive quotient {sR: is_mere_relation _ R} : Type@{max(i,j)} :=
    | class_of : A -> quotient.
```
but as I said, V.v and other files rely on the quotient living in the same type as the one we are taking a quotient of in a crucial way.

Is I understand, with the way quotients are implemented in the HoTT book, unless propositional resizing is assumed, the truncated coequalizer definition doesn't seem to work. Funnily enough the HoTT book gives another definition of the quotient, as the set of equivalence classes, and makes careful note about its use of propositional resizing.

The HoTT book talks about using the Rezk completion as a third construction of the quotient. This however doesn't seem like it will land the quotient in the correct universe. Namely the universe of the type we are taking a quotient of.

- What is going on here?
- Am I correct in saying that unless we have propositional resizing, it is not immediate that quotient lives in the claimed universe?
- How can we fix this?

Alizter:
Here is a start at cleaning up quotient.v. I have added a new file Colimits/Quotient.v for the new quotient.

I have left two quotients for now for reasons discussed in #1103.

I haven't managed to change the uses of quotient in hottclasses to the new one but @SkySkimmer said he could do that.

I also couldn't change the use of quotient in V.v due to universes getting too complicated.

This closes #1103, closes #1088.

This can be merged for now but we will need to change the use of quotient.v in hottclasses and V.v, after which we can remove the old quotient.v.

Please let me know how you like the new names, I am happy to change any with discussion.

Edit: I should also add that I had an attempt at adding universe annotation to Coeq.v, Pushout.v and Quotient.v. Not 100% convinced that these are the correct ones so I would appreciate any help here.

### DIFF
--- a/theories/Colimits/Quotient.v
+++ b/theories/Colimits/Quotient.v
@@ -1,0 +1,405 @@
+Require Import Basics.
+Require Import Types.
+Require Import HSet.
+Require Import HProp.
+Require Import UnivalenceImpliesFunext.
+Require Import TruncType.
+Require Import HIT.epi.
+Require Import HIT.Coeq.
+Require Import Truncations.
+
+Local Open Scope path_scope.
+
+
+(** * Quotient of a Type by an hprop-valued relation
+
+We aim to model:
+<<
+Inductive quotient : Type :=
+   | class_of : A -> quotient
+   | related_classes_eq : forall x y, (R x y), class_of x = class_of y
+   | quotient_set : IsHSet quotient
+>>
+*)
+
+(** This development should be further connected with the sections in the book; see below.*)
+
+Definition Quotient@{i j k} {A : Type@{i}} (R : relation@{i j} A)
+  `{is_mere_relation _ R} : Type@{k}
+  := Trunc@{k} 0 (Coeq@{k k}
+    (fun x : sig@{k k} (fun a : A => sig (R a)) => x.1)
+    (fun x : sig@{k k} (fun a : A => sig (R a)) => x.2.1)).
+
+Definition class_of@{i j k} {A : Type@{i}} (R : relation@{i j} A)
+  `{is_mere_relation _ R} : A -> Quotient@{i j k} R := tr o coeq.
+
+Definition qglue@{i j k} {A : Type@{i}} {R : relation@{i j} A}
+  `{is_mere_relation _ R} {x y : A}
+  : R x y -> class_of@{i j k} R x = class_of R y
+  := fun p => ap tr (cglue (x; y; p)).
+
+Definition Quotient_ind@{i j k l}
+  {A : Type@{i}} (R : relation@{i j} A) `{is_mere_relation _ R}
+  (P : Quotient@{i j k} R -> Type@{l}) {sP : forall x, IsHSet (P x)}
+  (pclass : forall x, P (class_of R x))
+  (peq : forall x y (H : R x y), qglue H # pclass x = pclass y)
+  : forall q, P q.
+Proof.
+  eapply Trunc_ind, Coeq_ind.
+  1: assumption.
+  intros [a [b p]].
+  refine (transport_compose _ _ _ _ @ _).
+  apply peq.
+Defined.
+
+Definition Quotient_ind_beta_qglue@{i j k l}
+  {A : Type@{i}} (R : relation@{i j} A) `{is_mere_relation _ R}
+  (P : Quotient@{i j k} R -> Type@{l}) {sP : forall x, IsHSet (P x)}
+  (pclass : forall x, P (class_of R x))
+  (peq : forall x y (H : R x y), qglue H # pclass x = pclass y)
+  (x y : A) (p : R x y)
+  : apD (Quotient_ind@{i j k l} R P pclass peq) (qglue p) = peq x y p.
+Proof.
+  refine (apD_compose' tr _ _ @ _).
+  refine (ap _ (Coeq_ind_beta_cglue _ _ _ _) @ _).
+  apply concat_V_pp.
+Defined.
+
+Definition Quotient_rec@{i j k l}
+  {A : Type@{i}} (R : relation@{i j} A) `{is_mere_relation _ R}
+  (P : Type@{l}) `{IsHSet P} (pclass : A -> P)
+  (peq : forall x y, R x y -> pclass x = pclass y)
+  : Quotient@{i j k} R -> P.
+Proof.
+  eapply Trunc_rec, Coeq_rec.
+  intros [a [b p]].
+  by apply peq.
+Defined.
+
+Definition Quotient_rec_beta_qglue @{i j k l}
+  {A : Type@{i}} (R : relation@{i j} A) `{is_mere_relation _ R}
+  (P : Type@{l}) `{IsHSet P} (pclass : A -> P)
+  (peq : forall x y, R x y -> pclass x = pclass y)
+  (x y : A) (p : R x y)
+  : ap (Quotient_rec@{i j k l} R P pclass peq) (qglue p) = peq x y p.
+Proof.
+  refine ((ap_compose tr _ _)^ @ _).
+  serapply Coeq_rec_beta_cglue.
+Defined.
+
+Arguments Quotient : simpl never.
+Arguments class_of : simpl never.
+Arguments qglue : simpl never.
+Arguments Quotient_ind_beta_qglue : simpl never.
+Arguments Quotient_rec_beta_qglue : simpl never.
+
+(* TODO: add to Notations.v *)
+Notation "A / R" := (Quotient (A:=A) R).
+
+(*
+Module Export Quotient.
+
+  Section Domain.
+
+    Context {A : Type} (R:relation A) {sR: is_mere_relation _ R}.
+
+    (** We choose to let the definition of quotient depend on the proof that [R] is a set-relations.  Alternatively, we could have defined it for all relations and only develop the theory for set-relations.  The former seems more natural.
+
+We do not require [R] to be an equivalence relation, but implicitly consider its transitive-reflexive closure. *)
+
+
+    (** Note: If we wanted to be really accurate, we'd need to put [@quotient A R sr] in the max [U_{sup(i, j)}] of the universes of [A : U_i] and [R : A -> A -> U_j].  But this requires some hacky code, at the moment, and the only thing we gain is avoiding making use of an implicit hpropositional resizing "axiom". *)
+
+    (** This definition has a parameter [sR] that shadows the ambient one in the Context in order to ensure that it actually ends up depending on everything in the Context when the section is closed, since its definition doesn't actually refer to any of them.  *)
+    Private Inductive quotient {sR: is_mere_relation _ R} : Type :=
+    | class_of : A -> quotient.
+
+    (** The path constructors. *)
+    Axiom related_classes_eq
+    : forall {x y : A}, R x y ->
+                        class_of x = class_of y.
+
+    Axiom quotient_set : IsHSet (@quotient sR).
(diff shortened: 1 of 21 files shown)


theories/Colimits/Quotient.v
```diff
@@ -0,0 +1,353 @@
...
+    apply (transitivity p).
+  Defined.
+
+  (* Quotient induction into a hprop. *)
+  Definition Quotient_ind_hprop
+    (P : A / R -> Type) `{forall x, IsHProp (P x)}
+    (dclass : forall x, P (class_of R x)) : forall q, P q.
+  Proof.
+    serapply (Quotient_ind R P dclass).
+    1: intro; apply trunc_succ.
```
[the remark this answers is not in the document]
Alizter: For some reason my typeclass resolution didn't detect that. I will commit your suggestion and see what travis says.

theories/Colimits/Quotient.v
```diff
@@ -0,0 +1,353 @@
...
+   | class_of R : A -> Quotient R
+   | qglue : forall x y, (R x y) -> class_of R x = class_of R y
+   | ishset_quotient : IsHSet (Quotient R)
+>>
+We do this by defining the quotient as a truncated coequalizer.
+
+*)
+
+Definition Quotient@{i j k} {A : Type@{i}} (R : relation@{i j} A)
+  `{is_mere_relation _ R} : Type@{k}
```
SkySkimmer: Why do we have the `is_mere_relation` argument here btw?
The only time we need a is_mere_relation argument is in the Equiv section to define in_class.

Alizter: @SkySkimmer This was included in the original definition, see the comments in HIT/quotient.v. I didn't want to break anything by not including it.

SkySkimmer: I'm pushing a commit removing it, there is very little breakage.

Alizter: I don't think there is anything wrong with removing it, but it means it is slightly different than the definition given in the book.

theories/Colimits/Quotient.v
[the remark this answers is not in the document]
Alizter: It seems removing that line has caused travis to fail sometimes?

SkySkimmer:
I tried porting hottclasses (13702a1a7bc855364ae77208222f5edaa5e07546) but there's a problem.
https://github.com/HoTT/HoTT/blob/79329719654d8850f7384991f7f22d771cfd90ee/theories/Classes/theory/rationals.v#L254
`ring_tac.ring_with_integers (NatPair.Z nat).` explodes (it tried eating all my memory then I killed it)
I guess computing with the new quotient could be harder? I don't know what to do about this.

Alizter:
@SkySkimmer I think `Quotient_ind_beta_qglue` needs to be opaque.

SkySkimmer:
>@SkySkimmer I think Quotient_ind_beta_qglue needs to be opaque.

It's used for the `Z_compute_path` lemma, but that lemma isn't used anywhere so it doesn't matter.

Alizter:
@SkySkimmer What is your version of coq? Mine is
```
The Coq Proof Assistant, version 8.10+alpha (August 2019)
compiled on Aug 2 2019 17:55:12 with OCaml 4.05.0
```
For some reason, the line your removed at 126 of Quotient expects 2 goals for me.

SkySkimmer:
5d82451d6d6f591ad81919d8e6529cee48474b9e master

Gaëtan Gilbert

Alizter:
@SkySkimmer This is really strange. I thought the ring tactic was supposed abstract away the computation as a simple data structure and then compute. Why on earth does it struggle unfolding this?

SkySkimmer:
The ring tactic builds a polynomial whose variables correspond to non-ring operations. Then it normalizes the polynomial. If the 2 sides have the same normalized polynomials then they're equal.

The normalization requires a computable ring to provide coefficients for the polynomial. Currently only natpair_integers provides this for hott.

Alizter:
@SkySkimmer Would we be able to keep certain things opaque in natpair_integers to avoid the unfolding nightmare?

Alizter:
@SkySkimmer OK thanks for your help anyways!

theories/Colimits/Quotient.v
mikeshulman: I think it's fine to be more general than the book.  The book already remarks in 6.10.1 that A doesn't need to be a set; allowing R to not be a mere relation is along the same lines.  One of the nice things about a formalization is that it's easy to see exactly what results depend on which hypotheses.

theories/Colimits/Quotient.v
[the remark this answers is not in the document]
mikeshulman: My version of Coq is the same as Ali's, and also can't find this instance by itself.  We could make the script work in either case by changing the line to something like

```
all: try (intro; apply trunc_succ).
```

mikeshulman:
It turns out the problem with universes in V.v was quite trivial: `bisimulation` was defined to take values in `hProp@{U'}` which should be `hProp@{U}`.  The parameter to `hProp` is not the universe that it *belongs* to, but the universe that its *elements* belong to.  All the code was correct, only the universe annotation was wrong; changing that one line makes it work.

I submitted this as a PR to @Alizter's branch.  While I was at it, I removed the nonce `RPushout` in `V.v` in favor of the generic `SPushout` that we now have.

mikeshulman:
Wouldn't it be better to just use a more efficient definition of the integers?  Like the binary ones defined in the Coq standard library (https://coq.inria.fr/distrib/current/stdlib/Coq.Numbers.BinNums.html)?  Even the unary integers defined in HoTT.Spaces.Int would be free of quotient issues.

Alizter:
@mikeshulman I was wondering this also. It seems kind of strange that the rationals rely on two quotients when we really only need one.

mikeshulman:
If we're ever going to do anything substantial with, say, spectral sequences or localization, it would be nice to have performant algebra.  Does hottclasses include algebraic localizations yet?

(Looks like 62b3612 fixed Travis; does it also work for you @SkySkimmer?)

theories/Colimits/Quotient.v
```diff
@@ -0,0 +1,352 @@
...
+    apply (transitivity p).
+  Defined.
+
+  (* Quotient induction into a hprop. *)
+  Definition Quotient_ind_hprop
+    (P : A / R -> Type) `{forall x, IsHProp (P x)}
+    (dclass : forall x, P (class_of R x)) : forall q, P q.
+  Proof.
+    serapply (Quotient_ind R P dclass).
+    all: try (intro; apply trunc_succ).
```
SkySkimmer: It works; but I think I'd prefer `1: try ...` + a comment about why we have to `try`.

mikeshulman: Why is `1:` preferable to `all:`?

SkySkimmer: Because there's no point in running the tactic on the second goal.

mikeshulman: `*shrug*` sure, if you want to change it that's fine with me.

Alizter: I am disconcerted by the fact that we need this at all. Why is there a difference between versions? Am I running on a version of coq which can't infer the typeclass?

Alizter:
@mikeshulman I don't think hottclasses has this yet.

I have some points I would like to clear up about localisation however:

In CORS they define a "localisation of (abelian) groups" to be something that satisfies the universal property of being a localisation (abelian) groups. Essentially the main theorem of CORS is showing that you can construct the localisation of a group G using the localisation of the space K(G,1).

There is however another construction in commutative algebra, used by algebraic geometers all the time. Which is the localisation of a ring or more generally an R-module. For abelian groups I would expect these to coincide, but CORS never really talks about this construction. For commutative rings: let R be a ring, and let D be a multiplicatively closed subset of R. Then define the congruence (r,d)~(s,e) iff x(er-ds)=0 for some x in D, where (r,s), (s,e) is in R x D. Now just take the quotient R x D / ~. This is basically just a small modification of the field of fractions construction, and is the "universal way" to make a set D into units in a ring R.

So when you ask does hottclasses have localisations yet, I am unsure which one you mean. We can formulate the universal property using hottclasses (which I have been working on doing). The constructions however I would only expect the later to come out of hottclasses since the former is what I am trying to formalise currently.

mikeshulman:
Localization in the sense used by CORS is the algebraic localization at the multiplicative set of the primes being localized away from.  The explicit construction of localization you gave is how to do it for a ring, then the case for modules can be obtained by tensoring with the localized ring.

Alizter:
All right, other than the fact this doesn't work for hottclasses this seems to be working fine. I don't think it would do any harm to leave hottclasses dependent on the old quotient. We can work on improving hottclasses' interaction with the new quotient in the future, but for now I feel like it would waste time.

Currently I am formalising abelianisation of a group, using hottclasses, but developed in the Algebra folder. Once we have this I will work on finishing the Eilenberg-MacLane spaces formalisation by avoiding the use of "coherent H-spaces" and just using the surjectivity of quotients maps.

mikeshulman:
I agree, we can just open a separate issue for "make hottclasses use new quotient" and merge this as-is.

spitters:
About effective algebra and HoTT. This is part of @andreaslyn PhD topic.
We may have time to look at this next week.

spitters:
We have the field of fractions here:
https://github.com/HoTT/HoTT/blob/master/theories/Classes/implementations/field_of_fractions.v
What kind of performant library of algebra are you imagining/would you need?
When designing math-classes we kept efficiency in mind, but we also have other methods available.

mikeshulman:
> What kind of performant library of algebra are you imagining/would you need?

The issue raised above is that with the integers defined as a quotient of pairs of naturals, with the new definition of quotient, the ring solver tactic explodes on simple calculations.  For the moment, I just meant we should fix this!  Using a more efficient notion of integers seems to me like an obvious approach, but I don't have a lot of experience with this sort of issues.

mikeshulman:
Can we merge this?  And maybe just add a to-do issue saying to update the use of quotients or integers in mathclasses in the future?

spitters:
Yes, LGTM, please feel free to merge after conflicts are resolved and you've added the todo.

Alizter:
rebased

Alizter:
@spitters we have rebased and travis is happy so we can merge.

PR #1118, opened 2019-10-29
renamed BuildIsEquiv to Build_IsEquiv and BuildEquiv to Build_Equiv

Alizter:
As the title says.

I also capitalised in Cantor.v and added the equivalence going the other way.


Alizter:
This is guaranteed to break other PRs to if we merge we should restart travis everywhere.

PR #1106, opened 2019-10-18
new and improved issig

Alizter:
Here is the new issig tactic from #1093.

Here are the biggest timing changes. As you can see it is near negligible. I may have even moved the mouse around a bit which may account for some of the changes.

<details>
  <summary>Timing (Click me to open)</summary>
<p>

```
After    | File Name                                                 | Before   || Change    | % Change
-------------------------------------------------------------------------------------------------------
7m40.72s | Total                                                     | 7m38.17s || +0m02.55s | +0.55%  
-------------------------------------------------------------------------------------------------------
0m29.72s | Categories/Category/Sigma/Univalent                       | 0m29.66s || +0m00.05s | +0.20%  
0m22.40s | Homotopy/BlakersMassey                                    | 0m22.35s || +0m00.04s | +0.22%  
0m19.81s | Classes/implementations/natpair_integers                  | 0m19.59s || +0m00.21s | +1.12%  
0m16.78s | Cubical/Cube                                              | 0m16.77s || +0m00.01s | +0.05%  
0m14.30s | Modalities/ReflectiveSubuniverse                          | 0m14.28s || +0m00.02s | +0.14%  
0m13.81s | Spaces/BAut/Bool                                          | 0m13.70s || +0m00.11s | +0.80%  
0m12.66s | Comodalities/CoreflectiveSubuniverse                      | 0m12.67s || -0m00.00s | -0.07%  
0m10.32s | Spaces/Torus/TorusEquivCircles                            | 0m10.41s || -0m00.08s | -0.86%  
0m09.90s | Modalities/Topological                                    | 0m09.90s || +0m00.00s | +0.00%  
0m08.90s | Spaces/Finite                                             | 0m08.64s || +0m00.25s | +3.00%  
0m08.59s | Idempotents                                               | 0m08.53s || +0m00.06s | +0.70%  
0m08.33s | Categories/Adjoint/Pointwise                              | 0m08.39s || -0m00.06s | -0.71%  
0m07.63s | Spaces/No/Core                                            | 0m07.60s || +0m00.03s | +0.39%  
0m07.17s | Modalities/Lex                                            | 0m07.21s || -0m00.04s | -0.55%  
0m07.13s | Colimits/Colimit_Flattening                               | 0m07.16s || -0m00.03s | -0.41%  
0m06.63s | Classes/theory/premetric                                  | 0m06.42s || +0m00.20s | +3.27%  
[...]

PR #1122, opened 2019-11-03
Basic group theory, abelianization

### LINKED ISSUE #1107: Many hottclasses classes are properties rather than structures.
In hottclasses we have many structures such as Group which I think should really be called IsGroup. If we wanted to talk about the type of groups, we would have to use the sigma type `sig Group` which isn't valid since `Group` needs other data too. I propose we rename some of these structures X to IsX in order to make this clear. If we want to discuss the type of Groups we would then have the appropriate total space `Group` which would consist of all the group data.

This would be more inline with the HoTT book too. A similar thing happens with pointed types, equivalences etc.

Alizter:
As of the time of submitting this PR there will be a lot of commits that come from #1106. The only commits that matter here are the last two. I trust that once #1106 is merged, I can rebase and make this easier to review. You should be able to click on the individual commits to only see the changes there.

I can summarize the changes as follows:

 * Changed X to IsX for many properties in hottclasses. This may have not been done perfectly and I don't think I caught them all but I have done most of them. This closes #1107. I have also shown that many of these properties are hprops, i.e. (what used to be Group but now) IsGroup is a hprop.

* I have started a file called Group.v in Algebra. The purpose of this file is to be a basic start on group related things. It defines what a group is (basically a record which is the total space of IsGroup). There are basic properties of groups in there and documented too. If you want to do group theory in the HoTT library, you can just import this file and you should have everything important. Later if we develop any actual group theory, say Lagrange's theorem, we can store such results in a separate directory. But you would expect that all group related things use Group.v.

* Subgroup.v defines subgroups, normal subgroups and collects results about them. Notably I have shown that normal subgroups give a congruence.

* Congruence.v defines a congruence. Classically this is required to be an equivalence relation, but I have just defined it to mean op preserving. This might bite me later so I have made a note of it here and in the file.

* QuotientGroup.v defines the group given by the quotient of a group by a congruence relation. It also defines the quotient of a group by a normal subgroup as the quotient by the congruence corresponding to the normal subgroup.

* AbelianGroup.v defines abelian groups, the property of being an abelianization and a construction of abelianization using a HIT. We define this as a set-coequalizer and it allows us to prove that abelianizations exist and are notably surjective. This was a key part in the argument that was needed to finish #1086. So once this gets merged I will send another PR for Eilenberg-MacLane spaces that is complete.

There are things I would like to do but I couldn't manage.

* Define order of a group. I imagine this will only require Card.v and when the cardinality is finite, we need notions of divisibility. This will eventually let us prove Lagrange's theorem.

* Show that the type of group isomorphisms is equivalent to the path space of groups. This will look something like:
```coq
Definition equiv_path_group {U : Univalence} {G H : Group}
  : GroupIsomorphism G H <~> G = H.
Proof.
  refine (_ oE (issig_GroupIsomorphism _ _)^-1).
  eqp_issig_contr issig_group.
Admitted.
```
but I didn't have enough time to finish it. Since the property of being a group is a proposition we essentially need to show that the sigma type containing a type, an identity element, an inverse map and an operation are equivalent to the path space using univalence.

### DIFF
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -1,0 +1,342 @@
+Require Import Basics.
+Require Import Types.
+Require Import Truncations.
+Require Import HIT.Coeq.
+Require Import Algebra.Group.
+Require Import Algebra.Subgroup.
+Require Import Cubical.
+Import TrM.
+
+(** * Abelian groups *)
+
+(** Definition of an abelian group *)
+
+Class AbGroup := {
+  abgroup_type : Type;
+  abgroup_sgop :> SgOp abgroup_type;
+  abgroup_unit :> MonUnit abgroup_type;
+  abgroup_inverse :> Negate abgroup_type;
+  abgroup_isabgroup :> IsAbGroup abgroup_type;
+}.
+
+(** We want abelian groups to be coerced to the underlying type. *)
+Coercion abgroup_type : AbGroup >-> Sortclass.
+
+(** The underlying group of an abelian group. *)
+Definition group_abgroup : AbGroup -> Group.
+Proof.
+  intros [G].
+  serapply (Build_Group G).
+Defined.
+
+(** We also want abelian groups to be coerced to the underlying group. *)
+Coercion group_abgroup : AbGroup >-> Group. 
+
+(** Definition of Abelianization.
+
+  Given a map F that turns any group into an abelian group, and a unit homomorphism eta_X : X -> F X. This data is considered an Abelianization if and only if for all maps X -> A, there exists a unique g such that h == g o eta X. *)
+Definition IsAbelianization (F : Group -> AbGroup)
+  (eta : forall X, GroupHomomorphism X (F X))
+  := forall (X : Group) (A : AbGroup) (h : GroupHomomorphism X A),
+    Contr (exists (g : GroupHomomorphism (F X) A), h == g o eta X).
+
+Existing Class IsAbelianization.
+
+(** Here we define abelianization as a HIT. Specifically as a set-coequalizer of the following to maps: (a, b, c) |-> a (b c) and (a, b, c) |-> a (c b).
+
+From this we can show that Abel G is an abelian group.
+
+In fact this models the following HIT:
+
+HIT Abel (G : Group) := 
+ | ab : G -> Abel G
+ | ab_comm : forall x y z, ab (x & (y & z)) = ab (x & (z & y)).
+
+We also derive ab and ab_comm from our coequalizer definition, and even prove the induction and computation rules for this HIT.
+
+This HIT was suggested by Dan Christensen.
+*)
+
+Section Abel.
+
+  (** Let G be a group. *)
+  Context (G : Group).
+
+  (** We locally define a map uncurry2 that lets us uncurry A * B * C -> D twice. *)
+  Local Definition uncurry2 {A B C D : Type}
+    : (A -> B -> C -> D) -> A * B * C -> D.
+  Proof.
+    intros f [[a b] c].
+    by apply f.
+  Defined.
+
+  (** The type Abel is defined to be the set coequalizer of the following maps G^3 -> G. *)
+  Definition Abel
+    := Tr 0 (Coeq
+      (uncurry2 (fun a b c => a & (b & c)))
+      (uncurry2 (fun a b c => a & (c & b)))).
+
+  (** We have a natural map from G to Abel G *)
+  Definition ab : G -> Abel.
+  Proof.
+    intro g.
+    apply tr, coeq, g.
+  Defined.
+
+  (** This map has to satisfy the condition ab_comm *)
+  Definition ab_comm a b c
+    : ab (a & (b & c)) = ab (a & (c & b)).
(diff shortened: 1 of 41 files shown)


Alizter:
Travis is failing for some reason.

spitters:
I've restarted the build.

Alizter:
@JasonGross Do you know why some travis builds are failing here?

spitters:
@Alizter Do you have access to the Travis logs? I.e. click on Details. 

It says:
Welcome to Coq 8.11+alpha (November 2019)
Error: Unable to locate library

Not sure what caused this?

Alizter:
@spitters Exactly, I have no idea what this is happening. I will try and debug it later today.

Alizter:
@spitters It seems to have nothing to do with my branch here. I tried my own travis build on master and that failed too with the same error. This means travis isn't working on the entire library for some reason.

spitters:
It seems to have been this one. Not sure how, because Travis seemed happy
when I pulled your PR.
https://travis-ci.org/HoTT/HoTT/builds/604939678

Alizter:
@spitters I think it is an issue with master because I can't get travis to be happy with master at the moment. See #1123.

spitters:
As I said, the problem seems to have been introduced by your quotient PR.
(Not entirely sure how that is possible, but Travis send me an email about
that).

mikeshulman:
This is substantial!  I'm surprised that hottclasses doesn't have any of this basic stuff about group theory yet.

mikeshulman:
I didn't realize you were working towards Lagrange's theorem.  I actually have a proof of Lagrange's theorem sitting around that I never got around to cleaning up and merging: https://github.com/mikeshulman/HoTT/blob/groups/theories/Algebra/FiniteGroup.v .  It's based on groups defined as ooGroups that happen to be sets, but could presumably be modified to use the groups from hottclasses.

Alizter:
@mikeshulman That's great! I will see if I can add it to the library after this PR gets merged. Currently I am rewriting parts of my EM spaces formalization to work with this group theory.

Alizter:
@spitters Just to be clear. The current issue with travis is caused by the latest version of coq 8.11 and nothing to do with my PRs. Travis works for 8.10 as it should, but not currently 8.11.

The reason for this is someone is messing around with what needs to be installed by the makefile in 8.11 and since our makefile is from 8.10 it doesn't work. There is not much we can do apart from wait for 8.11 to be fixed.

Alizter:
I have rebased now that #1106 has been merged.

theories/Algebra/AbelianGroup.v
```diff
@@ -0,0 +1,342 @@
...
+  abgroup_type : Type;
+  abgroup_sgop :> SgOp abgroup_type;
+  abgroup_unit :> MonUnit abgroup_type;
+  abgroup_inverse :> Negate abgroup_type;
+  abgroup_isabgroup :> IsAbGroup abgroup_type;
+}.
+
+(** We want abelian groups to be coerced to the underlying type. *)
+Coercion abgroup_type : AbGroup >-> Sortclass.
+
```
spitters: Can't we use 
abgroup_type :> Type;
 above ? I know there was some difference in the how this is used in Classes and in Records, but I don't know the current status.

theories/Algebra/Group.v
```diff
@@ -0,0 +1,254 @@
...
+Proof.
+  refine ((right_identity x)^ @ ap _ q^ @ _).
+  refine (associativity _ _ _ @ _).
+  refine (ap (& y) p @ _).
+  apply left_identity.
+Defined.
+
+(** Inverses are involutive *)
+Check negate_involutive.
+
```
spitters: Please comment these Check's out, as they will pollute running the script with make.

Alizter: Is there a silent check we can use? We have a similar problem in Idempotents.v too.

theories/Algebra/AbelianGroup.v
Alizter: I couldn't get that to work when I tried it. :> seems to work when going from class to class but for whatever reason didn't work when going from class to type.

spitters:
Yes, this means subclass, not coercion, I thought it had changed.
In any case, feel free to pull.

theories/Algebra/Group.v
mikeshulman: It's unfortunate that the checks in Idempotents spit output, but they do serve a purpose of verifying the correct universe behavior.  Do these checks serve any similar purpose?

Alizter: There not as critical as the one in idempotents, but they will catch any fiddling around with hottclasses in the future. I will comment these ones out for now.

mikeshulman: Can you explain?  What sort of fiddling around?  Is it possible for a Check of a single bare identifier to fail?

Alizter: @mikeshulman All I mean is that I would like these definitions to exist. If they get renamed, moved or even deleted, there might be problems, especially if I am not using them currently.

Mostly they are serving as a documentation, so people don't have to go diving through hottclasses to use the group theory library.

I've commented them out now.

PR #1128, opened 2019-11-15
Various truncation lemmas

### LINKED ISSUE #1084: Swapping order of truncation
Is there an easy proof for the following lemma?
```coq
Require Import Basics Types HIT.Truncations.
Import TrM.

Definition Trunc_swap n m X : Tr n (Tr m X) <~> Tr m (Tr n X).
Proof.
Admitted.
```

I managed to prove it by defining a min function on truncindex, showing that truncations cummulate their indexes in a min and then showing that min can have its order swapped. This seems a bit long winded to me, but is ok I guess.

Alizter:
Here are various truncation lemmas, they mostly come from my EM spaces development. Instead of shoving them all in that PR I will give them their own PR.

I will start with the small changes:

 1. Connectedness of suspension added into Homotopy/Suspension.v. Peter Lumsdaine gave a proof of this in the old Freudenthal.v file. I included the proof here, which uses the wedge connectivity lemma.
 2. Added `psusp_2functor`, `pequiv_psusp_functor` to Pointed/pSusp.v. Also added two corollaries of the Freudenthal suspension theorem:
    - The map `loop_susp_unit` is 2n-connected.
    - We have that Tr 2n X <~> Tr 2n (loops susp X)
 3. Added a more general connected predecessor lemma to Truncation/Connectedness.v stating that if a type is (n+2+m)-connected then it is m-connected.
 4. Added the equivalence given by the functor pTr in Pointed/pTr.v.
 5. Added that f is O-connected then O_functor O f is also O-connected to Modalities/ReflectiveSubuniverse.v
 6. To TruncType.v, added the lemma ap_trunctype, which says `ap`ing `trunctype_type` onto `path_trunctype` turns into a `path_universe_uncurried`.

Now for the big ones:

 7. To Basics/Trunc.v I have added `trunc_index_leq` which is what you think. It comes with various lemmas too. I have added `trunc_index_min` which is the minimum of the trunc indexes. I have also added a predecessor function for trunc_index. There is also some cleanup in this file.

 8. In Truncations/Core.v I have proven Trunc_min which is the following equivalence:
`Trunc_min n m X : Tr n (Tr m X) <~> Tr (trunc_index_min n m) X`. And also ` Trunc_swap n m X : Tr n (Tr m X) <~> Tr m (Tr n X)`. This closes #1084.

### DIFF
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -64,7 +64,7 @@ Fixpoint trunc_index_inc (n : nat) (k : trunc_index)
 Definition trunc_index_pred : trunc_index -> trunc_index.
 Proof.
   intros [|m].
-  1: exact -2.
+  1: exact (-2).
   exact m.
 Defined.
 
@@ -119,14 +119,14 @@ Fixpoint trunc_index_min (n m : trunc_index)
   : trunc_index.
 Proof.
   destruct n.
-  1: exact -2.
+  1: exact (-2).
   destruct m.
-  1: exact -2.
+  1: exact (-2).
   exact (trunc_index_min n m).+1.
 Defined.
 
 Definition trunc_index_min_minus_two n
-  : trunc_index_min n -2 = -2.
+  : trunc_index_min n (-2) = -2.
 Proof.
   by destruct n.
 Defined.
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -17,6 +17,25 @@ Fixpoint trunc_index_add (m n : trunc_index) : trunc_index
 
 Notation "m +2+ n" := (trunc_index_add m n) : trunc_scope.
 
+Definition trunc_index_add_minus_two m
+  : m +2+ -2 = m.
+Proof.
+  induction m.
+  1: reflexivity.
+  cbn; apply ap.
+  assumption.
+Defined.
+
+Definition trunc_index_add_succ m n
+  : m +2+ n.+1 = (m +2+ n).+1.
+Proof.
+  revert m.
+  induction n; induction m.
+  1,3: reflexivity.
+  all: cbn; apply ap.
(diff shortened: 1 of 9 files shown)


Alizter:
Again travis will fail since as of the time of writing coq/coq#11075 still hasn't been merged into coq master. Its important that at least 3 of the jobs succeed.

theories/Basics/Trunc.v
```diff
@@ -17,26 +17,177 @@ Fixpoint trunc_index_add (m n : trunc_index) : trunc_index
...
+
+Global Instance trunc_index_leq_refl
+  : Reflexive trunc_index_leq.
+Proof.
+  intro n.
+  by induction n as [|n IHn] using trunc_index_ind.
+Defined.
+
+Global Instance trunc_index_leq_transitive
+  : Transitive trunc_index_leq.
```
mikeshulman: In general I don't think transitivity lemmas are very good as Instances; Coq has very little chance to guess the intermediate object.

mikeshulman:
I also checked that it compiles by fetching from your branch to my local install.

theories/Basics/Trunc.v
Alizter: I thought this allowed the `transitivity` tactic to pick them up.

Alizter: I don't think it is actually hinting instances of `trunc_index_leq`, just that it is a transitive relation. We do this all over the place.

mikeshulman: Oh right, `Transitive` is also a typeclass.  Sorry for the noise.

Alizter: No problem :-)

Alizter:
If #1133 gets merged before this I will need to update this PR.

Alizter:
rebased

Issue #1124, opened 2019-11-08
Integers in HoTT


Alizter:
I propose we start collecting different definitions of the integers in a folder in Spaces. Currently we have two main definitions floating around:
 * The integers as N + 1 + N defined in Spaces.Int.
 * The integers as N * N / ~ defined in natpair_integers.v in Classes/implementations

I think it will be beneficial for us to put these in one place, with some auxiliary files showing that they are all somehow equivalent.

Now one of the interesting things about HoTT is that we can come up with some more definitions of the integers. There are a few directly as HITs and of course another as the loop space of the circle.

Currently if somebody says: "Hey, I need to use the integers" they have to choose between Spaces.Int and natpair_integers. The definition in Spaces.Int is a bit tricky to work with, we don't even have a definition of addition or multiplication yet. And the natpair version is hidden away in hottclasses for everybody to miss.

Personally, I have a hard time reading the code in the natpair version too, since I am not used to that style. Not to mention, when using natpair, you have to import the NatPair module. OK, fair enough, we import modules for modalities. But then when you want to talk about the type Z, you have to specify which naturals too.

This seems like a very convoluted way to talk about the integers, and is definitely not what I would call easy to use. I instead propose we define the integers like we do in the HoTT book as the quotient of the naturals. I think natpair is doing the same thing but in a very general way which makes it difficult to use and understand.

I understand some of the motivation behind this generality however, since when you can provide whichever implementation of the integers or naturals you can plug in the efficient versions for computations.

I think we need to have a discussion over how we want to implement the integers.

mikeshulman:
While it's a cute idea to define "a notion of integer" as hottclasses does that can be "implemented" in many ways, I feel like for practical use we ought to just pick one definition of "the integers" and stick with it.  We already have enough transporting across equivalences, I don't want the problem of putting together two different developments to be made more difficult by the fact that they use different notions of integer.  The possibility of the other definitions can be made visible by, for each definition, proving that our chosen notion of integer satisfies that definition (e.g. has the appropriate induction principle), and moreover that any type satisfying that definition is equivalent to our chosen one.

[...]

PR #1133, opened 2019-11-19
Removed -2 and -1 from trunc notations

Alizter:
As discussed in #1124, I have removed -2 and -1 from being their own separate notations for trunc indices. Since this notation was previously reserved, it caused `Numeral Notation`s to break for -1 and -2.

The real change here is that when passing trunc indices to functions, we need to bracket negative indices, i.e. write `foo (-2)` instead of `foo -2`. it could be argued that the new way is easier on the eyes too.

I also cleaned up the .gitignore file a bit and added the .vok extension.


Alizter:
Again travis is failing since latest coq version doesn't work with old makefiles. No problem as long as at least 3 succeed.

theories/Classes/implementations/natpair_integers.v
```diff
@@ -489,7 +489,7 @@ apply Z_path;red;simpl.
 ring_with_nat.
 Qed.
 
-Definition Zle_hProp@{} : Z -> Z -> TruncType@{UN} -1.
+Definition Zle_hProp@{} : Z -> Z -> TruncType@{UN} (-1).
```
mikeshulman: Can we not use the notation `hProp` here?

theories/Spaces/Finite.v
```diff
@@ -910,12 +910,12 @@ Definition leq_inj_finite `{Funext} {X Y} {fX : Finite X} {fY : Finite Y}
            (f : X -> Y) (i : IsEmbedding f)
 : fcard X <= fcard Y.
 Proof.
-  assert (MapIn -1 f) by exact _. clear i.
+  assert (MapIn (-1)%trunc f) by exact _. clear i.
```
mikeshulman: It's unfortunate that `MapIn n` is not definitionally equal to `IsTruncMap n`.  I wonder if there's a way to achive that.

mikeshulman: (Because we have a notation `IsEmbedding` for `IsTruncMap (-1)`.)

theories/Basics/Trunc.v
```diff
@@ -32,7 +32,7 @@ Fixpoint trunc_index_inc (n : nat) (k : trunc_index) : trunc_index
 (** ** Truncatedness proper. *)
 
 (** A contractible space is (-2)-truncated, by definition. *)
-Definition contr_trunc_minus_two `{H : IsTrunc -2 A} : Contr A
+Definition contr_trunc_minus_two `{H : IsTrunc (-2) A} : Contr A
```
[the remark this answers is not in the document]
Alizter: Good point.

theories/Spaces/Finite.v
Alizter: Are you sure?

The following seems to work:

```coq
Require Import Basics Types.
Require Import Truncations.

Import TrM.

Context {A B : Type}.
Goal @MapIn (-1) A B = IsEmbedding.
Proof.
reflexivity.
```

mikeshulman: Oh, right, I guess it's not definitional equality that's the issue but rather typeclass resolution.

I suppose it would probably be a terrible idea to move the definition of modalities into `Basics` so that `IsTrunc` and `IsTruncMap` could just be defined as notations for the modality versions...

Alizter:
@mikeshulman How about:

 * `mere` for `Tr (-1)`
 * mere_functor for `Trunc_functor (-1)`
 * The notation `"Notation | x |" := (tr (-1) x)`.

The mere might get a bit confused with merely however...

theories/Spaces/Finite.v
Alizter: I don't understand why you are suggesting this however?

```coq
Require Import Basics Types Truncations.

Import TrM.

Context {n : trunc_index}.

Goal IsTrunc n = In n.
  reflexivity.
```
works. And so does:

```coq
Require Import Basics Types Truncations.

Import TrM.

Context {n : trunc_index} {A B : Type}.

Goal @IsTruncMap n A B = @MapIn n A B.
  reflexivity.
```

What is the issue you are talking about exactly?

Alizter: Ohhh never mind I see the problem now. I don't know if we can fix this. Perhaps we should create an issue.

mikeshulman:
I don't think I could support such ungrammatical use of "mere" either.  (-:  It was just a thought, maybe not worth pursuing.

theories/Spaces/Finite.v
Alizter: So I tested if typeclasses can pick it up and it seems it can. Specifically at line 913 of finite, we don't need `assert (MapIn (-1)%trunc f) by exact _. clear i.`.

Alizter: I have been meaning to clean up Finite.v for some time now. I will create an issue later and document this in particular so we don't forget.

mikeshulman: That's interesting!  I wonder if typeclass inference changed.  I doubt that line would have been there if typeclass inference used to be able to pick it up automatically.

Alizter: Yes, I have added a reminder to clean these up in #1135.

Alizter:
@spitters @JasonGross @andreaslyn Anyone else able to review?

theories/Basics/Trunc.v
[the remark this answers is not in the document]
JasonGross: I think we want something to show up when we do `Search Contr IsTrunc`, for someone new to the library.  But it would probably be worth having a comment to this effect, if we decide to keep it.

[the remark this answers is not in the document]
mikeshulman: Ah, yes, I can see that.  Sure, we can keep it with a comment.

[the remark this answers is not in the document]
Alizter: I will undo my removal. What comment should we add?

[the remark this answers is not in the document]
Alizter: I still can't fathom anybody searching for `Search Contr IsTrunc`. If (-2)-truncation is defined to be contractibility, why would a new comer ever look for a lemma that achieves this if it is true by definition?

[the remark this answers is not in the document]
mikeshulman: They might not know that it is true by definition.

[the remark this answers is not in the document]
Alizter: @mikeshulman What comment should we add next to it then?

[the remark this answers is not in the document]
mikeshulman: "This function is the identity, so there is never any need to actually use it, but it exists to be found in searches"?

Alizter:
I dropped the removal commit and added a comment instead.

Alizter:
After this is merged I will create a PR for the binary integers.

PR #1137, opened 2019-11-21
Binary positive integers

Alizter:
Here we have the binary positive integers. These are constructed by defining `Pos` the type of positive binary integers, and defining `Int` as essentially `Pos + 1 + Pos`.

The old int has been removed, and in it's place you will find `Int.v` is an index file for the folder `Int` in `Spaces`. Inside this folder are various things to do with the integers.

I have gone ahead and rewritten HomotopyGroup.v and updated `TorusHomotopy.v` and `Pi1S1.v`. We now have *group isomorphisms* rather than equivalences.

We also have the group `Z` in all it's glory, which can be found in Algebra. I mean to reorganise the contents of Algebra in the future, since they are getting a bit messy.

Pos and Int have the appropriate numeral notations, you need only open positive_scope or int_scope respectively to use these.

Binary integers are also really fast compared to unary ones, so we can actually do computations with them. This might be useful for people.


theories/Algebra/Group.v
```diff
@@ -252,3 +256,58 @@ Proof.
   all: intro; apply negate_involutive.
 Defined.
 
+(** * Direct product of group *)
+
+Definition group_prod : Group -> Group -> Group.
```
Alizter: should probably rename this grp_prod

theories/Algebra/Group.v
```diff
@@ -252,3 +256,58 @@ Proof.
...
+  { intros [g h].
+    apply path_prod; cbn.
+    1,2: apply left_inverse. }
+  (** Right inverse *)
+  { intros [g h].
+    apply path_prod; cbn.
+    1,2: apply right_inverse. }
+Defined.
+
+Definition groupiso_prod {A B C D : Group}
```
Alizter: grp_iso_prod

theories/Algebra/Z.v
```diff
@@ -0,0 +1,34 @@
...
+    exact int_add_0_l.
+  + (** Right identity *)
+    exact int_add_0_r.
+  + (** Left inverse *)
+    exact int_add_negation_l.
+  + (** Right inverse *)
+    exact int_add_negation_r.
+  + (** Commutativity *)
+    exact int_add_comm.
+Defined.
```
Alizter: add newline here

theories/Homotopy/HomotopyGroup.v
```diff
@@ -2,124 +2,194 @@ Require Import Basics.
...
+  { intro.
+    refine (_^ @ _ @ _).
+    1: apply homotopygroup_functor_compose.
+    2: apply homotopygroup_functor_idmap.
+    apply homotopygroup_2functor.
+    apply peissect. }
+Defined.
 
-Local Open Scope pointed_scope.
+(* TODO: move these to a Utf8 file and make them more global. *)
```
Alizter: actually do this

theories/Homotopy/HomotopyGroup.v
```diff
@@ -2,124 +2,192 @@ Require Import Basics.
...
-  Global Instance PiOp_rightInv : RightInverse PiOp PiInverse PiUnit.
-  Proof.
-    intro x; strip_truncations; cbn.
-    apply ap, concat_pV.
-  Defined.
+(** Definition of the nth homotopy group *)
+Definition HomotopyGroup (n : nat) (X : pType) : HomotopyGroup_type n.
+Proof.
+  destruct n.
+  1: exact X.
```
mikeshulman: This should be `Tr 0 X`.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -2,124 +2,192 @@ Require Import Basics.
...
+    apply (ap tr).
+    apply concat_pV.
+Defined.
 
-  Global Instance pi_is_Group : IsGroup (Pi n.+1 X).
-  Proof.
-    repeat split; exact _.
-  Defined.
+(** HomotopyGroup can be a bit of a mouthful so we define a notation for it. *)
+Global Notation Pi := HomotopyGroup.
```
mikeshulman: What is the point of using the name `HomotopyGroup` to start with, if we then define a global notation for it?

theories/Homotopy/HomotopyGroup.v
Alizter: huh weird that I missed that.

theories/Homotopy/HomotopyGroup.v
Alizter: I think I will change it to what I mentioned in the other thread. HomotopyGroup for the group, and Pi for the set.

mikeshulman: Hang on, I think it may not be necessary, I'm working on something.

theories/Homotopy/HomotopyGroup.v
Alizter: @mikeshulman I won't change this now since you are fixing it in #1142.

mikeshulman:
APOE?

Alizter:
@mikeshulman What does that mean?

Alizter:

andreaslyn:
I do not have time to review at the moment. If no one else reviews, then I will find some time in next week.

Alizter:
@andreaslyn OK no problem, thanks for letting us know!

Alizter:
@JasonGross are you available for review?

andreaslyn:
@Alizter It is easier to review your PRs if you make them smaller.

Alizter:
@andreaslyn You’re right, sorry about that. My new PRs are more bite sized.

PR #1142, opened 2019-11-29
Truncated Whitehead's Principle (based on a start by Alizter)

### LINKED ISSUE #1141: Whitehead's theorem
I've been trying to prove Whithead's theorem for truncated types. I managed to reduce it down to this one lemma but the path algebra got too difficult for me. I would be interested if anybody can prove this. The book is a bit vague about changing the basepoint of a loop space.

```coq

Require Import Basics.
Require Import Types.
Require Import Pointed.
Require Import Fibrations.
Require Import EquivalenceVarieties.
Require Import Algebra.Group.
Require Homotopy.HomotopyGroup.
Require Import Truncations.
Import TrM.

Local Open Scope nat_scope.

Section BookLemmata.

  Context `{Univalence} {A B : Type} {f : A -> B}.

  (** 8.8.1 *)
  Theorem book_8_8_1
    {i  : IsSurjection (Trunc_functor 0 f)}
    {ii : forall x y, IsEquiv (@ap _ _ f x y)}
    : IsEquiv f.
  Proof.
    apply (equiv_isequiv_ap_isembedding f)^-1 in ii.
    serapply isequiv_surj_emb.
    serapply BuildIsSurjection.
    cbn; intro b.
    pose proof (@center _ (i (tr b))) as p.
    revert p.
    apply Trunc_functor.
    apply sig_ind.
    serapply Trunc_ind.
    intros a p.
    apply (equiv_path_Tr _ _)^-1 in p.
    strip_truncations.
    exists a.
    exact p.
  Defined.

  Lemma book_8_8_2
    {i  : IsEquiv (Trunc_functor 0 f)}
    {ii : forall x,
      IsEquiv (@loops_functor (Build_pType A x) (Build_pType B (f x))
        (Build_pMap (Build_pType A x) (Build_pType B (f x)) f idpath))}
    : IsEquiv f.
  Proof.
    serapply book_8_8_1.
    intros x y.
    apply isequiv_inhab_codomain.
    intro p.
    apply (ap (@tr 0 _)) in p.
    apply (ap (Trunc_functor 0 f)^-1) in p.
    pose proof ((eissect (Trunc_functor 0 f) (tr x))^
      @ p @ eissect (Trunc_functor 0 f) (tr y)) as q.
    clear p.
    apply (equiv_path_Tr _ _)^-1 in q.
    strip_truncations.
    destruct q.
    cbn in ii.
    serapply isequiv_homotopic.
    { intro p.
      exact (1 @ (ap f p @ 1)). }
    1: apply (ii x).
    exact (fun _ => concat_1p _ @ concat_p1 _).
  Defined.

End BookLemmata.

(** Truncated Whitehead's theorem *)

(** We don't use pi_functor from HomotopyGroup.v since it contains more information than we need. The type changes depending on n > 0, and cannot be made to type check even if we just need the underlying map. *)
Definition map_pi_functor k {A B : Type} (f : A -> B) (x : A)
  : Tr 0 (iterated_loops k (Build_pType A x))
    -> Tr 0 (iterated_loops k (Build_pType B (f x))).
Proof.
  apply Trunc_functor.
  apply iterated_loops_functor.
  serapply Build_pMap.
  1: apply f.
  reflexivity.
Defined.

Local Tactic Notation "transparent" "assert"
                "(" ident(name) ":" constr(type) ")" :=
                simple notypeclasses refine (let name := (_ : type) in _). 

Definition isequiv_unfold_map_pi_functor
  (k : nat) {A B : Type} (f : A -> B) (x : A) (p : x = x)
  : IsEquiv (map_pi_functor k.+1 f x)
  -> IsEquiv (map_pi_functor k (loops_functor
    (Build_pMap (Build_pType A x) (Build_pType B (f x)) f idpath)) p).
Proof.
  intro e.
  transparent assert (F : (iterated_loops k.+1 (Build_pType A x) <~>
    (iterated_loops k (Build_pType (loops (Build_pType A x)) p)))).
  { refine (_ o*E pequiv_unfold_iterated_loops _ _).
    apply pequiv_iterated_loops_functor.
    serapply Build_pEquiv'.
    1: apply (equiv_concat_l p).
    apply concat_p1. }
  transparent assert (G : (iterated_loops k.+1 (Build_pType B (f x)) <~>
    iterated_loops k (Build_pType (loops (Build_pType B (f x))) ((loops_functor
      (Build_pMap (Build_pType A x) (Build_pType B (f x)) f 1)) p)))).
  { refine (_ o*E pequiv_unfold_iterated_loops _ _).
    apply pequiv_iterated_loops_functor.
    serapply Build_pEquiv'.
    1: apply (equiv_concat_l (1 @ (ap f p @ 1))).
    apply concat_p1. }
  simple notypeclasses refine
    (@isequiv_commsq _ _ _ _ _ _ _ _ _ e _ _).
  1: apply Trunc_functor, F.
  1: apply Trunc_functor, G.
  2,3: exact _.
  etransitivity.
  1: symmetry; serapply O_functor_compose.
  etransitivity.
  2: serapply O_functor_compose.
  serapply O_functor_homotopy.
  cbv beta zeta.
  intro q.
  simpl.

Admitted.

Theorem whitehead `{H : Univalence} {A B : Type} n
  {H0 : IsTrunc n A} {H1 : IsTrunc n B} (f : A -> B)
  {i  : IsEquiv (Trunc_functor 0 f)}
  {ii : forall (x : A) (k : nat), IsEquiv (map_pi_functor k f x)}
  : IsEquiv f.
Proof.
  revert A B H0 H1 f i ii.
  induction n.
  1: intros; apply isequiv_contr_contr.
  intros A B H0 H1 f i ii.
  notypeclasses refine (@book_8_8_2 H A B f i _).
  intro x.
  pose proof (@istrunc_paths _ _ H0 x x) as h0.
  pose proof (@istrunc_paths _ _ H1 (f x) (f x)) as h1.
  apply (IHn _ _ h0 h1 _ (ii x 1)).
  intros p k.
  cbv zeta beta.
  unfold iterated_loops_functor.
  apply isequiv_unfold_map_pi_functor.
  apply ii.
Defined.


```

mikeshulman:
Closes #1141


Alizter:
LGTM apart from the fact that I rewrote a lot of homotopy group in #1137. There I made the decision to have the type of Pi depend on n. The advantage however is that we could coerce the group Pi to the underlying set.

What should the type Pi n X really be? Should it just be a set, which can later be turned into a group (or an abelian group)? Or should it be a group (or abelian group) that can always be turned into a set?

theories/Pointed/Loops.v
```diff
@@ -190,6 +194,37 @@ Proof.
   apply peissect.
 Defined.
 
+(** A version of [unfold_iterated_loops] that's an equivalence rather than an equality.  We could get this from the equality, but it's more useful to construct it explicitly since then we can reason about it.  *)
+Definition unfold_iterated_loops' (n : nat) (X : pType)
+  : iterated_loops n.+1 X <~>* iterated_loops n (loops X).
+Proof.
+  induction n.
+  1: reflexivity.
+  change (iterated_loops n.+2 X)
```
Alizter: This change isn't needed.

mikeshulman: It's not needed on line 37 either, but I think it was included there for the reader's benefit, so I kept it here too.

mikeshulman:
Hmm... I didn't think too hard about that before.  My instinct right now is that trying to be "too clever" in a way like that will bite us later on more than it's worth.  Also usually the unbundled approach seems to be better when it's possible.  Why did you make that change?

I wonder whether @fpvandoorn has any opinions.

Alizter:
> Why did you make that change?
It let's you write stuff like `GroupIsomorphism (Pi n X) (Pi n Y)` as opposed to `GroupIsomorphism (Build_Group (Pi n X) _) ...`.

I think we might be able to have the best of both worlds. If we define `Pi n X` to be the truncated loop space and then define `HomotopyGroup n X` to be the group.

mikeshulman:
Is there a problem with staying unbundled for the morphisms too?

Alizter:
>Is there a problem with staying unbundled for the morphisms too?

I agree that being a homomorphism is a mere proposition so there is no harm in having a typeclass instance floating around. However when we talk about isomorphisms in group theory, very few people (that I suppose I have come in contact with) say "A bijection which happens to be a homomorphism" they just go ahead and say "isomorphism".

I recently managed to prove that under univalence, group isomorphisms are equivalent to paths of groups (which isn't that easy to do, most of the proof is rearranging sigma types).

I think it's the same issue with equivalences and a function being an equivalence. We use both all over the place and it seems to work fine. There is something about packing up data that makes it easier to work with and reason abstractly.

In conclusion, I cannot say if unbundled homomorphisms are better, but the answer is probably similar to why we use bundled equivalences rather than having isequiv's float around.

mikeshulman:
The way people talk informally about mathematics is often a very poor guide to the best way to formalize it in a proof assistant.  However, the point about `Equiv` is a reasonable one, I can imagine we may want to chain together group isomorphisms in a similar way.

mikeshulman:
By adding a uniform coercion from `HomotopyGroup_type` to `pType` and similarly for the functoriality, I was able to merge this on top of #1137 without too much change.

mikeshulman:
I did simplify the names in `HomotopyGroup.v` back to `pi` rather than `homotopygroup` everywhere, although we didn't discuss that yet; is there a good reason to be more verbose?

Alizter:
>is there a good reason to be more verbose?

I think I wanted to eventually put the Pi notation in a scope so that it would read as `HomotopyGroup` outside of the scope, but can be written as `Pi` for ease. My worry was that `Pi` is just too uninformative of a name. It could be confused with pi-types however unlikely.

mikeshulman:
I would say, let's cross that bridge if we come to it.  We never refer to pi-types in this library with "pi", we always say `forall` to match Coq's notation.  Since synthetic homotopy theory is a big part of this library, and in any case the notation `Pi` will only appear when people `Import Homotopy.HomotopyGroup`, I'm not too worried.

Alizter:
OK I am happy with this, obviously I think we should wait till #1137 has been merged. In which case I will double check this PR. I should be able to finish a proof of `B (Pi 1 X) <~> Tr 1 X` soon. In which case I will start working on free groups. Since I am using some ooGroup stuff it might be a good time to include the Lagrange stuff you mentioned before.

Thank you very much for taking a look at this!

Alizter:
@andreaslyn @spitters @JasonGross anyone else able to review?

fpvandoorn:
> I wonder whether @fpvandoorn has any opinions.

The question is whether algebraic structures should be bundled or unbundled. Has the `HoTT/HoTT` library already made a decision on that? 

In Lean 2 HoTT we chose the bundled approach
https://github.com/leanprover/lean2/blob/master/hott/algebra/homotopy_group.hlean
```lean
  definition homotopy_group [reducible] [constructor] (n : ℕ) (A : Type*) : Set* :=
  ptrunc 0 (Ω[n] A)

  notation `π[`:95  n:0 `]`:0 := homotopy_group n

  definition ghomotopy_group [constructor] (n : ℕ) [is_succ n] (A : Type*) : Group :=
  gtrunc (Ωg[n] A)

  definition aghomotopy_group [constructor] (n : ℕ) [is_at_least_two n] (A : Type*) : AbGroup :=
  agtrunc (Ωag[n] A)

  notation `πg[`:95  n:0 `]`:0 := ghomotopy_group n
  notation `πag[`:95 n:0 `]`:0 := aghomotopy_group n
```
Notice the type for the group (and abelian group) version of homotopy groups. Instead of saying that `pi (n+1) A` is a group, I say that `pi n A` is a group if `n` has a `is_succ` instance, which also allows us to write, e.g. `pi (1+n) A` or `pi ((n+1)+(m+1)) A` as an abelian group.

In Lean 3 we are using unbundled algebraic structures. However, we still use bundled homomorphisms and isomorphisms. The reason is that Lean's type-class inference is not good enough doing type class inference on functions. For example, if `f` and `g` are equivalences, it would be nice if type-class inference can figure out that `fun x => f (g x)` is an equivalence (which I think it was able to). However, the (or at least, a) problem was that when trying to figure out that `h` is an equivalence, it also tried to view this as a composition of `(fun x => x)` and `h`, and loop by applying the composition instance over and over again.

Making the type of `pi n X` depend on `n` is interesting, but looks pretty dangerous. I tried to do that in Lean 2 at some point (when doing the LES of homotopy groups), but turned away from that quickly. I do not remember exactly what the issues were, but it didn't seem to work very well when I tried it.

mikeshulman:
Thanks @fpvandoorn!  At the moment we're using hottclasses which generally treats algebraic structures as unbundled, although sometimes it seems we have to bundle them up, perhaps for similar reasons to those you mention.  I can't help feeling that the need to choose between bundled and unbundled is a misfeature of the current generation of proof assistants; there ought to be a clean way to deal with this sort of thing by designing a proof assistant with it in mind.

Since what we've got now is working so far, we may as well leave it for now, but keep in mind that we may have to change it as we develop more theory of homotopy groups.

JasonGross:
Note that the category theory portion of the library uses bundled structures, and this has been necessary for performance reasons.

Alizter:
I'll also note that I have erroneously defined some structures as typeclasses, such as Group. After more thought there really isn't any reason these should be typeclasses. I will therefore change them to Records soon.

spitters:
There's some initial work on going more freely between bundled and
unbundled.
( The first one has been implemented in agda/elisp)

https://alhassy.github.io/next-700-module-systems-proposal/thesis-proposal.pdf
http://www.ii.uib.no/~bezem/abstracts/TYPES_2019_paper_51

mikeshulman:
@spitters The first link is broken.

spitters:
Apparently, it moved here:
https://alhassy.github.io/next-700-module-systems/papers/gpce19_a_language_feature_to_unbundle_data_at_will.pdf

spitters:
BTW. Now that we have convenient tactics to convert back-and-forth to Sigma types, I wonder whether they can be used to convert between bundled and unbundled records.

They claimed to have a beautiful solution for this issue in Ahrend, but I don't fully understand it.
https://arend-lang.github.io/about/arend-features#class-system

mikeshulman:
Thanks, the paper looks interesting although I don't completely understand it.  It's not clear to me from the Arend documentation that they have anything new here.

I opened a new issue #1165 for talking about bundling.

PR #1146, opened 2019-12-03
Group iso is equivalent to path of groups

Alizter:
Here we prove that under univalence, the type of group isomorphisms between two groups G and H are equivalent to the path type between groups G and H.

I have also characterized "DPaths of sigma types as sigma types of DPaths".

The proof is long because it requires unpacking various parts of a largely nested sigma type and rearranging data to strip down the equivalence. No doubt there is a slicker method to do this.

### DIFF
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -70,13 +70,13 @@ Class GroupHomomorphism (G H : Group) := Build_GroupHomomorphism' {
 Coercion grp_homo_map : GroupHomomorphism >-> Funclass.
 (* Notation "G '->G' H" := (GroupHomomorphism G H) (at level 20). *)
 
-Definition issig_GroupHomomorphism {G H : Group} : _ <~> GroupHomomorphism G H
+Definition issig_GroupHomomorphism (G H : Group) : _ <~> GroupHomomorphism G H
   := ltac:(issig).
 
 Definition equiv_path_grouphomomorphism {F : Funext} {G H : Group}
   {g h : GroupHomomorphism G H} : g == h <~> g = h.
 Proof.
-  refine ((equiv_ap issig_GroupHomomorphism^-1 _ _)^-1 oE _).
+  refine ((equiv_ap (issig_GroupHomomorphism G H)^-1 _ _)^-1 oE _).
   refine (equiv_path_sigma_hprop _ _ oE _).
   apply equiv_path_forall.
 Defined.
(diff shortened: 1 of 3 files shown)


mikeshulman:
Were you unable to use the tools in `PathAny.v`?  They're designed for right-nested sigma-types, but after rearranging sigmas I would expect that they should work.

Alizter:
@mikeshulman I never attempted using PathAny. I will try and see if I can use it later this week.

Alizter:
@mikeshulman I think I am being a bit slow, but I fail to see how PathAny is supposed to help here. After rearranging the outer data in the sigma types so that they match identically, then using equiv_functor_sigma_id, I get left with showing:
```coq
1 subgoal
U : Univalence
G, H : Group
f : G -> H
e : IsEquiv f
IsMonoidPreserving f <~>
{n
: DPath SgOp (path_universe_uncurried {| equiv_fun := f; equiv_isequiv := e |})
    group_sgop group_sgop &
DPath (fun a : {x : _ & SgOp x} => {_ : MonUnit (pr1 a) & Negate (pr1 a)})
  (path_sigma_dp (path_universe_uncurried {| equiv_fun := f; equiv_isequiv := e |}) n)
  (group_unit; group_inverse) (group_unit; group_inverse)}
```

Both sides are hprops but the maps going each way are quite non-trivial.

If I had exactly the same data in both sides but just muddled up, I would agree that PathAny would be able to prove it.

Am I not understanding the role of PathAny? Can you be a bit more specific about what you mean?

mikeshulman:
I sent you a PR.

Alizter:
I like the PathAny tactic but it takes some getting used to.

theories/Algebra/Group.v
```diff
@@ -1,5 +1,6 @@
 Require Import Basics.
 Require Import Types.
+Require Import Cubical.
```
mikeshulman: Probably not needed any more.

mikeshulman:
Yes, it takes some practice to apply correctly.

theories/Algebra/Group.v
Alizter: Well I did a sneaky and used path_sigma_dp in my latest commit, mostly because I’m biased. I don't think it harms anything by keeping it.

mikeshulman: Hmm. I think I would prefer to stick closer to the style of the book when there isn't a clear benefit from using cubical technology, to keep things more readable.

Alizter: Fair enough, I will change it back.

Alizter:

spitters:
@andreaslyn has a general version of this result based on universal algebra.

mikeshulman:
Where is it?

andreaslyn:
> Where is it?

https://github.com/HoTT/HoTT/blob/bbf912e260fa1db39dd934e6b868fa64b8a474f7/theories/Classes/theory/ua_isomorphic.v#L161

mikeshulman:
Thanks!  (I just finally figured out that `ua_` on these files means Universal Algebra -- I'm used to it meaning Univalence Axiom!)

So the question is, would it be less work to obtain this result for groups by identifying them as particular universal algebras?  We would need to show that the type `Group` is equivalent to a suitable instance of the type `Algebra`, and similarly for homomorphisms and isomorphisms, right?  Has that been done?

andreaslyn:
It would be less work to identify group with the corresponding algebra and obtain the result, but I think it is fine to merge this for now, since the universal algebra machinery is not quite ready.

I am working on a branch adding support for equational laws. When this is done, by identifying group with the corresponding algebra, we will additionally obtain the three isomorphism theorems, quotient group, direct product, and subgroup.

Alizter:
@andreaslyn I look forward to it!

PR #1147, opened 2019-12-04
H-spaces, closes #984

### LINKED ISSUE #984: Proving Lemma 8.5.5

Alizter:
I have added the definition of H-space and the equivalences induced when the space is 0-connected.

This closes #984.

### DIFF
--- a/theories/Homotopy/HSpace.v
+++ b/theories/Homotopy/HSpace.v
@@ -5,6 +5,7 @@ Require Export Classes.interfaces.abstract_algebra.
 Import TrM.
 
 Local Open Scope trunc_scope.
+Local Open Scope mc_mult_scope.
 
 Class IsHSpace (X : pType) := {
   hspace_op :> SgOp X;
@@ -25,7 +26,7 @@ Section HSpaceProperties.
    `{IsConnected 0 A}.
 
   Global Instance isequiv_hspace_left_op
-    : forall (a : A), IsEquiv (a &).
+    : forall (a : A), IsEquiv (fun x => a * x).
   Proof.
     refine (conn_map_elim (-1) (unit_name hspace_id) _ _).
     + exact (conn_point_incl hspace_id).
@@ -36,7 +37,7 @@ Section HSpaceProperties.
   Defined.
 
   Global Instance isequiv_hspace_right_op
-    : forall (a : A), IsEquiv (& a).
+    : forall (a : A), IsEquiv (fun x => x * a).
   Proof.
     refine (conn_map_elim (-1) (unit_name hspace_id) _ _).
     + exact (conn_point_incl hspace_id).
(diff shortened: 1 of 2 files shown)


theories/Homotopy/HSpace.v
```diff
@@ -0,0 +1,55 @@
+Require Import Basics.
+Require Import Types.
+Require Import Truncations.
+Require Export Classes.interfaces.abstract_algebra.
+Import TrM.
+
+Local Open Scope trunc_scope.
+
+Class HSpace (X : pType) := {
```
mikeshulman: Should this be `IsHSpace`?

Alizter: I think it is a bit awkward like IsIdempotent, since this isn't a proposition. I'm probably ok with that change however.

mikeshulman: Oh... I was thinking by analogy with `IsGroup`, but I forgot that the actual data is contained in other typeclasses and that one just states the hprop axioms.

However, if we ever wanted to talk about the type of h-spaces, it would be natural to call that `HSpace` (just like with the type `Group` of groups), in which case the structure of being an h-space would have to be called something different.  So I think I still favor `IsHSpace`.

Alizter: Yes this seems reasonable. As I've said before, we would then expect `HSpace <~> (Smash X X -> X)` or something.

Alizter: Actually wait I think I disagree. Why would we ever want to talk about the type of H-spaces? I have only ever seen collections of H-spaces on some specific object. I think that's why I didn't use the Is.

Alizter: Actually wait, ignore what I just said. I am confusing myself and probably you too. I think you are right with IsHSpace.

Alizter: OK I am fine with changing it to IsHSpace

Alizter:
OK I have rebased and carried out Mike's suggestion.

Alizter:
Rebased once again for #1155

Alizter:

PR #1148, opened 2019-12-04
Classifying spaces

Alizter:
This relies on #1147, therefore only check the last commit for review.

Changes:

 * Added (1-truncated) classifying space of type with operation HIT using private inductive types.
 * Defined classifying space of a group.
 * Showed that loop space of BG is equivalent to G.
 * Some extra properties of abelianizations

Eventually we should replace this private inductive type and construct BG out of coequalizers.

### DIFF
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -37,10 +37,10 @@ Coercion group_abgroup : AbGroup >-> Group.
 (** Definition of Abelianization.
 
   Given a map F that turns any group into an abelian group, and a unit homomorphism eta_X : X -> F X. This data is considered an Abelianization if and only if for all maps X -> A, there exists a unique g such that h == g o eta X. *)
-Definition IsAbelianization (F : Group -> AbGroup)
-  (eta : forall X, GroupHomomorphism X (F X))
-  := forall (X : Group) (A : AbGroup) (h : GroupHomomorphism X A),
-    Contr (exists (g : GroupHomomorphism (F X) A), h == g o eta X).
+Definition IsAbelianization {G : Group} (G_ab : AbGroup)
+  (eta : GroupHomomorphism G G_ab)
+  := forall (A : AbGroup) (h : GroupHomomorphism G A),
+    Contr (exists (g : GroupHomomorphism G_ab A), h == g o eta).
 
 Existing Class IsAbelianization.
 
@@ -317,10 +317,10 @@ Section Abelianization.
   Defined.
 
   (** Finally we can prove that our construction abel is an abelianization. *)
-  Global Instance isabelianization_abel
-    : IsAbelianization abel abel_unit.
+  Global Instance isabelianization_abel {G : Group}
+    : IsAbelianization (abel G) (abel_unit G).
   Proof.
-    intros X A h.
+    intros A h.
     serapply Build_Contr.
     { srefine (_;_).
       { simple notypeclasses refine (Build_GroupHomomorphism _).
@@ -343,19 +343,19 @@ Section Abelianization.
 
 End Abelianization.
 
-Theorem groupiso_isabelianization
-  (F1 F2 : Group -> AbGroup)
-  (eta1 : forall X, GroupHomomorphism X (F1 X))
-  (eta2 : forall X, GroupHomomorphism X (F2 X))
-  {x : IsAbelianization F1 eta1}
-  {y : IsAbelianization F2 eta2}
-  (G : Group) : GroupIsomorphism (F1 G) (F2 G).
+Theorem groupiso_isabelianization {G : Group}
+  (A B : AbGroup)
+  (eta1 : GroupHomomorphism G A)
+  (eta2 : GroupHomomorphism G B)
+  {x : IsAbelianization A eta1}
+  {y : IsAbelianization B eta2}
+  : GroupIsomorphism A B.
 Proof.
   unfold IsAbelianization in x, y.
-  destruct (x G (F2 G) (eta2 G)) as [[a ah] ac].
-  destruct (y G (F1 G) (eta1 G)) as [[b bh] bc].
-  destruct (x G (F1 G) (eta1 G)) as [[c ch] cc].
-  destruct (y G (F2 G) (eta2 G)) as [[d dh] dc].
+  destruct (x B eta2) as [[a ah] ac].
+  destruct (y A eta1) as [[b bh] bc].
+  destruct (x A eta1) as [[c ch] cc].
+  destruct (y B eta2) as [[d dh] dc].
   serapply (Build_GroupIsomorphism _ _ a).
   serapply (isequiv_adjointify _ b).
   { apply ap10.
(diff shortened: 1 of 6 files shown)


theories/Algebra/AbelianGroup.v
```diff
@@ -339,4 +339,106 @@ Section Abelianization.
...
+  { apply ap10.
+    change (@grp_homo_map _ _ (grp_homo_compose b a)
+      = @grp_homo_map _ _ grp_homo_id).
+    refine (ap (@grp_homo_map _ _) _).
+    refine (ap pr1 ((cc (_; _))^ @ cc (grp_homo_id; _))).
+    1: exact (fun i => bh i @ ap b (ah i)).
+    reflexivity. }
+Defined.
+
+Global Instance isequiv_groupiso_isabelianization
```
mikeshulman: Why do you need to declare this separately, when typeclass inference can find it?

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,337 @@
...
+  We implement this is a private inductive type.
+*)
+Module Export ClassifyingSpace.
+
+  Section ClassifyingSpace.
+
+    Private Inductive ClassifyingSpace {X : Type} `{SgOp X} :=
+      | bbase : ClassifyingSpace.
+
+    Definition bloop {X : Type} `{SgOp X} : X -> bbase = bbase.
```
mikeshulman: Is there a reason to use `Definition...Admitted` instead of `Axiom` like we do for other HITs?

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,337 @@
...
+    Definition ClassifyingSpace_ind
+      (P : ClassifyingSpace G -> Type)
+     `{forall x, IsTrunc 1 (P x)}
+      (bbase' : P bbase)
+      (bloop' : forall x, DPath P (bloop x) bbase' bbase')
+      (bloop_pp' : forall x y,  DSquare P (sq_G1 (bloop_pp x y))
+        (bloop' (x & y)) ((bloop' x) @D (bloop' y)) 1 1) x : P x
+      := match x return (_ -> P x) with
+            bbase => (fun _ => bbase')
+         end bloop.
```
mikeshulman: This should be the hypothesis `bloop'`, not the constructor `bloop`.  Also, I think you need to similarly "trivially use" the other path-hypothesis `bloop_pp'`.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,337 @@
...
+    Definition ClassifyingSpace_ind_beta_bloop
+      (P : ClassifyingSpace G -> Type)
+     `{forall x, IsTrunc 1 (P x)}
+      (bbase' : P bbase) (bloop' : forall x, DPath P (bloop x) bbase' bbase')
+      (bloop_pp' : forall x y,  DSquare P (sq_G1 (bloop_pp x y))
+        (bloop' (x & y)) ((bloop' x) @D (bloop' y)) 1 1) (x : G)
+      : dp_apD (ClassifyingSpace_ind P bbase' bloop' bloop_pp') (bloop x)
+        = bloop' x.
+    Proof. Admitted.
+
```
mikeshulman: The private inductive module should end here, since the definition of the non-dependent eliminator doesn't require looking inside the dependent one.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,337 @@
...
+  1: reflexivity.
+  intro x.
+  apply dp_paths_FlFr.
+  apply path_ishprop.
+Defined.
+
+(** Now we focus on the classifying space of a group. *)
+
+(** The classifying space of a group is the following pointed type. *)
+Definition B (G : Group) := Build_pType (ClassifyingSpace G) bbase.
```
mikeshulman: I would put this in a notation scope or module, since the single letter `B` may be used for many different things.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,337 @@
...
+
+(** We can show that [bloop] takes the unit of the group to reflexivity. *)
```
mikeshulman: For `bloop_id` and `bloop_inv`, wouldn't it be cleaner to just observe that `bloop` is a semigroup homomorphism (essentially by definition) and then use `Build_GroupHomomorphism` with `grp_homo_unit` and `grp_homo_inv`?

theories/Homotopy/ClassifyingSpace.v
Alizter: I can make it more immediately clear which parts are implicit.

theories/Homotopy/ClassifyingSpace.v
Alizter: Good point.

Alizter:
I think I made some poor design choices here. This was written before I bundled up groups, so it has been poorly adapted from a time when it had to pick IsGroup from out of the air.

I am not really happy with the generality of the HIT, it doesn't make much sense to me. I think it would be better to stick to a group. It's true that monoids have "classifying spaces" but there is really a group completion being done in the background.

I think it would be cleaner if ClassifyingSpace just took in a Group and gave a pType. Then B can be a notation for it which is scoped.

Also I am not really sure what is going on with the private inductive type anymore. I wrote it a few months ago and (I think?) it seemed okay then. Now I am rereading it I don't understand it at all. And this doesn't bode well for literally anybody else.

 How might we go about writing BG as a bunch of (truncated) coequalizers?

Alizter:
What about something like this:

The 1-truncated coequalizer of two maps G x G -> loop susp (G + 1). Where the first map is given by `fun x y => merid (inl (x & y)) @ (merid (inr tt))^` and the second by `fun x y => merid (inl x) @ (merid (inr tt))^ @ merid inl y @ (merid (inr tt))^`.

This should work right?

mikeshulman:
Hmm, that looks to me like the G's are going in the wrong dimension.  You need the adjunct susp(G x G) --> susp (G+1).  But I'm not sure that's right either, there might be an off-by-one error.  The systematic hub-and-spoke decomposition would, I think, be something like the pushout of G x G <-- G x G x S^1 --> Susp (G+1).

theories/Homotopy/ClassifyingSpace.v
Alizter: Hmm, I remember seeing Admitted in the definition of truncation, but now looking at it again this was so we could have the instance.

I won't bother changing it however unless it really gets treated differently than axiom.

Alizter:
@mikeshulman IsAbelianization is really a property of the group G_ab and the map from G to G_ab. To get what you want, you could construct the abelianization G_ab any way you wish and then write `IsAbelianization G_ab` to get the property you want.

The advantage of this definition is that it doesn't choose any particular definition of abelianization.

theories/Algebra/AbelianGroup.v
Alizter: Not really sure, will get rid of.

Alizter:
@mikeshulman It occurred to me that Floris has a whole section in his thesis constructing 2-HITs (as he calls them) using the hub-and-spoke method. I have started formalising some of his idea so perhaps we can write down non-recursive 2-HITs in an easier way in the future.

Alizter:
OK I had a good look at Floris' formalization of 2-HITs and it looks like we can write down BG and Rezk completion via his general schemata. It only uses quotients (so in our case coequalizers).

Formalizing this will take longer than I am willing to let this wait. So I propose we do the following:

 1. Clean up classifying space and get it in working order using the private inductive type.
 2. Develop a theory of general 2-HITs like Floris does in his lean formalization, and then redefine BG to use this.

This will allow for a nice account of the hub-and-spoke method too.

mikeshulman:
Sounds fine.  The changes to make the private inductive here work should be fairly minimal.

theories/Algebra/AbelianGroup.v
Alizter: @mikeshulman It turns out it can't find it.

theories/Homotopy/ClassifyingSpace.v
Alizter: OK fixed.

theories/Algebra/AbelianGroup.v
Alizter: I'm not sure why typeclass search is dumb here so I won't change it. I don't think it does any harm being around.

mikeshulman: Where is it that Coq can't find it?  Clearly it can find it _here_ because of how you are able to prove the instance?

Alizter: So in the proof of `issurj_isabelianization` (ln 404) I apply this lemma directly. If I replace this with an `exact _`. Coq cannot find it.

theories/Homotopy/ClassifyingSpace.v
Alizter: OK, change B to B_group and added a notation in bg_scope that let's you write `B G`.

theories/Homotopy/ClassifyingSpace.v
Alizter: Done.

theories/Homotopy/ClassifyingSpace.v
Alizter: Done

theories/Algebra/AbelianGroup.v
mikeshulman: Well, I'm not quite sure what's going on, but if I remove the `simpl in p` line from `issurj_isabelianization`, the `refine` then finishes the whole proof.

mikeshulman:
Shouldn't `Homotopy.ClassifyingSpace` also be exported by `HoTT.v`?

theories/Algebra/AbelianGroup.v
Alizter: Huh, that's weird. Thanks for finding that.

theories/Homotopy/ClassifyingSpace.v
mikeshulman: What about a more informative name like maybe `pClassifyingSpace`?

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,350 @@
...
+    Definition ClassifyingSpace_ind
+      (P : ClassifyingSpace G -> Type)
+     `{forall x, IsTrunc 1 (P x)}
+      (bbase' : P bbase)
+      (bloop' : forall x, DPath P (bloop x) bbase' bbase')
+      (bloop_pp' : forall x y,  DSquare P (sq_G1 (bloop_pp x y))
+        (bloop' (x & y)) ((bloop' x) @D (bloop' y)) 1 1) x : P x
+      := match x return (_ -> P x) with
+            bbase => (fun _ => bbase')
+         end bloop'.
```
mikeshulman: This still needs to be applied to `bloop_pp'` as well.

Alizter:
OK I have completely removed the SgOp stuff, and redefined classifying space to work only on groups.

theories/Homotopy/ClassifyingSpace.v
Alizter: I have absolutely no idea how I wrote this down at the time. Do you mean something like this:
```coq
match x return (_ -> _ -> P x) with
            bbase => (fun _ _ => bbase')
         end bloop' bloop_pp'.
```

mikeshulman: Yes, something like that.

Alizter: Or better yet without the return.

Alizter:
@mikeshulman Also I updated HoTT.v

Alizter:
@mikeshulman I also changed the definition of IsAbelianization to the one you suggested. Now I can't get over how backwards the previous one was. This one is much cleaner. Thanks for the suggestion.

Alizter:
Oh dear. Putting ClassifyingSpace in HoTT.v causes the & notation to be imported everywhere breaking all our tests using sigma &. I think we will have to wait for #1155 .

Alizter:
For some reason the notation "'B' G" in ClassifyingSpace.v is breaking HoTTBookExercises. I don't know why though, it's scope is closed...

mikeshulman:
I believe that once something is declared as a Notation, it affects the _parser_ even when its scope is closed, because parsing happens before scopes are evaluated.  That's one reason for declaring all Notations in a single file (e.g. `Basics/Notation.v`).  One might argue that this is a bug in Coq, but I don't think it's going away any time soon.  The workaround is to be very very careful about using alphanumeric identifiers as part of Notations.  In this case it might work to instead do something like

```
Module Import ClassifyingSpaceNotation.
  Definition B G := pClassifingSpace G.
End ClassifyingSpaceNotation.
```

and then clients who want the `B` notation can say `Import ClassifyingSpaceNotation` instead of `Open Scope bg_scope`.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -0,0 +1,349 @@
...
+  apply path_ishprop.
+Defined.
+
+(** Now we focus on the classifying space of a group. *)
+
+(** The classifying space of a group is the following pointed type. *)
+Definition pClassifingSpace (G : Group)
+  := Build_pType (ClassifyingSpace G) bbase.
+
+Notation "'B' G" := (pClassifingSpace G) (at level 0) : bg_scope.
```
SkySkimmer: Why not
```suggestion
Notation B G := (pClassifingSpace G) (at level 0) : bg_scope.
```
which doesn't modify the parser? https://coq.github.io/doc/master/refman/user-extensions/syntax-extensions.html#abbreviations

Alizter: That's a good point, thanks for that!

Alizter: Hmm it seems I cannot put abbreviations in scopes however.

mikeshulman:
What's the status of this?  Is it blocking on the notation issue?  Did you try the module suggestion?

Alizter:
@mikeshulman I'll try the module suggestion in a bit.

Alizter:
Oh and #1147 needs to be merged before this.

Alizter:
rebased

Alizter:
@mikeshulman I tried the module suggestion. Are you able to review again?

mikeshulman:
Yes, LGTM.

PR #1149, opened 2019-12-04
Induction principle for Join

Alizter:
I have added an induction principle for Join including it's computation rule. This make it easier to work the the join.


mikeshulman:
LGTM.  But surely this doesn't actually depend on #1147?

Alizter:
I completely failed to rebase all of these branches properly and in my ignorance thought they were fine. Will fix tomorrow 0:-)

Alizter:
OK rebased.

PR #1150, opened 2019-12-04
Diamond lemmas for join and join bifunctor

Alizter:
This relies on #1149, check the last two commits to review.

In this PR we have the join bifunctor defined in the obvious way. We also have some "diamond lemmas". These are lemmas about squares whose edges comprise of jglues. These lemmas will be used in the Cayley-Dickson construction to come.

### DIFF
--- a/theories/Homotopy/Join.v
+++ b/theories/Homotopy/Join.v
@@ -103,7 +103,7 @@ Section Join.
     apply sq_refl_h.
   Defined.
 
-  Definition join_bifunctor {A B C D} (f : A -> C) (g : B -> D)
+  Definition functor_join {A B C D} (f : A -> C) (g : B -> D)
     : Join A B -> Join C D.
   Proof.
     serapply Join_rec.
@@ -113,9 +113,9 @@ Section Join.
     apply jglue.
   Defined.
 
-  Definition join_bifunctor_compose {A B C D E F}
+  Definition functor_join_compose {A B C D E F}
     (f : A -> C) (g : B -> D) (h : C -> E) (i : D -> F)
-    : join_bifunctor (h o f) (i o g) == join_bifunctor h i o join_bifunctor f g.
+    : functor_join (h o f) (i o g) == functor_join h i o functor_join f g.
   Proof.
     serapply Join_ind.
     1,2: reflexivity.
@@ -129,8 +129,8 @@ Section Join.
     reflexivity.
   Defined.
 
-  Definition join_bifunctor_idmap {A}
-    : join_bifunctor idmap idmap == (idmap : Join A A -> Join A A).
+  Definition functor_join_idmap {A}
+    : functor_join idmap idmap == (idmap : Join A A -> Join A A).
   Proof.
     serapply Join_ind.
     1,2: reflexivity.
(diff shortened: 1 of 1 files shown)


theories/Homotopy/Join.v
```diff
@@ -87,6 +87,80 @@ Section Join.
...
+
+  Definition join_natsq_h {A B : Type} {a a' : A} {b b' : B}
+    (p : a = a') (q : b = b')
+    : Square (jglue a b) (jglue a' b') (ap joinl p) (ap joinr q).
+  Proof.
+    destruct p, q.
+    apply sq_refl_h.
+  Defined.
+
+  Definition join_bifunctor {A B C D} (f : A -> C) (g : B -> D)
```
mikeshulman: I think the consistent name for this would be `functor_join` (see for instance `functor_prod`, `functor_sum`).  It's assumed that a functor is of the appropriate number of arguments.

mikeshulman:
Also, like #1149, this surely doesn't depend on #1147.

theories/Homotopy/Join.v
Alizter: What I did in my smash coherences branch was name the functorial action `X_functor` and the functor itself (which is a record) `functor_X`. I am fine with renaming this to `join_functor` but if we ever define the packaged up version of Bifunctor, then I think it would make sense to write `bifunctor_join` for that.

Observe that in Pointed, loops and psusp have functor come after. Same with ReflectiveSubuniverse, in which it would be very weird to write functor_O.

So I would argue that functor_prod and functor_sum are the wrong names.

Alizter:
rebased.

theories/Homotopy/Join.v
mikeshulman: What do you mean by "packaged up version"?

You're right that we're already not consistent about this.  But I don't see any argument for putting the `functor` afterwards; I don't see anything weird about `functor_O`.  In general we put what something *is* first in its name, e.g. `equiv_*` for things that are equivalences rather than `*_equiv`.

Alizter: The "packaged up version" I was alluding to is something like a record called Functor which contains the action on objects, the action on morphisms, a proof of identity preservation and a proof of composition preservation. To me all this data together, constitutes a (1-coherent) functor. Then it deserves to be a functor, i.e. start with functor_. Otherwise it's not exactly that but just the action part etc.

mikeshulman: Meh, I think that's nitpicky.  We have much more use for the concrete action of functors on morphisms than for packaged up records of them.  Especially given that your record isn't actually a functor at all, only the 1-dimensional fragment of one.

Alizter:
@mikeshulman Are you able to restart that travis job that never got off the ground, for some reason I have lost my ability to restart travis jobs. Though I still have merge permissions on this repo...

theories/Homotopy/Join.v
Alizter: Fair enough. I will change it to functor_join. I will also create an issue highlighting our inconsistency in naming.

mikeshulman:
Apparently I am.  Are you signed into travis with your github account?

Alizter:
Signing out then in again has fixed it. Strange...

Alizter:
Rebased

mikeshulman:
I think a mere rebase, that doesn't make any substantial changes other than trivial ones to ensure compatibility with master, doesn't need to be re-reviewed.

PR #1151, opened 2019-12-04
Various facts about spheres and H-space on S1

Alizter:
This relies on #1150 and #1147. See the last 5 commits to review.

Here's what I've done:
* Show that `Sphere 1` is 1-truncated
* Show that `Sphere n.+1` is n-connected
* Stop spheres from unfolding, closes #1018 
* Corrected istrunc for DPath's, and added the default dpath between any families of hprops connected by a path in the base space.
* Shown that a square type in a n.+2 truncated type is n-truncated.
* Shown that S1 has a H-space structure which is associative and commutative.

PR #1152, opened 2019-12-04
Cayley-Dickson (from Rijke-Buchholtz)

Alizter:
This relies on #1150, check the last 2 commits to review.

Here is the Cayley-Dickson construction as described in Rijke-Buchholtz. This gives a H-space on `Join (Sphere 1) (Sphere 1)` when combined with #1151, but we need join associativity to prove that this is `S3`. We could probably prove join associativity directly, there is an "easier" proof with the 3x3 lemma but we don't have this yet.

I'll come back and add some stuff to this later, such as moving some type classes to canonical_names.v and adding the S3 H-space exercise solution to HoTTBookExcersises.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -271,6 +271,7 @@ theories/Homotopy/BlakersMassey.v
 theories/Homotopy/Freudenthal.v
 theories/Homotopy/ClassifyingSpace.v
 theories/Homotopy/EMSpace.v
+theories/Homotopy/CayleyDickson.v
 
 #
 #   Pointed
--- a/theories/Classes/interfaces/canonical_names.v
+++ b/theories/Classes/interfaces/canonical_names.v
@@ -290,6 +290,20 @@ Proof.
   exact (trunc_equiv (SigEquivRel R) (issig_equiv_rel R)).
 Qed.
 
+Class Conjugate A := conj : A -> A.
+
+Class DistrOpp {A} `(SgOp A) `(Conjugate A)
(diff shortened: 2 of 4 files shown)


mikeshulman:
LGTM.  I also agree with moving those typeclasses into `canonical_names`.

FWIW, I feel like the fact that you can write down the definition of a Cayley-Dickson spheroid like that is a validation of the unbundled approach and the hottclasses design more generally.  Wouldn't it be much more annoying to make that definition with bundled structures?

Alizter:
OK, I've rebased and fixed some issues. I am however not finished with this file. There is still some more clean up needing to be done and perhaps a better notation for conjugation needs to be found.

A H-space structure on S3 can be constructed with this but it needs associativity of join.

Alizter:
Regarding unbundled structures, I think we should use them as much as possible. But there are some times when they are needed, such as the category of groups.

mikeshulman:
Is this PR ready to be reviewed/merged?

Alizter:
Not just yet. I would have made this PR a draft if I had known it was possible at the time. I just need to reorganise some stuff and give some lemmas some better names. I haven't had the time to do this yet.

Alizter:
@mikeshulman This is ready now, feel free to merge when travis is done.

mikeshulman:
Since you just recently changed some stuff, let's wait the 48 hours in case anyone else wants to look at it.

PR #1155, opened 2019-12-06
Replace & for sg_op with scopes for * and +, close #976

### LINKED ISSUE #976: Don't clobber & for sgop
HoTT.Classes.interfaces.canonical_names defines `&` to mean `sgop`, which breaks the standard notation `{ x:A & P x }` for Sigma-types.  Let's find another symbol for `sgop`.

mikeshulman:
My suggestion at https://github.com/HoTT/HoTT/issues/976#issuecomment-561823453 seems to work, except for the notation `(.* x)`.  But I'm not a real big fan of that notation anyway, so I don't mind using lambdas instead.

### DIFF
--- a/theories/Classes/theory/groups.v
+++ b/theories/Classes/theory/groups.v
@@ -1,7 +1,7 @@
 Require Import
   HoTT.Classes.interfaces.abstract_algebra.
 
-Local Open Scope mc_add_scope.
+Local Open Scope mc_mult_scope.
 
 Section group_props.
   Context `{IsGroup G}.
@@ -10,14 +10,14 @@ Section group_props.
   Global Instance negate_involutive : Involutive (-).
   Proof.
     intros x.
-    transitivity (mon_unit + x).
+    transitivity (mon_unit * x).
     2: apply left_identity.
-    transitivity ((- - x + - x) + x).
-    2: apply (@ap _ _ (fun y => y + x)),
+    transitivity ((- - x * - x) * x).
+    2: apply (@ap _ _ (fun y => y * x)),
           left_inverse.
-    transitivity (- - x + (- x + x)).
+    transitivity (- - x * (- x * x)).
     2: apply associativity.
-    transitivity (- - x + mon_unit).
+    transitivity (- - x * mon_unit).
     2: apply ap, symmetry, left_inverse.
     apply symmetry, right_identity.
   Qed.
@@ -36,7 +36,7 @@ Section group_props.
     apply symmetry, right_identity.
   Qed.
 
-  Global Instance group_cancelL : forall z : G, LeftCancellation (+) z.
+  Global Instance group_cancelL : forall z : G, LeftCancellation (.*.) z.
   Proof.
     intros z x y E.
     rewrite <- (left_identity x).
@@ -47,7 +47,7 @@ Section group_props.
     reflexivity.
   Qed.
 
-  Global Instance group_cancelR: forall z : G, RightCancellation (+) z.
+  Global Instance group_cancelR: forall z : G, RightCancellation (.*.) z.
   Proof.
     intros z x y E.
     rewrite <-(right_identity x).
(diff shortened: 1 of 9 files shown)


Alizter:
LGTM, my Cayley-Dickson branch is going to need a new symbol for conjugation however. I don't think `a* * b*` is going to cut it.

mikeshulman:
Would `a^* * b^*` work?

Alizter:
@mikeshulman It might but I feel like it is getting too cluttered. I might be able to get coq to bracket up `(a*)` so it is easier to read, but can be written without brackets. i.e. `(a*) * (b*)`.

theories/Classes/theory/groups.v
```diff
@@ -1,21 +1,23 @@
...
+
 Section group_props.
   Context `{IsGroup G}.
 
   (** Group inverses are involutive *)
   Global Instance negate_involutive : Involutive (-).
   Proof.
     intros x.
-    transitivity (mon_unit & x).
+    transitivity (mon_unit + x).
```
Alizter: Shouldn't most of these be `*` instead of `+`.

mikeshulman: Well, it's an interesting question which symbol to use for one abstract group in isolation.  True, `+` has undesired connotations of commutativity, but if `-` is the notation for inverses then it makes more sense to write `- x + x = 0` than `- x * x = 0`.  Perhaps the notation for inverses should also be scoped?

Alizter: Yes, I think changing the multiplicative inverses to `^-1` or something would be sensible.

Alizter: We should also consider notations for mon_unit as you say.

mikeshulman: Will declaring notations for `0` and `1` conflict with the fancy behavior of general numeral notations for integers?

Alizter: It might cause issue if someone opens up the multiplicative operation scope after opening the int scope. This would mean that 1 would become 0.

However the group operation on Int is only considered in Algebra/Z so we could possibly force a closure of the multiplicative scope on import of Algebra/Z and force the additive scope to open.

Alizter:
If we were to define a ring, we would have two different sg_op's floating around. How would we ensure that one gets printed additively and the other multiplicatively?

theories/Classes/theory/groups.v
Alizter: Numeral notations will work fine with the additive notations, no matter which is opened first. Multiplicative notations will be problematic however, but it would depend on which scope was opened first. If int_scope then mc_mult_scope then 1 would become 0, if mc_mult_scope then int_scope, then 0 would become 1. I would advise against both practices, but I don't think we can restrict notation scopes from being opened.

mikeshulman:
I kept the typeclasses `Mult` and `Plus` that bind multiplicative and additive scope respectively in the global `mc_scope`, so rings work just the way they used to without opening either general multiplicative or general additive scope.

It just occurred to me that possibly another way to prove things about an arbitrary semigroup and choose to write its operation additively (but allow users of my library to apply it to semigroups written multiplicatively) would be to assume only `SgOp` in my context but define a *local* `Plus` instance from it.  I wonder whether that would be cleaner.

mikeshulman:
Well, I tried that other way and it doesn't seem to work.  I'm not sure why, but in proving compose instances for rings it causes the wrong operation to get picked up in the middle.

Alizter:
@mikeshulman That's a shame. I'm sure there is some way to get what you want but probably requires fiddling around. For now I think what you have done is fine, we can think about how we might deal with this in the future.

mikeshulman:
> I might be able to get coq to bracket up `(a*)` so it is easier to read

I would expect that to conflict with comment syntax `(* ... *)`.

theories/Classes/theory/groups.v
mikeshulman: For now, I changed this file to use `*`.

Alizter:
@mikeshulman Oh yes, you're right. I'm sure I'll find something eventually.

PR #1158, opened 2019-12-07
Get rid of Global Generalizable Variables

mikeshulman:
Setting `Global Generalizable Variables` in `canonical_names` means that any file anywhere in the library that uses hottclasses implicitly inherits those generalizable variables, perhaps to the surprise of the user.  I think we should be less lazy and explicitly indicate in each file whether we want generalizable variables.

### DIFF
--- a/theories/Algebra/Congruence.v
+++ b/theories/Algebra/Congruence.v
@@ -7,6 +7,6 @@ Require Import Classes.interfaces.abstract_algebra.
 
 Local Open Scope mc_mult_scope.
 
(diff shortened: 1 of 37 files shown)


Alizter:
LGTM, I must confess I kind of liked being lazy with this. But I definitely think it is a bad kind of laziness.

mikeshulman:
Rebased.  APOE?  There are a lot of little changes but I think this should be quick to review.

spitters:
If you want to speed up the process during the next week, just go ahead
with the two of you without with for me.

from my phone

PR #1161, opened 2019-12-07
Eilenberg-MacLane spaces

Alizter:
This relies on #1148. See the last commit here.

I couldn't in the end make the argument slicker as suggested in #1086 since I got really confused what was going on. Instead I have opted to put all the important lemmas in EMSpace.v. Currently it works and no doubt there is a nicer way to do this argument but for what I want to do with it it is fine.

### DIFF
--- a/theories/Homotopy/EMSpace.v
+++ b/theories/Homotopy/EMSpace.v
@@ -1,0 +1,278 @@
+Require Import Basics.
+Require Import Types.
+Require Import Pointed.
+Require Import Cubical.DPath.
+Require Import Algebra.Group.
+Require Import Algebra.AbelianGroup.
+Require Import Truncations.
+Require Import Homotopy.Suspension.
+Require Import Homotopy.ClassifyingSpace.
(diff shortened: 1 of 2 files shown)


theories/Homotopy/EMSpace.v
```diff
@@ -0,0 +1,277 @@
...
+  Context `{Univalence}.
+
+  Fixpoint EilenbergMacLane (G : Group) (n : nat) : pType
+    := match n with
+        | 0    => Build_pType G _
+        | 1    => B G
+        | m.+1 => pTr m.+1 (psusp (EilenbergMacLane G m))
+       end.
+
+  Notation "'K(' G , n )" := (EilenbergMacLane G n).
```
mikeshulman: I'm worried that this will give rise to the same kind of parsing issues as `B G`.

Alizter: Writing `K` followed straight by a bracket seems to be a very uncommon thing to do. I think this one is safe.

PR #1169, opened 2019-12-28
Reduce approval requirements in STYLE.md, close #1166

### LINKED ISSUE #1166: A dearth of pairs of eyes
I feel like the two-pairs-of-eyes rule isn't working very well.  In principle I still like having two people review each PR, and when it actually happens there is often useful feedback.  However, when it takes weeks or more for a PR to get a second pair of eyes on it, whereas one or two people are actively working on development and would like to merge stuff more often than monthly, I feel like the rule is getting in the way of progress.  I hesitate to propose simply scaling back to one pair of eyes, but can anyone think of a better solution?  Perhaps we could set a time limit of a few days, say, after which, if no one has expressed any reservations, a PR can be merged with one approval?

mikeshulman:
Since we are changing the rules for approval of pull requests, I think *this* pull request should be approved under the old system of two pairs of eyes (or more).

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1148,10 +1148,11 @@ comment such as LGTM ("Looks Good To Me").  Currently the rules are:
   successfully.
 
   Note also that Travis doesn't automatically restart itself on a pull
-  request when the master branch changes, so if other pull requests
-  have been merged in the interval since a given pull request was
-  first submitted it may be necessary to manually restart Travis
-  before merging it.
+  request when the master branch changes.  Thus, if other pull
+  requests have been merged in the interval since a given pull request
+  was first submitted, it may be necessary to rebase that pull request
+  against the new master and then manually restart Travis, to make
+  sure before merging it that it won't break the master branch.
 
 - In the absence of objections, two approvals suffice for a pull
   request to be merged.  Thus, instead of giving a second approval one
--- a/STYLE.md
+++ b/STYLE.md
@@ -51,7 +51,7 @@
     - [Available tactics](#available-tactics)
   - [Contributing to the library](#contributing-to-the-library)
     - [Fork & Pull](#fork--pull)
-    - [Two pairs of eyes](#two-pairs-of-eyes)
+    - [Approval of pull requests](#approval-of-pull-requests)
     - [Commit messages](#commit-messages)
     - [Creating new files](#creating-new-files)
     - [Travis](#travis)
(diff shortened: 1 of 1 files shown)


Alizter:
In light of recent mess ups by me can we also add a note about restarting travis before every PR to be certain it isn't broken.

Other than that it looks fine to me.

mikeshulman:
Good idea, done.

STYLE.md
```diff
@@ -1129,17 +1129,40 @@ on each other.
...
+  confirms that it builds successfully.  Exceptions to this rule
+  sometimes have to be made if the Travis configuration is broken for
+  some unrelated reason, but in that case it is better if the
+  person(s) approving the pull request confirms locally that it builds
+  successfully.
+
+  Note also that Travis doesn't automatically restart itself on a pull
+  request when the master branch changes, so if other pull requests
+  have been merged in the interval since a given pull request was
+  first submitted it may be necessary to manually restart Travis
```
JasonGross: If we keep this wording, we should update the Travis script to merge the latest master before running. Otherwise, we should make a note here about rebasing the PR.

mikeshulman: Thanks, fixed.

STYLE.md
```diff
@@ -1129,17 +1129,41 @@ on each other.
...
+  discussion must be had and resolved).
+
+- In general, a pull request should not be merged unless Travis CI
+  confirms that it builds successfully.  Exceptions to this rule
+  sometimes have to be made if the Travis configuration is broken for
+  some unrelated reason, but in that case it is better if the
+  person(s) approving the pull request confirms locally that it builds
+  successfully.
+
+  Note also that Travis doesn't automatically restart itself on a pull
```
JasonGross:
```suggestion
  Note also that Travis doesn't automatically incorporate new changes from master on a pull
```
My suggested wording is not the best, but it is less misleading than the existing wording.

STYLE.md
```diff
@@ -1129,17 +1129,41 @@ on each other.
...
+  request when the master branch changes.  Thus, if other pull
+  requests have been merged in the interval since a given pull request
+  was first submitted, it may be necessary to rebase that pull request
+  against the new master and then manually restart Travis, to make
```
JasonGross:
```suggestion
  against the new master, to make
```
Travis should automatically restart itself when you force-push after rebasing.

JasonGross:
Beyond the scope of this PR, but it might be nice to have Travis pull the latest changes from master, so that we don't actually have to rebase, but only redstart Travis.  Maybe we should open an issue?  (I might make a PR doing this within the next couple of days.)

SkySkimmer:
I believe it already does

Gaëtan Gilbert

JasonGross:
Huh, you're right.  Looking at the build-logs, and [the documentation](https://docs.travis-ci.com/user/pull-requests/#my-pull-request-isnt-being-built( Travis seems to use GitHub's `/merge` refs feature, which is [apparently undocumented and will not be up to date unless a user visits the PR page or the web API is used to update the ref](https://discourse.drone.io/t/github-claims-that-merge-refs-are-undocumented-feature/1100).  It's not clear to me what instructions we should give to pull request authors.  (It's especially not clear to me whether or not the merge ref is updated when a user without merge permissions visits the PR page.)

STYLE.md
JasonGross: Suggestion retracted based on @SkySkimmer's comment

mikeshulman:
Is this PR okay as-is, @JasonGross, or does it need further changes?

PR #1176, opened 2020-01-05
[hoqidetop] Create a shim for coqidetop

artagnon:
The [vscoq](https://github.com/coq-community/vscoq) extension uses coqidetop.opt, so let's create a shim for this.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -4,6 +4,9 @@ bin_SCRIPTS = hoqtop hoqc hoqdep hoq-config
 if make_hoqide
   bin_SCRIPTS += hoqide
 endif
+if make_hoqidetop
+  bin_SCRIPTS += hoqidetop
+endif
 if make_hoqtopbyte
   bin_SCRIPTS += hoqtop.byte
 endif
--- a/configure.ac
+++ b/configure.ac
@@ -108,6 +108,34 @@ AS_IF([test "x$make_coqide" = "xno"],
 
 AM_CONDITIONAL(make_hoqide, [test "$make_coqide" = "yes"])
 
+# Checking for coqidetop, which can be optional
+make_coqidetop="no"
+AC_ARG_WITH([coqidetop],
+            [AS_HELP_STRING([--without-coqidetop], [disable support for coqidetop])],
+            [],
+            [make_coqidetop="yes"])
+
+AS_IF([test "x$make_coqidetop" = "xno"],
+      [AC_MSG_NOTICE([Will not make hoqidetop])],
+      [AC_ARG_VAR([COQIDETOP], [the absolute path of the coqidetop executable])
+       AS_IF([test "x$COQIDETOP" != "x"],
+             [AC_MSG_CHECKING([for coqidetop])
+              AC_MSG_RESULT([(preset) $COQIDETOP])
+              AC_SUBST([COQIDETOP])],
+             [AC_PATH_PROG([COQIDETOP],[coqidetop.opt],[no],[$COQBIN_PATH])])
+       AS_IF([test "x$COQIDETOP" = "xno"],
+             [AC_MSG_NOTICE([Could not find coqidetop])
+              make_coqidetop="no"],
+             [AC_MSG_NOTICE([Trusting that coqidetop version is $COQVERSION])
+              COQIDETOPVERSION="$COQVERSION"
+              AC_SUBST([COQIDETOPVERSION])
+              AS_IF([test "x$COQIDETOPVERSION" != "x$COQVERSION"],
+                    [AC_MSG_ERROR([Version mismatch between coqtop $COQVERSION and coqidetop $COQIDETOPVERSION])])
+             ])
+      ])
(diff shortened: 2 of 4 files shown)


artagnon:
On the CI failure: Looks like the PPA [jgross-h/coq-master-daily](https://launchpad.net/~jgross-h/+archive/ubuntu/coq-master-daily) doesn't provide coqidetop.opt. Both my opam coq 8.10 package, as well as the source tree (built with dune) provide codeidetop.opt, so I'm curious about what this PPA does.

@JasonGross : could you comment on this?

JasonGross:
The ppa does not provide CoqIDE because I could not get the various lablgtk3 dependencies to build on Trusty.  In particular, last I checked, the failure was "missing build dependency libgtk-3-dev (>= 3.18)", and I did not feel like backporting a newer version of all of gtk3 and it's associated packages.

Alizter:
@JasonGross Would it be difficult to move onto Xenial?

artagnon:
The target is called [coqide-server](https://github.com/coq/coq/blob/master/Makefile.dune#L26), which is independent of the `coqide` target, I think. Is it worth including in the PPA build, or should I make this optional?

JasonGross:
@Alizter we are talking about Travis.  Feel free to change `dist: trusty` in the `travis.yml` file.  Xenial is not new enough, though; only the Eoan packages support CoqIDE.  (I don't recall what exactly the errors were for versions between Trusty and Eoan, but I believe I did try to backport the requirements and gave up when I couldn't figure out how to make things work, or when it looked like I needed to backport a newer version of like, libc or something.)

@artagnon the ppa installs everything that is present when you configure without CoqIDE and run `make && make install`.  If you want coqidetop.opt present by default in the ppa, then perhaps Coq's build should build it by default even when configured without support for CoqIDE?  In any case, given that currently coqidetop.opt does not get installed when Coq is configured without support for CoqIDE, and given that we support this use case, coqidetop.opt must be optional.

artagnon:
Got it. Better?

mikeshulman:
What is coqidetop?

artagnon:
Roughly speaking, [coqidetop](https://github.com/coq/coq/blob/master/ide/idetop.ml) an extension of coqtop, with some conveniences that makes it easy for an IDE to interact with; it is used by coqide. Just like coqtop, it communicates via XML when `-main-channel` and `-control-channel` are [specified](https://github.com/coq-community/vscoq/blob/master/server/src/coqtop/CoqTop8.ts#L212-L234).

Historically speaking, there was an `-ideslave` option for `coqtop` (currently deprecated, I think), which would not offer a prompt, and simply listen and respond in XML. The motivation for `coqidetop` arose from wanting to split out that functionality.

mikeshulman:
But it's not used by proofgeneral?

artagnon:
No, PG [uses](https://github.com/ProofGeneral/PG/blob/master/coq/coq-system.el#L408) `coqtop -emacs`, yet another mode.

spitters:
I understand there is a long term effort to converge on these interfaces. @ejgallego knows more.

maximedenes:
Naive question: why does HoTT  need a shim for this? It seems to be primarily passing flags to `coqide`/`coqc`. Is this due to a deficiency of `_CoqProject` that would not allow to pass these flags while using a standard coq toplevel?

mikeshulman:
It's possible that we don't; IIRC the whole collection of shims was set up back before `_CoqProject` existed.

In exactly what circumstances does coq read a particular `_CoqProject` file and apply the options listed therein?  E.g. if someone is working on a development in another directory that imports the HoTT library, will the options listed in our `_CoqProject` carry over to their project automatically?

SkySkimmer:
Not sure how well putting -coqlib in _coqproject works. For sure you need -arg and it won't be used by coqdep, but I don't know what impact that has.

SkySkimmer:
Basically there is a choice between telling users to use hoqtop/hoqc/etc (and setting up vscode to use hoqidetop now), vs telling them to put some -coqlib and -R HoTT in their _CoqProject (with paths dependent on their install).

maximedenes:
At least, it looks like a valid use case, so we should try to support it. So I guess you are saying that `-coqlib` should become a first class `_CoqProject` flag, like `-R`.

maximedenes:
Would `-coqlib` still be required once https://github.com/HoTT/HoTT/issues/963 is done?

SkySkimmer:
I guess it would be replaced by -noinit

maximedenes:
Right, in fact this is what stdlib2 does. So I guess HoTT could follow the same path. Thanks for the explanations, I guess we can stay with shims for now, but it is useful to know how the situation could be further improved.

maximedenes:
Yes, I'm well aware, Emilio, Vincent and I co-authored that PR (if you mean https://github.com/coq/coq/pull/186).

spitters:
I believe there was more work after that, but I cannot currently find it.

PR #1177, opened 2020-01-05
[README] Update link to HoTT in Lean

artagnon:
A HoTT library was present in Lean2, but there is none for Lean3 in
active development. While at it, also point to the experimental Cubical
Agda, which started with Mortberg's cubicaltt work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,17 +13,20 @@ It draws many ideas from Vladimir Voevodsky's
 [Foundations](https://github.com/vladimirias/Foundations) library (which has since been
 incorporated into the [UniMath](https://github.com/UniMath/UniMath) library) and also
 cross-pollinates with the [HoTT-Agda](https://github.com/HoTT/HoTT-Agda) library.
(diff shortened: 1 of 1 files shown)


Alizter:
The lean2 library hasn't gone anywhere. It's true that support for HoTT was dropped for lean3. There is the Spectral library based on the lean2 library as well. There was an effort to try and get a HoTT library working for lean3 but I am not sure it went anywhere.

I propose we just update the link to point to Floris van Doorn's lean2 library and the Spectral library.

artagnon:
Better?

mikeshulman:
I'm not convinced we should try to list all the HoTT formalization projects here.  We can't expect to be comprehensive, so I think it would be better to focus on just establishing this library's place in the ecosystem.  Which of the other projects are important to mention for that purpose?

Alizter:
I'm ok with mentioning these libraries here. I agree that it shouldn't be our *duty* to list them, but I don't see any harm in merging this.

PR #1180, opened 2020-01-15
Basics of wild 1-categories

mikeshulman:
This is joint work with @alizter, @emilyriehl, @mpopie, and @tslilc who visited last week.  It's building on the basic approach of #1163, but reorganized and renamed a bit.  In particular:

- We realized that wild categories allow reusing the basic definitions at different dimensions in an enriched style.  This is quite satisfying!
- The terminology is now more coherent, roughly matching the standard higher-categorical notion of (n,r)-category (made wild).
- We got rid of the duplicate data for opposites again; we can consider putting it back, but the issue is that we also want to use this as the basis for a definition of bicategory, potentially even tricategory, etc., and if extra data is hanging around then it would need extra coherence laws.  It also becomes a problem when trying to construct displayed categories without having to explicitly give the redundant data.  So far, we haven't been bothered by the behavior of opposites.

This PR basically gets up to the Yoneda lemma and the definition of Type and pType as wild 1-categories.  We're working on constructing a wild 2-category of wild 1-categories, and hope to get to other stuff like displayed categories, Grothendieck constructions, closed/monoidal categories, and enrichment.

I had hoped that the technology of wild functors could replace the disparate proofs of functoriality in Types/Prod, Types/Sum, etc., but doing that turns out to create universe polymorphism issues (in places like Modalities and PropResizing/Nat that are finicky about universe annotations), since the resulting proofs have different universe parameters.  I'm still not sure whether there is a solution to that.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -43,6 +43,22 @@ theories/Basics/Trunc.v
 theories/Basics/Decidable.v
 
 #
+#   Wildcat
+#
+theories/WildCat.v
+theories/WildCat/Core.v
+theories/WildCat/UnitCat.v
+theories/WildCat/EmptyCat.v
+theories/WildCat/Prod.v
+theories/WildCat/Equiv.v
+theories/WildCat/Sum.v
+theories/WildCat/Opposite.v
+theories/WildCat/Type.v
+theories/WildCat/Induced.v
+theories/WildCat/FunctorCat.v
+theories/WildCat/Yoneda.v
+
+#
 #   Types
 #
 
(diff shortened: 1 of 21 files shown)


Alizter:
Since I also authored this PR does my review "count" as APOE?

mikeshulman:
I think it counts.  The point is just to have two people looking at the code, whether or not they also authored part of it doesn't seem relevant to me.

mikeshulman:
(But if anyone disagrees, please speak up.)

mikeshulman:
Oops!  Looks like it needed to be rebased before merging.

Issue #1154, opened 2019-12-06
Replace the hottclasses integers


mikeshulman:
Now that #1137 is merged, we should try replacing the natpair_integers in hottclasses with the binary ones.  Hopefully the binary ones will suffice for the ring tactics and enable the new Quotient to work.

spitters:
This [paper by Altenkirch and Scoccolla](http://luisscoccola.github.io/Luis%20Scoccola_files/int-as-hit.pdf) may be relevant

mikeshulman:
The Altenkirch-Scoccola paper is [now on the arxiv](https://arxiv.org/abs/2007.00167) (and the previous link appears broken).  It looks like as clever as their definition is, it is still unary, hence not computationally efficient.

PR #1181, opened 2020-01-16
Global Set Loose Hint Behavior "Strict"

mikeshulman:
This [new flag](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:opt.loose-hint-behavior) prevents typeclass instances from "leaking" out of un-imported modules and through transitivity of imported files.  I think this is the correct behavior: only instances from imported modules should be visible.  (It also fixes a problem in at least one place that the import of `Modalities.Identity` in `Truncations.Core` kills typeclass search.)

It turns out that we were (probably unknowingly) relying on this (mis)feature in a lot of places, so I had to change a lot of files.  And there are some that I still couldn't figure out how to fix.  Currently the following files are still giving errors with `make -k`:

- [x] `Categories/Pseudofunctor/Identity`
- [x] `Categories/Grothendieck/PseudofunctorToCat`
- [x] `Categories/SimplicialSets`
- [x] `Categories/ExponentialLaws/Law1/Law`
- [x] `Classes/tactics/ring_tac`
- [x] `Classes/tests/ring_tac11`
- [x] `Classes/implementations/natpair_integers`
- [x] `Classes/orders/sum`
- [x] `Classes/theory/ua_quotient_algebra`
- [x] `Classes/theory/rings`

I don't understand enough about the Categories and Classes libraries to fix these easily; can someone else look at it please?  I've checked "Allow edits from maintainers" on this draft PR so anyone who wants to work on fixing these files can just push to my branch.  Of course, once the files are fixed, there may be other files that depend on them that will also have their own errors; feel free to use the "task list" above to track the files that need work.

### DIFF
--- a/theories/Classes/implementations/natpair_integers.v
+++ b/theories/Classes/implementations/natpair_integers.v
@@ -14,6 +14,7 @@ Require Import
   HoTT.Classes.interfaces.naturals
   HoTT.Classes.interfaces.integers
   HoTT.Classes.theory.rings
+  HoTT.Classes.theory.apartness
   HoTT.Classes.orders.sum
   HoTT.Classes.orders.rings
   HoTT.Classes.tactics.ring_tac
@@ -22,7 +23,7 @@ Require Import
 
 Generalizable Variables B.
 
-Local Set Loose Hint Behavior "Lax".
+Import ring_quote.Quoting.Instances.
 Local Set Universe Minimization ToSet.
 
 Module NatPair.
--- a/theories/Classes/implementations/natpair_integers.v
+++ b/theories/Classes/implementations/natpair_integers.v
@@ -8,11 +8,13 @@ Require Import HoTT.HIT.quotient
   HoTT.Types.Sum
   HoTT.TruncType.
 Require Import
+  HoTT.Classes.implementations.peano_naturals
   HoTT.Classes.interfaces.abstract_algebra
   HoTT.Classes.interfaces.orders
   HoTT.Classes.interfaces.naturals
   HoTT.Classes.interfaces.integers
   HoTT.Classes.theory.rings
+  HoTT.Classes.orders.sum
   HoTT.Classes.orders.rings
   HoTT.Classes.tactics.ring_tac
   HoTT.Classes.interfaces.naturals
@@ -20,6 +22,7 @@ Require Import
 
 Generalizable Variables B.
 
+Local Set Loose Hint Behavior "Lax".
 Local Set Universe Minimization ToSet.
 
 Module NatPair.
--- a/theories/Classes/implementations/natpair_integers.v
+++ b/theories/Classes/implementations/natpair_integers.v
@@ -14,6 +14,7 @@ Require Import
   HoTT.Classes.interfaces.naturals
   HoTT.Classes.interfaces.integers
   HoTT.Classes.theory.rings
+  HoTT.Classes.theory.groups
   HoTT.Classes.theory.apartness
   HoTT.Classes.orders.sum
   HoTT.Classes.orders.rings
@@ -300,6 +301,8 @@ Arguments Tapart {_ _ _} _ _.
 Arguments to_ring N {_} B {_ _ _ _ _ _} / _.
 
 End PairT.
+
+Import PairT.
 
 Section contents.
 Universe UN UNalt.
--- a/theories/Classes/implementations/natpair_integers.v
+++ b/theories/Classes/implementations/natpair_integers.v
@@ -1118,4 +1118,38 @@ Global Instance Z_zero_product@{} : ZeroProduct Z
 
 End contents.
 
+Module Instances.
+  Global Existing Instance T_set.
+  Global Existing Instance inject.
+  Global Existing Instance Tle_hprop.
+  Global Existing Instance Tlt_hprop.
+  Global Existing Instance Tapart_hprop.
+  Global Existing Instance Z_of_pair.
+  Global Existing Instance Z_of_N.
+  Global Existing Instance R_dec.
+  Global Existing Instance Z0.
+  Global Existing Instance Z1.
+  Global Existing Instance Z_plus.
+  Global Existing Instance Z_mult.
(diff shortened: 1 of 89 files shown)


Alizter:
So I managed to fix a few of the hottclasses files. I got stuck with natpair_integers.v and tests/ring_tac.v however. I tried to have a go at the simplicial sets one from Categories but I couldn't make sense of it either.

JasonGross:
Shouldn't this be easy to debug mostly automatically?  Do a `Set Typeclasses Debug` at the top of each file that's not building, compile each file on master and pipe the output to a file, and on this branch and pipe the output to a file, and then run `diff` on the two build logs.  (I can try to take a look at the categories files, but won't get a chance for a couple of days, probably.)

Alizter:
@JasonGross Thanks for the suggestions, this has made it easier to debug.

Something strange is happening in natpair_integers.v. It uses ring_quot which are some ring tactics. But I think they are intentionally assuming that the module `Quoting` will leak instances.

One solution is to import Quoting but this seems wrong. What if we locally set `Local Set Loose Hint Behavior "Lax". `?

Alizter:
OK, I have managed to fix most of hottclasses. The categories library however seems to have some namespacing issues. In ExponentialLaws/Law1.Law.v `inverse` is not being recognised as `Functor.inverse`. Not sure what is causing this...

mikeshulman:
I would prefer not to locally set the flag back to lax.  If it seems wrong to import all of Quoting (why?), what about putting the relevant instances in a different module that can be imported without the rest of Quoting?

Alizter:
>I would prefer not to locally set the flag back to lax

I don't see any disadvantages to doing this. We keep (what seems to be) the intended behavior of the hottclasses library without having to too much. I would be interested to know why you don't think it is a good idea.

> If it seems wrong to import all of Quoting (why?)

Well I could be wrong, but it seems to me that the reason that stuff is in a module is precisely so it cannot be seen by users of the library.

>relevant instances in a different module that can be imported without the rest of Quoting

This seems like a sensible idea, but might require some work to get done. Perhaps we could locally set lax and then open an issue to remove uses of lax?

The difference between typeclasses leaking in hottclasses and leaking in the rest of the library is that it seems to be intentional in hottclasses and unintentional in the rest of the library. I want to be as minimally disruptive as possible when it comes to hottclasses which shouldn't be difficult since it is quite separate from the rest of the library.

SkySkimmer:
When I made the ring tactic for hottclasses I just did enough to make it work. I certainly didn't consider whether I was using lax hints, I just saw that whatever I tried first worked and so I moved on. Feel free to change it if convenient.

Alizter:
OK very well. I will import Quoting and move on.

JasonGross:
>  I will import Quoting and move on.

I think it would be better to `Export Quoting` (or make an inner module with just the hints and export that) from the file defining the ring tactics.  That way you don't have to Import Quoting just to use the tactics

mikeshulman:
FWIW, importing all of Quoting doesn't work, because inside Quoting `Plus` is a constructor of `Expr`, while in the rest of hottclasses it is a typeclass.

mikeshulman:
I put the instances from `Quoting` in a submodule that can be imported separately.  Now `natpair_integers` gets farther, but breaks on something else.

(Didn't we want to replace `natpair_integers` with `Spaces/Int` anyway (#1154)?  Probably that's more work than we want to do right now, though.)

I don't want to set loose hint behavior to lax because that is the Wrong Behavior, and we shouldn't ever rely on it or encourage anyone else to rely on it.

Alizter:
Just so I don't forget. When debugging put:
```coq
Set Typeclasses Debug.
Set Loose Hint Behavior "Lax".
```
before any problematic lines to see what instances where being found before the strict setting.

Alizter:
I seperated the instances in natpair_integers.NatPair into NatPair.Instances which lets us import the instances from natpair_integers without importing everything else. I wish there was a special import command for coq which let you import only the instances.

Alizter:
I just noticed that `Set Loose Hint Behavior "Warn".` tells us exactly which modules haven't been imported.

Alizter:
All fixed.

theories/Classes/implementations/natpair_integers.v
```diff
@@ -297,6 +302,8 @@ Arguments to_ring N {_} B {_ _ _ _ _ _} / _.
 
 End PairT.
 
+Import PairT.
```
mikeshulman: Not that it really matters (or needs changing), but just FYI, instead of `Module X... End X. Import X.` you can just say `Module Import X. ... End X.`.

theories/Classes/theory/integers.v
```diff
@@ -141,6 +144,8 @@ Qed.
 
 Global Instance int_dec : DecidablePaths Z | 10.
 Proof.
+Set Typeclasses Debug.
+Set Loose Hint Behavior "Lax".
```
mikeshulman: Forgot to remove this?

theories/Classes/implementations/natpair_integers.v
```diff
@@ -1111,4 +1118,38 @@ Global Instance Z_zero_product@{} : ZeroProduct Z
 
 End contents.
 
+Module Instances.
+  Global Existing Instance T_set.
```
mikeshulman: Note that you can declare a bunch of existing instances at once with `Global Existing Instances`, as long as you don't want to give them different priorities.

mikeshulman:
Awesome, thanks!

theories/Classes/implementations/natpair_integers.v
Alizter: What do you mean by this exactly? I couldn't work out how to list them.

mikeshulman: Look at how I did it in `ring_quote.Quoting.Instances`.

Alizter: Oh, I see. I missed the s at the end of Instances.

PR #1184, opened 2020-01-25
Replace eqp_issig_contr tactic with a lemma

jdchristensen:
The `eqp_issig_contr` tactic for characterizing identity types can be replaced by a lemma.  The advantages of this are that the type of the lemma provides clear documentation for what it can do, and the lemma only needs to be proved once, instead of each time the tactic is used.  I also split out part of the functionality into another lemma which might be of independent use.  All the uses of the old tactic have been converted to the new lemma, just by adding a `revert` and an `apply`.

### DIFF
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -233,7 +233,7 @@ Definition equiv_path_group {U : Univalence} {G H : Group}
 Proof.
   refine (equiv_compose'
     (B := sig (fun f : G <~> H => IsMonoidPreserving f)) _ _).
-  { eqp_issig_contr issig_group.
+  { revert G H; apply (eqp_issig_contr issig_group).
     + intros [G [? [? [? ?]]]].
       exists 1%equiv.
       exact _.
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -233,7 +233,7 @@ Definition equiv_path_group {U : Univalence} {G H : Group}
 Proof.
   refine (equiv_compose'
     (B := sig (fun f : G <~> H => IsMonoidPreserving f)) _ _).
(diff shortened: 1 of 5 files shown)


mikeshulman:
Why is the `revert` necessary?

jdchristensen:
If instead of

  `revert G H; apply (eqp_issig_contr issig_group)`

you try

  `apply (eqp_issig_contr issig_group G H)`

then Coq doesn't infer the implicit type family.

Alizter:
So `apply` can be pretty dumb when it comes to arguments. `rapply` should work if not `serapply` definitely will.

jdchristensen:
No, that's not the issue.  Even if you do something like

  `refine (eqp_issig_contr issig_group _ _ G H).`

Coq isn't able to figure out the type family given just one instance of it.  But

  `revert G H; apply eqp_issig_contr issig_group`

is easy to type and works.

Alizter:
@jdchristensen Intriguing! I tried it myself and it looks like coq really fails to unify when there isn't a `forall a0 a1` around. It's even weirder that it can't unify with the arguments explicitly given!

Alizter:
I'm happy with this PR, but I would also like to suggest a name change. What about changing `eqp_issig_contr` to `equiv_path_issig_contr`? Of course, if @mikeshulman is happy with it.

jdchristensen:
Ok, I made those changes.

mikeshulman:
Taking @Alizter's comment as an approval, I'll go ahead and merge.  BTW, thanks for all these contributions, Dan!  I look forward to seeing the whole development you're working on; are you considering submitting it to the library as well?

jdchristensen:
Yes, the things I'm working on are intended for the library, but will take some time.

Alizter:
Thanks Dan!

PR #1185, opened 2020-01-25
Avoid rewrites in *_morphism proofs; make into Global Instances

jdchristensen:
For work in progress, I need to reason about these proofs, so avoid rewrites.

Also, some of these results were reproved elsewhere, so make them into Global Instances.  This sometimes causes `serapply BuildGroupHomomorphism` to spin, so you need to either add `notypeclasses` or provide the map.

### DIFF
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -153,11 +153,7 @@ Definition grp_homo_compose {G H K : Group}
   : GroupHomomorphism H K -> GroupHomomorphism G H -> GroupHomomorphism G K.
 Proof.
   intros f g.
-  serapply Build_GroupHomomorphism.
-  1: exact (f o g).
-  intros x y.
-  refine (ap f (grp_homo_op g _ _) @ _).
-  apply grp_homo_op.
+  serapply (Build_GroupHomomorphism (f o g)).
 Defined.
 
 Definition grp_homo_id {G : Group} : GroupHomomorphism G G.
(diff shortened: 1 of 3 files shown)


mikeshulman:
I wish there were a better way to deal with `compose` instances.  `isequiv_compose` and `isfunctor_compose` have similar issues.  It's certainly convenient to have them as instances, but they all sometimes seem to cause typeclass inference to spin.

mikeshulman:
I'm curious how you could need to reason about these proofs.  Aren't they elements of propositions?  Or are you getting to them before the ishset property of a group/monoid is imposed?

jdchristensen:
I'm working on non-set magma's, but those lemmas work fine since they don't use the hset condition.

mikeshulman:
It looks like it is `invert_sg_morphism` that's causing `Build_GroupHomomorphism` to spin.  It seems to happen literally all the time (I just ran into a bunch of other cases when merging master into the wildcat development branch).  So I'm inclined to say that that probably should not be a global instance.

jdchristensen:
Ok, I can make a new pull request that undoes that.  Should it be an "Instance" like it was before, without the Global keyword?  I'm not sure what that buys you.  I guess you can use "Existing Instance" in other files if you want it to be automatic?

mikeshulman:
I made a PR at #1188 and Jason has some suggestions I'm still thinking about.

Making it a `Local Instance` just means it functions as an instance inside that file or module, which might have been depended on there (I don't know).  Any other file can always declare it a `Local Existing Instance` even if it was only a `Definition` originally.

PR #1186, opened 2020-01-25
Add equiv_path_forall2 and define path_forall2 using it

jdchristensen:
I needed `equiv_path_forall2`, and it seemed slightly easier to define this directly and then define `path_forall2` using it compared to showing that the existing definition of `path_forall2` is an equivalence.  As a result, `path_forall2` moves from Overture to Forall.

If this is a problem, I can do it the other way.  `path_forall2` isn't used anywhere in the library, but others might use it.

Another question:  I had to remove `{_}` and `_` from the `Global Arguments path_forall2` line.  Not sure why they were there.  The `{_}` was probably for `Funext` which is now in the context.  Not sure about the trailing `_`.  `_ _` would make more sense.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -674,13 +674,6 @@ Definition path_forall `{Funext} {A : Type} {P : A -> Type} (f g : forall x : A,
 
 Global Arguments path_forall {_ A%type_scope P} (f g)%function_scope _.
 
-Definition path_forall2 `{Funext} {A B : Type} {P : A -> B -> Type} (f g : forall x y, P x y) :
-  (forall x y, f x y = g x y) -> f = g
-  :=
-  (fun E => path_forall f g (fun x => path_forall (f x) (g x) (E x))).
-
-Global Arguments path_forall2 {_} {A B}%type_scope {P} (f g)%function_scope _.
-
 (** *** Tactics *)
 
 (** We declare some more [Hint Resolve] hints, now in the "hint database" [path_hints].  In general various hints (resolve, rewrite, unfold hints) can be grouped into "databases". This is necessary as sometimes different kinds of hints cannot be mixed, for example because they would cause a combinatorial explosion or rewriting cycles.
--- a/theories/Types/Forall.v
+++ b/theories/Types/Forall.v
@@ -285,6 +285,24 @@ Proof.
   symmetry; apply transport_pp.
 Qed.
 
+(** ** Two variable versions for function extensionality. *)
+
+Definition equiv_path_forall2 {A B : Type} {P : A -> B -> Type} (f g : forall x y, P x y)
+  : (forall (a : A) (b : B), f a b = g a b) <~> f = g
+  := (equiv_path_forall f g) oE (equiv_functor_forall_id (fun a => equiv_path_forall (f a) (g a))).
+
+Definition path_forall2 {A B : Type} {P : A -> B -> Type} (f g : forall x y, P x y)
+  : (forall x y, f x y = g x y) -> f = g
+  := equiv_path_forall2 f g.
+
+Global Instance isequiv_path_forall2 `{P : A -> B -> Type} (f g : forall x y, P x y)
+  : IsEquiv (path_forall2 f g) | 0
+  := _.
+
+Global Arguments equiv_path_forall2 {A B}%type_scope {P} (f g)%function_scope.
+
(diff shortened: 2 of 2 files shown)


Alizter:
It doesn't seem that the trailing underscores do anything here. I think that the main reason there is an `Arguments` command at all is to bind function_scope to the function arguments.

mikeshulman:
Moving this to `Forall` is fine with me.  `Types` is supposed to come right after `Basics` (although in future `WildCat` might intervene), so there shouldn't be any problem with importing it whenever this lemma is needed.

Yes, the initial underscore was for `Funext`.  The trailing single underscore still works for me and makes sense: it refers to the unnamed argument of type `forall x y, f x y = g x y` (which is one argument, although it is itself a function taking two arguments).  I would probably still include this underscore myself, as a sort of documentation that there is an extra explicit argument there.

It also occurs to me that it would be a bit more in line with our conventions to call this `path_forall11` (e.g. `ap011` is for two-variable functions while `ap02` is for two-dimensional paths).  What do others think?

jdchristensen:
I'm about to push the underscore fix, and will wait to hear consensus about naming.

jdchristensen:
About the naming:  In Overture.v, we have:

```
Definition ap2 {A B C} (f : A -> B -> C) {x1 x2 y1 y2}:
  x1 = x2 -> y1 = y2 -> f x1 y1 = f x2 y2.
```

In PathGroupoids.v, we have:

```
Definition ap011 {A B C} (f : A -> B -> C) {x x' y y'} (p : x = x') (q : y = y')
: f x y = f x' y'
```

These do the same thing with different names.  Should one of them be removed, maybe replaced with a parsing-only notation?

jdchristensen:
`ap2` is currently used 23 times while `ap011` is used 3 times.  That doesn't mean that switching to ap011 is wrong, but it just an observation.

Alizter:
Let's make ap2 a notation for ap011. I myself am guilty of using ap2 instead of ap011 mostly because ap2 is easier to remember. It could be argued that ap2 is better suited for apping on a 2-cell.

We can open an issue to change all uses of ap2 to ap011.

jdchristensen:
If there is consensus on this, then I will switch *path_forall2 to *path_forall11, I guess?  With no zero?

Alizter:
I'm not really sure what the 0 means here. I'm sure @mikeshulman knows.

mikeshulman:
Regular `ap` is a notation for `ap01`, which means that we apply a function (0-dimensional) to a path (1-dimensional).  By contrast, `ap10` applies a path between functions (1-dimensional) to a point (0-dimensional).  Similarly, `ap11` applies a path between functions to a path, `ap02` applies a function (0-dimensional) to a path between paths (2-dimensional), and `ap011` applies a function (0-dimensional) to two paths (1-dimensional and 1-dimensional).  So according to that principle, `ap2` should be a notation for `ap02`, and if what we mean is `ap011` we should write that instead.  It looks like `ap2` was copied from HoTTClasses; presumably its author didn't realize we already had `ap011`.

I don't think there is anything in `path_forall` for the `0` to refer to, so `path_forall11` would make sense there.

jdchristensen:
Ok, I renamed them.  I'll let someone else do the ap2 renaming.  So this should be ready to merge.

fpvandoorn:
I think 
```
{A B : Type} {P : A -> B -> Type}
```
should be generalized to
```
{A : Type} {B : A -> Type} {P : forall x, B x -> Type}
```

fpvandoorn:
For reference, the corresponding result in Lean is
The corresponding result in Lean is
https://github.com/leanprover/lean2/blob/8072fdf9a0b31abb9d43ab894d7a858639e20ed7/hott/arity.hlean#L263
(and that file contains many similar results; `~2` is notation for pointwise paths for binary functions)

jdchristensen:
Good point!  The proofs go through unchanged.  I guess I won't use D11 in the name, since the original path_forall was already dependent.

jdchristensen:
(All the permutations of `ap` and `funext` make me think that tactics would be good here.  I see that there is `f_ap` already, but I haven't tried it.  Oh, I just tried the `funext` tactic, and it works well in places that I had used `path_forall11`.  But I don't see how it can replace `equiv_path_forall11`.)

Alizter:
@jdchristensen You could use Build_Equiv and construct the forward map with the funext tactic. Then hopefully, typeclass inference can work out it is an equivalence.

Issue #1091, opened 2019-09-20
Why is trunc_equiv opaque?


Alizter:
In `Trunc.v` there is a definition `trunc_equiv`:

```coq
(** Equivalence preserves truncation (this is, of course, trivial with univalence).
   This is not an [Instance] because it causes infinite loops.
   *)
Definition trunc_equiv A {B} (f : A -> B)
  `{IsTrunc n A} `{IsEquiv A B f}
  : IsTrunc n B.
Proof.
  generalize dependent B; generalize dependent A.
  simple_induction n n IH; simpl; intros A ? B f ?.
  - exact (contr_equiv _ f).
  - intros x y.
    exact (IH (f^-1 x = f^-1 y) (H (f^-1 x) (f^-1 y)) (x = y) ((ap (f^-1))^-1) _).
Qed.
```
Now this breaks my computation. Therefore I swapped `Qed` for `Defined`. The whole library compiles, apart from `Card.v` which gets stuck (and memory gets out of hand).

Specifically in Fibrations.v, we have the following definition:
```coq
Global Instance isequiv_ap_isembedding {A B} (f : A -> B) `{!IsEmbedding f}
  : forall x y, IsEquiv (@ap _ _ f x y).
Proof.
  intros x y. apply isequiv_fcontr,_.
Defined.
```
This uses trunc_equiv in some critical way. It then becomes a bit useless that such an equivalence exists since the inverse doesn't compute.

My questions are as follows:

 * Can we have a better computationally behaved version of `isequiv_ap_isembedding`?
 * Why is `trunc_equiv` so critical to memory usage in `Card.v`?

jdchristensen:
There's a related problem that is `eisadj_other` in Equivalences.v (formerly `other_adj`) is made transparent, then coq spins while checking Card.v. The problem is in the call to `strip_truncations` in the `reduce` tactic during the proof of `exp_zero_card`.  If we can fix this, then we can inline the proof of `eisadj_other` into `isequiv_inverse`, and use `eisadj (f^-1)` in place of `eisadj_other`.

PR #1187, opened 2020-01-25
Make center of contr_basedequiv compute

jdchristensen:
For some reason, with the current proofs of `contr_basedequiv` and `contr_basedequiv'`, the following cause Coq to spin:

```
Compute (center {Y : Type & nat <~> Y}).
Compute (center {Y : Type & Y <~> nat}).
Compute (@center {Y : Type & nat <~> Y} (@contr_basedequiv nat)).
Compute (@center {Y : Type & Y <~> nat} (@contr_basedequiv' nat)).
```

So I replaced their proofs with proofs that give explicit centers.

### DIFF
--- a/theories/Types/Universe.v
+++ b/theories/Types/Universe.v
@@ -481,17 +481,18 @@ Definition equiv_induction'_comp (P : forall U V, U <~> V -> Type)
 Global Instance contr_basedequiv {X : Type}
 : Contr {Y : Type & X <~> Y}.
 Proof.
-  refine (trunc_equiv' {Y : Type & X = Y}
-           (equiv_functor_sigma' (equiv_idmap Type)
-             (fun Y => equiv_equiv_path X Y))).
+  exists (X; equiv_idmap).
+  intros [Y f]; revert Y f.
+  exact (equiv_induction _ idpath).
 Defined.
 
 Global Instance contr_basedequiv' {X : Type}
 : Contr {Y : Type & Y <~> X}.
 Proof.
-  refine (trunc_equiv' {Y : Type & Y = X}
-           (equiv_functor_sigma' (equiv_idmap Type)
-             (fun Y => equiv_equiv_path Y X))).
+  serapply Build_Contr.
+  - exact (X; equiv_idmap).
+  - intros [Y f]; revert Y X f.
+    refine (equiv_induction' _ (fun T => idpath)).
 Defined.
 
 (** ** Truncations *)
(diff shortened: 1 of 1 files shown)


mikeshulman:
So why does `exists` work in the first case but not the second?  Apparently Coq is able to guess that `Equiv X Y` is the type family `Equiv X` applied to `Y`, but not that the type family `Equiv Y X` is the type family `fun Z => Equiv Z X` applied to `Y`.  Is that because the first doesn't need an abstraction?  It seems bizarre that the order of arguments matters like this.

jdchristensen:
I have no idea why `exists` doesn't work in the second case.  There are several ways to make it work, but this way avoided writing out the type family.  One can also use `exists (exist (fun ...) X equiv_idmap)` (from memory), but that seems harder to read.

jdchristensen:
Another asymmetry is that we don't have the other one-sided form of `equiv_induction`.  In this case, `equiv_induction'` worked easily, but we could also add something like `equiv_induction_sym`.

mikeshulman:
I think it'd be nice to include a comment in the file saying something about why the proofs are different, or at least remarking on it -- or perhaps change the first proof to match the second.

I see that again you're using the `revert` trick here to enable Coq to guess the type family for `equiv_induction`.  Did you figure that out yourself?

I'd be in favor of an `equiv_induction_sym` (or perhaps `equiv_induction_opp` or `equiv_induction_inv` -- the latter is closer to our names for inverses of equivalences, so perhaps more suggestive of what it means).

jdchristensen:
Ok, I added a comment explaining the problem with `exists`.  And I added `equiv_induction_inv`, which now makes the two new proofs more similar.  We could go further and define an `_inv` version of `path_ind`, or is it already somewhere?  I'm surprised it hasn't been needed.

jdchristensen:
I'm not sure whether I figured out the `revert` trick myself or saw it somewhere.

mikeshulman:
I think most of the time we only do path induction when both endpoints are free.

Alizter:
I am not sure why trunc_equiv' needs to be opaque, but it does break some things when it is made transparent. I highlighted the issue in #1091. The first file which blows up is Spaces/Card.v

jdchristensen:
I just thought of a way to simplify the three equiv_induction* proofs, and pushed an update.  Sorry to add this churn just when you were probably about to commit it.

Alizter:
@jdchristensen No problem, are you OK with merging this now? I will wait till travis is happy.

jdchristensen:
Yes, I'm happy now.  BTW, was your comment about trunc_equiv' meant for another issue?

Alizter:
@jdchristensen No, I think it is the reason contr_basedequiv' was not computing before.

jdchristensen:
Ah, I get it.  Thanks for explaining the issue.

PR #1188, opened 2020-01-26
remove global compose and invers for semigrop and monoid morphisms

mikeshulman:
These cause typeclass loops in many other places in the spectral sequences development where there is more algebra, so I think it's better not to have them as instances.  In `grp_homo_compose` it was enough to pose `compose_sg_morphism` as a hint.  (I don't understand why `grp_iso_inverse` still works without any such hint.)

### DIFF
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -152,6 +152,7 @@ Defined.
 Definition grp_homo_compose {G H K : Group}
   : GroupHomomorphism H K -> GroupHomomorphism G H -> GroupHomomorphism G K.
 Proof.
+  pose @compose_sg_morphism.
   intros f g.
   serapply (Build_GroupHomomorphism (f o g)).
 Defined.
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -152,7 +152,6 @@ Defined.
 Definition grp_homo_compose {G H K : Group}
   : GroupHomomorphism H K -> GroupHomomorphism G H -> GroupHomomorphism G K.
 Proof.
-  pose @compose_sg_morphism.
   intros f g.
   serapply (Build_GroupHomomorphism (f o g)).
 Defined.
--- a/theories/Classes/theory/groups.v
+++ b/theories/Classes/theory/groups.v
@@ -244,7 +244,8 @@ Section compose_mor.
     split; split.
   Defined.
 
-  Global Instance compose_sg_morphism : IsSemiGroupPreserving f -> IsSemiGroupPreserving g ->
+  (** Making these global instances causes typeclass loops. *)
+  Local Instance compose_sg_morphism : IsSemiGroupPreserving f -> IsSemiGroupPreserving g ->
     IsSemiGroupPreserving (g ∘ f).
   Proof.
     red; intros fp gp x y.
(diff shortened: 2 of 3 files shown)


JasonGross:
I would suggest instead using `Hint Mode` on `IsSemiGroupPreserving` to tell typeclass search when to trigger instances.  Alternatively, you can use `Hint Extern` to make these instances only trigger when function composition is already part of the goal.  (The reason it loops currently is that it'll say "oh, I know how to show `IsSemiGroupPreserving ?f`: just let `?f` be the composition of two `IsSemiGroupPreserving` functions")

JasonGross:
> I don't understand why grp_iso_inverse still works without any such hint.

You can figure this out with `Set Typeclasses Debug`

mikeshulman:
Thanks.  I think I can figure out the `Hint Extern` approach, but I don't understand what you're envisioning for `Hint Mode`.

JasonGross:
Based on [the documentation](https://coq.inria.fr/refman/proof-engine/tactics.html#coq:cmdv.hint-mode), if you do `Hint Mode IsSemiGroupPreserving - - ! : typeclass_instances` (how many implicit arguments does `IsSemiGroupPreserving` have?  If it's not 2, replace `- -` with as many `-`s as there are implicit arguments), this should fix the problem with the compose instance by saying that typeclass resolution for `IsSemiGroupPreserving` should never trigger unless the head of the function is known.  (If you do `+` instead of `!`, it means that argument has to be evar-free, rather than headed by a non-evar.). It's possible this will break things elsewhere, if you ever do want typeclass resolution to guess the function for you.

theories/Algebra/Group.v
```diff
@@ -152,6 +152,7 @@ Defined.
 Definition grp_homo_compose {G H K : Group}
   : GroupHomomorphism H K -> GroupHomomorphism G H -> GroupHomomorphism G K.
 Proof.
+  pose @compose_sg_morphism.
```
jdchristensen: Since we need to mention it by name, it's probably easier for the reader to follow if you put `serapply compose_sg_morphism` as the last line of the proof.

jdchristensen:
Another thing I suggest is removing the `(f:=idmap)` that I inserted on around line 249 of Group.v.  For some reason, that was needed when those results were Global Instances.

mikeshulman:
Ah, I see; `Classes/theory/groups.v` actually already had `Hint Extern`s declared for these instances.  I presume that's why `grp_iso_inverse` worked; the difference with `grp_homo_compose` is that the hint for `compose_sg_morphism` was defined using (the hottclasses unicode notation for) `Compose` (which is a `Definition`) rather than `compose` (which is a `Notation` and is what `g o f` means).  Adding a similar `Hint Extern` for `compose` works in `Algebra/Group` (as long as we keep the `f:=idmap`), but causes spinning in `HomotopyGroup.pi_functor`.  I don't have time to debug that further right now, or to experiment with `Hint Mode`, but I may later if no one else has done it yet.

jdchristensen:
I'd vote for making the two changes I suggested above and then merging it for now.

mikeshulman:
Oops, somehow I must have been in a confused state.  Actually it seems like just adding a `Hint Extern` for `compose` in addition to `Compose` makes everything work.  Now pushed.

PR #1191, opened 2020-01-26
Undo making various *_morphism definitions into Global Instances

jdchristensen:
Also undid the related changes elsewhere that are no longer needed.

This partially reverts #1185.

In my tests, I had to change both the "invert" results and the "compose" results in order to stop typeclass resolution from spinning, which makes sense if you think about it.

Currently, those results are still left as (non-Global) Instances.  Is that the right thing to do?


jdchristensen:
Oops, I missed that Mike already made a pull request.

PR #1192, opened 2020-01-26
results on pTr

Alizter:
Here are some results about pointed truncations that I had lying around. These proofs could probably be simpler in some palces. I'm not sure if univalence can be removed either.

### DIFF
--- a/theories/Pointed/pTrunc.v
+++ b/theories/Pointed/pTrunc.v
@@ -49,7 +49,6 @@ Proof.
   apply path_pmap.
   serapply Build_pHomotopy.
   1: intro; by strip_truncations.
-  apply moveL_pV.
   reflexivity.
 Defined.
 
--- a/theories/Pointed/pTrunc.v
+++ b/theories/Pointed/pTrunc.v
@@ -1,9 +1,11 @@
 Require Import Basics.
 Require Import Types.
+Require Import Spaces.Nat.
 Require Import Pointed.Core.
+Require Import Pointed.pMap.
 Require Import Pointed.pEquiv.
 Require Import Pointed.Loops.
-Require Import HoTT.Truncations.
+Require Import Truncations.
 Import TrM.
 
 Local Open Scope pointed_scope.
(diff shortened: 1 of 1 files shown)


theories/Pointed/pTrunc.v
```diff
@@ -1,9 +1,11 @@
 Require Import Basics.
 Require Import Types.
+Require Import Spaces.Nat.
```
jdchristensen: It looks like this can be removed.

theories/Pointed/pTrunc.v
```diff
@@ -16,6 +18,47 @@ Global Instance ispointed_tr (n : trunc_index) (A : Type) `{IsPointed A}
...
+Defined.
+
+Global Instance isequiv_ptr n (A : pType) `{IsTrunc n A}
+  : IsEquiv (@ptr n A) := _.
+
+Definition pequiv_ptr {n : trunc_index} {A : pType} `{IsTrunc n A}
+  : A <~>* pTr n A := Build_pEquiv _ _ ptr _.
+
+Definition pTr_rec n {X Y : pType} `{IsTrunc n Y}
+  (f : X ->* Y)
```
jdchristensen: Move f to previous line?

theories/Pointed/pTrunc.v
```diff
@@ -16,6 +18,47 @@ Global Instance ispointed_tr (n : trunc_index) (A : Type) `{IsPointed A}
...
+  : pTr n X ->* Y.
+Proof.
+  serapply Build_pMap.
+  1: exact (Trunc_rec f).
+  apply point_eq.
```
jdchristensen: I like bullet style myself, but am not sure what's more common.

jdchristensen:
I think univalence is needed where you use it.

theories/Pointed/pTrunc.v
```diff
@@ -16,6 +18,47 @@ Global Instance ispointed_tr (n : trunc_index) (A : Type) `{IsPointed A}
...
+  1: serapply pTr_rec.
+  { intro f.
+    destruct f as [f p].
+    apply (ap (Build_pMap _ _ f)).
+    apply concat_1p. }
+  intro f.
+  apply path_pmap.
+  serapply Build_pHomotopy.
+  1: intro; by strip_truncations.
+  apply moveL_pV.
```
jdchristensen: I think the `apply moveL_pV` line has to be removed.

jdchristensen:
I don't know if I'm allowed to approve things, but I think that when the build failure is fixed (and optionally the other stylistic things) it can be merged.

theories/Pointed/pTrunc.v
Alizter: I made it an inline definition now.

Alizter:
Ah, I was testing this on a version where the path in the pHomotopy is organised differently. That's why I inserted the apply moveL_V.

jdchristensen:
LGTM.

Alizter:
@jdchristensen by the way, when you go to the "Files changed" tab on the PR, in the top right corner there is a green button where you can submit reviews.

Alizter:
The current rule is to wait 48 hrs since the PR was submitted for another review, if nobody reviews by then I will merge anyway.

mikeshulman:
I should be able to look at this.

Alizter:
@mikeshulman Do you still want to look at this?

PR #1196, opened 2020-01-29
Simplify proof of other_adj in Equivalences.v

jdchristensen:
The first commit only moves code around, with _no changes_, except that I needed to remove the "Paths." prefix in a few other places to make the library compile.  The purpose of the move is to have `isequiv_ap` available when proving `other_adj`.

The second commit replaces the long proof of `other_adj` with a short proof that has no rewrites.

It's better to focus on the second commit to see what changed.

If I change the `Qed` at the end of `other_adj` to a `Defined`, then coq spins while checking Card.v.  The problem is in the call to `strip_truncations` in the `reduce` tactic during the proof of `exp_zero_card`.  It would be nice to make `other_adj` transparent, but if it's tricky to fix Card.v, then I'm happy with having this merged as is for now.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -101,108 +101,6 @@ Section IsEquivHomotopic.
 
 End IsEquivHomotopic.
 
-
-(** The inverse of an equivalence is an equivalence. *)
-Section EquivInverse.
-
-  Context {A B : Type} (f : A -> B) {feq : IsEquiv f}.
-  Open Scope long_path_scope.
-
-  Theorem other_adj (b : B) : eissect f (f^-1 b) = ap f^-1 (eisretr f b).
-  Proof.
-    (* First we set up the mess. *)
-    rewrite <- (concat_1p (eissect _ _)).
-    rewrite <- (concat_Vp (ap f^-1 (eisretr f (f (f^-1 b))))).
-    rewrite (whiskerR (inverse2 (ap02 f^-1 (eisadj f (f^-1 b)))) _).
-    refine (whiskerL _ (concat_1p (eissect _ _))^ @ _).
-    rewrite <- (concat_Vp (eissect f (f^-1 (f (f^-1 b))))).
-    rewrite <- (whiskerL _ (concat_1p (eissect f (f^-1 (f (f^-1 b)))))).
-    rewrite <- (concat_pV (ap f^-1 (eisretr f (f (f^-1 b))))).
-    apply moveL_M1.
-    repeat rewrite concat_p_pp.
-    (* Now we apply lots of naturality and cancel things. *)
-    rewrite <- (concat_pp_A1 (fun a => (eissect f a)^) _ _).
-    rewrite (ap_compose' f f^-1).
-    rewrite <- (ap_p_pp _ _ (ap f (ap f^-1 (eisretr f (f (f^-1 b))))) _).
-    rewrite <- (ap_compose f^-1 f).
-    rewrite (concat_A1p (eisretr f) _).
-    rewrite ap_pp, concat_p_pp.
-    rewrite (concat_pp_V _ (ap f^-1 (eisretr f (f (f^-1 b))))).
-    repeat rewrite <- ap_V; rewrite <- ap_pp.
-    rewrite <- (concat_pA1 (fun y => (eissect f y)^) _).
-    rewrite ap_compose', <- (ap_compose f^-1 f).
-    rewrite <- ap_p_pp.
-    rewrite (concat_A1p (eisretr f) _).
-    rewrite concat_p_Vp.
-    rewrite <- ap_compose.
-    rewrite (concat_pA1_p (eissect f) _).
-    rewrite concat_pV_p; apply concat_Vp.
-  Qed.
-
-  Global Instance isequiv_inverse : IsEquiv f^-1 | 10000
-    := Build_IsEquiv B A f^-1 f (eissect f) (eisretr f) other_adj.
-End EquivInverse.
-
-(** If the goal is [IsEquiv _^-1], then use [isequiv_inverse]; otherwise, don't pretend worry about if the goal is an evar and we want to add a [^-1]. *)
-Hint Extern 0 (IsEquiv _^-1) => apply @isequiv_inverse : typeclass_instances.
-
-(** [Equiv A B] is a symmetric relation. *)
-Theorem equiv_inverse {A B : Type} : (A <~> B) -> (B <~> A).
-Proof.
-  intro e.
-  exists (e^-1).
-  apply isequiv_inverse.
-Defined.
-
-Notation "e ^-1" := (@equiv_inverse _ _ e) : equiv_scope.
-
-Global Instance symmetric_equiv : Symmetric Equiv | 0 := @equiv_inverse.
-
-(** If [g \o f] and [f] are equivalences, so is [g].  This is not an Instance because it would require Coq to guess [f]. *)
-Definition cancelR_isequiv {A B C} (f : A -> B) {g : B -> C}
-  `{IsEquiv A B f} `{IsEquiv A C (g o f)}
-  : IsEquiv g
-  := isequiv_homotopic (compose (compose g f) f^-1)
-       (fun b => ap g (eisretr f b)).
-
-Definition cancelR_equiv {A B C} (f : A -> B) {g : B -> C}
-  `{IsEquiv A B f} `{IsEquiv A C (g o f)}
-  : B <~> C
-  := Build_Equiv B C g (cancelR_isequiv f).
-
-(** If [g \o f] and [g] are equivalences, so is [f]. *)
-Definition cancelL_isequiv {A B C} (g : B -> C) {f : A -> B}
-  `{IsEquiv B C g} `{IsEquiv A C (g o f)}
-  : IsEquiv f
-  := isequiv_homotopic (compose g^-1 (compose g f))
-       (fun a => eissect g (f a)).
-
-Definition cancelL_equiv {A B C} (g : B -> C) {f : A -> B}
-  `{IsEquiv B C g} `{IsEquiv A C (g o f)}
-  : A <~> B
-  := Build_Equiv _ _ f (cancelL_isequiv g).
-
-(** Combining these with [isequiv_compose], we see that equivalences can be transported across commutative squares. *)
-Definition isequiv_commsq {A B C D}
-           (f : A -> B) (g : C -> D) (h : A -> C) (k : B -> D)
-           (p : k o f == g o h)
-           `{IsEquiv _ _ f} `{IsEquiv _ _ h} `{IsEquiv _ _ k}
-: IsEquiv g.
-Proof.
-  refine (@cancelR_isequiv _ _ _ h g _ _).
-  refine (isequiv_homotopic _ p).
-Defined.
-
-Definition isequiv_commsq' {A B C D}
-           (f : A -> B) (g : C -> D) (h : A -> C) (k : B -> D)
-           (p : g o h == k o f)
-           `{IsEquiv _ _ g} `{IsEquiv _ _ h} `{IsEquiv _ _ k}
-: IsEquiv f.
-Proof.
-  refine (@cancelL_isequiv _ _ _ k f _ _).
-  refine (isequiv_homotopic _ p).
-Defined.
-
 (** Transporting is an equivalence. *)
 Section EquivTransport.
 
@@ -406,6 +304,12 @@ Definition isequiv_isequiv_postcompose {A B : Type} (f : A -> B)
 (* TODO *)
 *)
 
+(** ** Functorial action *)
+
+(** 'functor_path' is called [ap]. *)
+
+(** ** Equivalences between path spaces *)
+
 (** If [f] is an equivalence, then so is [ap f].  We are lazy and use [adjointify]. *)
 Global Instance isequiv_ap `{IsEquiv A B f} (x y : A)
   : IsEquiv (@ap A B f x y) | 1000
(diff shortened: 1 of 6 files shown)


theories/Basics/Equivalences.v
```diff
@@ -426,6 +330,101 @@ Global Instance isequiv_ap `{IsEquiv A B f} (x y : A)
...
+
+(** The inverse of an equivalence is an equivalence. *)
+Section EquivInverse.
+
+  Context {A B : Type} (f : A -> B) {feq : IsEquiv f}.
+
+  Theorem other_adj (b : B) : eissect f (f^-1 b) = ap f^-1 (eisretr f b).
+  Proof.
+    apply (equiv_inj (ap f)).
+    refine ((_ @ _ @ _)^ @ _).
```
mikeshulman: This is a much shorter and better proof than the one we had, but the way it's written right now I find very hard to follow.  It starts by breaking up the goal into a concatenation of four paths with totally unspecified intermediate points, then applies something to the fourth subgoal (which we can't even see completely) that instantiates one of the missing points in the first subgoal.  Then when it applies `ap_homotopic_id` the other endpoint of the first subgoal is _partially_ instantiated, but I don't immediately see what it gets instantiated to because the first subgoal disappears and is replaced by the remains-to-be-specified argument to `ap_homotopic_id` (in fact, it took me a while to figure out that that was what was appearing at that point).  Specifying that argument then completes the instantiation of that missing endpoint, which then appears in the second subgoal; and so on.

In general, I don't like proofs with lots of existential variables.  Among other things, this sort of causal flow of what gets instantiated by what is hard for me to follow.

Here are three reformattings of this proof (after the first `apply (equiv_inj (ap f))`) that I find easier to follow.  The first one applies lemmas in the same order as yours, but keeps a single subgoal throughout without any existential variables, so we can see it clearly as constructing a path step by step:
```
    refine (_ @ ap_compose _ _ _).
    refine (_ @ (ap_homotopic_id (eisretr f) _)^).
    refine (_ @ (concat_pp_V _ _)^).
    symmetry; apply eisadj.
```
The second one works from both sides instead of just one, allowing the use of `concat_A1p` instead of `ap_homotopic_id` (but requires moving `equiv_concat_r`, or some replacement for it, back into `Equivalences` as well):
```
    refine (_ @ ap_compose _ _ _).
    refine ((eisadj f (f^-1 b))^ @ _).
    apply (equiv_inj (equiv_concat_r (eisretr f b) _)); cbn; symmetry. 
    exact (concat_A1p (eisretr f) (eisretr f b)).
```
The third one is probably my favorite, despite being one line longer.  It starts by writing down the relevant naturality square (rather than sneakily setting things up for it to pop out of a hat), notices that something in it can be canceled on both sides, and then that the resulting path is a natural bridge between the two endpoints of the goal:
```
    assert (p := concat_A1p (eisretr f) (eisretr f b)).
    apply cancelR in p.
    refine (_ @ p^ @ _).
    - symmetry; apply eisadj.
    - apply ap_compose.
```

theories/Types/Paths.v
```diff
@@ -109,29 +109,6 @@ Proof.
   refine (H 1 q).
 Defined.
 
-(** ** Functorial action *)
-
-(** 'functor_path' is called [ap]. *)
-
-(** ** Equivalences between path spaces *)
-
-(** [isequiv_ap] is in Equivalences.v  *)
```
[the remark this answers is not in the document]
jdchristensen: Sure, I'll do that.

theories/Basics/Equivalences.v
jdchristensen: Thanks for the feedback.  If I were to prove this on paper, I would do it as a displayed sequence of equalities going from the RHS to the LHS.  I in particular find the last proof you propose a bit jarring, since it pulls something out of a hat that it isn't obvious we'll need.  Your first proof is fine, but I don't like having to know how many underscores to provide to each lemma.  I'm about to push a version that handles your complaints about the visibility of goals by using braces and by inlining the use of eisretr.  I added a comment which should also help make it clear to the reader what is happening, and moved the symmetry into each step so the goals are visibly going from right to left.  Let me know what you think.

jdchristensen:
Two questions:  should the name `other_adj` be changed to something more useful globally?  `eisadj'`?

And what about the `Qed` vs `Defined` issue?

theories/Basics/Equivalences.v
Alizter: It would be nice if we had a tactic that for a given relation, you supply a list of midpoints, and it generates goals for paths (or relations) between those midpoints. This would seem to more closely adhere to on paper reasoning. I think lean has something similar to this?

My personal taste, is that the first proof is the most readable. But I agree that it looks a bit intimidating with all the underscores. For me, this is something you just get used to.

Also don't forget, you can always add in comments in proofs if you think it is confusing!

theories/Basics/Equivalences.v
```diff
@@ -426,6 +330,101 @@ Global Instance isequiv_ap `{IsEquiv A B f} (x y : A)
...
+    4: apply ap_compose.
+    - serapply ap_homotopic_id.
+      intro; apply eisretr.
+    - simpl.  apply concat_pp_V.
+    - apply eisadj.
+  Qed.
```
Alizter: I think this is fine to make `Defined`. I think the only reason that it was `Qed` before is that we never really noticed that it was proof irrelevant was because we never needed it to be relevant elsewhere. I suspect that the length of the previous proof was the reason that compelled whoever to use Qed here.

Alizter:
@jdchristensen Regarding the name. `eisadj'` might suggest `eisadj` but taking equivalent but different arguments. What about `eisadj_other`?

theories/Basics/Equivalences.v
jdchristensen: @Alizter See my first comment about the problem in Card.v

jdchristensen:
I like `eisadj_other`.

theories/Basics/Equivalences.v
jdchristensen: @Alizter This is presumably related to #1091 about `trunc_equiv` needing to be opaque.  Maybe we can change Card.v or `strip_truncations` so that both `other_adj` and `trunc_equiv` can be made transparent.

jdchristensen:
@Alizter There's a build failure involving CayleyDickson.v, which wasn't in my repo when I tested my set-up.  But I don't see how my changes could make `Square` unavailable. 

```
HOQC theories/Homotopy/CayleyDickson
File "./theories/Homotopy/CayleyDickson.v", line 261, characters 12-18:
Error: The reference Square was not found in the current environment.
```

Alizter:
@jdchristensen That's my fault, I didn't restart travis before I merged CayleyDickson earlier today. I will submit a quickfix now.

Alizter:
Sorry about that I have created a PR #1198 so resolve that issue.

theories/Basics/Equivalences.v
jdchristensen: There is the `path_via` tactic which lets you specify one midpoint.  But I think that the new proof I pushed is pretty transparent.  It simply changes the RHS to the LHS in four easy steps.

theories/Basics/Equivalences.v
mikeshulman: Yes, I've been meaning to look back at that.  It ought to be possible to fix.

mikeshulman:
Does it really need a better name on its own, since it can always be referred to as `eisadj (f^-1)`?

jdchristensen:
Well, it's in the global namespace, and `other_adj` is pretty vague.

mikeshulman:
It doesn't need to be in the global namespace.  In fact we could just prove `isequiv_inverse` directly without stating `other_adj` as a lemma.

jdchristensen:
Good point.  Want me to change it to be like that?

jdchristensen:
I would get rid of the Section as well, then.

mikeshulman:
At the moment I don't see a downside of that, so sure, let's try it.

mikeshulman:
Actually, I realized there is a bit of a subtlety with opacity (`Defined` vs `Qed`).  The definition of `isequiv_inverse` needs to be transparent.  So unless we can fix the problem preventing `other_adj` from being opaque, we need to either define it separately as a lemma that we can prove with `Qed`, or surround the corresponding subgoal in the proof of `isequiv_inverse` with `abstract`.  (The latter has the disadvantage that all the periods need to be changed to semicolons, so that anyone wanting to step through it has to manually change them back.)

jdchristensen:
Actually, because of the problem in Card.v, the lemma needs to be opaque, but we probably don't want `isequiv_inverse` opaque.  I could maybe handle this with `abstract`, but then it seems like we couldn't step through the proof of the lemma anymore.  So maybe I should leave it like it is until the problem in Card.v is sorted out.

mikeshulman:
Haha, jinx!

jdchristensen:
Unless it's a quick fix to Card.v, let's leave that problem aside.  I can update #1091 to include inlining `other_adj` into `isequiv_inverse`.  And so I'll rename `other_adj` to `eisadj_other` in the meantime.  I'll push this after the build fix is merged, to trigger a new build of this PR.

Alizter:
@jdchristensen Alright, all fixed!

jdchristensen:
`other_adj` was also used in HoTTBook.v.

Alizter:
If we use `Defined` but then set it as opaque straight after, shouldn't this have a different behavior to `Qed`? AFAIK this just discourages coq from trying to unfold the term but it isn't really opaque. I could be wrong though.

mikeshulman:
An even more descriptive name would be `eisadj_isequiv_inverse`.

jdchristensen:
But the other way to look at it is that for a given equivalence `e`, there are two adjoint laws that hold, and this is the proof of the other one.  That's the context where I at one point needed this (and wanted it to compute), although after some rearranging I don't need it anymore.

mikeshulman:
Yes, that's true.

jdchristensen:
@Alizter Your approach might work, but if the problem in Card.v is fixed, then it won't be necessary.  So since there's no immediate need for this to be transparent, I think we should hold off on trying what you suggest.

theories/Basics/Equivalences.v
mikeshulman: This is a bit better, I guess.  I still like seeing the other endpoint at each step as we go through, rather than `?Goal0`.  I don't mind having to give the underscores, but you could avoid it with something like:
```
    refine (_ @ _); [ | apply ap_compose ].
    refine (_ @ _); [ | symmetry; apply (ap_homotopic_id (eisretr f)) ].
    refine (_ @ _); [ | symmetry; apply concat_pp_V ].
    symmetry; apply eisadj.
```
I don't find `assert (p := concat_A1p (eisretr f) (eisretr f b))` any less motivated than `apply (ap_homotopic_id (eisretr f))`.  Both of them pull a naturality square out of a hat when all we see in front of us is one side of that square.

jdchristensen: I think we're quibbling a bit here.  Remember that 24 hours ago, the proof was:

```
   (* First we set up the mess. *)
   rewrite <- (concat_1p (eissect _ _)).
   rewrite <- (concat_Vp (ap f^-1 (eisretr f (f (f^-1 b))))).
   rewrite (whiskerR (inverse2 (ap02 f^-1 (eisadj f (f^-1 b)))) _).
   refine (whiskerL _ (concat_1p (eissect _ _))^ @ _).
   rewrite <- (concat_Vp (eissect f (f^-1 (f (f^-1 b))))).
   rewrite <- (whiskerL _ (concat_1p (eissect f (f^-1 (f (f^-1 b)))))).
   rewrite <- (concat_pV (ap f^-1 (eisretr f (f (f^-1 b))))).
   apply moveL_M1.
   repeat rewrite concat_p_pp.
   (* Now we apply lots of naturality and cancel things. *)
   rewrite <- (concat_pp_A1 (fun a => (eissect f a)^) _ _).
   rewrite (ap_compose' f f^-1).
   rewrite <- (ap_p_pp _ _ (ap f (ap f^-1 (eisretr f (f (f^-1 b))))) _).
   rewrite <- (ap_compose f^-1 f).
   rewrite (concat_A1p (eisretr f) _).
   rewrite ap_pp, concat_p_pp.
   rewrite (concat_pp_V _ (ap f^-1 (eisretr f (f (f^-1 b))))).
   repeat rewrite <- ap_V; rewrite <- ap_pp.
   rewrite <- (concat_pA1 (fun y => (eissect f y)^) _).
   rewrite ap_compose', <- (ap_compose f^-1 f).
   rewrite <- ap_p_pp.
   rewrite (concat_A1p (eisretr f) _).
   rewrite concat_p_Vp.
   rewrite <- ap_compose.
   rewrite (concat_pA1_p (eissect f) _).
   rewrite concat_pV_p; apply concat_Vp.
```

While I agree in general that it can be helpful to see both sides of the equation, if you look at the first three steps of the above proof, they are all doing easy rewrites of the RHS: replacing `ap f (ap g ...)` with `ap (f o g) ...`, rewriting `ap h` when `h` is known to be homotopic to the identity, and cancelling paths.  I don't see how knowing the LHS is relevant.  I also don't think that the use of `ap_homotopic` can be a surprise, since the term you have at that point is `ap` applied to a function that is homotopic to the identity.

Regarding the proposal with `[ | ...]`, I considered that, but felt like it was unnecessarily raising the level of knowledge needed by the reader.

Maybe this is when I should mention my rewriting library.  With my tactics, the entire proof becomes:

```
    apply (equiv_inj (ap f)).
    rewriterV ap_compose.
    rewriter (ap_homotopic_id (eisretr f)).
    rewriter concat_pp_V.
    rewriter eisadj.
```

It produces a definitionally equivalent proof term, and you see both sides of the equation all the way along.  My plan was to bring these tactics up for discussion later this week, and I'd rather not hold up this pull request while we discuss my rewriting library.

theories/Basics/Equivalences.v
```diff
@@ -426,6 +326,101 @@ Global Instance isequiv_ap `{IsEquiv A B f} (x y : A)
...
+(** The inverse of an equivalence is an equivalence. *)
+Section EquivInverse.
+
+  Context {A B : Type} (f : A -> B) {feq : IsEquiv f}.
+
+  Theorem eisadj_other (b : B) : eissect f (f^-1 b) = ap f^-1 (eisretr f b).
+  Proof.
+    apply (equiv_inj (ap f)).
+    (* We will prove the equality as a composite of four paths, working right to left. *)
+    refine (_ @ _ @ _ @ _).
```
JasonGross: If you do
```suggestion
    refine (_ @ _ @ _ @ _); revgoals.
```
then you won't need to prefix the subgoals with numbers.  Feel free to do this, or not, as you'd prefer.

theories/Basics/Equivalences.v
mikeshulman: Well, it seems we have different aesthetics for tactic proofs.  (-:  But I'll stop complaining.

theories/Basics/Equivalences.v
jdchristensen: Thanks, that's a nice tool to know for this right-to-left situation!  I'd like to make this change, but will that reopen the discussion or be quickly accepted?  :smile:

mikeshulman: Speaking for myself, I'll quickly accept it.

PR #1190, opened 2020-01-26
renamed cubical library

Alizter:
Renamed Square and Cube to PathSquare and PathCube respectively. We have a library of squares in development in the wildcat branch currently so this is in preparation for that. This closes mikeshulman/HoTT#38.

PR #1198, opened 2020-01-29
Fixed CayleyDickson

Alizter:


Alizter:
So my inability to restart travis before merging a PR has caught up to me yet again. I merged #1152 earlier today, but it wasn't really compatible with #1190. This should fix that.

PR #1203, opened 2020-01-31
Add Theorem 4.7.7 explicitly

jdchristensen:
Is it worth having this iff statement explicitly when the two halves are already in the library?  I got confused and had trouble finding the pieces, which is why I'm proposing adding it.  An alternative would be to change HoTTBook.v to refer to both implications.  (Before this commit, it only pointed to one of the two implications.)


jdchristensen:
An argument in favour is that the two reverse implications are in different files.

PR #1204, opened 2020-02-02
Speed up typeclass search in .../Category/Sigma/Univalent.v

jdchristensen:
Reduces the file's type checking time from 17s to < 3s on my system.

### DIFF
--- a/theories/Categories/Category/Sigma/Univalent.v
+++ b/theories/Categories/Category/Sigma/Univalent.v
@@ -31,6 +31,8 @@ Section onobjects.
   : IsCategory (sigT_obj A Pobj).
   Proof.
     intros s d.
+    (* This makes typeclass search go faster. *)
+    pose @isequiv_compose.
     refine (isequiv_homotopic
               ((issig_full_isomorphic (sigT_obj A Pobj) _ _ o (issig_full_isomorphic A _ _)^-1)
                  o (@idtoiso A s.1 d.1)
@@ -114,8 +116,9 @@ Section onmorphisms.
   : IsCategory A'.
   Proof.
     intros s d.
+    (* Use [equiv_compose] rather than "o" to speed up typeclass search. *)
     refine (isequiv_homotopic
-              (iscategory_sigT_mor_helper^-1 o @idtoiso _ _ _)
+              (equiv_compose iscategory_sigT_mor_helper^-1 (@idtoiso _ _ _))
               _).
     intro x; apply path_isomorphic; cbn.
     destruct x; refine (path_sigma' _ 1 (contr _)).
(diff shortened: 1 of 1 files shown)


Alizter:
Great job Dan! I've tried before to speed up these files but with no success.

mikeshulman:
Yeah, that's great!

jdchristensen:
One of the things I noticed is that typeclass search took a long time to get around to trying `isequiv_compose`. Would something like the following be helpful globally? Or would Coq match one of the functions with `idmap`, causing looping?  Jason mentioned something about making Coq only apply a rule when the functions are known.

I tried this in Univalent.v, and it helped a bit in one spot but not enough.

```
    Hint Extern 0 (IsEquiv (_ o _)) => class_apply @isequiv_compose : typeclass_instances.
    Hint Extern 0 (IsEquiv (fun x => ?f (?g x))) => class_apply @isequiv_compose : typeclass_instances.
```

jdchristensen:
Also, why does closing the `on_both` section at the end of the file take 0.3s?  That and

`apply (ap (path_sigma_uncurried ...`

are currently the slowest two lines in the file.

mikeshulman:
I don't know why closing sections is sometimes slow.

I suspect that part of the problem with `isequiv_compose` is that `compose` is a notation rather than a definition.  We had a big discussion about that choice a long time ago and decided in #643 to do it that way, but I bet Coq would have an easier time applying typeclass inference to it if it were actually syntax.  Hottclasses does define a compose definition.

Alizter:
When closing a section, the terms are abstracted to take the context as arguments. If the terms in the section are particularly long, this might be a bottleneck. This is what I suspect.

mikeshulman:
It's hard to believe that simply adding a few abstractions in front of a term should be very slow.  But I guess I don't know how terms are stored.

SkySkimmer:
It also needs to replace the named variables by debruijn indices and add new arguments to occurences of constants, ie it needs to traverse the terms.
If you replace Defined by Qed it won't need to traverse the terms, so if it speeds up the section closing that means the slowness is very likely because of the traversal, otherwise very unlikely.

jdchristensen:
I don't know much about typeclass inference, but at least in this file it doesn't seem like it has trouble applying isequiv_compose.  It's that it tries so many other things before trying it.  Hence the suggestion to have a high priority hint to try isequiv_compose when it makes sense to.  Currently, isequiv_compose has priority 1000 and has no Hints as far as I can see.

Alizter:
I think the low priority was set so that coq didn't try isequiv_compose on idmap. Is there a way to tell coq to never try idmap?

JasonGross:
If you make `isequiv_idmap` (or whatever) not an instance and instead do `Hint Extern 0 (IsEquiv idmap) => ...` then it will never instantiate unknown things with idmap.

jdchristensen:
I think Coq inserts `fun x => x` as needed, without it being an instance, so I don't think we can avoid that.  Moreover, we _want_ the identity map to be inferred in many situations.  We just don't want isequiv_compose to use it.

A vanilla coq test of my first assertion is:

```
Axiom dumb : forall (A B C : Type) (g : B -> C) (f : A -> B), forall a1 a2, g (f a1) = g (f a2).

Theorem test {A B} (f : A -> B) : forall a1 a2, f a1 = f a2.
Proof.
  apply dumb.
Defined.

Print test.
(*
test = 
fun (A B : Type) (f : A -> B) => dumb A A B f (fun a2 : A => a2)
     : forall (A B : Type) (f : A -> B) (a1 a2 : A), f a1 = f a2
*)
```

jdchristensen:
@SkySkimmer It's hard to test your theory in this case, because the section doesn't typecheck unless most of the proofs end with Defined.

PR #1205, opened 2020-02-02
Equivalences and split-eso functors of 0-groupoids

mikeshulman:
Eventually we'll probably want equivalences of more general kinds of wild categories, which will use HasEquivs.  But 0-groupoids are a start (and already useful).

(This isn't on the wildcat branch yet.)


mikeshulman:
This is on top of #1206, just the last commit is additional.  (I didn't realize that at first -- the fact that `.vo` files hang around and get loaded by Coq even when I check out a different branch where their corresponding `.v` file doesn't exist always bites me.)

PR #1206, opened 2020-02-02
Path-groupoids and sigma-types as wild categories

mikeshulman:
I thought we had path-groupoids already, but I couldn't find them.  Maybe we just talked about it but didn't do it?

I know @mpopie worked on sigma-types during the wild categories week, but I think maybe it never got committed?  (And perhaps was discarded as eventually becoming an instance of displayed categories.)  I needed just the basic facts about them for an upcoming PR, and I think the definitions have changed, so I figured it was easier to reprove them.


Alizter:
Can you rebase this to make it easier to review?

mikeshulman:
yes, done.

PR #1207, opened 2020-02-02
The universal property of suspension, without funext

mikeshulman:
On top of #1205; see last 3 commits only (or wait to review until that's merged).

It's perhaps a bit of a stretch to say that this is an application of wild categories, when it's only using wild 0-groupoids that are just setoids, but I think the wild-category-theoretic perspective is helpful.

I need this right now just for separated localizations, but I expect it may be useful in Pointed too, e.g. using the `pForall` in the wildcat branch.

### DIFF
--- a/theories/Colimits/Pushout.v
+++ b/theories/Colimits/Pushout.v
@@ -151,6 +151,24 @@ Definition pushout_sym {A B C} {f : A -> B} {g : A -> C}
   : Pushout f g <~> Pushout g f :=
 equiv_adjointify pushout_sym_map pushout_sym_map sect_pushout_sym_map sect_pushout_sym_map.
 
+(** ** Functoriality *)
+
+Definition functor_pushout
+           {A B C} (f : A -> B) (g : A -> C)
+           {A' B' C'} (f' : A' -> B') (g' : A' -> C')
+           (h : A -> A') (k : B -> B') (l : C -> C')
+           (p : k o f == f' o h) (q : l o g == g' o h)
+  : Pushout f g -> Pushout f' g'.
+Proof.
+  unfold Pushout; serapply functor_coeq.
+  - exact h.
+  - exact (functor_sum k l).
+  - intros a; cbn.
+    apply ap, p.
+  - intros a; cbn.
+    apply ap, q.
+Defined.
+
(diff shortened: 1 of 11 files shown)


Alizter:
I was going to bring this up soon, but can we make IsGraph a parameter of the definition of wild category. This will allow us to define diagrams, then having certain (co)limits etc. Then we can show that pushouts are functors whenever they exist in a wild cat.

mikeshulman:
Yes, we should certainly do that.  Probably that should be done in the wildcat branch and then merged into master?

Alizter:
Can you rebase this?

mikeshulman:
Done

theories/Colimits/Pushout.v
```diff
@@ -183,6 +201,32 @@ Section EquivPushout.
 
 End EquivPushout.
 
+(** ** Contractibility *)
+
+(** The pushout of a span of contractible types is contractible *)
+
+Global Instance contr_pushout {A B C : Type} `{Contr A, Contr B, Contr C}
```
Alizter: I don't think there is an easy answer for the converse of this is there? I am not aware of any necessary conditions that makes a pushout contractible, but there are of course lots of sufficient ones you could choose.

theories/Pointed/pSusp.v
```diff
@@ -32,10 +32,9 @@ Definition psusp (X : pType) : pType
 
 (** ** Suspension Functor *)
 
-(* Definition of suspension functor *)
+(* Definition of pointed suspension functor *)
```
Alizter: Wasn't there a discussion about moving this into Suspension.v at some point? I suppose we could do that elsewhere.

theories/Colimits/Pushout.v
mikeshulman: No.

theories/Pointed/pSusp.v
mikeshulman: I did define the unpointed `functor_susp` in `Suspension.v`.

Alizter: I meant moving all the stuff about pointed suspensions into suspensions and out of Pointed.

PR #1208, opened 2020-02-02
Separated subuniverses

mikeshulman:
On top of #1207, see last commit only.  This defines RSUs of separated types and proves some results from section 2 of CORS.  For universe reasons I had to define "being the separated subuniverse" as a relation between two RSUs rather than an operation on them, and include it in the definition of a family of RSUs, which in turn meant that in order to prove anything about them I had to characterize them for all our examples, including localization; hence #1207.  The most interesting stuff here is the proofs from CORS in `ReflectiveSubuniverse.v`; thanks to @jdchristensen for helping me correct and simplify the proof of 2.26 in particular.

### DIFF
--- a/theories/Modalities/Localization.v
+++ b/theories/Modalities/Localization.v
@@ -1,9 +1,10 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 (** * Localization *)
 
-Require Import HoTT.Basics HoTT.Types.
+Require Import HoTT.Basics HoTT.Types HoTT.Cubical.
 Require Import Extensions EquivalenceVarieties.
 Require Import ReflectiveSubuniverse Accessible.
+Require Import Homotopy.Suspension.
 
 Local Open Scope nat_scope.
 Local Open Scope path_scope.
@@ -270,6 +271,13 @@ Proof.
   apply Xloc.
 Defined.
 
+(** A basic operation on local generators is the pointwise suspension. *)
+Definition susp_localgen (f : LocalGenerators@{a}) : LocalGenerators@{a}.
+Proof.
+  econstructor; intros i.
+  exact (functor_susp@{a a a a a a a a a a a a} (f i)).
+Defined.
+
 (** ** Localization as a HIT *)
 
 Module Export LocalizationHIT.
@@ -390,6 +398,38 @@ Module Localization_ReflectiveSubuniverses <: ReflectiveSubuniverses.
     apply ext_localize_ind@{a i j i k}; intros ?.
     apply ooextendable_over_const.
     apply Q_inO.
+  Defined.
+
+  (** The separated subuniverse corresponding to localization at [f] is localization at the pointwise suspension of [f].  This (and the following two definitions that prove it) is CORS Lemma 2.15, except that we can't use exactly the same proof since we have to avoid funext.  Instead we use the non-funext version proven using wild 0-groupoids in [Homotopy.Suspension]. *)
+  Definition IsSepFor@{u a} (O' O : ReflectiveSubuniverse@{u a}) : Type@{u}
+    := paths@{u} (Loc (susp_localgen (unLoc O))) O'.
+
+  Definition inO_paths_from_inSepO@{u a i iplus}
+             (O' O : ReflectiveSubuniverse@{u a}) (sep : IsSepFor O' O)
+             (A : Type@{i}) (A_inO : In@{u a i} O' A) (x y : A)
+    : In@{u a i} O (x = y).
+  Proof.
+    destruct O as [f]; destruct sep; unfold In, IsLocal in *; intros i; cbn in *.
+    specialize (A_inO i).
+    (** Unfortunately, [ooextendable_iff_functor_susp] has a ridiculous number of universe parameters.  Fortunately, most of them can be [i], and it's fairly easy to figure out which have to be [a].  But a few of them have to be something strictly larger than [i], which is why we included such a universe parameter [iplus] in these fields of [ReflectiveSubuniverse]. *)
+    assert (e := fst (ooextendable_iff_functor_susp@{a a i i a a a i i iplus i a a a i i iplus i i i i i i i i i i i i iplus i i i iplus i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i} (f i) _) A_inO (x,y)).
+    cbn in e.
+    refine (ooextendable_postcompose' _ _ _ _ e).
+    intros b.
+    symmetry; apply equiv_dp_const.
+  Defined.
+
+  Definition inSepO_from_inO_paths@{u a i iplus}
+             (O' O : ReflectiveSubuniverse@{u a}) (sep : IsSepFor O' O)
+             (A : Type@{i}) (e : forall (x y : A), In@{u a i} O (x = y))
+    : In@{u a i} O' A.
+  Proof.
+    destruct O as [f]; destruct sep; unfold In, IsLocal in *; intros i; cbn in *.
+    apply (ooextendable_iff_functor_susp@{a a i i a a a i i iplus i a a a i i iplus i i i i i i i i i i i i iplus i i i iplus i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i} (f i)).  
+    intros [x y].
+    refine (ooextendable_postcompose' _ _ _ _ (e x y i)).
+    intros b.
+    apply equiv_dp_const.
   Defined.
 
 End Localization_ReflectiveSubuniverses.
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -1,7 +1,7 @@
 (* -*- mode: coq; mode: visual-line -*- *)
 Require Import HoTT.Basics HoTT.Types.
 Require Import UnivalenceImpliesFunext EquivalenceVarieties Extensions HProp Fibrations NullHomotopy Pullback.
-Require Import HoTT.Tactics.
+Require Import HoTT.Tactics PathAny.
 Require Import HIT.Coeq Colimits.Pushout.
 Require Import Tactics.RewriteModuloAssociativity.
 
@@ -109,13 +109,14 @@ Module Type ReflectiveSubuniverses.
   (** As mentioned above, an implementation of this module type is a *family* of reflective subuniverses, indexed by the below type [ReflectiveSubuniverse].  If we just wrote [ReflectiveSubuniverse : Type], then it would end up parametrized by one universe, but in many examples the natural definition of the parametrizing type involves also a smaller universe, which would cause problems with Coq's strict polymorphism enforcement for module type implementations.  Thus, we use [Type2] instead, which takes two universe parameters. *)
   Parameter ReflectiveSubuniverse@{u a} : Type2@{u a}.
 
-  (** The universe parameters occurring in the definitions here play one of four roles, which we indicate consistently by [u], [a], [i], and [j].
+  (** The universe parameters occurring in the definitions here play one of the following roles, which we indicate consistently by the same letters.
 
   - [u] is the size of the parametrizing type [ReflectiveSubuniverse] (and, later, also [Modality]).
   - [a] is the size of smaller type-data occurring in that type, such as the family of generators for a localization.  This generally must be strictly smaller than [u].
   - [i] is the size of a type that we are reflecting or testing to be in the subuniverse.  This is generally at least as big as [a].
   - [j] is the size of a type that we are eliminating into (out of a type in [i]) with a universal property.  Also generally at least as big as [a].
-  - [k] is a universe at least as large as both [i] and [j], in which statements about types in both of them can live. *)
+  - [k] is a universe at least as large as both [i] and [j], in which statements about types in both of them can live.
+  - [iplus] is a universe that is strictly larger than [i]. *)
 
   Parameter O_reflector@{u a i} : forall (O : ReflectiveSubuniverse@{u a}),
                             Type2le@{i a} -> Type2le@{i a}.
@@ -133,7 +134,7 @@ Module Type ReflectiveSubuniverses.
   Parameter inO_equiv_inO@{u a i j k} :
       forall (O : ReflectiveSubuniverse@{u a}) (T : Type@{i}) (U : Type@{j})
              (T_inO : In@{u a i} O T) (f : T -> U) (feq : IsEquiv f),
-        (** We add an extra universe parameter that's bigger than both [i] and [j].  This seems to be necessary for the proof of repleteness in some examples, such as easy modalities. *)
+        (** We add an extra universe parameter that's at least as large as both [i] and [j].  This seems to be necessary for the proof of repleteness in some examples, such as easy modalities. *)
         let gei := ((fun x => x) : Type@{i} -> Type@{k}) in
         let gej := ((fun x => x) : Type@{j} -> Type@{k}) in
         In@{u a j} O U.
@@ -147,6 +148,22 @@ Module Type ReflectiveSubuniverses.
   Parameter extendable_to_O@{u a i j k}
   : forall (O : ReflectiveSubuniverse@{u a}) {P : Type2le@{i a}} {Q : Type2le@{j a}} {Q_inO : In@{u a j} O Q},
       ooExtendableAlong@{i i j k} (to O P) (fun _ => Q).
+
+  (** As defined in CORS, for a reflective subuniverse [O], the [O]-separated types are those whose identity types are in [O].  These form another reflective subuniverse [O'], but in general it might not be in the same family.  Thus, instead of considering [O] -> [O'] as an operation on reflective subuniverses, we consider it a relation between them; [IsSepFor O' O] says that [O'] is the [O]-separated subuniverse.  In addition, we can't just *define* this relation to say that the types in [O'] are the [O]-separated ones, because we want that to be true universe-polymorphically.  Thus, we assert the relation [IsSepFor] as part of the data of a family of reflective subuniverses, along with two more parameters saying that the types in [O'] are the [O]-separated ones at all universe levels.  *)
+  Parameter IsSepFor@{u a} : forall (O' O : ReflectiveSubuniverse@{u a}), Type@{u}.
+
+  (** These two have an extra universe parameter [iplus] that's strictly bigger than [i], because our current proofs of them in the case of localization seem to produce such a universe.  (It's unclear how necessary that is.) *)
+  Parameter inO_paths_from_inSepO@{u a i iplus}
+    : forall (O' O : ReflectiveSubuniverse@{u a}) (sep : IsSepFor O' O)
+             (A : Type@{i}) (A_inO : In@{u a i} O' A) (x y : A),
+      let gt := (Type1@{i} : Type@{iplus}) in
+      In@{u a i} O (x = y).
+
+  Parameter inSepO_from_inO_paths@{u a i iplus}
+    : forall (O' O : ReflectiveSubuniverse@{u a}) (sep : IsSepFor O' O)
+             (A : Type@{i}),
+      let gt := (Type1@{i} : Type@{iplus}) in
+      (forall (x y : A), In@{u a i} O (x = y)) -> In@{u a i} O' A.
 
 End ReflectiveSubuniverses.
 
@@ -173,7 +190,7 @@ Module Export Coercions.
 
 End Coercions.
 
-(** We assumed repleteness of the subuniverse in the definition.  Of course, with univalence this would be automatic, but we include it as a hypothesis since this is the only appearance of univalence in the theory of reflective subuniverses and non-lex modalities, and most or all examples can be shown to be replete without using univalence. *)
+(** We assumed repleteness of the subuniverse in the definition.  Of course, with univalence this would be automatic, but we include it as a hypothesis since most of the theory of reflective subuniverses and modalities doesn't need univalence, and most or all examples can be shown to be replete without using univalence. *)
 Arguments inO_equiv_inO {O} T {U} {_} f {_}.
 
 (** Being in the subuniverse is a mere predicate (by hypothesis) *)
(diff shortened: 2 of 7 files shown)


mikeshulman:
One other thing worth noting is that CORS 2.26 now replaces the characterization of path-types in `Truncation/Core.v`.

theories/Modalities/Localization.v
```diff
@@ -392,6 +400,38 @@ Module Localization_ReflectiveSubuniverses <: ReflectiveSubuniverses.
...
+
+  Definition inO_paths_from_inSepO@{u a i iplus}
+             (O' O : ReflectiveSubuniverse@{u a}) (sep : IsSepFor O' O)
+             (A : Type@{i}) (A_inO : In@{u a i} O' A) (x y : A)
+    : In@{u a i} O (x = y).
+  Proof.
+    destruct O as [f]; destruct sep; unfold In, IsLocal in *; intros i; cbn in *.
+    specialize (A_inO i).
+    (** Unfortunately, [ooextendable_iff_functor_susp] has a ridiculous number of universe parameters.  Fortunately, most of them can be [i], and it's fairly easy to figure out which have to be [a].  But a few of them have to be something strictly larger than [i], which is why we included such a universe parameter [iplus] in these fields of [ReflectiveSubuniverse]. *)
+    assert (e := fst (ooextendable_iff_functor_susp@{a a i i a a a i i iplus i a a a i i iplus i i i i i i i i i i i i iplus i i i iplus i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i i} (f i) _) A_inO (x,y)).
```
Alizter: Why are there so many universes?

jdchristensen:
Even just the last commit is pretty large, and I don't understand universes or modules very well, so this is hard going for me.  I'll try to find time to look it over, but it might not be for a few days.

mikeshulman:
@jdchristensen Yeah, I wasn't expecting you to try to figure out the universes and modules.  The interesting stuff from CORS is at about lines 2020-2200 of `ReflectiveSubuniverse`; I hope it should be fairly comprehensible if you just ignore the universe annotations.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2017,174 @@ Section ConnectedMaps.
...
+  (** Proposition 2.26 of CORS *)
+  Definition path_SepO {X : Type@{i}} (x y : X)
+    : O (x = y) -> (to O' X x = to O' X y)
+    := O_rec (O := O) (@ap X (O' X) (to O' X) x y).
+
+  Global Instance isequiv_path_SepO `{Univalence} {X : Type@{i}} (x y : X)
+    : IsEquiv (path_SepO x y).
+  Proof.
+    pose (Oxeq := fun y : X => (O (x = y); O_inO (x = y)) : Type_ O).
+    assert (Pb := fun y => (O_rec_beta@{Ou Oa i j j i j} Oxeq y)..1).
```
jdchristensen: Why `assert` and not `pose` here?  The docs say that `assert` in this form is equivalence to `pose proof`, but I don't see how that's different from `pose`.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2017,174 @@ Section ConnectedMaps.
...
+        apply (equiv_functor_sigma' 1%equiv); intros q; cbn.
+        exact (equiv_moveL_transport_V idmap (Pb y) _ q). }
+      intros p; destruct p.
+      reflexivity. }
+    { apply O_indpaths; intros p.
+      refine (_ @ (O_rec_beta _ _)^).
+      apply moveR_equiv_V.
+      destruct p; cbn.
+      reflexivity. }
+  Defined.
```
jdchristensen: I'm finding this proof hard to follow.  Maybe a few comments would help?

jdchristensen:
This is really great!  Thanks for working on it.  I have started reviewing it, but haven't finished.  I have to run, so I'll send my comments so far, and then will write more later.

Incidentally, it was easier to review this after I ran the `jdc-make-universes-invisible` emacs function below:

```
(defun jdc-make-invisible (regexp)
  "Make regions matching regexp invisible."
  (interactive)
  (save-excursion
    (let ((inhibit-point-motion-hooks t)
	  (case-fold-search t))
      (goto-char (point-min))
      (while (re-search-forward regexp nil t)
	(let* ((start (match-beginning 0))
	       (end (match-end 0)))
	  (overlay-put (make-overlay start end) 'invisible t))))))

(defun jdc-make-universes-invisible ()
  "Hide universe annotations in Coq files."
  (interactive)
  (jdc-make-invisible "@\{[a-z|<>= ]*\}"))
```

I'm not too happy with the `IsSepFor` relation, since it should be a defined construction.  I guess this is a side-effect of not being able to define parametrized modules.  And you can't prove existence of a new module given an existing one.  So now I see how this is another argument against the current set-up for organizing reflective subuniverses.

The proof of `inO_equiv` seems more complicated than expected, but I'm glad it's done.  And I think I understand that by proving it in general, you were able to remove it from Modality.v.  Nice.

More later.

mikeshulman:
Nice bit of emacs hackery there!

> I'm not too happy with the IsSepFor relation, since it should be a defined construction. I guess this is a side-effect of not being able to define parametrized modules. And you can't prove existence of a new module given an existing one. So now I see how this is another argument against the current set-up for organizing reflective subuniverses.

Yes.

> The proof of inO_equiv seems more complicated than expected

Yes, I had that impression too.  I was basically copying the proof in CORS; the issue is that an equivalence is defined with two homotopies whereas the fiber in question involves an equality between two pairs of functions, so you get funext redexes and also path_prod redexes.  Here's an alternative proof that uses a double fiber to avoid the path_prod redexes (I doubt there is any way to avoid the funext ones):

```
    Global Instance inO_equiv `{Funext} (A : Type@{i}) (B : Type@{j})
           `{In@{Ou Oa i} O A} `{In@{Ou Oa j} O B}
      : In@{Ou Oa k} O (A <~> B).
    Proof.
      refine (inO_equiv_inO@{Ou Oa k k k} _ (issig_equiv@{i j k} A B)).
      refine (inO_equiv_inO@{Ou Oa k k k} _ (equiv_functor_sigma equiv_idmap@{k}
                                 (fun f => equiv_biinv_isequiv@{i j k l} f))).
      pose (X := prod@{k k} (A->B) (prod@{k k} (B->A) (B->A))).
      pose (Y := hfiber@{k k} (fun fgh : X => (snd (snd fgh)) o (fst fgh)) idmap).
      pose (Z := hfiber@{k k} (fun fghp : Y => (fst fghp.1) o (fst (snd fghp.1))) idmap).
      refine (inO_equiv_inO'@{k k k} Z _). (** Introduces some extra copies of [k] by typeclass inference. *)
      { subst X Y Z; exact _. }
      srefine (equiv_adjointify _ _ _ _).
      - intros [[[f [g h]] p] q]; cbn in *.
        exact (f ; ((h ; ap10 p) , (g ; ap10 q))).
      - intros [f [[g p] [h q]]].
        exact (((f,(h,g)) ; path_arrow _ _ p) ; path_arrow _ _ q).
      - intros [f [[g p] [h q]]]; cbn.
        apply ap, path_prod; cbn.
        all:apply ap; apply path_forall, ap10_path_arrow.
      - intros [[[f [g h]] p] q]; cbn in *.
        serapply path_sigma; cbn.
        { serapply path_sigma; [ reflexivity | ].
          apply eta_path_arrow. }
        { unfold path_sigma; cbn.
          rewrite (transport_pr1_path_sigma_uncurried
                     _ (fun fgh:X => fst fgh o (fst (snd fgh)) = idmap)).
          cbn. apply eta_path_arrow. }
    Defined.
```

At least the definitions of the maps back and forth are now transparently just rearranging sigma-types and applying funext.  But in the second homotopy now a no-op transport appears and has to be gotten rid of, so the proof doesn't end up significantly shorter.  I'm not sure which I prefer.

> And I think I understand that by proving it in general, you were able to remove it from Modality.v.

Yes.  I think I didn't realize until I read CORS that arbitary RSUs were closed under `Equiv`.  Of course the proof for a modality is much easier, but since every modality is an RSU, once we've proven it for RSUs we don't need to re-prove it at all for modalities.

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: `assert (a := X)` is roughly equivalent to `pose (a := X); clearbody a`: it forgets the actual definition of `a` and remembers only its type.  `pose proof` behaves like `assert` but doesn't seem to let you give a name to the term.  (This is another instance, like `revert`/`generalize`/`generalize dependent`, where a tactic that sounds like a variant of another one actually behaves more like a variant of a differently named one.)

I sometimes use `assert` in this way instead of `pose` when the proof doesn't depend on the definition of the term, since I think it makes the context somewhat easier to read.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2017,174 @@ Section ConnectedMaps.
...
+    rapply inSepO_from_inO_paths.
+  Defined.
+
+  (** Proposition 2.18 of CORS.  Since we have universe-polymorphic RSUs, we can state it in the simpler form "[Type_ O] is in [O']".  Note the universe parameters: [Type_@{Ou Oa j i}] is a subtype of [Type@{i}], which is an element of [Type@{j}] (so that [i<j]).  The extra parameter [k] is required by separatedness. *)
+  Global Instance inSepO_typeO@{i j k} `{Univalence} : In O' (Type_@{Ou Oa j i} O).
+  Proof.
+    rapply inSepO_from_inO_paths@{Ou Oa j k}.
+    1:exact _.
+    (** For some reason, if we keep [A] and [B] in [Type_ O] and coerce them to [Type] (or write [A.1] and [B.1]), instead of destructing them into elements of [Type], then the universes get collapsed to [i=a] and [j=u].  I can't figure out why that is; maybe a bug in Coq? *)
+    intros [A ?] [B ?].
```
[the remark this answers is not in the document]
mikeshulman: Thanks.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2017,174 @@ Section ConnectedMaps.
...
+    pose (inO_equiv@{Ou Oa i i i j i i} O A B).
+    serapply (inO_equiv_inO'@{Ou Oa i j j} O (A <~> B) (equiv_compose'@{i j j} _ _)).
```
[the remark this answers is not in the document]
mikeshulman: Thanks.

theories/Modalities/ReflectiveSubuniverse.v
SkySkimmer: There's `pose proof term as name`

mikeshulman: Huh.  I don't suppose you have any idea why they decided against the more obvious `pose proof (a := X)`?

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Ok, I added some; perhaps too many.  (-:  I'm doing my best to follow the sketch you laid out by email for correcting the proof in CORS.

theories/Modalities/ReflectiveSubuniverse.v
jdchristensen: I now understand and agree with the motivation to hide the definition and retain only the type.  When I see `assert` I initially expect a goal to be generated, which is why I found its use here odd.  The `pose proof` form also sounds odd to be, because it does the opposite of what it says, as it shows the type but not the proof...  So staying with `assert` is fine with me.

theories/Modalities/ReflectiveSubuniverse.v
jdchristensen: Thanks for the comments!

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Yeah, I agree the naming leaves a lot to be desired.  I suspect that what whoever had in mind is a distinction between "elements of sets", for which the term matters, and "proofs of propositions", for which the term doesn't matter, so that `pose proof` is posing a "proof" whose value is irrelevant whereas `pose` is posing a term whose value is relevant.  But that could be way off.

jdchristensen:
I'm not sure which version I prefer.  They both have pros and cons.  If they are roughly the same, then I'd say to follow CORS, since someone (like me) might follow along in the paper to help understand the formalization.  If you stick with your original proof, one idea is to split the definition of `U` into three lines.  It ends up about the same size, but I think it's easier to parse in steps and without the nested `fst` and `snd` functions:

```
      transparent assert (c : (prod@{k k} (A->B) (prod@{k k} (B->A) (B->A)) -> prod@{k k} (A -> A) (B -> B))).
      1:intros [f [g h]]; exact (h o f, f o g).
      pose (U := hfiber@{k k} c (idmap, idmap)).
```

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2083,39 +2083,55 @@ Section Separated.
...
+    (* First we define a family [P : O' X -> Type] of [O]-modal types that extends [O (x = ?) : X -> Type], using the fact that the universe of [O]-modal types is [O']-modal.  [Pb] is its non-definitional computation rule that we will have to transport across.  *)
     pose (Oxeq := fun y : X => (O (x = y); O_inO (x = y)) : Type_ O).
     assert (Pb := fun y => (O_rec_beta@{Ou Oa i j j i j} Oxeq y)..1).
     set (P := O_rec Oxeq) in *; subst Oxeq; cbn in *.
+    (* We will first construct an equivalence [P y' <~> (to O' X x = y')] by a version of encode-decode, then show that when [y' = to O' X y] this equivalence is homotopic to [path_SepO]. *)
     serapply isequiv_homotopic'.
-    { refine (_ oE equiv_path _ _ (Pb y)^).
+    { (* First we generalize from [y:X] to [y':O' X]. *)
+      refine (_ oE equiv_path _ _ (Pb y)^).
       generalize (to O' X y); clear y.
```
jdchristensen: Insert `as y'` before the semicolon?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2083,39 +2083,55 @@ Section Separated.
...
+      (* Now by a standard result, it suffices to show that [P] is a pointed predicate whose total space. *)
```
jdchristensen: Add "is contractible" at the end.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2083,39 +2083,55 @@ Section Separated.
...
-      exists (to O' X x; transport idmap (Pb x)^ (to O _ 1%path)).
-      apply sig_ind.
+      (* The same point supplies the center of contraction. *)
+      exists (to O' X x; transport idmap (Pb x)^ (to O _ idpath)).
+      intros [y' p]; revert y' p.
+      (* To construct a contraction, we first observe that it suffices to return to the case [y:X]. *)
       refine ((extension_from_inO_sigma
                  O' (fun (w:O' X) => forall (p: (P w).1),
                          (to O' X x ; transport idmap (Pb x)^ (to O (x = x) 1%path))
                          = (w ; p)) _).1).
```
jdchristensen: The long type family argument is just repeated from the goal, so I think it's more clear to replace it with `_`.  That doesn't slow things down, but either way, this line takes ~0.3s.  Can we pose a hint or do something else to speed it up a bit?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2083,39 +2083,55 @@ Section Separated.
...
-      revert p; refine ((extension_from_inO_sigma O _ _).1).
-      { refine (inO_equiv_inO' _ (to O' X x = to O' X y) _).
-        pose (e := fun p:O (x = y) => (equiv_path_sigma (fun a => (P a).1)
-                                                        (to O' X x; transport idmap (Pb x)^ (to O (x = x) 1%path))
-                                                        (to O' X y; transport idmap (Pb y)^ p))^-1%equiv).
+      (* Now we have to transport [p] again across the computation rule [Pb]. *)
+      equiv_intro (equiv_path _ _ (Pb y)^) p; revert p.
+      (* We claim that once again, it suffices to assume [p:x=y]. *)
+      refine ((extension_from_inO_sigma O _ _).1).
+      { cbn.
```
jdchristensen: Put this `cbn` on a line before the `refine`?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2083,39 +2083,55 @@ Section Separated.
...
-        intros p.
+        refine ((equiv_sigma_contr _)^-1%equiv). intros p.
         refine (@contr_equiv' _ _ _ (@contr_basedpaths (O (x = y)) _)).
         apply (equiv_functor_sigma' 1%equiv); intros q; cbn.
         exact (equiv_moveL_transport_V idmap (Pb y) _ q). }
+      (* Now it's easy by path induction. *)
       intros p; destruct p.
       reflexivity. }
+    (* Finally, since [path_SepO] is the unique map making a commutative triangle with [to O' X], to show that the above equivalence is homotopic to [path_SepO] it suffices to show that it also makes that triangle commute. *)
     { apply O_indpaths; intros p.
```
jdchristensen: Adding `unfold equiv_path_from_contr.` right after the `{` makes it easier to see what's going on in the next few steps.

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Intriguing; I'm pretty sure wouldn't have given the argument in the first place if it weren't necessary, but I guess the proof has changed since that time so that it's no longer necessary.

mikeshulman:
All good suggestions, thanks!

theories/Modalities/ReflectiveSubuniverse.v
SkySkimmer: @herbelin may know about historical naming decisions.

herbelin: Hi, the naming was indeed as @mikeshulman is suspecting (`pose proof` is posing a "proof" whose value is irrelevant whereas `pose` is posing a term whose value is relevant).

I'm not against `pose proof (a := X)` which could be implemented too (by analogy with `pose` and `assert`. I did not retain it because there is the more general (but maybe less intuitive) form `pose proof X as pattern`.

herbelin: As it seems to be a natural expectation, I finally made coq/coq#11522 to support the `pose proof (a := X)` syntax as an alternative to `pose proof X as a` and `assert (a := X)`.

mikeshulman: Thanks!

Alizter:
Can you rebase this again?

mikeshulman:
Done.

Of course, since Dan is still working on reading this one over, don't merge it until he's done.

Alizter:
Of course.

theories/Modalities/Localization.v
Alizter: Makes you wish we had implicit universe arguments.

mikeshulman: Isn't that what happens when you don't give any universe annotation?

Alizter: I'm thinking something more specific like the arguments command. We should be able to choose which universe parameters to give, from which coq can infer the rest. At the moment its all or nothing.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2016,187 @@ Section ConnectedMaps.
...
+      generalize (to O' X y) as y'; clear y.
+      (* Now by a standard result, it suffices to show that [P] is a pointed predicate whose total space is contractible. *)
+      apply equiv_path_from_contr.
+      (* The point is just the image of [idpath x], but we have to transport it across [Pb] to put it in the right type. *)
+      1:exact (transport idmap (Pb x)^ (to O _ idpath)).
+      (* The same point supplies the center of contraction. *)
+      exists (to O' X x; transport idmap (Pb x)^ (to O _ idpath)).
+      intros [y' p]; revert y' p.
+      (* To construct a contraction, we first observe that it suffices to return to the case [y:X]. *)
+      refine ((extension_from_inO_sigma O' _ _).1).
```
jdchristensen: This line can be sped up with `pose @inSepO_sigma; pose @inO_paths_from_inSepO; pose @inO_forall.`  All three are necessary.

Alizter:
I will merge once you get a chance to look at Dan's recent comments.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2016,187 @@ Section ConnectedMaps.
...
+  Global Instance SepO_inverts_functor_hfiber `{Univalence}
+         {Y X : Type@{i} } (f : Y -> X) (x : X)
+    : O_inverts O (functor_hfiber@{i i i i} (fun y => (to_O_natural O' f y)^) x).
+  Proof.
+    serapply isequiv_homotopic'.
+    - unfold hfiber.
+      refine (_ oE (equiv_inverse (equiv_O_sigma_O O _))).
+      pose (fun oy:O' Y => O_functor@{Ou Oa i i i i i} O' f oy = to O' X x).
+      refine (equiv_functor_sigma_to_SepO@{i i iplus}
+                (fun oy:O' Y => O_functor@{Ou Oa i i i i i} O' f oy = to O' X x) oE _).
```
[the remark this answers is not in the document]
mikeshulman: Not sure why those were still hanging around.  In fact I don't think the `pose` was doing anything at all; it must have been left over from some experimentation.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1919,6 +2016,187 @@ Section ConnectedMaps.
...
+      apply equiv_O_functor. 
+      refine (equiv_functor_sigma' equiv_idmap _); intros y; cbn.
```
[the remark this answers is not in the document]
mikeshulman: Thanks for that reminder!

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Okay, I suppose that's worth it.  I `clear`ed them afterwards so they don't clutter up the context for the rest of the proof.  It would be nice if there were a cleaner way to pass hints to typeclass inference.

mikeshulman:
Ok, all done.  Thanks very much Dan!

PR #1210, opened 2020-02-03
minor simplifications to Sigma/Univalent.v

jdchristensen:


jdchristensen:
I checked that the typeclass inference was still fast with the remaining two pose commands, but slow if either one is removed.

PR #1214, opened 2020-02-05
Make equiv_sigma_assoc and contr_basedpaths easier to use

mikeshulman:
I finally realized that the reason Coq is often unable to guess the type family `Q` in `equiv_sigma_assoc` is that its domain is a sigma-type.  Defining a variant `equiv_sigma_assoc'` in which `Q` is instead a curried function of two (dependent) variables seems to remove the need to ever give it explicit arguments.  (Indeed, I'm tempted to make `P` and `Q` globally implicit arguments to both of them, but I didn't go that far yet.)

I also realized that the reason Coq is sometimes unable to find `contr_basedpaths` by typeclass inference is that it's looking at an eta-short form `sig (@paths A x)` which it's unable to unify with `{y:A & x = y}` since the latter is a notation for the eta-long form `sig (fun y => @paths A x y)`.  (I believe that when Coq displays `{y:_ & x = y}` it means it has the eta-short form `sig (@paths A x)`, although I don't know exactly why.)   Defining an eta-short variant seems to allow typeclass inference to find it all the time.

### DIFF
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -45,6 +45,10 @@ Proof.
   intros [y p]; apply path_basedpaths.
 Defined.
 
+(* Sometimes we end up with a sigma of a one-sided path type that's not eta-expanded, which Coq doesn't seem able to match with the previous instance. *)
+Global Instance contr_basedpaths_etashort {X : Type} (x : X) : Contr (sigT (@paths X x)) | 100
+  := contr_basedpaths x.
+
 Definition path_basedpaths' {X : Type} {x y : X} (p : y = x)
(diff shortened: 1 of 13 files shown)


mikeshulman:
Actually first I tried to just replace the old version with the new one, but that didn't work.  I think a reasonable rule of thumb is that if what's visible in the goal is the left-hand side of the equivalence, where Q is curried, then try the primed version; whereas if what's visible in the goal is the right-hand side, where Q is uncurried, then try the unprimed version.  However, I'm not sure that that always works.

Alizter:
LGTM. Lets merge when travis is done.

Issue #1209, opened 2020-02-03
Ditch the modules; free the modalities


mikeshulman:
Currently our library for modalities (including reflective subuniverses) uses Coq's modules and module types: instead of the obvious approach of each modality being an element of a record type called (say) `Modality`, each "family of modalities" (truncations, localizations, etc.) is a module that instantiates a particular "module type".  The reason for doing it this way was that an individual element of a record type cannot be universe-polymorphic, although the record type itself can be; thus with the obvious approach we would have a separate type of "modalities on universe U" for each U.  Which is not a problem for many purposes, but it becomes an issue when we want to say things like "the universe of modal types is modal" (for a lex modality) or "the universe of modal types is separated" (for any RSU); the latter can be finessed as in CORS, but the former less so.  In addition, when working with practical examples (notably truncation) we want to know that a type X has only one truncation, which happens to live in the same universe as X, but satisfies a polymorphic elimination principle mapping into types in any universe.  I wrote a long blog post about this [almost 5 years ago](https://homotopytypetheory.org/2015/07/05/modules-for-modalities/).

This way of representing modalities does however have notable problems.  For instance:
- It prevents us from talking about modalities that might exist only on one universe.  (E.g. modalities on an [impredicative universe](https://homotopytypetheory.org/2018/11/26/impredicative-encodings-part-3/) generated by a large family.)
- It prevents us from viewing reflective subuniverses as reflective subcategories of the wild category of types.
- It forces us to keep track of universe parameters explicitly in many places, which is at best annoying.  This is necessary because an instantiation of a module type has to have _exactly_ the same universe parameters as the module type was declared with.
- It makes it hard to relate modalities from different "families" to each other.  For instance, I recently wanted to define a notion of "exact sequence" parametrized by a modality and specialize it to both the identity modality (yielding fiber sequences) and the (-1)-truncation (yielding ordinary exact sequences of pointed sets / groups), but those modalities don't naturally live in the same family.
- It's inflexible and hard to extend.  For instance, to define separated subuniverses in #1208 I had to add a relation of "being separated for" to the definition of a family of modalities, and then instantiate it at all the example families.
- It presents a high bar to entry for people wanting to use modalities, let alone generalize and improve the modalities library: they have to understand both modules and universe polymorphism, which are two of the most finicky and least-understood aspects of Coq (even by the Coq developers).
[...]

PR #1216, opened 2020-02-06
removed funext from proof of O_inverts_functor_sigma_to_SepO

### LINKED ISSUE #1213: Remove funext from O_inverts_functor_sigma_to_SepO (CORS 2.27)
Based on an email discussion with @jdchristensen and Luis, there is a simpler proof.  By `extendable_toSepO_inO` (CORS 2.22) and `conn_map_from_extension_elim`, the unit of L'-localization is L-connected.  Now apply `conn_map_functor_sigma` (7.5.12 in the book).

Alizter:
closes #1213

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -2122,23 +2122,27 @@ Section Separated.
     : O (x = y) <~> (to O' A x = to O' A y)
     := Build_Equiv _ _ _ (isequiv_path_SepO x y).
 
+  Global Instance conn_map_toSepO@{i} {X : Type@{i} }
+    : IsConnMap O (to O' X).
+  Proof.
+    apply conn_map_from_extension_elim@{Ou Oa i i i i}.
+    intros Q f h.
+    exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
+  Defined.
+
   (** Lemma 2.27 of CORS. *)
   Global Instance O_inverts_functor_sigma_to_SepO@{i j}
          {X : Type@{i} } (P : O' X -> Type@{j})
     : O_inverts O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
   Proof.
     apply O_inverts_conn_map@{Ou Oa i i i i i i i i i}.
-    apply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
-    1: exact _.
-    apply conn_map_from_extension_elim@{Ou Oa i i i i}.
-    intros Q f h.
-    exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
+    serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
   Defined.
 
-  Definition equiv_functor_sigma_to_SepO@{i j jplus} `{Funext}
+  Definition equiv_functor_sigma_to_SepO@{i j jplus}
              {X : Type@{i} } (P : O' X -> Type@{j})
     : (O {x : X & P (to O' X x)}) <~> (O {ox : O' X & P ox})
-    := Build_Equiv _ _ _ (O_inverts_functor_sigma_to_SepO@{i j jplus} P).
+    := Build_Equiv _ _ _ (O_inverts_functor_sigma_to_SepO@{i j} P).
 
   (** Corollary 2.29 of CORS *)
   Global Instance SepO_inverts_functor_hfiber `{Univalence}
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -2130,22 +2130,16 @@ Section Separated.
     exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
   Defined.
 
+  (** A strengthening of Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
   Global Instance conn_map_functor_sigma_toSepO@{i j}
          {X : Type@{i} } (P : O' X -> Type@{j})
-    : IsConnMap O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
-  Proof.
-    serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
-  Defined.
-
-  (** Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
-  Global Instance O_inverts_functor_sigma_to_SepO@{i j}
-         {X : Type@{i} } (P : O' X -> Type@{j})
-    : O_inverts O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)) := _.
+    : IsConnMap O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap))
+    := conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i} _ _ _.
 
   Definition equiv_functor_sigma_to_SepO@{i j}
              {X : Type@{i} } (P : O' X -> Type@{j})
     : (O {x : X & P (to O' X x)}) <~> (O {ox : O' X & P ox})
-    := Build_Equiv _ _ _ (O_inverts_functor_sigma_to_SepO@{i j} P).
+    := Build_Equiv _ _ (O_functor O (functor_sigma@{i j i j} _ _)) _.
 
   (** Corollary 2.29 of CORS *)
   Global Instance SepO_inverts_functor_hfiber `{Univalence}
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -2139,7 +2139,7 @@ Section Separated.
     serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
   Defined.
 
-  Definition equiv_functor_sigma_to_SepO@{i j jplus}
+  Definition equiv_functor_sigma_to_SepO@{i j}
              {X : Type@{i} } (P : O' X -> Type@{j})
     : (O {x : X & P (to O' X x)}) <~> (O {ox : O' X & P ox})
     := Build_Equiv _ _ _ (O_inverts_functor_sigma_to_SepO@{i j} P).
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -2126,11 +2126,11 @@ Section Separated.
     : IsConnMap O (to O' X).
   Proof.
     apply conn_map_from_extension_elim@{Ou Oa i i i i}.
-    intros Q f h.
+    intros Q Q_inO h.
     exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
   Defined.
 
-  (** Lemma 2.27 of CORS. *)
+  (** Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
   Global Instance O_inverts_functor_sigma_to_SepO@{i j}
          {X : Type@{i} } (P : O' X -> Type@{j})
     : O_inverts O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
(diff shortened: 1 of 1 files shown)


mikeshulman:
I think I'd incline towards stating the L-connectivity of the L'-unit as a separate lemma first, for clarity.

Alizter:
OK, I will do that.

mikeshulman:
And actually we might as well upgrade the lemma to state `L`-connectivity: since `O_inverts_conn_map` is a global instance the CORS version should be automatically derivable.

mikeshulman:
There's something more general going on here too, although it may be hard to represent in the library unless we do #1209.  Consider two arbitrary reflective subuniverses `L` and `L'`.  Then I expect that the following are equivalent:

- The sigma of a family of `L`-modal types over an `L'`-modal type is `L'`-modal.
- Sections of `L`-modal families extend along the `L'`-unit.
- Sections of `L`-modal families extend uniquely along the `L'`-unit.
- The `L'`-unit is `L`-connected.

When `L`=`L'`, this reduces to the equivalence between various conditions (from the book or RSS) for an RSU to be a modality, while these results when `L'`=`Sep(L)` in CORS have essentially the same proofs (starting with 2.21, which proves the first one).

Note that these conditions also imply that every `L`-modal type is `L'`-modal.  So if we write them as `L << L'` then `<<` is a refinement of the containment order `<=` on RSUs, and `L` is a modality iff `L << L`.  This kind of reminds me of the [way below](https://ncatlab.org/nlab/show/continuous+poset) relation from order theory, and the fact that an open `U` in a locale is compact iff `U << U`.

I expect there is a similar binary relation `<<<` that generalizes some of the conditions for lex-ness from RSS and also CORS 2.18 and 2.26, so that `L <<< Sep(L)` always, and `L` is lex iff `L <<< L`.

Alizter:
Typeclass inference cannot guess the correct universe levels here however.

Alizter:
Also did you know that you can give underscores to universe parameters? I just found this out.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
     : O (x = y) <~> (to O' A x = to O' A y)
     := Build_Equiv _ _ _ (isequiv_path_SepO x y).
 
-  (** Lemma 2.27 of CORS.  This is the proof given in CORS, which requires funext.  An alternative proof not requiring funext should be possible, using the fact that [to O' X] is [O]-connected. *)
-  Global Instance O_inverts_functor_sigma_to_SepO@{i j jplus} `{Funext}
+  Global Instance conn_map_toSepO@{i} {X : Type@{i} }
+    : IsConnMap O (to O' X).
+  Proof.
+    apply conn_map_from_extension_elim@{Ou Oa i i i i}.
+    intros Q f h.
```
jdchristensen:
```suggestion
    intros Q Q_inO h.
```

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
...
+    exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
```
jdchristensen: I think it would be more clear to have a helper function that goes from ooExtendableAlong to ExtensionAlong.  Maybe it could even be a typeclass instance.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
...
+  Defined.
+
+  (** Lemma 2.27 of CORS. *)
```
jdchristensen:
```suggestion
  (** Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
```

jdchristensen: I also think it would be better to state the stronger result that the map is L-connected.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
...
-    apply (O_inverts_by_yoneda@{Ou Oa j j j j jplus j j j j} O).
-    intros Z ?.
-    serapply isequiv_homotopic'.
-    { refine (_ oE _ oE _).
-      3:apply equiv_inverse.
-      1,3: serapply equiv_sigT_ind@{i j j jplus}.
-      serapply equiv_toSepO_inO@{i j j jplus}. }
-    reflexivity.
+    apply O_inverts_conn_map@{Ou Oa i i i i i i i i i}.
+    serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
```
jdchristensen: Those universe variables are so annoying!

Alizter: It might be possible to reduce their number by carefully annotating everything.

jdchristensen:
> There's something more general going on here too, although it may be hard to represent in the library unless we do #1209. 

Would something like the following work?

```
Module RSUPairs (Os O's : ReflectiveSubuniverses).

    Module Import Os_Theory := ReflectiveSubuniverses_Theory Os.
    Module Import O's_Theory := ReflectiveSubuniverses_Theory O's.

    Definition related (O : Os.ReflectiveSubuniverse) (O' : O's.ReflectiveSubuniverse)
      := forall (X : Type) `{In O' X} (P : X -> Type) `{forall x, Os.In O (P x)},
         In O' { x : X & P x }.
```

It seems like `In` by default means `O's.In`, so the other use of `In` needs to be prefixed with `Os`.

Maybe this technique could be used to handle results about separated RSUs without the extra data being stored within a single family of RSUs.

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Perhaps.  That sounds like a different PR though.

theories/Modalities/ReflectiveSubuniverse.v
mikeshulman: Yes, there's something of a tradeoff -- if you postpone universe annotations until the last possible moment when you actually _need_ them, you generally find that all of your lemmas have an absurd number of universe parameters you have to pass explicitly.  But the other extreme is to annotate all the lemmas to keep their universe parameters manageable, and in that case before you know it everything in the library has universe annotations.

mikeshulman:
That might work.  But at the moment I'd rather put effort into finding out whether #1209 will work.

theories/Modalities/ReflectiveSubuniverse.v
spitters: @mattam82 started working on a universe minimization algorithm, but I believe the work is currently stalled.

jdchristensen:
> Typeclass inference cannot guess the correct universe levels here however.

Then maybe stating both the stronger L-connected version and the L-equivalence corollary makes sense?  (I didn't realize that the universe variables can get in the way what seems like a trivial typeclass inference...)

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
...
   Defined.
 
-  Definition equiv_functor_sigma_to_SepO@{i j jplus} `{Funext}
+  Definition equiv_functor_sigma_to_SepO@{i j jplus}
```
mikeshulman: You can probably get rid of the `jplus` universe here too, right?  (It's interesting that the new proof doesn't require that extra large universe any more too.)

mikeshulman:
Yes, I agree with stating both.

theories/Modalities/ReflectiveSubuniverse.v
Alizter:
```suggestion
  Definition equiv_functor_sigma_to_SepO@{i j}
```

Alizter:
I think that's all done. Let me know if I missed something.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,27 @@ Section Separated.
...
+  Proof.
+    apply conn_map_from_extension_elim@{Ou Oa i i i i}.
+    intros Q Q_inO h.
+    exact (fst (extendable_toSepO_inO@{i i i Ou Ou Ou} _ 1%nat) h).
+  Defined.
+
+  (** Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
+  Global Instance O_inverts_functor_sigma_to_SepO@{i j}
          {X : Type@{i} } (P : O' X -> Type@{j})
     : O_inverts O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
```
jdchristensen: Mike and I both thought that you should state the stronger result using `IsConnMap` as well as the special case `O_inverts`.  This splits the two-line proof into two one-line proofs, which seems silly, but it would be good to state the stronger result.  Are you sure that you can't *only* state the stronger result, with the `O_inverts` version being found by typeclass resolution as needed?

Alizter: @jdchristensen Am I confused or is the IsConnMap result you are talking about above this as conn_map_toSepO? If you do a typeclass search in the proof of O_inverts_functor_sigma_to_SepO you will see that it doesn't use the correct universes, so I have to hand hold it. Unless I am not following what either you are saying at all.

jdchristensen: The line above these comments could be strengthened to have `IsConnMap` instead of `O_inverts`.  The proof would then be one line shorter, and the current statement would be a special case, using typeclass inference.

Alizter: Oh I see now, sorry you were being completely clear. I will have a look at it tomorrow.

jdchristensen:
```suggestion
    : IsConnMap O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
```
(The name of the lemma would change too, of course.)

Alizter:
OK so I have separated the lemmas. The proof of `conn_map_functor_sigma_toSepO` can be found by typeclasses but there are still many unbound universe parameters. I have had to give it's proof explicitly, though it is only one line.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -2124,25 +2124,30 @@ Section Separated.
...
-      serapply equiv_toSepO_inO@{i j j jplus}. }
-    reflexivity.
+    serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
   Defined.
 
-  Definition equiv_functor_sigma_to_SepO@{i j jplus} `{Funext}
+  (** Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
+  Global Instance O_inverts_functor_sigma_to_SepO@{i j}
+         {X : Type@{i} } (P : O' X -> Type@{j})
+    : O_inverts O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)) := _.
```
jdchristensen: I still think it might be better to drop `O_inverts_functor_sigma_to_SepO`, since it will be deduced using typeclass resolution.  For example, when we prove something is contractible, we don't also record the fact that it is an hprop, since that can be deduced. So I would propose that this section be replaced with:

```
  (** A strengthening of Lemma 2.27 of CORS.  We provide a different proof that doesn't require [Funext]. *)
  Global Instance conn_map_functor_sigma_toSepO@{i j}
         {X : Type@{i} } (P : O' X -> Type@{j})
    : IsConnMap O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap)).
  Proof.
    serapply conn_map_functor_sigma@{Ou Oa i i i i i i i i i i i i i i i i i i i i i Ou i i}.
  Defined.

  Definition equiv_functor_sigma_to_SepO@{i j}
             {X : Type@{i} } (P : O' X -> Type@{j})
    : (O {x : X & P (to O' X x)}) <~> (O {ox : O' X & P ox})
    := Build_Equiv _ _ (O_functor O (functor_sigma@{i j i j} (Q := P) (to O' X) (fun x => idmap))) _.
```

The last underscore in the last line is where the deleted lemma is deduced.

jdchristensen: In fact, the last line can be simplified to
```
    := Build_Equiv _ _ (O_functor O (functor_sigma@{i j i j} _ _)) _.
```

Alizter: Nice!

Alizter:
There doesn't seem to be any getting away from those universe parameters.

mikeshulman:
Needs rebasing.

Alizter:
rebased

mikeshulman:
Did you compile it after rebasing?  It doesn't work for me.

Alizter:
There seem to be more universes this time...

Alizter:
@mikeshulman All good now.

PR #1218, opened 2020-02-06
replaced ap2 with ap011 and removed ap2

### LINKED ISSUE #1193: Delete ap2 and replace its uses with ap011
ap2 would be more properly a shorthand for ap02.  See discussion at #1186.

Alizter:
closes #1193

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -732,7 +732,7 @@ Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a 
   : D x2 (p # y) (p # z)
   :=
   match p with idpath => w end.
-
+(** ap011 is defined in Overture.v *)
 (** It would be nice to have a consistent way to name the different ways in which this can be dependent.  The following are a sort of half-hearted attempt. *)
 
 Definition ap011D {A B C} (f : forall (a:A), B a -> C)
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -732,7 +732,16 @@ Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a 
   : D x2 (p # y) (p # z)
   :=
   match p with idpath => w end.
-(** ap011 is defined in Overture.v *)
+
+(** *** [ap] for curried two variable functions *)
+
+Definition ap011 {A B C} (f : A -> B -> C) {x x' y y'} (p : x = x') (q : y = y')
+  : f x y = f x' y'.
+Proof.
+  destruct p, q.
+  reflexivity.
+Defined.
+
 (** It would be nice to have a consistent way to name the different ways in which this can be dependent.  The following are a sort of half-hearted attempt. *)
 
 Definition ap011D {A B C} (f : forall (a:A), B a -> C)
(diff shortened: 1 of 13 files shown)


theories/Basics/PathGroupoids.v
```diff
@@ -733,12 +733,6 @@ Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a
   :=
   match p with idpath => w end.
 
-(** *** [ap] for multivariable functions *)
-
-Definition ap011 {A B C} (f : A -> B -> C) {x x' y y'} (p : x = x') (q : y = y')
-: f x y = f x' y'
-:= ap11 (ap f p) q.
-
```
jdchristensen: Maybe add a comment saying that `ap011` is defined in Overture.v?

theories/Cubical/PathCube.v
```diff
@@ -699,8 +699,9 @@ Defined.
 (* Uncurry a function in sq_ap2 *)
 Definition sq_ap_uncurry {A B C} (f : A -> B -> C)
   {a a' : A} (p : a = a') {b b' : B} (q : b = b')
-  : PathCube (sq_ap (uncurry f) (sq_prod hr vr)) (sq_ap2 f p q)
-  (ap_uncurry _ _ _) (ap_uncurry _ _ _) (ap_uncurry _ _ _) (ap_uncurry _ _ _).
+  : PathCube (sq_ap (uncurry f) (sq_prod hr vr)) (sq_ap011 f p q)
+    (sq_G1 (ap_uncurry _ _ _)) (sq_G1 (ap_uncurry _ _ _))
+    (sq_G1 (ap_uncurry _ _ _)) (sq_G1 (ap_uncurry _ _ _)).
```
jdchristensen: Just curious why `ap_uncurry` was changed, making this part a bit more complicated.

Alizter: Well, it was a degenerate square before. We can now use ap_uncurry as a path, which might be useful elsewhere in the future. I would like to change the definition of sq_G1 to something simpler, at the time I was trying to get the proof of it being an equivalence to be found by typeclasses but now I am thinking this has made it unnecessarily complicated.

theories/Basics/PathGroupoids.v
```diff
@@ -733,12 +733,6 @@ Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a
   :=
   match p with idpath => w end.
```
Alizter:
```suggestion
(** ap011 is defined in Overture.v *)
```

theories/Basics/PathGroupoids.v
mikeshulman: Wouldn't it make more sense to keep the definition in PathGroupoids?  There isn't anything that needs it in Overture, is there?

theories/Cubical/PathCube.v
jdchristensen: Make sense.

theories/Basics/PathGroupoids.v
jdchristensen: If it isn't needed in Overture, then I agree with Mike.

theories/Cubical/PathCube.v
mikeshulman: The discussion with Floris has cemented in me a feeling that almost any function we know to always be an equivalence should be just explicitly defined as an `Equiv` in the first place and coerced to a function, rather than defined as a function and hoped to be inferred to be an equivalence.  I'm even tempted to go back and do this to basic operations like `moveR_Vp` and maybe even `transport`.

theories/Basics/PathGroupoids.v
Alizter: OK will do.

PR #1220, opened 2020-02-06
More systematic use of equiv_functor variants with identity equivalences

mikeshulman:
`equiv_functor_sigma_id` and its friends, used appropriately, can eliminate a lot more explicit type family arguments.

### DIFF
--- a/theories/EquivalenceVarieties.v
+++ b/theories/EquivalenceVarieties.v
@@ -64,7 +64,7 @@ Proof.
   (* First we turn homotopies into paths. *)
   refine (contr_equiv' { g : B -> A & f o g = idmap } _).
   - symmetry.
-    refine (equiv_functor_sigma' 1 _); intros g.
+    apply equiv_functor_sigma_id; intros g.
     exact (equiv_path_forall (f o g) idmap).
     (* Now this is just the fiber over [idmap] of postcomposition with [f], and the latter is an equivalence since [f] is. *)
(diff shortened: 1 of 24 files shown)


theories/EquivalenceVarieties.v
```diff
@@ -134,52 +127,34 @@ Local Definition equiv_fcontr_isequiv' `(f : A -> B)
...
-    (fun a =>
-      exists p, transport (fun a' : A => f a' = f a) p (r (f a)) = 1%path)
-    _ _ 1
-    (fun a => equiv_path_sigma (fun a' => f a' = f a)
-      (g (f a);r (f a)) (a;1%path)))^-1).
+  refine (equiv_functor_forall_id _ oE _).
+  1:intros; apply equiv_sigT_ind.
+  refine (equiv_flip _ oE _).
+  refine (equiv_functor_forall_id _ oE _). 
+  1:{ intros.
```
Alizter:
```suggestion
  { intros.
```

jdchristensen:
This is great!

But there seems to be a problem compiling IncoherentIdempotent.v.  More generally, it would be good to check that this didn't increase the compile time of any other files in a substantial way.

mikeshulman:
I added a timing diff to the commit (compared to current master).  Mostly the timing changes are negligible, but apparently `BlakersMassey` got a smidgen faster.

Issue #1136, opened 2019-11-21
Clean up HIT/epi.v and HIT/iso.v


Alizter:
It would be nice to clean these files up and maybe put them somewhere more appropriate. One option might be Epi.v and Iso.v in theories/ or better yet have their own folder, say `Morphisms`.

I am pretty sure IsSurjection -> IsEpi doesn't need univalence, but currently it is required.

Can we also get away with a version of IsEpi which has homotopies rather than paths?

I have created this issue so I don't forget to do this in the future.

PR #1223, opened 2020-02-07
snapply : simple notypeclasses refine version of rapply

Alizter:
**EDIT**: See my comments near the bottom. I have rewritten most of this now.
---

I have written a tactic that is a `simple notypeclasses refine` version of rapply. I considered calling it `snrapply` but that isn't as easy to remember. It seems that a similar effect can be done with `unshelve eapply`. There is also a notation for `simply notypeclasses refine` called `snrefine`.

Hopefully this makes typing some proofs easier, but I am still a bit unsure if we should have this at all. At the moment there are several other tactics that do something similar:

 * refine
 * refine
 * rapply
 * erapply
 * serapply
 * eapply
 * apply
etc.

There seem to be four main things we are trying to do:

 * Specify whether we want typeclass search
 * Specify whether we want existential variables
 * Specify avoiding having to write underscores
 * Specify the unification algorithm (at least I'm not sure if apply and refine use the same one).

On top of all of this, coq seems to guess typeclasses for implicit arguments so any of the apply tactics followed by a function with implicits cleared (@f) will completely change the behaviour.


mikeshulman:
I'm okay with this, I guess, but I think it should be `snrapply`.  And probably `ntc_refine` should be changed to `nrefine`.  Consistency is what makes things easy to remember.

jdchristensen:
I'm confused about the "e" in "serapply".  Shouldn't we simply allow holes in *all* of the *rapply tactics?  Can we inline the Ltac definition into the Tactic Notation to reduce duplication?  If not, maybe the Ltac definitions could have names that make it clear they are internal?

Alizter:
I tried to make a notation to fill in these underscores but I am running into issues with ltac. I will have another go tomorrow, but if I don't make any progress I will just rename the tactics to be more consistent for now.

Alizter:
@jdchristensen I'm in favour for allowing holes in all tactics, but if someone has done something like rely on rapply failing in their tactic it might break things. Currently rapply is defined as it is defined in coq, but there is nothing stopping us from making ours use existential variables all the time. This would leave us with `rapply`, `srapply`, `nrapply` and `snrapply` which makes somewhat more sense.

JasonGross:
Tactic Notations cannot be recursive.  We can get around this using Ltac's `let rec`.  Not clear to me why you'd want to do this rather than having them be `Local Ltac`, though...

Alizter:
@JasonGross I'm not trying to do anything recursive here. I'm trying to do something like:
```coq
Ltac do_with_holes x p :=
  x p ||
  x (p _) ||
  x (p _ _) ||
  x (p _ _ _) ||
  x (p _ _ _ _) ||
  x (p _ _ _ _ _) ||
  x (p _ _ _ _ _ _) ||
  x (p _ _ _ _ _ _ _) ||
  x (p _ _ _ _ _ _ _ _).
```
But trying the following doesn't work:
```coq

Goal forall A B (f : A -> B) (e : IsEquiv f), A <~> B.
intros x y f e.
do_with_holes refine Build_Equiv.
```
And I'm not quite sure why. I tried a similar thing with Tactic Notation but I can't get that to work either.

JasonGross:
There are two issues:
1. You need to write `ltac:(fun x => refine x)` instead of `refine`; all tactics with special grammar (so all tactic notations and any tactic defined in OCaml that takes arguments) must be fully applied
2. You need to write `uconstr:(p _)` rather than `(p _)`, etc, or else Coq will try to interpret it as a constr and fill all the holes before passing it to the tactic.

(You might also need to pass it `uconstr:(Build_Equiv)` or `(@Build_Equiv)` or else make `do_with_holes` a tactic notation that takes `tactic3(x)` (and then you don't need `ltac:()`) and `uconstr(p)`.)

Alizter:
@JasonGross Still can't get this to work:
```coq

Tactic Notation "do_with_holes" tactic3(x) uconstr(p) :=
(*   let p := uconstr:(p) in *)
  x p ||
  x (p _) ||
  x (p _ _) ||
  x (p _ _ _) ||
  x (p _ _ _ _) ||
  x (p _ _ _ _ _).

  Goal forall A B (f : A -> B) (e : IsEquiv f), A <~> B.
  intros x y f e.
  do_with_holes ltac:(fun x => refine x) (@Build_Equiv).
```

Alizter:
OK so I have removed serapply and erapply, their jobs are now done by a smarter rapply and srapply.

The naming scheme is simpler now: `rapply`, `nrapply`, `srapply`, `snrapply`.

jdchristensen:
I like this pattern a lot more!  A few comments:

There are some large .crashcoqide files committed.  Remove?

HIT.epi is added to AbelianGroup.v.  Can this be removed?

STYLE.md should be updated, as it documents the old names.

For the future (not part of this PR):  I wonder how many `srapply`s could be changed to `snrapply`s without a problem, potentially speeding up compilation.

theories/Basics/Overture.v
```diff
@@ -750,98 +750,100 @@ Global Arguments hfiber {A B}%type_scope f%function_scope y.
...
-(** [erapply lem] is like [apply lem] (rather, [rapply lem]), but it allows holes in [lem] *)
-Tactic Notation "erapply" open_constr(term) := rapply term.
-
-(** [erapply' lem] is like [apply lem] (rather, [rapply' lem]), but it allows holes in [lem] *)
-Tactic Notation "erapply'" open_constr(term) := rapply' term.
-
-(** A shorter name for [simple refine], useful for things like [isequiv_adjointify]. *)
+(** Applying a tactic to a term with increasingly many arguments *)
+Tactic Notation "do_with_holes" tactic3(x) uconstr(p) :=
+  x uconstr:(p) ||
```
JasonGross: This one isn't actually necessary; the `uconstr:` is a noop here.

mikeshulman:
Hmm, I need to think about this idea of changing the meaning of 'rapply'.

mikeshulman:
BTW, the line `Require HIT.epi` is not actually new, it was just moved to the top of the file from the middle of a proof (!) futher down.  But `Require` without an `Import` is strange -- the only reason for that should be if there is a potential naming clash involving the required file, in which case we should perhaps consider doing some renaming.  What's going on?

theories/Basics/Overture.v
Alizter: Would it be a problem to keep? I kind of like the aesthetics.

Alizter:
@mikeshulman I don't think anything clever since I wrote it. Since Dan started #1225 I don't think it is necessary anymore.

The whole of epi needs to be cleaned as I made note of in #1136. I might do this later this week.

jdchristensen:
Right, the HIT.epi line will disappear once this is merged with my new proof of the theorem that used it.  Still, it's bad form to change the location of the import, buried in this PR.

theories/Basics/Overture.v
jdchristensen: I think it's better to remove it.

Alizter:
@jdchristensen Yes, I agree. I was being sloppy with that.

jdchristensen:
I guess it would be comforting to know that the build time doesn't change significantly with these changes.

Also, STYLE.md will need to be updated.

Alizter:
@jdchristensen Yes, I will take a look at updating STYLE.md later.

Alizter:
rebased

theories/Basics/Overture.v
mikeshulman: If it were removed, I would be confused reading it thinking that for some reason the 0-argument case was being treated differently.  How about adding a comment to the effect that it is a no-op?

mikeshulman:
Can someone please explain (again) the meaning of `uconstr` and `open_constr`, and how the argument of an `Ltac` is treated (IIRC labels like `uconstr` can only be applied to the arguments of a `Tactic Notation`, but I can never remember how the arguments of an `Ltac` are treated)?

I am particularly confused because currently we have `erapply` defined to take an `open_constr`, but `serapply` defined to take a `uconstr`.  I thought that the `e` in both cases was supposed to indicate the same thing?

mikeshulman:
Also, is it just me, or is `open_constr` [entirely missing](https://coq.inria.fr/distrib/current/refman/user-extensions/syntax-extensions.html#grammar-token-tactic_argument_type) from the Coq refman?

Alizter:
@mikeshulman 
- `unconstr` is a kernel term that doesn't need to be typed.
- `open_constr` is a kernel term that can have holes and is typed`.

Hence `uconstr` can have holes but nobody cares because it isn't being type checked. If it were a `constr` then it would have to be typed and contain no holes.

This is what I gathered from playing around with it this week.

theories/Basics/Overture.v
jdchristensen: Oh, I misunderstood and thought it was unnecessary on all of the internal lines.  I think it's fine to leave it on that one line given that it's needed on the others as well.

Alizter:
I've had a stab at STYLE.md hopefully I haven't missed anything.

jdchristensen:
I read a few mailing list posts to get the general idea, but sure wish there was clear documentation.  E.g.
https://sympa.inria.fr/sympa/arc/coq-club/2016-07/msg00101.html

jdchristensen:
How about:

```
- `nrefine`, `srefine`, `snrefine`:
  Defined in `Basics/Overture`, these are shorthands for
  `notypeclasses refine`, `simple refine`, and `simple notypeclasses refine`.

- `rapply`, `nrapply`, `srapply`, `snrapply`:
  Defined in `Basics/Overture`, these tactics use `refine`,
  `nrefine`, `srefine` and `snrefine`, except that additional holes
  are added to the function so they behave like `apply` does.
  The unification algorithm used by `apply` is different and often
  less powerful than the one used by `refine`, though it is
  occasionally better at pattern matching.
 
  Here are some tips:
  - If `apply` fails with a unification error you think it shouldn't
    have, try `rapply`.
  - If `rapply` loops on typeclass resolution, try `rapply'` or
    `nrapply'`. The former starts with as many arguments as possible
    and tries decreasing the number. The latter will stop Coq from doing
    a typeclass search.  Similarly, if `refine` loops, try `nrefine`.
  - If you don't want Coq to create evars for certain subgoals,
    add an `s` to the tactic name to make it use `simple refine`.
```

Alizter:
Thanks @jdchristensen

JasonGross:
> Also, is it just me, or is `open_constr` [entirely missing](https://coq.inria.fr/distrib/current/refman/user-extensions/syntax-extensions.html#grammar-token-tactic_argument_type) from the Coq refman?

Indeed, I've reported it just now at https://github.com/coq/coq/issues/11562

JasonGross:
> and how the argument of an `Ltac` is treated

Hah.  I believe there is a special interpretation scope which is some mix between `constr:(...)` and `ltac:(...)`.  In particular, I believe single identifiers are treated as if wrapped in `ltac:(...)`, but complicated expressions are treated as if wrapped in `constr:(...)`.

Note also another relevant fact: `open_constr` does not do typeclass resolution on holes, if I recall correctly.

jdchristensen:
I checked the build time, and the total time only went up by 1.5s (0.55%), which is in the noise.  No file changed by more than 0.2s.

mikeshulman:
So if I wanted to define my own tactic which is just a wrapper around `refine` and behaves exactly like it, would the way to do that be
```
Tactic Notation "myrefine" uconstr:(t) := refine t.
```
?  If so, then I think I'm okay with this naming scheme.

Is there any relationship between the meanings of `simple` in `simple refine` and `simple apply`?

Alizter:
@mikeshulman I don't think `simple refine` and `simple apply` are related.

mikeshulman:
I figured that was probably the case, but if so it's another case of a very unfortunate naming scheme.

jdchristensen:
Yes, your myrefine should behave the same as refine.  (But you have to drop the : after uconstr.)  To be sure, I just tested it in ReflectiveSubuniverses.v, changing all plain refines to myrefine and it worked.

Alizter:
@mikeshulman Are you still reviewing or can I merge?

mikeshulman:
Yeah, I guess we can merge.  But this is probably going to require some serious rebasing of other branches like wildcat.

Alizter:
@mikeshulman It should be pretty easy to change things. If you find and replace `serapply` to `srapply`, `erapply` to `rapply` and `ntc_rapply` to `nrapply` everything should work. If you need any help doing this for wild cat let me know.

Also if you have any reservations about this change do let me know. I am happy to wait a bit more until you make your mind up.

mikeshulman:
No, I have no reservations, now that I realized that the old `rapply` was not actually treating its argument in the same way that `refine` does.

PR #1224, opened 2020-02-08
Redefine IsTrunc as an instance of AccRSU in Overture

mikeshulman:
This is an experiment in the direction of #1209.  At first I tried @jdchristensen's suggestion of defining an `AccRSU` to be a family of formally suspended functions, but I couldn't find a way of doing it that didn't require a lot more changes to the library.  So I went back to my idea of making `AccRSU` an inductive type, which has the advantage that `IsTrunc n` can be a notation for `In (Tr n)` and yet `IsTrunc n.+1 A` can still be definitionally equal to `forall (x y:A), IsTrunc n (x = y)`, and similarly `IsTrunc (-2)` can be definitionally equal to the usual definition of `Contr` (which for us is actually `Contr_internal`).

I haven't touched the `Modalities` code yet; this is just an experiment in the feasibility of defining `IsTrunc` in `Overture` in terms of `AccRSU`.  It's surprising how much of the library compiles without a problem with this change.  However, there are a few places where weird things happen; I'm hoping maybe someone with better Ltac/typeclass-fu can help.

The main issue seems to be whether `Tr : trunc_index -> AccRSU` should be set to `simpl never` or to `simpl nomatch`.  Setting it to `simpl never` seems in principle like a better option to me, since we don't want `IsTrunc n.+1` to get simplified to `In (SepRSU (Tr n))` (as then instances for `IsTrunc` wouldn't match it any more, plus the reader who knows about truncations but not about RSUs would get confused).  And indeed, without `simpl never` we do get problems like that.  But setting it to `simpl never` produces other behavior that I don't understand.

Take a look at the proof of `trunc_sum` on line 885 of `Types/Sum.v`.  It's a bit confusing because `n` is defined to equal `n'.+2` (I don't know why it was written this way), but you can just do a `subst n` at the beginning.  Now the goal is `IsTrunc n'.+2 (A+B)`, which is definitionally equal to `forall (x y:A+B), IsTrunc n'.+1 (x=y)`, so we'd hope that the first line `intros a b` would change the goal to `IsTrunc n'.+1 (a=b)`.  And indeed this is what happens... but only if `Tr` is set to `simpl nomatch`!  If `Tr` is set to `simpl never`, then `intros x y` instead changes the goal to

```
 In
    (SepRSU
       ((fix Tr (n : trunc_index) : AccRSU :=
           match n with
           | -2 => ContrRSU
           | m.+1 => SepRSU (Tr m)
           end) n')) (a = b)
```

There are two confusing things about this, for me.  The first one is that I didn't expect `simpl never/nomatch` to affect the behavior of `intros`!  However, looking at the Coq reference manual I see that `intro` uses the the tactic `hnf` to reduce the goal to a forall, and while the manual doesn't mention that `simpl never/nomatch` might affect `hnf` (they are only mentioned in the subsequent section that describes `cbn` and `simpl`), it seems reasonable that they might.

The second and more confusing thing is that `simpl never` is causing _more_ simplification to occur than `simpl nomatch`!  A bit of experimentation suggests that this happens with `hnf` and `simpl`, but not `cbn`.  Under `simpl nomatch`, `hnf` reduces this goal correctly to `forall x y : A + B, IsTrunc n'.+1 (x = y)`, while `simpl` and `cbn` both reduce it correctly to `  forall (x y : A + B) (x0 y0 : x = y), IsTrunc n' (x0 = y0)`.  And under `simpl never`, `cbn` correctly leaves it untouched, but `hnf` and `simpl` both produce a nasty-looking `fix/match` term.

Can anyone explain what is happening?  Is this just more weird behavior of the old `simpl` algorithm (which perhaps `hnf` is using a variant of)?  More importantly, is there a way to make this work without the `change` statements?

### DIFF
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -1536,6 +1536,7 @@ Proof.
     + refine (H' a b).
     + apply H.
   - intros H' a b.
+    change (IsHProp (a = b)).
     eapply trunc_equiv.
     + apply (H' a b).
     + apply (@isequiv_inverse _ _ _ (H _ _)).
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -716,7 +716,7 @@ Lemma Book_3_11 `{Univalence} : ~ (forall A, Trunc (-1) A -> A).
   {
     intros A.
     apply Book_3_4_solution_1.
-    apply Trunc_is_trunc.
+    exact _.
   }
 
   (** There are no fixpoints of the fix-point free autoequivalence of 2 (called
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -101,7 +101,8 @@ Defined.
 Global Instance ishset_grouphomomorphism {F : Funext} {G H : Group}
   : IsHSet (GroupHomomorphism G H).
 Proof.
-  intros f g; apply (trunc_equiv' _ equiv_path_grouphomomorphism).
+  intros f g. change (IsHProp (f = g)).
+  apply (trunc_equiv' _ equiv_path_grouphomomorphism).
 Defined.
 
 (** * Some basic properties of group homomorphisms *)
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -595,7 +595,7 @@ Class Contr_internal (A : Type) := Build_Contr {
 
 Arguments center A {_}.
 
-(** *** Accessible reflective subuniverses *)
+(** *** Accessible reflective subuniverses and modalities *)
 
 Inductive AccRSU@{i} :=
 | ContrRSU : AccRSU
@@ -616,6 +616,24 @@ Fixpoint subuniv_accrsu (L : AccRSU) : Subuniverse
           end).
 
 Coercion subuniv_accrsu : AccRSU >-> Subuniverse.
+
+Inductive AccModality@{i} :=
+| ContrModality : AccModality
+| IdentityModality : AccModality
+| MeetModality : forall (I : Type@{i}), (I -> AccModality) -> AccModality
+| NulModality : Type@{i} -> AccModality
+| SepModality : AccModality -> AccModality.
+
+Fixpoint accrsu_modality (M : AccModality) : AccRSU
+  := match M with
+     | ContrModality => ContrRSU
+     | IdentityModality => IdentityRSU
+     | MeetModality J Ms => MeetRSU J (fun i => accrsu_modality (Ms i))
+     | NulModality T => LocRSU T Unit (fun _ => tt)
+     | SepModality M => SepRSU (accrsu_modality M)
+     end.
+
+Coercion accrsu_modality : AccModality >-> AccRSU.
 
 (** ** Contractibility and truncation levels *)
 
@@ -700,15 +718,15 @@ Definition trunc_index_to_int n :=
 
 Numeral Notation trunc_index int_to_trunc_index trunc_index_to_int : trunc_scope (warning after 5000).
 
-Fixpoint Tr (n : trunc_index) : AccRSU
+Fixpoint Tr (n : trunc_index) : AccModality
   := match n with
-     | -2 => ContrRSU
-     | m.+1 => SepRSU (Tr m)
+     | -2 => ContrModality
+     | m.+1 => SepModality (Tr m)
      end.
 
 Arguments Tr : simpl never.
 
-Notation IsTrunc n := (In (subuniv_accrsu (Tr n))).
+Notation IsTrunc n := (In (subuniv_accrsu (accrsu_modality (Tr n)))).
 
 (** We use the priciple that we should always be doing typeclass resolution on truncation of non-equality types.  We try to change the hypotheses and goals so that they never mention something like [IsTrunc n (_ = _)] and instead say [IsTrunc (S n) _].  If you're evil enough that some of your paths [a = b] are n-truncated, but others are not, then you'll have to either reason manually or add some (local) hints with higher priority than the hint below, or generalize your equality type so that it's not a path anymore. *)
 
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -488,16 +488,100 @@ Notation "f ^-1" := (@equiv_inv _ _ f _) : function_scope.
 Definition ap10_equiv {A B : Type} {f g : A <~> B} (h : f = g) : f == g
   := ap10 (ap equiv_fun h).
 
-(** ** Contractibility and truncation levels *)
+(** ** Empty and Unit types *)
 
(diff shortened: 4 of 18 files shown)


mikeshulman:
I should try that, I guess.  I had been hoping to keep `IsTrunc` as a notation, so that we wouldn't have to worry about passing back and forth between the definitionally-equal but syntactically-distinct `IsTrunc n` and `In (Tr n)` the way we do in the current library.

By the way, I didn't fix quite all the problems yet; `Categories/ExponentialLaws/Law1/Law.v` has a problem hidden somewhere its incomprehensible tactics that I didn't see how to fix, and so any other files that depend on it might theoretically have issues too that I didn't see.

jdchristensen:
One thing I don't understand about what you've done:  don't parts of the library rely on facts about truncation that are proved by proving them for all modalities?

mikeshulman:
Yes.  So far, I haven't made any connection between `AccRSU` and the Modalities directory; `AccRSU` is just a funny-looking type that's not used for anything (yet) except to define `IsTrunc`.

jdchristensen:
Ah, and then the new IsTrunc is still fed into the Modalities library, and results from there still apply to it.  Got it.

jdchristensen:
BTW, was the problem with my definition that the order of quantifiers was wrong?  I.e., when unfolding `IsTrunc n+1 A`, you got `forall i : Unit, forall x, y : X, ...` instead of `forall x, y : X, forall i : Unit, ...`?

mikeshulman:
Yes, I think that's roughly why `IsTrunc n.+1` didn't turn out equal to `forall x y, IsTrunc n (x=y)`.  That could perhaps be fixed by formally suspending the entire family all at once instead of one by one, but the problem of identifying `Trunc (-2)` with `Contr` would remain.

JasonGross:
The `hnf` tactic uses `simpl` under the hood to do fixpoint refolding.  You might be able to make it use `cbn` with `Set SimplIsCbn`?  (And if that doesn't work, I think you can report it as a bug in Coq, that this setting should also affect `hnf`.). This is a hard part of Coq to deal with, though; there is just enough flexibility to make you think you can do the things you want to do, without enough flexibility to actually achieve what you might want, often.

JasonGross:
> By the way, I didn't fix quite all the problems yet; Categories/ExponentialLaws/Law1/Law.v has a problem hidden somewhere its incomprehensible tactics that I didn't see how to fix, and so any other files that depend on it might theoretically have issues too that I didn't see.

If you `Set Ltac Debug. Set Ltac Batch Debug.`, create files with the resulting log from before and after this commit, and post the diff as a gist somewhere (ideally make a gist with one file which is the log before, and then commit to the gist changing the file's contents to be the log after), I'll take a look and see if I can give you insight.  (I'll try to take a look in the nearish future anyway.). You could also see if the tactics are using `simpl` anywhere and see if replacing them with `cbn` fixes things.  Otherwise the general way to debug `repeat match goal` tactics is to have two copies of the file open in different emacsen, one before and one after, and try to find the first time where `match goal` gives a different result (generally I do this by binary search on `n` in `do n match goal`).  Then try to figure out which branch is triggered in each goal (by doing binary search on how much of the trailing branches you can remove before it fails).  At this point you generally either have a pattern that triggers in one version but not the other, or a tactic that succeeds in one version but not the other, and you can debug that.

Alizter:
So it's the `exp_laws_t` tactic that is causing trouble. Here is the goal having executed `exp_laws_t` just after the `exists` on line 28 in theories/Categories/ExponentialLaws/Law1/Law.v, before Mike's commit:
```coq
H : Funext
zero : PreCategory
H0 : IsInitialCategory 0
one : PreCategory
Contr0 : Contr 1%category
H1 : forall s d : 1%category, Contr (morphism 1 s d)
H2 : IsTerminalCategory 1
C : PreCategory
c : Functor 1 C
1%morphism = (c _1 (center (morphism 1 (center 1%category) (center 1%category))))%morphism
```

here is the same goal after Mike's commit:

```coq
1 subgoal
H : Funext
zero : PreCategory
H0 : IsInitialCategory 0
one : PreCategory
Contr0 : Contr_internal 1%category
H1 : forall s d : 1%category, Contr_internal (morphism 1 s d)
H2 : IsTerminalCategory 1
C : PreCategory
c : Functor 1 C
x, x0 : 1%category
x1 : morphism 1 x x0
transport (fun y : C => morphism C (c _0 x)%object y) (ap c (contr x0))
  (transport (fun y : C => morphism C y (c _0 (center 1%category))%object)
     (ap c (contr x)) 1%morphism) = (c _1 x1)%morphism
```

Alizter:
@mikeshulman I've sent you a PR fixing this.

mikeshulman:
Thanks for the fix @Alizter.  It looks like `Set SimplIsCbn` doesn't affect `hnf` here, but I haven't yet found a good MWE; it seems to depend somehow on the intervening `In`.

On the other hand, if `hnf` _actually_ behaved like `cbn` then that wouldn't be what we want here either, since under `simpl never` it wouldn't even unfold enough to expose a `forall` so that the `intros` would fail.  I would naively expect that a tactic called `hnf` oughn't to need to pay attention to settings like `simpl never` or `simpl nomatch` -- isn't the HNF of a term uniquely well-defined theoretically, not requiring any judgment calls?

JasonGross:
Suppose you have
```coq
Fixpoint add (n m : nat) {struct n} : nat :=
  match n with
  | O => m
  | S n' => S (add n' m)
  end.
```
What is the hnf of `(S x) + y`.  Is it `S (x + y)`?  Or is it `S ((fix add (n m : nat) {struct n} : nat := match n with O => m | S n' => S (add n' m) end) x y)`?  This is the part of `hnf` that `simpl` (or `cbn`) are used for: determining when and how to refold fixpoints back into names.  This might seem like a simple problem, but it becomes significantly more complicated (and time-consuming) when you consider mutual fixpoints, nested fixpoints, fixpoints whose bodies are not fully reduced (e.g., have let binders or beta redexes), and constants which have arguments before `fix`, such as `List.map`, which is defined as
```coq
fun (A B : Type) (f : A -> B) =>
fix map (l : list A) : list B :=
  match l with
  | nil => nil
  | a :: t => f a :: map t
  end
```

mikeshulman:
Ah, I guess I still haven't fully internalized how much work it is to move around between definitionally-but-not-syntactically-equal things.

However, the fact that `hnf` has to evaluate at least far enough to expose a `forall`, whereas plain `cbn` doesn't evaluate at all, does mean it's hard to be sure exactly what it means to say that `SimplIsCbn` "applies to `hnf`" here.  Right?

JasonGross:
I think the actual problem is that cbn, simpl, and hnf are thoroughly underspecified, and [even the Coq dev's have trouble understanding them](https://github.com/coq/coq/issues/4555#issuecomment-337539974).  See https://github.com/coq/coq/issues/11619, https://github.com/coq/coq/issues/7389, https://github.com/coq/coq/issues/7674, https://github.com/coq/coq/issues/5167, https://github.com/coq/coq/issues/4555, https://github.com/coq/coq/issues/3989, https://github.com/coq/coq/issues/3988#issuecomment-337530357, https://github.com/coq/coq/issues/3987#issuecomment-337530348, https://github.com/coq/coq/issues/3488

JasonGross:
> whereas plain `cbn` doesn't evaluate at all, does mean it's hard to be sure exactly what it means to say that `SimplIsCbn` "applies to `hnf`" here. Right?

Here's one way of looking at `simpl` and `cbn`: They consist of two (possibly interwoven) algorithms: deciding what and when to unfold things, and deciding when and how to refold things.  One might think that the former part does not apply to `hnf`, but https://github.com/coq/coq/issues/11619 shows this is not so.  The latter part definitely does apply to `hnf`.  A priori, `simpl` and `cbn` might make different choices about how to do both algorithms, so at the very least we would say that `SimplIsCbn` applies to `hnf` if it changes the refolding algorithm from matching that of `simpl` to matching that of `cbn`.

mikeshulman:
Gah.

Issue #1221, opened 2020-02-07
Travis is failing


mikeshulman:
It seems to be sporadically (?) unable to find `libgraphviz4`?

JasonGross:
Travis is currently having issues with connecting to launchpad during apt-get update.  (My guess is that the package hoating sites rate-limit or see Travis as a DOS attack or something.)  Maybe we should use something like https://github.com/mit-plv/fiat-crypto/blob/master/etc/ci/sudo-apt-get-update.sh to make it clear when this is the issue with Travis.  This has happened maybe half a dozen times before, and they usually manage to fix the problem after a couple of weeks or months.  At this point I'm pretty cynical about reporting this issue to Travis, since the past *n* times haven't resulted in lasting fixes.  (Feel free to find or link to the relevant bug report on Travis though.)

Alizter:
It might be a good time to bring up that Travis isn't the only CI we have to use. Recently we have Github actions which seems to satisfy all our needs. I wouldn't mind working out how to set it up.
See: [https://knapsackpro.com/ci_comparisons/github-actions/vs/travis-ci](https://knapsackpro.com/ci_comparisons/github-actions/vs/travis-ci f) for a comparison with travis.

JasonGross:
Switching to GitHub Actions seems like a good idea, as long as we can get it to run on PRs.  I don't have time to look into it much right now, but if you figure out how to make it work, I'm happy to review and approve

PR #1227, opened 2020-02-10
Github actions

Alizter:
Here is a start on setting up the Github actions CI.

I couldn't workout how to get the build matrix running. There are some issues with Jason's ppa's sinece actions only has xenial but those are for trusty I believe.

I've created a draft pull request so we can work on it.


spitters:
Is this intended to replace Travis?

Alizter:
Yes see #1221.

JasonGross:
The ppa packages should exist on every stale version since precise.  (It's possible the newer ones start from Trusty rather than precise.)

Alizter:
If we merge #1228 we can see the logs for this.

JasonGross:
Oh, is it graphviz that's giving issues?  I think that on newer versions of Ubuntu, the default graphviz might be new enough?
On checking, it looks like we probably just need >= 2.38 to avoid a bug.  We've been using 2.39, disco has 2.40, and xenial has 2.38.  I can backport graphviz 2.39 to xenial if we need it.  But try it first without the ppa and see if it works.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+    # a
+    - name: Add repositories to apt
+      ## TODO: Cache dependencies?
+      run: |
+        # echo Adding repositories...
+        # sudo add-apt-repository avsm
+        sudo add-apt-repository 'ppa:jgross-h/graphviz'
+        sudo add-apt-repository 'ppa:jgross-h/coq-master-daily'
+    - name: Update apt
+      continue-on-error: true
```
JasonGross: No, we should not continue on error.  If apt-get update fails, we should re-run it.  Even if it exits with status 0, it sometimes it should sometimes be considered to have failed.  I'm not sure if we should bother with checking it now, or wait until it becomes a problem, but we definitely should not continue-on-error.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+name: Github Actions CI
+
+
+## These are the actions that trigger the CI to run
+on:
+  ## We want the CI to run on every push
+  push:
+    branches: 
+      - master
+      - actions
```
JasonGross: Why do we have a branch `actions` here?  Why not just let it run on all branches on push and pull_request?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+    # Here is the main script
+    steps:
+    # First we set up Python 2.7
+    - name: Set up Python 2.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 2.7
+    # Now we checkout our branch  
+    - uses: actions/checkout@v2
+    # a
```
JasonGross: This comment seems wrong.  Also we need to init submodules for the timing scripts.  I updated one of my other projects to GitHub Actions after you pointed out that it exists, so you can do submodule init as in https://github.com/mit-plv/fiat-crypto/blob/f2aa648c7948d4664a4af440524cf41cc18278ba/.github/workflows/coq.yml#L42

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+      run: |
+        # echo Adding repositories...
+        # sudo add-apt-repository avsm
+        sudo add-apt-repository 'ppa:jgross-h/graphviz'
+        sudo add-apt-repository 'ppa:jgross-h/coq-master-daily'
+    - name: Update apt
+      continue-on-error: true
+      run: sudo apt-get update
+    - name: Installing dependencies
+      continue-on-error: true
```
JasonGross: Seems sketchy

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+      run: ./autogen.sh
+    
+    - name: Run ./configure
+      run: ./configure
+    
+    - name: Timing configuration
+      run: ./etc/coq-scripts/timing/make-pretty-timed.sh V=1 -j2
+    
+    - name: Make
+      run: make
```
JasonGross: Note that the pretty-timed file runs `make`; we do this afterwards just to get the failures at the bottom of the log, and to propagate failure to the pipeline.  Maybe just add this to the step above?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+        ./etc/ci/before_script.sh
+        export COMMITISH="$(git rev-list HEAD -1)"
+
+    - name: Run ./autogen.sh
+      run: ./autogen.sh
+    
+    - name: Run ./configure
+      run: ./configure
+    
+    - name: Timing configuration
```
JasonGross: This doesn't configure timing.  This builds everything with timing on and then prints a timing table. (But I think it succeeded even if the build fails, so we need to run make again afterwards.)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+        # sudo add-apt-repository avsm
+        sudo add-apt-repository 'ppa:jgross-h/graphviz'
+        sudo add-apt-repository 'ppa:jgross-h/coq-master-daily'
+    - name: Update apt
+      continue-on-error: true
+      run: sudo apt-get update
+    - name: Installing dependencies
+      continue-on-error: true
+      run: |
+        sudo apt-get install aspcud
```
JasonGross: I think you want `-y`, maybe also `--allow-unauthenticated`, on all of these

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+        python-version: 2.7
+    # Now we checkout our branch  
+    - uses: actions/checkout@v2
+    # a
+    - name: Add repositories to apt
+      ## TODO: Cache dependencies?
+      run: |
+        # echo Adding repositories...
+        # sudo add-apt-repository avsm
+        sudo add-apt-repository 'ppa:jgross-h/graphviz'
```
JasonGross: Do you need `-y` here?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+        sudo apt-get install libocamlgraph-ocaml-dev
+        sudo apt-get install lua5.1
+    
+    - name: Upgrade and autoremove packages
+      run: |
+        # This will take a very long time
+        # sudo apt-get -y upgrade
+        sudo apt-get autoremove
+ 
+   # This should be conditional on the matrix
```
JasonGross: In what way?  If you want an example of how to do this, see https://github.com/mit-plv/fiat-crypto/blob/4d1217bdf3456db0a99ab7ddaecd3db07a94ceb7/.github/workflows/coq.yml#L72

.github/workflows/ci.yml
```diff
@@ -0,0 +1,108 @@
...
+    - name: Timing configuration
+      run: ./etc/coq-scripts/timing/make-pretty-timed.sh V=1 -j2
+    
+    - name: Make
+      run: make
+    
+    - name: Test install target
+      run: ./etc/ci/test-install-target.sh
+    
+    # TODO: UPDATE_QUICK_DOC
```
JasonGross: What's the TODO?

.github/workflows/ci.yml
Alizter: Yes, update was failing because of the ppa. I stuck this in just to get it to finish.

.github/workflows/ci.yml
Alizter: That is left over from my branch when I was testing it.

.github/workflows/ci.yml
Alizter: I saw something similar in the travis.yml, I put this comment in here in case anything needed to be done.

.github/workflows/ci.yml
Alizter: I need to set up environment variables like we have in travis.

.github/workflows/ci.yml
Alizter: Is there a better way to list these other than copy and paste?

.github/workflows/ci.yml
Alizter: Yes, it wasn't too obvious from the travis script. I will put make back into the other step.

.github/workflows/ci.yml
Alizter: It is. I don't really want any of these continue-on-error to be present.

.github/workflows/ci.yml
Alizter: Sounds good I will add that.

.github/workflows/ci.yml
JasonGross: You can put them all into the same `sudo apt-get install` command.  If you need it to depend on the build matrix, you can list the packages in an environment variable like `PACKAGES` and then run `sudo apt-get install -y --allow-unauthenticated $PACKAGES`

Alizter: I had them like that before, but it got annoying when one of them failed. I guess the trick is to make none of them fail. :-)

The dependency on the matrix is a useful feature.

Alizter:
@JasonGross I need some help with graphviz here.

JasonGross:
I've pushed a couple of commits, and GH actions seems to be basically working now.  (Though proviola is still running.)  I still have to update things so that the relevant build jobs will actually push updates to gh-pages, since the deployment infrastructure is different on GH Pages than on Travis.  Currently working on that.

Alizter:
I guess for now, this is fine. I would eventually like to add more operating systems since we do have users on Windows and possibly Mac. For now this looks good to me.

I think for a time we should have both travis and actions just to make sure they agree on most things.

@JasonGross do you know why travis is unhappy here?

JasonGross:
The issue on Travis:
```
$ sudo npm install -g doctoc
sudo: npm: command not found
```
Maybe we need another package?  Idk what changed though.......

Alizter:
@JasonGross Looks like travis was working before `Update install_doctoc.sh`.

JasonGross:
Seems kind-of insane if sudo npm fails to find npm, but npm works fine.  Can you add an npm package to .travis.yml?

Alizter:
Whoops sorry you are correct. I did this from my phone and just realised how silly that was. :-)

Alizter:
@JasonGross We are getting a very colourful error now.

Alizter:
What on earth does this mean:

```
npm http GET https://registry.npmjs.org/doctoc
npm http GET https://registry.npmjs.org/doctoc
npm http GET https://registry.npmjs.org/doctoc
npm ERR! Error: CERT_UNTRUSTED
npm ERR!     at SecurePair.<anonymous> (tls.js:1370:32)
npm ERR!     at SecurePair.EventEmitter.emit (events.js:92:17)
npm ERR!     at SecurePair.maybeInitFinished (tls.js:982:10)
npm ERR!     at CleartextStream.read [as _read] (tls.js:469:13)
npm ERR!     at CleartextStream.Readable.read (_stream_readable.js:320:10)
npm ERR!     at EncryptedStream.write [as _write] (tls.js:366:25)
npm ERR!     at doWrite (_stream_writable.js:223:10)
npm ERR!     at writeOrBuffer (_stream_writable.js:213:5)
npm ERR!     at EncryptedStream.Writable.write (_stream_writable.js:180:11)
npm ERR!     at write (_stream_readable.js:583:24)
npm ERR! If you need help, you may report this log at:
npm ERR!     <http://github.com/isaacs/npm/issues>
npm ERR! or email it to:
npm ERR!     <npm-@googlegroups.com>
npm ERR! [0mSystem Linux 4.4.0-101-generic
npm ERR! command "/usr/bin/nodejs" "/usr/bin/npm" "install" "-g" "doctoc"
npm ERR! cwd /home/travis/build/HoTT/HoTT
npm ERR! node -v v0.10.25
npm ERR! npm -v 1.3.10
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /home/travis/build/HoTT/HoTT/npm-debug.log
npm ERR! not ok code 0
$ exit 1
$ exit 1
The command "./etc/ci/before_script.sh" failed and exited with 1 during .
Your build has been stopped.
```

Alizter:
Also, update html is taking 28 minutes...

Alizter:
@JasonGross So I have added `npm config set strict-ssl false` just before npm installs doctoc. This is probably not ideal.

Alizter:
I don't know what else to try this is really bugging me.

Alizter:
So it seems the sudo version of npm is quite different:
```
$ npm --version
6.13.7
$ sudo npm --version
sudo: npm: command not found
$ sudo npm install -g doctoc
sudo: npm: command not found
$ exit 1
$ exit 1
```

artagnon:
You can run `sudo -E` to inherit the environment, I think.

JasonGross:
We could also remove Travis entirely and switch to just GitHub Actions.

Other things we might want to consider:
- making the generated documentation available as an artifact on the PR even before it's merged
- using some marketplace action to push the updates to documentation, rather than factoring it through scripts (though scripts have the benefit that if we move to a different CI, it'll be easier to port)

spitters:
At some point we could consider replacing proviola with jscoq.

Alizter:
OK I have removed travis.

mikeshulman:
I don't have the time to learn all of this syntax, but it seems to work, so LGTM I guess.

mikeshulman:
What do all the different checks mean?  Which one(s) are actually checking whether the PR compiles (and if there is more than one, why?) and which are doing what other things?

Alizter:
@mikeshulman These are all the jobs that travis was doing before. The one we really care about is the top one, labelled "build".

mikeshulman:
I've always wanted to know what all the jobs are doing, though.  Can we have a list somewhere?

Alizter:
@mikeshulman Sure, I will write one up.

JasonGross:
Three of the jobs are about checking the build with master (installed from my ppa), checking that we build with the pinned version of Coq, and checking that we build when we fast forward the submodule to master (this is redundant work the ppa job; maybe we should drop it).  The other jobs are about building and pushing the various bits of documentation (dependency graph, readme TOC, coqdoc, proviola, etc)

Alizter:
I was thinking of adding something to STYLE.md about the ci.

jdchristensen:
Every time I push to my fork of the HoTT repo, I get notified of a build failure during the "apt repositories to apt" step:
```
W: Conflicting distribution: https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease (expected xenial but got stable)
67
E: Repository 'https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease' changed its 'Origin' value from 'microsoft-ubuntu-xenial-prod xenial' to 'vscode stable'
68
E: Repository 'https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease' changed its 'Label' value from 'microsoft-ubuntu-xenial-prod xenial' to 'vscode stable'
69
E: Repository 'https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease' changed its 'Suite' value from 'xenial' to 'stable'
70
E: Repository 'https://packages.microsoft.com/ubuntu/16.04/prod xenial InRelease' changed its 'Codename' value from 'xenial' to 'stable'
71
##[error]The operation was canceled.
```
Any idea why I'm getting this?

Alizter:
@jdchristensen This is very strange. I tried your branch also and I got past the "add repositories to apt" stage. You should be able to restart your jobs when they have failed. When in the actions tab, there is a "Re-run jobs" button.

If this still persists let us know.

jdchristensen:
The error quoted above went away at random, and then there were other errors that were in fact correct and which I fixed, so now my branch builds.

Alizter:
@jdchristensen I suspect it was a one off. Recently github have been messing around with some behind the scenes stuff which may be the cause. If it is happening more often then let us know however.

PR #1228, opened 2020-02-10
Create ci.yml

Alizter:
Here is a blank Github actions file so that we can review my PR #1227 and see that it works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,6 @@
 name: CI
 
-on: [push]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
+name: CI
+
+on: [push]
```
JasonGross:
```suggestion
on: [push, pull_request]
```
Do we want this so that we can see how it behaves on your PR?

PR #1230, opened 2020-02-11
Reorganised finite and added some functions

### LINKED ISSUE #1034: How to define `nat -> Fin n.+1`
Hi, there is a map `nat -> Fin n.+1` that I need which is typically defined using modular arithmetic. How can I write this down for `Fin n.+1`?

We have a modulo function in `Coq.Init.Nat` but this is `nat -> nat -> nat`. Perhaps we need to show that this has a finite image, hence there is my desired map. But this isn't obvious to me either.

Another way I can think of is mapping `N -> Z -> Z/nZ` but the latter will require use of quotients and showing that these are finite.

What is the best way to go about defining this map?

Alizter:
So I have reorganised `Finite.v` into a folder in `Spaces` called `Finite`. It is now currently split as follows:
 * `Finite/Core.v`: Definition of `Fin`, some properties of `Fin` and some operations on `Fin`.
* `Finite/Equiv.v`: Equivalences between finite sets.
* `Finite/Finite.v`: Definition of `Finite` and results on finite types.
* `Finite/Quotient.v`: Quotient of finite sets.

Perhaps the splitting is not ideal. Let's discuss this and share ideas.

In Core I did manage to implement a modulo map from nat to Fin k. So this closes #1034. This function is called fin_nat, and I would expect that `fin_nat (nat_fin x) = x` though I couldn't prove it.


mikeshulman:
Thanks for taking this on.  I think splitting out `Core` and `Equiv` is reasonable, though I would name them differently: perhaps `Canonical` and `CanonicalEquiv`?

What was your reasoning for splitting out the fairly small file `Quotient` and not any of the other operations on finite sets in `Finite`?

Alizter:
My original motivation for separating Quotient was because it imports quite a lot of other things that most of Finite doesn't use. I'm happy to put them back in, perhaps by requiring them at the top and importing them later on.

Would `Fin` and `FinEquiv` be a good name? That seems to be what they are about. Canonical is also a tricky word.

Alizter:
Rebased.

Alizter:
There is still some more work to be done here. I have added the label so nobody accidentally merges it.

mikeshulman:
I guess you can't change it into a "Draft pull request" after it's made?

I could live with `Fin` and `FinEquiv`.  What's the reason for separating out parts that import different stuff?

Alizter:
@mikeshulman For some reason I did this last week, but I can't remember why. I thought it was to do with typeclass search being slower, but now looking at it it doesn't seem to be the case. I will put `Quotient.v` back into `Finite.v`.

Alizter:
Is there any point to having CoreflectiveSubuniverse.v in its own folder in Comodalities anymore? If it doesn't look likely that we will add any comodalities, should we not just put them in the modalities folder? It's not like everything in the modalities folder has to do with modalities anyway.

mikeshulman:
Looks like this needs rebasing.

Alizter:
rebased

mikeshulman:
Whenever I say "needs rebasing" I mean "needs rebasing and updating to compile with current master".  (-:

Alizter:
Yes, I am sorry for having an atrocious record of doing this :-)

I will fix this tomorrow.

theories/Spaces/Finite/Finite.v
```diff
@@ -0,0 +1,729 @@
...
+
+Definition detachable_finite_subset {X : Type} `{Finite X}
+  (P : X -> Type) `{forall x, IsHProp (P x)}
+  {Pf : Finite (sig P)}
+  : forall x, Decidable (P x).
+Proof.
+  intros x.
+  refine (decidable_equiv _ (hfiber_fibration x P)^-1 _).
+  refine (detachable_image_finite pr1 x).
+  apply (mapinO_pr1 (-1)%trunc). (* Why can't Coq find this? *)
```
Alizter: This line is still an issue and I cannot work out why.

mikeshulman: Did you try changing `(-1)%trunc` to `(Tr (-1))`?

Alizter: @mikeshulman That's not the issue here. It shouldn't be needed at all, typeclass search should find it on the previous line. Or at least, it should be able to be replaced by `exact _`.

mikeshulman: Oh, by "still an issue" you just meant that the comment `Why can't Coq find this?` still applies?

There are a bunch of places in the new modalities code where I think Coq should be able to find a typeclass instance and it can't.

Alizter: Yes, it seems to be a coq issue we should look into in the future. Do you prefer `(-1)%trunc` or `Tr (-1)` here?

Alizter: Another weird thing I should note is that previously we didn't need to specify the trunc index at all. Now we do.

mikeshulman: I left some others of these as `Tr (-1)`, mainly since I find `%` annotations unaesthetic.  But YMMV, I don't have strong feelings and I could see an argument the other way.

Alizter:
```suggestion
  apply (mapinO_pr1 (-1)%trunc). (* Why can't Coq find this instance? *)
```

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+(** Here is an example *)
+Goal Fin 5 -> Type.
+FinInd.
+Abort.
+
+(** ** Operations on Fin *)
+
+(** Sucessor of an element of a finite set. This is a periodic function. *)
+(** TODO: make this more readable? *)
+Fixpoint fsucc {n : nat} : Fin n -> Fin n
```
Alizter: This is the cyclic successor. Egbert outlined how to write this in agda in #1034. The coq definition is pretty horrible, but computes for the examples we tested. I'm not really sure how to make this easier to understand.

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+                | inl r''' => inl (F (inl r'''))
+                | inr tt => inr tt
+              end
+          end (@fsucc n') r'
+        | inr _ => fin_zero n'
+        end
+      end.
+
+(** The map from the natural numbers to a finite set. The modulo map. *)
+Fixpoint fin_nat (n : nat) (m : nat) : Fin n.+1
```
Alizter: The main use of fsucc is we can turn any nat into an element of a given finite set. This will be useful if we ever wish to define integers modulo n.

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+  := match m with
+      | 0 => fin_zero n
+      | S m => fsucc (fin_nat _ m)
+     end.
+
+(** TODO: Use and put in scope *)
+(** Here is a notation for a natural number in a finite set. It is suggestive of modulo classes. *)
+(* Notation "[ n ]" := (fin_nat _ n). *)
```
Alizter: Here is a proposed notation that I didn't add for now.

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+      specialize (IH k).
+      rewrite nat_plus_comm in IH.
+      exact (ap S IH).
+    + rewrite nat_plus_comm; reflexivity.
+Qed.
+
+(** TODO: finish *)
+(** Turning an element of a finite set into a natural number then back into a finite set is homotopic to the identity map. *)
+(*
+Definition fin_nat_nat_fin (n : nat)
```
Alizter: This was something I expect to be true on paper. I'm basically trying to say that the map `nat_fin` which takes an element of a finite set and turns it into a nat is an injection, i.e. has a left inverse.

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+Defined.
+
+(** If there is a map from an n element finite set into the empty type then n = 0. *)
+Definition fin_empty (n : nat) (f : Fin n -> Empty) : n = 0.
+Proof.
+  destruct n; [ reflexivity | ].
+  elim (f (inr tt)).
+Defined.
+
+(** The zero element in a non-empty finite set is the right most element. *)
```
Alizter:
```suggestion
(** The zero element in a non-empty finite set is the left most element. *)
```

theories/Spaces/Finite/Fin.v
```diff
@@ -0,0 +1,180 @@
...
+Fixpoint fin_zero (n : nat)
+  : Fin n.+1
+  := match n with
+      | O => inr tt
+      | S n' => inl (fin_zero n')
+      end.
+
+(** There is an injection from the n element finite set into the n.+1 element finite set,
+    mapping [i : Fin n] to the successor of [i] in [Fin n.+1]. *)
+Fixpoint fin_succ_inject (n : nat)
```
Alizter: This is a shift map into the bigger fin.

theories/Spaces/Finite/FinEquiv.v
```diff
@@ -0,0 +1,286 @@
...
+  - exact idpath.
+  - elim (e^-1 (inr tt)).
+  - elim (e (inr tt)).
+  - refine (ap S (IHn m _)).
+    exact (snd (fin_equiv_inv n m e)).
+Qed.
+
+(** Definition of the permutation [Fin n -> Fin n] which reverses the
+    order of the elements, [0 ↦ n-1], [1 ↦ n-2], etc. *)
+Fixpoint fin_rev {n : nat} : Fin n -> Fin n :=
```
Alizter: This reverses the elements of a finite set.

theories/Spaces/Finite/FinEquiv.v
```diff
@@ -0,0 +1,286 @@
...
+  : forall i : Fin n, fin_rev (fin_succ_inject n i) = inl (fin_rev i).
+Proof.
+  induction n.
+  - contradiction.
+  - cbn in *; intros [i|u].
+    + by destruct (IHn i)^.
+    + by destruct u.
+Defined.
+
+Lemma is_involution_fin_rev {n : nat}
```
Alizter: The reverse map is an involution

Alizter:
@mikeshulman So I did something really stupid here and split the files and added things at the same time. This is near impossible to review but I have commented on all the new functions that I could remember.

mikeshulman:
Well, sometimes it's hard; #1238 was also a combination of a structural change with some new added content.  But in this case it might be possible to go back and break it into two PRs, if you felt like it.

theories/Spaces/Finite/Fin.v
mikeshulman: You could define it with tactics instead of as a giant term.

Alizter: So what is going on here is that we recursively take the last element in `Fin n.+1` in the source and make it the first in `Fin n.+1` in the target, then  take the remaining `Fin n` from the inr in the source, apply to current function recursively, then shift inject it (`fin_succ_inject`) into `Fin n.+1` at the target.

Alizter: OK this needs to be redone with tactics and explanation.

PR #1232, opened 2020-02-14
Trunc index inc

jdchristensen:
1st commit:

Trunc.v: rewrite trunc_index_inc so `trunc_index_inc minus_two.+2 n` definitionally agrees with `nat_to_trunc_index n`.

Trunc.v: add `trunc_index_inc'` which inserts the successors in a different order.

Loops.v: Replace only use of `nat_to_trunc_index_2` with `trunc_index_inc`.

pTrunc.v: Use `trunc_index_inc'` instead of `+2+` to eliminate rewrites in `ptr_iterated_loops`.

Spectrum.v: use new argument order for `trunc_index_inc`.

2nd commit:

Loops.v: Add `isconnected_iterated_loops_functor`, using `trunc_index_inc'`.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -547,45 +547,12 @@ Notation "n .+4" := (n.+1.+3)%nat   : nat_scope.
 Notation "n .+5" := (n.+1.+4)%trunc : trunc_scope.
 Notation "n .+5" := (n.+1.+4)%nat   : nat_scope.
 Local Open Scope trunc_scope.
-Fixpoint nat_to_trunc_index (n : nat) : trunc_index
-  := match n with
-       | 0%nat => minus_two.+2
-       | S n' => (nat_to_trunc_index n').+1
-     end.
 
-Coercion nat_to_trunc_index : nat >-> trunc_index.
-
-Definition int_to_trunc_index (v : Decimal.int) : option trunc_index
-  := match v with
-     | Decimal.Pos d => Some (nat_to_trunc_index (Nat.of_uint d))
-     | Decimal.Neg d => match Nat.of_uint d with
-                        | 2%nat => Some minus_two
-                        | 1%nat => Some (minus_two.+1)
-                        | 0%nat => Some (minus_two.+2)
-                        | _ => None
-                        end
-     end.
-
-Fixpoint trunc_index_to_little_uint n acc :=
-  match n with
-  | minus_two => acc
-  | minus_two.+1 => acc
-  | minus_two.+2 => acc
-  | trunc_S n => trunc_index_to_little_uint n (Decimal.Little.succ acc)
-  end.
-
-Definition trunc_index_to_int n :=
-  match n with
-  | minus_two => Decimal.Neg (Nat.to_uint 2)
-  | minus_two.+1 => Decimal.Neg (Nat.to_uint 1)
-  | n => Decimal.Pos (Decimal.rev (trunc_index_to_little_uint n Decimal.zero))
-  end.
-
-Numeral Notation trunc_index int_to_trunc_index trunc_index_to_int : trunc_scope (warning after 5000).
+(** Further notation for truncation levels is introducted in Trunc.v. *)
 
 Fixpoint IsTrunc_internal (n : trunc_index) (A : Type) : Type :=
   match n with
-    | -2 => Contr_internal A
+    | minus_two => Contr_internal A
     | n'.+1 => forall (x y : A), IsTrunc_internal n' (x = y)
   end.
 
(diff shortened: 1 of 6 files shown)


mikeshulman:
Fascinating.  I had no idea that the primitive `fix` command took multiple arguments; I would naively have expected that the `Fixpoint` command would be desugared into something like your version, with arguments preceeding the structural one bound in an enclosing lambda and arguments following it incorporated into a function return type.  Do you (or anyone) have any idea why it was done this way instead?

This is a nice simplification, but I'm wary of depending on two things that aren't explicitly defined in terms of each other nevertheless accidentally-on-purpose turning out to be definitionally equal.  Could we just _define_ `nat_to_trunc_index n` to mean `trunc_index_inc minus_two.+2 n` (either by moving `trunc_index_inc` into `Overture` or `nat_to_trunc_index` into `Trunc.v`?

SkySkimmer:
`fix` must take multiple arguments as they may vary in recursive calls. Strictly speaking you could try to reorder arguments given before the structural one to be after it (arguments after the structural argument are just normal lambdas) but this doesn't work when there is a dependency.
eg consider `fix (n m : nat) (H : le n m) {struct H} ...`, subcalls will be with different values for `m` and `n`.

mikeshulman:
Hmm, I definitely see the point when you have an indexed inductive type.  And I suppose once you're allowing `fix` to take multiple arguments in that case, it's simpler to allow it arbitrarily.  Thanks for explaining!

mikeshulman:
I'm also curious whether there is a reason to reverse the order of arguments?  I think your version with an inner fixpoint can also be written with the arguments in the original order:
```
Definition trunc_index_inc (n : nat) (k : trunc_index)
  : trunc_index
  := (fix inner (n : nat) : trunc_index := match n with
      | O => k
      | S m => (inner m).+1
    end) n.
```

jdchristensen:
I don't think we can move nat_to_trunc_index out of Overture.v, since lots depends on it, but it's certainly find to move trunc_index_inc to Overture.v.  I had it that way at one point, but I wasn't sure where to draw the line.  trunc_index_inc' isn't needed there, but having two similar but not the same definitions separated is a bit of an issue.  If trunc_index_inc' goes there too, then what about the lemmas relating them?  Neither form of trunc_index_inc is use more than a couple of times, so it seems best to minimize their appearance in Overture.v.  Let me know what you prefer and I'll do it.

About the argument order, the definition is cleaner the new way, and it somehow matches my intuition better, but again I'm fine with either option.

jdchristensen:
How about moving *only* trunc_index_inc to Overture, with a comment where to find the related things?

mikeshulman:
The old order of arguments makes marginally more sense to me with the partially-applied `trunc_index_inc n` being "the function that increments truncation levels by n", but I'm not going to make you change it back.  (-:

I think the only things that would have to move out of `Overture` along with `nat_to_trunc_index` are the handful of definitions following it up until the `Numeral Notation` line.  Then there are just a couple of uses of "-2" later in `Overture` that would have to change to `minus_two`.  I kind of like that, actually -- there's too much in `Overture` already.  We could put these lines, along with both `trunc_index_inc`s, at the top of `Trunc.v`.

jdchristensen:
This form also has the correct definitional behaviour (with either argument order).  Not sure which one is easier to read.
```
Definition trunc_index_inc'' (n : nat) (k : trunc_index) : trunc_index.
Proof.
  induction n.
  - exact k.
  - exact IHn.+1.
Defined.
```
I should also mention that I tried for a long time to find a way to have only one form of trunc_index_inc, but there are spots that just turn out to be awkward with the wrong form.  (Including some work in progress.)  I even tried redefining nat_to_trunc_index using trunc_index_nat', but still there were wrinkles that couldn't be fixed.  OTOH, having both seems to be working well for my applications.

mikeshulman:
I lean a bit towards the fixpoint as easier to read -- the boilerplate syntax of `fix` and `match` is a bit heavy, but once you get past that, I think the explicit branches of the match make it more clear what's going on.

jdchristensen:
You were right, it wasn't hard to move things to Trunc.v, and I think that's an improvement.

mikeshulman:
Great!

Now that `nat_to_trunc_index` is defined in terms of `trunc_index_inc`, is there a reason to still define the latter with an inner fixpoint?  They will now be definitionally equal however we define the latter, right?

jdchristensen:
The way with the inner fix beta-reduces for all of the nat coercions, but that probably doesn't matter, and it's certainly easier to compare the two different trunc_index_inc functions when they are in the same style.  So I switched it.

Alizter:
Just going to restart travis before I merge.

Alizter:
@JasonGross why is travis complaining here?

jdchristensen:
All but one of the build jobs succeeded.  The last one failed in the timing portion.  Can we merge this anyways?

```
  File "etc/coq-scripts/timing/make-one-time-file.py", line 19, in <module>
    times_dict = get_times(sys.argv[1])
  File "/home/travis/build/HoTT/HoTT/etc/coq-scripts/timing/TimeFileMaker.py", line 38, in get_times
    return dict((name, reformat_time_string(time)) for name, time in times)
  File "/home/travis/build/HoTT/HoTT/etc/coq-scripts/timing/TimeFileMaker.py", line 38, in <genexpr>
    return dict((name, reformat_time_string(time)) for name, time in times)
  File "/home/travis/build/HoTT/HoTT/etc/coq-scripts/timing/TimeFileMaker.py", line 14, in reformat_time_string
    seconds, milliseconds = time.split('.')
ValueError: too many values to unpack
The command "./autogen.sh && ./configure && if [ -z "$UPDATE_QUICK_DOC" ]; then make strict-test && ./etc/coq-scripts/timing/make-pretty-timed.sh V=1 -j2 && make && ./etc/ci/test-install-target.sh && cat ./time-of-build-pretty.log && echo; fi && ./etc/ci/after_success.sh" exited with 1.
```

Alizter:
I've restarted that job. I will merge it when it is done again. It's still strange that travis failed on that job. It would be good to know what is going on. Travis has been causing some problems recently which is why we are working on moving to Github actions.

PR #1236, opened 2020-02-15
Add Lemma 1.23 from RSS to ReflectiveSubuniverses.v

jdchristensen:
Adds a converse to `O_inverts_by_yoneda`, and a converse to the converse (in a different sense).

### DIFF
--- a/theories/Extensions.v
+++ b/theories/Extensions.v
@@ -180,6 +180,20 @@ Section Extensions.
     - apply equiv_pathsplit_isequiv.
   Defined.
 
+  (* Without [Funext], we can prove a small part of the above equivalence.
+     We suspect that the rest requires [Funext]. *)
+  Definition extension_isequiv_precompose
+           {A : Type} {B : Type}
+           (f : A -> B) (C : B -> Type)
+    : IsEquiv (fun (g : forall b, C b) => g oD f) -> forall g, ExtensionAlong f C g.
+  Proof.
+    intros E g.
+    pose (e := Build_Equiv _ _ _ E).
+    exists (e^-1 g).
+    apply apD10.
+    exact (eisretr e g).
+  Defined.
+
   (** Postcomposition with a known equivalence.  Note that this does not require funext to define, although showing that it is an equivalence would require funext. *)
   Definition extendable_postcompose' (n : nat)
              {A B : Type} (C D : B -> Type) (f : A -> B)
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -629,6 +629,31 @@ Section Reflective_Subuniverse.
       rapply e.
     Defined.
 
+    (** The converse is also true.  This the first half of Lemma 1.23 of RSS. *)
+    Definition ooextendable_O_inverts
+               {A B : Type} (f : A -> B) `{O_inverts f}
+               (Z : Type) `{In O Z}
+      : ooExtendableAlong f (fun _ => Z).
+    Proof.
+      refine (cancelL_ooextendable _ _ (to O B) _ _).
+      1:srapply extendable_to_O.
+      refine (ooextendable_homotopic _ (O_functor f o to O A) _ _).
+      1:apply to_O_natural.
+      refine (ooextendable_compose _ (to O A) (O_functor f) _ _).
+      2:srapply extendable_to_O.
+      srapply ooextendable_equiv.
+    Defined.
+
+    (** And now the funext version *)
+    Definition isequiv_precompose_O_inverts `{Funext}
+               {A B : Type} (f : A -> B) `{O_inverts f}
+               (Z : Type) `{In O Z}
+      : IsEquiv (fun g:B->Z => g o f).
+    Proof.
+      srapply (equiv_extendable_isequiv 0).
+      exact (ooextendable_O_inverts f Z 2).
+    Defined.
+
     Definition to_O_inv_natural {A B : Type} `{In O A} `{In O B}
                (f : A -> B)
     : (to O B)^-1 o (O_functor f) == f o (to O A)^-1.
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -654,6 +654,28 @@ Section Reflective_Subuniverse.
       exact (ooextendable_O_inverts f Z 2).
     Defined.
 
+    (** This property also characterizes the types in the subuniverse, which is the other half of Lemma 1.23. *)
+    Definition inO_ooextendable_O_inverts (Z:Type@{k})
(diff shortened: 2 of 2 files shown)


mikeshulman:
I have a version of the first of these in my in-progress ditch-the-modules branch, which goes by way of a non-funext version:

```
    Definition ooextendable_O_inverts
               {A B : Type} (f : A -> B) `{O_inverts f}
               (Z : Type) `{In O Z}
      : ooExtendableAlong f (fun _ => Z).
    Proof.
      refine (cancelL_ooextendable _ _ (to O B) _ _).
      1:srapply extendable_to_O.
      refine (ooextendable_homotopic _ (O_functor f o to O A) _ _).
      1:apply to_O_natural.
      refine (ooextendable_compose _ (to O A) (O_functor f) _ _).
      2:srapply extendable_to_O.
      srapply ooextendable_equiv.
    Defined.

    (** And now the funext version *)
    Definition isequiv_precompose_O_inverts `{Funext}
               {A B : Type} (f : A -> B) `{O_inverts f}
               (Z : Type) `{In O Z}
      : IsEquiv (fun g:B->Z => g o f).
    Proof.
      srapply (equiv_extendable_isequiv 0).
      exact (ooextendable_O_inverts f Z 2).
    Defined.
```

jdchristensen:
I had no idea that your branch was progressing so well that it now contains results not in the original library!  That's great.  Should I adapt this PR using the code above?  Or do you think your branch will replace the current code soon?

mikeshulman:
I don't remember why I added that lemma -- I must have needed it for something.  It's obviously not directly related to the central change, although I ended up making more other changes at the same time than I originally intended.  I'm almost ready to share it; I just need to finish up a bit more about lex modalities.  However, we may need some discussion about various choices I made before merging it, so if you want these results now, probably better not to wait.  If you think the way I proved it seems reasonable, feel free to incorporate it into your PR.  (The other converse could also, of course, be proven in a non-funext version, if desired.)

jdchristensen:
It turns out that my other converse `yoneda_O_inverts` wasn't using `Funext`, but it was using the sort of hypothesis that you'd need `Funext` to obtain.  Replacing that assumption with an `ooExtendableAlong` hypothesis was easy.  But now, to recover my version, I need a converse to the result below that doesn't use `Funext`, and I couldn't find this.  Any suggestions?  Of course, I can just keep my existing proof, which is pretty straightforward.

```
  Definition isequiv_ooextendable `{Funext}
             {A B : Type} (C : B -> Type) (f : A -> B)
  : ooExtendableAlong f C -> IsEquiv (fun g => g oD f)
```

mikeshulman:
Ah, intriguing.  We have `equiv_ooextendable_isequiv` of course, but the fact that the backwards direction doesn't require funext (I assume it doesn't) isn't there.  It should be easy to prove though.

mikeshulman:
Wait, I'm confused.  Your `inO_yoneda_O_inverts` currently has a funext hypothesis, although it doesn't use it.  So using `equiv_ooextendable_isequiv` to prove it wouldn't actually change the hypotheses it's stated to have.  Even if it's probably possible to prove `inO_yoneda_O_inverts` without funext, since its hypothesis is difficult to obtain without funext it seems of questionable usefulness to do so.

jdchristensen:
Right, the current proof I have doesn't need the Funext hypothesis, so I could leave it like that and be done.  But since precomposition being an equivalence should imply ooExtendable, it's a one or two line proof to prove my statement using the ooExtendable one.  And it seems like an important fact, at least as motivation, to know that precomposition being an equivalence implies ooExtendable.

jdchristensen:
In fact, if the missing implication was a typeclass instance, we wouldn't even need to state the second version.

For now, I've force pushed how things look, for reference.

About universe annotations:  I've just been copying them from other theorem statements, but your functions don't have any and it all compiles fine.  If things work without them, is that best?

mikeshulman:
Currently, `ooExtendableAlong` is not even a typeclass.  We could consider making it one, but if we did I'm not sure which direction of that implication should be an instance (having both would of course cause loops).

I generally leave out universe annotations unless forced to put them in for some reason.

mikeshulman:
Let me rephrase my question: given that the hypothesis of `inO_isequiv_precompose_O_inverts` is nearly impossible to obtain without funext, why is it important to prove it in a way that doesn't use funext?

jdchristensen:
The current proof is the most obvious one you could come up with and doesn't need Funext, so it's not like avoiding Funext has a cost.  I had simply assumed it was known that without Funext, precomposition being an equivalence implies ooExtendable, so I thought I'd be able to combine that with the previous result.  But since that doesn't seem to be obvious (I played with it for a bit and kept getting stuck), I think we should just stick with the current proof.  I'll update the comments and repush.

jdchristensen:
Ah, I see what's going on.  The ooExtendable version is only using the n=1 case, ExtensionAlong.  And that much *does* follow from precomposition being an equivalence, easily.  E.g.
```
  Definition extension_isequiv_precompose
           {A : Type} {B : Type}
           (f : A -> B) (C : B -> Type)
    : IsEquiv (fun (g : forall b, C b) => g oD f) -> forall g, ExtensionAlong f C g.
  Proof.
    intros E g.
    pose (e := Build_Equiv _ _ _ E).
    exists (e^-1 g).
    apply apD10.
    exact (eisretr e g).
  Defined.
```
So maybe it's worth recording this lemma in Extensions.v and using it in the proof.

jdchristensen:
BTW, if the fact that precomposition is an equivalence implies ooExtendableAlong requires Funext, then it's not clear that you could prove ooExtendableAlong in practice without Funext.  Or are there non-trivial cases where you can?

mikeshulman:
I doubt that it does require funext.  But in any case there are plenty of examples where we can prove it without funext, such as truncation and localization modalities, using the dependent eliminator of a (H)IT.

mikeshulman:
Ok, well I'm surprised (though maybe I shouldn't have been); now I think that direction does also require funext.  To be 2-extendable means that if h and k are _homotopic_ when precomposed with f, then they are themselves homotopic -- but getting that implication from precomposition being an equivalence requires making the homotopy into an equality of functions.

To expand on my previous comment further, the main application of ooExtendableAlong is to define modalities without funext.  Of the RSUs/modalities we know, truncation, localization, and closed modalities are HITs, so they have a dependent eliminator that gives ooExtendableAlong without funext.  Most of the other RSUs/modalities that we know, such as double-negation and open modalities, are defined using function-types in the reflector, so it's natural that they do require funext to say anything about.  It will be interesting what happens with the separated RSU constructed as in CORS through the join construction.

mikeshulman:
Your suggestion to record a lemma using the n=1 case seems reasonable.

jdchristensen:
Right, I agree with everything you wrote.  I updated the changes and I think this should be ready to go.  I split the changes into two parts, the part you had already done and the new part, in case you want to add the new part to your other branch.

Until that other branch is merged, I'll avoid making other changes to this part of the library.

PR #1243, opened 2020-02-17
Shave about 2s off of four Modality/ files

jdchristensen:
There are still some very slow parts in ReflectiveSubuniverses.v, but fixing them will require more thought that I'll leave for someone else to tackle in another PR.

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -616,7 +616,7 @@ Section Reflective_Subuniverse.
           rewrite (to_O_natural (O_functor Ou)^-1 x).
           refine (to_O_natural f _ @ _).
           set (y := (O_functor Ou)^-1 x).
-          transitivity (O_functor Ou y); try apply eisretr.
+          transitivity (O_functor Ou y); [ | apply eisretr].
           unfold f, O_functor.
           apply O_rec_postcompose.
         - refine (O_indpaths _ _ _); intros x.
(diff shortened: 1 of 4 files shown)


theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -989,23 +988,23 @@ Section Reflective_Subuniverse.
...
+          rewrite <- !inv_pp, @to_O_natural_compose.
           rewrite concat_p_Vp, concat_Vp, concat_1p.
           rewrite (ap_compose coeq_to (to O (Coeq (O_functor f) (O_functor g)))).
           subst coeq_to; rewrite functor_coeq_beta_cglue.
           rewrite !ap_pp, <- !ap_compose, !inv_pp, !concat_p_pp, !ap_V, !inv_V.
-          rewrite concat_pp_V, concat_pV_p; reflexivity.
+          rewrite concat_pp_V. apply concat_pV_p.
           Close Scope long_path_scope.
-      Qed.
+      Qed.  (* This Qed is quite slow (~2s on one machine), and many of the rewrites above are slow too. *)
```
mikeshulman: I wonder if this long proof could be usefully decomposed into lemmas to speed it up.

JasonGross: Or it might be the case that it's the `cbn` at the beginning making things slow.  In general, the first look I look for `Qed` being slow is for reduction in the proof script, which is not recorded in the proof term.

jdchristensen: Both good ideas.  Some parts of that long proof could possibly be recorded as a strange specialized lemma about path algebra that can be proved trivially by path induction.

mikeshulman: For another approach, see #1245.

PR #1245, opened 2020-02-18
Experimenting with proving equiv_O_coeq using the Yoneda lemma

mikeshulman:
One of my hopes for the wild categories library was to allow more conceptual category-theoretic proofs of things like "reflections into subuniverses preserve colimits because they are left adjoints", by "chaining natural equivalences" and the Yoneda lemma.  Since the slowness of the current proof that reflectors preserve coequalizers was mentioned in #1243, I thought I would see whether we have enough wild category stuff to do that proof this way.

We did, almost.  I had to port the category-of-categories over from the wildcat branch, prove the universal property of coequalizers (did we really not have that yet?), and mess around a bit with the inclusion functor from the category of modal types into the category of all types -- in a cleaner development probably most of the `{...}`s in the proof would be separated out as lemmas.  But it works.

Advantages of this approach:

- More conceptual; easier to write and read.
- Faster.  It's hard to compare apples to apples, because I had to add other extra stuff to the library here, and also comment out some stuff in ReflectiveSubuniverse that doesn't work any more (see below), but a rough estimate told me that the compile time of ReflectiveSubuniverse is at least halved (!).

Disadvantages of this approach:

- With the version of the yoneda lemma in the current wild categories library, it requires funext.  But it would probably be possible to prove a yoneda lemma for functors taking values in wild groupoids rather than types.
- The resulting equivalence has worse computational behavior.  The computation lemmas `equiv_O_coeq_to_O` and `inverse_equiv_O_coeq_to_O` that used to have one-line proofs are no longer so easy.  I just commented them out (and the ones about pushouts derived from them) for now, so I don't know how hard they are now.

### DIFF
--- a/theories/HIT/Coeq.v
+++ b/theories/HIT/Coeq.v
@@ -52,7 +52,7 @@ Defined.
 
 Definition Coeq_unrec {B A} (f g : B -> A) {P}
            (h : Coeq f g -> P)
-  : {h : A -> P & h o f == h o g}.
+  : {k : A -> P & k o f == k o g}.
 Proof.
   exists (h o coeq).
   intros b. exact (ap h (cglue b)).
--- a/theories/HIT/Coeq.v
+++ b/theories/HIT/Coeq.v
@@ -48,6 +48,37 @@ Proof.
   refine (Coeq_ind_beta_cglue (fun _ => P) _ _ _).
 Defined.
 
+(** ** Universal property *)
+
+Definition Coeq_unrec {B A} (f g : B -> A) {P}
+           (h : Coeq f g -> P)
+  : {h : A -> P & h o f == h o g}.
+Proof.
+  exists (h o coeq).
+  intros b. exact (ap h (cglue b)).
+Defined.
+
+Definition isequiv_Coeq_rec `{Funext} {B A} (f g : B -> A) P
+  : IsEquiv (fun p : {h : A -> P & h o f == h o g} => Coeq_rec P p.1 p.2).
+Proof.
+  srapply (isequiv_adjointify _ (Coeq_unrec f g)).
+  - intros h.
+    apply path_arrow.
+    srapply Coeq_ind; intros b.
+    1:reflexivity.
+    cbn.
+    abstract (rewrite transport_paths_FlFr, concat_p1, Coeq_rec_beta_cglue, concat_Vp; reflexivity).
+  - intros [h q]; srapply path_sigma'.
+    + reflexivity.
+    + cbn.
+      rapply path_forall; intros b.
+      apply Coeq_rec_beta_cglue.
+Defined.
+
+Definition equiv_Coeq_rec `{Funext} {B A} (f g : B -> A) P
+  : {h : A -> P & h o f == h o g} <~> (Coeq f g -> P)
+  := Build_Equiv _ _ _ (isequiv_Coeq_rec f g P).
+
 (** ** Functoriality *)
 
 Definition functor_coeq {B A f g B' A' f' g'}
--- a/theories/WildCat/CatCat.v
+++ b/theories/WildCat/CatCat.v
@@ -1,0 +1,87 @@
+(* -*- mode: coq; mode: visual-line -*-  *)
+
+Require Import Basics.Overture.
+Require Import Basics.PathGroupoids.
+Require Import Basics.Notations.
+Require Import Basics.Contractible.
+Require Import Basics.Equivalences.
+Require Import WildCat.Core.
+Require Import WildCat.FunctorCat.
+Require Import WildCat.Equiv.
+Require Import WildCat.Induced.
+
+(** * Wild categories of wild categories *)
+
+(** ** The wild category of wild (0,1)-categories *)
+
+(** The category of wild (0,1)-categories is simplest, but minimally coherent. *)
+Record WildCat01 :=
+{
+  cat01_carrier : Type;
+  cat01_is01cat : Is01Cat cat01_carrier;
+}.
+
+(* note for morgan: this allows us to consider WildCats as types. *)
+Coercion cat01_carrier : WildCat01 >-> Sortclass.
+
+Global Existing Instance cat01_is01cat.
+
+Global Instance is01cat_wildcat01 : Is01Cat WildCat01.
+Proof.
+  srapply Build_Is01Cat.
+  + intros A B; exact (Fun01 A B).
+  + intros C. cbn in *.
+    exists idmap; exact _.
+  + intros A B C [G g] [F f].
+    exists (G o F); exact _.
+Defined.
+
+(** In fact [WildCat01] is probably a (wild) 1-category and even a (1,2)-category, but we haven't shown that or even defined the latter. *)
+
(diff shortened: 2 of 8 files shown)


theories/HIT/Coeq.v
```diff
@@ -48,6 +48,37 @@ Proof.
   refine (Coeq_ind_beta_cglue (fun _ => P) _ _ _).
 Defined.
 
+(** ** Universal property *)
+
+Definition Coeq_unrec {B A} (f g : B -> A) {P}
+           (h : Coeq f g -> P)
+  : {h : A -> P & h o f == h o g}.
```
jdchristensen:
```suggestion
  : {k : A -> P & k o f == k o g}.
```

theories/WildCat/CatCat.v
```diff
@@ -0,0 +1,87 @@
...
+(** ** The wild category of wild (0,1)-categories *)
+
+(** The category of wild (0,1)-categories is simplest, but minimally coherent. *)
+Record WildCat01 :=
+{
+  cat01_carrier : Type;
+  cat01_is01cat : Is01Cat cat01_carrier;
+}.
+
+(* note for morgan: this allows us to consider WildCats as types. *)
```
Alizter:
```suggestion
(* This allows us to consider WildCats as types. *)
```

theories/WildCat/CatCat.v
```diff
@@ -0,0 +1,87 @@
...
+
+Record WildCat :=
+{
+  cat_carrier : Type;
+  cat_isgraph : IsGraph cat_carrier;
+  cat_is01cat : Is01Cat cat_carrier;
+  cat_is1cat : Is1Cat cat_carrier
+}.
+
+(* note for morgan: this allows us to consider WildCats as types. *)
```
Alizter:
```suggestion
(* This allows us to consider WildCats as types. *)
```

jdchristensen:
That's very cool.  Do you think there's some way to improve the computational behaviour?

I haven't absorbed the main proof in this PR.  Does it factor into a part that shows that the coequalizer is characterized by the universal property?  That would seem like a useful stepping stone that might make things easier to understand.

mikeshulman:
By "coequalizer is characterized by the universal property" do you mean that given any type Z with the universal property, Z is equivalent to the coequalizer?  It doesn't currently factor through a lemma stated in that way, and I'm not sure offhand whether it could unless the notion of "coequalizer" were a wild-category one that could be applied to the category of modal types.

jdchristensen:
Right, thinking about it more I realize that it wouldn't factor through that.  Instead, it should factor through `O_inverts_from_isequiv_precompose` (or the Funext free version).  I took a stab at doing this naively, and it seems doable, although there's one last goal with some path algebra that I didn't work on.  (It's probably similar to the second last goal.)  Would it work to combine the wildcat approach with `O_inverts_from_isequiv_precompose`?

This is all quite fast:
```coq
      Definition coeq_cmp : Coeq f g -> Coeq (O_functor f) (O_functor g)
        := functor_coeq (to O B) (to O A)
                       (fun y => (to_O_natural f y)^)
                       (fun y => (to_O_natural g y)^).

      Local Notation O_inverts f := (IsEquiv (O_functor f)).

      Definition isequiv_O_coeq_cmp `{Funext} : O_inverts coeq_cmp.
      Proof.
        apply O_inverts_from_isequiv_precompose.
        intros Z Z_inO.
        srapply isequiv_adjointify.
        - intro h. pose (h' := O_rec (h o coeq)).
          srapply Coeq_rec.
          + exact h'.
          + cbn. apply O_indpaths; intro b.
            refine (_ @ _ @ _^); only 1,3: apply (ap h' (to_O_natural _ _)).
            refine (_ @ _ @ _^); only 1,3: apply O_rec_beta.
            rapply (ap h _); apply cglue.
        - intro h. cbn. set (h' := O_rec (h o coeq)). cbn.
          apply path_forall.
          srapply Coeq_ind; cbn.
          + intro a.
            apply (O_rec_beta (h o coeq)).
          + cbn.
            intro b.
            refine (_ @ _).
            1: rapply (transport_paths_FlFr (cglue b)).
            rewrite ap_compose.
            unfold coeq_cmp.
            rewrite functor_coeq_beta_cglue.
            rewrite inv_V.
            rewrite !ap_pp, <- !ap_compose.
            cbn.
            rewrite Coeq_rec_beta_cglue.
            rewrite O_indpaths_beta.
            rewrite ap_V.
            rewrite !inv_pp, !inv_V, !concat_p_pp.
            rewrite concat_Vp, concat_1p.
            rewrite !concat_pV_p.
            reflexivity.
        - (* Other composite, more path algebra. *)
```

jdchristensen:
It turns out to be cleaner to generalize:  functor_coeq applied to O-equivalences produces an O-equivalence.  This follows from the universal property of coequalizers along with the fact the O-equivalences give equivalences when mapping into a local type.  There's one spot where we need a dependent version of this, which should probably be a lemma.  It's all fast.
```coq
    Section OCoeq.
      Context {B A : Type} (f g : B -> A).
      Context {B' A' : Type} (f' g' : B' -> A').

      Local Notation O_inverts f := (IsEquiv (O_functor f)).

      Definition O_inverts_functor_coeq `{Funext}
                 (h : B -> B') (k : A -> A')
                 (p : k o f == f' o h) (q : k o g == g' o h)
                 `{O_inverts k} `{O_inverts h}
        : O_inverts (functor_coeq h k p q).
      Proof.
        apply O_inverts_from_isequiv_precompose.
        intros Z Z_inO.
        srapply isequiv_commsq.
        4,5: rapply (Build_Equiv _ _ _ (isequiv_Coeq_rec _ _ _)).
        { srapply equiv_functor_sigma'.
          - srapply (Build_Equiv _ _ _ (isequiv_precompose_O_inverts k Z)).
          - intro h'; cbn.
            refine (_ oE _).
            2: { nrefine (Build_Equiv _ _ (fun r => r oD h : h' o f' o h == h' o g' o h) _).
                 rapply (isequiv_extendable 0).
                 exact (snd (ooextendable_O_inverts h Z 3) _ _). }
            srapply equiv_functor_forall_id.
            intro b; cbn.
            exact (equiv_concat_r (ap h' (q b))^ _ oE equiv_concat_l (ap h' (p b)) _).
        }            
        { cbn.
          intros [h' r]; cbn.
          apply path_forall.
          srapply Coeq_ind; cbn.
          - reflexivity.
          - intro b.
            rewrite (transport_paths_FlFr (cglue b)).
            rewrite concat_p1.
            rewrite Coeq_rec_beta_cglue.
            rewrite ap_compose.
            rewrite Coeq_rec_beta_cglue.
            unfold functor_forall, "oD".
            rewrite !ap_pp.
            rewrite Coeq_rec_beta_cglue.
            do 2 rewrite <- (ap_compose coeq (Coeq_rec Z h' r)).
            cbn.
            rewrite !ap_V, !inv_pp, !inv_V, !concat_p_pp.
            rewrite !concat_pV_p.
            apply concat_pV.
        }
        all: try exact _.
      Defined.
```

jdchristensen:
I've put a cleaner version that avoids rewrites at #1252 for discussion.

PR #1246, opened 2020-02-18
bundled up equivalences in cubical library

### LINKED ISSUE #1217: bundle equivalences in cubical library
Since the discussion of #1165 I think it would be more efficient to bundle the equivalences in the cubical library. Currently a lot of time is being spent on typeclass search.

Alizter:
I have bundled up the equivalences in the cubical library. As discussed in #1217. So this closes #1217.

### DIFF
--- a/theories/Cubical/DPath.v
+++ b/theories/Cubical/DPath.v
@@ -23,28 +23,16 @@ Arguments DPath _ / _ _ _ : simpl nomatch.
 (* We can prove they are equivalent anyway *)
 Definition dp_path_transport {A : Type} {P : A -> Type}
            {a0 a1 : A} {p : a0 = a1} {b0 : P a0} {b1 : P a1}
-  : transport P p b0 = b1 -> DPath P p b0 b1.
+  : transport P p b0 = b1 <~> DPath P p b0 b1.
 Proof.
   by destruct p.
 Defined.
-
-Global Instance isequiv_dp_path_transport {A} (P : A -> Type) {a0 a1}
-  (p : a0 = a1) (b0 : P a0) (b1 : P a1)
-  : IsEquiv (@dp_path_transport A P a0 a1 p b0 b1).
-Proof.
-  srapply isequiv_adjointify; by destruct p.
-Defined.
-
-Definition equiv_dp_path_transport {A : Type} (P : A -> Type)
-           {a0 a1 : A} (p : a0 = a1) (b0 : P a0) (b1 : P a1)
-  : transport P p b0 = b1 <~> DPath P p b0 b1
-  := Build_Equiv _ _ (@dp_path_transport A P a0 a1 p b0 b1) _.
 
 Global Instance istrunc_dp {A : Type} {P : A -> Type} {n : trunc_index}
  {a0 a1} {p : a0 = a1} {b0 : P a0} {b1 : P a1} `{IsTrunc n.+1 (P a0)}
   `{IsTrunc n.+1 (P a1)} : IsTrunc n (DPath P p b0 b1).
 Proof.
-  apply (trunc_equiv _ dp_path_transport).
+  refine (trunc_equiv' _ dp_path_transport).
 Defined.
 
 Definition dp_ishprop {A : Type} (P : A -> Type) {a0 a1} {p : a0 = a1}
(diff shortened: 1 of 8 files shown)


mikeshulman:
Is there a system to when you're putting the `equiv_` prefix on a lemma that is an equivalence and when you don't?

Alizter:
I tried to reduce when I added equiv this time round to see if it is
understandable. Since unbundled versions don't exist, I thought i might as
well just call them without the equivs. If this is confusing however, I am
happy to change it.

mikeshulman:
I'm in favor of removing `equiv_` when we define something directly as an equivalence; since an equivalence can be coerced to a function it seems that not much is lost.  But why not do that in all cases like `dp_path_transport`?

Alizter:
I agree that I should change those too.

Alizter:
OK, rebased and taken in @mikeshulman 's suggestions.

mikeshulman:
This looks great, thanks.  Although I would prefer the diff noise from reindenting lines be kept out of it.  Perhaps we should think about making similar changes in `Types`.

I'm a little worried about having some of these lemmas available only in uncurried form (as is necessary, of course, to state them as equivalences).  But it doesn't seem to be causing problems yet?

Alizter:
>But it doesn't seem to be causing problems yet?
No, but the general rule of thumb is if `apply some_function` worked, but no longer does when changed to `apply equiv_some_function` due to uncurrying, `apply (equiv_some_function (_;_)` probably still does.

It might be worth making a notation `curry := (fun f x y => f (x; y))` to help with these situations.

>Perhaps we should think about making similar changes in Types.

What kinds of changes did you have in mind?

mikeshulman:
For instance, defining `path_prod` to be what is currently called `equiv_path_prod`, and so on.

jdchristensen:
I haven't looked closely at this, but renaming things to remove "equiv_" prefixes makes it a bit harder to find equivalences you are looking for.  Maybe that's a worthwhile compromise, but I'm just pointing it out.

Alizter:
Dan I absolutely agree with you. I have been pondering this in my head for
quite a while now and I don’t know what to think.

Perhaps I could add notations with the `equiv_` in front so that it also
works?

mikeshulman:
@jdchristensen You mean, because you can't any longer restrict your looking to things that start with `equiv_`?

I wouldn't mind indicating things that are equivalences somehow, but `equiv_` is so verbose.  For a while I've been thinking that whether or not we can actually use wild categories to define the functorial actions in `Types`, it would be nice to adapt the terminology and change the names like `functor_sigma` and `equiv_functor_sigma` to shorter things like `fmap_sigma` and `emap_sigma`.

jdchristensen:
Yes, I often search through the .v files for 'equiv_' (or many other things) to try to find what I want in the library.  I don't mind if it is something else like emap, or maybe something involving a capital E like in oE.

mikeshulman:
Well, emap makes sense for the action of a functor on equivalences, but not for other ways of constructing equivalences.  A capital E on a function name would, I think, look kind of ugly.  What about just a shorter prefix like `eq_`?  I could live with that I think.  In principle it might be mistaken for referring to "equality", but we never do that in this library (we always use "path" when talking about the identity type), so I don't think it would be a problem.

mikeshulman:
And `fmap_sig` and `eq_fmap_sig` would be reasonably short.

jdchristensen:
Sounds fine to me.  About the capital E:  I was wondering if notation 1 for the identity equivalence should be changed to 1E.  It's a bit ugly, but not as ugly as the 1%equiv's we sometimes need.  Or maybe eq_idmap.

mikeshulman:
I'm actually not entirely sold on using `1` for identities of any sort.  It always annoys me when I want to write the natural number `1`, or the trunc_index `1`, and Coq thinks I mean the identity path.

jdchristensen:
Exactly.  I don't think we should overload any nats, and I was also going to propose getting rid of 1 for idpath.  idpath is short.  Similarly, 0 for the zero modality is not useful enough to make it worth the trouble when you more likely mean (Tr 0).

mikeshulman:
I think the Agda library uses `idp`, which is even shorter, and can be combined with `idf` and `ide` for identity functions and equivalences.

What is the zero modality?

jdchristensen:
When Coq expects a Subuniverse and I try to supply `0`, I get an error about `Zero Subuniverse`.  I guess I was misinterpreting the error.  I thought `0` was overloaded as a notation for some particular Subuniverse.  But instead `0` is overloaded as `zero` in the `Zero` typeclass.  I didn't even know there was a `Zero` typeclass.  Can we get rid of that `0` notation, or localize it to where it is needed?

mikeshulman:
Hmm... that must be a mathclasses thing?

mikeshulman:
Would it work to say

```coq
Global Instance zero_subuniverse : Zero Subuniverse := Tr 0
```

?

jdchristensen:
I was also wondering about making Subuniverse have a Zero element, but the most natural subuniverse to call trivial would be the subuniverse of contractible types.  And I'm guessing that Zero is intended to have a different, more algebraic meaning.  So it would be a workaround for a notation that is getting in the way unnecessarily.  Can we limit the interpretation of 0 as a Zero element to a scope that we only open when needed?

About "eq_".  Would we use "iseq_" as well?  To be honest, I don't mind "equiv_" and "isequiv_", since they nicely describe the types they live in.

Alizter:
@jdchristensen Yes, the Zero class from hottclasses can definitely have its notation limited. In fact, I am probably going to eventually close all hottclasses notation scopes by default.

For now I would just close `mc_scope` or open `nat_scope` or `trunc_scope` on top of it so that it overrides the notation.

PR #1248, opened 2020-02-19
Import IsGraph-as-a-parameter from wildcat branch

mikeshulman:
The original definition of wild categories had the underlying directed graph as a field of the typeclass rather than a parameter.  This was kind of a kludge -- I was pretty sure even at the time that it ought to be a parameter, but the graph typeclass was introduced while other people were working on various stuff downstream and we didn't want to change the foundations too much.  Recently @Alizter went back and made IsGraph into a parameter the way it should be in the wildcat branch; this PR imports that change back into master.

### DIFF
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -365,10 +365,10 @@ Proof.
 Defined.
 
 (** The wild cat of Groups *)
-Global Instance isgraph_Group : IsGraph Group
+Global Instance isgraph_group : IsGraph Group
   := Build_IsGraph Group GroupHomomorphism.
 
-Global Instance is01cat_Group : Is01Cat Group :=
+Global Instance is01cat_group : Is01Cat Group :=
   (Build_Is01Cat Group _ (@grp_homo_id) (@grp_homo_compose)).
 
 Global Instance isgraph_grouphomomorphism {A B : Group} : IsGraph (A $-> B)
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -364,27 +364,32 @@ Proof.
   1,2: apply grp_homo_op.
 Defined.
 
-(** Group forms a 01Cat *)
+(** The wild cat of Groups *)
+Global Instance isgraph_Group : IsGraph Group
+  := Build_IsGraph Group GroupHomomorphism.
+
 Global Instance is01cat_Group : Is01Cat Group :=
-  (Build_Is01Cat Group GroupHomomorphism (@grp_homo_id) (@grp_homo_compose)).
+  (Build_Is01Cat Group _ (@grp_homo_id) (@grp_homo_compose)).
 
-Global Instance is01cat_GroupHomomorphism {A B : Group} : Is01Cat (A $-> B) :=
-  induced_01cat (@grp_homo_map A B).
+Global Instance isgraph_grouphomomorphism {A B : Group} : IsGraph (A $-> B)
+  := induced_graph (@grp_homo_map A B).
 
-Global Instance is0gpd_GroupHomomorphism {A B : Group}: Is0Gpd (A $-> B) := 
-  induced_0gpd (@grp_homo_map A B).
+Global Instance is01cat_grouphomomorphism {A B : Group} : Is01Cat (A $-> B)
+  := induced_01cat (@grp_homo_map A B).
 
-Global Instance is0functor_postcomp_GroupHomomorphism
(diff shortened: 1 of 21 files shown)


theories/Algebra/Group.v
```diff
@@ -364,27 +364,32 @@ Proof.
   1,2: apply grp_homo_op.
 Defined.
 
-(** Group forms a 01Cat *)
+(** The wild cat of Groups *)
+Global Instance isgraph_Group : IsGraph Group
```
Alizter:
```suggestion
Global Instance isgraph_group : IsGraph Group
```

theories/Algebra/Group.v
```diff
@@ -364,27 +364,32 @@ Proof.
...
+  := Build_IsGraph Group GroupHomomorphism.
+
 Global Instance is01cat_Group : Is01Cat Group :=
```
Alizter:
```suggestion
Global Instance is01cat_group : Is01Cat Group :=
```

theories/Pointed/pType.v
```diff
@@ -43,14 +59,11 @@ Defined.
 Global Instance hasmorext_ptype `{Funext} : HasMorExt pType.
 Proof.
   srapply Build_HasMorExt; intros A B f g.
-  refine (isequiv_homotopic (equiv_path_pmap f g)^-1 _).
-  intros []; reflexivity.
-Defined.
-
+Abort.
```
Alizter: Why have we gotten rid of this proof?

theories/WildCat/Opposite.v
```diff
@@ -19,30 +13,48 @@ Notation "A ^op" := (op A).
...
-    exact (h $@L p).
-  - intros a b c d f g h; exact (cat_assoc_opp h g f).
-  - intros a b f; exact (cat_idr f).
-  - intros a b f; exact (cat_idl f).
- Defined.
+Section Op.
+
+  Context (A : Type) `{Is1Cat A}.
+
+  Global Instance isgraph_op : IsGraph A^op.
```
Alizter: So the `{Is1Cat A}` here generates all the instances it relies on. But this doesn't mean isgraph_op relies on an `Is1Cat`. It might be a bit confusing when reading this so perhaps it is worth putting in a comment like: `(** [Is1Cat A] generates [IsGraph A] and so on using generalized variables. Close the section [Op] and [About] the definition of [isgraph_op] to see how this works. **)`.

theories/WildCat/Opposite.v
```diff
@@ -136,7 +148,7 @@ Defined.
 
 Definition transformation_op {A} {B} `{Is01Cat B}
            (F : A -> B) (G : A -> B) (alpha : F $=> G)
-  : (@Transformation (A^op) (B^op) (@isgraph_1cat _ (is01cat_op B))
+  : (@Transformation (A^op) (B^op) _
```
Alizter:
```suggestion
  : @Transformation A^op B^op _
```

theories/WildCat/Opposite.v
```diff
@@ -136,7 +148,7 @@ Defined.
...
                      (G : (A^op) -> (B^op)) (F : (A^op) -> (B^op))).
```
Alizter:
```suggestion
                     (G : A^op -> B^op) (F : A^op -> B^op).
```

theories/Pointed/pType.v
mikeshulman: I'm not sure what happened; it seems to have disappeared in 56ab502eac3dc855dd6906b5aa0118e702c2e071.  I've put it back.

theories/WildCat/Opposite.v
mikeshulman: Why here specifically?  We do this sort of thing all over the place.

Alizter: Because I added this one in and reading it again, I think I could have made it clearer.

mikeshulman: But since we do it all over the place, I don't think this is a good place to put such an explanation: if someone doesn't understand generalizing binders, they're going to be confused long before this.

mikeshulman:
Github actions seem to be much faster than Travis was.

theories/WildCat/Opposite.v
Alizter: OK

Alizter:
We can run a lot more jobs at the same time too. The update html/proviola stuff still takes time however.

PR #1252, opened 2020-02-19
Add O_inverts_functor_coeq

jdchristensen:
This allows us to prove `isequiv_O_coeq_cmp` without rewrites.

But the proof of `equiv_O_pushout_to_O_pushl` gets stuck.  See the comments there.  Any ideas what's going on?

This PR includes the universal property of coequalizers, from #1245 .

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -905,25 +905,31 @@ Section Reflective_Subuniverse.
 
     (** ** Coproducts *)
 
-    Definition equiv_O_sum {A B} :
-      O (A + B) <~> O (O A + O B).
+    Local Notation O_inverts f := (IsEquiv (O_functor f)).
+
+    Definition O_inverts_sum `{Funext} {A B A' B'} (f : A -> A') (g : B -> B')
+               `{O_inverts f} `{O_inverts g}
+      : O_inverts (functor_sum f g).
     Proof.
-      simple refine (equiv_adjointify _ _ _ _).
-      - apply O_rec; intros x.
-        exact (to O _ (functor_sum (to O A) (to O B) x)).
-      - apply O_rec; intros [x|x].
-        + (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4525, stack overflow in exact *)
-          let lem := constr:(fun A B => to O _ o @inl A B) in
-          exact (O_rec (lem _ _) x).
-        + let lem := constr:(fun A B => to O _ o @inr A B) in
-          exact (O_rec (lem _ _) x).
-      - apply O_indpaths; intros [x|x].
-        all:revert x; apply O_indpaths; intros x.
-        all:abstract (rewrite !O_rec_beta; reflexivity).
-      - apply O_indpaths; intros [x|x].
-        all:abstract (rewrite !O_rec_beta; cbn;
-                      rewrite !O_rec_beta; reflexivity).
+      apply O_inverts_from_isequiv_precompose.
+      intros Z Z_inO.
+      srapply isequiv_commsq'; cbn.
+      4,5: rapply ((equiv_sum_ind (fun _ => Z))^-1).
+      { srapply equiv_functor_prod'.
+        - srapply (Build_Equiv _ _ _ (isequiv_precompose_O_inverts f Z)).
+        - srapply (Build_Equiv _ _ _ (isequiv_precompose_O_inverts g Z)). }
+      1: reflexivity.
+      all: try exact _.
     Defined.
+
+    Definition equiv_O_functor_sum `{Funext} {A B A' B'} (f : A -> A') (g : B -> B')
+               `{O_inverts f} `{O_inverts g}
+      : O (A + B) <~> O (A' + B')
+      := Build_Equiv _ _ _ (O_inverts_sum f g).
+
+    Definition equiv_O_sum `{Funext} {A B} :
+      O (A + B) <~> O (O A + O B)
+      := equiv_O_functor_sum (to O A) (to O B).
 
     (** ** Coequalizers *)
 
@@ -973,6 +979,14 @@ Section Reflective_Subuniverse.
         all: try exact _.
       Defined.
 
+      Definition equiv_O_functor_coeq
+                 {B' A' : Type} (f' g' : B' -> A')
+                 (h : B -> B') (k : A -> A')
+                 (p : k o f == f' o h) (q : k o g == g' o h)
+                 `{O_inverts k} `{O_inverts h}
+        : O (Coeq f g) <~> O (Coeq f' g')
+        := Build_Equiv _ _ _ (O_inverts_functor_coeq f' g' h k p q).
+
       Definition coeq_cmp : Coeq f g -> Coeq (O_functor f) (O_functor g)
         := functor_coeq (to O B) (to O A)
                        (fun y => (to_O_natural f y)^)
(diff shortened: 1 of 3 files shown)


theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1086,9 +1031,11 @@ Section Reflective_Subuniverse.
         : equiv_O_pushout (to O (Pushout f g) (pushl b))
           = to O (Pushout (O_functor f) (O_functor g)) (pushl (to O B b)).
       Proof.
-        cbn.
-        unfold O_coeq_cmp, O_coeq_cmp_inverse, Pushout, pushl, push.
-        rewrite to_O_natural.
+        (*cbn.  Spins. *)
+        unfold equiv_O_pushout, equiv_O_coeq, coeq_cmp, Pushout, pushl, push.
+        unfold "oE".
```
Alizter: I didn't know you could unfold notations like this!

jdchristensen: Yep!

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1086,9 +1031,11 @@ Section Reflective_Subuniverse.
         : equiv_O_pushout (to O (Pushout f g) (pushl b))
           = to O (Pushout (O_functor f) (O_functor g)) (pushl (to O B b)).
       Proof.
-        cbn.
-        unfold O_coeq_cmp, O_coeq_cmp_inverse, Pushout, pushl, push.
-        rewrite to_O_natural.
+        (*cbn.  Spins. *)
```
Alizter: There is a huge amount of memory being used here.

Alizter: OK so doing this
```coq

        unfold equiv_O_pushout.
        set (p := (equiv_inverse
              (equiv_O_coeq (fun x : rsu_reflector O A => inl (O_functor f x))
                 (fun x : rsu_reflector O A => inr (O_functor g x))))).
        clearbody p.
        simpl.
```

before the proof, proves that the whole cbn spinning thing stems from the inversion proof of equiv_O_coeq. Something in there is causing coq to hang.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -928,128 +928,73 @@ Section Reflective_Subuniverse.
...
+      Definition coeq_cmp : Coeq f g -> Coeq (O_functor f) (O_functor g)
+        := functor_coeq (to O B) (to O A)
+                       (fun y => (to_O_natural f y)^)
+                       (fun y => (to_O_natural g y)^).
 
-      Global Instance isequiv_O_coeq_cmp
-      : IsEquiv O_coeq_cmp
-        := isequiv_adjointify _ O_coeq_cmp_inverse
-                              O_coeq_cmp_eisretr O_coeq_cmp_eissect.
+      Global Instance isequiv_O_coeq_cmp `{Funext} : O_inverts coeq_cmp.
```
Alizter: These `Funext`'s are already in the context. So there are multiple copies floating around. This could be causing coq some trouble.

jdchristensen: Good point, but I was having trouble before I made that mistake.  I'll fix it.

Alizter:
OK here is a proof. The trick is to not unfold `equiv_O_coeq`:
```coq
        unfold equiv_O_pushout.
        unfold "oE", equiv_compose.
        unfold equiv_fun.
        refine (_ @ _).
        { apply ap.
          apply ap.
          apply equiv_O_coeq_to_O. }
        unfold equiv_O_functor.
        refine (_ @ _).
        { apply ap.
          apply to_O_natural. }
        unfold equiv_O_coeq.
        unfold equiv_fun.
        apply moveR_equiv_V.
        simpl.
        rewrite O_rec_beta.
        simpl.
        unfold O_functor.
        rewrite O_rec_beta.
        reflexivity.
      Qed.
```

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1012,52 +1026,35 @@ Section Reflective_Subuniverse.
...
-        cbn.
-        rewrite O_rec_beta.
-        rewrite to_O_natural.
-        reflexivity.
-      Qed.
+        rapply to_O_natural.
+      Defined.
 
       Definition inverse_equiv_O_pushout_to_O_pushl (b : B)
         : equiv_O_pushout^-1 (to O (Pushout (O_functor f) (O_functor g)) (pushl (to O B b)))
```
Alizter: these `^-1` can sometimes cause coq to hang. Adding an `%equiv` makes them more reliable. Soon `^-1` will mean the `%equiv` version will be the default.

Alizter:
It seems github had some downtime today: https://www.githubstatus.com/

jdchristensen:
I found a 1-line proof of those problematic lemmas by cleaning up some earlier stuff.  Lots of rewrites are now gone from the file and things are shorter and faster.  The trade-off is that Funext is used in more places.  I had to add Funext to one section of BlakersMassey.  Now the whole library compiles.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -905,142 +905,101 @@ Section Reflective_Subuniverse.
 
     (** ** Coproducts *)
 
-    Definition equiv_O_sum {A B} :
-      O (A + B) <~> O (O A + O B).
+    Local Notation O_inverts f := (IsEquiv (O_functor f)).
+
+    Definition O_inverts_sum `{Funext} {A B A' B'} (f : A -> A') (g : B -> B')
```
Alizter: Is this the reason why Funext got introduced in BlakersMassey?

jdchristensen: I think it comes from the Funexts in the Pushout part of ReflectiveSubuniverse.v

jdchristensen:
Like with Mike's approach, these changes reduce the build time of ReflectiveSubuniverse.v by more than a factor of 2.  I'll run some timing tests.

mikeshulman:
Nice!  I suspect that the funext-free approach will be just as fast and maybe even better.

jdchristensen:
You mean using `ooextendable_O_inverts` instead of `isequiv_precompose_O_inverts`?  Yes, I agree that that would be better.  But we don't yet have all of the tools in place to work with ooextendable maps, e.g. things like equiv_functor_sigma, so I'm not sure how to go about this.  If someone wants to take a stab at it, please go ahead and feel free to push to this PR or create a new one for comparison.

The output of the timing is simple to summarize:  ReflectiveSubuniverse went down by 4-5s, the whole library went down by 3s, and no other file changed by more than 1s.

mikeshulman:
I hope I can try the ooextendable approach soon.  But I suppose we can merge this for now.

mikeshulman:
Hmm, sorry, I should have looked at this more closely.  Why did you change the proof of `equiv_O_sum`?  The direct approach in that case is pretty simple, and I didn't think it was very slow.

jdchristensen:
I needed the more general fact that functor_sum sends O-equivalences to O-equivalences in the new proof that pushouts do the same, so I proved it in parallel to how I proved that fact for coequalizers.  The proof of this more general fact is shorter than the original proof of equiv_O_sum, but does depend on Funext.  We could keep the original proof of equiv_O_sum and the new proof of the more general fact, but that feels a bit wasteful.  If you have a preferred way to handle this, let me know, or feel free to make changes.  I suspect that a lot of things can be improved in this area and that this PR is not the last word.

mikeshulman:
Makes sense.  How about #1260?

PR #1254, opened 2020-02-19
Modulo map for Fin

### LINKED ISSUE #1034: How to define `nat -> Fin n.+1`

Alizter:
This closes #1034

### DIFF
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -51,12 +51,14 @@ Proof.
   elim (f (inr tt)).
 Defined.
 
+(** The zeroth element of a non-empty finite set is the left most element. It also happens to be the biggest by termsize. *)
 Fixpoint fin_zero (n : nat) : Fin n.+1 :=
   match n with
   | O => inr tt
   | S n' => inl (fin_zero n')
   end.
 
+(** There is an injection from Fin n -> Fin n.+1 that maps the kth element to the (k+1)th element. *)
 Fixpoint fin_finS_inject (n : nat) : Fin n -> Fin n.+1 :=
   match n with
   | O => Empty_rec
@@ -68,6 +70,7 @@ Fixpoint fin_finS_inject (n : nat) : Fin n -> Fin n.+1 :=
       end
   end.
 
+(** This injection is an injection/embedding *)
 Lemma isembedding_fin_finS_inject (n : nat) : IsEmbedding (fin_finS_inject n).
 Proof.
   apply isembedding_isinj_hset.
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -51,6 +51,35 @@ Proof.
   elim (f (inr tt)).
 Defined.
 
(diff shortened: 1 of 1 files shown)


Alizter:
I like the second one, ill need to think about the name. Andreas wrote the
inject so I will discuss it with him tomorrow.

Alizter:
@mikeshulman OK I have gone with your name changes.

mikeshulman:
Great, thanks!

theories/Spaces/Finite.v
```diff
@@ -1044,6 +1076,61 @@ Section Enumeration.
...
+  + destruct n' as [|n''].
+    1: revert r; apply Empty_rec.
+    destruct r as [r|].
+    - hnf; apply inl.
+      apply fsucc.
+      hnf; apply inl.
+      exact r.
+    - exact (inr tt).
+  + exact (fin_zero n').
+Defined.
```
abooij: This construction of `fsucc_mod` looks a bit cumbersome to me. It should be as follows, I think:

- If `n=0` do anything (e.g. the identity map, which may compute better than `Empty_rec`).
- For `S n` destruct the incoming element of `Fin n.+1`: if it's `inl x`, output `fsucc _ x`, otherwise output `fin_zero _`.

theories/Spaces/Finite.v
```diff
@@ -51,6 +51,38 @@ Proof.
   elim (f (inr tt)).
 Defined.
 
+(** The zeroth element of a non-empty finite set is the left most element. It also happens to be the biggest by termsize. *)
+Fixpoint fin_zero (n : nat) : Fin n.+1 :=
```
abooij: In my experiments in #1261 it worked very well to have the `n` argument implicit. I'm not saying that is necessarily better, but in my case it was convenient.

theories/Spaces/Finite.v
Alizter: Thanks, this is an excellent suggestion!

theories/Spaces/Finite.v
Alizter: I agree.

Alizter:
Thanks @abooij, fsucc_mod is now way simpler! We had a discussion in #1034 and there was a map defined in agda by pattern matching that I tried to adapt. I like your version the best.

Alizter:
@mikeshulman `fsucc_mod` now has a much simpler definition.

Alizter:
Thanks for those suggestions @abooij

PR #1255, opened 2020-02-19
updated STYLE.md with notes about Github actions

Alizter:
I have removed mention of travis from STYLE.md and replaced it with mention of github actions. I have also added some documentation on the jobs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-[![Build Status](https://api.travis-ci.org/HoTT/HoTT.png?branch=master)](https://travis-ci.org/HoTT/HoTT)
+![Build Status](https://github.com/HoTT/HoTT/workflows/CI/badge.svg)
 
 [Homotopy Type Theory](http://homotopytypetheory.org/) is an interpretation of
 Martin-Löf’s intensional type theory into abstract homotopy theory. Propositional equality
--- a/STYLE.md
+++ b/STYLE.md
@@ -1220,7 +1220,7 @@ made to the central repository.
 
 When you submit a PR to github, Github Actions will start running in
 the background. You can see the checks at the bottom by clicking
-`Show all checks` or by clicking the `Checks` tab at the top inbetween
+`Show all checks` or by clicking the `Checks` tab at the top in between
 `Commits` and `Files changed`.
 
 There should be 7 different jobs with the following names. Here is
--- a/STYLE.md
+++ b/STYLE.md
@@ -54,7 +54,7 @@
     - [Approval of pull requests](#approval-of-pull-requests)
     - [Commit messages](#commit-messages)
     - [Creating new files](#creating-new-files)
-    - [Travis](#travis)
+    - [Github Actions](#github-actions)
     - [Git rebase](#git-rebase)
     - [Timing scripts](#timing-scripts)
   - [Bugs in Coq](#bugs-in-coq)
@@ -131,8 +131,8 @@ They are currently in several groups:
 
 A dependency graph of all the files in the library can be found on the
 [wiki][wiki]; this may be helpful in avoiding circular dependencies.
-It is updated automatically by Travis (see below) on every push to the
-master branch.
+It is updated automatically by Github Actions (see below) on every
+push to the master branch.
 
 [wiki]: https://github.com/HoTT/HoTT/wiki
 
@@ -394,13 +394,21 @@ names `H`, your name will result in a conflict.  Conversely, we
 sometimes give a hypothesis a name that won't be used, to pre-empt
 such conflicts, such as `{ua : Univalence}` or `{fs : Funext}`.
 
-One gotcha about typeclass arguments is that they cannot be inferred automatically when preceeded by non-implicit arguments.  So for instance if we write
+One gotcha about typeclass arguments is that they cannot be inferred
+automatically when preceeded by non-implicit arguments.  So for
+instance if we write
 
 ```coq
 Definition foo (A : Type) `{Funext}
 ```
 
-then the `Funext` argument will not generally be inferrable.  Thus, typeclass arguments should generally come first if possible.  In addition, note that when section variables are generalized at the close of a section, they appear first.  Thus, if anything in a section requires `Funext` or `Univalence`, those hypotheses should go in the `Context` at the top of the section in order that they'll come first in the eventual argument lists.
+then the `Funext` argument will not generally be inferrable.  Thus,
+typeclass arguments should generally come first if possible.  In
+addition, note that when section variables are generalized at the
+close of a section, they appear first.  Thus, if anything in a section
+requires `Funext` or `Univalence`, those hypotheses should go in the
+`Context` at the top of the section in order that they'll come first
+in the eventual argument lists.
 
 ### Truncation ###
 
@@ -740,7 +748,8 @@ Sometimes binders without type annotations, like `forall n, foo n`
 where `foo : nat -> Type0`, will produce a fresh universe for
 the variable's type, eg `forall n : (? : Type@{fresh}), foo n`,
 which will remain in the definition as a phantom type:
-`fresh |= forall n : nat, foo n`. Annotating the binder will get rid of it.
+`fresh |= forall n : nat, foo n`. Annotating the binder will get rid
+of it.
 See also [bug #4868](https://coq.inria.fr/bugs/show_bug.cgi?id=4868).
 
 ### Lifting and lowering ###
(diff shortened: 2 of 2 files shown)


STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+We use [Github Actions][actions] to check that pull requests do not
+break anything, and also to automatically update various things (such
+as the documentation, proviola, and dependency graph linked on the
+[project wiki][wiki]).  Normally you shouldn't need to know anything
+about this; Github Actions automatically checks every pull request
+made to the central repository.
+
+When you submit a PR to github, Github Actions will start running in
+the background. You can see the checks at the bottom by clicking
+`Show all checks` or by clicking the `Checks` tab at the top inbetween
```
jdchristensen:
```suggestion
`Show all checks` or by clicking the `Checks` tab at the top in between
```

STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+
+There should be 7 different jobs with the following names. Here is
+also a description of what they do.
+
+1. **CI / build (pull_request)** This is the main job. It simply
+installs the library and checks that everything compiles with the
+recommended version of coq. This takes around ~5m.
+2. **CI / build (update dep graphs, yes, yes) (pull_request)** This
+builds the dependency graphs which can be viewed on the
+[project wiki][wiki]. This takes around ~12m.
```
jdchristensen: Why does it take 12m to make a dependency graph?

STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+3. **CI / build (update quick doc, yes) (pull_request)** This updates
+the documentation and takes around ~7m.
+4. **CI / build (update html, yes) (pull_request)** This updates the
+proviola documentation. This can take up to ~30m.
+5. **CI / build (validate, yes) (pull_request)** Builds the library
+and then validates it afterwards. Takes around ~5m.
```
jdchristensen: Can you clarify how this is different from number 1?  Can it just be combined with 1?

STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+6. **CI / build (with submodule coq, yes) (pull_request)** This builds
+the library with the recommended version of coq in the submodule, then
+ fastforwards it to the master version of coq to check if it still
+ compiles. This takes around ~11m.
```
jdchristensen: Is the second half of 6 the same as 7?

jdchristensen:
Thanks for clarifying things!  I always wondered why there were so many jobs.  Would it make sense to only do the slow ones involving doc updates when merging with master, or something like that?

Alizter:
@jdchristensen Now that we have moved to Github actions, it might be possible to build documentation independently of a pull request. At the moment, we have simply adapted what travis was doing to github actions. As for your other question, I don't know. But @JasonGross probably does.

Alizter:
I've also changed the `Build Passing` bade in the README.md. @JasonGross have I done this correctly? I can't work out how to link it to the current build.

STYLE.md
```diff
@@ -1138,19 +1147,19 @@ comment such as LGTM ("Looks Good To Me").  Currently the rules are:
...
-  requests have been merged in the interval since a given pull request
-  was first submitted, it may be necessary to rebase that pull request
-  against the new master, to make
-  sure before merging it that it won't break the master branch.
+  Note also that Github Actions doesn't automatically restart itself
+  on a pull request when the master branch changes.  Thus, if other
+  pull requests have been merged in the interval since a given pull
+  request was first submitted, it may be necessary to rebase that pull
+  request against the new master, to make sure before merging it that
+  it won't break the master branch.
```
JasonGross: We should check to see if GitHub merges the PR with master.  If not, we should do it ourselves, or update the doc here accordingly.

STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+1. **CI / build (pull_request)** This is the main job. It simply
+installs the library and checks that everything compiles with the
+recommended version of coq. This takes around ~5m.
+2. **CI / build (update dep graphs, yes, yes) (pull_request)** This
+builds the dependency graphs which can be viewed on the
+[project wiki][wiki]. This takes around ~12m.
+3. **CI / build (update quick doc, yes) (pull_request)** This updates
+the documentation and takes around ~7m.
+4. **CI / build (update html, yes) (pull_request)** This updates the
+proviola documentation. This can take up to ~30m.
```
JasonGross: Also the coqdoc documentation, I think.

STYLE.md
```diff
@@ -1200,16 +1209,44 @@ You will probably also want to add your new file to `HoTT.v`, unless
...
+5. **CI / build (validate, yes) (pull_request)** Builds the library
+and then validates it afterwards. Takes around ~5m.
+6. **CI / build (with submodule coq, yes) (pull_request)** This builds
+the library with the recommended version of coq in the submodule, then
+ fastforwards it to the master version of coq to check if it still
```
JasonGross: What?  This seems wrong.  This job should not fastfoward the module.

README.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://api.travis-ci.org/HoTT/HoTT.png?branch=master)](https://travis-ci.org/HoTT/HoTT)
+![Build Status](https://github.com/HoTT/HoTT/workflows/CI/badge.svg)
```
JasonGross: This seems fine. You could use `https://github.com/HoTT/HoTT/workflows/CI/badge.svg?branch=master&event=push` if you want to be more specific (I'm not sure if the one you have now excludes pull requests)

STYLE.md
JasonGross: The second half of 6 should not exist.  If it does, that's a bug.

STYLE.md
JasonGross: It can be combined with 1, if we want.  The difference is that here we also run coqchk, after running coqc.

STYLE.md
JasonGross: Because there's [a separate dependency graph for every constant in the library](http://hott.github.io/HoTT/file-dep-graphs/).  It's currently broken, though, and we should fix that.

JasonGross:
> Would it make sense to only do the slow ones involving doc updates when merging with master, or something like that?

No, if some PR manages to somehow break only the documentation, we want to know that.  The builds are run in parallel, though, so it shouldn't add too much overhead.

jdchristensen:
> Would it make sense to only do the slow ones involving doc updates when merging with master, or something like that?

> No, if some PR manages to somehow break only the documentation, we want to know that. The builds are run in parallel, though, so it shouldn't add too much overhead.

Since problems with the docs seem to be rare, I'd be fine with only learning about issues with, say, the dependency graph or proviola after merging with master, if that means we can be a bit more nimble with code changes.

mikeshulman:
Do github actions run automatically on all forks (unlike travis)?

JasonGross:
> Do github actions run automatically on all forks (unlike travis)?

I think do, but am not sure

STYLE.md
Alizter: OK then I have misunderstood what is being done here.

Alizter:
I can't workout how to restart Github actions builds. I am also not sure our jobs are restarting when something is pushed to master. Currently this means we have regressed in functionality. @JasonGross how can we restart jobs before merging to make sure things still build?

JasonGross:
Does closing and reopening a PR work?

Alizter:
Hmmm that seems to work.

JasonGross:
It seems that closing and reopening a PR restarts the jobs.  I don't think it's possible to restart successful jobs in any other way, though if a job fails, you can rerun the workflow

Alizter:
@JasonGross I tried looking at the documentation for this but I couldn't find anything. Are we able to restart builds when master has been pushed to?

JasonGross:
You mean set an action that automatically restarts builds?  We can probably make a job that runs on push to master that gets a list of the open PRs using the GitHub API and then closes and reopens each one.

Alizter:
@JasonGross Is there any other way to trigger jobs other than opening and closing them?

mikeshulman:
What's the status of this?

JasonGross:
> @JasonGross Is there any other way to trigger jobs other than opening and closing them?

Not that I'm aware of.  But I haven't looked to deeply

STYLE.md
Alizter: What does 6 actually do.

JasonGross: It uses the submodule version of Coq rather than using master from my Launchpad packages

Alizter:
I'm closing this issue as it will be subsumed by future changes I plan to submit soon.

PR #1261, opened 2020-02-20
Implement a 1-dimensional Sperner's lemma

abooij:
Given a finite sequence of bits, if the first bit is 1 and the last bit is 0, then there is a point where the sequence swaps from 1 to 0. This proves that.

### DIFF
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -1115,6 +1115,11 @@ Ltac FinIndOn X := repeat
 (** This tactic can be used to generate n cases from a goal like forall (x : Fin n), _ *)
 Ltac FinInd := let X := fresh "X" in intro X; FinIndOn X.
 
+(** The 1-dimensional version of Sperner's lemma says that given any
+finite sequence of decidable hProps, where the sequence starts with
+true and ends with false, we can find a point in the sequence where
+the sequence changes from true to false. This is like a discrete
+intermediate value theorem. *)
 Section Sperner.
 
   Fixpoint sperners_lemma_1d {n} : forall (f : Fin (n.+2) -> DHProp)
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -26,6 +26,27 @@ Fixpoint Fin (n : nat) : Type
        | S n => Fin n + Unit
      end.
 
+Fixpoint Fin_left {n : nat} : Fin (S n)
+  := match n with
+     | 0 => inr tt
+     | S n => inl Fin_left
+     end.
+
+Definition Fin_right {n : nat} : Fin (S n) := inr tt.
+
+Definition Fin_inc_l {n : nat} (k : Fin n) : Fin (S n) := inl k.
+Fixpoint Fin_inc_r {n : nat} : Fin n -> Fin (S n)
+  := match n with
+     | 0 => fun k =>
+              match k with
+              end
+     | S n => fun k =>
+                match k with
+                | inl l => inl (Fin_inc_r l)
+                | inr tt => inr tt
+                end
+     end.
+
 Global Instance decidable_fin (n : nat)
 : Decidable (Fin n).
 Proof.
@@ -1108,3 +1129,22 @@ Ltac FinIndOn X := repeat
 
 (** This tactic can be used to generate n cases from a goal like forall (x : Fin n), _ *)
 Ltac FinInd := let X := fresh "X" in intro X; FinIndOn X.
+
+Section Sperner.
+
+  Fixpoint sperners_lemma_1d {n} : forall (f : Fin (n.+2) -> DHProp)
+        (left_true : f Fin_left)
+        (right_false : ~ f Fin_right),
+    {k : Fin n.+1 & f (Fin_inc_l k) /\ ~ f (Fin_inc_r k)}.
+  Proof.
+    intros ???.
+    destruct n as [|n].
+    - exists Fin_left. split; assumption.
+    - destruct (dec (f (Fin_inc_l Fin_right))) as [prev_true|prev_false].
+      + exists Fin_right. split; assumption.
+      + destruct (sperners_lemma_1d _ (f o Fin_inc_l) left_true prev_false) as [k' [fleft fright]].
+        exists (Fin_inc_l k').
+        split; assumption.
+  Defined.
+
+End Sperner.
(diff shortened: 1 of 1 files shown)


Alizter:
Currently #1254 has some of these functions too. Your `Fin_left` is `fin_zero` and your `Fin_inc_r` is `fin_finS_inject` whose name may be subject to change.

I'm not so familiar with Sperner's lemma however. Is it saying that if I have a finite sequence `f : Fin n.+2 -> DHProp` is a decidable family of props, and I know it starts as false then becomes true, then I know that at some point in the middle it must switch values?

So it's kind of like an intermediate value theorem for finite sequences of decidable props? It might be worth adding some comments explaining this.

Other than that it looks good to me. Thanks for the additions!

theories/Spaces/Finite.v
```diff
@@ -27,6 +27,27 @@ Fixpoint Fin (n : nat) : Type
        | S n => Fin n + Unit
      end.
 
+Fixpoint Fin_left {n : nat} : Fin (S n)
+  := match n with
+     | 0 => inr tt
+     | S n => inl Fin_left
+     end.
+
+Definition Fin_right {n : nat} : Fin (S n) := inr tt.
```
Alizter: I would call this `fin_last` or something.

theories/Spaces/Finite.v
```diff
@@ -27,6 +27,27 @@ Fixpoint Fin (n : nat) : Type
...
+
+Definition Fin_inc_l {n : nat} (k : Fin n) : Fin (S n) := inl k.
```
Alizter: I would call this `fin_incl`.

theories/Spaces/Finite.v
abooij: Sure! Then I will correspondingly rename `Fin_left` into `fin_first`.

Alizter: In #1254 I have given `Fin_left` the name `fin_zero`.

theories/Spaces/Finite.v
abooij: The justification for the `_l` subscript is that there are (at least) two ways to include `Fin n` into `Fin (S n)` (except when `n=0`, in which case there is only one inclusion). `Fin_inc_l` is the inclusion "on the left".

But maybe you were just making a remark about capitalization, which I would be happy to adjust.

abooij:
> Currently #1254 has some of these functions too. Your `Fin_left` is `fin_zero` and your `Fin_inc_r` is `fsucc`.

Ahh, what a coincidence, I was not aware of that PR. Thanks for cross-checking. I'd be more than happy to rebase this PR on top of that after it is merged.

> I'm not so familiar with Sperner's lemma however. Is it saying that if I have a finite sequence `f : Fin n.+2 -> DHProp` is a decidable family of props, and I know it starts as false then becomes true, then I know that at some point in the middle it must switch values?

Exactly, except I implemented the version that says if it _starts_ with true and then becomes false, then at some point it switches from true to false. (If this is sounds less logical, I can swap it.)
 
> So it's kind of like an intermediate value theorem for finite sequences of decidable props? It might be worth adding some comments explaining this.

Will do!

Alizter:
> Ahh, what a coincidence, I was not aware of that PR. Thanks for cross-checking. I'd be more than happy to rebase this PR on top of that after it is merged.

It should be merged soon.

> Exactly, except I implemented the version that says if it starts with true and then becomes false, then at some point it switches from true to false. (If this is sounds less logical, I can swap it.)

I have no preference. Are you planning to use this for something?

abooij:
> > Exactly, except I implemented the version that says if it starts with true and then becomes false, then at some point it switches from true to false. (If this is sounds less logical, I can swap it.)
> 
> I have no preference. Are you planning to use this for something?

Yes, I have an application in mind, in which indeed the values start with true and end with false. But it is not a "canonical" application of Sperner's lemma, so it is no good justification for which variant is better.

Alizter:
>Yes, I have an application in mind, in which indeed the values start with true and end with false. But it is not a "canonical" application of Sperner's lemma, so it is no good justification for which variant is better.

I think its fine to keep it as it is. It would probably be a good idea to have a comment explaining where it starts from too.

Thanks again for these additions!

Alizter:
@abooij Are you able to rebase this?

abooij:
I rebased this on top of master and implemented @Alizter's suggestions.

theories/Spaces/Finite.v
```diff
@@ -1109,3 +1115,22 @@ Ltac FinIndOn X := repeat
 
 (** This tactic can be used to generate n cases from a goal like forall (x : Fin n), _ *)
 Ltac FinInd := let X := fresh "X" in intro X; FinIndOn X.
+
+Section Sperner.
+
```
Alizter: I would write something like the following above here:
```coq
(** Sperner's lemma (the 1-dimensional version) says that given any finite sequence of decidable hProps, where the sequence starts with true and ends with false, then there must exist a point in between such that the value of the sequence changes. This is like a discrete intermediate value theorem. *)
```

abooij: Ah, thanks for reminding me. I added a comment.

theories/Spaces/Finite.v
```diff
@@ -1108,3 +1114,27 @@ Ltac FinIndOn X := repeat
 
 (** This tactic can be used to generate n cases from a goal like forall (x : Fin n), _ *)
 Ltac FinInd := let X := fresh "X" in intro X; FinIndOn X.
+
+(** The 1-dimensional version of Sperner's lemma says that given any
+finite sequence of decidable hProps, where the sequence starts with
+true and ends with false, we can find a point in the sequence where
+the sequence changes from true to false. This is like a discrete
+intermediate value theorem. *)
+Section Sperner.
```
mikeshulman: Is there a reason for putting it in a section?  I generally only use sections when I want to put common variables in a `Context` for more than one result.

Alizter: I thought about this as well, but sections also serve a purpose to the reader. At the moment, Sperner's lemma seems to come out of nowhere so maybe a section is ok here?

abooij: It was indeed just an organizational section. I'd be happy to remove the `Section`.

theories/Spaces/Finite.v
```diff
@@ -1108,3 +1114,27 @@ Ltac FinIndOn X := repeat
 
 (** This tactic can be used to generate n cases from a goal like forall (x : Fin n), _ *)
 Ltac FinInd := let X := fresh "X" in intro X; FinIndOn X.
+
+(** The 1-dimensional version of Sperner's lemma says that given any
```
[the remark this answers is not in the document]
abooij: Okay, I'll remove the newlines. Trivial comments are good :-)

PR #1262, opened 2020-02-20
Allow compilation in a multiple git working trees

abooij:
The `git worktree` command allows to have several working trees for a
single repository.  The additional working trees then get a _file_
`.git` containing the gitdir, rather than a full `.git` directory.  In
order to support this, the build files should additionally allow
`.git` to be a file rather than a directory.

### DIFF
--- a/etc/install_coq.sh
+++ b/etc/install_coq.sh
@@ -11,9 +11,9 @@ pushd "$DIR" 1>/dev/null || exit 1
 ROOT_DIR="$(git rev-parse --show-toplevel)"
 pushd "$ROOT_DIR" 1>/dev/null || exit 1
 
-if test ! -d .git
+if test ! ( -d .git -o -f .git )
 then
-    echo 'Error: .git directory does not exist.'
+    echo 'Error: we do not seem to be in a git working tree.'
     echo 'This script only works on a git clone of the HoTT repository.'
     exit 1
 fi
--- a/etc/install_coq.sh
+++ b/etc/install_coq.sh
@@ -11,7 +11,7 @@ pushd "$DIR" 1>/dev/null || exit 1
 ROOT_DIR="$(git rev-parse --show-toplevel)"
 pushd "$ROOT_DIR" 1>/dev/null || exit 1
 
-if test ! ( -d .git -o -f .git )
+if ! ( [ -d .git ] || [ -f .git ] )
 then
     echo 'Error: we do not seem to be in a git working tree.'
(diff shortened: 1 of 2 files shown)


etc/install_coq.sh
```diff
@@ -11,9 +11,9 @@ pushd "$DIR" 1>/dev/null || exit 1
 ROOT_DIR="$(git rev-parse --show-toplevel)"
 pushd "$ROOT_DIR" 1>/dev/null || exit 1
 
-if test ! -d .git
+if test ! ( -d .git -o -f .git )
```
SkySkimmer: Don't use `-o` and `test !`, use the shell's `||` and `!` instead, so
```suggestion
if ! ( [ -d .git ] || [ -f .git ] )
```

SkySkimmer: (or just use `-e`)

jdchristensen:
Can you explain how to use the worktree set-up?  How best to avoid rebuilding coq in the submodule for each worktree?

jdchristensen:
To answer my own question, this seems to work:
```
cd path/to/clone/of/HoTT
git worktree ../HoTT-branchname [commitish]    # various options, see man git-worktree
cd ../HoTT-branchname
./autogen.sh
./configure COQBIN="../HoTT/coq-HoTT/bin"
make -j4
```

mikeshulman:
Maybe we could record that method somewhere?  Such as the increasingly-misnamed `STYLE.md`?

jdchristensen:
To avoid having to clone coq, this seems to work:
```
cd path/to/clone/of/HoTT
git worktree add ../HoTT-branchname [commitish]    (various options, see man git-worktree)
cd ../HoTT-branchname
git submodule sync etc/coq-scripts
git submodule update etc/coq-scripts
./autogen.sh -skip-modules
./configure COQBIN="../HoTT/coq-HoTT/bin"
make -j4
```

jdchristensen:
Let's see if other people improve on my suggestion, and then record it somewhere.

abooij:
For what it's worth, there is a justification for compiling Coq separately for a different branch: that branch may be based on an older version of the HoTT library and therefore require an older version of Coq.

jdchristensen:
COQBIN needs to be an absolute path, so I have updated the two recipes above.

jdchristensen:
There was a typo in the steps above:  I changed "-skip-modules" to "-skip-submodules".

PR #1264, opened 2020-02-20
O_inverts_functor_coeq without funext, closes #1259

### LINKED ISSUE #1259: Prove O_inverts_functor_coeq with Funext-free Yoneda
I don't have time to do this right now, but here's an approach that I think will work:

1. Improve `WildCat/Yoneda` to talk about 0-groupoid-valued functors (and probably thereby remove the notion of "strong" 1-category.
1. Prove a version of `equiv_functor_sigma` for equivalences of 0-groupoids and displayed categories.
1. Prove a version of the universal property for coequalizers using 0-groupoids (see `Homotopy/Suspension` for a similar universal property for suspensions).
1. Prove that this universal property is natural with respect to `functor_coeq` (again, see `Homotopy/Suspension` for the idea)
1. Repeat the current proof using equivalences of 0-groupoids instead of equivalences of types.

mikeshulman:
This is on top of #1246; review the last two commits.

I found a way to do it directly with extendability, analogously to the proof for sums, without using wild categories at all.  It's a neat trick that I'm surprised never occurred to me before: just as we can make _sections_ commute definitionally by using dependent types, i.e. fibrations, we can make _extensions_ commute definitionally by using cofibrations.  Book HoTT doesn't have a way to talk internally about cofibrations the way it can about fibrations, but we can still replace any map by one that is a cofibration, namely its mapping cylinder, and make use of the definitional behavior of the latter.  This gives a way to make a bunch of squares commute strictly, and with a bit of cubical technology added the path algebra becomes tractable.

The resulting proof of `O_inverts_functor_coeq` is, of course, much shorter, since all the work is happening in `extension_functor_coeq`.  I'm not sure whether this way of factoring the proof is easier or harder to understand overall, given the complication in the latter.  Overall the timing change seems to be negligible.  But in addition to avoiding funext (which is cute but not really dispositive) the extendability lemmas may have other applications (e.g. to pushouts or suspensions).

I included one little application that I just noticed (although I expect it could easily be proven in the funext style too): when applying `O_inverts_functor_coeq` to the universe `L'` of `L`-separated types, instead of assuming `h` (the map between the domains of the parallel pairs being coequalized) is an `L'`-equivalence, it suffices for it to be `L`-connected -- together with `k`, the map between codomains, being an `L'`-equivalence, this suffices for `functor_coeq` to be an `L'`-equivalence.  I wonder if this is good for anything (and did we know it already?).

### DIFF
--- a/theories/Colimits/MappingCylinder.v
+++ b/theories/Colimits/MappingCylinder.v
@@ -87,8 +87,7 @@ Section MappingCylinder.
         symmetry; apply cyglue.
       + intros b; reflexivity.
       + intros a; cbn.
-        apply dp_path_transport.
-        rewrite transport_paths_FFlr.
+        apply dp_paths_FFlr.
         rewrite Cyl_rec_beta_cyglue.
         apply concat_pV_p.
     - intros b; reflexivity.
--- a/theories/Colimits/MappingCylinder.v
+++ b/theories/Colimits/MappingCylinder.v
@@ -1,0 +1,208 @@
+(* -*- mode: coq; mode: visual-line -*- *)
+
+(** * Mapping Cylinders *)
+
+Require Import HoTT.Basics HoTT.Types Cubical.DPath Cubical.PathSquare.
+Require Import HIT.Coeq Colimits.Pushout.
+Local Open Scope path_scope.
+
+(** As in topology, the mapping cylinder of a function [f : A -> B] is a way to replace it with an equivalent cofibration (dually to how [hfiber] replaces it with an equivalent fibration).  We can't talk *internally* in type theory about cofibrations, but we can say metatheoretically what they are: functions with the isomorphism extension property.  So while we can't literally say "let [f] be a cofibration" we can do a mostly equivalent thing and say "let [f] be a map and consider its mapping cylinder".  Replacing a map by a cofibration can be useful because it allows us to make more equalities true definitionally. *)
+
+(** ** Definitions *)
+
+(** We define the mapping cylinder as the pushout of [f] and an identity map.  Peter Lumsdaine has given a definition of HIT mapping cylinders that are dependent on the codomain, so that the second factor is not just an equivalence but a trivial fibration.  However, at the moment we don't have a need for that.  *)
+
+Definition Cyl {A B : Type} (f : A -> B) : Type
+  := Pushout idmap f.
+
+Section MappingCylinder.
+  Context {A B : Type} {f : A -> B}.
+
+  Definition cyl (a : A) : Cyl f
+    := pushl a.
+  
+  Definition cyr (b : B) : Cyl f
+    := pushr b.
+
+  Definition cyglue (a : A)
+    : cyl a = cyr (f a)
+    := pglue a.
+
+  Section CylInd.
+    Context (P : Cyl f -> Type)
+            (cyla : forall a, P (cyl a))
+            (cylb : forall b, P (cyr b))
+            (cylg : forall a, DPath P (cyglue a) (cyla a) (cylb (f a))).
+
+    Definition Cyl_ind : forall c, P c.
+    Proof.
+      srapply Pushout_ind.
+      - apply cyla.
+      - apply cylb.
+      - intros a; apply dp_path_transport^-1, cylg.
+    Defined.
+
+    Definition Cyl_ind_beta_cyglue (a : A)
+      : dp_apD Cyl_ind (cyglue a) = cylg a.
+    Proof.
+      unfold Cyl_ind.
+      refine ((dp_path_transport_apD _ _)^ @ _).
+      apply moveR_equiv_M.
+      rapply Pushout_ind_beta_pglue.
+    Defined.
+
+  End CylInd.
+
+  Section CylRec.
+    Context {P : Type} (cyla : A -> P) (cylb : B -> P) (cylg : cyla == cylb o f).
+
+    Definition Cyl_rec : Cyl f -> P.
+    Proof.
+      srapply Pushout_rec.
+      - apply cyla.
+      - apply cylb.
+      - apply cylg.
+    Defined.
+
+    Definition Cyl_rec_beta_cyglue (a : A)
+      : ap Cyl_rec (cyglue a) = cylg a.
+    Proof.
+      rapply Pushout_rec_beta_pglue.
+    Defined.
+
+  End CylRec.
+
+  Definition pr_cyl : Cyl f <~> B.
+  Proof.
+    (** Rather than adjointifying, we give all parts of the equivalence explicitly, so we can be sure of retaining the computational behavior of [eissect] and [eisretr].  However, it's easier to prove [eisadj] on the other side, so we reverse the equivalence first. *)
+    symmetry.
+    srapply Build_Equiv.
+    1:apply cyr.
+    srapply Build_IsEquiv.
+    - srapply Cyl_rec.
+      + exact f.
+      + exact idmap.
+      + reflexivity.
+    - srapply Cyl_ind.
+      + intros a; cbn.
+        symmetry; apply cyglue.
+      + intros b; reflexivity.
+      + intros a; cbn.
+        apply dp_path_transport.
+        rewrite transport_paths_FFlr.
+        rewrite Cyl_rec_beta_cyglue.
+        apply concat_pV_p.
+    - intros b; reflexivity.
+    - intros b; reflexivity.
+  Defined.
+
+  Definition ap_pr_cyl_cyglue (a : A)
+    : ap pr_cyl (cyglue a) = 1
+    := Cyl_rec_beta_cyglue _ _ _ a.
+
+  (** The original map [f] factors definitionally through [Cyl f]. *)
+  Definition pr_cyl_cyl (a : A) : pr_cyl (cyl a) = f a
+    := 1.
+
+End MappingCylinder.
+
+(** Sometimes we have to specify the map explicitly. *)
+Definition cyl' {A B} (f : A -> B) : A -> Cyl f := cyl.
+Definition pr_cyl' {A B} (f : A -> B) : Cyl f -> B := pr_cyl.
+
+(** ** Functoriality *)
+
+Section FunctorCyl.
+  Context {A B A' B': Type} {f : A -> B} {f' : A' -> B'}
+          {ga : A -> A'} {gb : B -> B'}
+          (g : f' o ga == gb o f).
+
+  Definition functor_cyl : Cyl f -> Cyl f'.
+  Proof.
+    srapply Cyl_rec.
+    - exact (cyl o ga).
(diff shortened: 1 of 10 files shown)


mikeshulman:
Hmm, actually one could argue that there's not so much difference between fibrations and cofibrations even in Book HoTT.

- We have a notion of a "structured fibration over `A`", namely a type family `B : A -> Type`, which can be made into a map `pr1 : sig B -> A` that is a fibration.  But we don't have a way to talk about whether a _given_ map "is a fibration", only a way to produce maps that *are* fibrations out of "structured fibration data".

- Similarly, we can consider a "structured cofibration under `A`" to mean just a morphism `f : A -> B`, which can be made into a map `cyl : A -> Cyl f` that is a cofibration.  Thus, the fact that we don't have a way to talk about whether a _given_ map "is a cofibration" is really no different from the situation for fibrations.

- Given fibration data `B : A -> Type` and `D : C -> Type` with a map `h : A -> C`, we have a notion of "structured map of fibrations over `h`", namely a family of maps `forall a:A, B a -> D (h a)`, which induces a map `sig B -> sig D` that makes the square commute strictly (although we don't have a way to express internally that a given square "commutes strictly").  Similarly, given cofibration data `f : A -> B` and `g : C -> D` and a map `h : A -> C`, we have a notion of "structured map of cofibrations under `h`", namely a map `k : B -> D` and a homotopy filling the square, which induces a map `Cyl f -> Cyl g` making the square commute strictly.

The last case is somewhat different in that "maps of fibrations" compose strictly associatively, whereas "maps of cofibrations" don't.  But it does make me wonder whether cofibrations would be useful for other things.  For instance, what if we defined a "pointed type" to be determined (as it is now) by a type `A` and an element `a:A`, but considered its "underlying type" to be not `A` but `Cyl (unit_name a)`?  Then the inclusions of all basepoints would "be" cofibrations in a certain sense, which is something that classical algebraic topologists often assume, and we have to do a lot of work because we can't assume.  I expect that a trick like this would not actually buy us very much, since we'd still have to be manipulating the homotopies "behind the scenes"; but I wonder whether there might be some small things that it would buy us.

theories/Colimits/MappingCylinder.v
```diff
@@ -0,0 +1,208 @@
...
+
+Definition Cyl {A B : Type} (f : A -> B) : Type
+  := Pushout idmap f.
+
+Section MappingCylinder.
+  Context {A B : Type} {f : A -> B}.
+
+  Definition cyl (a : A) : Cyl f
+    := pushl a.
+
```
Alizter:
```suggestion

```

theories/Colimits/MappingCylinder.v
```diff
@@ -0,0 +1,208 @@
...
+      + exact f.
+      + exact idmap.
+      + reflexivity.
+    - srapply Cyl_ind.
+      + intros a; cbn.
+        symmetry; apply cyglue.
+      + intros b; reflexivity.
+      + intros a; cbn.
+        apply dp_path_transport.
+        rewrite transport_paths_FFlr.
```
Alizter: We have `dp_paths_FFlr` for doing exactly this.

mikeshulman: Of course, right.

theories/Extensions.v
```diff
@@ -433,7 +435,68 @@ Section Extensions.
...
+
+(** If a family is extendable along a cofibration (i.e. a mapping cylinder), it is extendable definitionally. *)
+Definition cyl_extension {A B} (f : A -> B) (C : Cyl f -> Type)
+           (g : forall a, C (cyl a))
+           (ext : ExtensionAlong cyl C g)
+  : ExtensionAlong cyl C g.
+Proof.
+  srefine (Cyl_ind C g (ext.1 o cyr) _ ; _); intros a.
+  + pose (p := dp_apD ext.1 (cyglue a)).
+    exact (transport (fun u => DPath C _ u _) (ext.2 a) p).
```
jdchristensen: Since I'm not familiar with the dependent path part of the code, I find the version below easier to read.  But the dependent path method matches what is used in nearby areas, so might be best.
```coq
  srefine (Pushout_ind C g (ext.1 o cyr) _ ; _); intros a.
  + refine (_ @ apD ext.1 (cyglue a)).
    exact (ap _ (ext.2 a)^).
```

theories/Extensions.v
```diff
@@ -464,3 +527,124 @@ Definition ooextendable_functor_sum
...
+  + (** Since we're using cofibrations, this holds definitionally. *)
+    intros a; reflexivity.
+  + (** And this one is much simpler than it would be otherwise. *)
+    intros b.
+    apply ds_dp.
+    rapply ds_G1.
+    refine (dp_apD_compose' _ _ (ap_cyl_cylcoeq_cglue p q b) _ @ _).
+    apply moveR_equiv_V.
+    rapply (Coeq_ind_dp_beta_cglue C').
+Defined.
```
jdchristensen: This is impressive!  I tried to see how things would go with a direct method.  It doesn't seem impossible, but it got too complicated for me.  I'll show what I tried here in case anyone else is interested in playing with it.  Not sure if dependent path techniques would be enough to overcome this, or if the mapping cylinder trick is the key.
```coq
Lemma transport_nameme {A B} {x y : A} {b : B} (P : B -> Type) (f : A -> B)
  (p : x = y) (q : f y = b) (r : forall a, P (f a))
  : transport P q (r y) = transport P (ap f p @ q) (r x).
Proof.
  destruct p, q; reflexivity.
Defined.

Definition extension_functor_coeq {B A f g B' A' f' g'}
           {h : B -> B'} {k : A -> A'}
           {p : k o f == f' o h} {q : k o g == g' o h}
           {C : Coeq f' g' -> Type}
           (ek : ExtendableAlong 1 k (C o coeq))
           (eh : forall (u v : forall x : B', C (coeq (g' x))),
               ExtendableAlong 1 h (fun x => u x = v x))
           (s : forall x, C (functor_coeq h k p q x))
  : ExtensionAlong (functor_coeq h k p q) C s.
Proof.
  pose (re := (fst ek (s o coeq)).2).
  set (r := (fst ek (s o coeq)).1) in *.
  srefine (_ ; _).
  - srapply (Coeq_ind C).
    + exact r.
    + refine (fst (eh _ _) _).1.
      intro b.
      refine (_ @ _).
      1: rapply transport_nameme; exact (p b).
      refine (_ @ apD r (q b)).
      refine (_ @ _).
      2: symmetry; rapply (transport_compose C).
      refine (ap _ (re (f b)) @ _ @ (ap _ (re (g b)))^).
      refine (_ @ _).
      2: symmetry; rapply (transport_nameme C _ (cglue b) _ s).
      apply (ap (fun w => transport C w (s (coeq (f b))))).
      refine (_ @ whiskerR _ _).
      2: symmetry; rapply functor_coeq_beta_cglue.
      refine (_ @ _).
      2: rapply (ap (fun w => _ @ w @ _) (ap_V _ _)^).
      symmetry; apply concat_pV_p.
  - srapply Coeq_ind.
    + intro a; cbn.
      apply re.
    + intro b; cbn.
      rewrite transport_paths_FlFr_D.
      (* Goal is 51 lines long; not sure what to do next. *)
```

theories/Extensions.v
```diff
@@ -464,3 +527,124 @@ Definition ooextendable_functor_sum
...
+           {h : B -> B'} {k : A -> A'}
+           {p : k o f == f' o h} {q : k o g == g' o h}
+           {C : Coeq f' g' -> Type}
+           (ek : ooExtendableAlong k (C o coeq))
+           (eh : forall (u v : forall x : B', C (coeq (g' x))),
+               ooExtendableAlong h (fun x => u x = v x))
+  : ooExtendableAlong (functor_coeq h k p q) C
+  := fun n => extendable_functor_coeq n (ek n) (fun u v => eh u v n).
+
+(** In addition, to extend at level [n] into path-types of [C], it suffices to extend at level [n.+1] into [C] itself. *)
```
jdchristensen: I was confused by this comment at first.  I thought it was describing the statement of the next result, but it's really describing the proof.  Maybe say something like "Since extending at level [n.+1] into [C] implies extending at level [n] into path-types of [C], we get the following corollary."

jdchristensen:
The mapping cylinder trick is very clever!  It's too bad that the main proof ends up being very hard to understand, especially compared to the Funext proof.  For the Funext proof, you just test things by mapping out and use the universal property.  Is there any chance some more basic ingredients involving ooextendability can be proved and then pieced together in a simpler way?  The main hard argument is in a sense just showing that coequalizers respect equivalences, and if it requires this much technology, how will we handle more complicated colimits?  Maybe the universal property should be stated in an ooextendable way, and that can be used to deduce that coequalizers respect ooextendability?

theories/Extensions.v
mikeshulman: Yeah, I tried this at first too and gave up.  With dependent paths it's a bit more manageable, but I still found it nearly impossible to complete until I had the mapping cylinders idea.

jdchristensen:
`OO_inverts_functor_coeq` is also nice!  There are a lot of variations like this.  When you take [O'] to be [Sep O] in this, you get a slightly better result than we knew for [Sep O].  There are other statements that I also know how to prove only for RSUs of the form [Sep O].

mikeshulman:
Well, the colimit of a diagram on any graph can be defined using sigmas and coequalizers, so we should be able to deduce things about them easily.  Colimits over more general things than graphs are hard to deal with in general, but we haven't really needed them yet either; in practice it seems we can construct most things we need with coequalizers and things defined from them (perhaps by iteration, e.g. iterated suspension).

> Maybe the universal property should be stated in an ooextendable way, and that can be used to deduce that coequalizers respect ooextendability?

I think this is what I had in mind with the sketch in #1259.  The extendable version of the universal property would be an analogue of (and, in fact, should imply) `equiv0gpd_Susp_ind_inv`.  Until the wildcat library contains oo-categories, we won't be able to state the full oo-extendable property this way -- with wild 0-groupoids we get only 2-extendable -- but this is enough for most purposes.

I agree that in some sense that argument would be the most pleasing, although the version for suspensions still got kind of hairy having to transport back and forth across naturality.  Maybe we should keep #1259 open even after this is merged in case someone wants to try doing it that way later.

theories/Extensions.v
mikeshulman: Yes, that's better; thanks.

theories/Extensions.v
mikeshulman: I stuck with dependent paths, but I renamed `Cyl_ind` to `Cyl_ind_dp` since it uses dependent paths (to match the naming of `Coeq_ind_dp`), and I replaced the `transport` with an operation of concatenating dependent paths with non-dependent ones (which I've needed elsewhere too).

mikeshulman:
I also had to change the level of the dependent-path concatenation notations to make them work.

PR #1267, opened 2020-02-22
Applications of extensions to connected maps and preservation of colimits

mikeshulman:
One commit on top of #1264.  Using extendability more systematically simplifies the proofs that reflectors preserve products and pushouts and invert connected maps, and implies that colimits preserve connected maps (with enhancements for separated pairs of subuniverses).

It would also simplify the proof of `equiv_O_sigma_O`, but unfortunately the simpler proof of that has worse computational behavior, and we rely on its computational behavior in several places.  We could change those places to reason differently, but it's probably better to keep the good computational behavior and just note that the inverse is definitionally the same as what the simple proof would give.  To emphasize this, I changed the inverse to be syntactically the same as well, which required some trivial changes in a couple of places downstream.

In making one of those trivial changes to `BlakersMassey`, I managed to speed it up by ~30%, probably due to replacing some `rewrites` and a `cbn` with `refine`s.

### DIFF
--- a/theories/Extensions.v
+++ b/theories/Extensions.v
@@ -589,7 +589,7 @@ Definition HomotopyExtensionAlong {A B} {Q : B -> Type}
   := { q : forall (b:B) (v:Q b), C (b;v) & forall a v, q (f a) v = p a v }.
 
 Fixpoint HomotopyExtendableAlong (n : nat)
-         {A B} {Q : B -> Type} (f : A -> B) (C : sig Q -> Type)
+         {A B} {Q : B -> Type} (f : A -> B) (C : sig Q -> Type) : Type
   := match n with
      | 0 => Unit
      | S n => ((forall (p : forall (a:A) (v:Q (f a)), C (f a;v)),
--- a/theories/Extensions.v
+++ b/theories/Extensions.v
@@ -496,7 +496,176 @@ Definition cyl_ooextendable'
   := fun n => cyl_extendable' n f C (ext n).
 
 
+(** ** Extendability along [functor_prod] *)
+
+Definition extension_functor_prod
+           {A B A' B'} (f : A -> A') (g : B -> B')
+           (P : A' * B' -> Type)
+           (ef : forall b', ExtendableAlong 1 f (fun a' => P (a',b')))
+           (eg : forall a', ExtendableAlong 1 g (fun b' => P (a',b')))
+           (s : forall z, P (functor_prod f g z))
+  : ExtensionAlong (functor_prod f g) P s.
+Proof.
+  srefine (_;_).
+  - intros [a' b']; revert b'.
+    refine ((fst (eg a') _).1).
+    intros b; revert a'.
+    refine ((fst (ef (g b)) _).1).
+    intros a.
+    exact (s (a,b)).
+  - intros [a b]; cbn.
+    refine ((fst (eg (f a)) _).2 b @ _).
+    exact ((fst (ef (g b)) _).2 a).
+Defined.
+
+Definition extendable_functor_prod (n : nat)
+           {A B A' B'} (f : A -> A') (g : B -> B')
+           (P : A' * B' -> Type)
+           (ef : forall b', ExtendableAlong n f (fun a' => P (a',b')))
+           (eg : forall a', ExtendableAlong n g (fun b' => P (a',b')))
+  : ExtendableAlong n (functor_prod f g) P.
+Proof.
+  revert P ef eg; simple_induction n n IH; intros P ef eg; [ exact tt | split ].
+  - apply extension_functor_prod.
+    + intros b'; exact (fst (ef b'), fun _ _ => tt).
+    + intros a'; exact (fst (eg a'), fun _ _ => tt).
+  - intros h k; apply IH.
+    + intros b'; apply (snd (ef b')).
+    + intros a'; apply (snd (eg a')).
+Defined.
+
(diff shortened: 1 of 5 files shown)


theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -811,20 +809,31 @@ Section Reflective_Subuniverse.
...
 
-    (** Theorem 7.3.9: The reflector [O] can be discarded inside a reflected sum. *)
+    (** [functor_sigma] over [idmap] preserves [O]-equivalences. *)
+    Definition O_inverts_functor_sigma_id {A} {P Q : A -> Type}
+           (g : forall a, P a -> Q a) `{forall a, O_inverts (g a)}
+      : O_inverts (functor_sigma idmap g).
+    Proof.
+      apply O_inverts_from_extendable; intros Z Z_inO.
+      apply ooextendable_functor_sigma_id; intros a.
+      nrapply ooextendable_O_inverts; exact _.
```
mikeshulman: Curiously, `nrapply ooextendable_O_inverts; exact _` is noticeably faster than `rapply ooextendable_O_inverts` here.  Perhaps the `nrapply` does some elaboration that makes the later typeclass resolution quicker, whereas the `rapply` does things in another order?

mikeshulman:
Any idea what's going on with CI?  This branch builds fine for me locally.

jdchristensen:
For some reason there's an error in Extensions.v.  It builds fine for me with the submodule Coq.  The failing build uses the system Coq.  Maybe that's the difference?
```
HOQC theories/Extensions
File "./theories/Extensions.v", line 603, characters 41-42:
Error:
In environment
A : Type
B : Type
Q : B -> Type
f : A -> B
C : {x : _ & Q x} -> Type
n : nat
The term "f" has type "A -> B" while it is expected to have type 
"?A -> ?B" (unable to find a well-typed instantiation for 
"?A": cannot ensure that "Type" is a subtype of "Type0").

make: *** [theories/Extensions.vo] Error 1
```

theories/Modalities/ReflectiveSubuniverse.v
jdchristensen: I've noticed the same thing occasionally, and I don't know why it happens.  You could try "Typeclasses eauto := debug" and see if the number of lines produced in the two cases is substantially different.

Alizter:
So the other jobs in the CI are cancelled once one of them has an error. @JasonGross do you know if we can change this behaviour?

jdchristensen:
Coq version info:
```
checking for coqtop... /usr/bin/coqtop
checking coqtop version... 8.12+alpha
checking Coq library path... /usr/lib/coq
configure: COQBIN is /usr/bin
checking for coqc... /usr/bin/coqc
checking coqc version... 8.12+alpha
```

Alizter:
So this was built with the latest version of Coq 8.12 on the master branch. For whatever reason there is a regression in behaviour. I suspect that there is a universe issue since coq is complaining about Type0.

Alizter:
I am currently building coq 8.12 so I can poke around and find out what is going on.

theories/Extensions.v
```diff
@@ -433,7 +435,237 @@ Section Extensions.
...
+
+(** Unfortunately, the technology of [ExtensionAlong] seems to be insufficient to state a general, funext-free version of [extension_functor_sigma] with a nonidentity map on the bases; the hypothesis on the fiberwise map would have to be the existence of an extension in a function-type "up to pointwise equality".  With wild oo-groupoids we could probably manage it.  For now, we say something a bit hacky. *)
+
+Definition HomotopyExtensionAlong {A B} {Q : B -> Type}
+           (f : A -> B) (C : sig Q -> Type)
+           (p : forall (a:A) (v:Q (f a)), C (f a;v))
+  := { q : forall (b:B) (v:Q b), C (b;v) & forall a v, q (f a) v = p a v }.
+
+Fixpoint HomotopyExtendableAlong (n : nat)
+         {A B} {Q : B -> Type} (f : A -> B) (C : sig Q -> Type)
```
Alizter:
```suggestion
         {A B} {Q : B -> Type} (f : A -> B) (C : sig Q -> Type) : Type
```

Alizter:
The fixpoint without the type specified was being put in `Type0` for whatever reason. When I give it the type `Type` coq correctly does the polymorphic thing.

JasonGross:
> So the other jobs in the CI are cancelled once one of them has an error. @JasonGross do you know if we can change this behaviour?

I do not know.  (We could aways resurrect Travis though)

Alizter:
@JasonGross Actually I found a way: see #1268

Alizter:
@JasonGross I would be in favour of resurrecting Travis while we are transitioning with Github actions. There are a still a few quirks to work out with actions and I would hate to get in the way of progress in the library.

Alizter:
@mikeshulman Can you rebase?

mikeshulman:
I don't think we need to resurrect travis.  CI is not crucial to development, we can always ignore it.

mikeshulman:
rebased

mikeshulman:
In case anyone is interested, the extra complication in the simpler proof consists partly of the same "extracting two different 2-extendabilities from an oo-extendability" issue that arose elsewhere, but also a couple of transports in constant type families.  I expect the latter are probably impossible to eliminate.

mikeshulman:
Oops, forgot to pull before I `push -f`'d.

PR #1271, opened 2020-02-23
Shorter and cleaner proof of hprop_isequiv

mikeshulman:
Also slightly faster.

### DIFF
--- a/theories/Types/Equiv.v
+++ b/theories/Types/Equiv.v
@@ -14,48 +14,22 @@ Section AssumeFunext.
   Global Instance hprop_isequiv {A B} `(f : A -> B)
   : IsHProp (IsEquiv f).
   Proof.
+    (** We will show that assuming [f] is an equivalence, [IsEquiv f] decomposes into a sigma of two contractible types. *)
     apply hprop_inhabited_contr; intros feq.
-    (* We will show that if [IsEquiv] is inhabited, then it is contractible, because it is equivalent to a sigma of a pointed path-space over a pointed path-space, both of which are contractible. *)
-    refine (contr_equiv' { g : B -> A & g = f^-1 } _).
-    equiv_via ({ g:B->A & { r:g=f^-1 & { s:g=f^-1 & r=s }}}); apply equiv_inverse.
-    1:exact (equiv_functor_sigma_id (fun _ => equiv_sigma_contr _ )).
-    (* First we apply [issig], peel off the first component, and convert to pointwise paths. *)
-    refine (_ oE (issig_isequiv f)^-1).
-    refine (equiv_functor_sigma_id _); intros g; simpl.
-    equiv_via ({ r : g == f^-1 & { s : g == f^-1 & r == s }}).
-    (* Now the idea is that if [f] is an equivalence, then [g f == 1] and [f g == 1] are both equivalent to [g == f^-1]. *)
-    { refine (equiv_functor_sigma'
-                (equiv_functor_forall idmap (fun b p => (ap f)^-1 (p @ (eisretr f b)^)))
-                (fun r => equiv_functor_sigma'
-                            (equiv_functor_forall f (fun a p => p @ (eissect f a)))^-1 _));
-      intros s; simpl.
-      (* What remains is to show that under these equivalences, the remaining datum [eisadj] reduces simply to [r == s].  Pleasingly, Coq can compute for us exactly what this means. *)
-      apply equiv_inverse;
-        refine (equiv_functor_forall' (Build_Equiv _ _ f _) _);
-        intros a; simpl; unfold functor_forall.
-      rewrite transport_paths_FlFr.
-      (* At this point it's just naturality wrangling, potentially automatable.  It's a little unusual because what we have to prove is not just the existence of some path, but that one path-type is equivalent to another one, but we can mostly still use [rewrite]. *)
-      Open Scope long_path_scope.
-      rewrite ap_pp, !concat_p_pp, eisadj, <- !ap_V, <- !ap_compose.
-      rewrite (concat_pA1_p (eissect f) (eissect f a)^).
-      rewrite (concat_A1p s (eissect f a)^).
-      rewrite (concat_pp_A1 (fun x => (eissect f x)^) (eissect f a)).
-      (* Here instead of [whiskerR] we have to be a bit fancier. *)
-      refine (_ oE (equiv_ap (equiv_concat_r (eissect f a)^ _) _ _)^-1).
-      rewrite concat_pV_p.
-      refine (_ oE equiv_ap (ap f) _ _).
-      (* Now we can get rid of the [<~>] and reduce the question to constructing some path. *)
-      apply equiv_concat_l.
-      rewrite !ap_pp, !ap_V, <- !eisadj, <- ap_compose.
-      rewrite_moveL_Vp_p.
-      symmetry; exact (concat_A1p (eisretr f) (r (f a))).
-      Close Scope long_path_scope. }
-    (* The leftover goal is just nested applications of funext. *)
-    { refine (equiv_functor_sigma' (equiv_path_arrow g f^-1)
-                                   (fun r => equiv_functor_sigma' (equiv_path_arrow g f^-1) _));
-      intros s; simpl.
-      refine (_ oE equiv_path_forall r s).
-      exact (equiv_ap (path_forall g f^-1) r s). }
+    refine (contr_equiv' _ (issig_isequiv f oE (equiv_sigma_assoc' _ _)^-1)).
+    srefine (contr_equiv' _ (equiv_contr_sigma' _ (f^-1 ; eisretr f))^-1); only 2: exact _.
+    (** Each of these types is equivalent to a based path space. *)
+    - refine (contr_equiv' { g : B -> A & g = f^-1 } _).
+      apply equiv_functor_sigma_id; intros g.
+      refine (_ oE equiv_ap10 _ _).
+      apply equiv_functor_forall_id; intros b.
+      apply equiv_moveR_equiv_M.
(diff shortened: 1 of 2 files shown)


theories/Types/Equiv.v
```diff
@@ -14,48 +14,22 @@ Section AssumeFunext.
...
-      symmetry; exact (concat_A1p (eisretr f) (r (f a))).
-      Close Scope long_path_scope. }
-    (* The leftover goal is just nested applications of funext. *)
-    { refine (equiv_functor_sigma' (equiv_path_arrow g f^-1)
-                                   (fun r => equiv_functor_sigma' (equiv_path_arrow g f^-1) _));
-      intros s; simpl.
-      refine (_ oE equiv_path_forall r s).
-      exact (equiv_ap (path_forall g f^-1) r s). }
+    refine (contr_equiv' _ (issig_isequiv f oE (equiv_sigma_assoc' _ _)^-1)).
+    srefine (contr_equiv' _ (equiv_contr_sigma' _ (f^-1 ; eisretr f))^-1); only 2: exact _.
```
jdchristensen: These and other `refine`s are currently fast, but since the unsolved goals are Contr goals and we have a lot of Contr typeclass instances, they could become slow if we add more typeclass instances or if Coq's algorithm changes slightly.  So I'd prefer adding `n` to whichever ones accept it.

Also, you can switch from `equiv_contr_sigma'` to `equiv_contr_sigma` if the `Qed` at the end of `moveR_equiv_M` is changed to `Defined`, with one change below that I'll mark.

theories/Types/Equiv.v
```diff
@@ -14,48 +14,22 @@ Section AssumeFunext.
...
+    - refine (contr_equiv' { g : B -> A & g = f^-1 } _).
+      apply equiv_functor_sigma_id; intros g.
+      refine (_ oE equiv_ap10 _ _).
+      apply equiv_functor_forall_id; intros b.
+      apply equiv_moveR_equiv_M.
+    - refine (contr_equiv' { s : f^-1 o f == idmap & eissect f = s } _).
+      apply equiv_functor_sigma_id; intros s; cbn.
+      refine (_ oE equiv_apD10 _ _ _).
+      apply equiv_functor_forall_id; intros a.
+      refine (equiv_concat_l (eisadj f a) _ oE _).
```
jdchristensen: If you use equiv_contr_sigma above, then this becomes:
```suggestion
      refine (equiv_concat_l ((concat_1p _) @ (eisadj f a)) _ oE _).
```

theories/Types/Equiv.v
```diff
@@ -14,48 +14,22 @@ Section AssumeFunext.
...
+      rapply equiv_ap.
   Qed.
```
jdchristensen:
```suggestion
  Defined.
```

theories/Types/Equiv.v
jdchristensen:
```suggestion
    srefine (contr_equiv' _ (equiv_contr_sigma _)^-1).
```

theories/Types/Sigma.v
```diff
@@ -490,6 +490,11 @@ Proof.
     apply transport_pV.
 Defined.
 
+Definition equiv_contr_sigma' {A : Type} (P : A -> Type) `{Contr A} (a : A)
+  : { x : A & P x } <~> P a
+  := equiv_transport P _ _ (contr a) oE equiv_contr_sigma P.
```
jdchristensen: Instead of adapting each lemma to take a point different from the center of contraction, what about a lemma:
```coq
Definition contr_change_center (A : Type) (C : Contr A) (a : A) : Contr A.
```
with the one line proof that changes the center to a.  This can then be chained with any other lemma that uses the center of contraction.

theories/Types/Equiv.v
Alizter: Didn't this cause problems in Spaces/Card.v?

theories/Types/Equiv.v
```diff
@@ -14,48 +14,22 @@ Section AssumeFunext.
...
-    (* The leftover goal is just nested applications of funext. *)
-    { refine (equiv_functor_sigma' (equiv_path_arrow g f^-1)
-                                   (fun r => equiv_functor_sigma' (equiv_path_arrow g f^-1) _));
-      intros s; simpl.
-      refine (_ oE equiv_path_forall r s).
-      exact (equiv_ap (path_forall g f^-1) r s). }
+    refine (contr_equiv' _ (issig_isequiv f oE (equiv_sigma_assoc' _ _)^-1)).
+    srefine (contr_equiv' _ (equiv_contr_sigma' _ (f^-1 ; eisretr f))^-1); only 2: exact _.
+    (** Each of these types is equivalent to a based path space. *)
+    - refine (contr_equiv' { g : B -> A & g = f^-1 } _).
```
jdchristensen: Maybe we should have a global instance somewhere that `{ g : B -> A & g == k }` is contractible, under Funext?

theories/Types/Equiv.v
jdchristensen: This line is the end of the new proof, which no longer uses rewrites.  Not sure if `moveR_equiv_M` needs to be Qed for some reason.

Alizter: If these changes are added we will see if it effects performance I suppose.

mikeshulman: I tried this, but it caused a problem in `Descent.v`.  I didn't investigate why.

theories/Types/Sigma.v
mikeshulman: i can see how that would work in this case where both contractibility statements are being proven by hand.  But what about when the `Contr` instance is being found by typeclass inference?

theories/Types/Equiv.v
mikeshulman: We do; it's `contr_basedhomotopy` in `FunextVarieties`... which depends on this file, so it can't be used here.

Alizter: We could bring it forward if it is really needed.

theories/Types/Sigma.v
mikeshulman: I guess one thing we could do is pay more attention to the center when we define `Contr` instances.  If we have a clean proof of a `Contr` instance that ends up with a center that's not as convenient as it might otherwise be, we could incorporate `contr_change_center` into it.  If we did that everywhere (and all `Contr` instances were transparent), it might be sufficient.

theories/Types/Equiv.v
mikeshulman: It would only shorten this proof by one line.

theories/Types/Sigma.v
mikeshulman: Although since `Contr` is a notation for `IsTrunc (-2)`, sometimes `Contr` instances may be found by applying general `IsTrunc` lemmas, in which we wouldn't have the opportunity to apply `contr_change_center`.

theories/Types/Equiv.v
mikeshulman: Although that instance is so generally useful that it is kind of annoying to have it hidden away in `FunextVarieties`; it would be nice to have it somewhere in `Basics` or `Types`.

Alizter: I've always thought EquivlaenceVarieties should be part of Types to be honest.

mikeshulman: All three `Varieties` files are a tricky problem, as on the one hand they contain a few useful lemmas -- e.g. just as `FunextVarieties` contains `contr_basedhomotopy`, `EquivalenceVarieties` contains `fcontr_isequiv` and `isequiv_fcontr`.   But on the other hand, a lot of their time is spent on "dead-end" results that are mainly interesting theoretically and aren't ever going to get used for anything in the rest of the library.  I guess one solution would be to move the useful lemmas into other files so that the Varieties files themselves never need to be imported elsewhere.

Alizter: Yes, if things are being used from EquivlaenceVarieties then clearly they are not so dead end.

theories/Types/Sigma.v
jdchristensen: If I want to apply lemmaA that needs a proof of contractibility, I can do
```
rapply (lemmaA (contr_change_center a))
```
so that typeclass inference can fill in the contractibility argument for contr_change_center.  (Here contr_change_center has {A: Type} and `{IsContr A}.)  Or:
```
nrapply lemmaA. 1: rapply (contr_change_center a).
```

jdchristensen:
Feel free to merge after taking into account whichever of the suggestions you want.  They are just ideas.

theories/Types/Equiv.v
jdchristensen: @mikeshulman Do you mean the Qed here or the one in moveR_equiv_M?  It doesn't need to be part of this PR, but moveR_equiv_M has a simple proof that we should be able to make transparent, and then there's no need to change the center of contraction here.

theories/Types/Equiv.v
mikeshulman: UnivalenceImpliesFunext, TruncImpliesFunext, and IntervalImpliesFunext are also dead-end metatheory.  (Although EquivalenceVarieties is not exactly metatheory -- there may be some other more important stuff there.)

theories/Types/Equiv.v
mikeshulman: I mean the one here.

mikeshulman: `moveR_equiv_M` has always been transparent, hasn't it?

jdchristensen: Sorry, I meant the Qed at the end of contr_equiv.  Now that it is Defined, you don't need to change the center in this proof.  And then you just have to add the concat_1p shown above.

mikeshulman: I prefer changing the center and leaving out the concat_1p.

PR #1273, opened 2020-02-23
Move contr_basedhomotopy out of FunextVarieties, reducing the need to import it

mikeshulman:
On top of #1271.  The only remaining imports of `FunextVarieties` are in the other metatheory files `TruncImpliesFunext` and `IntervalImpliesFunext`, which need `Funext_type`.

### DIFF
--- a/theories/HoTT.v
+++ b/theories/HoTT.v
@@ -17,7 +17,7 @@ Require Export HoTT.HSet.
 Require Export HoTT.EquivGroupoids.
 Require Export HoTT.EquivalenceVarieties.
 
-Require Export HoTT.FunextVarieties.
+(* Require Export HoTT.FunextVarieties. *)
 Require Export HoTT.UnivalenceVarieties.
 Require Export HoTT.Extensions.
 Require Export HoTT.Misc.
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -24,7 +24,7 @@
 *)
 
 Require Import HoTT Coq.Init.Peano.
-Require Import UnivalenceImpliesFunext.
+Require Import FunextVarieties UnivalenceImpliesFunext.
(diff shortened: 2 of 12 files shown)


mikeshulman:
Oh, and HoTTBookExercises, which also includes some metatheory.

theories/HoTT.v
```diff
@@ -17,7 +17,7 @@ Require Export HoTT.HSet.
 Require Export HoTT.EquivGroupoids.
 Require Export HoTT.EquivalenceVarieties.
 
-Require Export HoTT.FunextVarieties.
+(* Require Export HoTT.FunextVarieties. *)
```
Alizter: Is this intentional?

Alizter:
Can you rebase?

theories/HoTT.v
mikeshulman: Yes, but we should discuss it.  The idea of some files being "dead-end metatheory" suggested to me that "users" of the library will never need them, so that they should not be exported by `HoTT.v`.  But opinions could differ; what do you think?

Alizter: Given how much useless stuff HoTT.v actually imports I don't think keeping it will do any harm.

JasonGross: I think we should export everything in HoTT.v. IIRC, the dep graphs really on HoTT.v having all the constants, but I might be misremembering

mikeshulman: HoTT.v intentionally leaves out all the files that prove univalence or funext or assert them axiomatically, as discussed at the bottom of the file.  I think it would be consistent with this for it to exclude the metatheory files; the point of HoTT.v is not to create dependency graphs but to make it easy for users to import the whole library.

Alizter: Very well, I am ok with leaving it out.

PR #1274, opened 2020-02-23
Isolate the metatheory files in a subdirectory

mikeshulman:
I didn't put `EquivalenceVarieties` in here because it's not really metatheory, and is not really a dead end in the same way.  (E.g. `Extensions` uses `PathSplit` from there.)  However, we may also want to move some of the more important facts out of it, to reduce the need to import it.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -156,12 +156,10 @@ theories/Comodalities/CoreflectiveSubuniverse.v
 #
 
 theories/HIT/Circle.v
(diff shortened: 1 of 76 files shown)


Alizter:
We had some similar discussions in #1101.

mikeshulman:
Yeah... I guess my opinion about UIF has changed a bit since then.  Although one of the issues discussed there was resolved by #1181 .

PR #1275, opened 2020-02-23
Simpler definition of equiv_functor_forall_covariant

mikeshulman:
This lemma isn't actually used anywhere else in the library, so it's unclear how important it is to have at all.


Alizter:
I've used this lemma in showing that pMap preserves equivalences I believe. I need at some point go back and update my smash coherence branch to use the wild cat library.

PR #1276, opened 2020-02-24
Tactic make_equiv for rearranging sigma and record types, close #1194

### LINKED ISSUE #1194: More general issig?
@jdchristensen suggested that maybe a variety of the `issig` tactic could automatically prove more general equivalences between any pair of nested sigma-types and records that are essentially just iterated destructs and reshufflings.

mikeshulman:
I'm making this a draft PR for now because I think we need to discuss how widely this tactic should be used.  For now, I tried using it almost everywhere I could find where it might conceivably be helpful, so we can see what it _could_ be used for.  I think some of those are clear wins, like `issig_pequiv'` and the lemmas in `Types/Prod` and `Types/Sigma`, but others are less clear.

Generally it seems that this tactic has to be prefixed with a `transitivity` stating exactly what you want the new sigma-type to look like, which is sometimes quite verbose.  Also, often what we want isn't exactly just a rearrangement of sigma-types, but also includes applying some other equivalence in the middle or at the top, often something like `equiv_path_sigma`, so this tactic is less helpful there than it might be otherwise.  So, everyone please opine on where you think this tactic should and shouldn't be used.

One place I could have used this tactic but didn't is `PropResizing/Nat`; my first attempt created universe problems (like pretty much any attempt to modify that darned file now) and I gave up.

Timing-wise, the diff seems mostly noise, except that `BlakersMassey` got over 50% slower.  So at least some of the uses of `make_equiv` in that file are almost certainly out, unless we can speed it up somehow; I haven't tried yet to figure out whether some of the uses there are okay.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -822,4 +822,4 @@ End Examples.
 >>
 *)
 
-(** Some "real-world" examples where [make_equiv_contr_basedpaths] simplifies things a lot includi [hfiber_compose] in [Fibrations.v], [hfiber_pullback_along] in [Limits/Pullback.v], and [equiv_Ocodeleft2plus] in [BlakersMassey.v]. *)
+(** Some "real-world" examples where [make_equiv_contr_basedpaths] simplifies things a lot include [hfiber_compose] in [Fibrations.v], [hfiber_pullback_along] in [Limits/Pullback.v], and [equiv_Ocodeleft2plus] in [BlakersMassey.v]. *)
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -542,3 +542,45 @@ Ltac ev_equiv :=
            | [ |- context[equiv_fun (equiv_inverse ?f) ?a] ] =>
              change ((equiv_inverse f) a) with (f^-1 a)
          end.
+
+(** The following tactic [make_equiv] builds an equivalence between two types built out of arbitrarily nested sigma and record types, not necessarily right-associated, as long as they have all the same underyling components.  This is more general than [issig] in that it doesn't just prove equivalences between a single record type and a single right-nested tower of sigma types, but less powerful in that it can't deduce the latter nested tower of sigmas automatically: you have to have both sides of the equivalence known. *)
+
+(* Perform [intros] repeatedly, recursively destructing all possibly-nested record types. *)
+Ltac decomposing_intros :=
+  let x := fresh in
+  intros x; cbn in x;
+  try match type of x with
+  | ?a = ?b => fail 1           (** Don't destruct paths *)
+  | forall y:?A, ?B => fail 1   (** Don't apply functions *)
+  | _ => elim x; clear x
+  end;
+  try decomposing_intros.
+
+(* A multi-success version of [assumption].  That is, like [assumption], but if there are multiple hypotheses that match the type of the goal, then after choosing the first one, if a later tactic fails we can backtrack and choose another one. *)
+Ltac multi_assumption :=
+  multimatch goal with
+    [ H : ?A |- _ ] => exact H
+  end.
+
+(* Build an element of a possibly-nested record type out of hypotheses in the context. *)
+Ltac build_record :=
+  first [ cbn; multi_assumption
+        | unshelve econstructor; build_record ].
+
+(* Construct an equivalence between two possibly-nested record/sigma types that differ only by associativity and permutation of their components.  We could use [Build_Equiv] and directly construct [eisadj] by decomposing to reflexivity as well, but often with large nested types it seems to be faster to adjointify. *)
+Ltac make_equiv :=
+  snrapply equiv_adjointify;
+    [ decomposing_intros; build_record
+    | decomposing_intros; build_record
+    | decomposing_intros; reflexivity
+    | decomposing_intros; reflexivity ].
+
+(** In case anyone ever needs it, here's the version that doesn't adjointify. *)
+Ltac make_equiv_without_adjointification :=
+  snrapply Build_Equiv;
+    [ decomposing_intros; build_record |
+      snrapply Build_IsEquiv;
+      [ decomposing_intros; build_record
+      | decomposing_intros; reflexivity
+      | decomposing_intros; reflexivity
+      | decomposing_intros; reflexivity ] ].
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -584,3 +584,48 @@ Ltac make_equiv_without_adjointification :=
       | decomposing_intros; reflexivity
       | decomposing_intros; reflexivity
       | decomposing_intros; reflexivity ] ].
+
+(** This version is willing to destruct paths as well, but only as a (multisuccess) second choice. *)
+Ltac decomposing_intros_with_paths :=
+  let x := fresh in
+  intros x; cbn in x;
+  multimatch type of x with
+  | ?a = ?b =>
+    (** Don't destruct paths *)
+    try decomposing_intros_with_paths
+  | forall y:?A, ?B =>
+    (** Don't apply functions *)
+    try decomposing_intros_with_paths
+  | _ =>
+    try (elim x; clear x);
+    try decomposing_intros_with_paths
+  | ?a = ?b =>
+    (** Destruct paths as a second choice *)
+    first [ destruct x
+          | move x before b;
+            generalize dependent b;
+            refine (paths_ind _ _ _)
+          | move x before a;
+            generalize dependent a;
+            refine (paths_ind_r _ _ _)
+          ];
+    try decomposing_intros_with_paths
+  | _ =>
+    try decomposing_intros_with_paths
+  end.
+
+(** This version is willing to create evars in hopes that a later idpath will determine them. *)
+Ltac build_record_with_evars :=
+  first [ cbn; multi_assumption
+        | unshelve econstructor; build_record_with_evars
+        | match goal with
+            |- ?G => let x := fresh in evar (x : G); exact x
+          end; build_record_with_evars ].
+
+Ltac make_equiv_contr_basedpaths :=
+  simple notypeclasses refine (equiv_adjointify _ _ _ _);
+    (** [solve [ unshelve X ]] ensures that [X] succeeds without leaving any leftover evars. *)
+    [ decomposing_intros_with_paths; solve [ unshelve build_record_with_evars ]
+    | decomposing_intros_with_paths; solve [ unshelve build_record_with_evars ]
+    | decomposing_intros_with_paths; reflexivity
+    | decomposing_intros_with_paths; reflexivity ].
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -578,7 +578,7 @@ Ltac make_equiv :=
     | decomposing_intros; exact idpath
     | decomposing_intros; exact idpath ].
 
-(** In case anyone ever needs it, here's the version that doesn't adjointify. *)
+(** In case anyone ever needs it, here's the version that doesn't adjointify. It's not the default, because it can be slow. *)
 Ltac make_equiv_without_adjointification :=
   simple notypeclasses refine (Build_Equiv _ _ _ _);
     [ decomposing_intros; build_record |
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -543,12 +543,14 @@ Ltac ev_equiv :=
              change ((equiv_inverse f) a) with (f^-1 a)
          end.
 
+(** ** Building equivalences between nested sigma and record types *)
+
 (** The following tactic [make_equiv] builds an equivalence between two types built out of arbitrarily nested sigma and record types, not necessarily right-associated, as long as they have all the same underyling components.  This is more general than [issig] in that it doesn't just prove equivalences between a single record type and a single right-nested tower of sigma types, but less powerful in that it can't deduce the latter nested tower of sigmas automatically: you have to have both sides of the equivalence known. *)
 
 (* Perform [intros] repeatedly, recursively destructing all possibly-nested record types. *)
 Ltac decomposing_intros :=
   let x := fresh in
-  intros x; hnf in x;
+  intros x; cbn in x;
   try match type of x with
   | ?a = ?b => fail 1           (** Don't destruct paths *)
   | forall y:?A, ?B => fail 1   (** Don't apply functions *)
@@ -570,7 +572,7 @@ Ltac build_record :=
 
 (* Construct an equivalence between two possibly-nested record/sigma types that differ only by associativity and permutation of their components.  We could use [Build_Equiv] and directly construct [eisadj] by decomposing to reflexivity as well, but often with large nested types it seems to be faster to adjointify. *)
 Ltac make_equiv :=
-  snrapply equiv_adjointify;
+  simple notypeclasses refine (equiv_adjointify _ _ _ _);
     [ decomposing_intros; build_record
     | decomposing_intros; build_record
     | decomposing_intros; exact idpath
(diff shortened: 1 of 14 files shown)


mikeshulman:
Note that the tactic is a bit modified from the one I suggested in #1194, to prevent it from destructing paths and applying functions.  Any suggested improvements to the tactic would also be welcome.

theories/Basics/Equivalences.v
```diff
@@ -526,3 +526,45 @@ Ltac ev_equiv :=
...
+
+(* Construct an equivalence between two possibly-nested record/sigma types that differ only by associativity and permutation of their components.  We could use [Build_Equiv] and directly construct [eisadj] by decomposing to reflexivity as well, but often with large nested types it seems to be faster to adjointify. *)
+Ltac make_equiv :=
+  snrapply equiv_adjointify;
+    [ decomposing_intros; build_record
+    | decomposing_intros; build_record
+    | decomposing_intros; reflexivity
+    | decomposing_intros; reflexivity ].
+
+(** In case anyone ever needs it, here's the version that doesn't adjointify. *)
```
jdchristensen: Why not provide only the second version?

theories/Fibrations.v
```diff
@@ -78,16 +78,14 @@ Definition hfiber_functor_hfiber {A B C D}
...
-  apply equiv_functor_sigma_id; intros r.
-  apply equiv_functor_sigma_id; intros s; cbn.
+  rapply (equiv_functor_sigma_id _ oE _ oE equiv_functor_sigma_id _).
+  1:intros; rapply equiv_path_sigma.
+  2:intros; symmetry; rapply equiv_path_sigma.
+  cbn.
+  transitivity {ar : {a : A & f a = b} & {s : h ar.1 = c & transport (fun x : B => k x = g c) ar.2 (((p ar.1)^)^ @ ap g s) = q^}}.
+  2:make_equiv.
+  apply equiv_functor_sigma_id; intros [a r].
+  apply equiv_functor_sigma_id; intros s.
```
jdchristensen: I'd say that the change here is neutral.  The proof is a bit shorter, but the transitivity line is much harder to write.

theories/Fibrations.v
```diff
@@ -140,30 +138,23 @@ Section UnstableOctahedral.
...
     apply equiv_functor_sigma_id; intros a; simpl.
-    refine (_ oE _); revgoals.
-    - refine (equiv_functor_sigma_id
-                (fun p => (equiv_path_sigma _ _ _)^-1)).
-    - cbn. refine (_ oE equiv_sigma_symm _).
-      apply equiv_sigma_contr; intros p.
-      destruct p; simpl; exact _.
+    refine (_ oE equiv_sigma_symm _).
+    apply equiv_sigma_contr; intros p.
+    destruct p; simpl; exact _.
```
jdchristensen: Why is this part changed?  It looks roughly equivalent to what was there before.

theories/Modalities/Lex.v
```diff
@@ -244,13 +244,12 @@ Section ImpliesLex.
...
-    refine (_ oE (equiv_sigma_assoc _ _)^-1).
-    apply equiv_functor_sigma_id; intros y; cbn.
     refine (_ oE (equiv_functor_sigma_id _)).
     2:intros; symmetry; apply equiv_path_sigma.
     cbn.
-    refine (_ oE equiv_sigma_symm _).
-    apply equiv_sigma_contr; intros p.
+    transitivity {yp : {y:Y & f y = x} & {q : g (f yp.1) = g x & transport (fun x' => g x' = g x) yp.2 (1 @ ap idmap q) = idpath } }.
+    1:make_equiv.
+    apply equiv_sigma_contr; intros [y p].
```
jdchristensen: Not sure this is worth it.

jdchristensen:
There are many places where make_equiv is a clear win!  Nice.

About BlakersMassey, I guess you could look at the html timing file to see what is slow, and undo the relevant changes.  Or just try undoing them one at a time until you find the one that fixes the build time.  You'd think that such a simple equivalence would be fast, so I wonder what's going on.  Is it the make_equiv tactic itself that is slow, or the resulting proofs?

I was also wondering about creating `make_equiv_via mid`, but I see that sometimes it is the first goal and sometimes the second that you run `make_equiv` on, so doing it on two lines is probably best.

(I also noticed that you using `transivitity` instead of `equiv_via`.  I guess we should probably use transitivity throughout the library, like the comment before `equiv_via` says?  We only use `equiv_via` in six places.)

jdchristensen:
LGTM once the speed issue is fixed.  It's optional to me whether the borderline uses are reverted.

theories/Basics/Equivalences.v
mikeshulman: Because the first version is sometimes faster, sometimes so much so that the second is wholly impractical.  I think it was in BlakersMassey that I noticed this, though, so it's possible that if we remove the really slow uses there it wouldn't make much difference to use only the second one.

theories/Fibrations.v
mikeshulman: Probably I was rearranging it to try to use `make_equiv` but failed and forgot to put it all back.

theories/Basics/Equivalences.v
jdchristensen: Just to be clear, you mean that the tactic itself is slow, not that further uses of the proofs it generates are slow, right?  When this happens, do you know if it's the last decomposing_intros that is slow, or the last reflexivity?  I'm surprised that either would be slow.  It seems more likely that build_record might be slow in some cases, since it might have to backtrack, but I don't see how the two versions would differ here.

If I'm right that it's the tactic that can be slow, can you update the comment to mention this?

mikeshulman: It's the last reflexivity; I think in one case I didn't even wait long enough for it to finish.  I think sometimes reflexivity is slow on large terms because it has to essentially compute them enough in order to check that they're equal, which can be nontrivial.

theories/Fibrations.v
mikeshulman: I tried simplifying it a bit without using the new tactic.

jdchristensen:
`make_equiv_contr_basedpaths` is pretty clever.  Is it reasonably fast?

When I see a powerful tactic like this, I always have trouble understanding it, and it's hard to figure out how it works and when it will succeed.  One thing that I think would help a lot as a kind of documentation would be to give a couple of examples where the tactics that eventually succeed are listed by hand, separated by periods.  Then it would be easy to step through the proof and see how the tactic works.  What do you think?

mikeshulman:
As Dan already noticed, I added another tactic, `make_equiv_contr_basedpaths`, that can (sometimes) also contract based path-spaces automatically.  It seems to be reasonably fast when it works, although I haven't timed it yet.  Your suggestion for some examples as documentation is a good one, I'll work on that when I get a chance.

I'm a little ambivalent about it myself, partly because of unease about how powerful it is; it may be doing automatically things that a human reader would instead want to see spelled out.  Also the equivalences that it defines are not, I think, definitionally the same as those we had constructed by hand; they do more direct destruction of paths rather than calling `transport`.  This already bit in a couple of places requiring changes to later proofs.  So while I'm glad that it seems to work, I'm not entirely sure whether I actually want to start using it throughout the library.

BTW, I haven't looked into the speed issue in BlakersMassey yet at all.

mikeshulman:
(It's possible that if we computed everything down to `match`es they would end up definitionally the same, I haven't checked -- but they certainly aren't syntactically the same in any useful sense.)

mikeshulman:
Speed-wise, compared to current master the only significant slowdown in this branch is still in BlakersMassey.

mikeshulman:
(But there is no significant speedup either.)

jdchristensen:
If it's well-documented and doesn't slow things down, then I think we should use it.

mikeshulman:
I simplified the proofs in BlakersMassey a bit more.  It should be possible to simplify them even more using `make_equiv_contr_basedpaths`, but currently that doesn't work; I haven't debugged it yet.  Unfortunately this simplification makes them even slower.

When running interactively, there are huge slowdowns at the end of each subgoal of `equiv_Ocodeleft2plus`.  This doesn't show up in `make TIMING=1` though, suggesting to me that what's slow there is just something like figuring out how to display the next subgoal.

jdchristensen:
Do you think there might be some way to hide some of the context from Coq while using make_equiv, so that it does less unnecessary unfolding?  It really seems like these sorts of equivalences should be quick.  Maybe using the generalize tactic would help Coq out?  Or maybe some of them can be pulled out into helper lemmas?  But that's a pain to do.  Or maybe there are some Coq commands that can give hints to not unfold things, locally?

mikeshulman:
I suspect one possibility that might speed things up is to use Records instead of 8-fold nested sigma-types.  I'm going to try that when I'm done figuring out why `make_equiv_contr_basedpaths` isn't working on `equiv_Ocodeleft2plus`.

mikeshulman:
Well, I found one thing that speeds up BlakersMassey considerably: making `frobnicate` as `Opaque` except in the one place where it needs to be evaluated.  But this is mostly unrelated to `make_equiv` which is still slowing things down, although the two effects combined mostly cancel each other out so the current version is about the same speed as master.

Alizter:
By the way, you can use the `time` tactic to time tactics. Say you had a line
```coq
  some_tactic.
```
if you prefix it with
```coq
  time some_tactic.
```
It will churn out timing data for that line.

jdchristensen:
In Proof General, you can also toggle a timing switch which then shows the time used for each complete step (up to a period).  But for some reason, it doesn't show the time for certain lines.

I like the trick you used to toggle frobnicate between transparent and opaque.  We should keep it in mind in case there are other places that it would help.  I'm still puzzled by make_equiv being slow and wonder if this trick might help.  Or generalizing some things to be more generic.

mikeshulman:
Ok, I figured out why `make_equiv_contr_basedpaths` doesn't work without the `unfold` (and also doesn't work if `codeleft2` and `Ocodeleft2a` are records).

Suppose that `a : A` is a section variable, and that our current goal `B` is a definition in the section that uses `a`.  Then the tactic `revert a` will fail, complaining that `a is used implicitly in B in conclusion`.  But the tactic `generalize dependent a` will succeed, adding `forall a0:A` in front of the goal, but also _leaving `a` as a hypothesis in the context_, and leaving `B` in the conclusion as depending on that `a` in the context rather than the `a0`.  Of course, since `B` hasn't yet been generalized over `a`, there's no way it could depend on `a0`; but I think this behavior is at best unintuitive and might qualify as a bug in Coq, especially since the documentation for `generalize dependent` gives no hint that it might happen.

How can I do a `generalize dependent` that will fail if I'm trying to generalize a section variable that the goal depends on?  Is there a way to test whether a particular hypothesis is a section variable?  Or whether it's been cleared after `generalize dependent`?

mikeshulman:
(In this case I was able to work around it by using `+` instead of `first`, but it would be faster to avoid going down the wrong path from the start.)

mikeshulman:
Turns out that with records, it's blindingly fast.  In this branch, BlakersMassey is now 5.5 seconds faster (40%) than in master, and the entire branch is 19 seconds faster than master.

I suppose for a real apples-to-apples comparison, I ought to check out how fast BlakersMassey would be with records and opaque frobnicate but without `make_equiv`.  I might or might not feel like actually doing that...

mikeshulman:
BTW, about 30% of the time in BlakersMassey is now spent on the `cbn` with transparent `frobnicate`.

mikeshulman:
Intriguing!  I just ran the apples-to-apples comparison.  In order to use records for BlakersMassey in master, I had to add `issig` lemmas for all the records (which `make_equiv` obviates the need for).  The end result is about the same speed as master, but the time is spent rather differently -- now 40% of the time is spent proving the `issig` lemmas!  I guess the lesson from BlakersMassey is that big nested sigma-types are slow, and with `make_equiv` we can sometimes avoid dealing with them entirely.

I'm now convinced that `make_equiv` and `make_equiv_contr_basedpaths` are fairly speedy and good things to have in our toolkit.  There are probably further possible improvements that can be made to them, but they're now at a point where I'm happy to merge them into master.  So I'll add some more examples like Dan requested and then open this PR for review.

JasonGross:
> By the way, you can use the `time` tactic to time tactics. Say you had a line
> 
> ```coq
>   some_tactic.
> ```
> 
> if you prefix it with
> 
> ```coq
>   time some_tactic.
> ```
> 
> It will churn out timing data for that line.

You can also `Set Ltac Profiling` and then later `Show Ltac Profile` (and use `Reset Ltac Profile` to wipe the slate) to get much more detailed timing info.  You can also set it at the top of a file and get the info for the whole file (or pass `-ltac-profile` (or maybe it's `-profile-ltac`?) to `coqc` to get the info on the command line).

JasonGross:
> How can I do a `generalize dependent` that will fail if I'm trying to generalize a section variable that the goal depends on? Is there a way to test whether a particular hypothesis is a section variable? Or whether it's been cleared after `generalize dependent`?

Do `revert dependent H; assert_fails (move H at top)` (or pick some other tactic that'll always succeed as long as `H` is in the context).  (Note that `revert dependent` is the same as `generalize dependent`, but only accepts hypothesis names.)

mikeshulman:
Thanks!

> Note that revert dependent is the same as generalize dependent, but only accepts hypothesis names.

Ah, fabulous!  I see that that is new in 8.10.  Now all we need is for `generalize dependent` to _not_ clear things from the context (like ordinary `generalize`) and the tactic names will actually make sense.

mikeshulman:
Hmm... `revert dependent` doesn't work for me; Coq says `Error: No such hypothesis: dependent`.

JasonGross:
Maybe our version of Init.Tactics stripped out too much stuff?  It's defined at https://github.com/coq/coq/blob/fe1335eb350c305142bf4be57c681891515a5dac/theories/Init/Tactics.v#L243-L244 .

mikeshulman:
Yes, that's certainly missing from our `Init.Tactics`.

Alizter:
I think it would be worth adding to Overture.

theories/Basics/PathGroupoids.v
```diff
@@ -710,6 +710,14 @@ Proof.
   (* subst u. rewrite X. *)
 Defined.
 
+(** Occasionally [paths_rect] shows up; this lets us turn it into transport. *)
+Definition paths_rect_transport {A : Type} (P : A -> Type) {x y : A}
+           (p : x = y) (u : P x)
+  : paths_rect A x (fun a _ => P a) u y p = transport P p u.
+Proof.
+  by destruct p.
```
SkySkimmer: Does reflexivity not work here?

Alizter: It does work.

mikeshulman: I suppose that's because they're both defined as the same underlying [match]?

mikeshulman:
Shouldn't we add it to `Init.Tactics` instead?

mikeshulman:
Also `assert_fails (move H at top)` gives me `Error: Syntax error: ';' or ',' or ')' expected after [constr:operconstr level 200] (in [constr:operconstr]).` on the `at`.

Alizter:
@mikeshulman Generally we are trying to move away from having the stripped down version of coq. There are people in the coq development team who are trying to get coq working with no standard library. Currently I don't think it is possible to remove our version of the standard library completely.

But when it comes to "user defined" tactics we should put them in Overture, just like we did for rapply.

mikeshulman:
I suppose that's true... but the fact that `revert dependent` is described in the reference manual (unlike, say, "rapply") means that the user (like me) may not think of it as "user defined" and be confused to see it in Overture.  Maybe we should have a separate file in `Basics` for "stuff we took from the standard library"?

Alizter:
Well I've always though that Overture was getting too big for its own good. There are several tactics in there which could all be moved into a Basics/Tactics.v file.

JasonGross:
I guess we also need to add https://github.com/coq/coq/blob/fe1335eb350c305142bf4be57c681891515a5dac/theories/Init/Tactics.v#L327-L334 to get `assert_fails`.  (The error that you got is because it's trying to parse `move H at top` as something other than a tactic.)

mikeshulman:
Thanks Jason, I figured that out too.  But we do currently have the `not` tactic which seems to work here; how does it differ in practice from `assert_fails`?

JasonGross:
Where is our `not` tactic defined?  My guess is that we might have an older version of this tactic that behaves incorrectly when the tactic it's trying to negate fully solves the goal.  It shouldn't matter in this particular case.

mikeshulman:
In overture:

```coq
(** [not tac] is equivalent to [fail tac "succeeds"] if [tac] succeeds, and is equivalent to [idtac] if [tac] fails *)
Tactic Notation "not" tactic3(tac) :=
  (tryif tac then fail 0 tac "succeeds" else idtac); (* error if the tactic solved all goals *) [].
```

JasonGross:
I think `gfail` fails even if there are no goals, so it should accomplish the same thing as our `fail` with `; []`; I don't think it matters which one we do.  We should probably add `once` to our version, to have better behavior with multisuccess tactics.  For example, our version will loop on any tactic with infinitely many successes, and I think it'll have the wrong behavior on `idtac + fail`.

mikeshulman:
Should we rename ours to `assert_fails` to match the stdlib?

JasonGross:
> Should we rename ours to `assert_fails` to match the stdlib?

Probably a good idea

mikeshulman:
So... what exactly does [generalize dependent] do when applied to a term that's _not_ the name of a hypothesis?  Does it behave any differently from [generalize]?

mikeshulman:
Ok, now rebased on top of #1278, and with examples; ready for review!

JasonGross:
> So... what exactly does generalize dependent do when applied to a term that's not the name of a hypothesis? Does it behave any differently from generalize?

I think it might find occurrences of that term in all the hypotheses and generalize over them insofar as it can, similar to how induction works on things that are not hypotheses?  I'm not sure, you could test it.

theories/Basics/Equivalences.v
```diff
@@ -554,63 +556,254 @@ Ltac build_record :=
...
+    intros x; cbn in x.
+    elim x; clear x.
+    intros a2; cbn in a2.
+    intros r; cbn in r.
+    cbn; exact idpath.
+Defined.
+>>
+*)
+
+(** Some "real-world" examples where [make_equiv] simplifies things a lot include  the associativity/symmetry proofs in [Types/Sigma.v], [issig_pequiv'] in [Pointed/pEquiv.v], and [loop_susp_adjoint] in [Pointed/pSusp.v]. *)
```
jdchristensen:
```suggestion
(** Some "real-world" examples where [make_equiv] simplifies things a lot include the associativity/symmetry proofs in [Types/Sigma.v], [issig_pequiv'] in [Pointed/pEquiv.v], and [loop_susp_adjoint] in [Pointed/pSusp.v]. *)
```

theories/Basics/Equivalences.v
```diff
@@ -554,63 +556,254 @@ Ltac build_record :=
...
+      not (move a at top).
+      refine (paths_ind_r _ _ _).
+      intros b; cbn in b.
+      exact idpath.
+  Defined.
+End Examples.
+>>
+*)
+
+(** Some "real-world" examples where [make_equiv_contr_basedpaths] simplifies things a lot includi [hfiber_compose] in [Fibrations.v], [hfiber_pullback_along] in [Limits/Pullback.v], and [equiv_Ocodeleft2plus] in [BlakersMassey.v]. *)
```
jdchristensen:
```suggestion
(** Some "real-world" examples where [make_equiv_contr_basedpaths] simplifies things a lot include [hfiber_compose] in [Fibrations.v], [hfiber_pullback_along] in [Limits/Pullback.v], and [equiv_Ocodeleft2plus] in [BlakersMassey.v]. *)
```

jdchristensen:
Those examples are great.  Thanks!

theories/Basics/Equivalences.v
```diff
@@ -526,3 +526,284 @@ Ltac ev_equiv :=
...
+
+(* Construct an equivalence between two possibly-nested record/sigma types that differ only by associativity and permutation of their components.  We could use [Build_Equiv] and directly construct [eisadj] by decomposing to reflexivity as well, but often with large nested types it seems to be faster to adjointify. *)
+Ltac make_equiv :=
+  simple notypeclasses refine (equiv_adjointify _ _ _ _);
+    [ decomposing_intros; build_record
+    | decomposing_intros; build_record
+    | decomposing_intros; exact idpath
+    | decomposing_intros; exact idpath ].
+
+(** In case anyone ever needs it, here's the version that doesn't adjointify. *)
```
jdchristensen:
```suggestion
(** In case anyone ever needs it, here's the version that doesn't adjointify. It's not the default, because it can be slow. *)
```

mikeshulman: This is already mentioned in the previous comment, but I suppose it doesn't do any harm to mention it here too.

jdchristensen: Sorry, I missed the earlier comment.

PR #1277, opened 2020-02-24
Reals with locators

abooij:
(This PR depends on #1261.)

For a Cauchy complete Archimedean ordered field `F` (such as the Dedekind reals or the HoTT book reals), we may wonder which elements of `F` are limits of rationally-valued Cauchy sequences. This can be characterized succinctly by the existence of a _locator_ for `x : F`, where locators are defined as
```coq
Definition locator (x : F) := forall q r : Q, q < r -> (' q < x) + (x < ' r).
```

Additionally, having such a locator allows to compute with `x`, in the sense that we can compute, for instance, a signed-digit representation of `x`.

This, and more, is worked out my PhD thesis ([non-final version](https://www.cs.bham.ac.uk/~abb538/thesis-first_submission.pdf)).

I have formalized some results from my PhD thesis in the HoTT library using @SkySkimmer's work on the HoTTClasses library.

The development starts with a Cauchy complete Archimedean ordered field. Hence, this PR includes
* a definition of ordered fields (in `theories/Classes/interfaces/orders.v`),
* the Archimedean property (in `theories/Classes/interfaces/archimedean.v`, rather than `orders.v`, since it depends on an inclusion from the rationals in `interfaces/rationals.v`), and
* Cauchy completeness (in `theories/Classes/interfaces/cauchy.v` for similar reasons).

Then, in `theories/Analysis/Locator.v`, locators are defined as above, and this definition is shown to be equivalent to another. Locators for rationals are defined, and the majority of Theorem 6.7.6 in my thesis is implemented, namely the construction of locators for algebraic operations (if `x` has a locator then so does `-x`, if `x` and `y` do then so does `x+y`, etc).

The work in this PR could be used to develop real numbers in the HoTT library. Along the way, I have added some results to the HoTT library as needed, such as 1-dimensional Sperner's lemma (see #1261), some more facts about the finite types (such as an inclusion `Fin n -> nat` with corresponding path lemmas), a minor result about bounded search (namely that funext suffices to show that it also works for types equivalent to `nat`) (previously discussed in #944), some lemmas about orderings and lattices, some first lemmas about ordered fields, and some facts about decidable propositions.


mikeshulman:
Wow, this is impressive!  I'm looking forward to checking it out (but it may be a little while before I have time).

theories/Classes/interfaces/round.v
```diff
@@ -0,0 +1,13 @@
...
+        HoTT.Classes.interfaces.abstract_algebra
+        HoTT.Classes.interfaces.orders
+        HoTT.Classes.interfaces.naturals
+        HoTT.Classes.implementations.peano_naturals.
+
+Section round_up.
+
+  Class RoundUpStrict A `{IsSemiRing A} `{StrictSemiRingOrder A}
+    := round_up_strict : forall a : A, {n : nat & a < naturals_to_semiring nat A n}.
+  Global Arguments round_up_strict A {_} {_} {_} {_} {_} {_} {_} {_} {_} {_} {_} {_} _.
```
Alizter: You can write this as:
```suggestion
  Global Arguments round_up_strict A {_ _ _ _ _ _ _ _ _ _ _ _} _.
```

andrejbauer:
Now I *really* have to read that thick thesis that's sitting on my desk.

Alizter:
Can you rebase this?

abooij:
> Can you rebase this?

Done!

abooij:
> Therefore it seems to me that the Dedekind reals are the most likely candidate for a CCAOF in the near future. Have you got any plans to implement the reals?

I don't have such plans. In fact, I'm not aware of any complete formalization of the Dedekind reals including its structure as a Cauchy complete Archimedean ordered field. The work of [Andrej and his students](https://github.com/andrejbauer/dedekind-reals) is quite good has a way to go towards this goal.

Alizter:
@mikeshulman Do you still want to review this or should we go ahead and merge?

theories/Analysis/Locator.v
```diff
@@ -0,0 +1,802 @@
...
+        HoTT.Classes.orders.rings
+        HoTT.Classes.orders.dec_fields
+        HoTT.Classes.orders.semirings
+        HoTT.Classes.orders.lattices
+        HoTT.Classes.theory.rings
+        HoTT.Classes.theory.apartness
+        HoTT.Classes.theory.rationals.
+
+(* Strangely, it seems that combining the next import with the above
+list breaks some instance search? *)
```
mikeshulman, line 33: Does this mean that
```coq
Require Import A.
Require Import B.
```
behaves differently than
```coq
Require Import A B.
```
?  That sounds like a bug in Coq to me...

mikeshulman: Oh, and we don't usually put hard newlines in comments.  (-:

theories/Analysis/Locator.v
```diff
@@ -0,0 +1,802 @@
...
+      assert (qP_lower : {q : Q | P_lower q})
+        by refine (minimal_n_alt_type Q Q_eq P_lower _ P_lower_inhab).
+      destruct qP_lower as [q Pq].
+      exists (q - 1).
+      unfold P_lower in Pq. simpl in *.
+      apply (un_inl _ Pq).
+    Qed.
+
+    Local Definition lt1N (r : Q) : r < r + 1 := ltQposQ r 1.
+    (* Assume we have an enumeration of the rationals. *)
```
mikeshulman, line 268: I don't understand why this comment is placed here.

theories/Analysis/Locator.v
```diff
@@ -0,0 +1,802 @@
...
+      - apply inl. apply char_minus_right. rewrite <- preserves_negate. assumption.
+    Qed.
+
+    Section recip_pos.
+      Context (xpos : 0 < x).
+      Local Definition recip_nu := positive_apart_zero x xpos.
+
+      (* Note: If you are going to attemp to fill in the foles, you may
+         be interested in `dec_recip_to_recip` in
+         `Classes.theory.dec_fields` *)
```
mikeshulman, line 557: I also don't understand this comment (plus it has some typos).  `dec_recip_to_recip` is already used in this proof; what is it that one might attempt to fill in further?

theories/Analysis/Locator.v
abooij: Yes, that's what that meant. When I was writing that, I did not have time to figure out a minimal example of this, especially since it only occurred in this rather large file - I wouldn't know where to start looking.

I'll remove some newlines.

theories/Analysis/Locator.v
abooij: Me neither. That must've gotten left behind when I was moving code around.

theories/Analysis/Locator.v
abooij: Past me did not believe that these holes would actually get filled.

Present me will fix this by removing that comment entirely.

abooij:
I fixed some comments and newlines roughly along the lines suggested by @mikeshulman.

theories/Analysis/Locator.v
Alizter: I've made an issue for this at #1336.

PR #1278, opened 2020-02-26
Add some tactics from the stdlib to Overture

mikeshulman:
While I was at it, I removed a bunch of unnecessary imports of `HoTT.Tactics`.  Currently the tactics therein are only used in `Categories`, three places in `Classes`, and one place in `Diagrams/Diagram.v`.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -719,16 +719,16 @@ Global Arguments hfiber {A B}%type_scope f%function_scope y.
 
 (** *** More tactics *)
 
-(** Clear a hypothesis and also its dependencies.  Taken from Coq stdlib. *)
-
+(** Clear a hypothesis and also its dependencies.  Taken from Coq stdlib, with the performance-enhancing change to [lazymatch] suggested at [https://github.com/coq/coq/issues/11689]. *)
 Tactic Notation "clear" "dependent" hyp(h) :=
   let rec depclear h :=
   clear h ||
-  match goal with
+  lazymatch goal with
    | H : context [ h ] |- _ => depclear H; depclear h
   end ||
   fail "hypothesis to clear is used in the conclusion (maybe indirectly)"
  in depclear h.
+
 
 (** A version of [generalize dependent] that applies only to a hypothesis.  Taken from Coq stdlib. *)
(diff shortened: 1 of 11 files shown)


mikeshulman:
(See discussion at #1276)

theories/Basics/Overture.v
```diff
@@ -719,6 +719,21 @@ Global Arguments hfiber {A B}%type_scope f%function_scope y.
 
 (** *** More tactics *)
 
+(** Clear a hypothesis and also its dependencies.  Taken from Coq stdlib. *)
+
+Tactic Notation "clear" "dependent" hyp(h) :=
+  let rec depclear h :=
+  clear h ||
+  match goal with
```
JasonGross: Using lazymatch here would be more efficient.  I'll open an issue request on Coq

mikeshulman: Should we change it here now?

JasonGross: Reported as https://github.com/coq/coq/issues/11689

JasonGross: I vote yes, and leave a comment pointing at https://github.com/coq/coq/issues/11689 and noting the difference from the standard library

jdchristensen:
One thing I didn't understand about the discussion was the location of these tactics taken from the standard library.  Does the HoTT library import some portion of Coq's standard library?  If so, wouldn't it make sense for things that are taken from the standard library to be stored in a file with the same name as in the standard library, making it easy to compare the trimmed standard library against the original?

mikeshulman:
We do include a massively stripped-down copy of the Coq standard library in `coq/theories/`.  I think @Alizter 's argument was that we want to pretend as much as possible that that copy doesn't exist at all (as one day may be possible), so that our code (in `theories/`) is a replacement for the standard library rather than something built on top of it.  I can see your argument too though, I don't really feel strongly one way or the other.

spitters:
@jdchristensen 
@Alizter tells me we are getting close to resolving this one.
https://github.com/HoTT/HoTT/issues/963

Note that the stdlib itself is in flux. Ideally, we would target https://github.com/coq/stdlib2/ but that project seems stalled, partially because there are issues with type classes and canonical structures to be resolved.

Issue #1240, opened 2020-02-16
Change `f^-1` to only work for bundled equivalences


Alizter:
As discussed in #1165, #1238 and elsewhere, it would be beneficial for the notation `^-1` to only mean `equiv_inverse`. Otherwise currently coq spends a lot of time searching for IsEquiv instances when it really should be carrying them around. There are a few places in the library, notably my cubical library which I have made note of #1217, that use unbundled equivalences and invert them. These will have to be bundled up then inverted. We should expect to see a performance increase.

Alizter:
In #1252 I ran into issues where coq was typechecking `f^-1`. If I checked the entire proof it would hang, but if I stepped through it would succeed. `f^-1` is in hindsight a terrible notion because it causes coq to unfold `f` when looking for instances of `IsEquiv f`. This can get messy, especially if `f` has some troublesome insides. Whats even worse is that in many cases it is `^-1%equiv` that was really intended.

mikeshulman:
i started trying to do this, but there is a problem: `cbn` unfolds `f^-1 x` to `equiv_inv f x`, which we don't want to have to look at.  Is there a way to prevent that?

Alizter:
Well my plan was to have two notations like we currently do, but somehow make the `equiv` one the "default" one. I wonder if we could make an "isequiv_scope" or something. Then we would open equiv_scope on top of isequiv_scope.

mikeshulman:
Maybe this is a job for someone who understands scopes better than I do.

Alizter:
This doesn't seem like an easy thing to do after having tried it a few times. I will mess around with it tomorrow.

mikeshulman:
I wonder if @fpvandoorn has any ideas.

mikeshulman:
This isn't the first time I've been annoyed that Coq doesn't have a way to set a notation as "only displaying".

mikeshulman:
I guess one possibility -- probably not a good one -- is to define `Equiv` as a big record with all the fields in it directly, and then define `IsEquiv f` to mean that there is an `Equiv` whose underlying function is equal to `f`.

SkySkimmer:
>This isn't the first time I've been annoyed that Coq doesn't have a way to set a notation as "only displaying".

Isn't that what `only printing` is for? (the note just before https://coq.github.io/doc/master/refman/user-extensions/syntax-extensions.html#the-infix-command)

mikeshulman:
[...]

PR #1279, opened 2020-02-27
Make contr_equiv and trunc_equiv transparent; close #1091

### LINKED ISSUE #1091: Why is trunc_equiv opaque?

mikeshulman:
I still don't entirely understand everything that's going on, but it seems to work for now.

Making `contr_equiv` transparent caused a blowup on a `cbn` in `Spaces/Universe`, which I tracked down to Coq spending a long time looking for the center of `contr_baut_rigid`.  Using `contr_change_center`, as Dan suggested in #1271, to give the obvious center explicitly fixed that.

Making `trunc_equiv` also transparent caused, as Ali noted in #1091, a blowup in `Spaces/Card` -- not on `strip_truncations`, but on another `cbn`.  I'm not sure exactly why, but looking at the proof term I saw that Coq was unnecessarily passing back and forth between `IsTrunc n` and `In (Tr n)` (e.g. deducing truncatedness of `Unit` from `inO_unit`), so I tried to increase the cost of the hint `istrunc_inO_tr` to prevent this.

That got me into some weeds with the meaning of `Hint Immediate`.  As I noted at coq/coq#11697, the documentation suggests that `Hint Immediate` should be equivalent to `Hint Extern` with `simple apply term; trivial`, but this doesn't seem to be the case since the latter can leave over subgoals whereas `Hint Immediate` is not supposed to.  On the other hand, `Hint Immediate` does sometimes appear to leave over subgoals that can be solved by further typeclass inference in a way that `Hint Extern` with `simple apply term; solve [ trivial ]` does not; see the comments added to `Trunc_min` in `Truncations/Core.v`.  But the upshot is that I managed to increase the cost by making `istrunc_inO_tr` a `Hint Extern` instead, at the price of modifying the proofs of `Trunc_min` and `isconnected_join` in two places where, if I understand the documentation of `Hint Immediate` correctly, they _shouldn't_ have been working in the first place.

Fortunately, increasing this cost seemed to be sufficient to make `Spaces/Card` compile.

Then I had to also modify a couple of proofs in `HIT/surjective_factor` and `Classes/theory/rationals`.  I have no idea why these worked in the first place; they were calling `center _ c` with `c` a proof of contractibility, but the second argument of `center` is (and always has been) _implicit_, so this shouldn't make any sense.  Nor do I understand why the changes that I made here caused them to suddenly break.

I also had to modify a proof in `HoTTBookExercises`, which I also don't think should have worked in the first place.  It was calling `trunc_equiv _ e` to get a function of the form `IsTrunc A -> IsEquiv e -> IsTrunc B`, but the last arguments of `trunc_equiv` are implicit and _maximally inserted_, so it seems that Coq should have (as it now does) complained that it was unable to fill them in.  My only guess here is that somehow transparent and opaque constants behave differently with respect to insertion of implicit arguments. 

At this point the library compiled, but some things had gotten very slow:

```
After    | File Name                      | Before   || Change    | % Change 
-----------------------------------------------------------------------------
7m04.93s | Total                          | 5m59.32s || +1m05.60s | +18.25%  
-----------------------------------------------------------------------------
0m43.01s | Spaces/Card.vo                 | 0m02.04s || +0m40.96s | +2008.33%
0m09.91s | Pointed/pMap.vo                | 0m00.81s || +0m09.09s | +1123.45%
0m10.01s | Algebra/Group.vo               | 0m02.52s || +0m07.49s | +297.22% 
0m07.82s | Pointed/pType.vo               | 0m00.25s || +0m07.57s | +3028.00%
0m03.87s | Extensions.vo                  | 0m01.44s || +0m02.43s | +168.75% 
0m06.68s | Algebra/AbelianGroup.vo        | 0m05.68s || +0m01.00s | +17.60%  
```

So I tried adding `Opaque trunc_equiv`.  This sped `Card` right back up to where it used to be.

In `pMap`, `Group`, `pType`, and `Extensions` the culprit was typeclass inference searching for `equiv_isequiv`.  I was able to solve this by calling it explicitly, adding `%equiv`, and defining versions of the `moveX_equiv_Y` lemmas that act on bundled equivalences (more evidence for #1240).  But I'm not sure why these changes suddenly made those searches so much slower.

I still don't know what slowed down `AbelianGroup`.  The new slowest line is the `Defined` of `Abel_ind_beta_ab_comm`; I don't see any inverses of equivalences, and `Opaque` doesn't help.  But at least this slowdown is relatively minor.

So now at least we have a transparent `contr_equiv` and `trunc_equiv`.  But if anyone can help explain any of the points where I'm confused, I'd be much obliged.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1125,7 +1125,7 @@ Proof.
     refine ((equiv_concat_l (transport_paths_lr q p)^ p)^-1 oE _).
     refine ((equiv_concat_l (concat_p_pp _ _ _) _)^-1 oE _).
     apply equiv_moveR_Vp. }
-  assert (HK := trunc_equiv _ e^-1).
+  assert (HK := @trunc_equiv _ _ e^-1 (-1)).
   assert (u : forall (X:Type) (p:X=X), p @ 1 = 1 @ p).
   { intros X p; rewrite concat_p1, concat_1p; reflexivity. }
   pose (alpha := (fun X p => (idpath X ; u X p)) : K).
--- a/theories/Algebra/Group.v
+++ b/theories/Algebra/Group.v
@@ -406,7 +406,7 @@ Proof.
   srapply Build_HasMorExt.
   intros A B f g; cbn in *.
   snrapply @isequiv_homotopic.
-  1: exact equiv_path_grouphomomorphism^-1.
+  1: exact (equiv_path_grouphomomorphism^-1%equiv).
   1: exact _.
   intros []; reflexivity. 
 Defined.
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -22,12 +22,12 @@ Proof.
 Defined.
 
 (** It follows that any space of paths in a contractible space is contractible. *)
-(** Because [Contr] is a notation, and [Contr_internal] is the record, we need to iota expand to fool Coq's typeclass machinery into accepting supposedly "mismatched" contexts. *)
 
-Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000 := let c := {|
-  center := (contr x)^ @ contr y;
-  contr := path2_contr ((contr x)^ @ contr y)
-|} in c.
+Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000.
+Proof.
+  exists ((contr x)^ @ contr y).
+  exact (path2_contr ((contr x)^ @ contr y)).
+Defined.
 
 (** Also, the total space of any based path space is contractible.  We define the [contr] fields as separate definitions, so that we can give them [simpl nomatch] annotations. *)
 
@@ -102,3 +102,11 @@ Definition contr_retract {X Y : Type} `{Contr X}
            (r : X -> Y) (s : Y -> X) (h : forall y, r (s y) = y)
   : Contr Y
   := Build_Contr _ (r (center X)) (fun y => (ap r (contr _)) @ h _).
+
+(** Sometimes the easiest way to prove that a type is contractible doesn't produce the definitionally-simplest center.  (In particular, this can affect performance, as Coq spends a long time tracing through long proofs of contractibility to find the center.)  So we give a way to modify the center. *)
+Definition contr_change_center {A : Type} (a : A) `{Contr A}
+  : Contr A.
+Proof.
+  exists a.
+  intros; apply path_contr.
+Defined.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -172,17 +172,33 @@ Definition moveR_equiv_M `{IsEquiv A B f} (x : A) (y : B) (p : x = f^-1 y)
   : (f x = y)
   := ap f p @ eisretr f y.
 
+Definition moveR_equiv_M' `(f : A <~> B) (x : A) (y : B) (p : x = f^-1 y)
+  : (f x = y)
+  := moveR_equiv_M x y p.
+
 Definition moveL_equiv_M `{IsEquiv A B f} (x : A) (y : B) (p : f^-1 y = x)
   : (y = f x)
   := (eisretr f y)^ @ ap f p.
+
+Definition moveL_equiv_M' `(f : A <~> B) (x : A) (y : B) (p : f^-1 y = x)
+  : (y = f x)
+  := moveL_equiv_M x y p.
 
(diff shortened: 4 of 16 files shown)


mikeshulman:
(With the tweaks, the overall timing diff for this PR is negligible.)

Alizter:
Just going to check the CI is happy, then I will merge.

JasonGross:
You might be interested in doing `Hint Opaque contr_equiv : typeclass_instances` and/or `Argumemts contr_equiv : simpl never` (though this won't give you anything over `Opaque`).  Note also that you can get behavior between `Opaque` and `Transparent` with the `Strategy` command.

JasonGross:
> My only guess here is that somehow transparent and opaque constants behave differently with respect to insertion of implicit arguments.

This sounds like the wrong explanation.  I'd suspect there might be coercions involved (there's a recent Coq issue about coercions to funclass not inserting implicit arguments).  My next guess would be that it's inferring a surprising equivalence (like the identity equivalence) and hence having a thing that you expected to not be a function actually be a function.  Alternatively, opacity can influence what typeclasses can be resolved.  My final guess is that there are unification constraints caused by inserting evars for the implicit arguments, which can only be solved by unfolding the constant that got marked as opaque.

mikeshulman:
I have no intuition for what `Hint Opaque` means or in what cases I should use it.  Can you explain why you think it might be useful here?

I did see the `Strategy` command in the documentation, but my understanding of it from the documentation was that finite values of `Strategy` only differ in which of them is expanded _first_, whereas an `Opaque` constant will not be expanded _at all_; so that if the slow command is a `cbn` that's expanding everything eventually, a finite value of `Strategy` wouldn't help.  But maybe I am misunderstanding?

Thanks also for the suggestions for figuring out the implicit arguments.  I'll try to find the time to experiment with them.

JasonGross:
> I have no intuition for what `Hint Opaque` means or in what cases I should use it. Can you explain why you think it might be useful here?

`Hint Opaque foo : db` roughly means "treat `foo` as if it had been declared `Opaque` when doing resolution with `db`".  You can use `typeclass_instances` to specify that `foo` should be `Opaque` just for typeclass inference, or `rewrite` to specify that it should be `Opaque` for rewriting.  It might be useful if the performance issues are coming up in typeclass resolution due to transparency, and you think the constant should be transparent, but should never be automatically unfolded in typeclass resolution.  Another way of looking at it is that `Hint Opaque` is to typeclass resolution / rewrite / (e)auto what `simpl never` is to `cbn`/`simpl`.

> I did see the `Strategy` command in the documentation, but my understanding of it from the documentation was that finite values of `Strategy` only differ in which of them is expanded _first_, whereas an `Opaque` constant will not be expanded _at all_; so that if the slow command is a `cbn` that's expanding everything eventually, a finite value of `Strategy` wouldn't help. But maybe I am misunderstanding?

That's correct.  `Strategy` is only relevant for unification/conversion (which will both unfold even `Opaque` things, but only after unfolding everything else).

(Though note that order of unfolding can be relevant for `simpl` and `cbn`.  Consider simplifying `const5 (fact 100)`, where `const5` is the constant function returning 5 and `fact` is factorial.  There's a big difference between unfolding `const5` first vs unfolding `fact` first.)

PR #1281, opened 2020-02-27
removed unnecessary funext from AbelianGroup.v

Alizter:
I've removed some unnecessary Funext that was being used here. Unfortunately removing a section has caused some diff noise.

While this PR is open, I wouldn't mind adding some speed ups. I've tried to find some but it just seems to be slow. If anyone has any ideas how we can speed this up I am listening.

### DIFF
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -232,8 +232,8 @@ Section AbelGroup.
     revert x.
     Abel_ind_hprop x.
     cbn.
-    rewrite <- (left_identity (x * y)).
-    rewrite <- (left_identity (y * x)).
+    refine ((ap ab (left_identity _))^ @ _).
+    refine (_ @ (ap ab (left_identity _))).
     apply ab_comm.
   Defined.
 
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -151,7 +151,7 @@ Section Abel.
 
 End Abel.
 
-(** The first argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but the second usually cannot, so this tactic makes the common case fast. *)
+(** The [IsHProp] argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but [srapply] is slow, so we use this tactic instead. *)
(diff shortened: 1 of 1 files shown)


jdchristensen:
I was just about to submit a PR with speedups to this file.  But I can't easily see what you've changed because of the indentation.  Can you force push a revised commit without changing the indentation?  Also, can you explain why you reordered things in abel_sgop?

I think I'll submit a separate PR so you can see what I have done.  It should be easy to apply my changes to your version, and then my PR can be closed without merging.

Alizter:
@jdchristensen Ill wait until your PR is merged, then rebase on top of that.

mikeshulman:
Even when we remove or add a section, we should restrain ourselves from reindenting in the same PR that makes the substantive changes.

Alizter:
OK, i've reopened this and removed the indentation so you can better see what has been done.

Alizter:
OK I have rebased this on top of #1284

theories/Algebra/AbelianGroup.v
```diff
@@ -226,8 +227,10 @@ Section AbelGroup.
...
   Global Instance abel_commutative : Commutative abel_sgop.
   Proof.
-    Abel_ind_hprop x.
+    intro x.
     Abel_ind_hprop y.
+    revert x.
+    Abel_ind_hprop x.
     cbn.
     rewrite <- (left_identity (x * y)).
     rewrite <- (left_identity (y * x)).
```
jdchristensen:
```suggestion
    refine ((ap ab (left_identity _))^ @ _).
    refine (_ @ (ap ab (left_identity _))).
```
Not really related, but I noticed that these rewrites are easy to get rid of.  There are only two more in the file which are probably also easy to get rid of.

Alizter:
@jdchristensen This is a technique I learnt from the lean library. When they were defining the H-space structure on S1 they did a similar "intro before eliminating" so that they could define multiplication on S1 without funext. It means we use funext unnecessarily in a lot of places. This being one of them.

mikeshulman:
Of course, this sort of thing is also the reason why HITs imply funext.  So one might argue that instead of "removing" funext we're just factoring through the proof that HITs imply funext.  But I agree that it's nicer.

PR #1284, opened 2020-02-27
Speed improvements to AbelianGroup.v:

jdchristensen:
Replace the last five lines of `Abel_ind_beta_ab_comm` with `apply concat_V_pp`.

Get rid of `Section Abelianization`, since closing it was a bit slow.

Speed up use of `Abel_ind_hprop` with a tactic of the same name.  Adds up to a big savings.

### DIFF
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -151,7 +151,7 @@ Section Abel.
 
 End Abel.
 
-(** The first argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but the second usually cannot, so this tactic makes the common case fast. *)
+(** The [IsHProp] argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but [srapply] is slow, so we use this tactic instead. *)
 Local Ltac Abel_ind_hprop x := snrapply Abel_ind_hprop; [exact _ | intro x].
 
 (** We make sure that G is implicit in the arguments of ab and ab_comm. *)
@@ -288,8 +288,6 @@ Defined.
 
 (** Now we finally check that our definition of abelianization satisfies the universal property of being an abelianization. *)
 
-(* Section Abelianization.  Removed as closing the section is slow. *)
-
   (** We define abel to be the abelianization of a group. This is a map from Group to AbGroup. *)
   Definition abel `{Funext} : Group -> AbGroup.
   Proof.
@@ -330,8 +328,6 @@ Defined.
     Abel_ind_hprop x.
     apply p.
   Defined.
-
-(* End Abelianization. *)
 
 Theorem groupiso_isabelianization {G : Group}
   (A B : AbGroup)
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -120,11 +120,7 @@ Section Abel.
     apply dp_apD_path_transport.
     rewrite (apD_compose' tr).
     rewrite (Coeq_ind_beta_cglue _ _ _ (x, y, z)).
-    unfold Abel_ind.
-    refine (_ @ concat_1p _).
-    refine (concat_p_pp _ _ _ @ _).
-    apply whiskerR.
-    apply concat_Vp.
+    apply concat_V_pp.
   Defined.
 
   (** We also have a recursion princple. *)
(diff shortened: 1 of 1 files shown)


Alizter:
When we have a goal like `forall x y, P x y` where we know that `forall x y, IsHProp (P x y)` and we need to use something like `Abel_ind_hprop`, we can avoid funext by intoducing `x` apply `Abel_ind_hprop` and then reverting x then applying `Abel_ind_hprop` again.

theories/Algebra/AbelianGroup.v
```diff
@@ -155,6 +151,9 @@ Section Abel.
 
 End Abel.
 
+(** The first argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but the second usually cannot, so this tactic makes the common case fast. *)
```
mikeshulman: When you say the "first argument", do you mean the `IsHProp` argument (which is actually the second argument)?  It seems like the third argument `forall x, P (ab x)` is not usually going to belong a typeclass, so there shouldn't be any danger of typeclass inference spinning on it.  Is it just the cases when it does happen to belong to a typeclass that are slow?

theories/Algebra/AbelianGroup.v
```diff
@@ -281,26 +280,25 @@ End AbelGroup.
...
+  cbn.
   exists (tr (x; @idpath _ (ab x))).
   apply path_ishprop.
 Defined.
 
 (** Now we finally check that our definition of abelianization satisfies the universal property of being an abelianization. *)
-Section Abelianization.
 
-  Context `{Funext}.
+(* Section Abelianization.  Removed as closing the section is slow. *)
```
mikeshulman: What is it that makes closing a section slow?  Aren't there plenty of other sections in other files that aren't slow?

theories/Algebra/AbelianGroup.v
jdchristensen: Right, I meant the second argument.  I don't understand why `srapply` is slow, but it is in many, many cases in this file.  I'll rephrase the comment to:
```
(** The [IsHProp] argument of [Abel_ind_hprop] can usually be found by typeclass resolution, but [srapply] is slow, so we use this tactic instead. *)
```

theories/Algebra/AbelianGroup.v
jdchristensen: I can't figure it out.  Even with all context variables removed, closing that section is a bit slow.  (Only 0.2-0.3s, so not the end of the world.)  This is based on timing compilation of the whole file, so it isn't just that the time from Section closing gets shifted somewhere else.  Other Sections are quick to close.

theories/Algebra/AbelianGroup.v
mikeshulman: Got it.  Now I think I remember seeing that sort of behavior in other cases.  My guess was that `snrapply` manages to finish some unification that makes the later `exact _` fast at the typeclass resolution, whereas `srapply` may be trying to do the typeclass resolution before all the unification is done which makes it slower.

theories/Algebra/AbelianGroup.v
mikeshulman: Weird.

mikeshulman: Once the section is removed, there's no need to keep a comment about it in the file, is there?  Unless we're worried about someone later on deciding to factor out the `Funext` hypothesis into a section variable without realizing that it slows down the file?

jdchristensen: The sectioning in this file is mainly to help the reader understand the structure, so there is sort of a moral section here.  But you are probably right that we should instead just have an ordinary comment.  I'll update that.  Then I'll push a final commit that just undoes the indentation.

theories/Algebra/AbelianGroup.v
Alizter: I think the problem is that srapply is trying typeclass search on less arguments causing it to think more. If you use srapply' it would probably be faster.

jdchristensen: I tried srapply', and it took 4s and then failed with an error.  But if I try `simple refine` with exactly the right number of arguments, it is as fast as the tactic.  But `simple refine` with one too few arguments takes a bit of time to fail, so that explains why srapply is slow.  Mystery solved!

Alizter: So ideally we would have a very smart `srapply` that could guess how many arguments it needs from the beginning, without having to try with fewer arguments.

PR #1285, opened 2020-02-28
Remove funext entirely from AbelianGroup by using wild categories

mikeshulman:
It's even faster.

### DIFF
--- a/theories/Algebra/AbelianGroup.v
+++ b/theories/Algebra/AbelianGroup.v
@@ -4,6 +4,7 @@ Require Import HIT.Coeq.
 Require Import Algebra.Group.
 Require Import Algebra.Subgroup.
 Require Import Cubical.
+Require Import WildCat.
 
 Local Open Scope mc_mult_scope.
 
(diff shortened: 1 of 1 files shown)


theories/Algebra/AbelianGroup.v
```diff
@@ -308,75 +314,63 @@ Proof.
...
 (** Finally we can prove that our construction abel is an abelianization. *)
-Global Instance isabelianization_abel `{Funext} {G : Group}
+Global Instance isabelianization_abel {G : Group}
   : IsAbelianization (abel G) (abel_unit G).
 Proof.
-  intros A h.
-  srapply Build_Contr.
-  { srefine (_;_).
+  intros A; constructor.
+  { intros h; srefine (_;_).
```
jdchristensen:
```suggestion
  intros A. constructor.
  { intros h. srefine (_;_).
```
Since wildcats are fairly new, it's helpful to have a few more pauses when reading this.

theories/Algebra/AbelianGroup.v
```diff
@@ -386,18 +380,15 @@ Proof.
...
-Global Instance isabelianization_identity `{Funext} (A : AbGroup) : IsAbelianization A grp_homo_id.
+Global Instance isabelianization_identity (A : AbGroup) : IsAbelianization A grp_homo_id.
 Proof.
-  unfold IsAbelianization.
-  intros B h.
-  apply (Build_Contr _ (h; fun _ => idpath)).
-  intros [g p].
-  apply path_sigma_hprop; cbn.
-  by apply equiv_path_grouphomomorphism.
+  intros B; constructor.
```
jdchristensen:
```suggestion
  intros B. constructor.
```

theories/Algebra/AbelianGroup.v
```diff
@@ -308,75 +314,63 @@ Proof.
...
         refine (grp_homo_op _ _ _ @ _ @ (grp_homo_op _ _ _)^).
         apply (ap (_ *.)).
         refine (grp_homo_op _ _ _ @ _ @ (grp_homo_op _ _ _)^).
         apply commutativity. }
-      Abel_ind_hprop x.
+      intros y.
+      Abel_ind_hprop x; revert y.
       Abel_ind_hprop y.
       apply grp_homo_op. }
     cbn; reflexivity. }
```
jdchristensen:
```suggestion
    cbn. reflexivity. }
```
(You didn't change this line, but earlier changes make it less obvious what's happening here.)

theories/Algebra/AbelianGroup.v
```diff
@@ -386,18 +380,15 @@ Proof.
     (conn_map_homotopic _ _ _ p _)).
 Qed.
```
jdchristensen:
```suggestion
Defined.
```
Unrelated, but I see no reason for this to be Qed.

jdchristensen:
Really what's going on here is that abelian groups form a reflective subcategory of groups, and that by formal nonsense, all sorts of things are automatically true.  In this case, the subcategory is obtained by localizing with respect to the map from the free group on two generators to the free abelian group on two generators.  How hard would it be to prove these sorts of results in general?  Could a general framework even subsume the work on reflective subuniverses?

mikeshulman:
Right.  Results like `groupiso_isabelianization`, `homotopic_isabelianization`, `isabelianization_identity`, and `isequiv_abgroup_abelianization`, could certainly be proven right now about reflections into any subcategory of a wild 1-category, probably with almost the same proofs.

With the current definition of wild 1-category, reflections of this sort wouldn't quite subsume reflective subuniverses, because our reflective subuniverses are defined to have a completely funext-free universal property by eliminating into arbitrarily iterated path-types of hom-types, whereas our current wild 1-categories don't have the data necessary to even state such a thing.  (This distinction doesn't matter for groups, of course, since they're actually a "tame" 1-category with hom-sets rather than hom-types.)  But the `wildcat` branch in my fork does contain some experiments with a notion of "oo-category" that is equipped (coinductively) with cells at all dimension (but only a finite amount of coherence), so if we moved in that direction it's possible that we could subsume reflective subuniverses into reflective subcategories.  However, I'd still be a little worried about how easy or hard it would be to translate back and forth; the fact that I needed the `group_precomp` hack here suggests to me that the structure of the wild categories library is currently suboptimal (though I don't have any improvements in mind yet), and this might be problematic for the usability of reflective subuniverses and modalities in the rest of the library.

jdchristensen:
Even the existence of abelianization should follow from some general framework.

jdchristensen:
I was thinking in analogy with the locally presentable case classically, but didn't have anything precise in mind.  At least one level of generality up would be to show that just for the category of groups, one can localize with respect to families of maps.  The key tool both for groups and for Type is the existence of HITs, but I don't know if there's a way to abstract that.

mikeshulman:
One possibility would be to work at the level of universal algebra.  We already have some universal algebra developed in the mathclasses library, and I expect with HITs we could show that the category of algebras for any algebraic theory can be localized at any small family of maps.  That's not as general as locally-presentable even classically, but it's substantially more general than "groups and sets".

Another possibility would be to work with "limit-sketchable categories", which classically coincide with locally-presentable ones, but might be more concretely tractable constructively.

And of course there's the question of how wild the categories can be.

Issue #1272, opened 2020-02-23
Move useful lemmas out of EquivalenceVarieties to Types and discourage its use


Alizter:
All three `Varieties` files are a tricky problem, as on the one hand they contain a few useful lemmas -- e.g. just as `FunextVarieties` contains `contr_basedhomotopy`, `EquivalenceVarieties` contains `fcontr_isequiv` and `isequiv_fcontr`.   But on the other hand, a lot of their time is spent on "dead-end" results that are mainly interesting theoretically and aren't ever going to get used for anything in the rest of the library.  I guess one solution would be to move the useful lemmas into other files so that the Varieties files themselves never need to be imported elsewhere.

_Originally posted by @mikeshulman in https://github.com/HoTT/HoTT/pull/1271_

Alizter:
We shouldn't have any `Require Import EquivalenceVarieties` in the library. If we do, that means the lemma's used are note so "dead-end" and thus should be moved into Types/Equivalences.v or somewhere similar.

mikeshulman:
Maybe once we do this, we can move the Varieties files into a "dead-end subdirectory" to make it clear they shouldn't be imported.

Alizter:
Are you thinking somewhere like Contrib?

mikeshulman:
No, just maybe something like `theories/Metatheory`.

PR #1286, opened 2020-03-01
Split up EquivalenceVarieties

mikeshulman:
This is my proposed resolution of #1272.  However, if anyone has different opinions, please share them!

1. The majority of imports of `EquivalenceVarieties` are for the facts about contractible fibers.  I feel like these are really basic facts about equivalences in HoTT, and so one shouldn't have to import more than `Basics` and `Types` to get them.  So I moved them into `Types/Equiv`.

2. I realized on looking closer that the "different proofs" of hprop_isequiv in `EquivalenceVarieties` were really not that different; the proof in `Types/Equiv` was really just proving the lemmas internally.  So I removed them.  Since the lemmas in question were just moved into `Types/Equiv`, they can be used in `hprop_isequiv` there, so in a sense this is "merging the proofs together".

3. I had to move `hprop_trunc` from `HProp` (which depends on `Types`) into `Basics`.  But I think this is good anyway, since it's also a very basic fact that no one should have to remember where to find.  I had to modify the proof only slightly to avoid using facts from `Types`.

4. I also realized that instead of `forall b, Contr (hfiber f b)` we should really say just `IsTruncMap (-2) f`.  Making this an `Instance` also allows typeclass inference to solve a few more goals for us.

4. Finally, the remaining imports of `EquivalenceVarieties` really are using one of the other notions of equivalence (aside from one that was using `BiInv` unnecessarily instead of simply adjointifying), so those have to remain.  I don't think `BiInv`, `PathSplit`, and relational equivalences are important enough to put in `Basics` or `Types`.  But to make the "meaning of the imports" clearer, I suggest putting them in separate files in an `Equiv/` directory.

We could consider putting `Extensions` in `Equiv/` as well, since it's also a "notion of equivalence", albeit one that only applies to precomposition maps.  Opinions are welcome.


Alizter:
I agree with all of your points. I'm not sure about moving extensions just yet however.

Something that I was pondering last week: can `hprop_trunc`, or more specifically `contr_contr` be proven without funext? I am guessing no. But is `contr_contr` strong enough to prove funext?

mikeshulman:
I doubt that `contr_contr` can be proved without funext, but I would be surprised if it were strong enough to prove funext.

mikeshulman:
Specifically, `contr_contr` only tells you something about pi-types whose codomain is a path-type.  I don't see any way to deduce something from that about pi-types with arbitrary codomain.

theories/Types/Equiv.v
```diff
@@ -10,27 +10,84 @@ Local Open Scope path_scope.
...
+    apply equiv_functor_forall_id; intros a.
+    refine (equiv_concat_l (eisadj f a) _ oE _).
+    rapply equiv_ap.
   Qed.
 
+  (** Now since [IsEquiv f] and the assertion that its fibers are contractible are both HProps, logical equivalence implies equivalence. *)
+  Definition equiv_contr_map_isequiv {A B} (f : A -> B)
+    : IsTruncMap (-2) f <~> IsEquiv f.
+  Proof.
+    rapply equiv_iff_hprop.
```
Alizter: Nice proof!

theories/Basics/Trunc.v
```diff
@@ -400,4 +400,33 @@ Definition equiv_iff_hprop `{IsHProp A} `{IsHProp B}
   : (A -> B) -> (B -> A) -> (A <~> B)
   := fun f g => equiv_iff_hprop_uncurried (f, g).
 
+(** Truncatedness is an hprop. *)
+Global Instance hprop_trunc `{Funext} (n : trunc_index) (A : Type)
```
jdchristensen: Here the instance is called hprop_* and below it is called isprop_*.  Shouldn't they both be ishprop_*?  Or at least consistent with each other?

theories/Basics/Trunc.v
```diff
@@ -400,4 +400,33 @@ Definition equiv_iff_hprop `{IsHProp A} `{IsHProp B}
...
+    apply path_forall; intro x.
+    apply path_forall; intro y.
+    apply @path_contr.
+    apply IH, AH1.
+Qed.
+
+(** By [trunc_hprop], it follows that [IsTrunc n A] is also [m]-truncated for any [m >= -1]. *)
+
+(** Similarly, a map being truncated is also a proposition. *)
+Global Instance isprop_istruncmap `{Funext} (n : trunc_index) {X Y : Type} (f : X -> Y)
```
jdchristensen: This is the other spot I was referring to.

theories/Basics/Trunc.v
mikeshulman: Yes, they should.  This is a leftover from old less-consistent naming; I was just moving things around without thinking about renaming them.  But you're right, we might as well rename them at the same time.

PR #1293, opened 2020-03-10
moved Conjugation.v into contrib/old

### LINKED ISSUE #1292: Remove Conjugation.v
There is a file in theories called Conjugation.v. I had a look inside and it contains very few things. Shall we move this to contrib/old or something?

Alizter:
closes #1292


Alizter:
Whoops forgot to add this to the commit. Should be good now.

PR #1297, opened 2020-03-11
Added development of sequential colimits.

kristinas:


Alizter:
So I've added some commits to this PR. The first one is mostly changing Let statements into Local Definition statements. This stops the terms from hanging around in the context in the proofs below. This was causing a lot of speed issues.

I had to also open nat_scope at the very top to get this to work on my machine.

The second commit is just about making the imports more specific. Some of them were a bit ambiguous which could possibly cause coq some trouble in the future.

I think this highlights that some of the colimits files should be exporting files like diagrams by default, since they are very commonly imported together.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+Require Import Diagrams.Sequence.
+Require Import Diagrams.DDiagram.
+Require Import Diagrams.Cocone.
+Require Import Colimits.Colimit.
+Require Import Spaces.Nat.
+
+Local Open Scope nat_scope.
+Local Open Scope path_scope.
+
+Definition functor_forall_equiv_pb {A B : Type} {P : B -> Type} (f : A -> B) `{IsEquiv _ _ f}
```
Alizter: We have this. It is in Types/Forall.v under the name equiv_functor_forall_pb. It is the inverse of that equivalence.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+  : (forall a, P (f a)) -> (forall b, P b)
+  := fun h b => eisretr f b # h (f^-1 b).
+
+Definition functor_forall_equiv_pb_beta {A B : Type} {P : B -> Type} (f : A -> B)
```
Alizter: We don't have this in Types/Forall but we could adapt it into a lemma about equiv_functor_forall_pb

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+Defined.
+
+Record TypeSeq := {
+  typeSeq : nat -> Type;
+  typeSeqArr n : typeSeq n -> typeSeq n.+1
+}.
+
+Coercion typeSeq : TypeSeq >-> Funclass.
+
+Definition TypeSeqToDiagram : TypeSeq -> Diagram sequence_graph.
```
Alizter: This is basically an uncurried version of `Build_Sequence` which can be found in Diagrams/Sequence. `Sequence` is just defined to `Diagram sequence_graph`.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+  Proof.
+    srapply seq_colimit_uniqueness.
+    - exact (fun n a => 1).
+    - intros n [a b]; rewrite concat_1p, concat_p1, ap_compose, !SeqColimit_rec_beta_glue.
+      rewrite ap_pp, (Delta_proj _ (fibTypeSeqToTypeFam_beta n a)).
+      srapply (whiskerL _ _ @ concat_p1 _); rewrite (ap_compose _ _ _)^; simpl.
+      rewrite ap_const; exact 1.
+  Defined.
+
+End FibTypeSeq.
```
Alizter: Typically its best to end the file on a new line. You can directly commit this suggestion on github.
```suggestion
End FibTypeSeq.

```

Alizter:
I take it the corollary about path types commuting with sequential colimits isn't finished yet?

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+
+Local Open Scope nat_scope.
+Local Open Scope path_scope.
+
+Definition functor_forall_equiv_pb {A B : Type} {P : B -> Type} (f : A -> B) `{IsEquiv _ _ f}
+  : (forall a, P (f a)) -> (forall b, P b)
+  := fun h b => eisretr f b # h (f^-1 b).
+
+Definition functor_forall_equiv_pb_beta {A B : Type} {P : B -> Type} (f : A -> B)
+  `{IsEquiv _ _ f} (h : forall a, P (f a))
```
Alizter: If you want to avoid these extra arguments when writing generalized implicit arguments use the `!` before a name. This is one of those poorly documented coq features :-)
```suggestion
  `{!IsEquiv f} (h : forall a, P (f a))
```

mikeshulman: It's important to recognize, though, that that's not the primary purpose of `!`.  Rather, what it does is prevent implicit generalization.  This has the byproduct of allowing Coq to revert back to respecting the implicitness of arguments.

mikeshulman: It's important to recognize, though, that that's not the primary purpose of !. Rather, what it does is prevent implicit generalization. This has the byproduct of allowing Coq to revert back to respecting the implicitness of arguments.

theories/Colimits/SeqColimits.v
mikeshulman: Yes, we should do that.

mikeshulman:
I think a good rule of thumb would be that if you can't imagine making practical use of file `B` without file `A`, then `B` should export `A`.  For instance, `Modality` exports `ReflectiveSubuniverse` because so many of the theorems about modalities are actually theorems about reflective subuniverses.  It's hard to imagine talking about colimits without needing to talk about diagrams too, so I agree that it seems reasonable for colimit files to export diagrams.

mikeshulman:
Since this file is already in the `Colimits` directory, it seems a bit rendundant to call it `Colimits/SeqColimits`; how about `Colimits/Sequential`?

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+  Definition seq_colimit_sum_uniqueness E (F G : sig fibTypeSeqToTypeFam -> E)
+    : F o seqColimSumToSumSeqColim == G o seqColimSumToSumSeqColim -> F == G.
+  Proof.
+    intro h; srapply (SeqColimitSum_ind _ (fun _ _ _ => h _)); intros n a b.
+    srapply ((transport_compose _ _ _ _)^ @ _); exact (apD h (glue sigTypeSeq n (a;b))).
+  Defined.
+
+  (** The canonical map from the sequential colimit of Sigmas to the Sigma of
+      sequential colimits is an equivalence. *)
+  Definition seqColimSumToSumSeqColim_equiv : IsEquiv seqColimSumToSumSeqColim.
```
mikeshulman: Our standard naming convention for `IsEquiv` instances would name this `isequiv_seqColimSumToSumSeqColim`.  Do you want to make it an `Instance`?  It might also be helpful to have it bundled up into an `Equiv`.

theories/Colimits/SeqColimits.v
Alizter: Since there is literally nothing here to generalize, I suppose `{_ : IsEquiv f}` is the most correct thing to write.

mikeshulman: I don't think anything is more correct than anything else.  It's fine (IMHO) to use the backtick to avoid having to name the argument (even as an underscore), and it's fine (IMHO) to use the bang to avoid having to name the implicit arguments with underscores.  I just wanted to make sure any reader who isn't already familiar with `!` knows that it does _more_ than just allow omitting the ordinarily-implicit arguments.

kristinas:
> So I've added some commits to this PR. The first one is mostly changing Let statements into Local Definition statements. This stops the terms from hanging around in the context in the proofs below. This was causing a lot of speed issues.
> 
> I had to also open nat_scope at the very top to get this to work on my machine.
> 
> The second commit is just about making the imports more specific. Some of them were a bit ambiguous which could possibly cause coq some trouble in the future.
> 
> I think this highlights that some of the colimits files should be exporting files like diagrams by default, since they are very commonly imported together.

I get no error without nat_scope ....

theories/Colimits/SeqColimits.v
kristinas: Yes but the lemma about this definition requires funext in the library. In general I don't like these gratuitous uses of funext at a lot of places.

Alizter:
@kristinas  My guess is that you have an older compiled version of Nat sitting around. It used to open nat_scope by default, now we have to manually do this.

theories/Colimits/SeqColimits.v
kristinas: Yes, I did this intentionally because now maybe Sequence.v can be merged with my file.

Alizter: I think that is a good idea. I might be inclined to move your results about sequences into Diagrams/Sequence and keep the sequantial colimits stuff here though.

mikeshulman: Do you get a lot of benefit from working with `TypeSeq` instead of directly with `Sequence`?

theories/Colimits/SeqColimits.v
mikeshulman: The use of funext is necessary at the place where it appears in `Types/Forall` because what's being constructed there is an equivalence between Pi-types.  Your point is that the maps involved in this equivalence can be defined without funext, which is also reasonable.  I suggest simply moving this lemma to `Types/Forall`.

mikeshulman:
> @kristinas Did you do a `git pull origin master`? This will update the version on your computer to the online version (which is ahead).

This is another reason why having people other than the author pushing to a PR is dangerous...

kristinas:
I deleted my entire fork and forked the repo again, and recompiled everything so I am not sure why I would be getting the nat scope issue.

kristinas:
Not so much actually. I could possibly just get rid of it.

theories/Colimits/SeqColimits.v
kristinas: Not so much actually. I could possibly just get rid of it.

kristinas:
> I take it the corollary about path types commuting with sequential colimits isn't finished yet?

That's right, that would be the next step.

mikeshulman:
Looks like `Sequential.v` still needs to be added to `_CoqProject`.

@kristinas, would you prefer that we make fixes like this ourselves and push them to the PR branch (which we can do because you left checked the box that says "allow updates from maintainers"), or would you rather make them yourself (so that e.g. your local copy won't get out of sync with your fork)?

mikeshulman:
Also, it looks like right now you have both `SeqColimits` and `Sequential`; you need to `git rm` the first one.

theories/Colimits/Sequential.v
```diff
@@ -0,0 +1,518 @@
...
+
+    (** The point case of the nested induction. *)
+    Local Definition F n a : forall x, E (inj _ n a; x)
+      := SeqColimit_ind _ _ (fun k => Q k n a) (fun k => R k n a).
+
+    (** The path case of the nested induction. *)
+    Local Definition G n a y : F n a _ = Delta _ (fibTypeSeqToTypeFam_beta n a) y # F n.+1 (f n a) y.
+    Proof.
+      revert y; srapply (SeqColimit_ind _ _ (fun k => Q_beta k n a)); intros k b.
+      srapply (Phi (glue A n a) (fib_seq_lift_colim_equiv_beta _ _ _)).
```
mikeshulman:
```suggestion
      snrapply (Phi (glue A n a) (fib_seq_lift_colim_equiv_beta _ _ _)).
```
shaves off a fraction of a second.

theories/Colimits/Sequential.v
```diff
@@ -0,0 +1,518 @@
...
+      rewrite (SeqColimit_ind_beta_glue _ _ _ (fun k : nat => R k n.+1 (f n a)) k _).
+      rewrite (SeqColimit_ind_beta_glue _ _ _ (fun k : nat => R k n a) k.+1 _).
+      rewrite (apD_compose' _ _ (Beta (pair_lift_assoc (n; a) k) b)).
+      srapply functor_forall_equiv_pb_beta.
```
mikeshulman:
```suggestion
      snrapply functor_forall_equiv_pb_beta.
```
again, shaves off a fraction of a second.

theories/Colimits/Sequential.v
```diff
@@ -0,0 +1,518 @@
...
+    Proof.
+      induction k as [ | k h].
+      - intros n a b; srapply (_ @ t n a b).
+        srapply (Eta (seqColimSumToSumSeqColim_beta_glue n a b)).
+      - intros n a; apply (functor_forall_equiv_pb (coe (ap B (pair_lift_assoc _ k)))).
+        intro b; srapply (_ @ (Q_beta k n a b)^).
+        srapply (_ @ ap _ (h n.+1 (f n a) b)).
+        srapply (_ @ (Epsilon (glue A n a) (fib_seq_lift_colim_equiv_beta _ _ _))).
+        srapply (_ @ ap _ (ap _ (Q_beta _ n a _))).
+        srapply (ap _ ((transport_compose _ _ _ _)^ @ _)^).
```
mikeshulman:
```suggestion
        srefine (ap _ ((transport_compose _ _ _ _)^ @ _)^).
```
I would use `srefine` rather than `srapply`, here and on the previous four lines.  `srapply` works, but when the term-with-holes being given has the goal type itself, I think it's clearer for the reader to use a form of `refine` (or `exact`) rather than a form of `apply` since the latter suggests that it's being applied to some additional arguments.

theories/Colimits/Sequential.v
Alizter: The explanation for this is that snrapply is a version of srapply that doesn't perform typeclass search (i.e. doesn't go looking for instances).

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+  - exact e.
+Defined.
+
+Section TypeSeq.
+
+  Context {A : TypeSeq}.
+
+  Local Definition f := typeSeqArr A.
+
+  Definition pair_lift (x : sig A) : sig A := match x with (n;a) => (n.+1; f n a) end.
```
mikeshulman: Maybe this and the other definitions in this section should either be `Local` or be given more globally-unique names referring to type sequences?

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+Section SeqColimitDef.
+
+  Context (A : TypeSeq).
+
+  Let f := typeSeqArr A.
+  Let S' := TypeSeqToDiagram A.
+
+  Definition SeqColimit : Type := Colimit S'.
+
+  Definition inj n (a : A n) : SeqColimit := @colim _ S' n a.
```
mikeshulman: Similarly here, probably `inj` and `glue` should either be `Local` or have more globally-unique names.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+  Proof.
+    srapply SeqColimit_ind.
+    - exact h.
+    - intros n a; generalize (H n a); generalize (h n a); destruct (glue n a).
+      intros p q; srapply ((concat_p1 _)^ @ _); srapply (_ @ (concat_1p _)); exact q^.
+  Defined.
+
+End SeqColimitDef.
+
+Definition seq_colimit_succ_equiv A : SeqColimit (succTypeSeq A) <~> SeqColimit A.
```
mikeshulman: According to general naming conventions this should probably be `equiv_seq_colimit_succ`.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+Defined.
+
+Definition functor_seq_colimit {A} {B} (M : TypeSeqMap A B) : SeqColimit A -> SeqColimit B.
+Proof.
+  destruct M as [h H]; srapply SeqColimit_rec.
+  - exact (fun n a => inj _ n (h n a)).
+  - exact (fun n a => ap (inj _ n.+1) (H n a) @ glue _ n (h n a)).
+Defined.
+
+Global Instance functor_seq_colimit_equiv `{Funext} {A} {B} (M : TypeSeqEquiv A B) :
```
mikeshulman: General naming convention would be `isequiv_functor_seq_colimit`.

theories/Colimits/SeqColimits.v
```diff
@@ -0,0 +1,517 @@
...
+    : TypeSeqEquiv (fibSeqToTypeSeq x^+) (succTypeSeq (fibSeqToTypeSeq x)).
+  Proof.
+    srapply Build_TypeSeqEquiv.
+    - srapply Build_TypeSeqMap.
+      * exact (fun k => coe (ap _ (pair_lift_assoc x k))).
+      * exact (fun k b => Beta (pair_lift_assoc x k) b). 
+    - intro k; srapply isequiv_path.
+  Defined.
+
+  Definition fib_seq_lift_colim_equiv (x : sig A)
```
mikeshulman: Naming convention would be `equiv_` something (I'm not sure what `fib_seq_lift_colim` refers to), and so on for other lemmas.

mikeshulman:
This looks great overall; thanks for the contribution!  I'm looking forward to the corollary about path-types too.

kristinas:
Thanks Mike! I decided to incorporate your suggestion and get rid of the TypeSequence record altogether. I am almost done with this modification, will submit the updated thing tomorrow.

theories/Colimits/SeqColimits.v
kristinas: Changed the names for this and all the other ones,

theories/Colimits/SeqColimits.v
kristinas: Made them into Local Definitions.

theories/Colimits/SeqColimits.v
kristinas: The two lemmas are still there, you can move them if you like.

kristinas:
One thing I wasn't able to resolve is the nat_scope issue. If I open it, I start having problems with the reflexivity path 1 conflicting with the numeral 1. I still do not understand why it doesn't work for Ali.

Alizter:
Never mind about the nat_scope issue. It looks like the CI is happy with your code so it must have been an issue on my end.

Alizter:
I'm mostly happy with this. I have a few things left:
 * Remove `functor_forall_equiv_pb` and replace it with the inverse of the equivalence `equiv_functor_forall_pb` where ever it is used.
 * Rename `functor_forall_equiv_pb_beta` to `equiv_functor_forall_pb_inverse_beta` and move it into Types/Forall.v. As Mike said, many things in that file already require funext so don't hesitate to use it in that file.

kristinas:
Well, using the inverse of the equivalence requires funext.

mikeshulman:
I opened a PR in your fork that implements my suggestion, moving `functor_forall_equiv_pb` to `Types/Forall`.

PR #1302, opened 2020-03-13
Basics of (oo,n)-categories

mikeshulman:
Here's my next iteration of coinductive wild oo-categories.

Change # 1: By getting rid of (oo,oo)-categories and considering only (oo,n)-categories for finite n, the proof that equivalences compose becomes much simpler, and little if any interesting generality is lost -- the only true (oo,oo)-categories that I know of are things like the (oo,oo)-category of (oo,oo)-categories, and at the moment I don't think that even has a "wild" version since the amount of coherence required to even _define_ a k-transfor seems to go up with k.

Change # 2: I simplified the notion of `HasEquivs` to include only the property `CatIsEquiv`, with a bundled version `CatEquiv` defined uniformly for all categories.  I think this is easier to read and work with; the only drawback is that at present, `@CatEquiv Type` doesn't coincide with `Equiv`.  However, I think that can be finessed, either by bringing the basic category definitions into `Basics`, or by using a general notion of `subtype` for both of them.  For examples other than `Type`, such as `pType`, there's no problem since they are defined later in the library and can simply use `CatEquiv` explicitly instead of whatever manual definition they already had.

Change # 3: There is now a "default" coherent definition of `HasEquivs` using bi-invertibility that can be used for any category if desired, thus simplifying the process of defining a category.  But I kept the option for users of defining a different one, so that if it's desired to use half-adjoint equivalences (e.g. for `Type`) that can be done.  (I tried making the default definition use half-adjoint equivalences, but that was _much_ harder than bi-invertibility since you need associativity and unit isomorphisms to even _state_ the adjoint coherence law.)

Change # 4: I defined dependent inserters directly rather than in terms of comma categories.  I think this way is clearer, partly because it doesn't require product categories.  We can then define comma categories in terms of them (but I didn't yet).

Change # 5: I added some results about truncated categories, including a general notion of "isomorphism" for (1,1)-categories that could be plugged into (for instance) groups, etc. rather than separately defining `GroupIsomorphism` and so on in each case separately.

Overall I'm trying to figure out whether this development can be made "lightweight" and comprehensible enough to be "the" categories library we use everywhere, replacing the current truncated-only definitions in `WildCat`.  An aspect of that would be to make it possible for users to use only finitely many dimensions, rather than being forced to work with infinitely many in all cases.  It's already possibly to talk about categories with only finitely many dimensions of cells, just by sticking in Unit at the top of the coinductive notion.  But there isn't yet a way to talk about _functors_ that only act on finitely many dimensions if their domain and codomain categories were defined to include infinitely many.  I think it should be possible to do this by parametrizing functors by a "coherence dimension", but I haven't tried yet.

I'm interested in any feedback, but particularly about the feasibility of this idea.  Could this actually be usable by anyone?  Or is it unavoidably just too scary-looking and heavyweight?

I do think this approach would have some practical advantages.  For instance, having the higher morphisms and equivalences around from the beginning deals more cleanly with a definition of 2-category than our previous version that required explicitly passing to cores on hom-categories.  And it allows us to formulate general notions of universal property (e.g. reflective subcategories) that specialize in particular cases to our funext-free versions (e.g. reflective subuniverses).

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -169,7 +169,7 @@ Notation "x .2" := (pr2 x) : fibration_scope.
 
 Definition uncurry {A B C} (f : A -> B -> C) (p : A * B) : C := f (fst p) (snd p).
 
-(** Composition of functions. *)
+(** *** Composition of functions. *)
 
 Notation compose := (fun g f x => g (f x)).
 
@@ -694,6 +694,8 @@ Class Irreflexive {A} (R : Relation A) :=
 Class Asymmetric {A} (R : Relation A) :=
   asymmetry : forall {x y}, R x y -> (complement R y x : Type).
 
+(** *** Unit type *)
+
 (** Likewise, we put [Unit] here, instead of in [Unit.v], because [Trunc] uses it. *)
 Inductive Unit : Type0 :=
     tt : Unit.
@@ -704,6 +706,90 @@ Definition Unit_rect := Unit_ind.
 
 (** A [Unit] goal should be resolved by [auto] and [trivial]. *)
 Hint Resolve tt : core.
+
+(** ** Natural numbers *)
+
+(** The natural numbers are defined in [Coq.Init.Datatypes] (we weren't able to excise them from the stdlib), and studied further in [Spaces.Nat].  But her we give some basic definitions. *)
+
+Fixpoint pred (n : nat) : nat
+  := match n with
+     | 0 => 0
+     | S n => n
+     end.
+
+Notation "n .-1" := (pred n) : nat_scope.
+
+(** *** Booleans *)
+
+(* coq calls it "bool", we call it "Bool" *)
+Local Unset Elimination Schemes.
+Inductive Bool : Type0 :=
+  | true : Bool
+  | false : Bool.
+Scheme Bool_ind := Induction for Bool Sort Type.
+Scheme Bool_rec := Minimality for Bool Sort Type.
+(* For compatibility with Coq's [induction] *)
+Definition Bool_rect := Bool_ind.
+
+Add Printing If Bool.
+
+Declare Scope bool_scope.
+
+Delimit Scope bool_scope with Bool.
+
+Bind Scope bool_scope with Bool.
+
+(** ** Coinductive streams *)
+
+CoInductive Stream (A : Type) := SCons
+{
+  head : A ;
+  tail : Stream A
+}.
+
+Arguments SCons {A} _ _.
+Arguments head {A} _.
+Arguments tail {A} _.
+
+CoFixpoint const_stream {A} (a : A) : Stream A
+  := SCons a (const_stream a).
+
+
+(** *** Typeclasses for case analysis *)
+
+(** Versions of [n = 0] and [n > 0] that are typeclasses and can be found automatically. *)
+Definition IsZeroNat (n : nat) : Type :=
+  match n with
+  | O => Unit
+  | S _ => Empty
+  end.
+Existing Class IsZeroNat.
+Global Instance iszeronat : IsZeroNat 0 := tt.
+
+Definition IsPosNat (n : nat) : Type :=
+  match n with
+  | O => Empty
+  | S _ => Unit
+  end.
+Existing Class IsPosNat.
+Global Instance isposnat (n : nat) : IsPosNat (S n) := tt.
+
+Definition IsTrue (b : Bool) : Type :=
+  match b with
+  | true => Unit
+  | false => Empty
+  end.
+Existing Class IsTrue.
+Global Instance istrue : IsTrue true := tt.
+
+Definition IsFalse (b : Bool) : Type :=
+  match b with
+  | true => Empty
+  | false => Unit
+  end.
+Existing Class IsFalse.
+Global Instance isfalse : IsFalse false := tt.
+
 
 (** *** Pointed types *)
 
--- a/theories/ooCat/Cat0.v
+++ b/theories/ooCat/Cat0.v
@@ -1,0 +1,664 @@
+(* -*- mode: coq; mode: visual-line -*-  *)
+
+Require Import Basics.
+Require Export ooCat.Glob.
+
+(** * 0-coherent oo-categories *)
+
+Generalizable Variables m n p A B C.
+
+(** ** Basic definition *)
+
+(** The implicit oo-category convention is in effect: [IsCat0 n] means "is an (oo,n)-category".  The postfix "0" means "0-coherent".  There are different choices about what this could mean:
+
+- It could mean "having only the structure of a (0,n)-category at each dimension": composition, identities, and in the case [n=0] inversion, but no axioms.
+- It could mean the maximal difference in dimension between the inputs and outputs of operations is 0, so it has all operations but no axioms.
+
+We currently choose the latter, which happens to also be "everything that can be stated without referring to equivalences". *)
+CoInductive IsCat0 (n : nat) (A : Type) `{IsGlob n A} : Type :=
+{
+  cat_id : forall (a : A), a $-> a ;
+  cat_comp : forall {a b c : A},
+      (b $-> c) -> (a $-> b) -> (a $-> c) where "g $o f" := (cat_comp g f);
+  isfunctor0_postcomp : forall (a b c : A) (g : b $-> c),
+      IsFunctor0 (fun f : a $-> b => g $o f) ;
+  isfunctor0_precomp : forall (a b c : A) (f : a $-> b),
+      IsFunctor0 (fun g : b $-> c => g $o f) ;
+  gpd_inv : forall (isz : IsZeroNat n) (a b : A),
+      (a $-> b) -> (b $-> a) ;
+  iscat0_hom : forall (a b : A), @IsCat0 (pred n) (a $-> b) _ ;
+}.
+
+Existing Class IsCat0.
+Arguments cat_id {n A _ _} a.
+Arguments cat_comp {n A _ _ a b c} g f.
+Notation "g $o f" := (cat_comp g f).
+
+(** When we're in a groupoid dimension, we use path-like syntax. *)
+Arguments gpd_inv {n A _ _ _ a b} f.
+Notation "f ^$" := (gpd_inv f).
+Notation "f $@ g" := (@cat_comp 0 _ _ _ _ _ _ g f).
+
+Global Existing Instances
+       iscat0_hom isfunctor0_postcomp isfunctor0_precomp.
+
+Hint Extern 1 (IsCat0 ?n (Hom _ _ _)) => change_dim n : typeclass_instances.
+Hint Extern 1 (IsCat0 ?n (Hom _ _ _)) => change_dim_0 n : typeclass_instances.
+
+Definition cat_postcomp `{IsCat0 n A}
+           (a : A) {b c : A} (g : b $-> c)
+  : (a $-> b) -> (a $-> c)
+  := fun f => g $o f.
+
+Definition cat_precomp `{IsCat0 n A}
+           {a b : A} (c : A) (f : a $-> b)
+  : (b $-> c) -> (a $-> c)
+  := fun g => g $o f.
+
+Definition cat_postwhisker `{IsCat0 n A} {a b c : A}
+           {f g : a $-> b} (h : b $-> c) (p : f $-> g)
+  : h $o f $-> h $o g
+  := fmap (cat_postcomp a h) p.
+Notation "h $<o p" := (cat_postwhisker h p).
+
+Definition cat_prewhisker `{IsCat0 n A} {a b c : A}
+           {f g : b $-> c} (p : f $-> g) (h : a $-> b)
+  : f $o h $-> g $o h
+  := fmap (cat_precomp c h) p.
+Notation "p $o> h" := (cat_prewhisker p h).
+
+(* These seem to be unnecessary; I guess [cat_postcomp] and [cat_precomp] are sufficiently transparent to typeclass search. *)
+(*
+Global Instance isfunctor0_postcomp' `{IsCat0 n A}
+       {a b c : A} (g : b $-> c)
+  : IsFunctor0 (cat_postcomp a g).
+Proof. 
+  exact _.
+Defined.
+
+Global Instance isfunctor0_precomp' `{IsCat0 n A}
+       {a b c : A} (f : a $-> b)
+  : IsFunctor0 (cat_precomp c f).
+Proof.
+  exact _.
+Defined.
+*)
+
+Definition iscat0_induced {A} `{IsCat0 n B} (F : A -> B)
+  : @IsCat0 n A (isglob_induced F).
+Proof.
+  unshelve econstructor; cbn; intros.
+  - eapply cat_comp; eassumption.
+  - apply cat_id.
+  - exact _.
+  - exact _.
+  - apply gpd_inv; assumption.
+  - exact _.
+Defined.
+
+(** ** Constant functors *)
+
+CoFixpoint isfunctor0_const `{IsGlob m A} `{IsCat0 n B} (x : B)
+  : IsFunctor0 (@const A B x).
+Proof.
+  simple notypeclasses refine (Build_IsFunctor0 _ _ _).
+  { intros a b f.
+    apply cat_id. }
+  intros a b.
+  rapply isfunctor0_const.
+Defined.
+
+Global Existing Instance isfunctor0_const.
+
+
+(** ** Truncatedness and invertibility *)
+
+(** Truncatedness can be used as a cap on the invertibility dimension *)
+CoFixpoint iscat0_cattrunc
+       {n : nat} `{IsCat0 m A} `{!CatTrunc n.-1 A}
+  : @IsCat0 n A isglob_reindex.
+Proof.
+  unshelve econstructor.
+  1: clear iscat0_cattrunc; intros a b c; cbn; apply cat_comp.
+  1: clear iscat0_cattrunc; intros; cbn; apply cat_id.
+  1-2: intros; cbn;
+    apply (@isfunctor0_reindex (pred m) (pred m) (pred n) (pred n));
+    exact _.
+  - destruct n as [|[|n]]; cbn in *.
+    1:intros; rapply cat_center.
+    1-2:intros [].
+  - intros; cbn.
+    rapply (iscat0_cattrunc (pred n) (pred m) (a $-> b)).
+    destruct n as [|n]; cbn in *.
+    + apply cattrunc_pred.
+    + exact _.
+Defined.
+
+(** ** Quasi-isomorphisms *)
+
+(** A morphism in an (oo,n)-category is an equivalence if either [n=0] or if it has an inverse up to equivalence.  Since this is not a coherent structure, we denigrate it with the name "quasi-isomorphism".  We define it coinductively, but in practice it suffices to reason about it inductively since all our categories are (oo,n)-categories for some finite n. *)
+
+CoInductive IsQIso `{IsCat0 n A} {a b : A} (f : a $-> b) :=
+{
+  qiso_inv : b $-> a ;
+  qiso_issect : (qiso_inv $o f) $-> (cat_id a) ;
+  qiso_isretr : (f $o qiso_inv) $-> (cat_id b) ;
+  isqiso_qiso_issect : @IsQIso _ (a $-> a) _ _ _ _ qiso_issect ;
+  isqiso_qiso_isretr : @IsQIso _ (b $-> b) _ _ _ _ qiso_isretr ;
+}.
+Existing Class IsQIso.
+Arguments qiso_inv {n A _ _ a b} f {_}.
+Arguments qiso_issect {n A _ _ a b} f {_}.
+Arguments qiso_isretr {n A _ _ a b} f {_}.
+Global Existing Instances isqiso_qiso_issect isqiso_qiso_isretr.
+
+Hint Extern 1 (@IsQIso ?n (Hom _ _ _) _ _ _ _ _) => change_dim n : typeclass_instances.
+Hint Extern 1 (@IsQIso ?n (Hom _ _ _) _ _ _ _ _) => change_dim_0 n : typeclass_instances.
+
+(** The inverse of a quasi-isomorphism is a quasi-isomorphism. *)
+Global Instance isqiso_inv `{IsCat0 n A}
+       {a b : A} (f : a $-> b) `{!IsQIso f}
+  : IsQIso (qiso_inv f).
+Proof.
+  unshelve econstructor.
+  1: exact f.
+  1: exact (qiso_isretr f).
+  1: exact (qiso_issect f).
+  all: exact _.
+Defined.
+
+(** Every morphism in a (-2)- or (-1)-truncated category is a quasi-isomorphism. *)
+CoFixpoint isqiso_catcontr `{IsCat0 n A, !CatContr A}
+           {a b : A} (f : a $-> b)
+  : IsQIso f.
+Proof.
+  unshelve econstructor.
+  1-3:apply cat_center.
+  1-2:rapply isqiso_catcontr.
+Defined.
+Global Existing Instance isqiso_catcontr.
+
+Global Instance isqiso_catprop `{IsCat0 n A, !CatProp A}
+       {a b : A} (f : a $-> b)
+  : IsQIso f.
+Proof.
+  unshelve econstructor.
+  1-3:apply cat_center.
+  1-2:apply isqiso_catcontr.
+Defined.
+
+(** It follows that in a 0-truncated category, having maps back and forth suffices. *)
+Global Instance isqiso_catposet `{IsCat0 n A, !CatPoset A}
+       {a b : A} (f : a $-> b) (g : b $-> a)
+  : IsQIso f.
+Proof.
+  unshelve econstructor.
+  1:exact g.
+  1-2:apply cat_center.
+  1-2:apply isqiso_catprop.
+Defined.
+
+(** And in a truncated (1,1)-category, an ordinary "isomorphism" suffices. *)
+Global Instance isqiso_cat1cat `{IsCat0 1 A, !Cat1Cat A}
+       {a b : A} (f : a $-> b) (g : b $-> a)
+       (s : g $o f $== cat_id a) (r : f $o g $== cat_id b)
+  : IsQIso f.
+Proof.
+  unshelve econstructor.
+  - exact g.
+  - exact s.
+  - exact r.
+  - apply isqiso_catposet.
+    exact (s^$).
+  - apply isqiso_catposet.
+    exact (r^$).
+Defined.
+
+(** To prove much of anything else about quasi-isomorphisms (even that the identity is one!) requires more coherence. *)
+
+
+(** ** Equivalences *)
+
+(** There are various possible coherent definitions of equivalence, some generally applicable and others more specifiec.  Rather than mandate a particular one in all cases, we give the user the freedom to choose whichever they prefer for a particular category by instantiating the following typeclass.  Note though that we don't actually *require* any coherence of the notion of equivalence, only that it is logically equivalent to quasi-isomorphism.  *)
+
+CoInductive HasEquivs (n : nat) (A : Type) `{IsCat0 n A} :=
+{
+  CatIsEquiv' : forall {a b : A}, (a $-> b) -> Type ;
+  catie_isqiso' : forall (a b : A) (f : a $-> b), IsQIso f -> CatIsEquiv' f ;
+  isqiso_catie' : forall (a b : A) (f : a $-> b), CatIsEquiv' f -> IsQIso f ;
+  hasequivs_hom : forall (a b : A), @HasEquivs (pred n) (a $-> b) _ _ ;
+}.
+Existing Class HasEquivs.
+Global Existing Instance hasequivs_hom.
+
+Hint Extern 1 (HasEquivs ?n (Hom _ _ _)) => change_dim n : typeclass_instances.
+Hint Extern 1 (HasEquivs ?n (Hom _ _ _)) => change_dim_0 n : typeclass_instances.
+
+(** A field can't be a working [Existing Class] due to a bug in Coq. *)
+Class CatIsEquiv `{HasEquivs n A} {a b : A} (f : a $-> b)
+  := catie_catie : @CatIsEquiv' n A _ _ _ a b f.
+
+Hint Extern 1 (CatIsEquiv ?n (Hom _ _ _)) => change_dim n : typeclass_instances.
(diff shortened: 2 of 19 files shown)


Alizter:
`make` spits out loads of warning because Fibrations is now an ambiguous name.

mikeshulman:
Bah

mikeshulman:
I've been thinking for a while that it might be better to rename `HoTT.Fibrations` to `HoTT.HFiber`.

JasonGross:
> `make` spits out loads of warning because Fibrations is now an ambiguous name.

This is why I always fully qualify everything.  Note that I have [a tool](https://github.com/JasonGross/coq-tools/blob/master/absolutize-imports.py) that should be able to do this automatically, if so desired.

theories/Basics/Overture.v
```diff
@@ -705,6 +707,90 @@ Definition Unit_rect := Unit_ind.
 (** A [Unit] goal should be resolved by [auto] and [trivial]. *)
 Hint Resolve tt : core.
 
+(** ** Natural numbers *)
+
+(** The natural numbers are defined in [Coq.Init.Datatypes] (we weren't able to excise them from the stdlib), and studied further in [Spaces.Nat].  But her we give some basic definitions. *)
+
+Fixpoint pred (n : nat) : nat
```
JasonGross:
```suggestion
Definition pred (n : nat) : nat
```

theories/Basics/Overture.v
```diff
@@ -705,6 +707,90 @@ Definition Unit_rect := Unit_ind.
...
+
+(** ** Coinductive streams *)
+
+CoInductive Stream (A : Type) := SCons
+{
+  head : A ;
+  tail : Stream A
+}.
+
+Arguments SCons {A} _ _.
```
JasonGross: Do we want a notation for `SCons`?  Also, why is `SCons` capitalized?

mikeshulman: Not sure why I capitalized it; I guess it shouldn't be.

I don't think it's common enough to deserve a notation, though.

Alizter:
cate_compose_temp has a universe error on one of the jobs. I suspect this is a coq bug.

Alizter:
> coherence required to even define a k-transfor seems to go up with k.

Exactly which coherence is this? Might it be possible to coinductively add the necessary coherence?

>Change # 4: I defined dependent inserters directly rather than in terms of comma categories. I think this way is clearer, partly because it doesn't require product categories. We can then define comma categories in terms of them (but I didn't yet).

Is there any point defining one in terms of the other? Why not just define them separately?

> It's already possibly to talk about categories with only finitely many dimensions of cells, just by sticking in Unit at the top of the coinductive notion. But there isn't yet a way to talk about functors that only act on finitely many dimensions if their domain and codomain categories were defined to include infinitely many.

Well I don't think that is the correct way to do finite dimensional cells. We are in HoTT, which means every type comes equipped with a full oo-groupoid structure. I would rather that "finite dimensional" categories begin using regular paths as higher cells. Then we can always characterize the action of functions on paths anyway using ap.

So if I were to define a 2-cat X as some oo-cat, I would define the two dimensions and then say the higher cells are just paths. When I construct functors X -> Y, then I need only supply the action on the first two dimensions since the action on paths is always given.

In general I think this is cool. It is a bit scary as you say though. It took me some time to get on board with the (oo,n) idea, but I see why it make some more practical sense than (oo,oo).

One of the things I noticed with the (oo,oo)-definition is that it is relatively easy to show pType is a globular type. I struggled a lot trying to show it was a Cat0.

Will it ever be possible to define some sort of functor category? If not, it looks less likely that we can characterise (co)limits.

mikeshulman:
> I would rather that "finite dimensional" categories begin using regular paths as higher cells.

Yes, that's another option which is certainly better than just using Unit, but it's more work -- not at the level of `IsGlob`, but at the level of `IsCat0` and beyond where you'll start having to explain how the operations you defined at lower dimensions act on those paths.  So I would say that's sort of a middle road between the total cop-out of Unit and the morally right path of giving a correct funext-and-univalence-free definition of all the higher cells.

I have no doubt that with the right approach we'll be able to characterize (co)limits.  That might or might not involve defining a functor category.

mikeshulman:
I tried pType; it does seem that the brute force approach isn't going to work.  I'm sure there's a way to do it in this case, but this makes me think that maybe we'd need another level of parametrization: someone might want to be able to define an `IsGlob` instance that goes up at all dimensions, but an `IsCat0` instance on the same type that only goes up some finite number of dimensions.  Is it going to get unmanageable to have an extra "dimension level" of parametrization on every typeclass?

mikeshulman:
I think I see how to get pType as a displayed 0-coherent category over Type, although I haven't managed to push it all the way through yet: there's a missing precomposition lemma, and a universe issue when trying to sigma it up, plus the 0-coherence lemma for sigmas of displayed categories isn't complete.

mikeshulman:
And, of course, dimension restrictions on the coherence of functors will infect categories as well, since we use functors to construct categories (inserters, comma categories, etc.).

It would be pleasing if we could simply parametrize categories by an (n,r) as in [(n,r)-category](https://ncatlab.org/nlab/show/(n,r)-category), where n can be infinity but r cannot.  Then n would be a limit on the dimensions of stuff, r a bound on invertibility, and of course there will still be a coherence index as always that would be the only difference with "classical" higher category theory.  Functors and transfors would have only an n bound, I guess.

There will probably be some `min`s involved when performing constructions.  Unless we behave like Coq's universe polymorphism and assume `m <= n` and `m <= p` instead of using `min(n,p)`.  Along those lines, I have been wondering whether it would be reasonable to make inequality of natural numbers and trunc indices into typeclasses in general.

Alizter:
>I have been wondering whether it would be reasonable to make inequality of natural numbers and trunc indices into typeclasses in general.

Trunc index inequality is currently a type class.

mikeshulman:
Oh my, you're right.  Is it used anywhere?

Alizter:
@mikeshulman I think I made it a typeclass when I was doing the Trunc_min stuff.

Alizter:
So for 2-categories there is a notion of "hom-wise core". Do we have something similar but for only 1-cells? I imagine such an example wouldn't fit into the (oo,n)-framwork.

mikeshulman:
Yes, of course you can.  The end result won't be describable as an (n,r)-category, but it makes sense to discard the noninvertible 1-cells and keep all the 2-cells between the invertible ones.

Alizter:
@mikeshulman Do you know of any examples of this being done in 2-cat literature?

mikeshulman:
Not offhand.

Alizter:
Hmm, an example where this might be done would be to take a monoidal category C. Then consider it's 2-category delooping BC. For whatever reason, you might be interested in the sub 2-category that consists of "objects with inverses". Then this "lower core" would correspond to some sort of categorification of taking the subtgroup of invertible elements in a monoid.

I also suspect there would be a Grothendieck-like construction that would correspond to formally inverting all the lower cells, though I am not sure what the higher cells for those would be.

Alizter:
@mikeshulman Could you rebase this?

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+{
+  Hom : A -> A -> Type@{u} ;
+  isglob_hom : forall (x y : A), IsGlob n.-1 (Hom x y) ;
+}.
+(* Technically this definition could allow two universe parameters, [A : Type@{u1}] and [Hom a b : Type@{u0}] with u0 <= u1.  But nearly all subsequent definitions force [u0 = u1].  Moreover, curiously, allowing [u0 != u1] here causes [iscat0_type] to fail with a universe error, while forcing [u0 = u1] here allows [iscat0_type] to succeed.  I don't fully understand why that is. *)
+
+Existing Class IsGlob.
+Arguments Hom {n} A {_} a b.
+Notation "a $-> b" := (@Hom _ _ _ a b).
+Notation "a $== b" := (@Hom 0 _ _ a b). (** In the invertible dimensions, we use a notation that looks more like paths or homotopies *)
```
JasonGross, line 41: Do we want also an only printing notation for `@Hom (pred 0) _ _ a b`?

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+
+(** ** Dependent/displayed globular types *)
+
+(** For now, we require no relation between the invertibility dimensions of [A] and [B]. *)
+CoInductive IsDGlob {m : nat} {A : Type@{u}} (n : nat) (B : A -> Type@{u})
+            `{IsGlob m A} : Type :=
+{
+  DHom : forall (a b : A) (f : a $-> b) (u : B a) (v : B b), Type@{u} ;
+  isdglob_dhom : forall (a b : A) (u : B a) (v : B b),
+      @IsDGlob (pred m) (a $-> b) (pred n) (fun f => DHom a b f u v) _ ;
```
JasonGross, line 71: nit: You use notation for `IsGlob`, why not also for `IsDGlob`?
```suggestion
      @IsDGlob m.-1 (a $-> b) n.-1 (fun f => DHom a b f u v) _ ;
```

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+}.
+
+Existing Class IsDGlob.
+Arguments DHom {m A n B _ _ a b} f u v.
+(** Since [DHom] has three arguments that generally need to be given explicitly, we don't try to give it an infix notation. *)
```
JasonGross, line 76: You could use something like `u $-{ f }-> v`?

mikeshulman: I think I tried that, but found it unreadable in practice.

JasonGross:
I've rebased this at https://github.com/JasonGross/HoTT/tree/pr/1302, fixed the build conflicts, and also fixed the universe issue (and reported it as https://github.com/coq/coq/issues/14057).

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+(** We can induce a globular structure by pullback along a function.  This is not an instance, of course; we have to choose when to apply it. *)
+Definition isglob_induced {A} `{IsGlob n B} (F : A -> B) : IsGlob n A
+  := Build_IsGlob n A (fun a b => F a $-> F b) _.
+
+
+(** ** Dependent/displayed globular types *)
+
+(** For now, we require no relation between the invertibility dimensions of [A] and [B]. *)
+CoInductive IsDGlob {m : nat} {A : Type@{u}} (n : nat) (B : A -> Type@{u})
+            `{IsGlob m A} : Type :=
```
JasonGross, line 67: Is it important that `IsDGlob` is indexed over the globular structure of `A`, rather than providing the globular structure of both `A` and `B` at once?  The reason I ask is because there seems to be a nice generalization to telescopes that makes `IsGlob` the base case and `IsDGlob` a sort of inductive step, but I think it requires bundling all the globular structure as fields rather than parameterizing over it, I think.

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+Proof.
+  unshelve econstructor.
+  - intros a b f u v; exact (DHom (fmap F f) u v).
+  - intros a b u v. 
+    refine (isdglob_compose _ (F a $-> F b) (fun g => DHom g u v)
+                            _ _ _ _ _ _ (fmap F) _).
+Defined.
+Existing Instance isdglob_compose.
+
+(** Tlsil has suggested that we could also define a dependent globular type to be a globular map into a globular type of spans.  This would make [isdglob_compose] an instance of [isfunctor0_compose].  But it would require defining globular maps before dependent globular types, with the effect that globular maps wouldn't be definitionally globular sections of constant dependent globular types. *)
```
JasonGross, line 290: I'm struggling to understand how this fits into the picture.  Can you elaborate on it a bit more?

The perspective I'm coming from is that it seems like all of the dependent globular theory should generalize to telescopes (i.e., from `forall a, B` to `forall a b ... c, D`), and I'm trying to understand what the telescope picture looks like here.  Does doing this in a telescope way require defining globular maps mutually with globular types?

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+Class IsTerminal `{IsGlob n A} (a : A) :=
+  catcontr_terminal : forall b, CatContr (b $-> a).
+Global Existing Instance catcontr_terminal.
+
+Class HasTerminal `{IsGlob n A} :=
+{
+  terminal_obj : A ;
+  isterminal_terminal : IsTerminal terminal_obj ;
+}.
+Global Existing Instance isterminal_terminal.
```
JasonGross, line 370: I think this should be parameterized over laxity, so that we get things to line up more judgmentally.  What do you think?

theories/ooCat/Laxity.v
```diff
@@ -0,0 +1,37 @@
...
+
+(** It may seem backwards for [colax] to mean a morphism "forwards" and [lax] a morphism "backwards", but that's what matches the standard terminology for natural transformations. *)
+Definition lHom (l : Laxity) `{HasEquivs n A} (a b : A) :=
+  match l with
+  | colax => (a $-> b)
+  | pseudo => (a $<~> b)
+  | lax => (b $-> a)
+  end.
+
+Definition lcat_id (l : Laxity) `{IsCat1 n A} (a : A)
```
JasonGross, line 31: Should we also add
```coq
Definition lcat_compose {l : Laxity} `{IsCat1 n A} {a b c : A} (f : lHom l a b) (g : lHom l b c) : lHom l a c
```
?

theories/ooCat/Cat0.v
```diff
@@ -0,0 +1,664 @@
...
+(** A morphism in an (oo,n)-category is an equivalence if either [n=0] or if it has an inverse up to equivalence.  Since this is not a coherent structure, we denigrate it with the name "quasi-isomorphism".  We define it coinductively, but in practice it suffices to reason about it inductively since all our categories are (oo,n)-categories for some finite n. *)
+
+CoInductive IsQIso `{IsCat0 n A} {a b : A} (f : a $-> b) :=
+{
+  qiso_inv : b $-> a ;
+  qiso_issect : (qiso_inv $o f) $-> (cat_id a) ;
+  qiso_isretr : (f $o qiso_inv) $-> (cat_id b) ;
+  isqiso_qiso_issect : @IsQIso _ (a $-> a) _ _ _ _ qiso_issect ;
+  isqiso_qiso_isretr : @IsQIso _ (b $-> b) _ _ _ _ qiso_isretr ;
+}.
```
JasonGross, line 148: Is there a reason for using this structure rather than talking about adjoint equivalences?  They're the same, right, just involving a reshuffling of data?

theories/ooCat/Cat0.v
```diff
@@ -0,0 +1,664 @@
...
+  - intros; cbn.
+    rapply (iscat0_cattrunc (pred n) (pred m) (a $-> b)).
+    destruct n as [|n]; cbn in *.
+    + apply cattrunc_pred.
+    + exact _.
+Defined.
+
+(** ** Quasi-isomorphisms *)
+
+(** A morphism in an (oo,n)-category is an equivalence if either [n=0] or if it has an inverse up to equivalence.  Since this is not a coherent structure, we denigrate it with the name "quasi-isomorphism".  We define it coinductively, but in practice it suffices to reason about it inductively since all our categories are (oo,n)-categories for some finite n. *)
```
JasonGross, line 139: I'm a bit confused by the `n=0` comment, since it seems that `gpd_inv` is not enough to fill `qiso_issect` and `qiso_isretr`?  I think this means we should be indexing "(oo,n)-category" starting with n = -1, not n = 0, so that the inverses we get are actually connected to the identity?  Or is there a deep/good reason that we're not considering categories to be truncated at levels above `n`?

JasonGross: Thinking a bit more about this, it seems like maybe `gpt_inv` should be removed entirely, and its content is subsumed by being truncated one level up?

theories/ooCat/EquivCat.v
```diff
@@ -0,0 +1,24 @@
...
+}.
+
+CoInductive IsEquivCat {m A n B} `{IsCat0 m A, HasEquivs n B}
+            (F : A -> B) `{!IsFunctor0 F} :=
+{
+  isequivcat_esssplit : EssSplit F ;
+  isequivcat_hom : forall (a b : A),
+      @IsEquivCat (pred m) (a $-> b) (pred n) (F a $-> F b)
+                  _ _ _ _ _ (fmap F) _;
+}.
```
JasonGross, line 24: What's special about this notion of equivalence, over any of the others?

theories/ooCat/Forall.v
```diff
@@ -0,0 +1,183 @@
...
+
+Generalizable Variables m n p A B C.
+
+(** * Forall (oo,n)-categories (with plain types as base) *)
+
+(** ** Forall (oo,n)-categories *)
+
+(** [forall] preserves globular types. *)
+CoFixpoint isglob_forall `(B : A -> Type) `(forall a, IsGlob n (B a))
+  : IsGlob n (forall a, B a).
```
JasonGross, line 14: It seems a bit strange to me that `Σ A B` needs globular structure on both `A` and `B`, but `Π A B` needs globular structure only on `B`.  Do you have intuition for this?

theories/ooCat/Sigma.v
```diff
@@ -0,0 +1,33 @@
...
+  - intros [a u] [b v] [c w] [g q] [f p].
+    exists (g $o f).
+    exact (q $oD p).
+  - intros [a u].
+    exists (cat_id a).
+    exact (dcat_id u).
+  - intros [a u] [b v] [c w] [g q].
+Abort.
+
+(* Global Existing Instance iscat0_sigma. *)
```
JasonGross, line 33: Why is this commented?

theories/ooCat/Transformation.v
```diff
@@ -0,0 +1,216 @@
...
+    + exact (GenDInserter (tail ls) (cat_postcomp (F a) tb o fmap F)
+                          (cat_precomp (G b) ta o fmap G)).
+    + exact (GenDInserter (tail ls) (cat_postcomp (F a) tb o fmap F)
+                          (cat_precomp (G b) ta o fmap G)).
+    + exact (GenDInserter (tail ls) (cat_postcomp (G a) tb o fmap G)
+                          (cat_precomp (F b) ta o fmap F)).
+  - intros a b ta tb; cbn.
+    unfold GenDInserter in ta, tb.
+    set (l := head ls) in *.
+    destruct l;
```
JasonGross, line 35: What's up with the `set`?

theories/ooCat/Transformation.v
```diff
@@ -0,0 +1,216 @@
...
+    apply isdglob_gendinserter.
+Defined.
+Global Existing Instance isdglob_gendinserter.
+
+(** We can forget the invertibility of an iso-inserter. *)
+Definition colax_pseudo_inserter
+           (ls : Stream Laxity) `{IsGlob m A} `{HasEquivs n B}
+           (F G : A -> B) `{!IsFunctor0 F, !IsFunctor0 G} (a : A)
+  : (GenDInserter (scons pseudo ls) F G a) -> (GenDInserter (scons colax ls) F G a)
+  := fun f => cate_fun f.
```
JasonGross, line 45: There's a more general version of this, where you can go from `pseudo` to any laxity, right?

theories/ooCat/Transformation.v
```diff
@@ -0,0 +1,216 @@
...
+  - cbn; intros a b u v.
+    apply isdfunctor0_idmap.
+Defined.
+
+(** And we can turn around a lax inserter into a colax one. *)
+Definition colax_lax_inserter
+           (ls : Stream Laxity) `{IsGlob m A} `{HasEquivs n B}
+           (F G : A -> B) `{!IsFunctor0 F, !IsFunctor0 G} (a : A)
+  : (GenDInserter (scons lax ls) G F a) -> (GenDInserter (scons colax ls) F G a)
+  := fun f => f.
```
JasonGross, line 64: Maybe we should have `flip : Laxity -> Laxity` or something and this should be more general?

theories/ooCat/Transformation.v
```diff
@@ -0,0 +1,216 @@
...
+  pose (iscatsect0_fmapD alpha alnat x y).
+  exact (iscatsect0_isdfunctor0_compose
+           (colax_lax_inserter ls ((cat_precomp (G y) (alpha x)) o (fmap' G x y))
+                                ((cat_postcomp (F x) (alpha y)) o (fmap' F x y)))
+           (@fmapD _ _ _ _ _ _ alpha alnat x y)).
+Defined.
+
+(** If we generalized comma categories (and hence inserters) to act on category-sections rather than just functors, we could in principle iterate this approach to define modifications and higher transfors, analogously to how we iterate [isglob_forall] to define all the higher structure of [Type] and similarly for [pType].  Such a generalization seems to require either that the displayed category is an isofibration, or that we define a more general notion of inserter over a given natural transformation, and either one seems to require 2-coherent categories.  In fact, since the level of coherence seems to go down by one each time we apply a comma construction, we probably wouldn't be able to use this to define a whole oo-category of oo-categories at any fixed coherence level.  *)
```
JasonGross, line 216: What's the relationship between comma categories and inserters?

theories/ooCat/Cat0.v
```diff
@@ -0,0 +1,664 @@
...
+  isqiso_catie' : forall (a b : A) (f : a $-> b), CatIsEquiv' f -> IsQIso f ;
+  hasequivs_hom : forall (a b : A), @HasEquivs (pred n) (a $-> b) _ _ ;
+}.
+Existing Class HasEquivs.
+Global Existing Instance hasequivs_hom.
+
+Hint Extern 1 (HasEquivs ?n (Hom _ _ _)) => change_dim n : typeclass_instances.
+Hint Extern 1 (HasEquivs ?n (Hom _ _ _)) => change_dim_0 n : typeclass_instances.
+
+(** A field can't be a working [Existing Class] due to a bug in Coq. *)
```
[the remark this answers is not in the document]
Alizter: coq/coq#12975 which will be fixed by coq/coq#9711

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+(** ** Dependent/displayed globular types *)
+
+(** For now, we require no relation between the invertibility dimensions of [A] and [B]. *)
+CoInductive IsDGlob {m : nat} {A : Type@{u}} (n : nat) (B : A -> Type@{u})
+            `{IsGlob m A} : Type :=
+{
+  DHom : forall (a b : A) (f : a $-> b) (u : B a) (v : B b), Type@{u} ;
+  isdglob_dhom : forall (a b : A) (u : B a) (v : B b),
+      @IsDGlob (pred m) (a $-> b) (pred n) (fun f => DHom a b f u v) _ ;
+}.
```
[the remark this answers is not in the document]
Alizter: @JasonGross I would be interested in this telescoping idea, could you write a small example?

theories/ooCat/Glob.v
```diff
@@ -0,0 +1,370 @@
...
+CoInductive IsCatSect0 {m A n B} `{IsDGlob m A n B} (F : forall a, B a) :=
+{
+  fmapD : forall (a b : A) (f : a $-> b), DHom f (F a) (F b) ;
+  iscatsect0_fmapD : forall (a b : A),
+      @IsCatSect0 (pred m) (a $-> b) (pred n) (fun f => DHom f (F a) (F b))
+                  _ _ (fmapD a b) ;
+}.
+
+Existing Class IsCatSect0.
+Arguments fmapD {_ _ _ _ _ _} F {_ a b} f.
```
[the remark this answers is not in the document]
Alizter: I did not know you could give multiple declarations of arguments! The [documentation](https://coq.inria.fr/refman/language/extensions/arguments-command.html#grammar-token-implicits_alt) still seems a bit sparse but this is cool feature.

theories/ooCat/Glob.v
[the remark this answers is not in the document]
JasonGross: I don't have it fully figured out, but here's what I have so far:
```coq
Set Universe Polymorphism.
Set Implicit Arguments.
Set Primitive Projections.
Record sig A P := exist { pr1 : A ; pr2 : P pr1 }.
Arguments sig (A P)%type.
Notation "{ x | P }" := (sig (fun x => P)) : type_scope.
Notation "{ x : A | P }" := (sig (A:=A) (fun x => P)) : type_scope.
Notation "{ ' pat | P }" := (sig (fun pat => P)) : type_scope.
Notation "{ ' pat : A | P }" := (sig (A:=A) (fun pat => P)) : type_scope.
Notation "n '.-1'" := (pred n) (at level 9, format "n '.-1'") : nat_scope.
Notation "n '.+1'" := (S n) (at level 9, format "n '.+1'") : nat_scope.
Notation "( x ; y )" := (exist _ x y) : fibration_scope.
Notation "( x ; .. ; y ; z )" := (exist _ x .. (exist _ y z) ..) : core_scope.
Fixpoint telescope (n : nat) : Type :=
  match n with
  | 0 => unit
  | n.+1 => { A : Type | A -> telescope n }
  end.
Fixpoint Sigma {n} : telescope n -> Type
  := match n with
     | 0 => fun _ => unit
     | n.+1 => fun '(A; B) => { a : A | Sigma (B a) }
     end.
Fixpoint Pi {n} : forall t : telescope n, (Sigma t -> Type) -> Type
  := match n with
     | 0 => fun _ B => B tt
     | n.+1 => fun '(A; t) B => forall a : A, Pi _ (fun s => B (a; s))
     end.
```
I haven't yet figured out what the right types are, but I think we need to bundle `IsDGlob` to get this to work.  The next step is to write something that takes in `n : nat` and spits out `{ A : Type | { B : A -> Type | { C : forall a, B a -> Type | ... } } }`.  I think I did this once before, and it's a bit tricky but doable.  This is the type that we parametrize the generalization of `IsGlob`/`IsDGlob` over.  Then we write a thing that takes in one of these sigmas and spits out the type of all the hom fields, again as a sigma.  Then, finally, we take in an element of this sigma and perform the appropriate transformation to turn it back into a thing we can pass to `IsGlob`.  In this case, passing `n := 0` will turn this coinductive into `IsGlob` judgmentally.  Passing `n:=1` should turn it into, roughly `{ IsGlob A | IsDGlob B }`.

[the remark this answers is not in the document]
JasonGross: I think something like this should work, but it's giving me universe inconsistencies:
```coq
Fixpoint Homs {n} : forall (tA tB : telescope n), Type
  := match n with
     | 0 => fun 'tt 'tt => unit
     | n.+1 => fun '(A; P) '(B; Q) => forall (a : A) (b : B), { Hom : Type | @Homs n (P a) (Q b) }
     end.
Fixpoint Homs_cps {n} : forall {tA tB}, (telescope n -> Type) -> @Homs n tA tB -> Type
  := match n with
     | 0 => fun 'tt 'tt F 'tt => F tt
     | n.+1
       => fun '(A; P) '(B; Q) cont Hom
          => forall (a : A) (b : B),
              @Homs_cps n (P a) (Q b) (fun t => cont (pr1 (Hom a b); (fun _ (* ??? *) => t))) (pr2 (Hom a b))
     end.
Inductive Homs_cps_ind : forall {n tA tB}, (telescope n -> Type) -> @Homs n tA tB -> Type :=
| Homs0 : forall F, F tt -> @Homs_cps_ind 0 tt tt F tt
| HomsS : forall {n} A P B Q cont Hom,
    (forall (a : A) (b : B), @Homs_cps_ind n (P a) (Q b) (fun t => cont (pr1 (Hom a b); (fun _ (* ??? *) => t))) (pr2 (Hom a b)))
    -> @Homs_cps_ind (n.+1) (A; P) (B; Q) cont Hom
.
CoInductive IsGlob@{u} (n : nat) (A : Type@{u}) : Type :=
  {
    Hom : A -> A -> Type@{u} ;
    isglob_hom : forall (x y : A), IsGlob n.-1 (Hom x y) ;
  }.
CoInductive IsDGlob@{u} (n : nat) {A : Type@{u}} (B : A -> Type@{u}) : Type :=
  {
    DHom0 : A -> A -> Type@{u}
    ; DHom1 : forall (a b : A) (f : DHom0 a b) (u : B a) (v : B b), Type@{u}
    ; isdglob_dhom : forall (a b : A) (u : B a) (v : B b),
        @IsDGlob (pred n) (DHom0 a b) (fun f => @DHom1 a b f u v) ;
  }.
CoInductive IsGenGlob (n : nat) {len} (t : telescope len) : Type :=
  {
    GHom : Homs t t
    ; isgenglob_ghom : Homs_cps_ind (IsGenGlob n) GHom
  }.
```
There's a `???` where I'm not sure that I set things up right.  I expect that you should be able to prove `IsGlob`  here equivalent to `IsGenGlob (len:=0)` and `IsDGlob` here (which is bundled) equivalent to `IsGenGlob (len:=1)`.  It also shouldn't be hard to generalize to a tuple of `nat`s rather than a single one, since the `nat`s are independent from the types.

theories/ooCat/Forall.v
Alizter: Could it be that when you construct paths in sigma types you need paths in both components, but paths in pi types need paths in the domain.

JasonGross: Paths in Pi types seem to need only paths in the target, but this isn't really the case (or it's only the case because the paths in the source are free).  A more symmetric presentation, which is equivalent, says that paths in Sigma types are Sigma of a path and a path over that path, and paths in Pi types are Pis from paths to paths over those paths.  It's just that the latter base paths have both endpoints free, so we can contract them by path induction.  But I'm not sure the same should be true of globular types, and it seems to me that this is implicitly using the path globular structure for the domain.

Alizter: I seem to recall a discussion with @mikeshulman about something similar with pType. There are a few `forall` pTypes you can construct depending which types you also require to be pointed. They are not equivalent.

Alizter: You might be able to find an answer in here: https://arxiv.org/pdf/1705.04296.pdf however.

theories/ooCat/Transformation.v
Alizter: I believe comma categories can be thought of as a "lax pullback" and inserters can be thought of as a "lax equalizer" then just as regular pullbacks can be defined as equalizers, comma categories can be defined as inserters.

Concretely, $f : A \to C$ and $g : B \to C$ has the comma category $(f / g)$ given as $\marthrm{Ins}(f \circ \pi_1, g \circ \pi_2)$.

Alizter: This PR is pretty dead, but I thought I might as well answer this question as I was looking through. Sorry for the late reply.

mikeshulman: It's correct that comma categories can be constructed from products and  inserters as you say, but it's not really correct to call them "lax pullbacks", or inserters "lax equalizers", although it's not uncommon.  Properly speaking a "lax pullback" should be the lax limit of a cospan, and similarly a "lax equalizer" should be the lax limit of a parallel pair.

PR #1304, opened 2020-03-14
renamed Fibrations.v to HFiber.v

### LINKED ISSUE #1303: rename `HoTT.Fibrations` to `HoTT.HFiber`.
I've been thinking for a while that it might be better to rename `HoTT.Fibrations` to `HoTT.HFiber`.

_Originally posted by @mikeshulman in https://github.com/HoTT/HoTT/pull/1302#issuecomment-598888126_

Alizter:
closes #1303


Alizter:
@mikeshulman are you able to review this?

mikeshulman:
Thanks for this, and sorry for the delay.  My main goal here is to make it intuitive which file one needs to import to get a particular result.  For instance, it makes sense that to get results about hfibers you need to import `HFiber`.  However, there are a few results currently in this file for which I don't think it's intuitive that you have to import `HFiber`, notably `isequiv_from_functor_sigma` and those involving `IsTruncMap`.  Is there somewhere else we could put those results that would make them easier to find?  I almost feel like those are such basic results that they should be in `Basics` or `Types` somewhere.

Alizter:
@mikeshulman What about moving HFiber to Types? Then we will never have to worry about importing it.

Alizter:
OK, I have rebased and reorganised HFiber. I've moved the "hfiber of functor" results to Types also.

theories/Types/Equiv.v
```diff
@@ -164,4 +164,25 @@ Section AssumeFunext.
   : (A <~> B) <~> (B <~> A)
     := Build_Equiv _ _ (@equiv_inverse A B) _.
 
+  (** If [functor_sigma idmap g] is an equivalence then so is g *)
```
mikeshulman, line 167: Why are these here rather than in `Types/Sigma`?

Alizter:
They require the fcontr results which rely on Types/Sigma. Maybe we should
kove fcontr to Basics?

mikeshulman:
Ah right.  That's probably okay, but maybe we should add a comment in `Types/Sigma` pointing the reader to `Types/Equiv` to find them.

theories/Types/Sigma.v
```diff
@@ -709,3 +709,50 @@ Definition pr1_path_path_sigma_hprop {A : Type} {P : A -> Type}
...
+           (g : forall a, P a -> Q a)
+           (b : A) (v : Q b)
+: (hfiber (functor_sigma idmap g) (b; v)) <~>
+   hfiber (g b) v.
+Proof.
+  refine (_ oE hfiber_functor_sigma P Q idmap g b v).
+  exact (equiv_contr_sigma
+           (fun (w:hfiber idmap b) => hfiber (g w.1) (transport Q (w.2)^ v))).
+Defined.
+
```
Alizter, line 750:
```suggestion

(** The converse and Theorem 4.7.7 can be found in Types/Equiv.v *)
```

Alizter: @mikeshulman Something like this?

mikeshulman: Yeah, looks good.

Alizter:
Alright before we merge I will have a look for unnecessary imports.

Alizter:
OK, I've removed the unnecessary imports of HFiber.v

Alizter:
So the ubuntu repositories are being hosted on azure. I've heard rumors that azure is under a lot of strain due to current events. It will probably go away tomorrow.

Alizter:
@mikeshulman Are we good to merge now?

PR #1305, opened 2020-03-20
Update INSTALL.md

jcommelin:
The INSTALL instructions didn't work for me on a recent Ubuntu. In particular, the `opam switch create`  command didn't work (because the version number had a typo), and without switching to the correct version I got many errors that I didn't understand. (First time ocaml user here.)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,6 +11,8 @@ Git and make can be installed through your package manager.  On Debian
 or any distribution with `apt-get` you can run the script
 `./etc/install_coq_deps.sh` in the HoTT directory (see step 2) which
 installs some dependencies automatically.
+(Beware: on Ubuntu you also need the dependencies `m4` and `autoconf`,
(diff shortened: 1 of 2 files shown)


Alizter:
@jcommelin Did you run the `etc/install_coq_deps.sh` script? That should install all the dependencies for coq. The opam switch feature is newer than the instructions which is why we haven't documented it yet.

Did you manage to finish the installation or do you need further help?

jcommelin:
I think I finished the installation succesfully now. I've run `make` and it compiled a long list of `.v` files (-;

The `etc/install_coq_deps.sh` script did not install everything I needed. But I assumed that it was catering to Debian. Should I edit the script?

Alizter:
@jcommelin Good to hear you managed to compile everything. If you could update the script that would be great.

jcommelin:
Ok, I've edited the script as well.

Alizter:
Thank you

PR #1309, opened 2020-03-31
Toda brackets

Alizter:
Here is a basic definition of Toda bracket. Since Toda brackets rely on the data of the two null homotopies, I've defined the set of Toda brackets to be a quotient.

This doesn't yet correspond to the classical usage however. It shouldn't be too difficult to show that (pointed) Toda brackets of S^n -> X -> Y -> Z embed into Pi n+1 Z.

Showing that Toda brackets of W -> X -> Y -> Z embed into ||Susp W -> Z||_0 required univalence.

I'm making this a draft for now, since I might add some other things to it.


Alizter:
I'm going to close this for now as I haven't got any plans to use it.

PR #1310, opened 2020-04-11
Make some proofs about pointed maps transparent.  Note X ->* Y is ptd.

jdchristensen:
Make a version of pointed_reduce without the rewrites, and use it in a bunch of places.  Also make a version that just handles a single map, to keep the context clean.  Give shorter default names for some of the destructed things, to make things more readable.

Unrelated, but also define the pointed type of pointed maps.

### DIFF
--- a/theories/Pointed/pMap.v
+++ b/theories/Pointed/pMap.v
@@ -41,32 +41,42 @@ Defined.
 Definition pmap_compose_assoc {A B C D : pType} (h : C ->* D)
   (g : B ->* C) (f : A ->* B) : (h o* g) o* f ==* h o* (g o* f).
 Proof.
-  pointed_reduce.
+  pointed_reduce'.
   simple refine (Build_pHomotopy _ _); cbn.
   - intros ?; reflexivity.
   - reflexivity.
-Qed.
+Defined.
 
 Definition pmap_precompose_idmap {A B : pType} (f : A ->* B)
 : f o* pmap_idmap ==* f.
 Proof.
-  pointed_reduce.
+  pointed_reduce_pmap f.
   simple refine (Build_pHomotopy _ _); cbn.
   - intros ?; reflexivity.
   - reflexivity.
-Qed.
+Defined.
 
 Definition pmap_postcompose_idmap {A B : pType} (f : A ->* B)
 : pmap_idmap o* f ==* f.
 Proof.
-  pointed_reduce.
+  pointed_reduce_pmap f.
   simple refine (Build_pHomotopy _ _); cbn.
   - intros ?; reflexivity.
   - reflexivity.
-Qed.
+Defined.
 
 (** ** Trivially pointed maps *)
 
 (** Not infrequently we have a map between two unpointed types and want to consider it as a pointed map that trivially respects some given point in the domain. *)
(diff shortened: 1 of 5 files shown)


mikeshulman:
Ali, can you tell if this will play well with the additional unmerged work done in `Pointed` in the wildcat branch?

mikeshulman:
ppMap is also defined there, IIRC, but presumably the definitions are identical and could be merged easily.

mikeshulman:
Other than that worry, this looks fine to me.

theories/Pointed/pMap.v
```diff
@@ -41,32 +41,42 @@ Defined.
...
 
 (** ** Trivially pointed maps *)
 
 (** Not infrequently we have a map between two unpointed types and want to consider it as a pointed map that trivially respects some given point in the domain. *)
 Definition pmap_from_point {A B : Type} (f : A -> B) (a : A)
   := Build_pMap (Build_pType A a) (Build_pType B (f a)) f 1%path.
+
+(** ** The type of pointed maps is itself pointed *)
+
+Definition constpmap {X Y : pType} : X ->* Y := Build_pMap _ _ (const (point Y)) idpath.
```
Alizter, line 76: Perhaps a better name would be `pmap_const`?

Alizter:
The big difference would be the definition of pHomotopy which meant Floris had to rewrite how some of the pointed_reduce tactics worked. It doesn't look like there are any major conflicts however.

theories/Pointed/pMap.v
Alizter: Or nevermind we can do this later if we need to.

jdchristensen: I'm flexible.

jdchristensen:
> The big difference would be the definition of pHomotopy which meant Floris had to rewrite how some of the pointed_reduce tactics worked. It doesn't look like there are any major conflicts however.

Can you point me to this?  In my work in progress, I also had to redefine pHomotopy to make certain things go smoothly.  Instead of `pointed_htpy (point A) @ point_eq g = point_eq f` I used `pointed_htpy (point A) = point_eq f@ (point_eq g)^`.  This has the advantage that when `f` and `g` are `constpmap`, the RHS reduces to `idpath` and so `A` is the same as a pointed map into the loop space.

I tried doing this in Pointed/Core.v, but then `pointed_reduce` didn't work, so instead I have to translate back and forth a bit.

Alizter:
@jdchristensen Here is the wildcat branch we were talking about: https://github.com/mikeshulman/HoTT/tree/wildcat/theories

The big change is that pHomotopy is now defined as a "dependent pointed map" which allows us to define higher homotopies similarly to how homotopies (between dependent maps) are dependent maps hence we can have homotopies between homotopies. I think Floris had to change the equation round like you described.

jdchristensen:
That's great.  It switches to exactly the convention I was forced to use (and does it in a dependent way).  Will the pointed stuff there get merged soon?  I'm doing lots of stuff with pointed types and could probably take advantage of the improvements.

mikeshulman:
I wanted to merge it a long time ago, but some people thought it wasn't ready.  I think we should go ahead and merge the Pointed stuff from that branch now, and if we need to tweak it further later we can do it then.

PR #1311, opened 2020-04-11
Fix typos in comments

jdchristensen:


mikeshulman:
Some of these are leftovers because some of the connectedness lemmas were originally proven only for the truncation modality and then generalized to modalities.

jdchristensen:
Right, I should have said "fixes to comments" or something like that.

Alizter:
@jdchristensen I've merged the latest master into the wildcat branch so that it now compiles.

PR #1312, opened 2020-04-12
Pointed from wildcat branch

Alizter:
Here is most of the stuff about pointed types that was developed in the wildcat branch.

The biggest change is in the definition of pMap.

We define a pointed family to be a type family over a pointed type with a section at the base point (which we call `dpoint`). We then define `pForall` to be sections of this type family with a proof that these sections agree at the base point with `dpoint`. This way `pMap` becomes a special case of `pForall` by taking the constant pointed family `pfam_const`. This approach also let's us define pHomotopy more generally for two pForall's similar to homotopies. In fact, we can define pHomotopy as a pForall, this has the advantage that higher homotopies are also pHomotopy. This let's us directly replicate the situation for pi-types and homotopies for pType.

Most of this was written and developed by Floris van Doorn.


theories/Pointed/Core.v
```diff
@@ -9,113 +10,158 @@ Local Open Scope path_scope.
...
+  := Build_pType (sig P) (point A; P.2).
 
-(* Pointed pi types, note that the domain is not pointed *)
-Definition pforall {A : Type} (F : A -> pType) : pType
+(** Pointed pi types, note that the domain is not pointed *)
+Definition pproduct {A : Type} (F : A -> pType) : pType
   := Build_pType (forall (a : A), pointed_type (F a)) (ispointed_type o F).
 
-(** The following tactics often allow us to "pretend" that pointed maps and homotopies preserve basepoints strictly.  We have carefully defined [pMap] and [pHomotopy] so that when destructed, their second components are paths with right endpoints free, to which we can apply Paulin-Morhing path-induction. *)
+(** The following tactic often allows us to "pretend" that pointed maps and homotopies preserve basepoints strictly.  We have carefully defined [pMap] and [pHomotopy] so that when destructed, their second components are paths with right endpoints free, to which we can apply Paulin-Morhing path-induction. *)
```
jdchristensen, line 152:
```suggestion
(** The following tactics often allow us to "pretend" that pointed maps and homotopies preserve basepoints strictly. *)
```

theories/HFiber.v
```diff
@@ -65,6 +65,50 @@ Proof.
   - intros a e; exact ((p a)^ @ ap k e).
 Defined.
 
+(* This doesn't really need to be defined as an instance, since typeclass search can already find it, but we state it for the reader's benefit. *)
```
jdchristensen, line 68: Will having this as an instance slow down typeclass search?  E.g. will Coq try two equivalent things, duplicating entire sub-trees of the search?

Alizter:
I've just force pushed since I forgot to add something to the last commit. The CI should be happy now.

theories/HFiber.v
Alizter: I don't think it will slow it down, but I am unsure. The way to really know is to `Set Typeclasses Debug` and find something that we might think will be effected. I have done similar "instances for the readers benefit" before and I haven't noticed any issues.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -95,35 +110,50 @@ Definition pi_functor_type (n : nat) (X Y : pType) : Type
      end.
 
 (* Every such map is, in particular, a pointed map. *)
-Definition pi_functor_type_pmap n X Y : pi_functor_type n X Y -> (Pi n X ->* Pi n Y)
+Definition pi_functor_type_pmap {n X Y} : pi_functor_type n X Y -> (pForall (Pi n X) (pfam_const (Pi n Y)))
```
mikeshulman, line 113: Is the remark lower down about coercions the reason for writing this as a `pForall` instead of `pMap`?  It's kind of ugly.  Would it work to just make `A ->* B` a `Notation` for `pForall A (pfam_const B)`?

theories/Homotopy/HomotopyGroup.v
```diff
@@ -132,7 +162,7 @@ Proof.
   destruct n; intros x.
   - apply Trunc_functor_idmap.
   - etransitivity.
-    + apply O_functor_homotopy, iterated_loops_functor_idmap.
+    + apply O_functor_homotopy. exact (iterated_loops_functor_idmap _ (n.+1)).
```
mikeshulman, line 165: Is this for speed reasons?

theories/Pointed/Loops.v
```diff
@@ -34,22 +35,22 @@ Defined.
 Definition istrunc_loops {n} (A : pType) `{IsTrunc n.+1 A}
   : IsTrunc n (loops A) := _.
 
-(** Similarly for connectedness. *)
+(* (** Similarly for connectedness. *)
```
mikeshulman, line 38: What's the problem here?

theories/WildCat/PointedCat.v
```diff
@@ -0,0 +1,50 @@
...
+Global Existing Instance isinitial_zero_object.
+Global Existing Instance isterminal_zero_object.
+
+Definition zero_morphism {A : Type} `{IsPointedCat A} {a b : A} : a $-> b
+  := (isinitial_zero_object b).1 $o (isterminal_zero_object a).1.
+
+Declare Scope pointedcat_scope.
+Local Open Scope pointedcat_scope.
+
+Notation "1" := zero_object : pointedcat_scope.
```
mikeshulman, line 19: Wait a minute... the zero object is called "1"?

theories/Pointed/pFiber.v
```diff
@@ -21,34 +21,87 @@ Definition pfib {A B : pType} (f : A ->* B) : pfiber f ->* A
...
-  - cbn.
-    apply concat_Vp.
+    symmetry; apply point_eq. }
+  cbn; apply concat_Vp.
 Defined.
 
-(** The triple-fiber functor is equal to the negative of the loopspace functor. *)
-Definition pfiber2_loops_functor {A B : pType} (f : A ->* B)
-: loops_inv _ o* pfiber2_loops f o* pfib (pfib (pfib f))
-  ==* loops_functor f o* pfiber2_loops (pfib f).
```
mikeshulman, line 38: I forget, where did this lemma move to (or what did it become)?

theories/WildCat/PointedCat.v
jdchristensen: More generally, I think we should avoid using natural numbers as notations for other things (and should undo some of the other instances of this in the library).

theories/HFiber.v
jdchristensen: I did a test, and it definitely causes Coq to duplicate its search through the sub-tree.  If you think about it, there's no efficient way for Coq to avoid this.  It would have to cache every state it has been in through the search, and then compare each new state to the states in the cache *up to definitional* equality, which is an expensive test in general.  Moreover, each redundant path *multiplies* the slow down.  So we definitely want to avoid redundant typeclass instances.  My test code is below.

```coq
Require Import HoTT.

Class Class1 := { dummy1 : nat }.

Class Class2 := { dummy2 : Empty }.

Class Class3 := { dummy3 : Type }.

Axiom ax12 : Class1 -> Class2.

Axiom ax23 : Class2 -> Class3.

Global Existing Instance ax12.

Global Existing Instance ax23.

Set Typeclasses Debug.

Definition test1 : Class1 -> Class3 := _.
(* 11 lines of output. *)

Definition test2 : Unit -> Class3.
Proof.
  refine _.
  (* 10 lines of output *)
Abort.

Global Instance ax12' : Class1 -> Class2 := _.

Definition test2 : Unit -> Class3.
Proof.
  refine _.
  (* 21 lines of output *)
Abort.

Global Instance ax23' : Class2 -> Class3 := _.

Definition test2 : Unit -> Class3.
Proof.
  refine _.
  (* 32 lines of output, 2^2 cases tried. *)
Abort.
```

Alizter: @jdchristensen Thank you for doing this test! I've always been unsure about how coq does this, and in this case it looks like my hunch was incorrect. I have lots of *questionable* uses of typeclasses which might be a problem in the future.

theories/Pointed/Loops.v
Alizter: I'm not sure why this was commented out. But since Dan pointed out above that we probably don't want to be writing redundant instances for speed reasons, I would probably lean on keeping this as a comment.

theories/WildCat/PointedCat.v
Alizter: I definitely wrote this. I'll get rid of this since it seems silly now. I agree with Dan about using "1" for notations. Though sometimes it makes path algebra easier to read and write.

theories/Pointed/pFiber.v
Alizter: This got accidentally removed it seems.

theories/WildCat/PointedCat.v
Alizter: Looking back it now, I just remembered why I called it "1". The example I had in mind was pType hence the zero object of pType is pUnit. I also called the zero morphism "0" so "1" seemed like a good name at the time.

It is pretty standard to call the zero morphism "0" when working with abelian groups. So I am not sure if we should remove that. The notation I used for the zero object however has no precedent that I am aware of so I will remove it.

mikeshulman:
I think it's much more common to call the zero object "0".  But I think
that would be a bad idea for us too.

theories/Homotopy/HomotopyGroup.v
Alizter: It doesn't seem to work anymore.

theories/Homotopy/HomotopyGroup.v
Alizter: Yes, I believe it is written that way to get the coercion to work. I will see if making pMap a notation is better.

Alizter: I've removed pMap and made `->*` a notation for pForall.

Alizter:
Interesting feature, underneath the reviewers listing on the side there is a "convert to draft button". It looks like we can now post-draft PR's!

I have drafted this one since I am not happy with the quality of the code in some places. There are various names which are very long/strange. Once I have had some time to mull over various things I'll publish this again. Please continue reviewing however!

theories/Pointed/pHomotopy.v
```diff
@@ -1,52 +1,60 @@
 Require Import Basics.
 Require Import Types.
 Require Import Pointed.Core.
+Require Import WildCat.
 
 Local Open Scope pointed_scope.
 
-(* pointed homotopy is a reflexive relation *)
-Global Instance phomotopy_reflexive {A B} : Reflexive (@pHomotopy A B).
+(** TODO: in various places we pointed_reduce. We should avoid doing so. *)
```
jdchristensen, line 8:
```suggestion
(** TODO: in various places we use pointed_reduce. We should avoid doing so. *)
```

theories/WildCat/PointedCat.v
jdchristensen: I think you should remove "0" for the zero morphism.  We end up having to annotation so many natural numbers with %scope that it's longer than typing something like zero.

theories/Pointed/Loops.v
Alizter: I'll uncomment this, perhaps it was an accident.

theories/Pointed/Core.v
```diff
@@ -9,113 +10,158 @@ Local Open Scope path_scope.
...
+  (h : f ==* g) : h (point A) @ dpoint_eq g = dpoint_eq f.
+Proof.
+  apply moveR_pM.
+  exact (dpoint_eq h).
+Defined.
+
+(** The underlying homotopy of a pointed homotopy between non-dependent pointed maps. *)
+Definition pointed_htpy' {A B} (f g : A ->* B) (p : pHomotopy f g)
+  : forall x:A, f x = g x
+  := fun x => pointed_htpy p x.
```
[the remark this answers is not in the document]
Alizter: I'm uncertain why this was added. It doesn't seem to be used anywhere.

theories/Pointed/Core.v
```diff
@@ -9,113 +10,158 @@ Local Open Scope path_scope.
...
-  exact (point A; P.2).
-Defined.
+(** Pointed sigma types *)
+Definition psigma {A : pType} (P : pFam A) : pType
+  := Build_pType (sig P) (point A; P.2).
 
-(* Pointed pi types, note that the domain is not pointed *)
-Definition pforall {A : Type} (F : A -> pType) : pType
+(** Pointed pi types, note that the domain is not pointed *)
+Definition pproduct {A : Type} (F : A -> pType) : pType
```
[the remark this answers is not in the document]
Alizter: I'm not sure really what to call this. This was formalized by me using the ideas in [arXiv:1311.4002](https://arxiv.org/abs/1311.4002). There are now two separate things which we would call "pointed forall". Now since `pForall` is probably the "real" forall here, I settled on `pproduct` since it is roughly a product of pType's indexed by A. These things don't have good names that I am aware of.

theories/Pointed/Core.v
```diff
@@ -135,16 +181,564 @@ Ltac pointed_reduce_pmap f
...
+
+(** ** Various operations with pointed homotopies *)
+
+Definition phomotopy_refl {A : pType} {P : pFam A} (f : pForall A P) : f ==* f
+  := Build_pHomotopy (fun x => 1) (concat_pV _)^.
+
+(** [pHomotopy] is a reflexive relation *)
+Global Instance phomotopy_reflexive {A : pType} {P : pFam A}
+  : Reflexive (@pHomotopy A P)
+  := phomotopy_refl.
```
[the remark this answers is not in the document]
Alizter: I think making it an existing instance later is the best practice.

theories/Pointed/Core.v
```diff
@@ -135,16 +181,564 @@ Ltac pointed_reduce_pmap f
...
+Definition path_pforall `{Funext} {A : pType} {P : pFam A} {f g : pForall A P}
+  : (f ==* g) -> (f = g) := equiv_path_pforall f g.
+
+Definition phomotopy_path `{Funext} {A : pType} {P : pFam A} {f g : pForall A P}
+  : (f = g) -> (f ==* g) := (equiv_path_pforall f g)^-1 % equiv.
+
+Definition phomotopy_path_path_pforall `{Funext} {A : pType} {P : pFam A}
+  {f g : pForall A P} (p : f ==* g)
+  : phomotopy_path (path_pforall p) ==* p
+  := phomotopy_path (eissect (equiv_path_pforall f g) p).
```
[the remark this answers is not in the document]
Alizter: I'll get rid of this since it isn't used anywhere.

theories/Pointed/Core.v
```diff
@@ -135,16 +181,564 @@ Ltac pointed_reduce_pmap f
...
+Definition path_pforall_phomotopy_path `{Funext} {A : pType} {P : pFam A}
+  {f g : pForall A P} (p : f = g)
+  : path_pforall (phomotopy_path p) = p
+  := eisretr (equiv_path_pforall f g) p.
+
+Definition equiv_path_pmap `{Funext} {A B : pType} (f g : A ->* B)
+  : (f ==* g) <~> (f = g) := equiv_path_pforall f g.
+
+Definition path_pmap `{Funext} {A B : pType} {f g : A ->* B}
+  : (f ==* g) -> (f = g) := equiv_path_pmap f g.
```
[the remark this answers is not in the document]
Alizter: Seems reasonable.

theories/Pointed/Core.v
```diff
@@ -135,16 +181,564 @@ Ltac pointed_reduce_pmap f
...
+  refine (equiv_functor_sigma' (equiv_idmap _) _); intros p.
+  refine (_^-1 oE equiv_path_inverse _ _).
+  apply equiv_moveR_1M.
+Defined.
+
+Definition path_pforall `{Funext} {A : pType} {P : pFam A} {f g : pForall A P}
+  : (f ==* g) -> (f = g) := equiv_path_pforall f g.
+
+Definition phomotopy_path `{Funext} {A : pType} {P : pFam A} {f g : pForall A P}
+  : (f = g) -> (f ==* g) := (equiv_path_pforall f g)^-1 % equiv.
```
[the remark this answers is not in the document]
Alizter: In the second definition, since you assume funext, why not prove the stronger assumption that they are equal?

theories/Pointed/Core.v
```diff
@@ -725,7 +725,7 @@ Proof.
   + intros ? ? p. exact (phomotopy_compose_Vp p).
 Defined.
 
-(** The forgetful map from pType to Type is a forgetful 0-functor *)
+(** The forgetful map from pType to Type is a 0-functor *)
```
jdchristensen, line 728: You missed the "is a forgetful 1-functor" a few lines later.

theories/Pointed/Loops.v
```diff
@@ -351,25 +366,25 @@ Defined.
 Local Notation "( X , x )" := (Build_pType X x).
 
 (* We can convert between families of loops in a type and loops in Type at that type. *)
-Definition loops_type `{Univalence} (A : Type)
-  : loops (Type,A) <~>* (A <~> A, equiv_idmap).
+Definition loops_type@{u0 u1 u2} `{Univalence} (A : Type@{u0})
+  : pEquiv@{u1 u1 u2} (loops (Type@{u0},A)) (A <~> A, equiv_idmap).
```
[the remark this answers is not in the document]
Alizter: Without them there are universe inconsistencies.

theories/Pointed/Core.v
jdchristensen: Ah, just saw the next commit.

theories/Algebra/Group.v
```diff
@@ -83,9 +83,10 @@ Class GroupHomomorphism (G H : Group) := Build_GroupHomomorphism' {
 Coercion grp_homo_map : GroupHomomorphism >-> Funclass.
 
 (* Group homomorphisms are pointed maps *)
-Definition pmap_GroupHomomorphism {G H : Group} (f : GroupHomomorphism G H) : pMap G H
+Definition pmap_GroupHomomorphism {G H : Group} (f : GroupHomomorphism G H)
+  : pForall G (pfam_const H)
```
jdchristensen, line 87: Does [G ->* H] work here?  (If so, it could fit on the previous line.)

theories/Pointed/Core.v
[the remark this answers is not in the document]
jdchristensen: Sounds fine.  It seems good to avoid having definitionally equal synonyms for no reason.

theories/Pointed/Core.v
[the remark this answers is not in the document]
jdchristensen: No particular reason.  The Funext is there so that equiv_path_pmap' is even defined.  And maybe it's good to not use it to promote == to = if we don't need =?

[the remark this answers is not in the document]
Alizter: We also have this map more generally in any wild category `GpdHom_path`. We can probably give it a better name. Having morphism extensionality (`HasMorExt`) means that this map is invertible.

theories/Homotopy/HomotopyGroup.v
mikeshulman: That's weird, any idea why?

theories/HFiber.v
mikeshulman: Good to know.  I guess then results like this should just be `Definition`s rather than `Instance`s?

I do think it's possible that there are situations in which declaring a theoretically-unnecessary instance may oppositely speed things up.  But in the absence of evidence in favor of that, we should probably not declare unnecessary instances.

theories/Pointed/Loops.v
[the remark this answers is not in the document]
mikeshulman: I think it's usually worth leaving a comment to that effect when it happens, for the benefit of the reader who may wonder the same thing.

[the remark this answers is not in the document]
Alizter: When I originally wrote this, I don't remember putting universe annotations. They seem to have been added later. I think we require more universes when working with pForall which might have caused coq some trouble here.

theories/HFiber.v
Alizter: It would be good to open an issue on this topic. At the moment my knowledge of how typeclasses works is basically a bunch of loosely held together heuristics.

theories/WildCat/PointedCat.v
Alizter: On the other hand we will only have that issue if you open pointedcat_scope which I'm not sure many people will do.

theories/Homotopy/HomotopyGroup.v
Alizter: I don't have any idea.

theories/HFiber.v
jdchristensen: > Good to know. I guess then results like this should just be `Definition`s rather than `Instance`s?

I think so.

> I do think it's possible that there are situations in which declaring a theoretically-unnecessary instance may oppositely speed things up. But in the absence of evidence in favor of that, we should probably not declare unnecessary instances.

I suspect that that might be true, e.g. when an Instance gives a short-cut that composes several other instances.  But in the case where two instances are essentially the same, I'm doubtful that there is any benefit, and think it's likely that things will be slower.  The slowdowns may be hard to measure with just one or two cases, but we are all now aware of how exponential growth works...

theories/Pointed/Loops.v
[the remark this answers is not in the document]
Alizter: I've named the universe parameters but I haven't yet put a comment explaining their necessity.

Alizter:
Also what exactly are we seeing here with pType? What do we call an "internal forall"? I seem to recall seeing a similar construction for spectra as well.

mikeshulman:
Basically, it's a CwF.  (-:

I've thought about whether a general structure of this sort could be
formulated internally, but always ended up concluding that it needs more
strictness than is internally expressible.  However, I never got to the
point of writing anything down, so I could very well be wrong.

jdchristensen:
Maybe `pproduct` could be changed to `forallp` or `pforallp` or `ppforall`?  `ppforall` would mean the Pointed type of Pointed maps, hence two p's.

Alizter:
@jdchristensen We use ppforall as a notation for ppForall which is the pointed type pForall. Things get very confusing :)

jdchristensen:
Ok, I finally see that there are four kinds of `forall`s, depending on whether the functions are pointed and the overall type of functions is pointed.  The current naming is:
```
                            type unpointed         type pointed
functions unpointed         forall                 pproduct
functions pointed           pForall                ppForall (with ppforall notation)
```
Can we come up with a pattern here?  E.g. we could put a p at the end when the type is pointed and at the start when the functions are pointed, giving `forall`, `forallp`, `pforall`, `pforallp`.  As far as I can think of, using the leading p in this way roughly matches our other uses, e.g. pType, pHomotopy.  The things *in* the type are pointed, but the type of those things is not.  Or maybe there's another scheme that doesn't leave one oddly named case?

(I'm not too concerned about the capitalization, which has more to do with using a notation to get binding.)

mikeshulman:
I think the intent in the name `pproduct` was not to view it as fitting in this table, but as a special case of the product of a family of objects in any (wild) category.  We haven't defined the latter yet, but once we do it will also include `forall`.  By contrast, `pForall` and its pointed version aren't as simply instances of a categorical construction.

jdchristensen:
@Alizter I think there are just a few things left to address, right?  The duplicate instance, the duplicate names for reflexivity, symmetry and transitivity of pointed homotopy, a comment about universe annotations, and maybe a few things I missed.  I also don't like the notation "0" being used for the zero morphism.  Is there anything else you plan to do?

Alizter:
There are some issues I encountered today. The universes of pForall now have to be larger (with respect to the previous pMap case) since we have a type family. I also had to explicitly define a coercion `pType@{i} -> pType@{j}` in order to lift pointed types to a bigger universe. Most of my universe issues arose when I was using lemmas like `local_global_looping`. I am not sure if this is going to be a big issue in the future but it might be something to keep note of.

For example, psphere in HIT/Spheres.v has way too many universes parameters at the moment. You can obviously modify the definition and force it to take a single parameter but it surprised me today.

There are a few places left in the library where I might do some renaming which will require me to read some more code in detail.

Hopefully tomorrow I can finish this and finish implementing the rest of the code suggestions.

mikeshulman:
It's a good thing we ditched the modules for modalities; we should be much
less fragile to extra universe parameters now.

theories/Pointed/Core.v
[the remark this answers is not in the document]
Alizter: @jdchristensen Do you still want me to add this map?

Alizter:
Here is an organization question. Should pTrunc and pSusp be kept in Pointed or be moved to Truncations and Susp respectively? We could probably also expand pTrunc to be about pointed types and reflective subuniverses in general, in which case we can do things like modally connected covers.

theories/Pointed/Core.v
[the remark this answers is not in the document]
jdchristensen: If it fits easily into the new framework, then I think it would be handy.

theories/HFiber.v
JasonGross: Another case where giving a redundant instance explicitly is when it changes the order in which that particular path is tried in the larger search tree.

My rule of thumb is that duplicate instances are fine iff their typeclass subproblems are expected to either never fail or else to fail immediately without invoking longer chains of typeclass search.

Alizter:
#1313 has broken things here but should be fixable with a rebase

theories/HFiber.v
jdchristensen: That seems reasonable, Jason, and I think it rules out this case, since there are lots of ways to try to prove `IsEquiv` and we have experience showing that they can take some time to fail.

jdchristensen:
This LGTM once that one redundant instance is changed to a Definition.

Alizter:
The last job for the CI keeps failing but I can't work out why. @JasonGross any idea what is going on?

mikeshulman:
We could consider making this case (and others like it) a `Hint Immediate
... : typeclass_instances`.

Alizter:
Hmm it seems something has changed in coq master which has caused our last job to fail. I don't think this is really an issue for us since we are targeting coq 8.11 and people using 8.12 are not our problem (yet).

JasonGross:
No, if this PR doesn't build on master, we should not merge it yet; the contract of developments on Coq's CI is that we don't break the build with master.

JasonGross:
Ah, but this seems to be a genuine issue with Coq master having broken the `install-devfiles` target, so maybe it's okay to merge anyway.  I've reported this as https://github.com/coq/coq/issues/12115.  We might be able to work around this by adding `-k` to `sudo make install-binaries install-devfiles` in `etc/ci/install_coq.sh`

PR #1313, opened 2020-04-13
Surreals: [No] is a letin so has no polymorphic universes

SkySkimmer:


Alizter:
Has this always been the case or was something changed in coq 8.10 (or before)?

spitters:
LGTM, but shouldn't (or can't) the type checker give a warning about this?

SkySkimmer:
It will be an error after https://github.com/coq/coq/pull/11978

spitters:
Thanks! Is this the only place in the hott library where this will raise an error?

SkySkimmer:
yes

Gaëtan Gilbert

mikeshulman:
Can someone actually explain what's going on here please?

SkySkimmer:
`No` is a section letin `Let No := GenNo S.` but for some reason it was used as `No@{i}` which makes no sense.

mikeshulman:
Why does it make no sense?

JasonGross:
In Coq `let`-binders cannot be universe-polymorphic.  (One reason is that if they were, then clearing the body of a `let`-binder would result in a lambda with non-prenex universe polymorphism, which is not supported.)  So the universes get fixed at `Let` declaration time.  If you want this to not happen, I think you can use `Notation` instead.

mikeshulman:
Ah, I see, thanks.  It still ends up polymorphic after the section is closed, but the universe is fixed for the entire section; right?

JasonGross:
Yes

PR #1316, opened 2020-04-19
Use pointed_reduce' and remove Funext hyp where possible

jdchristensen:
The pointed_reduce' tactic avoids rewrites, and in many cases works in place of pointed_reduce with little or no change.  Also, several results had unnecessary Funext hypotheses.

### DIFF
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -172,14 +172,14 @@ Defined.
 
 (** *** Homomorphic properties *)
 
-(** The following tactic often allows us to "pretend" that phi preserves basepoints strictly.  This is basically a simple extension of [pointed_reduce] (see Pointed.v). *)
+(** The following tactic often allows us to "pretend" that phi preserves basepoints strictly.  This is basically a simple extension of [pointed_reduce_rewrite] (see Pointed.v). *)
 Ltac grouphom_reduce :=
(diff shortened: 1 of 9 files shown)


jdchristensen:
The build failures have to do with installing dependencies...

Alizter:
@JasonGross Any idea what is going on with github actions? I'm guessing there are some difficulties with the servers hosting the packages for ubuntu?

Alizter:
Is there any point in keeping `pointed_reduce`? It seems that `pointed_reduce'` is doing most of the work anyway, so I would probably incline on calling that `pointed_reduce`. The only extra thing that is being done is `rewrite ?concat_1p, ?concat_p1` which isn't exactly difficult to remember or use. I don't think it needs to be a seperate tactic. This way we also don't have to worry about using pointed_reduce whenever since it is just a routine induction on the path data that comes with pointed stuff.

jdchristensen:
I think it's handy to have both, but maybe the one without rewrites should be named [pointed_reduce] and the one with rewrites should get a less convenient name, like [pointed_reduce_rewrite]?

Alizter:
I would be okay with that naming scheme.

jdchristensen:
Ok, unless anyone else objects, I'll add that renaming to this PR.

jdchristensen:
Rebased on #1319.

PR #1317, opened 2020-04-19
Prove = instead of ==* in some cases

jdchristensen:
This is on top of #1316, so just look at the second commit.

In many cases, we can prove that two pointed functions are equal rather than just pointed homotopic, without using Funext.  The proofs are similar or in some cases shorter.  The advantage of this is that we get a strictly stronger result.  The disadvantage is that it means that it's hard to predict whether various lemmas produce an equality or a homotopy, and you sometimes have to apply phomotopy_path to convert.  Is this a good idea?

### DIFF
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -224,9 +224,9 @@ Notation "p ^*" := (phomotopy_symmetric _ _ p) : pointed_scope.
 (** [pHomotopy] is a transitive relation *)
 Global Instance phomotopy_transitive {A B} : Transitive (@pHomotopy A B).
 Proof.
-  intros x y z p q.
+  intros f g h p q.
   snrefine (Build_pHomotopy (fun x => p x @ q x) _).
-  nrefine (dpoint_eq p @@ dpoint_eq q @ concat_pp_p _ _ _ @ _).
+  nrefine (dpoint_eq p @@ dpoint_eq q @ (concat_pp_p _ _ _ @ _)).
   nrapply whiskerL; nrapply concat_V_pp.
 Defined.
 
--- a/theories/Pointed/pHomotopy.v
+++ b/theories/Pointed/pHomotopy.v
@@ -8,19 +8,17 @@ Local Open Scope pointed_scope.
 (** Some higher homotopies *)
 
 Definition phomotopy_inverse_1 {A : pType} {P : pFam A} {f : pForall A P}
-  : (phomotopy_reflexive f)^* ==* phomotopy_reflexive f.
+  : (phomotopy_reflexive f)^* = phomotopy_reflexive f.
 Proof.
-  srapply Build_pHomotopy.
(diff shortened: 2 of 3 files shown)


jdchristensen:
I force pushed something with no changes to trigger a rebuild, but it failed again.

mikeshulman:
In general I tend to stick to homotopies, even when equality is possible,
unless there's a reason to expect that equality would be useful.  Do you
have any applications in mind for the equalities?

Alizter:
You don't need to force push to trigger a rebuild, when a job fails we have the option of restarting the jobs. If you click `Details` you will see there is a re-run jobs button under the edit button.

I'm not sure what is going on with the CI today, but we will find out soon enough.

jdchristensen:
I don't know yet if I have a specific need for equalities, so this is more of a general question.  Having equalities makes many other tools available, but if in practice it usually works out to have only homotopies, then maybe that makes sense.  It would be a shame, though, if we later used Funext to convert some of these homotopies to equalities when it turns out it isn't needed.  If we keep these as homotopies, would it be worth adding a comment to each one saying that it can be upgraded to an equality without Funext?

jdchristensen:
Actually, the first thing I just tried to do with the new pointed homotopy stuff doesn't work because of this issue.  I want to know that the map `phomotopy_path` from `loops (Y ->** Z)` to `Y ->* loops Z` is a map of magmas, which means that it preserves the composition operations up to path equality.  And that's essentially the `=` version of `phomotopy_path_pp`.  I can of course just prove the stronger version locally, since it's a one-line proof.

jdchristensen:
Maybe we should provide both versions, with the `==*` versions having the natural names and the occasional `=` versions have a suffix (a prime, or `_eq`, or something else)?

mikeshulman:
Yes, if we need both versions then I would lean towards providing both.

jdchristensen:
Ok, I'll work on this more after I experiment more and see what I need.  For now, I've converted it to a draft PR.  Any suggestions for the name of the `=` version of `phomotopy_path_pp`, for example?

Alizter:
@jdchristensen The name that comes to mind is `path_phomotopy_path_pp` but it might be a bit silly.

mikeshulman:
I don't know.  Ideally I think a suffix would be better than a prefix, so that it's more clearly a version of the same lemma.  Perhaps `_aspath`?

mikeshulman:
oops

jdchristensen:
I agree with the suffix preference, and think `_aspath` is a nice idea.

mikeshulman:
Is this still in-progress?

jdchristensen:
At some point, some changes like this are going to get merged from the Hurewicz work-in-progress, but I had been waiting to see what was needed.  Right now the Hurewicz work is on hold while I teach covid-style, so in the meantime I think I'll just close this.

PR #1320, opened 2020-04-20
make coqocaml rather than coqlight

JasonGross:
As per https://github.com/coq/coq/issues/12115#issuecomment-616498778

Hopefully fixes the CI issue on master.

Closes coq/coq#12115


Alizter:
Thanks for the fix!

PR #1321, opened 2020-05-22
Adding characterization of path spaces in sequential colimits

kristinas:


jdchristensen:
It's great to see this!  Thanks for doing it.

I'm not sure if I'll have time to properly review it soon, so I'll just make a few very general comments.  First, it would be helpful if there were more comments in the file.  A long comment near the top could explain the overall goals of the file and the structure of the argument, and other comments throughout could help the reader see what is coming next.

Very minor things:  I prefer to avoid doubled bullets like `--` and `**` unless the single bullets are all used up.  And I have a personal dislike for using `1` for `idpath`, so I'd prefer to see `reflexivity` instead of `exact 1`.

Alizter:
@jdchristensen The argument is written up here for reference: https://www.cs.cornell.edu/~ks858/papers/sequential_colimits_homotopy.pdf

jdchristensen:
It would be good to cite the paper in the .v file and reference theorem numbers from the paper at various points in the formalization.

kristinas:
Okay, will do. There is an encode_decode lemma that should probably be placed somewhere else (it's about the contractibility of a total space). Also, there are two lemmas about adding natural numbers. There are very similar lemmas in the nat.v file, but there they end with Qed so I cannot actually use their computational properties.

As an aside, it appears that if I say Local Definition, then the definition is also accessible from outside the Section where I define it, whereas if I say Let then this is not the case. I intended the local definitions to be local to just the sections where they are defined, but Ali observed that if I use Let instead, then the code is really slow.

kristinas:
What kind of bullet should I use after +, -, * have been used up?

mikeshulman:
If I ever end up with more than three levels of nested bullets, I usually think about refactoring the proof into lemmas.  (-:  Another possibility that I often use is nested braces `{ ... }`.

mikeshulman:
You can feel free to add or modify stuff in `Spaces/Nat.v` too.

mikeshulman:
The meaning of "Local" in Coq is inconsistent across the commands that it can modify.  According to [the refman](https://coq.inria.fr/refman/language/gallina-specification-language.html#coq:cmdv.local-definition), a `Local Definition` means local to the current *module* (in the sense of "never imported even when the module is imported), not the current section.  I think this serves the same purpose of keeping it out of the global namespace so that you can feel free to use short names, though.

jdchristensen:
Not sure why this was closed and another PR was opened.  I think it would be better to reopen this one, so the discussion stays in one place.

kristinas:
I don't know how to update this pull request.

jdchristensen:
Since you created the pull request from your master branch, every commit to your master branch gets added to the pull request.  So it already includes your updates to the comments.  (Thanks for doing that!)  In the future, it's better to create a separate branch for a pull request, so that you can also work on other things without interfering with it.

To reopen this pull request, I think you first need to close the new one.

theories/Colimits/Sequential.v
```diff
@@ -223,116 +407,88 @@ Section FibSequence.
...
-      {psi : coe (ap Y p) = F} {q} (theta : I psi (apD (fun x y => G (x;y)) p) = q) y
-      : apD G (Delta p psi y) = (q y)^.
-    Proof.
-      destruct p; destruct psi; destruct theta; exact 1.
-    Defined. 
-
-    (** The point-point case of the nested induction. *)
-    Local Definition Q k : forall n a b, E (inj _ n a; inj _ k b).
+    (** The point-point case of the nested induction; corresponds to "h" in the paper. *)
+    Let Q k : forall n a b, E (inj _ n a; inj _ k b).
```
jdchristensen, line 434: Is there a reason to use `Let` instead of `Local Definition`?  I'm not really sure how they differ or affect the proof term, but it seems odd to use `Let` when a `Proof` is needed.

theories/Colimits/Sequential.v
```diff
@@ -223,116 +407,88 @@ Section FibSequence.
...
-        srefine (_ @ ap _ (ap _ (Q_beta _ n a _))).
-        srefine (ap _ ((transport_compose _ _ _ _)^ @ _)^).
-        exact (apD _ (Beta _ b)).
+      - intros n a b; srefine (_ @ ap _ (h n.+1 (f n a) b)).
+        srapply (Epsilon (glue A n a) (colimitSuccSeqToColimitSeq_beta_glue _ _ _)).
     Defined.
 
-    (** The point case of the nested induction. *)
-    Local Definition F n a : forall x, E (inj _ n a; x).
+    (** The point case of the nested induction; correcponds to "g" in the paper. *)
```
jdchristensen, line 455: Typo: "correcponds"

theories/Colimits/Sequential.v
```diff
@@ -372,45 +528,105 @@ Section FibSequence.
...
       srapply (whiskerL _ _ @ concat_p1 _); rewrite (ap_compose _ _ _)^; simpl.
-      rewrite ap_const; exact 1.
+      rewrite ap_const; reflexivity.
   Defined.
 
 End FibSequence.
+
+(** Lemma 7.3; this probably belongs to a different file. *)
+Definition encode_decode `{Funext} {A} {B : A -> Type} a (b : B a) :
+  Contr (sig B) <~> forall x, IsEquiv (fun p : a = x => p # b).
```
jdchristensen, line 577: One direction of this is `equiv_path_from_contr` in PathAny.v, although it doesn't identify the equivalence with transport.  I think this could go in that file, either before `equiv_path_from_contr`, with `equiv_path_from_contr` as a trivial consequence, or after `equiv_path_from_contr`, using it for the proof (and maybe updating it to be more specific about the equivalence).

theories/Colimits/Sequential.v
```diff
@@ -372,45 +528,105 @@ Section FibSequence.
...
+    := Build_Sequence (fun k => a1^++k = a2^++k) (fun k p => ap (f k) p).
+
+  Let B : FibSequence A.
+  Proof.
+    srapply Build_FibSequence.
+    - intros [k b]; exact (a1^++k = b).
+    - intros [k b]; exact (ap (f k)).
+  Defined.
+ 
+  Let g := fibSequenceArr B.
```
jdchristensen, line 606: This is unused.

theories/Colimits/Sequential.v
```diff
@@ -1,86 +1,263 @@
...
+  Defined.
+
+  Local Definition nat_plus_n_O n : (n + 0)%nat = n.
+  Proof.
+  induction n as [ | n p].
+    - reflexivity.
+    - exact (ap S p).
+  Defined.
+
+  Local Definition nat_plus_n_Sm n m : (n + m.+1)%nat = (n + m)%nat.+1.
```
jdchristensen, line 183: If the above two are elsewhere but are Qed, then you should see if it works to simply make them Defined in the other location.

theories/Colimits/Sequential.v
```diff
@@ -372,45 +528,105 @@ Section FibSequence.
...
     - exact (fun n a => 1).
     - intros n [a b]; rewrite concat_1p, concat_p1, ap_compose, !Colimit_rec_beta_colimp.
-      rewrite ap_pp, (Delta_proj _ (fibSequenceToTypeFam_beta n a)).
+      rewrite ap_pp, (Delta_proj _ (fibSeqToTypeFam_beta_glue n a)).
       srapply (whiskerL _ _ @ concat_p1 _); rewrite (ap_compose _ _ _)^; simpl.
-      rewrite ap_const; exact 1.
+      rewrite ap_const; reflexivity.
   Defined.
 
 End FibSequence.
```
jdchristensen, line 573: This line is the slowest in the file (~1.8s).  It would be nice to figure out a way to speed it up.  Not having a Section at all is one option.  I've heard that using Qed for some results can help, as can avoiding cbn/simpl.  Does changing from Let's to Definition's help?

theories/Colimits/Sequential.v
```diff
@@ -1,86 +1,263 @@
...
+    intros p q; srefine ((concat_p1 _)^ @ _); srefine (_ @ (concat_1p _)); exact q^.
+  Defined.
+
+  (** The successor sequence from Lemma 3.6. *)
+  Definition succSeq : Sequence
+    := Build_Sequence (fun k => A k.+1) (fun k => f k.+1).
+
+  (** The shifted sequence from Lemma 3.7. *)
+  Definition shiftSeq n : Sequence
+    := Build_Sequence (fun k => A (k+n)%nat) (fun k => f (k+n)%nat).
```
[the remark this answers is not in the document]
mikeshulman: What about (outside the `Section`)
```coq
Fixpoint shiftSeq n (A : Sequence) : Sequence
  := match n with
     | 0 => A
     | S n' => succSeq (shiftSeq n' A)
     end.
```
?  (I haven't checked whether any of the rest of the file would work this way.)

theories/Colimits/Sequential.v
mikeshulman: As for how they differ, my understanding is that when the `Section` is closed, a `Let` no longer exists but becomes a `let`-binding in the proof-terms of everything else defined in the section.  Whereas a `Local Definition` is a definition like any other, and actually pays no attention to sections: what `Local` means in that context is that it doesn't pollute the namespace of other *modules* that import the current module (or the current file, if there is no other current module) -- but can still be referred to everywhere by its FQ name.

theories/Colimits/Sequential.v
```diff
@@ -1,86 +1,263 @@
...
 Local Open Scope path_scope.
 
-Section Lift.
+Section Sequence.
 
-  Context {A : Sequence}.
+  Context (A : Sequence).
 
-  Local Definition f n := @arr _ A n _ 1.
+  Let f n := @arr _ A n _ 1.
```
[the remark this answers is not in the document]
kristinas: Introduced a notation.

kristinas:
Thanks Dan! I added the Corollary about sequential colimits preserving truncation levels as you suggested.

theories/Colimits/Sequential.v
```diff
@@ -1,86 +1,263 @@
+(** We present a proof of the conjecture that sequential colimits in HoTT apropriately
+    commute with Σ-types. As a corollary, we characterize the path space of a sequential
+    colimit as a sequential colimit of path spaces. For the written account of these results
+    see https://www.cs.cornell.edu/~ks858/papers/sequential_colimits_homotopy.pdf. *)
```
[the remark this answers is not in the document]
kristinas: I'm not entirely sure what to do about this.

theories/Colimits/Sequential.v
kristinas: I actually only need the one direction so I replaced the usage of the encode-decode lemma by equiv_path_from_contr.

theories/Colimits/Sequential.v
kristinas: The corresponding lemmas in nat.v are the symmetrical versions, e.g., nat_plus_n_Sm n m has type (n + m)%nat.+1 = (n + m.+1)%nat. This is very inconvenient for my code and it's easier to just duplicate the definitions.

theories/Colimits/Sequential.v
kristinas: I removed the majority of the sections according to your suggestions,

theories/Colimits/Sequential.v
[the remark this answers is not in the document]
kristinas: Why do we want to unify these? The successor sequence is an instance of adding a numeral n (in this case 1) on the left, so that the sequence becomes k |-> A (n + k) [this only makes sense for numerals]. 

The shift sequence is an instance of adding a natural number on the right, so that the sequence becomes k |-> A (k + n) [this makes sense for arbitrary n:N].

theories/Colimits/Sequential.v
Alizter, line 4:
```suggestion
(** We present a proof of the conjecture that sequential colimits in HoTT apropriately commute with Σ-types. As corollary, we characterize the path space of a sequential colimit as a sequential colimit of path spaces. For the written account of these results see https://www.cs.cornell.edu/~ks858/papers/sequential_colimits_homotopy.pdf. *)
```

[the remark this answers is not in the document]
jdchristensen: Make it one long line.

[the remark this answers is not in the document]
Alizter: I added a suggestion comment at the end of this thread.

theories/Colimits/Sequential.v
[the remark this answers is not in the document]
jdchristensen: If they are unified, then you only need to prove things about the more general case.  The way it is now, you have duplicated proofs, one handling 1 + n and one handling n + k, which makes it harder for the reader.  Now maybe there is no good way to unify them that works with the rest of the file, in which case, that's fine.  But if there is a way, then it would be good to use it.

theories/Colimits/Sequential.v
```diff
@@ -1,416 +1,600 @@
...
+          { intros A trH a b.
+            srefine (trunc_equiv' _ (equiv_inverse (equiv_concat_l (glue A _ a) _))).
+            srapply (@trunc_equiv' _ _ _ k (IHm (succ_seq A) _ (@arr _ A 0%nat _ 1%path a) b)).
+            srapply (equiv_ap (colim_succ_seq_to_colim_seq A)). }}
+        { intros n m p b; srapply path_ishprop; srapply ishprop_istrunc. }
+      * intros A trH a; srapply (functor_forall_equiv_pb (colim_succ_seq_to_colim_seq A)).
+        intro x; srapply (@trunc_equiv' _ _ _ k (IHn (succ_seq A) _ a x)); srapply equiv_ap.
+    + intros n m p a; srapply path_ishprop; srapply trunc_forall.
+      intro x; srapply ishprop_istrunc.
+Defined.
```
jdchristensen, line 600: You should put a newline at the end of the file.

jdchristensen:
The last commit is impossible to review.  First, you accidentally committed changes to hundreds of files in submodules.  You need to undo the commit and commit only the relevant file.

Second, you combined a change that affects the indentation of most of the file with changes to content, so the diff gives us no information about the changes to the content.  Changes to indentation should always be committed on their own, and other changes should be committed in logical pieces.

theories/Colimits/Sequential.v
jdchristensen: I had thought that you should be able to use the existing definitions, followed by `^` to reverse the path, but I see that that doesn't work without further changes.  But thinking about it more, I think the directions that you have make more sense.  And the corresponding lemmas in Nat.v aren't used anywhere else.  So I suggest changing the lemmas in Nat.v to go in your direction and to end with Defined.

jdchristensen: (Or will this change be a pain to *users* of the HoTT library?)

mikeshulman: Why not include both directions in Nat.v?

theories/Colimits/Sequential.v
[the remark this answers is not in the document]
mikeshulman: Did anyone try my suggestion?

theories/Colimits/Sequential.v
jdchristensen: Because we don't usually state `lemma` and `lemma^` in the library, since the second one is redundant.  Imagine if many proofs of equality also had the inverse proved separately...  In this PR, the computational behaviour of the reversed lemmas works with the current proof.  It probably wouldn't be too hard to make it work with the original lemmas as stated in Nat.v, but I'm not sure.  I do think that the reversed lemmas are the more natural ones, since we usually state lemmas in the direction that one would typically use them with `rewrite`.

theories/Colimits/Sequential.v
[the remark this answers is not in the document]
jdchristensen: I think I did, and the rest of the file didn't work.  But I suspect it could be adapted to use your method.

theories/Colimits/Sequential.v
mikeshulman: Not usually, but sometimes we do.  For instance, we have `concat_pp_p` and `concat_p_pp`.  If we want to include this direction for purposes of this PR, but also keep the old directions around for the benefit of not breaking the code of possible users of the library, I don't think there's anything wrong with that.

jdchristensen: I think it should be avoided.  For example, as currently defined, we don't know that `concat_pp_p` and `concat_p_pp` are inverses of each other.  At the very least, if we do this, I think we should define the second one to be `^` applied to the first one.  In the current situation, since the existing lemmas are defined with `Qed`, there would be no harm if they were defined to be `^` applied to the lemmas in this PR.  I'd still prefer to not have such redundancy unless it is really needed, as it makes it harder to choose names and find things (with `concat_pp_p` and `concat_p_pp` being a counterexample, though).

So my preference is to change this PR to use the existing lemmas (with `Defined`), but I would be ok with the new lemmas being moved to Nat.v, and having the old lemmas defined in terms of the new ones using `^`.

kristinas:
But using the existing lemmas is actually more work than redefining them
locally (as something that isn’t exported) so I would prefer to completely
avoid having to use them in their current direction.

theories/Colimits/Sequential.v
```diff
@@ -1,416 +1,600 @@
...
+    + intro n; revert trH; revert A; induction n as [ | n IHn].
+      * intros A trH a; srapply Colimit_ind.
+        { intros m b; revert b; revert a; revert trH; revert A; induction m as [ | m IHm].
+          { intros A trH a b.
+            srefine (trunc_equiv' _ (equiv_inverse (equiv_path_colim _ a b))). }
+          { intros A trH a b.
+            srefine (trunc_equiv' _ (equiv_inverse (equiv_concat_l (glue A _ a) _))).
+            srapply (@trunc_equiv' _ _ _ k (IHm (succ_seq A) _ (@arr _ A 0%nat _ 1%path a) b)).
+            srapply (equiv_ap (colim_succ_seq_to_colim_seq A)). }}
+        { intros n m p b; srapply path_ishprop; srapply ishprop_istrunc. }
```
jdchristensen, line 595: This line is a lot faster with `snrapply`:
```suggestion
        { intros n m p b; snrapply path_ishprop; snrapply ishprop_istrunc. exact _. }
```
The `exact _` finds `Funext`.

theories/Colimits/Sequential.v
```diff
@@ -1,416 +1,600 @@
...
+      * intros A trH a; srapply (functor_forall_equiv_pb (colim_succ_seq_to_colim_seq A)).
+        intro x; srapply (@trunc_equiv' _ _ _ k (IHn (succ_seq A) _ a x)); srapply equiv_ap.
+    + intros n m p a; srapply path_ishprop; srapply trunc_forall.
```
jdchristensen, line 598: Similar here:
```suggestion
    + intros n m p a; snrapply path_ishprop; snrapply trunc_forall. 1: exact _.
```

theories/Colimits/Sequential.v
jdchristensen: (I'm keeping this in the correct thread, but am responding to Kristina's message below.)

If using the existing lemmas is hard, then I think your lemmas should go in Nat.v, with the existing ones defined using them.

kristinas: Since this seemed to be a sticky point, I decided to reuse the existing lemmas in their current form even though it resulted in a bit more work.

theories/Colimits/Sequential.v
kristinas: Why doesn't the snrapply find Funext by itself?

kristinas:
Unfortunately I really don't know how to proceed with this PR; I don't know why the changes to other modules happened - I never modified anything in other files - or how to undo them. I tried to fork the repository anew but it doesn't seem to have done anything. I find git incredibly difficult to work with .

theories/Colimits/Sequential.v
jdchristensen: The n means to not do typeclass inference, and when that isn't done, Funext isn't found.  But without the n, Coq tries to figure out other arguments using typeclass search, which in this case takes a bit of time.  So sometimes we tell it to not use typeclass search, and then help fill in some parts.

jdchristensen:
First, save a copy of Sequential.v somewhere else for safe keeping.  Then, if you still just have the one bad commit in your repo, you can do `git reset HEAD~1` to remove it.  The files should stay unchanged.  Then you should be able to do `git commit Sequential.v` to commit only that file.  This doesn't address the fact that several things were changed at once there, but I think we can live with that.  `git log -p` can be used to verify that you committed the changes you thought you committed. 
 It's always good to check before pushing to the remote.  Then `git push -f` should force push the new commit to github in place of the old one.

kristinas:
That didn't work; it kept saying I cannot commit because there needs to be a pull first but when I tried the pull that failed too. I deleted the repo and forked it again but now I cannot even get Coq to run. Following the instructions in the library, I get the following error:

krist@LAPTOP-06DP60CA ~/HoTT
$ etc/install_coq.sh
$ git submodule sync
Synchronizing submodule url for 'coq-HoTT'
Synchronizing submodule url for 'etc/coq-dpdgraph'
Synchronizing submodule url for 'etc/coq-scripts'
$ git submodule update --init --recursive
~/HoTT/coq-HoTT ~/HoTT ~/HoTT/etc ~/HoTT
$ ./configure -local
      0 [main] ocamlrun 1420 child_info_fork::abort: address space needed by 'd
lunix.so' (0x400000) is already occupied
      0 [main] ocamlrun 1421 child_info_fork::abort: address space needed by 'd
lunix.so' (0x400000) is already occupied
      0 [main] ocamlrun 1422 child_info_fork::abort: address space needed by 'd
lunix.so' (0x400000) is already occupied
      1 [main] ocamlrun 1423 child_info_fork::abort: address space needed by 'd
lunix.so' (0x400000) is already occupied
      0 [main] ocamlrun 1424 child_info_fork::abort: address space needed by 'd
lunix.so' (0x400000) is already occupied
I can not automatically find the name of your architecture.
Give me a name, please [win32 for Win95, Win98 or WinNT]:

jdchristensen:
@kristinas  Sorry, I have no idea what's going on there.  If you post your OS version, maybe someone else can help you out?

Alizter:
@kristinas You are on windows using cygwin if I remember correctly? Windows users do not need to run the `etc/install_coq.sh` script since that will build coq locally. This cannot be done in cygwin so my solution was to use the prebuilt windows binaries.

This is the version of coq you should be using:
https://github.com/coq/coq/releases/tag/V8.11.2
Download and install these. These are just regular windows packages nothing to do with cygwin.

If you have the windows binaries for coq installed, your configure step should look like this: `./configure COQBIN="/cygdrive/c/Coq/bin/"`. Afterwards, you should be able to run make in (the HoTT directory) with no issues.

kristinas:
Thanks Ali, that fixed the problem. I have made a new commit,

jdchristensen:
@kristinas Did you forget to push (-f) the new commit?

jdchristensen:
No, I think the issue is that you committed to a new fork of the repo, which doesn't seem to be linked to the pull request anymore.  I don't know how to fix this.

Alizter:
@kristinas Since you've deleted your fork, this PR has become orphaned. The way to fix this will be to open another PR. We can close this PR once the other one is merged.

Alizter:
In the future try to use *branches*. You can create as many of these and delete them. This is what I would typically use for a PR, I would create a new branch locally on my machine, commit my changes, push to github (which would create a branch there), then from github I would open a new PR merging my branch into HoTT/master.

kristinas:
Thanks Ali. Why is it so complicated though? I didn't even modify anything outside that one file I was working on - at least not that I know of.

kristinas:
I created a new branch and used it to submit a PR

PR #1323, opened 2020-05-28
[Exercise 1.4] Write Fixpoint for iterN, and define recN in terms of it

artagnon:
I have used two well-founded Fixpoints for defining `iterN` and `recN`, but I have no idea if I've answered the question in the book correctly.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -124,7 +124,21 @@ End Book_1_3_sig.
 (* ================================================== ex:iterator *)
 (** Exercise 1.4 *)
 
+Section Book_1_4.
+  Variable C : Type.
 
+  Fixpoint iterN (c0 : C) (cs : C -> C) (n : nat) : C :=
+  match n with
+  | O => c0
+  | S n => cs(iterN c0 cs n)
+  end.
+
+  Fixpoint recN (c0 : C) (cs : nat -> C -> C) (n : nat) : C :=
+  match n with
+  | O => c0
+  | S n => cs n (recN c0 cs n)
+  end.
+End Book_1_4.
 
 (* ================================================== ex:sum-via-bool *)
 (** Exercise 1.5 *)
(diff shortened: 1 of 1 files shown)


mikeshulman:
Exercise 1.4 says to *assume as given* the iterator and *use this only* to construct the recursor.  Not to define both of them as fixpoints or using the induction principle.

artagnon:
Indeed, and I found [this](https://math.stackexchange.com/questions/2488061/defining-the-recursor-for-natural-numbers-using-iterator-hott-book-exercise#) elaboration. I'll work on it, thanks.

mikeshulman:
I don't think that's right.  Try evaluating it on some examples.

artagnon:
```coq
  Let C := nat.
  Compute recN (S O) (fun n x => x + n) (S (S O)).
```

computes to `5`. Isn't that the expected result?

mikeshulman:
No.

jdchristensen:
No, that should produce 2.  More generally, `recN c0 cs 2` should produce `cs 1 (cs 0 c0)`.  Notice that `cs` is called with different first arguments each time, and is never called with first argument 2.

Maybe you need to read Section 1.9 of the book more closely before attempting this exercise?

jdchristensen:
The new version is wrong in multiple ways:  it calls `cs` with a constant first argument instead of a decreasing first argument;  it calls `cs` 2n times instead of n times;  and the inner function ignores its argument.  Please unroll your next definition by hand to make sure, for example, that `recN c0 cs 5` produces `cs 4 (cs 3 (cs 2 (cs 1 (cs 0 c0))))`.  And then test on an example.

In fact, the exercise says "Show that the defining equations of the recursor hold propositionally for this function, using the induction principle for N."  This will check that you've got it right, so you should do this before pushing another commit.

Solving this exercise requires an extra idea, to carry the needed information through the iteration.

mikeshulman:

PR #1324, opened 2020-05-28
[Exercise 1.5] Completed with lots of help from @herbelin

artagnon:
~~Seems simple enough, but I wonder if I've addressed the question in the book correctly.~~

Edit: I think I've solved it correctly this time.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -129,7 +129,26 @@ End Book_1_3_sig.
 (* ================================================== ex:sum-via-bool *)
 (** Exercise 1.5 *)
 
+Section Book_1_5.
+  Definition sum (A B : Type) := { x : Bool & if x then A else B }.
 
+  Notation "'inl' a" := (true; a) (at level 0).
+  Notation "'inr' b" := (false; b) (at level 0).
+
+  Definition ind (A B : Type) (C : sum A B -> Type) (f : forall a, C (inl a))
+   (g : forall b, C (inr b)) : forall x : sum A B, C x := fun x => match x with
+   | inl a => f a
+   | inr b => g b
+   end.
+
+  Theorem inl_red {A B : Type} {C : sum A B -> Type} f g { a : A }
+  : ind A B C f g (inl a) = f a.
+  Proof. reflexivity. Defined.
+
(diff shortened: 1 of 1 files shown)


artagnon:
Thanks also to @SkySkimmer for helping turn the definitions of `inl` and `inr` into notations that work.

contrib/HoTTBookExercises.v
```diff
@@ -129,7 +129,26 @@ End Book_1_3_sig.
...
+  Notation "'inl' a" := (true; a) (at level 0).
+  Notation "'inr' b" := (false; b) (at level 0).
+
+  Definition ind (A B : Type) (C : sum A B -> Type) (f : forall a, C (inl a))
+   (g : forall b, C (inr b)) : forall x : sum A B, C x := fun x => match x with
+   | inl a => f a
+   | inr b => g b
+   end.
+
+  Theorem inl_red {A B : Type} {C : sum A B -> Type} f g { a : A }
```
Alizter, line 144: I would advise against making these opaque with Qed. Since these equalities are true by definition we would like to see that reflexivity is the proof term. If we have to later reason about `inl_red` then we would have trouble since it is opaque. Changing Qed to Defined should do the trick. In general we tend not to use Qed in the library unless we know that it absolutely is not going to be unfolded. Typically this applies for paths that are hprops.

Of course since this is an exercise this shouldn't be used anywhere else, but if people will read these exercises they might as well see better practices.

jdchristensen: I would also change `auto` to `reflexivity`, so the reader can see immediately that these are true by definition.

jdchristensen:
Once those two comments are addressed, it LGTM.

artagnon:
Thanks @Alizter and @jdchristensen.

jdchristensen:
LGTM.  While working on a pull request, it's usually best to add additional commits rather than replacing commits.  If needed, things can be squashed or cleaned up at the end.

artagnon:
Noted.

mikeshulman:
Note that `Section`s in Coq do not do namespacing (that's what `Module`s are for).  So I think it would be better not to clobber global names like `sum` and `ind`.

PR #1326, opened 2020-05-28
[Exercise 1.11] Can be solved easily with some insight

artagnon:
~~Not sure if this provides any insight.~~


Alizter:
I think it would be better to provide some insight. Here is a proof that I think would be more beneficial to people reading this later:
```coq
  Theorem dblneg : forall A, ~~~A -> ~A.
  Proof.
    intros A f a.
    apply f.
    intros g.
    apply g.
    exact a.
  Defined.
```

artagnon:
@Alizter Thanks!

jdchristensen:
LGTM.

PR #1331, opened 2020-06-04
[Exercise 2.11] Solved using equiv_compose

artagnon:
~~Again, I'm not sure if I've answered the question in its entirety.~~

I've modified the solution to use `IsPullback`, and it looks obviously correct this time around.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -484,7 +484,14 @@ End TwoTen.
 (* ================================================== ex:pullback *)
 (** Exercise 2.11 *)
 
+(* Already solved as pullback_corec; we copy it here with different variable
+names for completeness *)
 
+Section Book_2_11.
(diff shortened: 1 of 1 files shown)


mikeshulman:
`IsPullback` is the *statement* that a given commutative square is equivalent to "the" concretely defined pullback.  This is not the same as the definition of "pullback square" given in Ex. 2.11.

Alizter:
I think exercise 2.11 is asking if the square obtained from the canonical pullback `pullback_commsq` satisfies `IsPullback` which should be evident.

mikeshulman:
Exercise 2.11 is asking if `pullback_commsq` satisfies *a* definition of "being a pullback", but that definition is different from our `IsPullback`.

artagnon:
I wasn't able to figure out how to use `pullback_commsq`, but the solution hopefully looks better now?

mikeshulman:
No, you still aren't actually engaging with the definition of "pullback square" *as given in the exercise*.

mikeshulman:

PR #1332, opened 2020-06-04
[Exercise 2.12] Solve using primitives in Pullback.v

artagnon:
This is a work-in-progress, and I wonder if this is the most
straightforward definition to supply proof for.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -526,7 +526,7 @@ End TwoTen.
 (* ================================================== ex:pullback-pasting *)
 (** Exercise 2.12 *)
 
-
+Definition Book_2_12 := @HoTT.Limits.Pullback.ispullback_pasting_outer.
 
 (* ================================================== ex:eqvboolbool *)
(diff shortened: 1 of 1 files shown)


Alizter:
No, we don't have this in the main library. I did prove it a few times before, but I never managed to merge those branches.

Alizter:
This is the key thing to use when proving the 2-pullbacks lemma. Egbert mentioned this proof to me.

Alizter:
#1370

Alizter:
@artagnon We have the pasting lemma for pullbacks in Pullback.v now.

artagnon:
On second thought, I think `ispullback_pasting_outer` is proving something subtly different from the problem statement.

mikeshulman:
Right; that's what I was saying at #1331.  `ispullback_pasting_outer` *looks* like the statement of exercise 2.12, but is different because the meaning of "is a pullback square" defined in exercise 2.11 is different from `IsPullback`.

mikeshulman:

PR #1333, opened 2020-06-08
[Exercise 1.12] All three parts are relatively straightforward

artagnon:
The third part seems to be somewhat tricky.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -211,7 +211,7 @@ Definition Book_1_10 := ack.
 (** Exercise 1.11 *)
 
 Section Book_1_11.
-  Theorem dblneg : forall A, ~~~A -> ~A.
+  Theorem dblneg : forall A, (~~~A) -> ~A.
   Proof.
     intros A f a; apply f.
     intros g; apply g.
@@ -222,7 +222,26 @@ End Book_1_11.
 (* ================================================== ex:tautologies *)
 (** Exercise 1.12 *)
 
+Section Book_1_12.
+  Theorem Book_1_12_part1 : forall A B, A -> (B -> A).
+  Proof.
+    intros ? ? a ?.
+    exact a.
+  Defined.
 
+  Theorem Book_1_12_part2 : forall A, A -> ~~A.
+  Proof.
+    intros A a f.
+    exact (f a).
+  Defined.
+
+  Theorem Book_1_12_part3 : forall A B, ((~A) + (~B)) -> ~(A * B).
+  Proof.
+    intros A B [na | nb] [a b].
+    - exact (na a).
+    - exact (nb b).
+  Qed.
+End Book_1_12.
 
 (* ================================================== ex:not-not-lem *)
 (** Exercise 1.13 *)


contrib/HoTTBookExercises.v
```diff
@@ -222,7 +222,24 @@ End Book_1_11.
...
+  Theorem Book_1_12_part1 : forall A B, A -> (B -> A).
+  Proof.
+    intros ? ? a ?.
+    exact a.
+  Defined.
 
+  Theorem Book_1_12_part2 : forall A, A -> ~~A.
+  Proof.
+    intros ? b c.
+    destruct c. exact b.
```
mikeshulman, line 235: I would probably say `apply c` rather than `destruct` here.  In fact I'm surprised that `destruct` works.

jdchristensen: Most clear might be something like
```coq
  intros A a f.
  exact (f a).
```

jdchristensen:
For part 3, `split` might be the tactic you want.

artagnon:
`split` doesn't seem to work as `A + (~ B)` is not an inductive goal with 1 constructor. I'm currently stuck at:

```coq
    intros A B f g.
    destruct g. apply f.
```

contrib/HoTTBookExercises.v
```diff
@@ -231,13 +231,15 @@ Section Book_1_12.
 
   Theorem Book_1_12_part2 : forall A, A -> ~~A.
   Proof.
-    intros ? b c.
-    destruct c. exact b.
+    intros A a f.
+    exact (f a).
   Defined.
 
   Theorem Book_1_12_part3 : forall A B, (~A + ~B) -> ~(A * B).
```
jdchristensen, line 238: It turns out that this exercise is stated incorrectly.  There should be parentheses around `~A`.  (Or maybe the precedence of `~` and `+` should be reversed.)  When this is done, `destruct f` will work below.  (But `f` is not the right letter.  A name like `ab` or `x` would be better.

mikeshulman:
Huh.  I agree, that's a very unintuitive precedence order for `~` and `+`.  But it looks like we inherited it from the Coq stdlib.  Anyone know why they did it that way?

contrib/HoTTBookExercises.v
```diff
@@ -235,12 +235,13 @@ Section Book_1_12.
...
 
-  Theorem Book_1_12_part3 : forall A B, (~A + ~B) -> ~(A * B).
+  Theorem Book_1_12_part3 : forall A B, ((~A) + (~B)) -> ~(A * B).
   Proof.
-    intros A B f g.
-    destruct g. apply f.
-    (* TODO: split doesn't work on `+`? *)
-  Abort.
+    intros A B ab prod.
+    destruct prod. destruct ab.
```
jdchristensen, line 241: Coq chooses a strange name `n` when destructing `ab`.  Moreover, it's bad to rely on the chosen name later, since it could change.  `destruct ab as [f|f]` would probably be better, with `n` changed to `f` below.

jdchristensen:
Thanks!  LGTM.

artagnon:
Would you like me to squash all the commits into one for merge?

contrib/HoTTBookExercises.v
```diff
@@ -222,7 +222,27 @@ End Book_1_11.
...
+    intros A a f.
+    exact (f a).
+  Defined.
 
+  Theorem Book_1_12_part3 : forall A B, ((~A) + (~B)) -> ~(A * B).
+  Proof.
+    intros A B ab prod.
+    destruct prod. destruct ab as [f|f].
+    - exact (f fst).
+    - exact (f snd).
```
peterlefanulumsdaine, line 243: It’s not good to make use of auto-generated intro names (here `fst` and `snd`).  Also, `ab` seems better for the variable of type `A*B` than of type `~A + ~B`.  So perhaps the intros/destruct lines could be changed to e.g.:
```
intros A B nab ab.
destruct ab as [a b]. destruct nab as [na | nb].
```
or alternatively, doing the destruct as part of the introduction:
```
intros A B [na | nb] [a b].
```

artagnon: Thanks for the lovely explanation; fixed now.

PR #1334, opened 2020-06-08
[Exercise 1.13] Solved quite easily

artagnon:
We've used a single 'auto' in a place where no more insight is necessary.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -227,7 +227,14 @@ End Book_1_11.
 (* ================================================== ex:not-not-lem *)
 (** Exercise 1.13 *)
 
-
+Section Book_1_13.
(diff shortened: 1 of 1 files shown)


mikeshulman:
No `auto` allowed!

artagnon:
No idea how to solve this without `auto`, since `split` doesn't work as expected.

mikeshulman:
Well that's some good evidence that additional insight *was* required after all...

artagnon:
Well, we could assume LEM or IsHProp, but those are concepts introduced in later chapters. To be honest, I don't know why `auto` works.

mikeshulman:
To be honest, I don't know why `auto` works here either.  We could try to figure it out, but a more productive use of time would be to actually solve the exercise.

We appreciate receiving correct solutions to the exercises to include in the library.  But this isn't really the right forum to ask for help in solving the exercises in the first place.

PR #1340, opened 2020-06-22
Squares in wildcats

Alizter:
Here is a small library of squares in a wildcat that was written by @fpvandoorn. Having a library like this quickly becomes essential, especially when you start composing commutative squares.

I've marked this PR as a draft since there are some things I would like to discuss first.

Firstly, I am not completely convinced with the naming conventions here. In Cubical/PathSquare I stuck to the convention `Square l r t b`. But here squares are instead `Square t b l r`. Since the former is a special case of the latter, we would eventually want some interaction between the structures. I believe having different conventions will be very confusing.

Which convention should we stick with?

Secondly, there are various definitions in this file that need better names.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -45,6 +45,7 @@ theories/Basics/Decidable.v
 #
 #   Wildcat
 #
+
 theories/WildCat.v
 theories/WildCat/Core.v
 theories/WildCat/UnitCat.v
@@ -64,6 +65,7 @@ theories/WildCat/Yoneda.v
 theories/WildCat/TwoOneCat.v
 theories/WildCat/NatTrans.v
 theories/WildCat/PointedCat.v
+theories/WildCat/Square.v
 
 #
 #   Types
--- a/theories/Basics/Notations.v
+++ b/theories/Basics/Notations.v
@@ -27,6 +27,14 @@ Reserved Notation "f ^*" (at level 3, format "f '^*'").
 Reserved Notation "f ^-1*" (at level 3, format "f '^-1*'").
 Reserved Notation "p ^" (at level 3, format "p '^'").
 Reserved Notation "g o*E f" (at level 40, left associativity).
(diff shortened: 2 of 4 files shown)


fpvandoorn:
* Thanks for making a PR with this.
* I'm also fine with the order of arguments you propose. They are "equivalent", but do greatly affect naming. It is very important that the convention is consistent, so they should definitely be the same here as in `PathSquare`.
* About the naming: I'm happy if any of these things are renamed. I tried to base my naming on the naming scheme here, but there are probably some functions whose name can be improved
* About the notation: I feel strongly that we should have notation for these operations. I haven't used Squares all that much on the `wildcat` branch yet, but there are some cases where I think it looks nicer with notation:
```coq
exact ((smap_square g n $@h smap_square f n) $@vR loops_functor_compose _ _).
```
In Lean I used these notations all the time.
* Could you add me as co-author? https://help.github.com/en/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors

Alizter:
@fpvandoorn This PR should allow you to commit to my branch by default. Let me know if it doesn't let you push however.

jdchristensen:
LGTM, once people are happy with the names / argument order.

mikeshulman:
I'm in favor of adding the notations.  As for naming, the main problem I see is that these nice short names may be globally ambiguous.  One possibility would be to prefix them all; another would be to put them in a module that other users could choose to either import or use FQ names.

Alizter:
What about giving them nice long names and sticking all the short names (as notations) in a module?

Alizter:
I've changed the square order to left right top bottom and added the notations back in. I've also gone ahead and added some comments as to what is going on. I've also marked this PR as ready for review.

Alizter:
I count 1 LGTM, so I will go ahead and merge.

PR #1341, opened 2020-06-22
successor structures

Alizter:
Here is a definition of "successor structure", i.e. type with an endofunction. This is a useful notion to consider when thinking about exact sequences, spectra and spectral sequences. This was originally written by @fpvandoorn.


peterlefanulumsdaine:
Looks good to me too.

jdchristensen:
LGTM.

PR #1342, opened 2020-06-22
Exact sequences

Alizter:
Here is a formalization of exact sequences written by @mikeshulman.

* We start off by defining when two (composable) pointed maps are a (very short) complex.
* We then construct the canonical map from the domain of the first map to the fiber of the second.
* Then various properties are shown to be preserved when various functors are applied to complexes.
* A definition of "n-exact" for any modality n (not just truncation index). A complex is considered n-exact when the canonical map from the domain of the first map to the fiber of the second is n-connected.
* Fiber sequences correspond to "oo-exact" sequences where oo is just the identity modality.

Importantly by the end of the file, LES of loops and homotopy groups are constructed. These are however exact as pointed sets rather than groups.

I'm making this a draft for now since it relies on #1341 and #1340.

### DIFF
--- a/theories/Homotopy/ExactSequence.v
+++ b/theories/Homotopy/ExactSequence.v
@@ -1,0 +1,424 @@
+(* -*- mode: coq; mode: visual-line -*-  *)
+Require Import Basics Types.
+Require Import SuccessorStructure.
+Require Import WildCat.
+Require Import Pointed.
+Require Import ReflectiveSubuniverse Modality Modalities.Identity Modalities.Descent.
+Require Import Truncations.
+Require Import HFiber.
+
+Local Open Scope succ_scope.
+Open Scope pointed_scope.
+
+(** * Exact sequences *)
+
+(** ** Very short complexes *)
+
+(** A (very short) complex is a pair of pointed maps whose composite is the zero map. *)
+Definition IsComplex {F X Y} (i : F ->* X) (f : X ->* Y)
+  := (f o* i ==* pconst).
+
+(** This induces a map from the domain of [i] to the fiber of [f]. *)
+Definition cxfib {F X Y : pType} {i : F ->* X} {f : X ->* Y}
+           (cx : IsComplex i f)
+  : F ->* pfiber f.
+Proof.
+  srapply Build_pMap.
+  - intros x; exact (i x ; cx x).
+  - cbn. refine (path_sigma' _ (point_eq i) _); cbn.
+    refine (transport_paths_Fl _ _ @ _).
+    apply moveR_Vp.
+    exact ((concat_p1 _)^ @ point_htpy cx).
+Defined.
+
+(** ...whose composite with the projection [pfib : pfiber i -> X] is [i].  *)
+Definition pfib_cxfib {F X Y : pType} {i : F ->* X} {f : X ->* Y}
+           (cx : IsComplex i f)
+  : pfib f o* cxfib cx ==* i.
+Proof.
+  srapply Build_pHomotopy.
+  - intros u; reflexivity.
+  - cbn.
+    rewrite ap_pr1_path_sigma; hott_simpl.
+Defined.
+
+(** Truncation preserves complexes. *)
+Definition iscomplex_ptr (n : trunc_index)
+       {F X Y : pType} (i : F ->* X) (f : X ->* Y) (cx : IsComplex i f)
+  : IsComplex (ptr_functor n i) (ptr_functor n f).
+Proof.
+  refine ((ptr_functor_pmap_compose n i f)^* @* _).
+  refine (_ @* ptr_functor_pconst n).
+  apply ptr_functor_homotopy; assumption.
+Defined.
+
+(** Loopspaces preserve complexes. *)
+Definition iscomplex_loops
+       {F X Y : pType} (i : F ->* X) (f : X ->* Y) (cx : IsComplex i f)
+  : IsComplex (loops_functor i) (loops_functor f).
+Proof.
+  refine ((loops_functor_compose f i)^* @* _ @* loops_functor_pconst).
+  apply loops_2functor; assumption.
+Defined.
+
+Definition iscomplex_iterated_loops
+           {F X Y : pType} (i : F ->* X) (f : X ->* Y)
+           (cx : IsComplex i f) (n : nat)
+  : IsComplex (iterated_loops_functor n i) (iterated_loops_functor n f).
+Proof.
+  induction n as [|n IHn]; [ exact cx | ].
+  apply iscomplex_loops; assumption.
+Defined.  
+
+(** Passage across homotopies preserves complexes. *)
+Definition iscomplex_homotopic_i {F X Y : pType}
+           {i i' : F ->* X} (ii : i' ==* i) (f : X ->* Y) (cx : IsComplex i f)
+  : IsComplex i' f.
+Proof.
+  exact (pmap_postwhisker f ii @* cx).
+Defined.
+
+Definition iscomplex_homotopic_f {F X Y : pType}
+           (i : F ->* X) {f f' : X ->* Y} (ff : f' ==* f) (cx : IsComplex i f)
+  : IsComplex i f'.
+Proof.
+  exact (pmap_prewhisker i ff @* cx).
+Defined.
+
+Definition iscomplex_homotopic_cancelR {F X Y Y' : pType}
+           (i : F ->* X) (f : X ->* Y) (e : Y <~>* Y') (cx : IsComplex i (e o* f))
+  : IsComplex i f :=
+  (compose_V_hh e (f o* i))^$ $@ 
+    cat_postwhisker _ ((cat_assoc _ _ _)^$ $@ cx) $@ precompose_pconst _.
+
+(** And likewise passage across squares with equivalences *)
+Definition iscomplex_squaric_i {F F' X X' Y : pType}
+           (i : F ->* X) (i' : F' ->* X')
+           (g : F' <~>* F) (h : X' <~>* X) (p : Square i' i g h)
+           (f : X ->* Y)
+           (cx: IsComplex i f)
+  : IsComplex i' (f o* h).
+Proof.
+  refine (pmap_compose_assoc _ _ _ @* _).
+  refine (pmap_postwhisker f p @* _).
+  refine ((pmap_compose_assoc _ _ _)^* @* _).
+  refine (pmap_prewhisker g cx @* _).
+  apply postcompose_pconst.
+Defined.
+
+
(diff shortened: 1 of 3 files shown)


theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+Defined.
+
+Definition iscomplex_homotopic_f {F X Y : pType}
+           (i : F ->* X) {f f' : X ->* Y} (ff : f' ==* f) (cx : IsComplex i f)
+  : IsComplex i f'.
+Proof.
+  exact (pmap_prewhisker i ff @* cx).
+Defined.
+
+Definition iscomplex_homotopic_cancelR {F X Y Y' : pType}
```
jdchristensen, line 88: Maybe drop `homotopic_` from the name?

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+           (i : F ->* X) (f : X ->* Y) (e : Y <~>* Y') (cx : IsComplex i (e o* f))
+  : IsComplex i f :=
+  (compose_V_hh e (f o* i))^$ $@ 
+    cat_postwhisker _ ((cat_assoc _ _ _)^$ $@ cx) $@ precompose_pconst _.
+
+(** And likewise passage across squares with equivalences *)
+Definition iscomplex_squaric_i {F F' X X' Y : pType}
```
jdchristensen, line 95: Why `squaric`?

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+(** And also passage across squares with equivalences. *)
+Definition isexact_squaric_i n  {F F' X X' Y : pType}
+           (i : F ->* X) (i' : F' ->* X')
+           (g : F' <~>* F) (h : X' <~>* X) (p : Square i' i g h)
+           (f : X ->* Y)
+           `{IsExact n F X Y i f}
+  : IsExact n i' (f o* h).
+Proof.
+  exists (iscomplex_squaric_i i i' g h p f cx_isexact); cbn.
+  simple notypeclasses refine (cancelR_equiv_conn_map n (C := pfiber f) _ _).
```
jdchristensen, line 158:
```suggestion
  snrefine (cancelR_equiv_conn_map n (C := pfiber f) _ _).
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+Definition iscomplex_pfib {X Y} (f : X ->* Y)
+  : IsComplex (pfib f) f.
+Proof.
+  srapply Build_pHomotopy.
+  - intros [x p]; exact p.
+  - cbn. exact (concat_p1 _ @ concat_1p _)^.
+Defined.
+
+Global Instance isexact_pfib {X Y} (f : X ->* Y)
+  : IsExact oo (pfib f) f.
```
jdchristensen, line 207: Where is `oo` defined?

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+  := Build_LongExactSequence
+       (Tr k) N (fun n => pTr k.+1 (S n))
+       (fun n => ptr_functor k.+1 (les_fn S n)) _.
+
+
+(** ** LES of loop spaces and homotopy groups *)
+
+Local Notation "'0'" := (inl (inl (inr tt))).
+Local Notation "'1'" := (inl (inr tt)).
+Local Notation "'2'" := (inr tt).
```
jdchristensen, line 396: Are these only used one place?  If so, I think it would be better to inline them.  If they are kept, I'd prefer to not conflict with the naturals.  E.g. `n0`, `n1`, `n2`?

jdchristensen: Or `zero`, `omega`, `omega2`, if we think of these as ordinals (and if I'm interpreting things correctly).

theories/Homotopy/ExactSequence.v
Alizter: Line 163 of Truncations.Core. It is a notation for `purely`.

theories/Homotopy/ExactSequence.v
Alizter: Since these are local notations, they are not visible outside of the file.

jdchristensen: I know.  But I still don't like them...

theories/Homotopy/ExactSequence.v
jdchristensen: Ah, ok.  Is there a reason it's defined there rather than in Identity.v?  As it is, I then have to search for `purely` to ultimately find out what it means.  And why have synonyms like this?  `purely` is only 6 characters and `oo` can mean many things.

(Sorry for the side conversation.)

mikeshulman: I think `oo` made more sense when we wrote just `n` in more places for the truncation modality rather than `Tr n`.  I'd be fine with getting rid of it (in another PR, presumably).

theories/Homotopy/ExactSequence.v
mikeshulman: Yeah.  Maybe that's left over from a previous point at which this lemma included a homotopy.  Will fix.

mikeshulman:
```suggestion
Definition iscomplex_cancelR {F X Y Y' : pType}
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,424 @@
...
+           {f : X ->* Y} {f' : X' ->* Y'}
+           (g : F' <~>* F) (h : X' <~>* X) (k : Y' <~>* Y) 
+           (p : Square i' i g h)
+           (q : Square f' f h k)
+           `{IsExact n F X Y i f}
+  : IsExact n i' f'.
+Proof.
+  pose (I := isexact_squaric_i n i i' g h p f).
+  pose (I2 := isexact_homotopic_f n i' q).
+  exists (iscomplex_homotopic_cancelR i' f' k cx_isexact).
```
mikeshulman, line 184:
```suggestion
  exists (iscomplex_cancelR i' f' k cx_isexact).
```

theories/Homotopy/ExactSequence.v
mikeshulman: By analogy to `homotopic`, this lemma passes the notion of being a complex across a square rather than a homotopy.  Certainly an unlovely "word".  Can anyone think of a better one?

theories/Homotopy/ExactSequence.v
mikeshulman: They're not zero, omega, omega2; the ordinal in question is 3.omega (which is isomorphic to omega) rather than omega.3.  But sure, it's probably fine to just inline them.

theories/Homotopy/ExactSequence.v
jdchristensen: I think one would normally say that `i` is equivalent to `i'` (in the category of arrows), so `iscomplex_equiv_i` (or `equivto` or `byequiv`) might indicate the idea more clearly.  To me the equivalences are the assumption to highlight.

jdchristensen:
LGTM, whether or not `squaric` is changed.

theories/Homotopy/ExactSequence.v
mikeshulman:
```suggestion
Definition iscomplex_equiv_i {F F' X X' Y : pType}
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,420 @@
...
+
+(** And also passage across squares with equivalences. *)
+Definition isexact_squaric_i n  {F F' X X' Y : pType}
+           (i : F ->* X) (i' : F' ->* X')
+           (g : F' <~>* F) (h : X' <~>* X) (p : Square i' i g h)
+           (f : X ->* Y)
+           `{IsExact n F X Y i f}
+  : IsExact n i' (f o* h).
+Proof.
+  exists (iscomplex_squaric_i i i' g h p f cx_isexact); cbn.
```
mikeshulman, line 157:
```suggestion
  exists (iscomplex_equiv_i i i' g h p f cx_isexact); cbn.
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,420 @@
...
+  nrefine (cancelR_isequiv_conn_map n _ e).
+  1: apply equiv_isequiv.
+  refine (conn_map_homotopic n (cxfib (cx_isexact)) _ _ _).
+  intro u. simpl. srapply path_hfiber.
+  1: reflexivity.
+  refine (concat_1p _ @ concat_V_pp _ _)^.
+Defined.
+
+(** And also passage across squares with equivalences. *)
+Definition isexact_squaric_i n  {F F' X X' Y : pType}
```
mikeshulman, line 150:
```suggestion
Definition isexact_equiv_i n  {F F' X X' Y : pType}
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,420 @@
...
+Definition isexact_square_if n  {F F' X X' Y Y' : pType}
+           {i : F ->* X} {i' : F' ->* X'}
+           {f : X ->* Y} {f' : X' ->* Y'}
+           (g : F' <~>* F) (h : X' <~>* X) (k : Y' <~>* Y) 
+           (p : Square i' i g h)
+           (q : Square f' f h k)
+           `{IsExact n F X Y i f}
+  : IsExact n i' f'.
+Proof.
+  pose (I := isexact_squaric_i n i i' g h p f).
```
mikeshulman, line 182:
```suggestion
  pose (I := isexact_equiv_i n i i' g h p f).
```

theories/Homotopy/ExactSequence.v
```diff
@@ -0,0 +1,420 @@
...
+Definition connecting_map {F X Y} (i : F ->* X) (f : X ->* Y)
+           `{IsExact oo F X Y i f}
+  : loops Y ->* F
+  := i_fiberseq (connect_fiberseq i f).
+
+Global Instance isexact_connect_R {F X Y} (i : F ->* X) (f : X ->* Y)
+       `{IsExact oo F X Y i f}
+  : IsExact oo (loops_functor f) (connecting_map i f).
+Proof.
+  refine (isexact_squaric_i (Y := F) oo
```
mikeshulman, line 349:
```suggestion
  refine (isexact_equiv_i (Y := F) oo
```

Alizter:
I will merge this once CI is happy.

PR #1343, opened 2020-06-23
Adapt w.r.t. coq/coq#12572.

ppedrot:
The Coq PR fixes #12571 which was observed in HoTT, according to some puzzled comments in the code. This PR should be backwards compatible.

### DIFF
--- a/theories/Homotopy/EMSpace.v
+++ b/theories/Homotopy/EMSpace.v
@@ -125,11 +125,9 @@ Section LicataFinsterLemma.
     set (Q := fun a b => tr (n:=1) (merid (a * b))
       = tr (merid b @ (merid mon_unit)^ @ merid a)).
     srapply (@wedge_incl_elim_uncurried _ (-1) (-1) _
-      mon_unit _ _ mon_unit _ Q _ _ x y).
-    { intros a b.
-      cbn; unfold Q.
-      apply istrunc_paths.
-      exact _. }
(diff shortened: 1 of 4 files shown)


jdchristensen:
Thanks for this!  LGTM.  Can you explain what changed in Coq?  (I tried following the link to the unreleased changelog, but didn't find anything there.)

Alizter:
From what I read, it seems that local variables (let variables) were not being unfolded by typeclass search.

jdchristensen:
Actually, one comment:  since it looks like these changes are meant to handle different behaviours by different versions of Coq, there should be comments in the files saying this.  Otherwise, someone might try to simplify these by either entirely removing the relevant part, or removing the `try` or `all:`, depending what works on their version of Coq.  The comments could also say that in the future (when version X no longer needs to be supported) one can simplify the code by doing one of those two options.

Alizter:
@jdchristensen Unless I am mistaken, a change that breaks compatibility with future versions would be picked up by the CI since we also test against coq master. When we release for the version of coq this patch appears in we can finally get rid of the try statements like you say.

ppedrot:
Do you still want me to add comments?

Alizter:
That would be great thanks!

ppedrot:
Done.

jdchristensen:
Perfect, thanks!

PR #1344, opened 2020-06-24
Cleanup contrib/old (closes #1338)

### LINKED ISSUE #1338: What is the role of `contrib/old`?  Keeping old code present but not in makefile seems bad.
I have an external development based over the HoTT library, which uses the lemma `einv_V` from `EquivGroupoids.v`, which got moved to `contrib/old` in #1287/#1288.  The motivation given there is that `EquivGroupoids.v` wasn’t used by anything else in the library, and should be deprecated for internal use, because it depends on `Funext` more than is desirable.

This is rather disruptive to external developments (more so than most moves/changes within the library), since files in `contrib/old` are not built by `make`, and so there’s now no clean way to import them.

**Generally, it feels like having `contrib/old` around but not in the makefile is a mistake.**  Files there aren’t importable, and will go stale, so are essentially no better off than deleted files (which are always still accessible in repo history anyway) — but moving files to `contrib/old` *feels* much less drastic than deleting them, and so gets done with less compunction.

**Am I missing something here — are there any advantages to keeping files around in `contrib/old` rather than deleting them?**  (On a quick search, I can’t find any previous discussion of this point.)

Assuming people agree that keeping non-maintained code around is bad, I propose:

- for the files in `contrib/old` that haven’t staled too much and seem potentially still of interest: bring them up to date and add them to `_CoqProject`

- for files that have staled too much to be easily resurrected, or that seem very clearly bad/obsolete: delete them entirely (and anyone who wants can still resurrect them from repo history later)

peterlefanulumsdaine:
Fixes the biggest headache of #1338 : all `.v` files in `contrib/old` are either deleted or else are back in the `_CoqProject`/`Makefile`.


peterlefanulumsdaine:
Merging this, since it’s had no objections, one approval, and 48 hours.

PR #1347, opened 2020-06-24
make strict now checks contrib

Alizter:


Alizter:
I'll publish this once contrib has been cleaned up. For now it will fail since there are files in contrib that are not found in _CoqProject.

peterlefanulumsdaine:
Ah, so we already had a check for this, just had forgotten to include `contrib`?

Might it be more robust to have `ALL_HOTT_VFILE` really include *all* `*.v` files in the repo, except as deliberately excluded ones?  E.g. preparing #1344 , I was checking coverage using a local 
```
find . -type f -name "*.v" \
	-not -path "./coq-HoTT/*" \
	-not -path "./coq/*" \
	-not -path "./etc/*"
```

Alizter:
It might make sense to include files in `/coq/` but `/coq-HoTT/` is a submodule, so I don't really want to go poking around there. Eventually coq is allowing us to completely ignore the standard library anyway, so over time `/coq/` will fade away. It also seems that the `.v` files in `/etc/` are all in the `coq-scripts` submodule, hence I don't think we should care about those.

peterlefanulumsdaine:
Sorry if I was unclear.  I completely agree, the submodules and `etc/` should be excluded from this check, as in my example search.   My suggestion is: the check may be more robust if `ALL_HOTT_VFILES` is defined as (in pseudo-code) “all `*.v`-files in the repository, excluding the submodules and `etc/`” instead of the current definition as “all `*.v`-files in `contrib` and `theories`”.

Alizter:
@peterlefanulumsdaine Ah I see. I've changed the script to check all files not in /etc/ nor /coq-HoTT/

peterlefanulumsdaine:
Great — looks good to me once it passes the CI, which I guess should happen once #1344 is merged and this is rebased over or merged with that?

Alizter:
@peterlefanulumsdaine Yes, we should wait till contrib has been cleaned up.

peterlefanulumsdaine:
Now that #1344 is merged, do we need to rebase this or merge in the newer master to get Travis to re-build it?  Anyhow, as above: LGTM once Travis builds this.

Alizter:
I've rebased. By the way, our CI is not called Travis anymore, Travis has retired and we are using Github actions.

I'll merge when the CI is happy.

Issue #1345, opened 2020-06-24
Re-home or delete the material of `EquivGroupoids`?


peterlefanulumsdaine:
The file `theories/EquivGroupoids` was pseudo-deleted (i.e. removed from the `Makefile`) in #1287 / #1288, and then restored in #1344  (if that gets merged) following discussion in #1338 .  Roughly summarising the discussion from those threads:

- I (@peterlefanulumsdaine) find the material in this file useful, and would prefer that it be retained in the library, not deleted (but I don’t mind at all if it gets moved, reorganised, split up, etc).

- @mikeshulman and @Alizter don’t find this useful, and are particularly unhappy with its current location

So it sounds like probably it should be reorganised or moved to a better location?  Or possibly deleted, if people have more substantial motivation for deletion than was mentioned in the cursory discussion at #1287 before pseudo-deletion.

mikeshulman:
It sounded to me like we had an emerging consensus on moving its contents to `Types/Equiv`.  The fact that you've found it useful is reason enough for me to keep it in the library.  It might be worth thinking a bit more about naming conventions for these lemmas, but I don't have a lot of brain cycles to spare for that right now, so I'd be fine with just plunking them in `Types/Equiv` as-is now and making a note somewhere that it might be worth thinking about renaming them in the future.

peterlefanulumsdaine:
That sounds good to me!

PR #1350, opened 2020-06-28
Reinstate EquivGroupoids in export list HoTT.v

peterlefanulumsdaine:
Cf. #1345. This reverts fdedfe6, but was missed when restoring `EquivGroupoids` in #1344 .

### DIFF
--- a/theories/HoTT.v
+++ b/theories/HoTT.v
@@ -13,6 +13,7 @@ Require Export HoTT.Truncations.
 Require Export HoTT.HFiber.
 Require Export HoTT.HProp.
 Require Export HoTT.HSet.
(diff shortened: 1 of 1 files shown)


mikeshulman:
I suppose this is ok for now, but it'll have to be re-reverted again when we move the content of EquivGroupoids into Types/Equiv.

peterlefanulumsdaine:
@mikeshulman: yes, of course. But it seems good to me to keep `HoTT.v` up-to-date with the current state of the library either way? And changes needed due to moving or removing a file  will be caught by the build — it’s only adding or re-adding a file that adds an extra burden to keep track of.

Issue #1335, opened 2020-06-10
Make ~ bind tighter than +


mikeshulman:
The discussion at #1333 suggests that the current precedence order, which we inherited from Coq, is more or less accidental, and it might be better if `~` bound tighter than binary operators like `+` and `*`.  Let's try it!

Alizter:
This is the file where the notations are set:
https://github.com/HoTT/HoTT/blob/master/coq/theories/Init/Notations.v

Here are some relevant lines:
```coq
Reserved Notation "~ x" (at level 75, right associativity).
Reserved Notation "x + y" (at level 50, left associativity).
```

What do we suggest the new levels be? I suppose we should change the level of ~ to something like 1. Any suggestions?

peterlefanulumsdaine:
Level 35 sounds good to me.  I’ve tried this out in [`negation-precedence`](https://github.com/peterlefanulumsdaine/HoTT/tree/negation-precedence), and it works very well — it makes things strictly more human-readable, at least to my eye. In particular, there were a few places where the old precedence was actively misleading: e.g. in `Bool.v`, expressions like `~false = true`, which (at least to me) looks like it means `(~false) = true` (with `~` as boolean negation), but actually meant `~(false = true)`.

Regarding alternative precedences: the other big difference would be to make `~` bind tighter than function application, i.e. level <10.  That seems good when `~` is under a function, e.g. `PropHolds ~(x < y)` is nice, but bad when `~ ` is outside a function, e.g. `~ equiv p q` is a little hard to parse.

Within the precedence range between 10 (function application) and 40 (where binary operators/relations start), I see very little difference — not much to choose between e.g. 35 and 25.

jdchristensen:
It looks to me like all of the changes in #1351 apply even if the precedences aren't changed, so to me writing things in that form would be good style, whether or not we change the precedence.  Changing the precedence would force people to use that style, which I think is more clear, but should also be balanced against having less compatibility with standard Coq.  A crazy idea would be to check that the library builds with *either* precedence, so nobody can be confused by the meaning of what we write.

mikeshulman:
Yes, precedence 35 looks good.  I agree that everything in #1351 is an improvement.

For comparison, what are some things we could write with the new precedence that would not parse correctly with the old precedence?  Are they things we'd want to permit or prefer to forbid?

peterlefanulumsdaine:
[...]

PR #1351, opened 2020-06-28
Changed ‘~’ and ‘~~’ to bind tightly (level 35)

peterlefanulumsdaine:
A proposal addressing #1335.  Marking as draft so it doesn’t get merged until we’ve discussed whether further tweaks are desirable.


peterlefanulumsdaine:
I’ve fixed an earlier omission: I’ve now lowered the precedence of `¬` (synonym of `~`)and `!!` (negation on decidable props) to level 35, to match with `~`.

There are now no unary constructors of level > 35, at least not among the reserved notations in `Basics/Notations.v`.  There are some binary constructors of level ≤35, but as far as I can see none of them are type-valued relations/operations, so they should not interact misleadingly with `~`.

So I think this is ready to merge.

PR #1353, opened 2020-07-03
Reorganization of group theory library + added some group theory

Alizter:
In this PR we restructure the `Algebra` directory to include two new subdirectories: `AbGroups` and `Groups`. We have also added various bits of group theory.

Here is a quick rundown of the changes:

* `theories/Algebra/Groups.v` is an indexing file for group theory files.
* `theories/Algebra/Groups/Group.v` contains: the definition of a group. (Note that this has been changed from a Class to a Record). The definition of a group homomorphism and isomorphism (which have also been changed from a Class to a Record). 
R067    theories/Algebra/Subgroup.v     theories/Algebra/Groups/Subgroup.v
* `theories/Algebra/Groups/Image.v` contains the definition of the image of a group homomorphism. The image is defined as a Subgroup which comes with its inclusion map. We also define `grp_image_in` which is the restriction of the codomain of the original map to the image.
* `theories/Algebra/Groups/Kernel.v` the kernel of a subgroup. Together with a proof that it is a normal subgroup. 
* `theories/Algebra/Groups/QuotientGroup.v`(used to be `theories/Algebra/QuotientGroup.v`). This file underwent some cleanup and has a proof of the first isomorphism theorem for groups at the end.

* `theories/Algebra/AbGroups.v` is an indexing file for the Abelian groups library.
* `theories/Algebra/AbGroups/AbelianGroup.v` (what used to be `theories/Algebra/AbelianGroup.v`) is the main file for abelian groups. It contains constructions for groups specialized to abelian groups such as quotients, images and the first isomorphism theorem. `AbGroup` is shown to be a pointed wild category.
* `theories/Algebra/AbGroups/Abelianization.v` contains the definition of being an Abelianization and a construction of one.
* `theories/Algebra/AbGroups/Z.v` the group Z is defined as an abelian group (this can of course always be coerced to a group).

Finally the rest of the files are just compatibility changes. 
* `_CoqProject`
* `theories/HoTT.v`
* `theories/Homotopy/ClassifyingSpace.v`
* `theories/Homotopy/EMSpace.v`
* `theories/Homotopy/HomotopyGroup.v`
* `theories/Homotopy/Pi1S1.v`
* `theories/Homotopy/WhiteheadsPrinciple.v`
* `theories/Spaces/Torus/TorusHomotopy.v`


mikeshulman:
LGTM.

I forgot why we define a "subgroup of G" to be a group with an injective map to G rather than a predicate on G that's closed under the operations?

Alizter:
Both setups have their advantages and are ultimately equivalent in some sense. "subgroups as injections" allow us to say things like "Z2 is a subgroup of S3". At the time I thought this would be more useful than having to write "Z2 is isomorphic to a subgroup of S3".

It may be time to change this however.

Alizter:
@jdchristensen Do you still want to review this or are you ok letting us merge?

theories/Algebra/Groups/Group.v
```diff
@@ -214,7 +246,7 @@ Global Instance transitive_groupisomorphism
 Proof.
   intros G H K f g.
   srapply Build_GroupIsomorphism.
-  1: srapply grp_homo_compose.
+  1: srapply (grp_homo_compose g f).
```
jdchristensen, line 249: Would `exact` work?  If so, it's more clear to the reader.

theories/Algebra/AbGroups.v
```diff
@@ -0,0 +1,8 @@
+(** Theory *)
+
+Require Export HoTT.Algebra.AbGroups.AbelianGroup.
+Require Export HoTT.Algebra.AbGroups.Abelianization.
+
+(** Examples *)
+
+Require Export HoTT.Algebra.AbGroups.Z.
```
jdchristensen, line 8: Add final newline.

theories/Algebra/Groups/Group.v
```diff
@@ -309,9 +341,16 @@ Proof.
   all: intro; apply negate_involutive.
 Defined.
 
+(** Trivial group *)
+Definition grp_trivial : Group.
```
jdchristensen, line 345: Choice of names:  `TrivialAbGroup` vs `grp_trivial`?

theories/Algebra/Groups/Group.v
Alizter: Yes. Since previously group homomorphisms were a typeclass they were getting picked up by `srapply`. Since I demoted them back to records, it doesn't really make sense to use srapply anymore.

Alizter:
```suggestion
  1: exact (grp_homo_compose g f).
```

theories/Algebra/AbGroups.v
Alizter:
```suggestion
Require Export HoTT.Algebra.AbGroups.Z.

```

theories/Algebra/Groups/Group.v
Alizter: I think `grp_trivial` is the better name. I will change `TrivialAbGroup` to `abgroup_trivial`. I could also define `TrivialAbGroup` in terms of `grp_trivial` as well.

Alizter: I've changed the name to `abgroup_trivial` and it is now defined in terms of `grp_trivial`.

PR #1357, opened 2020-07-08
removed unused folders in coq

Alizter:
These folders were kept since historically coq complained without them. This doesn't seem to be the case anymore.


mikeshulman:
LGTM

PR #1358, opened 2020-07-09
fixed typo in reflexivity tactic causing universe increase

### LINKED ISSUE #1356: Using `reflexivity` with `Equiv` increases universe
Using the `reflexivity` tactic with an `Equiv` goal causes a strictly larger universe to be postulated. This has unfortunately propagated universe growth in various areas. The following snippet demonstrates the problem:

```coq
Require Import Basics.

Axiom A@{i} : Type@{i}.

Axiom foo@{i} : A@{i} <~> A@{i}.

Definition bar@{i} : A@{i} <~> A@{i}.
Proof.
  reflexivity.
Fail Defined.
Abort.

Definition bar@{i} : A@{i} <~> A@{i}.
Proof.
  exact equiv_idmap.
Defined.
```

Since the instance for reflexivity for `Equiv` has to mention a relation `Type@{i} -> Type@{i} -> Type@{i}` it is no surprise that `reflexivity` increases the universe. However, this does mean various places where `reflexivity` is used to prove equivalences may be better served with specifying `equiv_idmap` exactly.

Is there a way to get `reflexivity` to not introduce a new universe here?

Alizter:
There was a typo in the `reflexivity` tactic which caused universes to increase when proving an `Equiv` goal. Now `reflexivity` applies the reduced term just like `symmetry` and `transitivity` do.

closes #1356

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -24,6 +24,8 @@ Notation "1" := equiv_idmap : equiv_scope.
 
 Global Instance reflexive_equiv : Reflexive Equiv | 0 := @equiv_idmap.
 
+Arguments reflexive_equiv /.
+
 (** The composition of equivalences is an equivalence. *)
 Global Instance isequiv_compose `{IsEquiv A B f} `{IsEquiv B C g}
   : IsEquiv (compose g f) | 1000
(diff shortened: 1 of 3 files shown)


SkySkimmer:
>There was a typo in the reflexivity tactic which caused universes to increase when proving an Equiv goal. 

This change needs to be combined with the Arguments command to actually reduce universes IIRC

Alizter:
@SkySkimmer Yep, its on its way.

jdchristensen:
Just out of curiosity, what does `Arguments /.` do?  I tried to search for this but failed.

jdchristensen:
LGTM.

Alizter:
@jdchristensen  It tells `simpl` and `cbn` to always unfold once that many arguments are applied.

PR #1361, opened 2020-07-11
Define GrpPullback, AbPullback and prove corecursion principle

jarlg:
Implements pullbacks of groups in [Groups/GrpPullback.v](https://github.com/jarlg/HoTT/blob/GrpPullback/theories/Algebra/Groups/GrpPullback.v) by equipping `Pullback f g` (from Limits/Pullback.v) with a group structure whenever `f`, `g` are group homomorphisms. If the groups involved are abelian, so is the pullback -- this is what [AbGroups/AbPullback.v](https://github.com/jarlg/HoTT/blob/GrpPullback/theories/Algebra/AbGroups/AbPullback.v) amounts to.

I'm unsure about which instances should be Global vs Local. Currently only the [Group instance for `Pullback f g`](https://github.com/jarlg/HoTT/blob/3eeee0b774f59343a14cd970ca7865f9d590e25f/theories/Algebra/Groups/GrpPullback.v#L91) is Global and the [two instances in AbPullback.v](https://github.com/jarlg/HoTT/blob/3eeee0b774f59343a14cd970ca7865f9d590e25f/theories/Algebra/AbGroups/AbPullback.v#L11) as well.

Should [`isequiv_grp_pullback_corec`](https://github.com/jarlg/HoTT/blob/3eeee0b774f59343a14cd970ca7865f9d590e25f/theories/Algebra/Groups/GrpPullback.v#L158) and [`isequiv_ab_pullback_corec`](https://github.com/jarlg/HoTT/blob/3eeee0b774f59343a14cd970ca7865f9d590e25f/theories/Algebra/AbGroups/AbPullback.v#L43) be instances?

### DIFF
--- a/theories/Algebra/AbGroups/AbPullback.v
+++ b/theories/Algebra/AbGroups/AbPullback.v
@@ -1,0 +1,49 @@
+Require Import Basics Types Limits.Pullback Cubical.PathSquare.
+Require Import Algebra.Groups Algebra.AbGroups.AbelianGroup.
+Require Import WildCat.
+
+(** Pullbacks of abelian groups. *)
+
+Section AbPullback.
+  (* Variables are named to correspond with Limits.Pullback. *)
+  Context {A B C : AbGroup} (f : B $-> A) (g : C $-> A).
+
+  Global Instance ab_pullback_commutative
+    : Commutative (@group_sgop (GrpPullback f g)).
+  Proof.
+    unfold Commutative.
+    intros [b [c p]] [d [e q]].
+    apply equiv_path_pullback; simpl.
+    refine (commutativity _ _; commutativity _ _; _).
+    apply equiv_sq_path.
+    apply path_ishprop.
+  Defined.
+
+  Global Instance isabgroup_ab_pullback
+    : IsAbGroup (GrpPullback f g) := {}.
+
+  Definition AbPullback
+    : AbGroup := Build_AbGroup (GrpPullback f g) _ _ _ _.
+
+  Proposition ab_pullback_corec {X : AbGroup}
+              (b : X $-> B) (c : X $-> C)
+              (p : f o b == g o c)
+    : X $-> AbPullback.
+  Proof.
+    exact ((@grp_pullback_corec A B C f g X) b c p).
+  Defined.
+
+  Corollary ab_pullback_corec' {X : AbGroup}
+    : {b : X $-> B & {c : X $-> C & f o b == g o c}}
+      -> (X $-> AbPullback).
+  Proof.
+    intros [b [c p]]; exact (ab_pullback_corec b c p).
+  Defined.
+
+  Theorem isequiv_ab_pullback_corec {X : AbGroup} `{Funext}
+    : IsEquiv (@ab_pullback_corec' X).
+  Proof.
+    srapply isequiv_grp_pullback_corec.
+  Defined.
+
+End AbPullback.
--- a/theories/Algebra/AbGroups/AbPullback.v
+++ b/theories/Algebra/AbGroups/AbPullback.v
@@ -25,26 +25,7 @@ Section AbPullback.
   Definition AbPullback
     : AbGroup := Build_AbGroup (GrpPullback f g) _ _ _ _.
 
-  Proposition ab_pullback_corec {X : AbGroup}
-              (b : X $-> B) (c : X $-> C)
-              (p : f o b == g o c)
-    : X $-> AbPullback.
-  Proof.
-    exact ((grp_pullback_corec f g) b c p).
-  Defined.
-
-  Corollary ab_pullback_corec' (X : AbGroup)
-    : {b : X $-> B & {c : X $-> C & f o b == g o c}}
-      -> (X $-> AbPullback).
-  Proof.
-    exact ((grp_pullback_corec' f g) X).
-  Defined.
+  (** The corecursion principle is inherited from Groups; use grp_pullback_corec and friends from Groups/GrpPullback.v. *)
 
 End AbPullback.
-
-Theorem isequiv_ab_pullback_corec `{Funext} {A B C X : AbGroup}
-        (f : B $-> A) (g : C $-> A)
-  : IsEquiv (ab_pullback_corec' f g X).
-Proof.
-  exact ((isequiv_grp_pullback_corec f g) X).
-Defined.
+  
--- a/theories/Algebra/AbGroups/AbPullback.v
+++ b/theories/Algebra/AbGroups/AbPullback.v
@@ -9,7 +9,7 @@ Section AbPullback.
   Context {A B C : AbGroup} (f : B $-> A) (g : C $-> A).
 
   Global Instance ab_pullback_commutative
-    : Commutative (@group_sgop (GrpPullback f g)).
+    : Commutative (@group_sgop (grp_pullback f g)).
   Proof.
     unfold Commutative.
     intros [b [c p]] [d [e q]].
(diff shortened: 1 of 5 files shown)


theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,49 @@
...
+
+  Definition AbPullback
+    : AbGroup := Build_AbGroup (GrpPullback f g) _ _ _ _.
+
+  Proposition ab_pullback_corec {X : AbGroup}
+              (b : X $-> B) (c : X $-> C)
+              (p : f o b == g o c)
+    : X $-> AbPullback.
+  Proof.
+    exact ((@grp_pullback_corec A B C f g X) b c p).
```
jdchristensen, line 33:
```suggestion
    exact ((grp_pullback_corec f g) b c p).
```

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,49 @@
...
+  Defined.
+
+  Corollary ab_pullback_corec' {X : AbGroup}
+    : {b : X $-> B & {c : X $-> C & f o b == g o c}}
+      -> (X $-> AbPullback).
+  Proof.
+    intros [b [c p]]; exact (ab_pullback_corec b c p).
+  Defined.
+
+  Theorem isequiv_ab_pullback_corec {X : AbGroup} `{Funext}
```
jdchristensen, line 43: We like to have `Funext` args at the front, so:
```suggestion
  Theorem isequiv_ab_pullback_corec `{Funext} {X : AbGroup}
```
However, because of the `Context` at the start of the `Section`, we still end up with `Funext` in the middle of the list of arguments.  So it would probably be better to end the section before this result.

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,49 @@
...
+    : X $-> AbPullback.
+  Proof.
+    exact ((@grp_pullback_corec A B C f g X) b c p).
+  Defined.
+
+  Corollary ab_pullback_corec' {X : AbGroup}
+    : {b : X $-> B & {c : X $-> C & f o b == g o c}}
+      -> (X $-> AbPullback).
+  Proof.
+    intros [b [c p]]; exact (ab_pullback_corec b c p).
```
jdchristensen, line 40: Making these all parallel references to the results in GrpPullback makes the pattern clear:
```suggestion
    exact (grp_pullback_corec' f g).
```

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,49 @@
...
+  Defined.
+
+  Theorem isequiv_ab_pullback_corec {X : AbGroup} `{Funext}
+    : IsEquiv (@ab_pullback_corec' X).
+  Proof.
+    srapply isequiv_grp_pullback_corec.
```
jdchristensen, line 46:
```suggestion
    exact (isequiv_grp_pullback_corec f g).
```
(Just for the sake of the pattern.)

theories/Algebra/Groups/GrpPullback.v
```diff
@@ -0,0 +1,175 @@
...
+
+  Lemma grp_pullback_corec_pr2 {X : Group}
+        (b : X $-> B) (c : X $-> C)
+        (p : f o b == g o c)
+    : grp_pullback_pr2 f g $o grp_pullback_corec f g b c p = c.
+  Proof.
+    apply equiv_path_grouphomomorphism; reflexivity.
+  Defined.
+
+  Theorem isequiv_grp_pullback_corec `{Funext} (X : Group)
```
jdchristensen, line 158:
```suggestion
  Theorem isequiv_grp_pullback_corec (X : Group)
```
`Funext` is in the Context.

theories/Algebra/AbGroups/AbPullback.v
jarlg: Thanks! I've fixed this in the next commit.

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,50 @@
...
+    apply (grp_pullback_corec' f g).
+  Defined.
+
+End AbPullback.
+
+Theorem isequiv_ab_pullback_corec `{Funext} {A B C X : AbGroup}
+        (f : B $-> A) (g : C $-> A)
+  : IsEquiv (ab_pullback_corec' f g X).
+Proof.
+  apply (isequiv_grp_pullback_corec f g).
```
jdchristensen, line 49: It doesn't matter much, but doesn't `exact` work in place of the last three `apply`s?  `exact` gives a bit more info to the reader.

jarlg: Yep, I just needed to supply the last parameter 'X'.

jarlg: I realize now I had a confusion earlier, the fact that 'exact' works here tells me these three last results are indeed redundant. In the interest of code brevity I'm going to remove them and refer to the grp variants.

theories/Algebra/Groups/GrpPullback.v
```diff
@@ -0,0 +1,175 @@
...
+    unfold grp_pullback_sgop; simpl.
+    apply equiv_path_pullback; simpl.
+    refine (right_inverse _; right_inverse _; _).
+    apply equiv_sq_path.
+    apply path_ishprop.
+  Defined.
+
+  Global Instance isgroup_grp_pullback : IsGroup (Pullback f g) := {}.
+
+  Definition GrpPullback : Group
```
Alizter, line 93: I would suggest naming this `grp_pullback`.

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -0,0 +1,31 @@
...
+    apply equiv_path_pullback; simpl.
+    refine (commutativity _ _; commutativity _ _; _).
+    apply equiv_sq_path.
+    apply path_ishprop.
+  Defined.
+
+  Global Instance isabgroup_ab_pullback
+    : IsAbGroup (GrpPullback f g) := {}.
+
+  Definition AbPullback
```
Alizter, line 25: This might also be better named as ab_pullback

Alizter:
Local instances are fine when constructing a group like you have. When you have a group `G` all its typeclass data are automatically instances anyway so there is no reason to have two sets of identical instances.

theories/Algebra/Groups/GrpPullback.v
jarlg: I think the current naming is consistent with [STYLE.md](https://github.com/HoTT/HoTT/blob/master/STYLE.md#general-principles), and prefixing with the property being proven is how I see it usually written in Algebra/. Could you explain your reasoning here? To me, `grp_pullback` sounds like the function creating the pullback (like `grp_prod` which creates the product of two groups, in Groups/Group.v).

jarlg: I think the renaming is fine; it's inline with `grp_prod`, `grp_image`, `grp_kernel`. The outlier seems to be quotient groups, for which the constructor is `QuotientGroup G N` (and `QuotientAbGroup G N`).

Alizter: That's a good point, I should rename that.

jarlg: Cool! Is there an explanation why `Pullback` (from Limits/Pullback.v) is capitalized, whereas in Algebra/ the "constructors" aren't? From my reading of [STYLE.md](https://github.com/HoTT/HoTT/blob/master/STYLE.md#capitalization-and-spacing) it seems types should be capitalized. Since all the groups in question (products, images, kernels, pullbacks) are types, shouldn't the constructors be capitalized? I.e. `QuotientGroup` is the one adhering to STYLE.md, it seems to me.

PR #1365, opened 2020-07-13
Free groups

### LINKED ISSUE #1139: Free groups
I would like to construct the free group of a type A. I will probably need A to be a hset. The universal property of the free group F(-) will look something like this:

GroupHom(F(A), G) <~> (A -> G)

Now my plan for doing this is as follows:

 1. Define the free group of a set A as F(A) := Pi 1 (Susp (A + 1)). 
 2. Start with a group homomorphism h : F(A) -> G
 3. Promote this to the world of oo-groups using the classifying space construction: BF(A) -> BG, whereby it becomes a pointed map.
 4. Observe that B ( Pi 1 X) <~> ||X||_1 which can be proven using the truncated Whitehead theorem from the HoTT book. Hence we have a pointed map: Susp (A + 1) -> BG.
 5. Use the loop-susp adjunction to get a pointed map:  A + 1 -> loops BG, and observe that loops BG is pointed equivalent to G (pointed by identity element).
 6. Observe that pointed maps A + 1 -> X are equivalent to maps A -> X. So we are done.

So we can get a recursion principal from this definition of free group for "free". The work that needs to be done is as follows:

1. Prove the truncated Whitehead's theorem (The proof is in the book).
2. Prove an isomorphism : Pi 1 X -> G implies a pointed equivalence X -> B G.
3. Show that classifying spaces can turn groups into oo-groups.

I am making a note of this argument here so I don't forget it in the future, as of now I have a formalization of classifying spaces but nothing about B (Pi 1 (X)).

Alizter:
Here is a beginning on free groups. We add a new folder to Algebra/Groups called `FreeGroup`. This includes two files `KrausAltenkirch.v` and `LoopSusp.v`. These are two separate constructions of free groups which are defined in Group.v. The file Algebra/Groups/FreeGroup.v contains properties of free groups and exports KrausAltenkirch.v as the main implementation.

Here is a quick summary of the main file changes:

* `theories/Algebra/FreeGroup.v` contains properties of free groups (UP). Anybody wanting material on free groups would import this file or just Group.v which imports all group theory.
* `theories/Algebra/FreeGroup/KrausAltenkirch.v` contains a clever HIT for the free group on a type X. It is basically a quotient of the usual list construction, but with minimal constructors. This is a truncated version of a HIT found in Kraus-Altenkirch's "Free Higher Groups in Homotopy Type Theory". It is shown that this HIT satisfies the correct UP. This is also how we define the free group on X.
* `theories/Algebra/FreeGroup/LoopSusp.v` contains a definition of the free group on X as the fundamental group of the suspension of X with a basepoint attached. I wasn't able to finish the proof that this is the free group, but I kept this around anyway in case anybody else want's to finish it.
* In `ClassifyingSpace.v` we prove that the delooping of the fundamental group of a 0-connected 1-truncated type is equivalent to the original type.
* In `ooGroup.v` we show that ooGroup is a wildcat and that there is an inclusion (at least) 1-functor from Group to ooGroup.

Closes #1139

### DIFF
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -1,0 +1,45 @@
+Require Import Basics Types.
+Require Import Groups.Group.
+Require Import WildCat.
+
+(** Free groups are defined in Group.v. In this file we prove properties of free groups. There are various constructions of free groups to choose from, all of which can be found in the FreeGroup folder. *)
+
+(** In this file, and in the rest of the library we choose a modified version of the free group which can be found in Kraus-Altenkirch 2018 arXiv:1805.02069. This is a very simple HIT in a similar manner to the abelianization HIT used in Algebra.AbGroup.Abelianization. *)
+Require Export Algebra.Groups.FreeGroup.KrausAltenkirch.
+
+(** Properties of free groups *)
+
+(** We can state the universal property of free groups as an equivalence: (F(A) $-> G) <~> (A -> G) *)
+Theorem equiv_isfreegroup_rec `{Funext} (G F : Group) (A : Type) (i : A -> F)
+  `{IsFreeGroupOn A F i}
+  : (F $-> G) <~> (A -> G).
+Proof.
+  snrapply Build_Equiv.
+  { intros f.
+    exact (f o i). }
+  nrapply isequiv_contr_map.
+  intro f.
+  unfold hfiber.
+  snrapply contr_equiv'.
+  1: exact (FactorsThroughFreeGroup A F i G f).
+  { rapply equiv_functor_sigma_id.
+    intro g.
+    apply equiv_path_forall. }
+  exact _.
+Defined.
+
+(** The above theorem is true regardless of the implementation of free groups. This let's us state the more specific theorem about the free groups themselves. This can be read as [FreeGroup] is left adjoint to the forgetful functor [group_type]. *)
+Theorem equiv_freegroup_rec `{Funext} (G : Group) (A : Type)
+  : (FreeGroup A $-> G) <~> (A -> G).
+Proof.
+  rapply equiv_isfreegroup_rec.
+Defined.
+
+(** TODO: Nielsen-Schreier theorem: Subgroups of free groups are free. Proofs of this statement are non-trivial. We can prove it using covering spaces which haven't yet been considered in this library. In fact, every known proof requires the axiom of choice in some crucical way. *)
+
+(** Here is a sketch of such a proof. If F is a free group on a type X, then it is the fundamental group of the suspension of (X + 1) (This coule be by definition). A subgroup is then the fundamental group of a covering space of Susp (X + 1). This space is a connected 1-type and using choice we can show it has a spanning tree (since it is a topological graph). By shrinking the spanning tree we get that this cover is also the suspension of some non-empty type, hence is a free group. This "proof" is however a sketch and there may be serious problems when allowing groups to be free over arbitrary types. *)
+
+(** TODO: If F(A) $<~> F(B) then the cardinalities of [A] and [B] are the same. *)
+
+(** A proof might go like this: If F(A) $<~> F(B) then so are their abelianizations F(A)^ab $<~> F(B)^ab. It can be shown that F(A)^ab = Z^A hence we need only compare ranks of free abelian groups. It is not clear to me at the time of writing that this is now easier to prove however. *)
+
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -28,7 +28,7 @@ Proof.
   exact _.
 Defined.
 
-(** The above theorem is true regardless of the implementation of free groups. This let's us state the more specific theorem about the free groups themselves. This can be read as [FreeGroup] is left adjoint to the forgetful functor [group_type]. *)
+(** The above theorem is true regardless of the implementation of free groups. This lets us state the more specific theorem about the canonical free groups. This can be read as [FreeGroup] is left adjoint to the forgetful functor [group_type]. *)
 Theorem equiv_freegroup_rec `{Funext} (G : Group) (A : Type)
   : (FreeGroup A $-> G) <~> (A -> G).
 Proof.
@@ -57,4 +57,3 @@ Proof.
     apply equiv_path_forall. }
   exact _.
 Defined.
-
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -35,11 +35,26 @@ Proof.
   rapply equiv_isfreegroup_rec.
 Defined.
 
-(** TODO: Nielsen-Schreier theorem: Subgroups of free groups are free. Proofs of this statement are non-trivial. We can prove it using covering spaces which haven't yet been considered in this library. In fact, every known proof requires the axiom of choice in some crucical way. *)
+Global Instance ishprop_isfreegroupon `{Funext} (F : Group) (A : Type) (i : A -> F)
+  : IsHProp (IsFreeGroupOn A F i).
+Proof.
+  unfold IsFreeGroupOn.
+  apply trunc_forall.
+Defined.
 
-(** Here is a sketch of such a proof. If F is a free group on a type X, then it is the fundamental group of the suspension of (X + 1) (This coule be by definition). A subgroup is then the fundamental group of a covering space of Susp (X + 1). This space is a connected 1-type and using choice we can show it has a spanning tree (since it is a topological graph). By shrinking the spanning tree we get that this cover is also the suspension of some non-empty type, hence is a free group. This "proof" is however a sketch and there may be serious problems when allowing groups to be free over arbitrary types. *)
-
-(** TODO: If F(A) $<~> F(B) then the cardinalities of [A] and [B] are the same. *)
-
-(** A proof might go like this: If F(A) $<~> F(B) then so are their abelianizations F(A)^ab $<~> F(B)^ab. It can be shown that F(A)^ab = Z^A hence we need only compare ranks of free abelian groups. It is not clear to me at the time of writing that this is now easier to prove however. *)
+(** Both ways of stating the universal property are equivalent. *)
+Definition equiv_isfregroupon_isequiv_precomp `{Funext} (F : Group) (A : Type) (i : A -> F)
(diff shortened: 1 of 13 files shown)


theories/Homotopy/ClassifyingSpace.v
```diff
@@ -349,3 +389,211 @@ Section HSpace_bg.
...
+        1: exact 1.
+        { induction n.
+          1: exact tt.
+          rapply trunc_index_leq_transitive. }
+        exact _. }
+      apply equiv_istrunc_contr_iterated_loops.
+      induction n; exact _. } }
+  (** Finally we need to show that this equivalence is pointed *)
+  reflexivity.
+Defined.
```
Alizter, line 599:
```suggestion
Defined.

```

theories/Pointed/pEquiv.v
```diff
@@ -108,3 +108,54 @@ Proof.
...
+    apply eisadj. }
+  intros h.
+  rapply equiv_path_pforall.
+  snrapply Build_pHomotopy.
+  { intro x; cbn.
+    apply eissect. }
+  pointed_reduce.
+  unfold moveR_equiv_V.
+  hott_simpl.
+Defined.
```
Alizter, line 161:
```suggestion
Defined.

```

theories/WildCat/Equiv.v
```diff
@@ -323,3 +323,29 @@ Proof.
...
+      apply cate_issect. }
+    apply cat_idr. }
+  intros g.
+  apply path_hom.
+  refine (cat_assoc _ _ _ $@ _).
+  refine (_ $@ _).
+  { rapply cat_postwhisker.
+    apply cate_isretr. }
+  apply cat_idr.
+Defined.
```
Alizter, line 351:
```suggestion
Defined.

```

theories/Algebra/ooGroup.v
```diff
@@ -283,3 +285,34 @@ Section Subgroups.
...
+  snrapply Build_Is1Functor.
+  1: exact @functor2_pclassifyingspace.
+  1: exact functor_pclassifyingspace_idmap.
+  exact functor_pclassifyingspace_compose.
+Defined.
+
+
+
+
+
```
Alizter, line 318:
```suggestion
```

theories/Algebra/Groups/Group.v
```diff
@@ -511,3 +513,26 @@ Proof.
...
+(** A group is free if there exists a generating type on which it is a free group *)
+Class IsFreeGroup (F_S : Group)
+  := isfreegroup : exists S i, IsFreeGroupOn S F_S i.
+
+Global Instance isfreegroup_isfreegroupon (S : Type) (F_S : Group) (i : S -> F_S)
+  {H : IsFreeGroupOn S F_S i}
+  : IsFreeGroup F_S
+  := (S; i; H).
+
+
```
Alizter, line 538:
```suggestion
```

theories/Algebra/Groups/FreeGroup/LoopSusp.v
```diff
@@ -0,0 +1,128 @@
...
+    1: apply equiv_freegroup_rec.
+    intro f.
+    unfold equiv_freegroup_rec.
+    unfold transitive_equiv.
+    unfold equiv_compose.
+    unfold equiv_isequiv.
+    (** It's not obvious how this can be finished. *)
+  Abort.
+
+End AssumeUnivalence.
```
Alizter, line 128:
```suggestion
End AssumeUnivalence.

```

theories/Algebra/Groups/FreeGroup/KrausAltenkirch.v
```diff
@@ -0,0 +1,361 @@
...
+  Defined.
+
+  (** Typeclass search can already find this but we leave it here as a definition for reference. *)
+  Definition isfreegroup_freegrup `{Funext} : IsFreeGroup FreeGroup := _.
+
+End Reduction.
+
+
+
+
```
Alizter, line 361:
```suggestion
```

andrejbauer:
It seems misguided to call a formalisation of free groups by its author names. Nobody will be able to guess what's in the file.

Alizter:
@andrejbauer If you want free groups you don't have to know about the files with the authors names. Just import `theories/Algebra/FreeGroup.v`.

The file with the author names is just a specific implementation of free groups. It is named after the authors where I first saw this HIT (or a similar one) being described. The specific paper is mentioned in the file `theories/Algebra/FreeGroup.v`.

There are other implementations of free groups we could do such as the loop space of the suspension of a type with a basepoint added. This can be found in `theories/Algebra/FreeGroups/LoopSusp.v`. I ran into a few issues when implmenting free groups this way that I was not able to resolve so it is an unfinished implementation.

Alizter:
Looking back at this it would perhaps be better to make a module that describes what an implementation of a free group ought to have and have the files in `FreeGroups` implement this module. This would also make it easier to swap out implementations of free groups other than hoping they follow the same naming convention. Since we don't have more than one working implementation currently, and the fact that I highly doubt there will be a better implementation in the future, I don't think this is worth doing.

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -0,0 +1,60 @@
+Require Import Basics Types.
+Require Import Groups.Group.
+Require Import WildCat.
+
+(** Free groups are defined in Group.v. In this file we prove properties of free groups. There are various constructions of free groups to choose from, all of which can be found in the FreeGroup folder. *)
```
mikeshulman, line 5: This comment seems a bit misleading if there are actually only two constructions in the folder, of which only one is complete.  (-:

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -0,0 +1,60 @@
...
+  unfold hfiber.
+  snrapply contr_equiv'.
+  1: exact (FactorsThroughFreeGroup A F i G f).
+  { rapply equiv_functor_sigma_id.
+    intro g.
+    apply equiv_path_forall. }
+  exact _.
+Defined.
+
+(** The above theorem is true regardless of the implementation of free groups. This let's us state the more specific theorem about the free groups themselves. This can be read as [FreeGroup] is left adjoint to the forgetful functor [group_type]. *)
```
mikeshulman, line 31:
```suggestion
(** The above theorem is true regardless of the implementation of free groups. This lets us state the more specific theorem about the canonical free groups. This can be read as [FreeGroup] is left adjoint to the forgetful functor [group_type]. *)
```

theories/Algebra/Groups/Group.v
```diff
@@ -26,11 +26,13 @@ Record Group := {
   group_isgroup : IsGroup group_type;
 }.
 
-Arguments group_type {_}.
+(* Arguments group_type {_}. *)
```
mikeshulman, line 29: It seems odd to leave this here but commented out.  Should there be an explanatory comment on why?

theories/Algebra/Groups/Group.v
```diff
@@ -511,3 +513,24 @@ Proof.
...
+
+(** Universal property of a free group on a set (type). *)
+Class IsFreeGroupOn (S : Type) (F_S : Group) (i : S -> F_S)
+  := contr_isfreegroupon : forall (A : Group) (g : S -> A),
+      Contr (FactorsThroughFreeGroup S F_S i A g).
+Global Existing Instance contr_isfreegroupon.
+
+(** A group is free if there exists a generating type on which it is a free group *)
+Class IsFreeGroup (F_S : Group)
+  := isfreegroup : exists S i, IsFreeGroupOn S F_S i.
```
mikeshulman, line 531: I forget (and the reader probably does too) -- is this `exists` a truncated or untruncated one?  If untruncated, better to use sigma notation.

theories/Pointed/pEquiv.v
```diff
@@ -108,3 +108,54 @@ Proof.
   apply pmap_postwhisker.
   symmetry; apply peisretr.
 Defined.
+
+Definition equiv_pequiv_precompose `{Funext} {A B C : pType} (f : A <~>* B)
```
mikeshulman, line 112: Couldn't these be done in wildcat generality?

theories/Algebra/Groups/Group.v
Alizter: Untruncated

theories/Algebra/Groups/Group.v
Alizter: I will get rid of this all together. Since `group_type` is a coercion, you won't see it anyway, therefore it doesn't make much sense for it to have implicit arguments. For the rare times you actually need to specify something with it, it is more convenient to have access to the arguments.

theories/Pointed/pEquiv.v
Alizter: I could only find the precomposition one in the wildcat library . I've gone ahead and added the postcomp equivalence for wildcats too. These two lemmas are now stated in terms of wildcat. I will keep them so it's easier to find, especially for people not too familiar with the wildcat library.

Alizter:
Rebased.

theories/Algebra/Groups/FreeGroup.v
Alizter: I've added a little honesty.

Alizter:
I've renamed KrausAltenkirch to ListQuotient.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -264,3 +264,23 @@ Lemma pmap_pi_functor {X Y : pType} (f : X ->* Y) (n : nat)
 Proof.
   srapply Build_pHomotopy. 1: reflexivity. apply path_ishprop.
 Defined.
+
+(** Homotopy groups of truncations *)
+
+(** The fundamental group 1st truncation of X is isomorphic to the fundamental group of X *)
```
jdchristensen, line 270: Should say:  "The fundamental group of the 1-truncation of X ..."

jdchristensen:
```suggestion
(** The fundamental group of the 1-truncation of X is isomorphic to the fundamental group of X *) 
```

jdchristensen:
The list of files changed shows thousands of files in two submodules.  Can those be removed from this PR?

Does it make sense to include the "implementation" that isn't complete?

Other than that, this looks great to me (LGTM).

jdchristensen:
Sorry, I didn't notice that this was already merged...

mikeshulman:
Sorry, I missed noticing the submodule update.  What was that about?

Alizter:
Sorry about that, when I rebased my submodules seem to have been stuck in the past. I'll send a PR updating all of them for good.

PR #1370, opened 2020-07-16
Pasting for pullbacks

### LINKED ISSUE #1012: Two pullbacks lemma
So currently in my formalisation of Blakers-Massey I am in need of the two pullbacks lemma (or pasting law for pullbacks):

Given:

```
A --> B --> C
|     |     |
V     V     V
D --> E --> F
```

where the right square is a pullback, then the outer square is a pullback if and only if the left square is a pullback.

It seems this is not so easy to prove, although [a formalisation in coq exists](https://github.com/peterlefanulumsdaine/hott-limits). This however is a bit different to how we have defined pullbacks so I am wondering what the best way to prove this will be.

Alizter:
In this PR we prove the pasting lemma for pullbacks in both directions. We also show that a commutative square is a pullback if and only if `functor_hfiber` is an equivalence.

closes #1012

### DIFF
--- a/theories/Limits/Pullback.v
+++ b/theories/Limits/Pullback.v
@@ -353,7 +353,7 @@ Section Pasting.
     pose (e2' := isequiv_functor_hfiber_ispullback _ e2 b).
     snrapply isequiv_commsq'.
     7: apply isequiv_idmap.
-    4: apply (functor_hfiber_compose b).
+    4: apply (functor_hfiber_compose H K b).
     1,2: exact _.
   Defined.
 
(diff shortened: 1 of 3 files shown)


theories/Limits/Pullback.v
```diff
@@ -112,8 +113,8 @@ Proof.
   symmetry; apply inv_V.
 Defined.
 
-(** If the induced map on fibers is an equivalence, then a square is a pullback. *)
-Definition ispullback_isequiv_functor_hfiber {A B C D}
+(** If the induced maps on fibers is an equivalence, then a square is a pullback. *)
```
mikeshulman, line 116:
```suggestion
(** If the induced maps on fibers are equivalences, then a square is a pullback. *)
```

Alizter:
I should note that I should really have used the wildcat squares as to slowly replace CommutativeSquares.v. But currently, I am undecided on many things in the squares library so I have added horiztonal composition of squares for the existing implementation.

Issue #1348, opened 2020-06-24
Rename S1 to Circle


Alizter:
I think it would be a good idea to distinguish between the names `S1` and `Circle`.

* We don't define S1 to be `S 1`. This *could* be confusing to users.
* HoTT vernacular typically denotes this type as the circle, so for new users of the library, searching circle might be more logical than S1.
* The name S1 gets cluttered. We might want to use it to denote the 1st sphere.

peterlefanulumsdaine:
I don’t have strong feelings on this.  I wouldn’t quite see it as needing changing for my part, but if you want to change it then I don’t oppose that.

mikeshulman:
Could we make a notation or definition alias so that both names could be found by search?

PR #1372, opened 2020-07-19
Moved Spheres and Circle to Spaces, renamed S1 to Circle, S2 to TwoSphere + general cleanup

Alizter:
I've renamed S1 to Circle, S2 to TwoSphere and relaxed various names in Spheres.v. I've also moved Circle, TwoSphere and Spheres to the Spaces folder.

I'm marking this PR as a draft for now as Mike wanted a notation for Circle in #1348. I haven't worked out how to do this properly yet.


Alizter:
I've made an `only parsing` notation for `Circle` called `S1` which means it can be searched.

closes #1348.

PR #1374, opened 2020-07-27
Update to Coq V8.12

Alizter:
Coq 8.12 has been released. I've updated the Coq submodule and updated INSTALL.md.

This is a draft for now as there are a lot of new warnings we will have to resolve.

When I make the library on my machine I get the following warnings:

<details>
<summary> `make -j16 > /dev/null` </summary>
  
```
*** Warning: in file coq/theories/Init/Prelude.v, declared ML module numeral_notation_plugin has not been found!
*** Warning: in file coq/theories/Init/Notations.v, declared ML module ltac_plugin has not been found!
File "./coq/theories/Init/Peano.v", line 37, characters 0-30:
Warning: Declaring arbitrary terms as hints is fragile; it is recommended to
declare a toplevel constant instead [fragile-hint-constr,automation]
File "./coq/theories/Init/Peano.v", line 38, characters 0-38:
Warning: Declaring arbitrary terms as hints is fragile; it is recommended to
declare a toplevel constant instead [fragile-hint-constr,automation]
File "./coq/theories/Init/Prelude.v", line 19, characters 0-91:
Warning: Deprecated Numeral Notation for Decimal.uint, Decimal.int or
Decimal.decimal. Use Numeral.uint, Numeral.int or Numeral.numeral
respectively. [decimal-numeral-notation,deprecated]
File "./coq/theories/Init/Prelude.v", line 21, characters 0-85:
Warning: Deprecated Numeral Notation for Decimal.uint, Decimal.int or
Decimal.decimal. Use Numeral.uint, Numeral.int or Numeral.numeral
respectively. [decimal-numeral-notation,deprecated]
File "./coq/theories/Init/Prelude.v", line 25, characters 0-79:
Warning: Deprecated Numeral Notation for Decimal.uint, Decimal.int or
Decimal.decimal. Use Numeral.uint, Numeral.int or Numeral.numeral
respectively. [decimal-numeral-notation,deprecated]
File "./coq/theories/Init/Peano.v", line 96, characters 0-34:
Warning: Declaring arbitrary terms as hints is fragile; it is recommended to
declare a toplevel constant instead [fragile-hint-constr,automation]
File "./coq/theories/Init/Peano.v", line 97, characters 0-50:
Warning: Declaring arbitrary terms as hints is fragile; it is recommended to
declare a toplevel constant instead [fragile-hint-constr,automation]
File "./coq/theories/Init/Peano.v", line 136, characters 0-35:
[...]

Issue #1367, opened 2020-07-13
Cleanup nat


Alizter:
There are various places scattered throughout the library having to do with `nat`. I propose that we start collecting these in Spaces.Nat similar to how Spaces.Int is organized.

I recently had a look if it is possible to completely remove the coq stdlib, it seems to be the case but one of the stumbling blocks is nat. Currently various files (mostly in mathclasses) rely on Coq.Init.Peano and these lemmas need to be collected elsewhere in the main library and merged with existing results. For example, there are two different definitions of less than or equal to for naturals, and lemmas about those etc. Once this has been done, it will allow the transition from stdlib to no stdlib be much smoother.

One other thing that we can do is move the definition of `nat` into the main library and perhaps rename it `Nat` to fit with the library naming conventions.

spitters:
https://github.com/HoTT/HoTT/issues/1424

PR #1376, opened 2020-08-05
fix fragile-hint-constr warning

Alizter:
I've removed the warning suppression(s) introduced in #1374 for `fragile-hint-constr` and fixed all offenders.


coq/theories/Init/Peano.v
```diff
@@ -34,8 +34,10 @@ Local Notation "0" := O.
 
 Definition eq_S := f_equal S.
 
-Hint Resolve (f_equal S): v62.
-Hint Resolve (f_equal (A:=nat)): core.
+Local Definition f_equal_S := f_equal S.
+Local Definition f_equal_nat := f_equal (A:=nat).
+Hint Resolve f_equal_S : v62.
```
JasonGross, line 39: You could remove this one, possibly even both of them.

Alizter: I tried that initially, but some of the proofs here use `auto` and don't seem to work without it. This file will be completely removed in the future anyway, once I finish a PR for #1367.

PR #1377, opened 2020-08-06
[Group] (amalgamated) free products, presentations

Alizter:
Here is a WIP formalization of presentations. We construct the amalgamated free product of a group as a certain HIT (built from various coequalizers). From that we can define group presentations together with finiteness conditions on generators and relations.

I am making this a draft PR as I hope to also add a characterization of group homomorphisms between presented groups. i.e. every group homomorphism is determined by its action on generators and relations.


Alizter:
I was using the new `pose proof (X := foo)` syntax which is only available in 8.12. But since #1380 isn't merged yet, the submodule still on 8.11 where this syntax doesn't exist. I've therefore swapped it out for the equivalent using `assert`.

theories/Algebra/Groups/Presentation.v
```diff
@@ -0,0 +1,116 @@
...
+Record GroupPresentation := {
+  (** We have a type of generators *)
+  gp_generators : Type ;
+  (** An indexing type for relators *)
+  gp_rel_index : Type ;
+  (** The relators are picked out amongst elements of the free group on the generators. *)
+  gp_relators : gp_rel_index -> FreeGroup gp_generators;
+}.
+
+(** Note: A relator is a relation in the form of "f(x) = 1", any relation "f(x) = g(x)" can become a relator "f(x)g(x)^-1 = 1". *)
```
mikeshulman, line 20: Why `f(x)`?  It's not the value of a function, is it?

theories/Algebra/Groups/Presentation.v
```diff
@@ -0,0 +1,116 @@
...
+
+(** Given the data of a group presentation we can construct the group. This is sometimes called the presented group. *)
+Definition group_gp : GroupPresentation -> Group.
+Proof.
+  intros [X I R].
+  exact (GroupCoeq
+    (FreeGroup_rec I (FreeGroup X) R)
+    (FreeGroup_rec I (FreeGroup X) (fun x => mon_unit))).
```
mikeshulman, line 28:
```suggestion
    (FreeGroup_rec I (FreeGroup X) (fun x => group_unit))).
```
?

theories/Algebra/Groups/Presentation.v
```diff
@@ -0,0 +1,116 @@
...
+
+Coercion presentation : HasPresentation >-> GroupPresentation.
+
+(** Here are a few finiteness properties of group presentations. *)
+
+(** A group presentation is finitely generated if its generating set is finite. *)
+Class FinitelyGeneratedPresentation (P : GroupPresentation)
+  := finite_gp_generators : Finite (gp_generators P).
+
+(** A group presentation is finitely related if its relators indexing set is finite. *)
```
mikeshulman, line 45: I've never heard of this notion before, although I see now that it exists elsewhere on the Internet at least.  Do you know of any use for it other than as half of "finitely presented"?

theories/Algebra/Groups/FreeProduct.v
```diff
@@ -0,0 +1,749 @@
...
+   | omega_H : forall (x y : list (H + K)),
+      amal_eta (x ++ [inl mon_unit] ++ y) = amal_eta (x ++ y)
+   | omega_K : forall (x y : list (H + K)),
+      amal_eta (x ++ [inr mon_unit] ++ y) = amal_eta (x ++ y).
+
+  We will build this HIT up sucessively out of coequalizers. *)
+
+(** We will call M [amal_type] and prefix all the constructors with [amal_] (for amalgmated free product). *)
+
+(** Here are some operations on lists from the coq stdlib *)
```
mikeshulman, line 32: Maybe these should go elsewhere?

theories/Algebra/Groups/FreeProduct.v
```diff
@@ -0,0 +1,749 @@
...
+      apply word_concat_w_nil. }
+    simpl.
+    destruct x; refine (_ @ (word_concat_w_ww _ _ _)^); f_ap.
+  Defined.
+
+
+  (** There are five source types for the path constructors. We will construct this HIT as the colimit of five forks going into [Words]. We can bundle up this colimit as a single coequalizer. *)
+
+  (** Source types of path constructors *)
+  Definition pc1 : Type := Words * H * H * Words.
```
mikeshulman, line 121: I think these definitions shouldn't be global.

theories/Algebra/Groups/Presentation.v
Alizter: I meant it as an f for formula rather than an f for function.

theories/Algebra/Groups/Presentation.v
Alizter: Looks like that didn't quite work. I will fix this in a bit.

theories/Algebra/Groups/FreeProduct.v
Alizter: I agree that these should go elsewhere. I did the same thing in #1365 but I can't get away with it this time :-)

For now I would recommend keeping it as it is until we can write down a small library of list stuff.

theories/Algebra/Groups/FreeProduct.v
Alizter: I agree.

It occurs to me that there is an idea of [search blacklisting](https://coq.inria.fr/refman/proof-engine/vernacular-commands.html#coq:table.search-blacklist). This would allow us to have global names that don't get picked up by `Search`. The local names already get picked up by search, but are also qualified by the module.

Currently we have a few blacklisted prefixes in our coq stdlib.

theories/Algebra/Groups/Presentation.v
Alizter: Weirdly `mon_unit` can correctly infer which group it is talking about whereas `group_unit` cannot. I had to tell `group_unit` which group it was the unit of here.

Maybe that's not so surprising since `mon_unit` looks for (things contained in) `IsGroup` instances whereas `group_unit` can't really do a typeclass search since `Group` is not a class.

Anyhow, I've fixed it.

theories/Algebra/Groups/Presentation.v
mikeshulman: But then what is `x`?

theories/Algebra/Groups/FreeProduct.v
mikeshulman: I think it would be better to just use modules.  Namespacing was invented precisely to manage large software developments, after all.

theories/Algebra/Groups/FreeProduct.v
mikeshulman: Why not just start the list library now with the stuff we've needed so far?  It doesn't have to be comprehensive, we can add to it as we need it.

theories/Algebra/Groups/Presentation.v
Alizter: I can't remember where I heard about groups being finitely related, however there is a analogous situation for modules. I remember that you can show any finitely related module is the direct sum of a finitely presented module and a free module.

I think I can conjure up some examples however. Take the free group on (at least) countably many generators F and take the free product with any finitely presented group. This should give you a finitely related group that is not finitely generated. I can't say if this is an interesting thing to do however.

theories/Algebra/Groups/FreeProduct.v
Alizter: I've created an issue #1384. For now I will keep them where they are.

theories/Algebra/Groups/Presentation.v
Alizter: I'll come up with a better comment 0:-)

theories/Algebra/Groups/FreeGroup/ListQuotient.v
```diff
@@ -13,37 +13,37 @@ Section Reduction.
...
-  Definition change_sign : A + A -> A + A
+  Local Definition change_sign : A + A -> A + A
     := fun x => match x with
                 | inl a => inr a
                 | inr a => inl a
                 end.
 
   (** We introduce a local notation for [change_sign]. It is only defined in this section however. *)
-  Notation "x ^" := (change_sign x).
+  Local Notation "x ^" := (change_sign x).
```
mikeshulman, line 26: As you probably know, this change doesn't do anything, since notations are already local to sections.  I'm divided about whether it's a good or a bad idea to include it anyway.  On the one hand it emphasizes to the reader that the notation is local; but on the other hand it might create the mistaken impression that non-local notations survive sections.

Alizter: Is it any different from writing `Global Instance` outside of a section? I mean we never actually have to do it, but we include it as good practice.

theories/Algebra/Groups/Presentation.v
Alizter: Actually, upon pondering this some more it seems you can characterize finitely related groups in a similar way to how you do it for modules. If you have a group presentation <G|R> where R is finite, then you know that only finitely many elements of G occur in words in R. This should allow you to write <G|R> as the free product of a (still infinitely generated) free group and a finitely presented group. Thus every finitely related group is the free product of a free group and a finitely presented group modulo details.

theories/Algebra/Groups/FreeGroup/ListQuotient.v
mikeshulman: Maybe not.  Perhaps it's just residual annoyance at the coq devs for using the same keywords `Local`/`Global` to mean different things for different commands.

PR #1381, opened 2020-08-10
added Zulip badge to README.md

Alizter:


mikeshulman:
Zulip join links are temporary, aren't they?  So probably it doesn't make sense to include one in our readme, unless we plan to update it every few weeks.

Alizter:
@mikeshulman You're right. Earlier I couldn't get the URL to work without an invite. It seems to be okay now. I'll force push this through.

mikeshulman:
I believe one does need an invite link to create an account and access zulip.  Given that, and that we can't post a permanent invite link, is it really worthwhile adding this to our readme?

Alizter:
Our Zulip organization is not invite only. The new link is: `https://hott.zulipchat.com/` which should be accessible by anyone. I think it is worth adding this badge as some people may want to ask questions about the library without creating a new issue etc.

mikeshulman:
ok

PR #1382, opened 2020-08-14
[Overture] Get discriminate working

artagnon:
Register a few things so that the discriminate tactic starts working.

Co-authored-by: Hugo Herbelin <Hugo.Herbelin@inria.fr>

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -266,7 +266,7 @@ Proof.
   intros y p.
   destruct p.
   exact u.
-Defined.                                  
+Defined.
 
 (** We declare a scope in which we shall place path notations. This way they can be turned on and off by the user. *)
 
@@ -671,6 +671,7 @@ Ltac path_via mid :=
 
 (** We put [Empty] here, instead of in [Empty.v], because [Ltac done] uses it. *)
 Inductive Empty : Type0 := .
+Register Empty as core.False.type.
 
 Scheme Empty_ind := Induction for Empty Sort Type.
 Scheme Empty_rec := Minimality for Empty Sort Type.
@@ -708,7 +709,9 @@ Definition Unit_rect := Unit_ind.
 Hint Resolve tt : core.
 
 Register Unit as core.IDProp.type.
+Register Unit as core.True.type.
 Register tt as core.IDProp.idProp.
+Register tt as core.True.I.
 
 (** *** Pointed types *)
 
--- a/theories/Tests.v
+++ b/theories/Tests.v
@@ -88,3 +88,43 @@ Module Issue_973.
   end.
 
 End Issue_973.
+
+
+Module PR_1382.
+  (* Tests for discriminate tactic *)
+
+  Goal O = S O -> Empty.
+   discriminate 1.
+  Qed.
+
+  Goal forall H : O = S O, H = H.
+   discriminate H.
+  Qed.
+
+  Goal O = S O -> Unit.
+  intros H. discriminate H. Qed.
+  Goal O = S O -> Unit.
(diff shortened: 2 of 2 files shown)


theories/Basics/Overture.v
```diff
@@ -699,6 +700,8 @@ Class Asymmetric {A} (R : Relation A) :=
 
 (** Likewise, we put [Unit] here, instead of in [Unit.v], because [Trunc] uses it. *)
 Inductive Unit : Type0 := tt : Unit.
+Register Unit as core.True.type.
```
Alizter, line 703: Can we put all the `Register` statements for `Unit` together? Starting from 713?

Alizter:
LGTM modulo my comments. Do you also have any examples of using the discriminate tactic?

artagnon:
Yeah, the simplest reduced case example is `0 <> 1`. We were trying to prove properties of natural numbers like `Theorem O_or_S n : {m : nat | S m = n} + {0 = n}.` (see [this](https://coq.inria.fr/distrib/current/stdlib/Coq.Arith.Peano_dec.html)) when we ran into this.

artagnon:
`inversion` doesn't work either (`core.eq.congr` undefined). I might have a remedy within the next couple of days. If the tactic works like in vanilla Coq, it should be a good addition.

Alizter:
I think you need to register the following:
```coq
   Register ap as core.eq.congr.
   Register concat as core.eq.trans.
   Register inverse as core.eq.sym.
   Register idpath as core.eq.refl.
   Register paths_ind as core.eq.ind.
   Register paths as core.eq.type.
```

Some of these are sort of registered in Overture.v already, they just need to have the name `core.identity` changed to `core.id`. Even after trying this I couldn't get the `inversion` tactic to work, it just seems to duplicate the hypothesis.

Alizter:
If you need to inversion tactic you'll just have to use transport in the way that was explained in the Zulip chat. We have examples of inversion principles in `theores/Spaces/Int/Core.v` proven using transport.

spitters:
@Alizter 
> the internal ocaml varbiable core.eq.ind need the induction principle to eliminate into prop families

Is that essential, or just a legacy problem?
In general, @ejgallego has been very doing great work on getting the tactics to be independent of the stdlib. So, he might have some insights too.

SkySkimmer:
>`   Register paths_ind as core.eq.ind.`

Shouldn't it be `as core.eq.rect`?

artagnon:
The build breaks when I put in these `Register` lines:

```
File "./theories/Basics/PathGroupoids.v", line 579, characters 2-23:
Error: Found no subterm matching "inv_V q" in the current goal.
```

Alizter:
@artagnon This seems to be caused by `Register paths as core.eq.type.`

I'm not sure why this is however.

Alizter:
> > `   Register paths_ind as core.eq.ind.`
> 
> Shouldn't it be `as core.eq.rect`?

Well without registering `core.eq.ind` the `inversion` tactic spits an error.

Alizter:
For whatever reason `Register paths as core.eq.type.` causes rewrite to break. It looks like rewrite doesn't see the argument as an equality to rewrite along but as a term to match and rewrite.

Alizter:
It looks like there is a coq bug here, but I am not sure. The coq developers are mostly on holiday at the moment, so it will be probably a few weeks before we can find a fix.

Alizter:
I've added a test to make sure that discriminate continues to work. I'll merge this once the CI is happy. The problems with inversion seem to be specific to it so we should continue that discussion in it's PR. It's also worth noting we now have some upstream fixes that will help compatibility coq/coq#12847.

But for now, everything here looks and works fine (from what I've tested).

PR #1383, opened 2020-08-17
[Overture] Get inversion working

artagnon:
The series of Register commands breaks build at the moment.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -227,11 +227,13 @@ Scheme paths_rec := Minimality for paths Sort Type.
 Arguments paths_rec [A] a P f y p.
 
 Register idpath as core.identity.refl.
+Register idpath as core.eq.refl.
 
 (* See comment above about the tactic [induction]. *)
 Definition paths_rect := paths_ind.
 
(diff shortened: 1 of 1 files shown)


Alizter:
I wouldn't replace the identity variables with the eq ones. That was a mistake in my post which I have since edited.

Alizter:
What's `core.id.*`? I am only aware of `core.eq.*` and `core.identity.*`.

Alizter:
The current error stems from the fact that registering `core.eq.type` changes the behaviour of rewrite.

artagnon:
> What's core.id.*? I am only aware of core.eq.* and core.identity.*.

My bad. I think I misunderstood your comment on another thread.

mikeshulman:
What's the status of this?  (And actually, what's the goal?)

Alizter:
@mikeshulman The `inversion` tactic can prove inequalities between differing constructors of inductive types. On the coq end things are being changed (such as the registering of the empty type etc.) to allow us to use it.

For now however, it doesn't seem like this PR actually does what it says, so I am in favour of closing it until things change upstream.

PR #1385, opened 2020-08-20
removed "oo" notation for purely

### LINKED ISSUE #1373: Remove "oo" as a notation for purely
I think `oo` made more sense when we wrote just `n` in more places for the truncation modality rather than `Tr n`.  I'd be fine with getting rid of it (in another PR, presumably).

_Originally posted by @mikeshulman in https://github.com/HoTT/HoTT/pull/1342_

Alizter:
Hopefully it still makes sense. I've also replaced the terminology "oo-exact" with "purely-exact" but there is probably something better to say.

closes #1373

### DIFF
--- a/theories/Homotopy/ExactSequence.v
+++ b/theories/Homotopy/ExactSequence.v
@@ -193,7 +193,7 @@ Proof.
   rewrite (concat_A1p (eisretr k)), concat_V_pp. reflexivity.
 Defined.
 
-(** When [n] is the identity modality [oo], so that [cxfib] is an equivalence, we get simply a fiber sequence.  In particular, the fiber of a given map yields an oo-exact sequence. *)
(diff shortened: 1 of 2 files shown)


theories/Homotopy/ExactSequence.v
```diff
@@ -193,7 +193,7 @@ Proof.
   rewrite (concat_A1p (eisretr k)), concat_V_pp. reflexivity.
 Defined.
 
-(** When [n] is the identity modality [oo], so that [cxfib] is an equivalence, we get simply a fiber sequence.  In particular, the fiber of a given map yields an oo-exact sequence. *)
+(** When [n] is the identity modality [oo], so that [cxfib] is an equivalence, we get simply a fiber sequence.  In particular, the fiber of a given map yields an purely-exact sequence. *)
```
mikeshulman, line 196:
```suggestion
(** When [n] is the identity modality [purely], so that [cxfib] is an equivalence, we get simply a fiber sequence.  In particular, the fiber of a given map yields an purely-exact sequence. *)
```

mikeshulman:
I can't decide what I think about "purely exact".

PR #1393, opened 2020-09-01
Define and characterize projective types

jarlg:
Co-authored-by: Dan Christensen <jdc@uwo.ca>

Defines projective types and characterizes them as types for which the axiom of choice holds, or alternatively types into which every epi (merely) splits. Under `AC_oo,-1` (in the sense of the HoTT Book, Exercise 7.8) projective types correspond to sets.

Questions:
- Is `Spaces/Projective.v` a good placement? `Types/Projective.v` didn't work out due to circular imports.
- Should the `AC_oo,neg1` section at the end of `Projectivity.v` be placed elsewhere? An alternative placement would be in `HSet.v`, in which case the definition of `choice` in `Projectivity.v` would need to be factored out into it's own file to avoid circular imports. (In this case, `Spaces/Finite.v` could also import this definition of choice, and `finite_choice` would implement `choice X` for finite `X`.)

Lastly, the proposition `projective_cover_AC` solves exercise 7.8 from the HoTT Book for n = oo, so a comment was added about that.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1619,9 +1619,17 @@ End Book_7_1.
 
 (* ================================================== ex:acnm-surjset *)
 (** Exercise 7.9 *)
-(* Solved for AC_(oo,-1) by projective_cover_AC in Spaces/Projective.v *)
 
+Section Book_7_9.
+  (** Solution for the case (oo,-1). *)
 
+  Axiom AC_oo_neg1 : forall X : hSet, Choice X.
+
+  Definition Book_7_9_oo_neg1 `{Univalence} (A : Type)
+    : merely (exists X : hSet, exists p : X -> A, IsSurjection p)
+    := @HoTT.Projective.projective_cover_AC AC_oo_neg1 _ A.
+
+End Book_7_9.
 
 (* ================================================== ex:acconn *)
 (** Exercise 7.10 *)
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1620,16 +1620,10 @@ End Book_7_1.
 (* ================================================== ex:acnm-surjset *)
 (** Exercise 7.9 *)
 
-Section Book_7_9.
-  (** Solution for the case (oo,-1). *)
-
-  Axiom AC_oo_neg1 : forall X : hSet, Choice X.
-
-  Definition Book_7_9_oo_neg1 `{Univalence} (A : Type)
-    : merely (exists X : hSet, exists p : X -> A, IsSurjection p)
-    := @HoTT.Projective.projective_cover_AC AC_oo_neg1 _ A.
-
-End Book_7_9.
+(** Solution for the case (oo,-1). *)
+Definition Book_7_9_oo_neg1 `{Univalence} (AC_oo_neg1 : forall X : hSet, Choice X) (A : Type)
+  : merely (exists X : hSet, exists p : X -> A, IsSurjection p)
+  := @HoTT.Projective.projective_cover_AC AC_oo_neg1 _ A.
 
 (* ================================================== ex:acconn *)
 (** Exercise 7.10 *)
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1621,7 +1621,7 @@ End Book_7_1.
 (** Exercise 7.9 *)
 
 (** Solution for the case (oo,-1). *)
-Definition Book_7_9_oo_neg1 `{Univalence} (AC_oo_neg1 : forall X : hSet, Choice X) (A : Type)
+Definition Book_7_9_oo_neg1 `{Univalence} (AC_oo_neg1 : forall X : hSet, IsProjective' X) (A : Type)
   : merely (exists X : hSet, exists p : X -> A, IsSurjection p)
   := @HoTT.Projective.projective_cover_AC AC_oo_neg1 _ A.
 
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1619,6 +1619,7 @@ End Book_7_1.
 
 (* ================================================== ex:acnm-surjset *)
 (** Exercise 7.9 *)
+(* Solved for AC_(oo,-1) by projective_cover_AC in Spaces/Projective.v *)
 
 
 
--- a/theories/HSet.v
+++ b/theories/HSet.v
@@ -121,6 +121,13 @@ Proof.
   exact (ap pr1 q).
 Defined.
 
+Definition isinj_section {A B : Type} {s : A -> B} {r : B -> A}
+      (H : r o s == idmap) : isinj s.
+Proof.
+  intros a a' alpha.
+  exact ((H a)^ @ ap r alpha @ H a').
+Defined.
+
 Lemma isembedding_isinj_hset {A B : Type} `{IsHSet B} (m : A -> B)
 : isinj m -> IsEmbedding m.
 Proof.
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -94,7 +94,7 @@ Section Subuniverse.
   : MapIn O g.
   Proof.
     intros b.
-    refine (inO_equiv_inO (hfiber f b)
+    exact (inO_equiv_inO (hfiber f b)
                           (equiv_hfiber_homotopic f g p b)).
   Defined.
 
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -51,6 +51,14 @@ Definition inO_equiv_inO' {O : Subuniverse}
   : In O U
   := inO_equiv_inO T f.
 
+Definition inO_equiv_inO'' `{Funext} (O : Subuniverse) {T U : Type} (f : T <~> U)
+  : In O T <~> In O U.
+Proof.
+  apply equiv_iff_hprop.
+  - intro T_inO; exact (inO_equiv_inO' T f).
+  - intro U_inO; exact (inO_equiv_inO' U f^-1%equiv).
+Defined.
+
 (** The universe of types in the subuniverse *)
 Definition Type_@{i j} (O : Subuniverse@{i}) : Type@{j}
   := @sig@{j i} Type@{i} (fun (T : Type@{i}) => In O T).
@@ -69,6 +77,54 @@ Definition equiv_path_TypeO@{i j} {fs : Funext} O (T T' : Type_ O)
 (** Types in [TypeO] are always in [O]. *)
 Global Instance inO_TypeO {O : Subuniverse} (A : Type_ O) : In O A
   := A.2.
+
+(** ** Properties of Subuniverses *)
+
+(** A map is O-local if all its fibers are. *)
+Class MapIn (O : Subuniverse) {A B : Type} (f : A -> B)
+  := inO_hfiber_ino_map : forall (b:B), In O (hfiber f b).
+
+Global Existing Instance inO_hfiber_ino_map.
+
+Section Subuniverse.
+  Context (O : Subuniverse).
+
+  (** Anything homotopic to a local map is local. *)
+  Definition mapinO_homotopic {A B : Type} (f : A -> B) {g : A -> B}
+             (p : f == g) `{MapIn O _ _ f}
+  : MapIn O g.
+  Proof.
+    intros b.
+    refine (inO_equiv_inO (hfiber f b)
+                          (equiv_hfiber_homotopic f g p b)).
+  Defined.
+
+  (** The projection from a family of local types is local. *)
+  Global Instance mapinO_pr1 {A : Type} {B : A -> Type}
+         `{forall a, In O (B a)}
+  : MapIn O (@pr1 A B).
+  Proof.
+    intros a.
+    exact (inO_equiv_inO (B a) (hfiber_fibration a B)).
+  Defined.
+
+  Lemma equiv_forall_inO_mapinO_pr1 `{Funext} {A : Type} (B : A -> Type)
+    : (forall a, In O (B a)) <~> MapIn O (@pr1 A B).
+  Proof.
+    unfold MapIn.
+    apply equiv_functor_forall_id; intro a.
+    apply inO_equiv_inO''.
+    exact (hfiber_fibration a B).
+  Defined.
+
+  (** Being a local map is an hprop *)
+  Global Instance ishprop_mapinO `{Funext} {A B : Type} (f : A -> B)
+  : IsHProp (MapIn O f).
+  Proof.
+    apply trunc_forall.
+  Defined.
+
+End Subuniverse.
 
 (** *** Reflections *)
 
@@ -1210,6 +1266,14 @@ Section ConnectedTypes.
   : IsConnected O A -> IsConnected O B
     := isconnected_equiv A f.
 
+  (** The O-connected types form a subuniverse. *)
+  Definition Conn : Subuniverse.
+  Proof.
+    rapply (Build_Subuniverse (IsConnected O)).
+    simpl; intros T U isconnT f isequivf.
+    exact (isconnected_equiv T f isconnT).
+  Defined.
+
   (** Connectedness of a type [A] can equivalently be characterized by the fact that any map to an [O]-type [C] is nullhomotopic.  Here is one direction of that equivalence. *)
   Definition isconnected_elim {A : Type} `{IsConnected O A} (C : Type) `{In O C} (f : A -> C)
   : NullHomotopy f.
@@ -1310,13 +1374,6 @@ End ConnectedTypes.
 
 (** ** Modally truncated maps *)
 
-(** A map is "in [O]" if each of its fibers is. *)
-
-Class MapIn (O : ReflectiveSubuniverse) {A B : Type} (f : A -> B)
-  := inO_hfiber_ino_map : forall (b:B), In O (hfiber f b).
-
-Global Existing Instance inO_hfiber_ino_map.
-
 Section ModalMaps.
   Context (O : ReflectiveSubuniverse).
 
@@ -1325,25 +1382,6 @@ Section ModalMaps.
   : MapIn O f.
   Proof.
     intros b; exact _.
-  Defined.
-
-  (** Anything homotopic to a modal map is modal. *)
-  Definition mapinO_homotopic {A B : Type} (f : A -> B) {g : A -> B}
-             (p : f == g) `{MapIn O _ _ f}
-  : MapIn O g.
-  Proof.
-    intros b.
-    refine (inO_equiv_inO (hfiber f b)
-                          (equiv_hfiber_homotopic f g p b)).
-  Defined.
-
-  (** The projection from a family of modal types is modal. *)
-  Global Instance mapinO_pr1 {A : Type} {B : A -> Type}
-         `{forall a, In O (B a)}
-  : MapIn O (@pr1 A B).
-  Proof.
-    intros a.
-    refine (inO_equiv_inO (B a) (hfiber_fibration a B)).
   Defined.
 
   (** A slightly specialized result: if [Empty] is modal, then a map with decidable hprop fibers (such as [inl] or [inr]) is modal. *)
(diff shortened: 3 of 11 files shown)


theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+    refine (tr (pullback_pr1 o s; _)).
+    intro x.
+    refine (pullback_commsq p f (s x) @ _).
+    apply (ap f).
+    apply E.
+Defined.
+
+(** ** Projectivity and choice *)
+
+Definition choice (X : Type) := forall P : X -> Type,
```
Alizter, line 36:
```suggestion
Definition choice (X : Type) : Type := forall P : X -> Type,
```

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
+Require Import Basics Types HProp HSet HFiber Truncations.
+Require Import Modalities.Descent Modalities.Separated.
+Require Import Limits.Pullback Spaces.Finite.
+
+
+(** * Projective types *)
+
+Definition IsProjective (X : Type)
```
Alizter, line 8:
```suggestion
Definition IsProjective (X : Type) : Type
```

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+  := forall A B : Type, forall f : X -> B, forall p : A -> B,
+        IsSurjection p -> merely (exists s : X -> A, p o s == f).
+
+(** A type X is projective if and only if surjections into X merely split. *)
+Proposition equiv_isprojective_surjections_split `{Funext} (X : Type)
```
Alizter, line 13: I take it Funext is only used to show the RHS is a hprop correct?

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+    apply E.
+Defined.
+
+(** ** Projectivity and choice *)
+
+Definition choice (X : Type) := forall P : X -> Type,
+      (forall x, merely (P x)) -> merely (forall x, P x).
+
+(* The following two lemmas make the proof of equiv_isprojective_choice below easier to follow. *)
+Lemma equiv_contr_hprop `{Funext} (A : Type) `{IsHProp A}
```
Alizter, line 40: This should probably go in `theories/Basics/Trunc.v` if its not in there already.

Alizter:
I've been meaning to add a choice file for a while, but it was never really needed for what I was up to. I think a file in theories collecting all choice principles (type-theoretic and otherwise) should be created.

I'm also unsure if Spaces it the best placement for this since it is about a property of spaces rather than an example itself. I think for now putting it in theories would be a good idea, and then moving properties like `Fin n` being protective to the Spaces/Finite.v file.

But let's first see what others say, I'd be interested to see if anybody else has thoughts on organisation.

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+      destruct S as [s h].
+      rapply (istrunc_embedding_trunc s).
+      apply isembedding_isinj_hset.
+      exact (isinj_section h).
+    - intro ishsetX.
+      apply (equiv_isprojective_choice X)^-1.
+      rapply AC.
+  Defined.
+
+End AC_oo_neg1.
```
Alizter, line 142:
```suggestion
End AC_oo_neg1.

```
We usually add spaces to the end of files. I can't really remember why but it is something we do :)

Alizter:
Also in general, Types is meant for the "core basic types" of type theory. I think we are generally inclined to put other types in "Spaces".

theories/Spaces/Projective.v
Alizter: Also, should `choice` be `Choice` to follow the convention on types?

Alizter:
Also I think that a file called Projective in a folder Spaces might send the wrong message, i.e. a file about RP^n.

theories/HSet.v
```diff
@@ -121,6 +121,13 @@ Proof.
   exact (ap pr1 q).
 Defined.
 
+Definition isinj_section {A B : Type} {s : A -> B} {r : B -> A}
+      (H : r o s == idmap) : isinj s.
```
Alizter, line 125: Perhaps use `Sect`?

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+(** * Projective types *)
+
+Definition IsProjective (X : Type)
+  := forall A B : Type, forall f : X -> B, forall p : A -> B,
+        IsSurjection p -> merely (exists s : X -> A, p o s == f).
+
+(** A type X is projective if and only if surjections into X merely split. *)
+Proposition equiv_isprojective_surjections_split `{Funext} (X : Type)
+  : IsProjective X <~>
+    (forall (Y:Type), forall (p : Y -> X),
```
Alizter, line 15:
```suggestion
    (forall (Y : Type), forall (p : Y -> X),
```

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+          IsSurjection p -> merely (exists s : X -> Y, p o s == idmap)).
+Proof.
+  srapply equiv_iff_hprop.
+  - intro isprojX. unfold IsProjective in isprojX.
+    intros Y p S.
```
Alizter, line 20:
```suggestion
  - intro isprojX Y p S.
    unfold IsProjective in isprojX.
```

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+(* The following two lemmas make the proof of equiv_isprojective_choice below easier to follow. *)
+Lemma equiv_contr_hprop `{Funext} (A : Type) `{IsHProp A}
+  : Contr A <~> A.
+Proof.
+  rapply equiv_iff_hprop.
+  - apply center.
+  - rapply contr_inhabited_hprop.
+Defined.
+
+Lemma equiv_isconnmap_merely `{Funext} {X : Type} (P : X -> Type)
```
Alizter, line 48: Maybe a comment here saying "If the first projection map from a sigma type of a family is split surjective then all sections of the family are merely inhabited." or something to that effect.

theories/Spaces/Projective.v
```diff
@@ -0,0 +1,142 @@
...
+Proof.
+  apply (equiv_isprojective_choice (Fin n))^-1.
+  rapply finite_choice.
+Defined.
+
+Proposition isprojective_unit `{Funext} : IsProjective Unit.
+Proof.
+  apply (equiv_isprojective_choice Unit)^-1.
+  intros P S.
+  specialize S with tt.
```
Alizter, line 91: I wasn't aware that this was a thing. I guess it just destructs S right?

jarlg: Here `S` has type `forall x : Unit, merely (P x)` and specializing at `tt` means evaluating at `tt`, giving `S : merely (P tt)`. I see specializing more as function application (or variable binding) than destructing.

theories/Spaces/Projective.v
jarlg: No, also the LHS (`IsProjective X`) needs Funext to be an hprop. But then that's the sole need for Funext here.

theories/Spaces/Projective.v
Alizter: Oh I see now, I was confused. The syntax that I am aware of is `specialize (S tt).``

contrib/HoTTBookExercises.v
```diff
@@ -1619,6 +1619,7 @@ End Book_7_1.
 
 (* ================================================== ex:acnm-surjset *)
 (** Exercise 7.9 *)
+(* Solved for AC_(oo,-1) by projective_cover_AC in Spaces/Projective.v *)
```
mikeshulman, line 1622: I think this should be a definition like the other sections in this file.  It's okay that it doesn't solve the full exercise, you can name with an appropriate suffix to make this clear.

theories/HSet.v
mikeshulman: I actually would prefer to get rid of `Sect` entirely.  I can never remember what order the arguments of `Sect` come in, whereas with `r o s == idmap` it's clear that `s` is the section and `r` the retraction.

theories/Spaces/Projective.v
mikeshulman: Agreed.

theories/Spaces/Projective.v
mikeshulman: What we put at the end of a file is a newline character, not an empty blank line.

mikeshulman:
Regarding organization, I agree with @Alizter: `Types` is, generally speaking, for the type *formers*, while `Spaces` is for *particular* types.  This file proves general theorems, so it doesn't belong in either.

theories/Spaces/Projective.v
jdchristensen: In that case, maybe we should weaken the claim to `<->` instead of `<~>`?  The same might have to be done one or more places below.  It would then be handy to prove a lemma that `<->` is transitive (or is that already somewhere?)

theories/Spaces/Projective.v
jdchristensen: I agree, so I'll mark this as resolved.

theories/Spaces/Projective.v
mikeshulman: There could be both a `<->` result without funext and a `<~>` version with funext, if the latter seems important enough to spare uses the trouble of applying `equiv_iff_hprop` themselves.

mikeshulman: We have `iff_compose` in `Overture`.

theories/Spaces/Projective.v
jdchristensen: "A family of types is pointwise merely inhabited if and only if the corresponding fibration is surjective."  And I think we should reverse the two sides of the equivalence, having the simply claim first.  This also matches how it is used later.

mikeshulman: This lemma doesn't involve projectivity, so I would put it elsewhere.  `Truncations/Core`?

theories/HSet.v
Alizter: I agree, let's open an issue about that.

jdchristensen: Yes, I can never remember the order either.

theories/Spaces/Projective.v
jdchristensen: `ObjectClassifier.v` is where the bijection is discussed, but that file is a bit off the beaten track, so somewhere in Truncations would also make sense.  But this lemma is really mixing two things.  First, for *any* property `Q`, `Q` holds for the types in the family iff it holds for the fibers of the associated fibration, since we have equivalences.  Second, there is a minor translation between `merely` and `Tr (-1)`, which `equiv_contr_hprop` handles.  So the result is a bit of an awkward one as currently stated, but maybe it's still useful in this form?

mikeshulman: Well, you found it useful, right?  (-:

The fact about an arbitrary property Q uses univalence, whereas particular properties are equivalence-invariant without univalence, so we wouldn't want to be deriving the specific results from the general one in practice.

jdchristensen: Good point about univalence.  Still, one could consider factoring this into a result about `Tr n` for any `n`, and then a separate result about `merely` being equivalent to `Tr (-1)` being inhabited (which might already exist somewhere?)

mikeshulman: It's related to `Truncations.Core.Build_IsSurjection`.

Actually I think the general thing to prove would be about any `Subuniverse`, which is basically a property of types that's closed under equivalence.  In `ReflectiveSubuniverse` we have `mapinO_pr1` which is one direction of this lemma in that generality; it assumes the subuniverse is reflective for historical reasons but doesn't need to.

theories/Spaces/Projective.v
jarlg: Yep, you're right.

theories/Spaces/Projective.v
jarlg: I'll name the `<->` result (which doesn't assume Funext) `iff_isprojective_surjections_split` and then deduce the equiv version using Funext. This usage of "iff" doesn't seem to me like a misnomer since we have e.g. `iff_functor_forall` in `Types/Forall.v`, but let me know if I'm wrong.

theories/HSet.v
jarlg: Sounds like this usage is fine and the relevant issue has been opened, so resolving.

jarlg:
The placement in Spaces/ was due to lack of a better idea. Since both @Alizter and @mikeshulman agree, I'll move Projective.v to theories/ and refactor results into better files.

contrib/HoTTBookExercises.v
jarlg: Alright, I did this in the last commit just added.

theories/Spaces/Projective.v
jarlg: I think commit cae55fa contains the lemma you're alluding to @mikeshulman, specifically `equiv_forall_inO_mapinO_pr1`. To apply the lemma in this situation requires viewing `O`-truncated types as a subuniverse for a reflective subuniverse `O`, so that statement was added too. Since it has proven useful here, the exact statement needed was added as `equiv_merely_issurjection` to `Truncations.Core` near `Build_IsSurjection`.

contrib/HoTTBookExercises.v
```diff
@@ -1620,7 +1620,16 @@ End Book_7_1.
 (* ================================================== ex:acnm-surjset *)
 (** Exercise 7.9 *)
 
+Section Book_7_9.
+  (** Solution for the case (oo,-1). *)
 
+  Axiom AC_oo_neg1 : forall X : hSet, Choice X.
```
mikeshulman, line 1626: We shouldn't assume Axioms in this file.  It's better to state the definition using this "axiom" as a hypothesis.

mikeshulman:
This is improved, thanks!  I would still probably prefer to see `equiv_forall_inO_mapinO_pr1` and `equiv_merely_issurjection` stated as iff's without funext first, and only later the equiv version derived with funext (if it's really needed -- is it?)

contrib/HoTTBookExercises.v
jarlg: Okay, that's an easy fix. Is there a concrete reason why, or is it just bad style? Not arguing, just wanting to understand.

I can see that if another exercice want to use this one, the Axiom might be problematic. Is that it?

jdchristensen: I think you just want to make it `Context`.

jdchristensen:
And expanding on Mike's comment, if those are changed to iff, then other consequences about projectivity could also be stated as iff (with equiv versions assuming Funext?)

contrib/HoTTBookExercises.v
jarlg: I'll make it `Context '{Univalence} {AC_oo_neg1 : forall X : hSet, Choice X}`, so univalence stays first.

mikeshulman: Right, an `Axiom` is a permanent assertion, as opposed to a `Context` which is a local hypothesis.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -51,14 +51,6 @@ Definition inO_equiv_inO' {O : Subuniverse}
   : In O U
   := inO_equiv_inO T f.
 
-Definition inO_equiv_inO'' `{Funext} (O : Subuniverse) {T U : Type} (f : T <~> U)
-  : In O T <~> In O U.
-Proof.
-  apply equiv_iff_hprop.
-  - intro T_inO; exact (inO_equiv_inO' T f).
-  - intro U_inO; exact (inO_equiv_inO' U f^-1%equiv).
-Defined.
```
jarlg: `inO_equiv_inO''` was added in a previous commit, but it was no longer needed after this refactoring.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -89,6 +81,13 @@ Global Existing Instance inO_hfiber_ino_map.
 Section Subuniverse.
   Context (O : Subuniverse).
 
+  (** Being a local map is an hprop *)
+  Global Instance ishprop_mapinO `{Funext} {A B : Type} (f : A -> B)
+  : IsHProp (MapIn O f).
+  Proof.
+    apply trunc_forall.
+  Defined.
```
jarlg, line 89: This was simply moved up to deduce `equiv_forall_inO_mapinO_pr1` from the iff version below.

jarlg:
Both `equiv_forall_inO_mapinO_pr1` and `equiv_merely_isprojective` have now been deduced from their iff versions. This had the the nice consequence of `Fin n` being projective without `Funext`. While not strictly needed, I've opted to keep the equiv versions of various results for convenience.

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -70,6 +70,63 @@ Definition equiv_path_TypeO@{i j} {fs : Funext} O (T T' : Type_ O)
...
+
+  (** Anything homotopic to a local map is local. *)
+  Definition mapinO_homotopic {A B : Type} (f : A -> B) {g : A -> B}
+             (p : f == g) `{MapIn O _ _ f}
+  : MapIn O g.
+  Proof.
+    intros b.
+    refine (inO_equiv_inO (hfiber f b)
```
Alizter, line 97: It's not a necessary change but since there are no holes in this refine, its probably better to just write `exact` to emphasise that you have given an exact term.

Alizter (approved):
LGTM, Thanks for the additions!

theories/Modalities/ReflectiveSubuniverse.v
jarlg: This result (along with `mapinO_pr1` below, and `ishprop_mapinO` above) was just moved up, so I didn't consider changing them. If there are more things to change I'll make to sure to include this.

jdchristensen, line 97:
```suggestion
    exact (inO_equiv_inO (hfiber f b)
```
I agree with Ali, and you can commit this change right in github.

jarlg: Done!

jarlg: Alright, done!

theories/Projective.v
```diff
@@ -0,0 +1,130 @@
...
+          IsSurjection p -> merely (exists s : X -> Y, p o s == idmap)).
+Proof.
+  exact (equiv_iff_hprop_uncurried (iff_isprojective_surjections_split X)).
+Defined.
+
+
+(** ** Projectivity and choice *)
+
+Definition Choice (X : Type) : Type := forall P : X -> Type,
+    (forall x, merely (P x)) -> merely (forall x, P x).
```
mikeshulman, line 46: Can we find another name for this property?  Maybe `IsProjective'` or something (since it's equivalent to `IsProjective`)?  In topos theory the word "choice object" is sometimes used with a [dual meaning](https://ncatlab.org/nlab/show/choice+object), which is not equivalent to projectivity.

theories/Projective.v
```diff
@@ -0,0 +1,130 @@
...
+  Proposition projective_cover_AC `{Univalence} (A : Type)
+    : merely (exists X:hSet, exists p : X -> A, IsSurjection p).
+  Proof.
+    pose (X := BuildhSet (Tr 0 A)).
+    pose proof ((equiv_isprojective_choice X)^-1 (AC X)) as P.
+    pose proof (P A X idmap tr _) as F; clear P.
+    strip_truncations.
+    destruct F as [f p].
+    apply tr; refine (X; (f; _)).
+    unfold IsConnMap, IsConnected.
```
mikeshulman, line 102: This and the following two lines are essentially `apply BuildIsSujection`.

mikeshulman (approved):

theories/Projective.v
jdchristensen: `IsProjective'` seems reasonable to me, along with a comment saying something like "In topos theory, an object X is said to be projective if the axiom of choice holds when making choices indexed by X."

theories/Projective.v
jdchristensen: You still need the intro, and the unfolds are just there to help the reader, so it's a choice between `apply BuildIsSurjection. intro a` and `intro a. rapply contr_inhabited_hprop`.  I prefer the second once (i.e. leaving the proof like it is), as it seems easier to follow.

jdchristensen (approved):

theories/Projective.v
jarlg: I think the intent of `BuildIsSurjection` is clearer, and it's what I would reach for in the future upon seeing the goal `IsConnMap (Tr (-1)) f` (which I should read as `IsSurjection f`). The unfolds really only help the reader with the current proof in my opinion, and without them `BuildIsSurjection` can even go into the refine above making it `refine (X; (f; BuildIsSurjection f _)).` which I think is readable and saves 3 lines. I'll commit that tonight unless there's further input.

theories/Projective.v
jarlg: Yep, will change to `IsProjective'`.

theories/Projective.v
jdchristensen: Sure, sounds fine to me.

jarlg: Great!

PR #1398, opened 2020-11-09
bump coq-dpdgraph

Alizter:


Alizter:
I'm going to break convention and merge this when this is green. This is only updating a submodule (coq-dpdgraph) so that our CI works with the latest version of coq. Hopefully this will make #1327 green which is required for the main coq ci.

Alizter:
Or forget this I will push a commit to #1327 instead.

PR #1400, opened 2020-11-29
With Coq PR #13513, "intros []" uses "destruct" rather than "case" and becomes "smarter".

herbelin:
As a consequence, a clear on a residual dependent argument of an inductive type becomes useless. We leave a try so as to be backwards-compatible. You may prefer a clean non-compatible script though.

cf coq/coq#13513


Alizter:
Can you rebase this? We had some issues with CI that should be fixed now.

Alizter (approved):

PR #1402, opened 2020-12-01
Update version of setup-python

### LINKED ISSUE #1401: CI failing due to unsecure commands
It seems github had some issues with environment variables: https://thenewstack.io/github-actions-design-flaw-leaves-security-hole-for-remote-code-execution/

See githubs blog post on the issue: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

As a result, it seems we need the `ACTIONS_ALLOW_UNSECURE_COMMANDS` to be set to true for github actions to succeed. I think we might need to modify our current script to not use unsecure commands where ever they are.

The current build succeeded 22 days ago but failed today.

Alizter:
This fixes #1401

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
     steps:
     # First we set up Python 2.7
     - name: Set up Python 2.7
-      uses: actions/setup-python@v1.1.1
+      uses: actions/setup-python@v2
       with:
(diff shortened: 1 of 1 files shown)


Alizter:
Since this is a small fix and only effects the CI I will merge it without review. Feel free to complain if you feel otherwise.

Issue #1387, opened 2020-08-26
Subgroups vs subgroups upto isomorphism


Alizter:
When designing the group theory library, I defined subgroups to be subgroups upto isomorphism. I thought that this would better fit the informal usage i.e. Z2 is a subgroups of S4 etc.

I don't think this is correct anymore however. It generalizes our definition of normal subgroup slightly. I think this may be a problem as evidenced here: https://mathoverflow.net/a/370104/54401

In that answer they construct a group whose subgroups are "normal upto isomorphism" but not normal. I think this is plenty evidence that "subgroup upto isomorphism" is strictly a weaker notion than subgroup.

To fix this we need to replace the definition of subgroup with one that uses a predicate. As a result the type of subgroups of a group G in a universe U will live in a strictly larger universe, but the actual group we get from this will still be in U.

mikeshulman:
No, that question is about subgroups that are *abstractly* isomorphic to a normal one (i.e. not respecting the inclusion).  Our subgroups are *equipped with* an inclusion, so the resulting notion of "normal" is the same as the usual one.

Alizter:
Oh I see where I was confused now. Being a normal subgroup is really a property of the inclusion map. I was confused because I thought we could always compose the subgroup with the isomorphism and get another normal subgroup, but that is no longer the same subgroup.

Alright, I will close this issue.

mikeshulman:
Which isn't to say that we shouldn't switch.  As I think I said, I would have been inclined to do it the other way from the start.  The fact that we *can* talk about subsets directly, rather than injective functions, is one of the advantages of type theory over ETCS.

Also, am I confused or do *both* definitions of subgroup end up living in a higher universe (though, under propositional resizing, being equivalent to one in the same universe)?

Alizter:
I've thought about this a bit more over the passing months, and I am more and more opting to redesign and define subgroups as a special kind of predicate on groups.

Currently we talk about subgroups like they are groups, but really we are talking of properties of their inclusion map. We can remove a lot confusion just by making it a property of a predicate on groups.

@jarlg I wonder what your opinion on this is. since you said you will be using the groups library. I personally think, we should seriously consider redesigning it before too much stuff depends on it.

jdchristensen:
[...]

PR #1403, opened 2020-12-01
Characterise paths of subgroups, relate to trivial kernels and exactness

jarlg:
The first commit characterises paths of subgroups as isomorphisms respecting the inclusions, and proves that the kernel of a group homomorphism is trivial if and only if it's an embedding. There was some related discussion at #1387, and I've opted to stick with the existing formalization of subgroups instead of developing a version using predicates. If there's an argument for switching, I'd like to know.

The second commit relates the above to (pure) exactness of a complex `0 -> A -> B` of groups in `equiv_grp_isexact_isembedding`.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -417,6 +417,14 @@ Proof.
   apply ap, eisretr.
 Defined.
 
+Definition equiv_ap_inv `(f : A -> B) `{IsEquiv A B f} (x y : B)
+  : (f^-1 x = f^-1 y) <~> (x = y)
+  := (@equiv_ap B A f^-1 _ x y)^-1%equiv.
+
+Definition equiv_ap_inv' `(f : A <~> B) (x y : B)
+  : (f^-1 x = f^-1 y) <~> (x = y)
+  := (equiv_ap' f^-1%equiv x y)^-1%equiv.
+
 (** If [g \o f] and [f] are equivalences, so is [g].  This is not an Instance because it would require Coq to guess [f]. *)
 Definition cancelR_isequiv {A B C} (f : A -> B) {g : B -> C}
   `{IsEquiv A B f} `{IsEquiv A C (g o f)}
--- a/theories/Homotopy/ExactSequence.v
+++ b/theories/Homotopy/ExactSequence.v
@@ -106,6 +106,23 @@ Proof.
   apply postcompose_pconst.
 Defined.
 
+(** Any pointed map induces a trivial complex. *)
+Definition iscomplex_trivial {X Y : pType} (f : X ->* Y)
+  : IsComplex (@pconst pUnit X) f.
+Proof.
+  srapply Build_pHomotopy.
+  - intro x; cbn.
+    exact (point_eq f).
+  - cbn; symmetry.
+    exact (concat_p1 _ @ concat_1p _).
+Defined.
+
+Local Existing Instance ishprop_phomotopy_hset.
+
+(** If Y is a set, then IsComplex is an HProp. *)
+Instance ishprop_iscomplex_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (i : F ->* X) (f : X ->* Y)
+  : IsHProp (IsComplex i f) := {}.
+
 
 (** ** Very short exact sequences and fiber sequences *)
 
@@ -117,6 +134,17 @@ Class IsExact (n : Modality) {F X Y : pType} (i : F ->* X) (f : X ->* Y) :=
 }.
 
 Global Existing Instance conn_map_isexact.
+
+Definition issig_isexact (n : Modality) {F X Y : pType} (i : F ->* X) (f : X ->* Y)
+  : _ <~> IsExact n i f := ltac:(issig).
+
+(** If Y is a set, then IsExact is an HProp. *)
+Instance ishprop_isexact_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (n : Modality) (i : F ->* X) (f : X ->* Y)
+  : IsHProp (IsExact n i f).
+Proof.
+  rapply (transport IsHProp (x := { cx : IsComplex i f & IsConnMap n (cxfib cx) })).
+  apply path_universe_uncurried; issig.
+Defined.
 
 (** Passage across homotopies preserves exactness. *)
 Definition isexact_homotopic_i n  {F X Y : pType}
--- a/theories/Homotopy/ExactSequence.v
+++ b/theories/Homotopy/ExactSequence.v
@@ -120,7 +120,7 @@ Defined.
 Local Existing Instance ishprop_phomotopy_hset.
 
 (** If Y is a set, then IsComplex is an HProp. *)
-Instance ishprop_iscomplex_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (i : F ->* X) (f : X ->* Y)
+Global Instance ishprop_iscomplex_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (i : F ->* X) (f : X ->* Y)
   : IsHProp (IsComplex i f) := {}.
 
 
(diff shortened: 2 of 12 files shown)


theories/Basics/Equivalences.v
```diff
@@ -417,6 +417,14 @@ Proof.
   apply ap, eisretr.
 Defined.
 
+Definition equiv_ap_inv `(f : A -> B) `{IsEquiv A B f} (x y : B)
```
Alizter, line 420:
```suggestion
Definition equiv_ap_inv (f : A -> B) `{IsEquiv A B f} (x y : B)
```

theories/Basics/Equivalences.v
```diff
@@ -417,6 +417,14 @@ Proof.
...
+  : (f^-1 x = f^-1 y) <~> (x = y)
+  := (@equiv_ap B A f^-1 _ x y)^-1%equiv.
+
+Definition equiv_ap_inv' `(f : A <~> B) (x y : B)
```
Alizter, line 424:
```suggestion
Definition equiv_ap_inv' (f : A <~> B) (x y : B)
```

theories/Homotopy/ExactSequence.v
```diff
@@ -118,6 +135,17 @@ Class IsExact (n : Modality) {F X Y : pType} (i : F ->* X) (f : X ->* Y) :=
 
 Global Existing Instance conn_map_isexact.
 
+Definition issig_isexact (n : Modality) {F X Y : pType} (i : F ->* X) (f : X ->* Y)
+  : _ <~> IsExact n i f := ltac:(issig).
+
+(** If Y is a set, then IsExact is an HProp. *)
+Instance ishprop_isexact_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (n : Modality) (i : F ->* X) (f : X ->* Y)
```
Alizter, line 142:
```suggestion
Global Instance ishprop_isexact_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (n : Modality) (i : F ->* X) (f : X ->* Y)
```

theories/Modalities/Identity.v
```diff
@@ -27,3 +27,10 @@ Proof.
     + intros _ [].
     + intros; exact tt.
 Defined.
+
+(** A purely connected map is an equivalence. *)
+Instance isequiv_purely_conn `{Funext} {A B : Type} (f : A -> B) {conn : IsConnMap purely f}
```
Alizter, line 32:
```suggestion
Global Instance isequiv_purely_conn `{Funext} {A B : Type} (f : A -> B) {conn : IsConnMap purely f}
```

theories/Pointed/pHomotopy.v
```diff
@@ -72,3 +72,11 @@ Proof.
 Defined.
 
 Notation "p @@* q" := (phomotopy_hcompose p q).
+
+(** Pointed homotopies in a set form an HProp. *)
+Instance ishprop_phomotopy_hset `{Univalence} {X Y : pType} `{IsHSet Y} (f g : X ->* Y)
```
Alizter, line 77:
```suggestion
Global Instance ishprop_phomotopy_hset `{Univalence} {X Y : pType} `{IsHSet Y} (f g : X ->* Y)
```

theories/Homotopy/ExactSequence.v
```diff
@@ -106,6 +106,23 @@ Proof.
...
+  - intro x; cbn.
+    exact (point_eq f).
+  - cbn; symmetry.
+    exact (concat_p1 _ @ concat_1p _).
+Defined.
+
+Local Existing Instance ishprop_phomotopy_hset.
+
+(** If Y is a set, then IsComplex is an HProp. *)
+Instance ishprop_iscomplex_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (i : F ->* X) (f : X ->* Y)
```
Alizter, line 123:
```suggestion
Global Instance ishprop_iscomplex_hset `{Univalence} {F X Y : pType} `{IsHSet Y} (i : F ->* X) (f : X ->* Y)
```

Alizter (commented):
Just some minor style suggestions. Writing `Global` before an instance can make it more clear that it really is a global instance. The behaviour of just writing `Instance` can change dramatically, whether or not it is contained in a section.

Alizter:
The CI isn't anything to worry about. It seems to have duplicated and failed some jobs.

theories/Basics/Equivalences.v
jarlg: I had the tick there because `equiv_ap` (and `equiv_ap'`) both have one. It's needed here as well, since neither `A` nor `B` are explicitly defined but they are generalizable variables. (See [the docs](https://coq.inria.fr/refman/language/extensions/implicit-arguments.html#implicit-generalization) .)

theories/Homotopy/ExactSequence.v
jarlg: Good catch. I left them undeclared since I wasn't sure whether to make them global or not, but I think global is reasonable in this case?

theories/Modalities/Identity.v
```diff
@@ -1,5 +1,5 @@
 (* -*- mode: coq; mode: visual-line -*- *)
-Require Import HoTT.Basics HoTT.Types.
+Require Import HoTT.Basics HoTT.Types HoTT.Truncations.
```
Alizter, line 2: I'm not sure how I feel about the Identity modality relying on Truncations which in turn relies on other things in the modalities library.

Alizter (approved):
LGTM

theories/Modalities/Identity.v
jarlg: I agree that it has potential to get messy, and the main reason this works is that `Modalities/` is quite modular in the sense that we usually only import specific files from there. In order to get this working i had to remove the (unused) import of `Modalities.Identity` in `Truncations.Core`.

I guess an alternative solution would be to have a file `Truncations.Purely` which would import `Modalities.Identity` and prove `isequiv_purely_conn`. That might be more appropriate? Let me know and I'll make that change.

jdchristensen: Actually, this import isn't needed at all in the file.  If you remove it, you can just add `exact conn` as the last line of the last proof in the file.

theories/Modalities/Identity.v
```diff
@@ -27,3 +27,10 @@ Proof.
     + intros _ [].
     + intros; exact tt.
 Defined.
+
+(** A purely connected map is an equivalence. *)
+Instance isequiv_purely_conn `{Funext} {A B : Type} (f : A -> B) {conn : IsConnMap purely f}
+  : IsEquiv f.
+Proof.
+  rapply equiv_contr_map_isequiv.
```
jdchristensen, line 35: You can add `exact conn.` here to make the proof go through.  Also, you can use `isequiv_contr_map` which more directly gives what you need.  One question:  there is a `Hint Immediate` for `isequiv_contr_map`.  If it was instead a `Global Instance`, would `isequiv_purely_conn` not be needed at all?
```coq
  rapply isequiv_contr_map.
  exact conn.
```

jdchristensen (approved):

theories/Modalities/Identity.v
jdchristensen: In fact, I think `isequiv_contr_map` and the new `isequiv_purely_conn` are exactly the same fact.  The types of the hypotheses `IsConnMap purely f` and `IsTruncMap (-2) A B f` are definitionally equal.

jdchristensen: Coq accepts:
```coq
Definition isequiv_purely_conn' `{Funext} {A B : Type} (f : A -> B) {conn : IsConnMap purely f}
  : IsEquiv f := isequiv_contr_map f.
```

jarlg: Thanks, this was clarifying. The comment on line 33 in `Types/Equiv.v` says we can't have both `contr_map_isequiv` and `isequiv_contr_map` (which are inverse of each other) as instances, as I'm guessing that would create an infinite loop during instance search. It seems the choice was made to make one direction an instance, and the direction we're currently considering only a hint.

I'm opting to inline the proof (which you gave just above), and it seems propagating this through some of my more recent changes is improving typeclass resolution which was starting to stall.

jdchristensen: Right, I guess you inadvertently created a loop in the typeclass resolution, so just inlining this where needed is better.  LGTM.

Issue #1394, opened 2020-09-03
Get rid of `Sect`


Alizter:
Reasons for keeping:
 * ???

Reasons for removing:
 * Requires unfold in various proofs
 * Hard to remember which way the arguments go
 * Scares away new users

---

>I actually would prefer to get rid of `Sect` entirely.  I can never remember what order the arguments of `Sect` come in, whereas with `r o s == idmap` it's clear that `s` is the section and `r` the retraction.

_Originally posted by @mikeshulman in https://github.com/HoTT/HoTT/pull/1393#discussion_r483120592_

PR #1406, opened 2020-12-29
Remove Sect

Alizter:
As discussed in #1394.

In summary Sect is hard to read/remember how to use.

I have also removed pSect since it wasn't used so often.

Also did some minor cleanup in the exercises.


mikeshulman (approved):

jdchristensen (approved):
Nice, I prefer things this way.

Issue #479, opened 2014-09-09
A strange error (or lack thereof)


mikeshulman:
Commit d238e03e3eb is on top of PR #470.  It causes compilation to fail in `Pushout.v` with the error

```
File "./theories/hit/Pushout.v", line 77, characters 2-51:
Error:
Cannot infer an internal placeholder of type "TruncType.TruncType minus_two" in environment:

A : hSet
B : hSet
f : A -> B
```

I have two questions:
1. Why is this?  Why do we get an error in `Pushout` by changing `ReflectiveSubuniverse` (which is not even imported directly into `Pushout`) to import (but not use anything from) `Functorish`?
2. Why did the offending line `Definition setcone := Truncation 0 (pushout f !).` ever work in the first place?  As far as I can tell, `!` is a notation for the unique map from any type to _any_ contractible type, so how did Coq guess in this case that the contractible type should be `Unit`?

PR #1407, opened 2021-01-02
Using Hint Cut to avoid infinite loops with bidirectional Instances

jarlg:
Using `Hint Cut` it's possible to avoid infinite loops induced by having for example both directions of an equivalence be instances. In the library, the usual solution is to make one direction a `Hint Immediate` (or a `Hint Extern` that either fails or succeeds). That may be more efficient, but it's less expressive. This is a draft to discuss the feasibility of using `Hint Cut` like this, and there are some comments inline explaining the changes.

Here `Hint Cut` is just tested out for the equvalences `inO_tr_istrunc` / `istrunc_inO_tr` and `mapinO_tr_istruncmap` / `istruncmap_mapinO_tr`. The main changes are to `Truncations.Core`, and the rest are minor changes either required or to save a few seconds. The new instances open new "proof branches" which aren't always productive, but I think it's possible to clean up. The killer app is `isembedding_pr1_hset`, which Coq is now able to solve by itself.

With these changes, my build time is (on `hint-cut`):
```sh
real    4m57.762s
user    11m11.561s
sys     1m21.895s
```
versus on `master`
```sh
real    3m57.741s
user    9m23.791s
sys     1m21.934s
```

I haven't looked into profiling in Coq, but I think with some cleaning up the gap could be reduced substantially.

The benefit of the proposal should be clear. A possible concern is that these `Hint Cut`s disallow *any* repeated application of a Hint (or Instance), even with different hypotheses. In theory this rules out some valid proofs, but I haven't observed this in practice (and of course, those proofs can still be carried out manually).

### DIFF
--- a/theories/Algebra/Groups/Subgroup.v
+++ b/theories/Algebra/Groups/Subgroup.v
@@ -113,7 +113,8 @@ Defined.
 Corollary ishprop_path_subgroup `{U : Univalence} {G : Group} {H K : Subgroup G}
   : IsHProp (H = K).
 Proof.
-  rapply (equiv_transport IsHProp _ _).
+  (** jarlg: Without "n" here, it unifies (IsHProp ?x) with (IsHprop Unit). Can that be avoided? *)
+  nrapply (transport IsHProp).
   - apply equiv_path_universe.
     exact (equiv_path_subgroup H K).
   - apply equiv_hprop_allpath.
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -712,8 +712,6 @@ Register tt as core.True.I.
 (** A space is pointed if that space has a point. *)
 Class IsPointed (A : Type) := point : A.
 
-Typeclasses Transparent IsPointed.
-
 Arguments point A {_}.
 
 Record pType :=
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -712,6 +712,8 @@ Register tt as core.True.I.
 (** A space is pointed if that space has a point. *)
 Class IsPointed (A : Type) := point : A.
 
+Typeclasses Transparent IsPointed.
+
 Arguments point A {_}.
 
 Record pType :=
--- a/theories/Pointed/Loops.v
+++ b/theories/Pointed/Loops.v
@@ -180,7 +180,8 @@ Proof.
   intros p; eapply isconnected_equiv'.
   - refine (hfiber_loops_functor f p oE _).
     symmetry; apply hfiber_ap.
-  - exact _.
+    (** jarlg: "exact _." works here but is a bit slow. *)
+  - apply isconnected_paths.
 Defined.
 
 Definition isconnected_iterated_loops_functor `{Univalence}
--- a/theories/Truncations/Connectedness.v
+++ b/theories/Truncations/Connectedness.v
@@ -128,7 +128,7 @@ Definition merely_path_is0connected `{Univalence}
 : merely (x = y).
 Proof.
   refine ((equiv_path_Tr x y)^-1 (path_contr (tr x) (tr y))).
-  exact H0.
+  assumption.
 Defined.
 
 Definition is0connected_merely_allpath `{Univalence}
--- a/theories/Truncations/Connectedness.v
+++ b/theories/Truncations/Connectedness.v
@@ -128,6 +128,7 @@ Definition merely_path_is0connected `{Univalence}
 : merely (x = y).
 Proof.
   refine ((equiv_path_Tr x y)^-1 (path_contr (tr x) (tr y))).
+  exact H0.
 Defined.
 
 Definition is0connected_merely_allpath `{Univalence}
--- a/theories/Truncations/Core.v
+++ b/theories/Truncations/Core.v
@@ -133,9 +133,11 @@ Proof.
   assumption.
 Defined.
 
-(** To avoid infinite loops we cut proofs which apply any of the above Instances twice. Note that the cuts match arbitrary applications of the Instances, so this disallows proofs that would non-circularly use the Instances with different hypotheses. In practice this rarely happens however. *)
-Hint Cut [( _* ) inO_tr_istrunc ( _* ) inO_tr_istrunc] : typeclass_instances.
+(** To avoid infinite loops we cut proofs which apply second Instance twice. Note that the cuts match arbitrary applications of the Instances, so this disallows proofs that would non-circularly use this Instance with different hypotheses. In practice this rarely happens. It turns out to slightly slow down compilation of the library to add the same cut rule for inO_tr_istrunc. *)
 Hint Cut [( _* ) istrunc_inO_tr ( _* ) istrunc_inO_tr] : typeclass_instances.
+(** The next two cuts are purely for efficiency, eliminating immediate round-trips between the two inverse Instances. *)
+Hint Cut [( _* ) inO_tr_istrunc istrunc_inO_tr] : typeclass_instances.
+Hint Cut [( _* ) istrunc_inO_tr inO_tr_istrunc] : typeclass_instances.
 
 (** We do the same for [IsTruncMap n] and [MapIn (Tr n)]. *)
 Global Instance mapinO_tr_istruncmap {n : trunc_index} {A B : Type}
@@ -153,8 +155,9 @@ Proof.
 Defined.
 
 (** See comment for the 'Hint Cut's above. *)
-Hint Cut [( _* ) mapinO_tr_istruncmap ( _* ) mapinO_tr_istruncmap] : typeclass_instances.
 Hint Cut [( _* ) istruncmap_mapinO_tr ( _* ) istruncmap_mapinO_tr] : typeclass_instances.
+Hint Cut [( _* ) mapinO_tr_istruncmap istruncmap_mapinO_tr] : typeclass_instances.
+Hint Cut [( _* ) istruncmap_mapinO_tr mapinO_tr_istruncmap] : typeclass_instances.
 
 (** jarlg: An example of what we gain by having the Instance istruncmap_mapinO_tr. *)
 Lemma isembedding_pr1_hset `{Funext} {X : Type} {Y : hSet} (f : X -> Y) (y : Y)
--- a/theories/Truncations/Core.v
+++ b/theories/Truncations/Core.v
@@ -127,20 +127,15 @@ Proof.
   assumption.
 Defined.
 
-(** Having both of these as [Instance]s would cause infinite loops. *)
-Definition istrunc_inO_tr {n : trunc_index} (A : Type) `{In (Tr n) A}
-  : IsTrunc n A.
+Global Instance istrunc_inO_tr {n : trunc_index} (A : Type) `{In (Tr n) A}
+  : IsTrunc n A | 1000.
 Proof.
   assumption.
 Defined.
 
-(** Instead, we make the latter an immediate instance, but with high cost (i.e. low priority) so that it doesn't override the ordinary lemmas about truncation.  Unfortunately, [Hint Immediate] doesn't allow specifying a cost, so we use [Hint Extern] instead. *)
-(** Hint Immediate istrunc_inO_tr : typeclass_instances. *)
-(** See https://github.com/coq/coq/issues/11697 *)
-Hint Extern 1000 (IsTrunc _ _) => simple apply istrunc_inO_tr; solve [ trivial ] : typeclass_instances.
-(** This doesn't seem to be quite the same as [Hint Immediate] with a different cost either, though; see the comment in the proof of [Trunc_min] below.  *)
-
-(** Unfortunately, this isn't perfect; Coq still can't always find [In n] hypotheses in the context when it wants [IsTrunc].  You can always apply [istrunc_inO_tr] explicitly, but sometimes it also works to just [pose] it into the context. *)
+(** To avoid infinite loops we cut proofs which apply any of the above Instances twice. Note that the cuts match arbitrary applications of the Instances, so this disallows proofs that would non-circularly use the Instances with different hypotheses. In practice this rarely happens however. *)
+Hint Cut [( _* ) inO_tr_istrunc ( _* ) inO_tr_istrunc] : typeclass_instances.
+Hint Cut [( _* ) istrunc_inO_tr ( _* ) istrunc_inO_tr] : typeclass_instances.
 
 (** We do the same for [IsTruncMap n] and [MapIn (Tr n)]. *)
 Global Instance mapinO_tr_istruncmap {n : trunc_index} {A B : Type}
@@ -150,14 +145,23 @@ Proof.
   assumption.
 Defined.
 
-Definition istruncmap_mapinO_tr {n : trunc_index} {A B : Type}
+Global Instance istruncmap_mapinO_tr {n : trunc_index} {A B : Type}
   (f : A -> B) `{MapIn (Tr n) _ _ f}
-  : IsTruncMap n f.
+  : IsTruncMap n f | 1000.
 Proof.
   assumption.
 Defined.
 
-Hint Immediate istruncmap_mapinO_tr : typeclass_instances.
+(** See comment for the 'Hint Cut's above. *)
+Hint Cut [( _* ) mapinO_tr_istruncmap ( _* ) mapinO_tr_istruncmap] : typeclass_instances.
+Hint Cut [( _* ) istruncmap_mapinO_tr ( _* ) istruncmap_mapinO_tr] : typeclass_instances.
+
+(** jarlg: An example of what we gain by having the Instance istruncmap_mapinO_tr. *)
+Lemma isembedding_pr1_hset `{Funext} {X : Type} {Y : hSet} (f : X -> Y) (y : Y)
+  : IsEmbedding (pr1 : hfiber f y -> X).
+Proof.
+  exact _.
+Defined.
 
 (** ** A few special things about the (-1)-truncation. *)
 
@@ -256,9 +260,7 @@ Proof.
     symmetry.
     srapply equiv_tr.
     srapply trunc_leq.
-    3:{ (** Strangely, if [istrunc_inO_tr] were a [Hint Immediate], rather than our [Hint Extern], then typeclass inference would be able to find this all on its own, although the documentation for [Hint Immediate] suggests that it shouldn't because the following tactic doesn't solve it completely. *)
-        simple apply istrunc_inO_tr; trivial.
-        exact _. }
+    3: exact _.
     apply trunc_index_min_leq_left.
 Defined.
 
@@ -282,7 +284,7 @@ Global Instance O_eq_Tr (n : trunc_index)
 Proof.
   split; intros A A_inO.
   - intros x y; exact _.
-  - exact _.
+  - assumption.
 Defined.
 
 (** It follows that [Tr n <<< Tr n.+1].  However, it is easier to prove this directly than to go through separatedness. *)
--- a/theories/Algebra/Groups/Kernel.v
+++ b/theories/Algebra/Groups/Kernel.v
@@ -106,14 +106,14 @@ Local Existing Instance ishprop_path_subgroup.
 Proposition equiv_kernel_isembedding `{Univalence} {A B : Group} (f : A $-> B)
   : (grp_kernel f = trivial_subgroup) <~> IsEmbedding f.
 Proof.
-  srapply equiv_iff_hprop.
+  apply equiv_iff_hprop.
   - intros phi b.
     apply hprop_inhabited_contr; intro a.
     rapply (contr_equiv' _ (equiv_grp_hfiber _ _ a)^-1%equiv).
     rapply (transport Contr (x:=grp_trivial)).
     exact (ap (group_type o subgroup_group A) phi^).
   - intro isemb_f.
-    rapply equiv_path_subgroup.
+    apply equiv_path_subgroup.
     srefine (grp_iso_inverse _; _).
     + srapply Build_GroupIsomorphism.
       * exact grp_homo_const.
--- a/theories/HIT/FreeIntQuotient.v
+++ b/theories/HIT/FreeIntQuotient.v
@@ -74,7 +74,8 @@ Section FreeIntAction.
       unfold loop.
       exact (Coeq_rec_beta_cglue _ _ _ _).
     - intros xu yv.
-      refine (trunc_equiv' (n := -1) _ (equiv_path_sigma _ xu yv)).
+      (** jarlg: This one finishes but takes a while without "n". *)
+      nrefine (trunc_equiv' (n := -1) _ (equiv_path_sigma _ xu yv)).
       destruct xu as [x u], yv as [y v]; cbn.
       apply hprop_allpath.
       intros [p r] [q s].
--- a/theories/Spaces/Finite.v
+++ b/theories/Spaces/Finite.v
@@ -884,7 +884,6 @@ Proof.
   refine (decidable_equiv _ (hfiber_fibration x P)^-1 _).
   (* The try clause below is only needed for Coq <= 8.11 *)
   refine (detachable_image_finite pr1 x); try assumption.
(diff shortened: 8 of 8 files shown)


jdchristensen:
If this can be made to work without impacting the build time too much, it'll be a clear win.  And I suspect that that should be possible.  Take a look at etc/coq-scripts/timing/README.md and scripts in that directory for info about timing tools.  Examples:
```bash
# Compare build times to an earlier commit.  Defaults to previous commit.
# This does two full builds, so takes some time.
export PREV_COMMIT=<hash>
./etc/coq-scripts/timing/make-pretty-timed-only-diff-tip.sh
sort -r -k 12 time-of-build-both.log | head
```
To see the timing per line in a file, use:
```bash
make timing-html/HoTT.Modalities.ReflectiveSubuniverse.html 
```
for example, and then view the file in a web browser.  Coloured bars overlaying each line show how long that line took relative to the slowest line in that file (full width).

Within Coq, you can prefix a line with `Time` to see how long it takes.  And in Proof General, you can set Coq > Settings > Time Commands in order to time all commands.

theories/Truncations/Core.v
```diff
@@ -127,20 +127,15 @@ Proof.
...
-(** Hint Immediate istrunc_inO_tr : typeclass_instances. *)
-(** See https://github.com/coq/coq/issues/11697 *)
-Hint Extern 1000 (IsTrunc _ _) => simple apply istrunc_inO_tr; solve [ trivial ] : typeclass_instances.
-(** This doesn't seem to be quite the same as [Hint Immediate] with a different cost either, though; see the comment in the proof of [Trunc_min] below.  *)
-
-(** Unfortunately, this isn't perfect; Coq still can't always find [In n] hypotheses in the context when it wants [IsTrunc].  You can always apply [istrunc_inO_tr] explicitly, but sometimes it also works to just [pose] it into the context. *)
+(** To avoid infinite loops we cut proofs which apply any of the above Instances twice. Note that the cuts match arbitrary applications of the Instances, so this disallows proofs that would non-circularly use the Instances with different hypotheses. In practice this rarely happens however. *)
+Hint Cut [( _* ) inO_tr_istrunc ( _* ) inO_tr_istrunc] : typeclass_instances.
+Hint Cut [( _* ) istrunc_inO_tr ( _* ) istrunc_inO_tr] : typeclass_instances.
```
jdchristensen, line 138: I'm surprised by the form of these `Hint Cut`s (and missed this point when we discussed this yesterday).  I was expecting to cut proofs when you had done one of these and then the other, but instead you are blocking them when you do one of these twice.  Strangely, if you switch to what I expected (for both orders), the `exact _` a few lines below seems to loop.  Ah, as you said yesterday, this is because `inO_tr_istrunc` has other inverses as well (e.g. `inO_contr` when the truncation index is -2).  So these `Hint Cut`s will definitely rule out some proofs that could have been found with the old set-up...    What about only keep the second `Hint Cut` of each pair (disallowing the new Instance from being used twice)?  The library builds with this change, but I didn't check the timing.

jdchristensen:
Here are some bigger-picture thoughts.

`IsTrunc n A` and `In (Tr n) A` are *definitionally* equal, and the Instances we are considering are just using reflexivity to switch between them.  Is there a better solution?

* Can we have only one of these forms in the library so no translation is needed?  Or maybe have one of the forms except in some small portion of the library?
* Or would `Hint Transparent` be a better tool here than Instances, since we are dealing with definitionally equal things?
* Or maybe we could have a tactic that switches the goal *and the hypotheses* to one of the two forms, and we could arrange that typeclass inference runs this tactic?

To be honest, the layers involving `IsTrunc`, `IsTrunc_internal`, `In`, `In_internal`, `Contr` and `Contr_internal` are very confusing...

theories/Truncations/Core.v
jarlg: I think cuts like these are the way to go, as they clearly avoid looping. If you change it like you said, blocking the application of one after the other, then you also need to add a cut for `inO_contr` and I think that will make the `exact _.` you refer to go through. (I got all of `Truncations.Core` working with that approach yesterday, but won't recompile now to try again.) Keeping track of all such cuts will make a mess however.

With just the second cut my build time is a bit worse:
```sh
real    5m13.113s
user    11m43.226s
sys     1m24.836s
```
Maybe having both saves time. And I've yet to see any examples of where such cuts rule out proofs in practice.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -113,7 +113,8 @@ Defined.
 Corollary ishprop_path_subgroup `{U : Univalence} {G : Group} {H K : Subgroup G}
   : IsHProp (H = K).
 Proof.
-  rapply (equiv_transport IsHProp _ _).
+  (** jarlg: Without "n" here, it unifies (IsHProp ?x) with (IsHprop Unit). Can that be avoided? *)
+  nrapply (transport IsHProp).
```
jdchristensen, line 117: That's really weird.  Why would it arbitrarily choose `Unit` to be the other end of the path?  In any case, the `n` prefix should be the default except when we want typeclass inference to happen, so the change is fine.

theories/Basics/Overture.v
```diff
@@ -712,6 +712,8 @@ Register tt as core.True.I.
 (** A space is pointed if that space has a point. *)
 Class IsPointed (A : Type) := point : A.
 
+Typeclasses Transparent IsPointed.
```
jdchristensen, line 715: I think this change should be put in a separate PR that illustrates why it is helpful.  This PR compiles fine without it.

theories/Pointed/Loops.v
```diff
@@ -180,7 +180,8 @@ Proof.
   intros p; eapply isconnected_equiv'.
   - refine (hfiber_loops_functor f p oE _).
     symmetry; apply hfiber_ap.
-  - exact _.
+    (** jarlg: "exact _." works here but is a bit slow. *)
```
jdchristensen, line 183: It's not too important, but maybe some costs can be adjusted to make it faster?

theories/Truncations/Connectedness.v
```diff
@@ -128,6 +128,7 @@ Definition merely_path_is0connected `{Univalence}
 : merely (x = y).
 Proof.
   refine ((equiv_path_Tr x y)^-1 (path_contr (tr x) (tr y))).
+  exact H0.
```
jdchristensen, line 131: I wonder why this isn't found, but was before.  In any case, `assumption.` is better, since `H0` is a generated name.

theories/Basics/Overture.v
jarlg: Okay, but it pertains to the discussion in the sense that having more stuff being instances might require adding things like this. Here's a minimal working example that can be added just after `Arguments point A {_}.` in Overture.v:

```coq
Record pSet := {
  pset_type :> Type;
  pset_ishset : IsHSet pset_type;
  pset_ispointed : IsPointed pset_type;
}.

Arguments pset_ispointed {_}.

Existing Instance pset_ishset.
Existing Instance pset_ispointed.

Example needs_transparent {A : pSet} : IsHProp (point A = pset_ispointed).
Proof.
  typeclasses eauto. (* fails if IsPointed isn't Transparent. *)
Defined.
```

Should I add this to the PR or save it for later?

theories/Algebra/Groups/Subgroup.v
jarlg: This is what Coq finds:
```coq
Debug: 1: looking for (IsHProp ?x) with backtracking
Debug: 1.1: simple apply @istrunc_inO_tr on (IsHProp ?x), 1 subgoal(s)
Debug: 1.1-1 : (In (Tr (-1)) ?x)
Debug: 1.1-1: looking for (In (Tr (-1)) ?x) with backtracking
Debug: 1.1-1.1: simple apply inO_unit on (In (Tr (-1)) ?x), 0 subgoal(s)
Debug: Finished run 1 of resolution.
```
I think a good solution to this would be that `inO_unit` shouldn't be allowed to unify an evar `?x` with `Unit`. That way the instance is always found whenever we know we're talking about `Unit`, but it won't suggest `Unit` for a general `?x`. I'm trying to use `Mode Hint` to set that up, but haven't managed so far..

theories/Pointed/Loops.v
jarlg: Probably; I'd be wililng to investigate optimizations once we've agreed whether `Hint Cut` is feasible at all.

theories/Truncations/Connectedness.v
jarlg: Previously the proof was to jump back and forth once along one of the equivalences (returning to the starting point), then apply `assumption.` Not sure why `assumption.` isn't just tried from the get-go, or if something actually happened after those jumps.

theories/Algebra/Groups/Subgroup.v
jarlg: Let me just record that replacing the `inO_unit` instance with this works and compiles:
```coq
    Hint Extern 0 (In ?O Unit) => simple apply inO_unit; solve [ trivial ] : typeclass_instances.
```
It seems the matching for `Hint Extern` is stricter. But Coq ends up finding something else and producing a weirder goal; `nrapply` is probably the way to go.

theories/Truncations/Core.v
jdchristensen: I think we should also add cuts that prevent going back and forth with nothing in between, like this:
```coq
Hint Cut [( _* ) inO_tr_istrunc istrunc_inO_tr] : typeclass_instances.
Hint Cut [( _* ) istrunc_inO_tr inO_tr_istrunc] : typeclass_instances.
```
These should be combined with either both of the hints in this PR, or maybe just the second one.  I still have some hope that we can just have the second cut and get the build time to be good.

BTW, once you find a file whose build time is affected a lot by these hints, you could do some testing using `time make -j<N> path/to/file.vo` which will just build that file and the dependencies, making it easier to experiment quickly.

theories/Truncations/Connectedness.v
Alizter: Instead of refine you could use exact no?

theories/Algebra/Groups/Subgroup.v
jdchristensen: I agree with you that `inO_unit` should be fixed to not apply unless `Unit` is already there.  It might speed up some other things.  Once that is done, then the Instance that applies and produces a strange goal is `Global Existing Instance O_inO.`  Again, it seems like this shouldn't be used unless we have a type already in mind, but I'm not sure how to enforce that.  One of the typeclass experts can probably tell us.

What is happening here is that our more powerful Instances are turning up what I'd say are existing buglets in the typeclass instances we already have, so I wouldn't be deterred by this.

theories/Basics/Overture.v
jdchristensen: I think the `Typeclasses Transparent IsPointed.` should be saved for later.  The example shouldn't be needed, since we already have a real-world example that will illustrate why it's useful to have IsPointed be transparent.

theories/Truncations/Connectedness.v
jdchristensen: No, exact gets stuck with the same Contr goal that the hypothesis H0 solves.  H0 could probably be passed in using @ notation or something like that, but that's not any simpler.

theories/Truncations/Core.v
jarlg: My interpretation of the docs (and kleene star) is that the pattern `( _*)` will match zero-or-more hint applications, so "back and forth with nothing in between" is already covered. Or have your observed the opposite?

Thanks for the additional info!

jdchristensen: I am proposing:
```coq
Hint Cut [( _* ) inO_tr_istrunc ( _* ) inO_tr_istrunc] : typeclass_instances.  (* Optional, hopefully can be skipped. *)
Hint Cut [( _* ) istrunc_inO_tr ( _* ) istrunc_inO_tr] : typeclass_instances.
Hint Cut [( _* ) inO_tr_istrunc istrunc_inO_tr] : typeclass_instances.
Hint Cut [( _* ) istrunc_inO_tr inO_tr_istrunc] : typeclass_instances.
```
The last two are not covered by the first two, since the last two involve both instances, not a repeated instance.  Allowing immediate back and forth is almost certainly a waste, and will cause a large blow up in the search space when it is possible.  So I think the last two cuts should speed things up and cause zero harm.  I tested the "zero harm" part, but didn't test the speed change.

jarlg: Sorry, I read too quickly. I agree that's a desired cut, and I'll look at the performance implications when I get to profiling.

jdchristensen: Ok, I tested the speed change.  Adding two additional cuts like the last two above, for both inO and mapinO, saves around 5% in the building of the dependencies of Truncations/Core.v.  So while it helps, it doesn't come close to bringing us back to the former speed.  I used `time make -j4` on a 4 core machine.
```
Time for building the dependencies of Truncations/Core.v on jdc with "make -j4".

With Jarl's 2 Hints in both sections:

real    1m33.105s
user    3m36.243s
sys     0m17.414s

With 4 Hints in both sections:

real    1m26.591s
user    3m26.560s
sys     0m17.416s

HoTT-master:

real    1m3.541s
user    2m45.068s
sys     0m17.422s
```
To be clear, this is without doing a `make clean`, so it focuses only on the part of the library that gets rebuilt when Core.v changes, i.e., the part that depends on these instances.

jarlg: Nice, that's great! There's also the aspect of the real-time impact of having lots of cuts, since these are run on every hint application, but we can get to profiling that another time.

Feel free to add the cuts to the draft, or I'll do it once I'm on a clean branch.

theories/Truncations/Connectedness.v
Alizter: In which case you can decorate the `refine` with `by` which just tries `assumption` and `trivial` after running the tactic after it.

jarlg: Oh, that's nice. Thanks!

theories/Algebra/Groups/Subgroup.v
jarlg: Adding the following mode hint to Overture restricts proof search for `IsTrunc ?n ?X` (e.g. `IsHProp ?X` with `n` known) to only search for types when the truncation index `n` is known (+) and the head of the desired type is known (!).
```coq
Hint Mode IsTrunc + ! : typeclass_instances.
```
This stops search for `IsHProp ?x` since the head of `?x` isn't known, and is compatible with the library (i.e. it builds without any changes). The `rapply` here then just stops immediately, and so can be left as-is.

jdchristensen: That `Hint Mode` setting seems reasonable to me, but I hope some experts will chime in.  It should in theory speed up typeclass search and avoid spurious matches like you observed.  BTW, if `nrapply` works, I think it's better to use it even if `rapply` also works.  It'll be faster and more robust, and it tells the reader that no magic is happening.

jdchristensen: I'm not sure, though, whether the `+` makes sense.  Maybe it's ok for the truncation index to be existential?

theories/Truncations/Core.v
jdchristensen: Interestingly, it's actually a bit faster to drop the first Hint Cut.  I'm not sure why this would be the case, but I did both runs twice to make sure the timing was consistent.
```
With last 3 Hints in both sections:

real    1m23.791s
user    3m20.929s
sys     0m17.414s
```

theories/Truncations/Connectedness.v
jdchristensen: I don't think you need to be so fancy.  `refine (...); assumption.` works fine.  But I like the period better than the semicolon, so the reader can see which assumption is getting used here.  So I'd prefer to just replace `exact H0` with `assumption`.

theories/Algebra/Groups/Subgroup.v
jarlg: I should have included a complete proposal for `IsTrunc`, which is:
```coq
Hint Mode IsTrunc + ! : typeclass_instances.
Hint Mode IsTrunc - + : typeclass_instances.
```
This builds fine. The last hint means we only search of index when the type is known, which is needed e.g. in Truncations/Core.v. Maybe the + should be relaxed to ! for the last hint, but it isn't currently needed.

jdchristensen: Maybe just the second `Hint Mode` is enough?  Or just `Hint Mode IsTrunc - ! : typeclass_instances.` which covers both cases and more?

jarlg: The second hint alone would mean the type must always be known, which breaks e.g Categories/Category/Sigma/Core.v line 232 and Homotopy/BlakersMassey.v line 212 (which are hefty lines admittedly).

I'd be fine with your one-liner, though it's more relaxed than currently needed. Would be interesting to know if there are performance implications to multiple `Hint Mode`s.

jdchristensen: I'd say we should be more relaxed unless we know a reason not to be.  IIRC, the only bad case we know is where a completely unknown type was instantiated with `Unit`, and the `- !` version should avoid that.

jarlg: I was thinking about this the wrong way, trying to make the modes as strict as possible in order to optimize current proof search. But your heuristic is better, we should be as relaxed as possible and only remove the known unwanted proof searches. So I'm in favor of your one-liner then!

mikeshulman: This reminds me of the conversation at #479.  I was under the impression that the `Local Set Typeclasses Strict Resolution` at the top of `Overture` would prevent classes declared therein, such as `IsTrunc`, from doing this sort of undesired guessing.  But it seems like `Hint Mode !` is newer than `Set Typeclasses Strict Resolution`, so maybe it's the preferred way of doing things now?  Does anyone understand the relation between these two options?

mikeshulman:
> To be honest, the layers involving IsTrunc, IsTrunc_internal, In, In_internal, Contr and Contr_internal are very confusing.

I've thought for a long time that the layers of Contr_internal / Contr / IsTrunc_internal / IsTrunc are too clever by half.  The comment preceding the definition of `IsTrunc_internal` in `Overture` explains why all four of them exist, but it would be worth experimenting to see whether they could be simplified with not much loss of convenience.

As for In_internal / In, I believe the only reason for that separation is a Coq bug that a field of a record can't (effectively) be declared as an `Existing Class`.  I don't know whether there's any chance of fixing that bug (or even whether it might have been fixed already!).

mikeshulman:
> IsTrunc n A and In (Tr n) A are definitionally equal, and the Instances we are considering are just using reflexivity to switch between them. ... Can we have only one of these forms in the library so no translation is needed?

You may remember that I tried briefly to achieve this in #1209 and #1224, but we decided to back off.

Alizter:
>As for In_internal / In, I believe the only reason for that separation is a Coq bug that a field of a record can't (effectively) be declared as an Existing Class. I don't know whether there's any chance of fixing that bug (or even whether it might have been fixed already!).

I had opened an issue here https://github.com/coq/coq/issues/12975

Currently there is a draft PR trying to fix this: https://github.com/coq/coq/pull/9711

I've pinged the relevant devs so it can be added to coq soon.

jdchristensen:
> I've thought for a long time that the layers of Contr_internal / Contr / IsTrunc_internal / IsTrunc are too clever by half. The comment preceding the definition of `IsTrunc_internal` in `Overture` explains why all four of them exist, but it would be worth experimenting to see whether they could be simplified with not much loss of convenience.

The reasons for Contr_internal and IsTrunc_internal seem to be different, and I don't really understand them.  For IsTrunc_internal, the comment says that the reason is so that `intros x y` works.  But it seems to already work with the internal version, so maybe Coq has changed since that design decision was made?

```coq
Definition test (A : Type) (n : trunc_index) : IsTrunc_internal n.+1 A.
intros x y.  (* Works. *)
```

Even if I'm misunderstanding the issue, surely we could just have a lemma that unfolds IsTrunc_internal that we apply when we need it.

The only reason given for Contr_internal is "to not pollute typeclass search", but I don't know precisely what would go wrong without it.

> As for In_internal / In, I believe the only reason for that separation is a Coq bug that a field of a record can't (effectively) be declared as an `Existing Class`. I don't know whether there's any chance of fixing that bug (or even whether it might have been fixed already!).

`Existing Class In_internal` doesn't itself give an error, but (with appropriate changes to this section), later parts of the file still fail.  This is minor, but it would be nice if it wasn't needed.

mikeshulman:
> For IsTrunc_internal, the comment says that the reason is so that `intros x y` works. But it seems to already work with the internal version

I believe the reason it works for the internal version is that the internal version is not declared as `simpl never`, whereas `IsTrunc` is declared as `simpl never`.  Now that I read the comment again I see that it says `simpl nomatch` in several places where it should say `simpl never`, so that is probably the source of confusion!

> The only reason given for Contr_internal is "to not pollute typeclass search", but I don't know precisely what would go wrong without it.

I'm not sure either.

> `Existing Class In_internal` doesn't itself give an error, but (with appropriate changes to this section), later parts of the file still fail.

Yes, that's what i meant by "can't (effectively) be declared": Coq accepts the declaration, but it doesn't work.

jdchristensen:
> > For IsTrunc_internal, the comment says that the reason is so that `intros x y` works. But it seems to already work with the internal version
> 
> I believe the reason it works for the internal version is that the internal version is not declared as `simpl never`, whereas `IsTrunc` is declared as `simpl never`. Now that I read the comment again I see that it says `simpl nomatch` in several places where it should say `simpl never`, so that is probably the source of confusion!

Oh, right.  With `simpl never` you are right that `intros x y` fails.  But still, but can just make a lemma that does this:

```coq
Arguments IsTrunc_internal n A : simpl never.

Lemma conv (A : Type) (n : trunc_index) :
  (forall x y : A, IsTrunc_internal n (x = y))
  -> IsTrunc_internal n.+1 A.
exact idmap.
Defined.

Definition test (A : Type) (n : trunc_index) : IsTrunc_internal n.+1 A.
apply conv.
```

Is it worth a definitional wrapper so one can type `intros x y` instead of `apply lemma_with_good_name`?  Or  maybe there are other consequences that aren't mentioned in that comment.

mikeshulman:
Or a tactic that executes a `change`, to avoid polluting the proof term with an idmap.  This is the sort of thing I had in mind re: experimenting to see whether they could be simplified with not much loss of convenience.

jdchristensen:
> > IsTrunc n A and In (Tr n) A are definitionally equal, and the Instances we are considering are just using reflexivity to switch between them. ... Can we have only one of these forms in the library so no translation is needed?
> 
> You may remember that I tried briefly to achieve this in #1209 and #1224, but we decided to back off.

Right.  #1224 is especially relevant.  But maybe we should be willing to accept some other compromises in order to avoid the extra layer.  I don't mind if `IsTrunc n A` is displayed using `In` or in some other reasonable way.  Or maybe we keep `IsTrunc` separate, but simply have a convention that except for the parts of the library that come before Modalities, we never refer to IsTrunc, but instead use the In form.  So IsTrunc is an internal implementation detail used while bootstrapping things.  And we duplicate typeclass instances currently stated for IsTrunc in the In form.  Or maybe there's a reasonable change to Coq that could let us simplify this.  Just throwing out ideas here.

JasonGross:
Back when I was designing all of this, `Hint Cut` wasn't a thing.  So in order to make typeclass search reasonable, the classes need to form a dag (that is, there should be a partial order of classes such that every instance returning a class `C` has only hypotheses mentioning classes <= C).  With `Hint Cut` this can be relaxed, though I don't think it's enough to have typeclass search work well when instances are added ad-hoc; there should be a plan about the entire typeclass search and how any given instance is going to be used in conjunction with other instances.

jdchristensen:
I've discussed this with Jarl, and we think the approach in this PR is probably not the best way forward, although it's good to have it in the github history as a possible tool for dealing with loops in typeclass inference.  I think the other ideas for trying to avoid having so many equivalent ways of expressing definitionally equal things would be more suitable, if any of them are possible.

PR #1409, opened 2021-01-13
mention in INSTALL that you need to install zarith from opam

mikeshulman:
I had to do this before I could compile our currently pinned Coq.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -49,7 +49,7 @@ You may also need to run
 to continue working in the same terminal.  Then you can install the
 required dependencies:
 
-    opam install ocaml camlp5 lablgtk3 lablgtk3-sourceview3 ocamlfind num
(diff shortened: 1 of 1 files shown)


jdchristensen:
Yes, I had to do that a couple of months ago too.  I thought it was something messed up in my config.  BTW, I didn't install the gtk packages.  It's mentioned earlier on the page that these are for coqide, but maybe we should separate them out in the cut-and-paste part, with a repeat of that remark?

jdchristensen:
The patch itself LGTM.

mikeshulman:
Yeah, that makes sense.

Alizter:
LGTM

Alizter:
So the configure script from the main coq repo wasn't correctly raising the fact that zarith wasn't installed. https://github.com/coq/coq/issues/13041

Now that that has been fixed, we should update our configure script to also do this.

PR #1413, opened 2021-01-17
update to V8.13

Alizter:


mikeshulman:
Looks good.  Shall we add explicit hint localities, to get rid of the warnings?
<https://coq.inria.fr/refman/changes.html#hintwarning>

Alizter:
I've done a crude conversion of hints to global hints. This meant that coq was able to finish some proofs better, so I shortened those proofs.

I've also changed a few other things for various warnings to be solved.

SkySkimmer:
>I've done a crude conversion of hints to global hints. This meant that coq was able to finish some proofs better, so I shortened those proofs.

That's surprising, adding global isn't supposed to change the effect other than suppressing the warning.

theories/Algebra/Groups/ShortExactSequence.v
```diff
@@ -40,8 +40,6 @@ Proof.
     + rapply (transport IsHProp (x:= grp_trivial)).
       apply path_universe_uncurried.
       rapply Build_Equiv.
```
Alizter, line 42: Here coq was completely able to guess the equivalence.

theories/Homotopy/HSpace.v
```diff
@@ -28,22 +28,20 @@ Section HSpaceProperties.
     : forall (a : A), IsEquiv (fun x => a * x).
   Proof.
     refine (conn_map_elim (-1) (unit_name hspace_id) _ _).
-    + exact (conn_point_incl hspace_id).
```
Alizter, line 31: coq was able to guess this

theories/Spaces/BAut/Rigid.v
```diff
@@ -104,8 +104,7 @@ Proof.
     assert (MH : forall (a:A) (f:X*A -> X*A) (x:X),
                fst (f (x,a)) = fst (f (x,a0))).
     { refine (conn_map_elim (Tr n) (unit_name a0) _ _).
-      - apply conn_point_incl; assumption.
```
Alizter, line 107: coq was able to guess this

theories/Truncations/Connectedness.v
```diff
@@ -76,12 +76,12 @@ Proof.
   apply O_lex_leq_Tr.
 Defined.
 
-Hint Immediate conn_point_incl : typeclass_instances.
-
 (** Note that [OO_cancelR_conn_map] and [OO_cancelL_conn_map] (Proposition 2.31 of CORS) generalize the above statements to 2/3 of a 2-out-of-3 property for connected maps, for any reflective subuniverse and its subuniverse of separated types.  If useful, we could specialize that more general form explicitly to truncations. *)
 
 End Extensions.
 
+#[global] Hint Immediate conn_point_incl : typeclass_instances.
```
Alizter, line 83: For some reason, before coq didn't have the same behaviour with this hint

theories/Categories/Category/Morphisms.v
```diff
@@ -193,6 +194,7 @@ Section iso_equiv_relation.
     | abstract iso_comp_t @right_inverse ].
   Defined.
 
+  #[local]
   Hint Immediate isisomorphism_inverse : typeclass_instances.
```
SkySkimmer, line 198: No need to add the attribute in sections btw

SkySkimmer:
Oh, because HoTT uses `Global Set Loose Hint Behavior "Strict".` we should use export, not global.

mikeshulman:
That's what I was going to say, we should use export.

What's the reason for the weird syntax `#[global] Hint` instead of something like `Global Hint` like all other Coq commands?

theories/Categories/Category/Morphisms.v
mikeshulman: In other cases our convention is to add locality markers even when they are redundant, so as to avoid forgetting them when they are not, e.g. we write `Global Instance` all the time even in places where globality is the default.

mikeshulman:
I assume that with "export", the places where "global" allowed Coq to guess something new would also be made guessable by an additional `Require Import` at the top of the file to import the missing hint(s).

Alizter:
OK should I replace global with export then?

mikeshulman:
But even with `export` instead of `global` Coq can still guess some extra things, hm?  That's interesting.

I would be in favor of keeping the `local` attributes as "checked documentation" inside of sections, even though they are redundant.  What do you think?

Alizter:
> I would be in favor of keeping the local attributes as "checked documentation" inside of sections, even though they are redundant. What do you think?

I think that this is a good idea too. I'll see what I can do today.

coq/theories/Init/Datatypes.v
```diff
@@ -68,7 +68,7 @@ Notation "A /\ B" := (prod A B) (only parsing) : type_scope.
 Notation and := prod (only parsing).
 Notation conj := pair (only parsing).
 
-Hint Resolve pair inl inr : core.
+#[global] Hint Resolve pair inl inr : core.
```
SkySkimmer, line 71: Could be one of the remaining global somehow

Alizter:
I caught a few more warnings, and changed to keep the local attribute.

mikeshulman (approved):

PR #1416, opened 2021-02-02
Formalization of IW-types

Alizter:
Here is a formalisation of Indexed W-types based on [Jasper Hugunin's formalisation](https://github.com/jashug/IWTypes). The contents are summarised as follows:

* Did some cleanup in W-type.
* We define IW-types
* We show that IW-types can be reduced to W-types with definitionally computing rules
* We characterise path types of IW-types as another IW-type
* We characterise the path type of fiber of the index map of an IW-type
* We show that an IW-type is n-truncated if the index map is n-truncated
* We show that an IW-type has decidable equality if the fibers of the index map have decidable equality and a condition on the family of arities holds.

AFAIK some of these results are not written down anywhere in the literature, so it's nice to be able to collect them together like this.

When I started this around half a year ago, I also thought about writing a plugin that would take a given coq inductive type and spit out an equivalent IW-type. It might be a useful machine to have when proving lemmas about inductive types. But I haven't thought about investing any time into yet, since I don't know if it's really useful to do.


Alizter:
OK CI has failed, @JasonGross do you know what is going on?

JasonGross:
It seems that ubuntu-latest now refers to 20.04 rather than 18.04, I'm guessing?  You can either explicitly ask for `ubuntu-18.04` or else figure out how to fix the broken/missing packages.

mikeshulman:
Here's a question that just occurred to me; maybe you have a thought @Alizter, or maybe @jashug does.  The path types of *any* type X are always IW-types in a trivial (non-recursive) way, generated by reflexivity at each element of X.  Can this characterization in the case X=IW somehow be viewed as substituting the universal property of IW into the domain of reflexivity?

jashug:
@mikeshulman 

Can you elaborate on what you mean by "substitute the universal property of IW into the domain of reflexivity"?

<details>
<summary>Some scribbles as I tried to work this out, do you recognize anything in here?</summary>

There is an equivalence
`path X x1 x2 =equiv= (Inductive IWpath_triv : X -> Type := refl : 1 -> (0 -> IWpath_triv x1) -> IWpath_triv x1) x2`,
is that what you are referring to with "The path types of any type X are always IW-types in a trivial (non-recursive) way, generated by reflexivity at each element of X."?

And we have the universal property of
`(Inductive IW : I -> Type := sup (x : A) (ch : forall (y : B a), IW (j x y)) : IW (i x))` saying that for any other pair of `T : I -> Type` and `supT : forall (x : A) (ch : forall (y : B a), T (j x y)), T (i x))`, there exists a unique function `rec : forall l, IW l -> T l` such that `rec (sup x ch) = supT x (rec o ch)`.

Equivalently, the dependent induction principle is that for every motive `P : forall l, IW l -> Type` and induction step `IS : forall (x : A) (ch : forall (y : B a), IW (j x y)) (IH : forall y, P (j x y) (ch y)), P (i x) (sup x ch)`, there exists (and we can prove it is unique) a function `el : forall l (a : IW l), P l a` such that `el (i x) (sup x ch) = IS x ch (el o ch)`.

The type of reflexivity in this special case is `refl : forall (l : I) (a : IW l), path (IW l) a a`.
If we set `P l a = path (IW l) a a`, this has the same type as `el` above. Thus, we should set `IS x ch IH = refl _ (sup x ch)`? But that doesn't make use of the inductive hypothesis, we might rather like `IS x ch (IH : forall y, path (IW (j x y)) (ch y) (ch y)) = ap_(sup x) (funext IH)`, or something. And we can take this second definition of `IS`, because
`refl _ (sup x ch) = ap_(sup x) (funext (refl o ch))`.

This looks similar to the characterization in this formalization
`path (IW l) a b =equiv= (Inductive IWpath : forall l, IW l -> IW l -> Type := supIWpath : forall (x : A) (chl chr : forall (y : B a), IW (j x y)), (forall y, IWpath (j x y) (chl y) (chr y)) -> IWpath (i x) (sup x chl) (sup x chr)) l a b`,
but I'm not sure if it looks like substituting the universal property of IW into the domain of reflexivity.
</details>

Alizter:
I've also changed the CI to use ubuntu-18.04. Currently there are some packages that are obsolete in ubuntu 20 that we need to fix, but this is out of the scope of this PR.

theories/Types/WType.v
```diff
@@ -0,0 +1,53 @@
...
+
+Definition w_label {A B} (w : W A B) : A :=
+  match w with
+  | w_sup x y => x
+  end.
+
+Definition w_arg {A B} (w : W A B) : B (w_label w) -> W A B :=
+  match w with
+  | w_sup x y => y
+  end.
```
JasonGross, line 18: Why separate these out?  The bug in Coq has been fixed, primitive projections now work fine with recursive records.

Alizter: I get other problems down the line, for example switching to:
```coq

Inductive W (A : Type) (B : A -> Type) : Type
  := w_sup 
  {
    w_label : A ;
    w_arg   : B w_label -> W A B ;
  }.

Arguments w_label {A B}.
Arguments w_arg {A B}.
```

prevents the issig lemma from going through.

mikeshulman:
@jashug That's the direction I was thinking in.  But rather than defining reflexivity by IW-induction, I was thinking of saying, suppose we're given a motive `P : forall (l:I) (x y:IW l), (x = y) -> Type` and we want to prove `forall l x y p, P l x y p`.  We can first apply path induction on `p` to reduce the goal to `forall (l:I) (x:IW l), P l x x (refl x)`.  Then we can apply IW-induction on `x` to reduce this to something else.  Is this two-step process similar to what we would get by replacing `x = y` by the IW-type it's equivalent to and using *its* induction principle?

jashug:
@mikeshulman Ah, I think I see what you are getting at.

IW-induction on the goal `forall (l:I) (a:IW l), P l a a (refl a)` turns it into the goal
```
x : A
ch : forall y, IW (j x y)
IH : forall y, P (j x y) (ch y) (ch y) (refl (ch y))
----------------------------------------------
P (i x) (sup x ch) (sup x ch) (refl (sup x ch))
```

Whereas using the induction principle you get from the IW-type equivalent to `(x = y)` takes a goal `forall l x y p, P l x y p` and turns it into
```
x : A
chl chr : forall y, IW (j x y)
chp : forall y, chl y = chr y
IH : forall y, P (j x y) (chl y) (chr y) (chp y)
--------------------------------------------
P (i x) (sup x chl) (sup x chr) (ap (sup x) (funext chp))
```
Does that make sense? I think there is a significant difference here, but of course everything is interderivable.

mikeshulman:
Yes, that's what I meant!  You can then get from the second goal to the first by applying funext to `chp` and then doing a path induction on it.  So those goals don't seem very different to me.

jashug:
Indeed you can, and there should be a step in this formalization that does exactly that (I think in section `Ind`, around line 175, or specifically the proof `IWPath_ind_refl`).

mikeshulman:
Is there a different IW-type that's also equivalent to paths in our original IW-type and whose native induction principle would be your first goal rather than your second?

jashug:
Hmn, I wanted to say no, but I think you can get close with
`Inductive IWpaths : forall l : I, IW l -> IW l -> Type := sup_refl (x : A) (ch : forall y, IW (j x y)) : IWpaths (i x) (sup x ch) (sup x ch).`
Here we take the usual definition of paths as an inductive family, and apply the equivalence
`{l : I & IW l} = {x : A & forall y, IW (j x y)}` in the domain of `refl`, just as you originally suggested.
That takes a goal `forall l x y p, P l x y p` and turns it into
```
x : A
ch : forall y, IW (j x y)
-----------------------
P (i x) (sup x ch) (sup x ch) (sup_refl x ch)
```
which is close to the first goal/code-block in my above response, but has no inductive hypothesis, because this is a non-recursive inductive family.

In words, this just says that any path with two unconstrained endpoints and unconstrained index `l : I` is just `refl (sup x ch)`.
But I'm not sure that is interesting. My "criterion for success" when writing the original version of this formalization included something like being a properly recursive inductive family, with an infinitary inductive argument indexed by "B x" or something derived from it.

mikeshulman:
Hmm, yes, I think you're right, that is not very interesting.

theories/Types/IWType.v
```diff
@@ -0,0 +1,577 @@
...
+  intros I A B i j h l.
+  intros x y.
+  refine (trunc_equiv' _ (equiv_path_iwtype I A B i j l x y) (n := n.+1)).
+  apply IHn.
+  intros [k [a b]].
+  (** The crucial step is to characterize the fiber of [i'] which was done previously. *)
+  apply (trunc_equiv' _ (equiv_path_hfiber_index I A B i j k a b)).
+Defined.
+
+(** TODO: Clean up rest of file *)
```
Alizter, line 427: So I need some help doing this part, since I am not really sure what to do with all these little steps marching towards proving decidable equality. Should we separate these as separate lemmas? If so where would we put them? I could probably make a few more of these definitions `Local` come to think of it, as I really don't see them being useful outside of this file.

Alizter:
Which modalities/RSUs do IW-types "preserve"? The n-truncated index map giving n-truncated IW-types is interesting, but it doesn't work for n=-2, so clearly it doesn't work for all modalities.

theories/Types/WType.v
mattam82: If you make a primitive record for a recursive type, you don't get eta-equality (that would lead to undecidable conversion checking) and hence no dependent elimination. But you could write it this way to avoid the issue:
```
Set Primitive Projections.
Record W_record (A : Type) (B : A -> Type) (I : Type) : Type
  := w_sup 
  {
    w_label : A ;
    w_arg   : B w_label -> I ;
  }.
Arguments w_sup {A B I}.
Arguments w_label {A B I}.
Arguments w_arg {A B I}.

Inductive W A B := 
  | fold (unfold : W_record A B (W A B)) : W A B.
Arguments fold {A B}.

Definition unfold {A B} (w : W A B) := 
  let 'fold f := w in f.

Lemma w_eq {A B} (w : W A B) : w = fold (w_sup (w_label (unfold w)) (w_arg (unfold w))).
Proof.
  destruct w. cbn. reflexivity.
Qed.

Goal forall A B (P : W A B -> Type), (forall (a : A) (w_arg : B a -> W A B), P (fold (w_sup a w_arg))) ->
  forall w, P w.
Proof.
  intros.
  destruct w. apply X.
Qed.
```
Not sure it would be the best formalization choice though.

Alizter:
I've rebased and done a little cleanup. I must admit that I don't really understand how the decidable paths for IW-types proof works, so I haven't been able to clean it up much.

This is ready to be merged.

Alizter:
OK I've cleaned up the decidable paths proof. This is now ready to be reviewed properly.

Alizter:
rebased

mikeshulman (approved):
Sorry to take so long to look at this.

PR #1419, opened 2021-02-27
Algebra, homomorphism and isomorphism

andreaslyn:
This is the first PR developing infinitary universal algebra. The next PR will contain a predicate `IsEquational` for algebras with equations.

Later PRs will define (univalent) category of algebras and construct direct product, subalgebra and quotient. There is moreover a proof which shows that constructively the quotient map cannot generally be surjective.

I think we should remove or deprecate the old (finitary) universal algebra when the infinitary has been merged, since the infinitary universal algebra supersedes the finitary.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -327,6 +327,13 @@ theories/Spectra/Spectrum.v
 theories/Spectra/Coinductive.v
 
 #
+#   Universal Algebra
+#
+
+theories/Algebra/Universal/Algebra.v
+theories/Algebra/Universal/Homomorphism.v
+
+#
 #   Algebra
 #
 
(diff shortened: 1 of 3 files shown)


mikeshulman:
LGTM, thanks!

Are you sure that the infinitary version supersedes the finitary one for practical use as well as theoretical purposes?  E.g. if I wanted to define, say, a Lie algebra as a special kind of universal algebra, would there be any advantage to using the finitary library rather than the infinitary one (e.g. some extra definitional equalities, perhaps)?

mikeshulman:
(BTW, I verified locally that this compiles, since the CI is stil broken.)

andreaslyn:
Good point. There may be a practical advantage in that the finitary universal algebra works directly with curried operations `A -> A -> A`. The corresponding operation in the infinitary univarsal algebra would be of the form `(Fin 2 -> A) -> A`. In my repo, I have a function to turn a curried function `A -> ... -> A` into the uncurried form `(Fin n -> A) -> A`, but even with such conversions, I anticipate that the uncurried operations will make it less convenient.

I am planning to add some algebra instances, to serve as examples, and also to implement a practical interface on the way. For example, shorthands to define nullary, unary and binary operations, and for common laws, associativity, commutativity, unit, etc.

andreaslyn:
The CI is passing now after a rebase.

PR #1420, opened 2021-02-28
Update ci.yml

Alizter:
Hopefully this fixes the ci issue. This seems to be because of a change in the npm packaging.


Alizter:
Actually, I just remembered that I fixed this issue in #1416 but never merged that commit. Let me see if I can cherrypick it.

PR #1422, opened 2021-03-04
[coq] Overlay for coq/coq#13617

ejgallego:
You may want to make this change conditional on the Coq version, but
it is out of my league. Additionally, there is quite a bit of cruft
around here that it is not needed at all.

The build process without stdlib should be very simple these days.


Alizter:
I think we will wait for the version to make this change. We need to do a bit more work to remove our copy of the coq stdlib which should be possible now.

There is also the discussion of switching to dune that needs to be had here, but I need some coqdoc support in dune before that can be done.

Alizter:
Can you rebase?

ejgallego:
> Can you rebase?

Hi @Alizter , I think this is not needed anymore, waiting on the CI of the upstream PR to confirm.

ejgallego:
Indeed, this is not needed at all anymore.

PR #1428, opened 2021-03-25
Algebra operations

andreaslyn:
Started work on algebra operations

Added Finite/FinSeq.v

Make some definitions monomorphic to fix universe blowup in Finite/FinNat.v


theories/Algebra/Universal/Operation.v
```diff
@@ -0,0 +1,219 @@
...
+  HoTT.Spaces.Finite
+  HoTT.Spaces.Nat
+  HoTT.Spaces.Finite.Vect.
+
+Import notations_algebra.
+
+(** Functions [head_dom'] and [head_dom] are used to get the first
+    element of a nonempty operation domain [a : forall i, A (ss i)]. *)
+
+Monomorphic Definition head_dom' {σ} (A : Carriers σ) (n : nat)
```
andreaslyn, line 21: I am making certain definitions monomorphic to avoid unverse blowup.

theories/Spaces/Finite/Fin.v
```diff
@@ -112,14 +112,14 @@ Proof.
     + rewrite path_fin_fsucc_incl, path_nat_fin_incl.
       apply IHn.
     + reflexivity.
-Qed.
+Defined.
```
andreaslyn, line 115: I need the computation for `operation_curry`.

theories/Spaces/Finite/FinNat.v
```diff
@@ -79,7 +79,8 @@ Proof.
                    (hset_path2 1 (@path_succ_finnat n u u.2)) idpath).
 Defined.
 
-Definition is_bounded_fin_to_nat {n} (k : Fin n) : fin_to_nat k < n.
+Monomorphic Definition is_bounded_fin_to_nat {n} (k : Fin n)
```
andreaslyn, line 82: Monomorphic to avoid universe blowup

theories/Spaces/Finite/Vect.v
```diff
@@ -0,0 +1,212 @@
...
+  HoTT.HSet
+  HoTT.DProp
+  HoTT.Spaces.Finite.Fin
+  HoTT.Spaces.Finite.FinInduction
+  HoTT.Spaces.Nat.
+
+(** Finite-dimensional vector. Note that the induction principle
+    [vect_ind] uses funext and does not compute. *)
+
+Definition Vect (n : nat) (A : Type) : Type := Fin n -> A.
```
andreaslyn, line 13: I am not sure whether vector is a good name for this.

theories/Spaces/Finite/Vect.v
```diff
@@ -0,0 +1,212 @@
...
+  : vtail (vcons a v) = v.
+Proof.
+  funext i.
+  apply compute_vtail'.
+Defined.
+
+(** A non-empty vector is the [vcons] of its head and tail,
+    [path_expand_vcons'] and [path_expand_vcons]. *)
+
+Lemma path_expand_vcons' {A : Type} (n : nat)
```
andreaslyn, line 87: The remaining lemmas are here to prove `vect_ind`. I do not think that I will need it, but I formalised `vect_ind` for completeness.

theories/Spaces/Finite/Fin.v
mikeshulman: That's curious, since it's a path in `nat`.  Is it not possible to just use the fact that `nat` is a set?

theories/Spaces/Finite/Vect.v
mikeshulman: Well, it's at least a traditional name in dependent type theory, right?  Although I guess it's more common to define as an indexed family than as a function type.  But yeah, it might be confusing if we start also doing linear algebra.  (-:  Any other suggestions?

mikeshulman: I'm curious why you decided to define them this way rather than as an inductive family?  It seems like the latter would use less funext, for one thing.

theories/Spaces/Finite/Fin.v
andreaslyn: It is because I want the definitional equality from `operation_curry`: https://github.com/HoTT/HoTT/blob/9e8e09146747077eb81c6f152bc5f04b4d7cd5f5/theories/Algebra/Universal/Operation.v#L141

theories/Spaces/Finite/Vect.v
andreaslyn: I think that `Vect` is a traditional name in dependent type theory, but perhaps not suitable here. An alternative is `FinSeq` for finite sequence.

In the multi-sorted universal algebra that I am developing, this type occurs as the of sorts for the domain of finitary operations: https://github.com/HoTT/HoTT/blob/9e8e09146747077eb81c6f152bc5f04b4d7cd5f5/theories/Algebra/Universal/Operation.v#L77

In my initial approach I defined the inductive family and proved the two equivalent, but I do not need the inductive family, so I decided to leave it out.

andreaslyn: I use funext to derive the induction principle for the type, which I just included for completeness. I do not expect to need that for universal algebra.

mikeshulman: `FinSeq` could be good.  Why is this definition preferable to the inductive one?

andreaslyn: I think the inductive family is often preferable, but in my particular case I need this one to instantiate the `SymbolTypeOf` record: https://github.com/HoTT/HoTT/blob/bce95e43fe3407706ac0eae1bd9ddaf7d42fb2b9/theories/Algebra/Universal/Algebra.v#L23 when `Arity` is `Fin n`.

I am fine with having both notions of `FinSeq`. The "default" inductive definition in `HoTT.Spaces.FinSeq` and this one in `HoTT.Spaces.Finite.FinSeq`. What do you think?

mikeshulman:
Sure.  No need to add the inductive one until someone needs or wants it.

andreaslyn:
I have rebased and changed `Vect` to `FinSeq`.

Alizter:
The universe issues here are interesting. We can fix some issues by changing `Fin` in `Finite/Fin.v` to the following:
```coq
Fixpoint Fin (n : nat) : Type0
  := match n with
       | 0 => Empty
       | S n => Fin n + Unit
     end.
```
It makes sense to stick it in the lowest universe, just like we do with `nat`. So `Fin` doesn't depend on universes.

This let's us reduce the universes in `FinSeq` as a result. In general I would opt for:
```coq
Definition FinSeq@{u} (n : nat) (A : Type@{u}) : Type@{u} := Fin n -> A.
```
This means that `FinSeq` only has one universe.

This should reduce some of the blowup I observed.

andreaslyn:
Good idea. I have added those changes.

andreaslyn:
Are we OK with merging this?

mikeshulman:
Looks fine to me.

PR #1429, opened 2021-03-25
Remove coq prelude

Alizter:
This PR is a work in progress of removing the coq prelude from the HoTT library.

At the moment the coq prelude has been completely removed with just a boilerplate Prelude.v file that remains.

There is still much more work to do, but I would like some feedback on what has been done so far:

* All types are defined in `Basics` now.
* Most notations (except for a few conflicting ones) are reserved in Basics.Notations avoiding level clashes.
* Overture.v has been partially reorganized
* Most of the tactics have been moved into Basics.Tactics
* Lemmas about nat have been collected in Spaces.nat whilst the definition is now in Overture.v

The following still needs to be done/considered:
* Renaming `list`, `nat`, `option`, `prod` to match our style
* Begin using the `-noinit` flag with `coqc` and generally simplify the makefile process (this is a big one)
* Remove excess code. We have a lot of unused code inherited from our fragment of the coq prelude, mostly tactic stuff. We should look into cleaning this up.
* Possibly split up Overture even further? I feel like we could have a separate file for all the coq flags so as not to clutter the theory developed in Overture. This would also have the advantage of not scaring new users since Overture is likely one of the first files seen.
* Update the documentation that exists here.
* Organise Spaces.Nat as a lot of things have been copied with little thought into structure. We can organise it like Spaces.Pos or Spaces.Int.


theories/Algebra/Universal/Algebra.v
```diff
@@ -158,9 +158,8 @@ Defined.
 Arguments path_path_algebra {_} {σ} {A B}%Algebra_scope (p q r)%path_scope.
 
 Module notations_algebra.
-
+  (** Conflicts with invers notation so cannot be reserved *)
```
mikeshulman, line 161:
```suggestion
  (** Conflicts with inverse notation so cannot be reserved *)
```

theories/Basics/Decimal.v
```diff
@@ -1,12 +1,4 @@
-(************************************************************************)
-(*         *   The Coq Proof Assistant / The Coq Development Team       *)
-(*  v      *   INRIA, CNRS and contributors - Copyright 1999-2018       *)
-(* <O___,, *       (see CREDITS file for the list of authors)           *)
-(*   \VV/  **************************************************************)
-(*    //   *    This file is distributed under the terms of the         *)
-(*         *     GNU Lesser General Public License Version 2.1          *)
-(*         *     (see LICENSE file for the text of the license)         *)
-(************************************************************************)
```
mikeshulman, line 9: Since these files are being essentially copied verbatim from the Coq prelude, I wonder if we should retain the original copyright notice, perhaps slightly modified?

theories/Basics/Notations.v
```diff
@@ -1,9 +1,88 @@
...
+
 
 (** Work around bug 5569, https://coq.inria.fr/bugs/show_bug.cgi?id=5569, Warning skip-spaces-curly,parsing seems bogus *)
 Local Set Warnings Append "-skip-spaces-curly".
 
+(** ML Tactic Notations *)
+Declare ML Module "ltac_plugin".
+Global Set Default Proof Mode "Classic".
+
+(** These are the notations whose level and associativity are imposed by Coq *)
```
mikeshulman, line 15: What does "imposed" mean here?

mikeshulman:
Looks fine to me overall.

spitters:
Thanks for this undertaking. I believe it will simplify things quite a bit. Looks good overall.

theories/Basics/Utf8.v
```diff
@@ -1,5 +1,18 @@
...
+  (at level 200, x binder, y binder, right associativity).
+Reserved Notation "'∃'  x .. y , P"
+  (at level 200, x binder, y binder, right associativity).
+Reserved Notation "'λ'  x .. y , t"
+  (at level 200, x binder, y binder, right associativity).
+
+Reserved Notation "x ∧ y" (at level 80, right associativity).
+Reserved Notation "x → y" (at level 99, y at level 200, right associativity).
+Reserved Notation "x ↔ y" (at level 95, no associativity).
+(*Notation "¬ x" (at level 75, right associativity).*)
```
andreaslyn, line 13: Why are some notations commented out?

andreaslyn (commented):
Nice work! I think it is best to get this merged before other major changes, such as renaming `list` or splitting Overture further.

theories/Basics/Notations.v
Alizter: This is from our copy of the prelude, I am assuming that Andrej wrote them way back when. It's probably not relevant now.

coq/theories/Init/Prelude.v
```diff
@@ -1,28 +1 @@
...
-Number Notation Number.uint Number.uint_of_uint Number.uint_of_uint
-  : dec_uint_scope.
-Number Notation Number.int Number.int_of_int Number.int_of_int
-  : dec_int_scope.
-
-(* Parsing / printing of [nat] numbers *)
-Number Notation nat Nat.of_num_uint Nat.to_num_uint (abstract after 5001) : nat_scope.
-
-Add Search Blacklist "_admitted" "_subproof" "Private_".
+(** I am empty for a reason *)
```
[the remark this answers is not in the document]
Alizter: The current makefile set up expects this file to exist. If I change this now, there is a whole rabbit hole to go down so I will leave it like this for now.

theories/Basics/Utf8.v
Alizter: They were commented out before. We can add them back in if they don't break anything.

coq/theories/Init/Prelude.v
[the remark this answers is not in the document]
mikeshulman:
```suggestion
(** This file is empty; it only exists because the current makefile expects it to. *)
```

theories/Basics/Utf8.v
andreaslyn: OK I was just curious. Not a big deal I think.

theories/Basics/Notations.v
Alizter: But I think it used to be the case that coq actually expected these notations to exist to work correctly. This is no longer the case.

theories/Algebra/Universal/Algebra.v
```diff
@@ -158,9 +158,8 @@ Defined.
 Arguments path_path_algebra {_} {σ} {A B}%Algebra_scope (p q r)%path_scope.
 
 Module notations_algebra.
-
+  (** Conflicts with inverse notation so cannot be reserved *)
```
andreaslyn, line 161: Right! I will change this notation soon, so we can reserve it.

theories/Classes/theory/ua_subalgebra.v
```diff
@@ -127,6 +127,7 @@ Section subalgebra.
 End subalgebra.
 
 Module subalgebra_notations.
+  (** Cannot reserve since it messes with sigma type notation *)
```
andreaslyn, line 130: When I am at it, I will change this notation as well.

theories/Basics/Decimal.v
Alizter: Here is a Zulip discussion about the license:
https://coq.zulipchat.com/#narrow/stream/237977-Coq-users/topic/Question.20on.20licensing
I am being told that our BSD license is incompatible with the coq GPL one found in those files. It is suggested that we just rewrite those files, but this seems kind of pointless, since I would just end up writing the exact same thing.

andreaslyn (approved):

theories/Classes/theory/ua_subalgebra.v
Alizter: Yep, these should be the only two notations I was unable to reserve.

Alizter:
OK, we need to resolve the issues with licensing before we merge, however as of now this discussion is still on going.

Alizter:
@spitters I have added the copyright banner back in for Decimal.v and Hexadecimal.v with a little note stating that we have modified those files for use in this library. Currently our BSD-2 license states that:
```
The HoTT library is distributed under the BSD 2-clause licence, included below, except for
files which contain a specific copyright notice and licencing information. The members of
the HoTT development team and their contributions are described in the file CREDITS.txt.
```
I take this to mean we can do as I have done.

We also have some of the authors of this file giving their permission in the Zulip thread:

https://coq.zulipchat.com/#narrow/stream/237977-Coq-users/topic/Question.20on.20licensing

So I will merge this when it is green. Licensing can always be changed at a later date to be more coherent, but hopefully this resolves some of the issues that @mikeshulman raised.

spitters:
@Alizter with f2a298d you seem to be strictly improving the current status quo and we seem to be inline with the spirit of the authors of those files. LGTM

PR #1433, opened 2021-03-31
Added logical path for contrib

Alizter:
We are currently trying to make a coq platform release. In order to do this, we need to give a logical path to the contrib folder or else it gets installed weirdly.


Alizter:
CI is green, merging without review since we need this upstream

PR #1436, opened 2021-04-02
Predicate subgroups

Alizter:
This is a WIP attempt at switching to predicates for subgroups as discussed in #1387.

This should have a few advantages when it comes to formalization.

I've also implemented a basic hint database for groups, which can automate a lot of routine arguments that we had. With some more careful design we can make this a powerful and useful tool.

This should also resolve issues we were having with #1369 though I haven't been able to confirm this yet.

When it comes to actually constructing subgroups, the predicate approach is far more efficient. There is no need to construct another group and check that the inclusion map is a homomorphism, etc.

PR #1440, opened 2021-04-02
auto for groups

Alizter:
We create a new hint database for basic group lemmas and introduce a new tactic `grp_auto` that uses them.

An example of where this is useful is the construction of the direct product. If you find any other places this could be used let me know!

We can of course make this tactic much smarter with rewriting etc. but I am unsure how useful this will really be. It might also be worth noting that if a group has a decidable word problem, then we ought to have a decision procedure to solve equalises for us. Again, unsure if this is worth it.

This comes from #1436.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -31,8 +31,8 @@ Arguments group_sgop {_}.
 Arguments group_unit {_}.
 Arguments group_inverse {_}.
 Arguments group_isgroup {_}.
-(** We never need to unfold the proof that something is a group *)
-Opaque group_isgroup.
+(** We should never need to unfold the proof that something is a group. *)
+Global Opaque group_isgroup.
 
 (** We coerce groups back to types. *)
 Coercion group_type : Group >-> Sortclass.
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Group.v
```diff
@@ -471,32 +501,9 @@ Proof.
...
-  { intros [g h].
-    apply path_prod; cbn.
-    1,2: apply left_inverse. }
-  (** Right inverse *)
-  { intros [g h].
-    apply path_prod; cbn.
-    1,2: apply right_inverse. }
+  repeat split.
+  1: exact _.
+  all: grp_auto.
```
andreaslyn, line 506: Cool

andreaslyn (commented):
LGTM

jdchristensen (approved):

theories/Algebra/Groups/Group.v
```diff
@@ -41,6 +41,33 @@ Global Existing Instances group_sgop group_unit group_inverse group_isgroup.
...
+#[export] Hint Immediate grp_law_right_identity : group_db.
+Local Definition grp_law_left_inverse {G : Group} (x : G) := left_inverse x.
+#[export] Hint Immediate grp_law_left_inverse : group_db.
+Local Definition grp_law_right_inverse {G : Group} (x : G) := right_inverse x.
+#[export] Hint Immediate grp_law_right_inverse : group_db.
+
+(** Given path types in a product we may want to decompose. *)
+#[export] Hint Extern 5 (@paths (_ * _) _ _) => (rapply path_prod) : group_db.
+(** Given path types in a sigma type of a hprop family (i.e. a subset) we may want to decompose. *)
+#[export] Hint Extern 6 (@paths (sig _) _ _) => (rapply path_sigma_hprop) : group_db.
```
Alizter, line 65: TODO: Think about these

Alizter: Is `rapply` necessary?

PR #1444, opened 2021-04-08
Absolutize require imports

Alizter:
In testing the Coq Platform release of the HoTT library, we have come across the issue that somebody who wants to step through a file cannot without recompiling. This is something we didn't notice before because we develop in the HoTT library itself.

It would therefore be beneficial for external users to be able to do this.

One solution is to kill all ambiguity with the `Require Imports`. I've gone ahead and run Jason's [absolutize-imports script](https://github.com/JasonGross/coq-tools/blob/master/absolutize-imports.py) and this has in turn decorate the whole library with full logical names.


mikeshulman:
Ugh.  You said "one solution"; are there others?

Alizter:
The third solution is to get the semantics of coq to work, but I am unsure of the technical details that would be needed for this to happen. @MSoegtropIMC and @Zimmi48 were having a discussion on Zulip about ways this might be done.

Alizter:
I am not too fond of this change myself, but in guise of the workshop next week, it would be beneficial to people who want to get familiar with the library to be able to step through already compiled files.

mikeshulman:
I don't understand the issue.  What is different between locally compiled files and distributed compiled ones?

Alizter:
This is what @MSoegtropIMC wrote earlier on Zulip:
>one note: as a test I did put a _CoqProject file with the proper options into the Coq Platform install folder of HoTT. This allows people to open and run HoTT files from the install folder in CoqIDE - which can be a great help to understand things. But this does not work for all files, because some files reference HoTT files without the HoTT prefix. It would help users to change all files such that this does work. As examples:
>
>Does not work: Homotopy/Join.v
Does work: ExcludedMiddle.v
It is currently not possible to have dynamic absolute paths in _CoqProject files nor is it possible to specify relative paths in -Q and -R options, so there is no easy other fix than changing all files to consistently use Require Import Hott.XYZ.

mikeshulman:
Where is the zulip discussion?  i can't find it.

Alizter:
I've marked this as draft for now as there are a few things left to fix.

The zulip discussion is here:
https://coq.zulipchat.com/#narrow/stream/250632-Coq-Platform.20devs.20.26.20users/topic/Coq-HoTT.20in.20the.20platform.3F

mikeshulman:
Thanks.  I always forget that zulip only shows the streams on the left that i've "subscribed" to.

PR #1445, opened 2021-04-08
Make module references in Require statements absolute

MSoegtropIMC:
This PR makes all module references absolute, so that it is easy to step through files with an installed HoTT.

This change was done with a script by @JasonGross with the following commands from a HoTT git folder:
```
# build to have dependency information
git clean -dxf
coq_makefile -f _CoqProject -o CoqMakefile
make -f CoqMakefile -j 16
# patch requires
python2 ../coq-tools/absolutize-imports.py --update-all -f _CoqProject
# test if it still builds
git clean -dxf
coq_makefile -f _CoqProject -o CoqMakefile
make -f CoqMakefile -j 16
```
where the tools from (https://github.com/JasonGross/coq-tools) have been cloned into a sibling folder of the HoTT folder.


MSoegtropIMC:
Duplicate of #1444

jdchristensen:
While so many imports are changing, is there also a tool that checks for unneeded imports?

andreaslyn:
I have taken a quick journey through this. Towards the end I had to load the diffs manually. LGTM.

MSoegtropIMC:
> While so many imports are changing, is there also a tool that checks for unneeded imports?

I don't think this can be derived from the dependency information created during compilation. But it isn't that hard to make a tool which simply removes each require and test compiles. HoTT compiles pretty fast, so that this shouldn't take that long.

MSoegtropIMC:
Btw.: @mattam82 mentioned in the zulip discussion on this topic that other build systems might not be able to handle the explicit HoTT prefix. I tested this only with the `coq_makefile` based method from the opam file.

Alizter:
It would be nice to be able to minimise requires up to a point though. I think we previously discussed that it is generally good to assume the whole of Basics and Types are imported. See #1054

MSoegtropIMC:
@JasonGross : really a nice collection of tools!

@Alizter : I guess it wouldn't be hard to add a list of modules to not remove e.g. via a Json file.

jdchristensen:
I have no objections to merging this now.  But if someone wants to run that tool to check for unused imports, it would make sense to do both sets of changes at once, since these changes might create some merge conflicts with other work.  OTOH, if nobody wants to try this, this suggestion shouldn't hold up going ahead with this PR if it is otherwise fine.

MSoegtropIMC:
Let me run the tool.

Also is it clarified if all ways to build HoTT (dune ...) can handle this?

mikeshulman:
The conclusion I took away from the zulip discussion was that we can leave things as they are for now, and hope that the `From HoTT` syntax for requires can be improved so that we can write unambiguous imports without so much excess verbiage.

andreaslyn:
Let us close this for now.

MSoegtropIMC:
It would be good to solve this before the next HoTT school, though - in one way or another.

PR #1446, opened 2021-04-12
Removed eisadj_other + some optimizations

### LINKED ISSUE #1282: Make eisadj_other transparent
Originally raised by @jdchristensen at #1091:

If `eisadj_other` in Equivalences.v (formerly `other_adj`) is made transparent, then coq spins while checking Card.v. The problem is in the call to `strip_truncations` in the `reduce` tactic during the proof of `exp_zero_card`.  If we can fix this, then we can inline the proof of `eisadj_other` into `isequiv_inverse`, and use `eisadj (f^-1)` in place of `eisadj_other`.

Alizter:
I've inlined the proof of `eisadj_other` as requested in #1282.

I've also gone ahead and made `eisadj` opaque. This will tell coq to delay unfolding it as long as possible when simplifying. This will stop equivalences from turning into super large terms. Note that marking a term as Opaque is different from being Qed, as Qed will never be unfolded.

Doing this unfortunately caused ClassifiyngSpace.v to blow up. I have therefore replaced a large simplification with a manual one which has saved quite bit of time.

```
   After |  Peak Mem | File Name                                                    |   Before |  Peak Mem ||    Change || Change (mem) | % Change | % Change (mem)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
5m39.28s | 585436 ko | Total Time / Peak Mem                                        | 6m38.31s | 666804 ko || -0m59.02s ||    -81368 ko |  -14.81% |        -12.20%
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
0m17.77s | 583772 ko | Homotopy/ClassifyingSpace.vo                                 | 0m28.22s | 666804 ko || -0m10.44s ||    -83032 ko |  -37.03% |        -12.45%
```

Finally, in Card.v (where the problems first began) I noticed that the tactic `simpl_ops`, was unfolding the term `exp_card` which was unnecessary and costly. It is much faster now.

closes #1282

### DIFF
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -532,9 +532,10 @@ Definition Book_4_2_1 := @HoTT.Basics.Overture.IsEquiv.
 (* ================================================== lem:coh-equiv *)
 (** Lemma 4.2.2 *)
 
-Definition Book_4_2_2 := fun A B f isf x =>
+(* The proof of Lemma 4.2.2 is embedded in the proof of isequiv_inverse. *)
+Definition Book_4_2_2 := fun (A B : Type) (f : A -> B) (feq : IsEquiv f) (y : B) =>
   @HoTT.Basics.Overture.eisadj B A f^-1
-    (@HoTT.Basics.Equivalences.isequiv_inverse A B f isf) x.
+    (@HoTT.Basics.Equivalences.isequiv_inverse A B f feq) y.
 
 (* ================================================== thm:equiv-iso-adj *)
 (** Theorem 4.2.3 *)
@@ -2337,4 +2338,3 @@ Definition Book_11_6_17_iv := @HoTT.Spaces.No.Core.lt_le_trans.
 (** Example 11.6.18 *)
 
 Definition Book_11_6_18 := @HoTT.Spaces.No.Addition.plus.
-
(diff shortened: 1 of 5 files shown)


mikeshulman (approved):
Thanks!

contrib/HoTTBook.v
```diff
@@ -532,7 +532,9 @@ Definition Book_4_2_1 := @HoTT.Basics.Overture.IsEquiv.
 (* ================================================== lem:coh-equiv *)
 (** Lemma 4.2.2 *)
 
-Definition Book_4_2_2 := @HoTT.Basics.Equivalences.eisadj_other.
+Definition Book_4_2_2 := fun A B f isf x =>
+  @HoTT.Basics.Overture.eisadj B A f^-1
+    (@HoTT.Basics.Equivalences.isequiv_inverse A B f isf) x.
```
jdchristensen, line 537: Maybe it's worth helping the reader parse this by giving the types of the arguments and using the same variable name (y) as in the HoTT book:
```suggestion
(* The proof of Lemma 4.2.2 is embedded in the proof of isequiv_inverse. *)
Definition Book_4_2_2 := fun (A B : Type) (f : A -> B) (feq : IsEquiv f) (y : B) =>
  @HoTT.Basics.Overture.eisadj B A f^-1
    (@HoTT.Basics.Equivalences.isequiv_inverse A B f feq) y.
```
Coq can also infer a lot of that, but I guess it's worth being explicit as a way to point the reader to where the proof is.

jdchristensen (approved):
Nice!  For some reason, in my test with `-j4`, the build was only 5% faster.  But that's still good.

Alizter:
Rebasing since githubs conflict resolution does a merge

PR #1449, opened 2021-04-12
added 5.8.4 (iv) -> (v)

### LINKED ISSUE #1036: Do we have HoTT 5.8.4
It is not in the HoTT book file, but more specifically do we have a lemma saying that if we have the equivalence in 5.8.4 (iv) then we have the contractility in (v)? I couldn't find anything in Types.Sigma or Types.Path...

Alizter:
closes #1036

### DIFF
--- a/theories/Types/Sigma.v
+++ b/theories/Types/Sigma.v
@@ -756,3 +756,19 @@ Proof.
   intros a v.
   exact (trunc_equiv' _ (hfiber_functor_sigma_idmap _ _ _ _ _)).
 Defined.
+
+(** Theorem 5.8.4 (iv) -> (v) *)
+(** Contractibility of the total space of a reflexive relation is equivalent to asking that transport induces an equivalence between paths and the relation. *)
+Definition contr_sigma_refl_rel {A B : Type}
+  (R : A -> A -> Type) (r0 : forall a, R a a)
+  : (forall a b, IsEquiv (fun p : a = b => transport (R a) p (r0 a)))
+  -> forall a, Contr (sig (R a)).
+Proof.
+  intros f a.
+  rapply contr_equiv'.
+  { rapply equiv_functor_sigma_id.
+    intros b.
+    nrapply Build_Equiv.
+    exact (f a b). }
+  apply contr_basedpaths.
+Defined.
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -772,7 +772,7 @@ Definition Book_5_8_2_iv_implies_iii := @HoTT.PathAny.equiv_path_from_contr.
 (* ================================================== thm:ML-identity-systems *)
 (** Theorem 5.8.4 *)
 
-
+Definition Book_5_8_4_iv_implies_v := @HoTT.Types.Sigma.contr_sigma_refl_rel.
 
 (* ================================================== thm:equiv-induction *)
 (** Corollary 5.8.5 *)


mikeshulman (approved):

theories/Types/Sigma.v
```diff
@@ -756,3 +756,19 @@ Proof.
   intros a v.
   exact (trunc_equiv' _ (hfiber_functor_sigma_idmap _ _ _ _ _)).
 Defined.
+
+(** Theorem 5.8.4 (iv) -> (v) *)
+(** Contractibility of the total space of a reflexive relation is equivalent to asking that transport induces an equivalence between paths and the relation. *)
```
jdchristensen, line 761: The comment says "is equivalent to" but only one direction is proved.

theories/Types/Sigma.v
```diff
@@ -756,3 +756,19 @@ Proof.
...
+Definition contr_sigma_refl_rel {A B : Type}
```
jdchristensen, line 762: `B` is unused, right?

theories/Types/Sigma.v
```diff
@@ -756,3 +756,19 @@ Proof.
   intros a v.
   exact (trunc_equiv' _ (hfiber_functor_sigma_idmap _ _ _ _ _)).
 Defined.
+
+(** Theorem 5.8.4 (iv) -> (v) *)
```
jdchristensen, line 760: I think it would be better to prove the corresponding implication in 5.8.2, which proves the implication separately for each `a : A`.  This is the converse of `equiv_path_from_contr` in PathAny.v, and might be a more natural home for this?

theories/Types/Sigma.v
Alizter: Ah yes, I didn't bother to do the other direction.

theories/Types/Sigma.v
Alizter:
```suggestion
Definition contr_sigma_refl_rel {A : Type}
```

theories/Types/Sigma.v
Alizter:
```suggestion
(** If transport between paths and a reflexive relation is an equivalence then the total space of the relation is contractible. *) 
```

mikeshulman: The comments should probably mention somewhere that the two statements *are* equivalent, even if we're only proving one direction here.

jdchristensen: Just to be clear, the other direction is already in the library, as `equiv_path_from_contr`, in the stronger form like 5.8.2.   (It's stronger because it says `forall a, (P a -> Q a)` instead of `(forall a, P a) -> (forall a, Q a)`, and moreover allows the relation to only be defined at `a`.)  So I hope we phrase this direction in the corresponding way, to have both directions in that form.  Of course, it's also good to state the special case in the form 5.8.4 as well.

Alizter: I don't think that is true. `equiv_path_from_contr` asserts that it is sufficient for *any* equivalence to exist, so I don't think the converse is provable. On the other hand, 5.8.4 asserts that a specific map is an equivalence, hence a hprop thus why we get an equivalence in the end.

jdchristensen: It's true that `equiv_path_from_contr` could be strengthened to say that the equivalence is given by transport, but I think this follows by path induction anyways.  But my main point is that the result you proved can be strengthened by proving the implication separately for each `a : A`.  This works:

```coq
Definition contr_sigma_refl_rel' {A : Type}
  (a : A) (R : A -> Type) (r0 : R a)
  : (forall b, IsEquiv (fun p : a = b => transport R p r0))
  -> Contr (sig R).
Proof.
  intros f.
  rapply contr_equiv'.
  { rapply equiv_functor_sigma_id.
    intros b.
    nrapply Build_Equiv.
    exact (f b). }
  apply contr_basedpaths.
Defined.
```
With `a` factored out like this, you see how close it is to
```coq
Definition equiv_path_from_contr {A : Type} (a : A) (P : A -> Type)
           (Prefl : P a)
           (cp : Contr {y:A & P y} )
           (b : A)
  : P b <~> a = b.
```
with some minor differences in notation.  But, even better, you can strengthen and simplify your result even more by just assuming that you have *any* equivalence:
```coq
Definition contr_sigma_refl_rel'' {A : Type}
  (a : A) (R : A -> Type) (r0 : R a)
  (f : forall b, (a = b) <~> R b)
  : Contr (sig R).
Proof.
  rapply contr_equiv'.
  1: exact (equiv_functor_sigma_id f).
  apply contr_basedpaths.
Defined.
```

Alizter: Oh I see now! That is cool. I'll get rid of my lemma then tomorrow, and include that in PathAny.v as an inverse to the existing lemma there.

jdchristensen: I think it makes sense to have the 5.8.2 version and then use it to prove the 5.8.4 version.  And while I don't see much use in having "transport is an equivalence" as a hypothesis (compared to the more general "there is an equivalence"), it does seem like it could be handy to have it as a conclusion, if you felt like making the conclusion of `equiv_path_from_contr` more precise.

mikeshulman:
Is this PR still waiting for anything?

Alizter:
@mikeshulman No, I just didn't have the time to work on it again. I'll mark it as draft for now.

Alizter:
Not going to finish this anytime soon so closing.

PR #1450, opened 2021-04-12
standardize naming of istrunc_* lemmas

### LINKED ISSUE #1226: Naming of trunc_sigma, trunc_forall etc.
We have some old names for these lemmas. Wouldn't it make more sense for them to be called istrunc_sigma, istrunc_forall. Are there any others?

Alizter:
I've tried my best to standardise the `trunc_*` lemmas. I may have missed a few, but I guess we can change them when we come across them. They should now be `istrunc_*`.

 * In `Types/` you will find the obvious changes like `trunc_forall` to `istrunc_forall` for each type.
 * We change `trunc_equiv` to `istrunc_isequiv_istrunc` and `trunc_equiv'` to `istrunc_equiv_istrunc`. This is how the corresponding modality lemma is named, as mentioned in #1226.
 * In renaming `trunc_equiv` I have occasionally swapped which lemma is being used, this ought to save some typeclass search.

closes #1226

### DIFF
--- a/theories/Spaces/No/Core.v
+++ b/theories/Spaces/No/Core.v
@@ -903,9 +903,7 @@ Section NoCodes.
   Proof.
     revert x y.
     refine (No_ind_hprop _ _); intros L R ? xL xR xcut xHL xHR.
-    (** TODO: Why can't Coq find [trunc_arrow] here? *)
-    refine (@No_ind_hprop _ _
-              (fun y => @trunc_prod _ _ trunc_arrow _ trunc_arrow) _).
+    refine (@No_ind_hprop _ _ _ _).
     intros L' R' ? yL yR ycut yHL yHR. split.
     - intros x_le_y.
       rewrite le'_cut in x_le_y.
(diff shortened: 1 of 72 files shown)


Alizter:
rebased

theories/Spaces/No/Core.v
```diff
@@ -903,9 +903,7 @@ Section NoCodes.
   Proof.
     revert x y.
     refine (No_ind_hprop _ _); intros L R ? xL xR xcut xHL xHR.
-    (** TODO: Why can't Coq find [trunc_arrow] here? *)
-    refine (@No_ind_hprop _ _
-              (fun y => @trunc_prod _ _ trunc_arrow _ trunc_arrow) _).
+    refine (@No_ind_hprop _ _ _ _).
```
mikeshulman, line 906: Hmm, I wonder what changed here?

mikeshulman (approved):

theories/Spaces/No/Core.v
andreaslyn: Type class search has probably improved after this was made.

andreaslyn:
LGTM

Issue #1435, opened 2021-04-01
Fancier documentation


JasonGross:
The [proviola docs](http://hott.github.io/HoTT/proviola-html/toc.html) are nice, but I wonder if we want more modern documentation.  For example, with [Alectryon](https://plv.csail.mit.edu/blog/alectryon.html), we could get pretty-printed MathJax displays:
![image](https://user-images.githubusercontent.com/396076/113332142-c92d6800-92ee-11eb-908e-ee02582d2880.png)

spitters:
I think that would be nice. Proviola is showing it's age...

PR #1453, opened 2021-04-13
Excise the alternate standard library, use -noinit

JasonGross:
We now use `-noinit` rather than passing `-coqlib`.  I've also fixed
some of the previously dead paths such as `README.markdown`, and updated
various files to know about the recent update of binding `contrib` to
`HoTT.Contrib` rather than `""`.

N.B. I haven't tested the install target, so there's a chance I broke
something.

This will make it easier for me to conceptualize how to do https://github.com/HoTT/HoTT/issues/1435, since we will now look for Coq plugins (such as coq-serapi) in the standard place rather than in some custom location.


Alizter:
Why did we have `coq/tools/coqdoc/coqdoc.css`?

Alizter:
We can also use the `coq_makefile` install target instead. How difficult would it be to untangle the makefile and use coq_makefile instead?

I should note that coq_makefile is perfectly capable of building and installing the library, and that we even have an opam release that works this way.

JasonGross:
> Why did we have `coq/tools/coqdoc/coqdoc.css`?

I think it's because `coqdoc` errors without access to this file (which it copies over into the html directory), and at this point I hadn't set up `configure.ac` to be clever enough at copying things.  Note that we also have [`etc/coqdoc.sty`](https://github.com/HoTT/HoTT/blob/master/etc/coqdoc.sty), which we can probably remove.

> We can also use the `coq_makefile` install target instead. How difficult would it be to untangle the makefile and use coq_makefile instead?

Not too difficult, I think, maybe a couple of hours of work?  I don't want to do it in this PR, though.  I think there are a number of questions such as "do we still want to have `hoqc`?" and "do we still want to install LICENSE.txt CREDITS.txt INSTALL.md README.md (and if so where do they go?)".  The changes here will not make it any harder than it currently is to untangle the Makefile.

Alizter:
> > Why did we have `coq/tools/coqdoc/coqdoc.css`?
> 
> I think it's because `coqdoc` errors without access to this file (which it copies over into the html directory), and at this point I hadn't set up `configure.ac` to be clever enough at copying things. Note that we also have [`etc/coqdoc.sty`](https://github.com/HoTT/HoTT/blob/master/etc/coqdoc.sty), which we can probably remove.

OK, I'll make an issue about removing that.

> > We can also use the `coq_makefile` install target instead. How difficult would it be to untangle the makefile and use coq_makefile instead?
> 
> Not too difficult, I think, maybe a couple of hours of work? I don't want to do it in this PR, though. I think there are a number of questions such as "do we still want to have `hoqc`?" and "do we still want to install LICENSE.txt CREDITS.txt INSTALL.md README.md (and if so where do they go?)". The changes here will not make it any harder than it currently is to untangle the Makefile.

That's fine not to do it in this PR. I've made an issue #1457 for changing the makefile. I've created #1458 regarding the hoq scripts. I am unsure about installing the documentation, ideally we want to do it the same way other coq projects do.

Alizter:
Regarding the install, it looks like it works fine from looking at the CI.

PR #1462, opened 2021-04-15
Add problem matchers for GH Actions

JasonGross:
This will display warnings somewhat prominently in PRs, idk if we want
this.


mikeshulman:
If the diff for this PR is an example of what this would look like, it looks more annoying than helpful.  Most of the warnings shown are not relevant to the PR in question.

JasonGross:
Indeed, the idea would be to fix or suppress all existing warnings, and then these matchers would show us only the warnings introduced by a new PR.  Of course, if we merged the PR without fixing the warnings, they'd then show up on every subsequent PR.  (Granted, the warnings unrelated to the PR show up below all the files modified by the PR.)  Do you think it's worth fixing the hint locality warnings, suppressing the alectryon warnings, and then merging something like this?  Or would it not be useful?

mikeshulman:
Well, we should fix the hint locality warnings in any case.  I don't have a good sense for whether there are "real" warnings that would be useful to see when reviewing a PR.  Do you have any in mind?

JasonGross:
I think the most useful thing this would bring is that when the CI fails with an error, we don't need to go digging into the logs to find the error, and instead it will show up prominently at that location.  (If you want, I can make the matchers only catch errors and not warnings, and induce one artificially to demonstrate?  Alternatively, I can write some complicated script that will check whether or not the warning/error is in a file that was touched by the current PR/commit and only display it in the interface in that case, but I think that will be a moderately more complicated task.)  I think the warnings are mostly useful for code hygiene; since we don't frequently go digging though the build logs, this will let us know more prominently when some warning is triggered and motivate us to fix it sooner.  There's a sense in which none of Coq's warnings are any more "real" than the hint locality ones, though I suppose there are ones that are occasionally useful (the one coming to mind is the relatively new warning about having a named and unused variable which catches multiple cases in a pattern match, hinting that you might have misspelled a constructor name).

mikeshulman:
Having it show errors certainly does seem useful!  I suppose we could try out having it show all the warnings and see what it's like.

JasonGross:
I've made the annotations less noisy and more useful.  The Alectryon warnings are no longer displayed, we display warnings only from the master job (and display the full text of the warning, not just the first line) so there are no duplicates, though we display errors from all jobs.  I think this is ready for merge if it looks good.  (I'll fix the existing warnings in another PR.)

Alizter (approved):

mikeshulman:
Sure, let's give it a try.

Zimmi48:
@JasonGross I believe that this PR has broken Coq's CI. We get `/bin/bash: etc/coq-scripts/github/reportify-coq.sh: No such file or directory`. This might be because we build with `./autogen.sh -skip-submodules`. IMHO you should keep supporting this use case.

Zimmi48:
I didn't realize the issue had already been located (and now the fix is merged).

andreaslyn (commented):
LGTM

PR #1464, opened 2021-04-16
Silence some warnings

JasonGross:


Alizter:
These don't seem to be compatible with our current version of coq.

JasonGross:
This apparently needs to wait until we bump the minimum version to 8.14

PR #1465, opened 2021-04-19
Fix for the Makefile when the submodule is not present

JasonGross:
Discovered at https://github.com/coq/coq/pull/13952/checks?check_run_id=2370784127 / https://github.com/coq/coq/pull/13952.


ppedrot:
IIUC this fixes the Coq CI so it's quite urgent.

Issue #1468, opened 2021-04-21
Decidable props


Alizter:
I'm currently trying to do some cleanup with the nat library but I am running into a bit of a problem: I don't know if I should be using DHProp. I see that previously discussions were had  in #654 and #683 about introducing decidable props for propositions. My opinion on the matter has seesawed somewhat so I would like to get a discussion going here on what to do.

Here is a common workflow when it comes to formalizing mathematics (I believe this is how ssreflect does it too):
 1. Define a proposition (say leq for nat as an inductive type).
 2. Define a boolean version of this proposition.
 3. Show that it is equivalent (in SSRs case that it "reflects").
 4. Get decidability from this

I've done this numerous times for various developments, but it always looks like I've done twice the amount of work I really need to.

Compare this with the DHProp version (which I think is happening):
 1. Define a DHProp for whatever it is you are doing
 2. It is automatically a hprop and decidable

Now this workflow looks nice, but I don't think it actually solves any problems. In fact, I don't see the difference between just working with booleans directly, since you can always coerce those into hprops.

I also don't like the fact that DHProp types actually compute. Sure this is neat where you can normalize `5 <= 7` and get a prop you know the answer to, but this is a nightmare when working in proofs. Sometimes you want to simplify a term, but coq simplifies the DHProps so things don't typecheck immediately (or at least take a long time too).

I feel as though it makes more sense to have a more rigid type, and this reflects how we work in practice. It seems to me that decidability is a property that we show after defining a property, not whilst defining it.

So I am interested in other peoples opinions, or perhaps even if I am missing something obvious.

spitters:
Quick response. Yes, that's what ssr does.
There is also the decide equality tactic which is used in software
foundations/logical foundations.

mikeshulman:
I don't understand what you mean by "DHProps compute".  A DHProp has an underlying *type*, not a boolean.  I guess if you define a DHProp by recursion the same way you would define a boolean-valued predicate, then it would compute, but you could just as well take the underlying type of a DHProp to be inductively defined.

mikeshulman:
Would the problem you're seeing be solved with finer-grained simplification tactics that you could instruct to leave the DHProps alone?  How about declaring them Opaque or `simpl never`?

Alizter:
[...]

PR #1469, opened 2021-04-21
Clean up natural number library

### LINKED ISSUE #1367: Cleanup nat

Alizter:
<s>This PR is a WIP attempt at cleaning up the natural numbers library.

This PR isn't currently ready for review but was opened so I can get some help.

@andreaslyn a lot of my changes are breaking the FinNat and FinInduction library. Could you take a look? The last commit should contain the changes I tried, but I am unable to get some of the proofs to go through. Especially in FinInduction.v

When I finish this PR I will force push better commits then currently exist.</s>

---

OK, this is ready for review. Here is what has been done.

The organization has changed a bit:
 * `nat` is now defined in `Overture.v`.
 * There is a file `Basics/Nat.v` that contains very basic functions about `nat`. These are necessery to have since numeral notations (conversion between decimal, hexadecimal, TruncIndex etc.) use them.
 * Definitions of basic arithmetic, orders and lemmas about `nat` can be found in `Spaces.Nat` as expected.

Importantly, we no longer have the arbitrary organization of Coq.Peano / Spaces.nat.

Notably, we define `leq` as an inductive type, no longer as a DProp. There is still some discussion going on in #1468 about it, but at least we have something to experiment with.

The other changes are compatibility changes. Thanks to @andreaslyn for helping with the `Finite` library.

There are still some more things we can do:
 * Rename `nat` to `Nat`. #1470 I feel it will be too noisy to do here
 * Split up Spaces.Nat into smaller files (separating the theory about `leq` for example). Similar to the organization in `Pos` and `Int`.
 * Reorganise some theory about `nat` found in the hottclasse library. I didn't do this yet since it seemed like a rabbit hole, and it isn't used in many places.

closes #1367

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -4,35 +4,10 @@
 (** Import the file of reserved notations so we maintain consistent level notations throughout the library *)
 Require Export Basics.Notations Basics.Datatypes Basics.Logic.
 
-
 Declare ML Module "number_string_notation_plugin".
-
-
 
 (** Keywords for blacklisting from search function *)
 Add Search Blacklist "_admitted" "_subproof" "Private_".
-
-(** Some basic tactics *)
-Ltac easy :=
-  let rec use_hyp H :=
-    match type of H with
-    | _ => try solve [inversion H]
-    end
-  with do_intro := let H := fresh in intro H; use_hyp H
-  with destruct_hyp H := case H; clear H; do_intro; do_intro in
-  let rec use_hyps :=
-    match goal with
-    | H : _ |- _ => solve [inversion H]
-    | _ => idtac
-    end in
-  let rec do_atom :=
-    solve [reflexivity | symmetry; trivial] ||
-    contradiction ||
-    (split; do_atom)
-  with do_ccl := trivial; repeat do_intro; do_atom in
-  (use_hyps; do_ccl) || fail "Cannot solve this goal".
-
-Tactic Notation "now" tactic(t) := t; easy.
 
 Create HintDb rewrite discriminated.
 #[export] Hint Variables Opaque : rewrite.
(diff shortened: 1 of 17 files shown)


Alizter:
The CI issue is to do with changes to coq master and our use of the submodule in the CI.

theories/Basics/Overture.v
```diff
@@ -765,14 +680,37 @@ Global Arguments path_forall {_ A%type_scope P} (f g)%function_scope _.
...
-Hint Resolve idpath : core.
+#[export] Hint Resolve idpath inverse : path_hints.
+#[export] Hint Resolve idpath : core.
 
 Ltac path_via mid :=
   apply @concat with (y := mid); auto with path_hints.
 
+(** ** Natural numbers *)
+
+(** Unfortunately due to a bug in coq #10766 we the induction tactic fails to work properly. We therefore have to use the autogenerated induction schemes and define the ones we want to use ourselves. *)
```
andreaslyn, line 691: Can we work around this by manually defining induction? `nat_ind .. := match ..`

mikeshulman (approved):

PR #1472, opened 2021-04-22
Clean up and speed up ClassifyingSpace.v

jdchristensen:
It's probably best to review the two commits separately.

The first commit greatly improves the proof of pequiv_pclassifyingspace_pi1.
The new proof is broken into conceptual steps.  It's shorter overall, and much faster.

The second commit does some other minor cleanups and also speeds up a few things.

The hspace part of the file is untouched, and is still slow, but overall the file typechecks
in less than half the time.

### DIFF
--- a/theories/Homotopy/ClassifyingSpace.v
+++ b/theories/Homotopy/ClassifyingSpace.v
@@ -133,19 +133,24 @@ Section Eliminators.
     intros; apply path_ishprop.
   Defined.
 
+  (** Similarly, when eliminating into an hprop, we only have to handle the basepoint. *)
+  Definition ClassifyingSpace_ind_hprop (P : ClassifyingSpace G -> Type)
+    `{forall x, IsTrunc (-1) (P x)} (bbase' : P bbase)
+    : forall x, P x.
+  Proof.
+    refine (ClassifyingSpace_ind_hset P bbase' _).
+    intros; rapply dp_ishprop.
+  Defined.
+
 End Eliminators.
 
-(** We can prove that the classifying space is 0-connected. *)
+(** The classifying space is 0-connected. *)
 Global Instance isconnected_classifyingspace {G : Group}
   : IsConnected 0 (ClassifyingSpace G).
 Proof.
   exists (tr bbase).
   srapply Trunc_ind.
-  srapply ClassifyingSpace_ind_hset; cbn.
-  1: reflexivity.
-  intro x.
-  apply dp_paths_FlFr.
-  apply path_ishprop.
+  srapply ClassifyingSpace_ind_hprop; reflexivity.
 Defined.
 
 (** Now we focus on the classifying space of a group. *)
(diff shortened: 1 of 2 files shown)


jdchristensen:
One question:  While working on this, and also on unrelated things, I've been having a lot of trouble getting Coq to unfold things of the form `e^-1%equiv` where `e` is an equivalence.  If I try `simpl` or `cbn`, they both usually spin.  I've tried dozens of `unfold` commands, but can't get the inverse function to appear.  Is this due to some recent change?  It's a real problem.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -298,26 +302,25 @@ Section EncodeDecode.
...
 
-Definition grp_iso_loopgroup_bg `{Univalence} (G : Group)
-  : GroupIsomorphism (LoopGroup (B G)) G.
-Proof.
-  snrapply Build_GroupIsomorphism'.
-  1: exact equiv_loops_bg_g.
-  intros x y.
-  apply encode_pp.
-Defined.
+End EncodeDecode.
```
jdchristensen, line 323: Above here the changes are just to the whitespace and the arguments from the Context.

Alizter:
Wow the new proof of `pequiv_pclassifyingspace_pi1` is much better!

Alizter:
@jdchristensen  Regarding the issues with equiv, I suspect I broke some things with #1446. Could you try putting `Transparent eisadj` in front of any simplifications and see if it makes any difference. Also if you could provide a small example so I can do some testing that would be great.

Alizter:
Also it might be worth defining BG out of known HITs. (Maybe not in this PR). Something like
```
|| Σ(G+1) x Σ(G+1) => ΣG||_1
```
i.e. a 1-truncated coequalizer of pointed suspensions?

Alizter (approved):
LGTM am very impressed with the new speed

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -508,86 +528,23 @@ Proof.
...
-        exact _. }
-      apply equiv_istrunc_contr_iterated_loops.
-      induction n; exact _. } }
-  (** Finally we need to show that this equivalence is pointed *)
-  reflexivity.
+  (** The pointed map [f] is the adjunct to the inverse of the natural map [loops X -> Pi1 X]. We define it first, to make the later goals easier to read. *)
+  transparent assert (f : (B (Pi1 X) ->* X)).
+  { srapply pClassifyingSpace_rec.
+    1: exact (equiv_tr 0 _)^-1%equiv.
+    intros x y.
```
jdchristensen, line 538: As an example of the problem with inverses, before this `intros x y` line try `simpl` or `cbn`.  They seem to go forever.  Adding `Transparent eisadj.` before doing these doesn't help.

jdchristensen: Well, not forever.  A minute or two for `simpl`.  This happens even if the inverse of the equivalence is explicitly defined.

mikeshulman:
Looks good, thanks!  Should we go ahead and merge this and worry about speed issues elsewhere?

Issue #1454, opened 2021-04-13
Coq with native compilation enabled cannot compile the library


JasonGross:
Every time I build, I get errors such as
```
theories/Homotopy/ExactSequence.vo (real: 28.06, user: 26.58, sys: 1.23, mem: 597200 ko)
HOQC theories/Spaces/No/Addition
Fatal error: exception Stack overflow
Error: Native compiler exited with status 2

Command exited with non-zero status 1
theories/Spaces/Spheres.vo (real: 123.56, user: 117.66, sys: 5.67, mem: 2844152 ko)
make: *** [Makefile:910: theories/Spaces/Spheres.vo] Error 1
HOQC theories/Spaces/Torus/TorusEquivCircles
Fatal error: exception Stack overflow
Error: Native compiler exited with status 2

Command exited with non-zero status 1
theories/Spaces/Torus/TorusEquivCircles.vo (real: 371.11, user: 288.58, sys: 77.41, mem: 4923976 ko)
make: *** [Makefile:910: theories/Spaces/Torus/TorusEquivCircles.vo] Error 1
HOQC theories/Colimits/Colimit_Pushout_Flattening
theories/Colimits/Colimit_Pushout_Flattening.vo (real: 43.00, user: 34.00, sys: 7.96, mem: 968808 ko)
HOQC theories/Colimits/Sequential
Fatal error: exception Stack overflow
Error: Native compiler exited with status 2

Command exited with non-zero status 1
theories/Colimits/Sequential.vo (real: 114.35, user: 103.08, sys: 10.99, mem: 2181364 ko)
make: *** [Makefile:909: theories/Colimits/Sequential.vo] Error 1
HOQC theories/Homotopy/CayleyDickson
Fatal error: exception Stack overflow
Error: Native compiler exited with status 2

Command exited with non-zero status 1
theories/Homotopy/CayleyDickson.vo (real: 74.73, user: 61.76, sys: 12.81, mem: 1872412 ko)
```
These files are painfully slow.  I expect this can be fixed by making select lemmas `Qed` rather than `Defined`.

JasonGross:
Btw, to test this out yourself, you need to `configure` Coq itself with `-native-compiler yes` when building from source.

JasonGross:
The current list of files like this:
[...]

PR #1474, opened 2021-04-22
pass `-native-compiler ondemand`

JasonGross:
This will make my life nicer while we wait for a real fix to https://github.com/HoTT/HoTT/issues/1454


Alizter (approved):
LGTM

PR #1477, opened 2021-04-23
Mark Tr as Opaque

Alizter:
I'm creating this draft so we can see if it fixes some issues highlighted in #1475.

Note that the following now works correctly:
```coq
Require Import HoTT.

Local Open Scope trunc_scope.

Definition test (X : pType) `{IsTrunc 1 X}
  : { bloop' : Pi1 X -> loops X &
                   forall x y : Pi1 X, bloop' (x * y) = bloop' x @ bloop' y }.
Proof.
    exists (equiv_tr 0 _)^-1%equiv.
    Time simpl.
```
But I also see that the behaviour of `cbn` is vastly different from `simpl`.

This is currently on top of #1472.

### DIFF
--- a/theories/Spaces/BAut/Rigid.v
+++ b/theories/Spaces/BAut/Rigid.v
@@ -7,6 +7,9 @@ Require Import Spaces.BAut.
 Local Open Scope trunc_scope.
 Local Open Scope path_scope.
 
+(** We mark Tr as transparent here as we need to unfold it. *)
+Transparent Tr.
+
 (** * Rigid types *)
 
 Class IsRigid (A : Type) := 
@@ -27,7 +30,7 @@ Global Instance contr_baut_rigid `{Univalence} {A : Type} `{IsRigid A}
   : Contr (BAut A).
 Proof.
   refine (contr_change_center (point (BAut A))).
-  refine (contr_trunc_conn (Tr 0)).
+  nrefine (contr_trunc_conn (Tr 0)); [|exact _].
   intros Z W; baut_reduce.
   refine (istrunc_equiv_istrunc (n := -1) (A <~> A)
                       (path_baut (point (BAut A)) (point (BAut A)))).
--- a/theories/Truncations/Core.v
+++ b/theories/Truncations/Core.v
@@ -63,6 +63,9 @@ Proof.
   - intros; reflexivity.
 Defined.
 
+(** We want to discourage coq from unfolding [Tr] as much as possible. *)
+Global Opaque Tr.
+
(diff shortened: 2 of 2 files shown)


JasonGross:
Are you sure we want to mark Tr as Opaque rather than having O_rec unfold early?  (I think we should be principled about this; what's the right level of abstraction for these things?)

Alizter:
I'm not sure if it is a good idea, hence I've made a draft PR. I'll also do some experimenting with making O_rec unfold. Currently it is set to `simpl never`.

jdchristensen:
The only difference I see between `cbn` and `simpl` is that `cbn` also unfolds the definition of `*`.

Despite what the manual says, I find that `simpl` is a better behaved version of `cbn` in many cases...

jdchristensen:
I manually applied the two changes from this PR to current master (marking Tr Opaque in Truncations/Core.v and Transparent in Rigid.v), and checked the no file had a significant change in build time.

Alizter:
@jdchristensen I take it that it is taking longer?

jdchristensen:
Sorry, there was a typo.  I meant that none of the files had any significant change in build time.

Alizter:
rebased

Alizter:
After thinking about it I am not too happy with making Tr opaque. If we ever need to cbv something in the future, we will have to remember to make Tr transparent again. This seems like it will cause more problems than it will solve.

jdchristensen:
Yes, that's what I was thinking in #1475 when I wrote "I can imagine situations in which this might not be ideal."  Alternatives are `simpl never` (which helps with `cbn` but not `simpl`) or making `O_rec` unfold more easily (which you said would require changes to some files).  It's a shame that no solution is ideal, as there are many times when I try to do `simpl` or `cbn`, but have to avoid them because they are so slow, with the result that the proof is harder to follow along with in Coq.  I gave one test case, but this arose several times over the last week, and I tested one other case and found that it was solved by making `Tr` opaque. Maybe the `simpl never` approach is a reasonable compromise?   Or maybe the changes needed for the `O_rec` solution aren't too invasive?

(I guess further discussion should move back to #1475.)

theories/Spaces/BAut/Rigid.v
```diff
@@ -27,7 +30,7 @@ Global Instance contr_baut_rigid `{Univalence} {A : Type} `{IsRigid A}
   : Contr (BAut A).
 Proof.
   refine (contr_change_center (point (BAut A))).
-  refine (contr_trunc_conn (Tr 0)).
+  nrefine (contr_trunc_conn (Tr 0)); [|exact _].
```
jdchristensen, line 33: When I tested this change, I didn't need to change this line.

Alizter: This was not needed, but a small speedup that I had observed I believe.

Alizter:
I'm closing this as it is unlikely to get merged. We can always reopen.

PR #1479, opened 2021-04-23
mark eisadj as globally Opaque

Alizter:
I'll do some timing checks later


jdchristensen:
I tested that putting `Opaque eisadj.` before the `Time simpl.` line in my test in #1475 brings the time down from 98s to 12.1s.  Still slightly slower than before the `eisadj` changes around April 10-12, but very similar.  (The timing is 11.1s with the Apr 10 commit.)  I didn't actually test this PR.  It would be interesting to see the changes to timing for the entire library before and after this PR, and maybe also comparing this PR to the Apr 10 state.

jdchristensen:
I tested the timing of the build with and without this change, and there was essentially zero difference.  (Max time difference for any file was 0.07s.)  Since this change is what was intended originally, I think we should go ahead and merge it.  While it doesn't affect the build time, it does in practice greatly speed up some steps while trying to prove things interactively.

mikeshulman:
Agreed.

PR #1480, opened 2021-04-23
More cleanups and speedups to ClassifyingSpace.v.

jdchristensen:
- Reverse the direction of the equivalence between ΩBG and G, so it
  goes in the natural direction bloop : G -> ΩBG.  This makes several
  things cleaner and allows us to drop encode_pp.
- Make IsEquiv bloop a Global Instance.
- Prove that it is natural.
- Remove many rewrites, especially in the H-space part.
  Replace one entire proof by reflexivity.
- Use nrapply instead of rapply in some places.

The entire file now takes 2s to typecheck on my system.

### DIFF
--- a/theories/Homotopy/ClassifyingSpace.v
+++ b/theories/Homotopy/ClassifyingSpace.v
@@ -283,6 +283,8 @@ Section EncodeDecode.
     apply bloop_id.
   Defined.
 
+  Definition pequiv_loops_bg_g := pequiv_g_loops_bg^-1*%equiv.
+
   (** We also have that the equivalence is a group isomorphism. *)
 
   (** First we show that the loop space of a pointed 1-type is a group *)
--- a/theories/Homotopy/ClassifyingSpace.v
+++ b/theories/Homotopy/ClassifyingSpace.v
@@ -261,45 +261,26 @@ Section EncodeDecode.
     apply bloop_id.
   Defined.
 
-  (** Universal property of BG *)
-  Theorem equiv_loops_bg_g : loops (B G) <~> G.
+  Global Instance isequiv_bloop : IsEquiv (@bloop G).
   Proof.
-    srapply equiv_adjointify.
+    srapply isequiv_adjointify.
     + exact (encode _).
-    + exact bloop.
+    + rapply decode_encode.
     + intro x.
       refine (codes_transport _ _ @ _).
       apply left_identity.
-    + intro.
-      apply (decode_encode bbase x).
   Defined.
 
-  Definition equiv_bloop := equiv_loops_bg_g^-1%equiv.
+  (** Defining property of BG *)
+  Definition equiv_g_loops_bg : G <~> loops (B G)
+    := Build_Equiv _ _ bloop _.
 
-  (** Pointed version of the universal property. *)
-  Theorem pequiv_loops_bg_g : loops (B G) <~>* Build_pType G _.
+  (** Pointed version of the defining property. *)
+  Definition pequiv_g_loops_bg : Build_pType G _ <~>* loops (B G).
   Proof.
     srapply Build_pEquiv'.
-    1: apply equiv_loops_bg_g.
-    reflexivity.
-  Defined.
-
-  Local Lemma encode_pp' (x : B G) (p q : bbase = x)
-    : encode _ (p @ q^) = transport (fun x : B G => codes x) q^ (encode x p).
-  Proof.
-    destruct q; cbn.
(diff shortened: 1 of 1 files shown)


jdchristensen:
I hadn't noticed that other files use the equivalence in the original direction.  I'll fix it...

Alizter:
I've removed funext from the HSpace structure btw if you are interested in adding it:
```coq
(** When G is an abelian group, BG is a H-space. *)
Section HSpace_bg.

  Context {G : AbGroup}.

  Definition bg_mul : B G -> B G -> B G.
  Proof.
    intro b.
    snrapply ClassifyingSpace_rec.
    1: exact _.
    1: exact b.
    { intro x.
      revert b.
      snrapply ClassifyingSpace_ind_hset.
      1: exact _.
      1: exact (bloop x).
      cbn; intro y.
      apply dp_paths_lr.
      refine (concat_pp_p _ _ _ @ _).
      apply moveR_Vp.
      refine ((bloop_pp _ _)^ @ _ @ bloop_pp _ _).
      apply ap, commutativity. }
    intros x y.
    revert b.
    srapply ClassifyingSpace_ind_hprop.
    exact (bloop_pp x y).
  Defined.

  Definition bg_mul_symm : forall x y, bg_mul x y = bg_mul y x.
  Proof.
    intros x.
    srapply ClassifyingSpace_ind_hset.
    { revert x.
      srapply ClassifyingSpace_ind_hset.
      1: reflexivity.
      intros x.
      apply sq_dp^-1, sq_1G.
      refine (ap_idmap _ @ _^).
      nrapply ClassifyingSpace_rec_beta_bloop. }
    intros y; revert x.
    snrapply ClassifyingSpace_ind_hprop; simpl.
    1: exact _.
    apply sq_dp^-1, sq_1G.
    refine (_ @ (ap_idmap _)^).
    nrapply ClassifyingSpace_rec_beta_bloop.
  Defined.

  Definition bg_mul_left_id
    : forall a : B G, bg_mul bbase a = a.
  Proof.
    intros a.
    refine (bg_mul_symm _ _ @ _).
    reflexivity.
  Defined.

  Definition bg_mul_right_id
    : forall a : B G, bg_mul a bbase = a.
  Proof.
    reflexivity.
  Defined.

  Global Instance ishspace_bg : IsHSpace (B G)
    := Build_IsHSpace _
          bg_mul
          bg_mul_left_id
          bg_mul_right_id.

End HSpace_bg.
```

Alizter (approved):
LGTM! These really are some much needed improvements.

jdchristensen:
Nice work on Funext!  I forgot to try to get rid of it.  I'll take a look tomorrow and push a new version then.

jdchristensen:
Your code is much cleaner.  It turns out that adding a couple of `simpl`s makes it both easier to read and faster, so I did so.  Now it's the same speed as the Funext version.  The slowest line is the Defined at the end of bg_mul_symm, at around 0.2s.  Probably not worth worrying about, so this LGTM.

Alizter:
This is a general trick that can be done when defining something like `X -> Y -> Z` be recursion on `X` and `Y`. If `X` requires truncatedness assumptions when inducting, then you are probably better off inducting on `Y` first and then `X` second. This is contrary to how many functions such as `plus` for `nat` are defined, but you get to avoid funext. I learnt this trick from Floris.

jdchristensen:
Yes, I've been using that trick for other things and have some ideas about abstracting it.  I'll open a separate issue to discuss them.

About the H-space structure, there's a completely different approach that one could take that is more conceptual and would make it easy to prove more coherence for the H-space structure.

First, prove that `B (grp_prod G H) <~>* B G * B H` (which is good to know in any case).  There is a natural map from left to right given by applying the functoriality of `B` to the two projection maps.  You can show it is an equivalence by applying `loops_functor` and checking that a diagram commutes.

If `G` is abelian, the group operation gives a group homomorphism `m : (grp_prod G G) -> G`.  Composing the inverse of the above equivalence with the map induced on classifying spaces gives a pointed map `BG * BG ->* BG` which is the H-space structure.  The pointedness gives the two unit laws.  It is obviously symmetric since `m` is equal to `m` precomposed with the swap homomorphism.  Similarly, it is associative and satisfies coherences since `m` does.

Not sure if I'll get a chance to try this approach, but I thought I'd record it here.  (I don't think we should hold off on merging this PR.  Even if someone wants to pursue that approach, we should get the current method into the git history.)

jdchristensen:
Unless there are any other comments, I'll merge this version in a bit using the 48 hour rule.

PR #1482, opened 2021-04-24
Parametrise IsProjective and IsQuotientChoosable

andreaslyn:
I have parametrised `IsProjective` by a modality, which was suggested by @Alizter as a way to generalise over all truncation levels including infinity.

I have changed name of `IsProjective'` to `IsChoosable`, to avoid confusion with `IsProjective`.

I have moreover defined `IsQuotientChoosable`, equivalent to `(0,-1)-projective`, which I will need later on. Although I do not use it, I have included `choose_quotient_ind`, which generalises `quotient_rec2`. I think it is a nice recursion principle for `quotient`.

### DIFF
--- a/theories/Colimits/Quotient/Choice.v
+++ b/theories/Colimits/Quotient/Choice.v
@@ -7,20 +7,20 @@ Require Import
   HoTT.Colimits.Quotient
   HoTT.Projective.
 
-(** The following is an alternative (0,-1)-projective predicate. Given a family of quotient sets [forall x : A, B x / R x], for [R : forall x : A, Relation (B x)], we merely have a choice function [g : forall x, B x], such that [class_of (g x) = f x]. *)
+(** The following is an alternative (0,-1)-projectivity predicate on [A]. Given a family of quotient equivalence classes [f : forall x : A, B x / R x], for [R : forall x : A, Relation (B x)], we merely have a choice function [g : forall x, B x], factoring [f] as [f x = class_of (g x)]. *)
 
-Definition IsQuotientChoosable (A : Type) :=
+Definition HasQuotientChoice (A : Type) :=
   forall (B : A -> Type), (forall x, IsHSet (B x)) ->
   forall (R : forall x, Relation (B x))
          (pR : forall x, is_mere_relation (B x) (R x)),
          (forall x, Reflexive (R x)) ->
          (forall x, Symmetric (R x)) ->
          (forall x, Transitive (R x)) ->
-  forall (f : forall x : A, B x / (R x)),
+  forall (f : forall x : A, B x / R x),
   hexists (fun g : (forall x : A, B x) =>
                    forall x, class_of (R x) (g x) = f x).
 
-Section choose_isquotientchoosable.
+Section choose_has_quotient_choice.
   Context `{Univalence}
     {A : Type} {B : A -> Type} `{!forall x, IsHSet (B x)}
     (P : forall x, B x -> Type) `{!forall x (a : B x), IsHProp (P x a)}.
@@ -79,12 +79,12 @@ Section choose_isquotientchoosable.
     : B a / RelClassEquiv a
     := (prechoose i a).1.
 
-End choose_isquotientchoosable.
+End choose_has_quotient_choice.
 
-(** The following section derives [IsTrChoosable 0 A] from [IsQuotientChoosable A]. *)
+(** The following section derives [HasTrChoice 0 A] from [HasQuotientChoice A]. *)
 
-Section isquotientchoosable_to_istr0choosable.
-  Context `{Univalence} (A : Type) (qch : IsQuotientChoosable A).
+Section has_tr0_choice_quotientchoice.
+  Context `{Univalence} (A : Type) (qch : HasQuotientChoice A).
 
   Local Definition RelUnit (B : A -> Type) (a : A) (b1 b2 : B a) : HProp
     := Build_HProp Unit.
@@ -118,7 +118,7 @@ Section isquotientchoosable_to_istr0choosable.
     by apply qglue.
   Qed.
 
-  Lemma isquotientchoosable_to_istr0choosable : IsTrChoosable 0 A.
+  Lemma has_tr0_choice_quotientchoice : HasTrChoice 0 A.
   Proof.
     intros B sB f.
     transparent assert (g : (forall a, B a / RelUnit B a)).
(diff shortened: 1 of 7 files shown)


Alizter:
@jarlg May be interested in this as the author of `Projective.v`.

Alizter:
You seem to be using the old quotient which we are trying to move away from. You will find the new one in `Colimits/Quotient.v`.

andreaslyn:
Thanks. I will change to the new quotient.

andreaslyn:
What do you think about adding a directory `Colimits/Quotient/Choice.v` for the `IsQuotientChoosable` stuff?

theories/HIT/quotient/choice.v
```diff
@@ -0,0 +1,336 @@
+Require Import
+  HoTT.Basics
+  HoTT.Types
+  HoTT.HSet
+  HoTT.Truncations
+  HoTT.TruncType
+  HoTT.HIT.quotient
+  HoTT.Projective.
+
+(** The following is an alternative (0,-1)-projective predicate.
```
Alizter, line 10: I think we mention in the STYLE.md guide that comments should not be hard wrapped. I believe this is for documentation reasons.

mikeshulman:
I think the main reason for not hard-wrapping comments is that different
screens have different widths, and hard-wrapped comments look awful when
viewed in a window that's narrower than the width they were wrapped to.

andreaslyn:
I have adapted to the new quotient.

Do you know of better names for `IsChoosable` and `IsQuotientChoosable`?

Alizter:
What's the reason for all the `Qed`'s? Is it just for performance?

theories/Projective.v
```diff
@@ -1,27 +1,41 @@
...
+(** To quantify over all truncation levels including infinity, we parametrize [IsProjective] by a [Modality]. When specializing to [IsProjective purely] we get an (oo,-1)-projective predicate. When specializing to [IsProjective (Tr n)] we get an (n,-1)-projective predicate. *)
+
+Definition IsProjective (O : Modality) (X : Type) : Type
+  := forall A, In O A -> forall B, In O B ->
+     forall f : X -> B, forall p : A -> B,
+     IsSurjection p -> merely (exists s : X -> A, p o s == f).
+
+(** An (oo,-1)-projective predicate [IsPureProjective]. *)
+
+Notation IsPureProjective := (IsProjective purely).
```
jarlg, line 20: The notion of "pure-projective" (modules) already exists in algebra, could we instead call this just `IsProjective` and make the parametrized version above `IsOProjective`?

andreaslyn: Good point. Do you have a suggestion for a name of `IsChoosable`?

jarlg (commented):
Nice additions! I've only really looked through `IsProjective.v`as I don't have context for the quotient stuff. I notice that Exercise 7.9 can now be done for all `n`, but I can do that in a later PR. Thanks!

theories/Projective.v
jarlg: Just saw this -- I suggested `HasSections` below.

theories/Projective.v
```diff
@@ -31,53 +45,84 @@ Proof.
...
 
-(** In topos theory, an object X is said to be projective if the axiom of choice holds when making choices indexed by X. *)
-Definition IsProjective' (X : Type) : Type := forall P : X -> Type,
-    (forall x, merely (P x)) -> merely (forall x, P x).
+(** In topos theory, an object X is said to be projective if the axiom of choice holds when making choices indexed by X. We will refer to this as [IsChoosable], to avoid confusion with [IsProjective] above. In similarity with [IsProjective], we parametrize it by a [Modality]. *)
+
+Class IsChoosable (O : Modality) (A : Type) :=
+  ischoosable :
+    forall (B : A -> Type), (forall x, In O (B x)) ->
+    (forall x, merely (B x)) -> merely (forall x, B x).
```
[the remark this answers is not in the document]
andreaslyn: That is quite a creative name, I am fine with `HasSections`.

andreaslyn:
I have rebased and implemented the suggestions from @jarlg. I decided to rename `IsQuotientChoosable` to `HasQuotientFactorings`. What do you think?

andreaslyn:
Yes `HasQuotientChoice` is better. I also realized that `HasQuotientFactorings` could be confused with image factorization.

jarlg:
Then `HasSections` should become `HasChoice` as well, I think. :)

theories/Colimits/Quotient/Choice.v
```diff
@@ -0,0 +1,326 @@
...
+  Proof.
+    apply hprop_allpath.
+    refine (Quotient_ind_hprop _ _ _).
+    intro r.
+    refine (Quotient_ind_hprop _ _ _).
+    intro s.
+    by apply qglue.
+  Qed.
+
+  Lemma has_quotient_choice_to_has_tr0_choice : HasTrChoice 0 A.
```
mikeshulman, line 121: Generally, our naming convention is that the result type is indicated first, followed by the inputs and relevant hypotheses.  E.g. `isequiv_transport` says that `transport` is an equivalence.  So I would name this something like `has_tr0_choice_quotientchoice`.

theories/Colimits/Quotient/Choice.v
```diff
@@ -0,0 +1,326 @@
...
+      exact (class_of _ f).
+    - specialize (qch B _ (RelUnit B) _ _ _ _ g).
+      strip_truncations.
+      apply tr.
+      apply qch.
+  Qed.
+
+End has_quotient_choice_to_has_tr0_choice.
+
+Lemma has_tr0_choice_to_has_quotient_choice (A : Type)
```
mikeshulman, line 137: Similarly, I'd rename this.

andreaslyn: Good idea. I have changed this including a couple of results in Projective.v, which I renamed to `isoprojective_ochoice` and `hasochoice_oprojective`.

mikeshulman:
Looks like this needs rebasing.

theories/Colimits/Quotient/Choice.v
```diff
@@ -0,0 +1,326 @@
...
+    exists (a h).
+    apply tr.
+    exists h.
+    exists 1.
+    apply E.
+  Defined.
+
+(** By projecting out of [chooseprop_quotient_ind] we obtain a generalization of [quotient_rec2]. *)
+
+  Lemma choose_quotient_ind
```
andreaslyn, line 262: This is quotient induction, but generalised to `forall i, A i / R i`.

andreaslyn:
CI is passing now.

mikeshulman (approved):
Unfortunately, I guess #1525 requires another rebase.

theories/Spaces/Finite/Finite.v
```diff
@@ -211,10 +211,9 @@ Defined.
 
 (** ** The finite axiom of choice, and projectivity *)
 
-Definition finite_choice {X} `{Finite X} (P : X -> Type)
-  : (forall x, merely (P x)) -> merely (forall x, P x).
+Definition finite_choice {X} `{Finite X} : HasChoice X.
 Proof.
-  intros f.
+  intros P oP f; clear oP.
```
Alizter, line 216: I'm not suggesting you change it, but you can use intro patterns to do this in one go.
```coq
intros P _ f.
```
Though, the way you have written it is arguably more readable.

Alizter (approved):

PR #1484, opened 2021-04-25
Make nrefine, nrapply, and snrapply find global axioms.

### LINKED ISSUE #1483: Make nrefine and friends handle Funext and Univalence?
We should be using nrefine, nrapply and snrapply when they suffice, as they are faster and more predictable since they don't do typeclass search.  But a frequent annoyance when using them is having to manually handle trivial Funext or Univalence goals.  What if we change nrefine to:
```coq
Tactic Notation "nrefine" uconstr(term) := notypeclasses refine term; try match goal with
    | |- Funext  => exact _
    | |- Univalence => assumption
  end.
```
Then those trivial goals that are quick to handle would be solved for us.  (The reason to use `exact _` for Funext is that we may have Univalence in the context.)

With the above done, it becomes more pleasant to try to speed up existing code by adding the letter `n` in various places.  Of course, a bunch of existing code would need to have some `exact _` lines removed if the above is done, but I'm willing to do that.

jdchristensen:
This is based on discussion in #1483.  It will make it easier to use the `n` form of our tactics.

I testing the timing, and no file has a significant change in build time.

Closes #1483


theories/Basics/Tactics.v
```diff
@@ -378,12 +378,27 @@ Tactic Notation "do_with_holes'" tactic3(x) uconstr(p) :=
...
   x uconstr:(p).
 
+(** We keep a list of global axioms that we will solve automatically, even when not using typeclass search. *)
+Unset Primitive Projections.
+Class IsGlobalAxiom (A : Type) := {}.
+Set Primitive Projections.
+Global Hint Mode IsGlobalAxiom + : typeclass_instances.
+
+(** We add [Funext] to the list here, and will later add [Univalence]. *)
+Instance is_global_axiom_funext : IsGlobalAxiom Funext := {}.
```
SkySkimmer, line 388:
```suggestion
Global Instance is_global_axiom_funext : IsGlobalAxiom Funext := {}.
```

theories/Basics/Tactics.v
```diff
@@ -679,7 +694,7 @@ Ltac issig :=
   let u := fresh "u" in
   let v := fresh "v" in
   (** We build an equivalence with 5 holes. *)
-  snrefine  (* We don't want typeclass search running. *)
+  simple notypeclasses refine  (* We don't want typeclass search running. *)
```
SkySkimmer, line 697: ?

theories/ExcludedMiddle.v
```diff
@@ -5,6 +5,9 @@ Require Import HoTT.Basics HoTT.Types.
 Monomorphic Axiom ExcludedMiddle : Type0.
 Existing Class ExcludedMiddle.
 
+(** Mark this axiom as a "global axiom", which some of our tactics will automatically handle. *)
+Instance is_global_axiom_excludedmiddle : IsGlobalAxiom ExcludedMiddle := {}.
```
SkySkimmer, line 9:
```suggestion
Global Instance is_global_axiom_excludedmiddle : IsGlobalAxiom ExcludedMiddle := {}.
```

theories/PropResizing/PropResizing.v
```diff
@@ -7,6 +7,10 @@ Local Open Scope path_scope.
 (** See the note by [Funext] in Overture.v regarding classes for axioms *)
 Monomorphic Axiom PropResizing : Type0.
 Existing Class PropResizing.
+
+(** Mark this axiom as a "global axiom", which some of our tactics will automatically handle. *)
+Instance is_global_axiom_propresizing : IsGlobalAxiom PropResizing := {}.
```
SkySkimmer, line 12:
```suggestion
Global Instance is_global_axiom_propresizing : IsGlobalAxiom PropResizing := {}.
```

theories/Types/Universe.v
```diff
@@ -34,6 +34,10 @@ Defined.
 (** See the note by [Funext] in Overture.v *)
 Monomorphic Axiom Univalence : Type0.
 Existing Class Univalence.
+
+(** Mark this axiom as a "global axiom", which some of our tactics will automatically handle. *)
+Instance is_global_axiom_univalence : IsGlobalAxiom Univalence := {}.
```
SkySkimmer, line 39:
```suggestion
Global Instance is_global_axiom_univalence : IsGlobalAxiom Univalence := {}.
```

theories/Basics/Tactics.v
jdchristensen: I explained this change in #1483.

SkySkimmer: Is it still needed after fixing the hint mode?

jdchristensen: Ah, now I understand.  You are right that it is no longer needed.
```suggestion
  snrefine  (* We don't want typeclass search running. *)
```

Alizter:
If these are "Global" axioms, what are "local" axioms? Do we have any use for them?

Alizter (approved):

PR #1490, opened 2021-04-26
disabled -native-compiler ondemand option due to noise

### LINKED ISSUE #1489: Noise coming from native compiler option
Unfortunately, the `-native-compiler ondemand` option creates a lot of noise when building the library with coq 8.13.2. This was introduced in #1474.

```
Warning: Native compiler is disabled, -native-compiler ondemand option
ignored. [native-compiler-disabled,native-compiler]
```

Currently there is no way to hide this warning but coq/coq#13821 fixes this issue.

I therefore think we should disable these arguments in our `_CoqProject` file and wait till coq 8.14 to re-enable them.

@JasonGross since this mostly affects you, is this the right way to go about it?

Alizter:
closes #1489

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -4,7 +4,7 @@
 #-arg -boot
 -arg -noinit
 -arg -indices-matter
--arg -native-compiler -arg ondemand
+#-arg -native-compiler -arg ondemand
 
 #
 #   Table of contents:
--- a/_CoqProject
+++ b/_CoqProject
@@ -4,7 +4,7 @@
 #-arg -boot
 -arg -noinit
 -arg -indices-matter
-#-arg -native-compiler -arg ondemand
+-arg -native-compiler -arg no
 
 #
 #   Table of contents:


_CoqProject
```diff
@@ -4,7 +4,7 @@
 #-arg -boot
 -arg -noinit
 -arg -indices-matter
--arg -native-compiler -arg ondemand
+#-arg -native-compiler -arg ondemand
```
SkySkimmer, line 7:
```suggestion
-arg -native-compiler -arg no
```
This should not warn and AFAIK we don't use native at all.

Alizter: @JasonGross want's to use native with the library. See #1474 and #1454.

JasonGross: I say it's better to leave it disabled for now; I'd prefer `-native-compiler no` to not passing any argument, at least until we make the library compile with the native compiler.  What's the warning, though?

Alizter: See #1489

Alizter:
@JasonGross does this look good?

JasonGross (approved):

PR #1491, opened 2021-04-26
add TIMING=1 by default to silence noise

### LINKED ISSUE #1487: Build noisier after Makefile changes?
There's lots of stuff in console now while building.  Maybe these are things that we should fix anyways?  If not, then I'd prefer to silence them somehow, as they make me think I've introduced a problem...

I filtered out the COQC lines, and this is the extra stuff:

```
NaiveFunext
     : Type
NaiveNondepFunext
     : Type
WeakFunext
     : Type
@Univalence_implies_FunextNondep : Univalence_type -> NaiveNondepFunext
     : Univalence_type -> NaiveNondepFunext
ExtendableAlong_Over
     : forall (n : nat) (A B : Type) (f : A -> B) 
       (C : B -> Type),
       (forall b : B, C b -> Type) -> ExtendableAlong n f C -> Type
ooExtendableAlong_Over
     : forall (f : ?A -> ?B) (C : ?B -> Type),
       (forall b : ?B, C b -> Type) -> ooExtendableAlong f C -> Type
where
?A : [ |- Type]
?B : [ |- Type]
fun X : Type => idem_idmap X : Idempotent X : Type
     : forall X : Type, Idempotent X
Susp nat : Type0
     : Type0
1 : Lift nat
     : Lift nat
lift'
     : ?A -> Lift' ?A
where
?A : [ |- Type]
lower'
     : Lift' ?A -> ?A
where
?A : [ |- Type]
isequiv_lift'
     : forall T : Type, IsEquiv lift'
fun A : Type => ispointed_susp : IsPointed (Susp A)
     : forall A : Type, IsPointed (Susp A)
ispointed_susp
     : IsPointed (Susp Type)
ispointed_susp
     : IsPointed (Susp Type0)
{_ : Unit & {_ : Unit & Unit}}
     : Type
prod_rect
     : forall (A B : Type) (P : A * B -> Type),
       (forall (fst : A) (snd : B), P (fst, snd)) -> forall p : A * B, P p
```

Alizter:
<s>closes # 1487</s>

### DIFF
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -19,7 +19,7 @@
 # (transitive) use of variables defined after this file's inclusion in
 # Makefile.coq
 include Makefile.coq.local-early
-.SECONDEXANSION:
+.SECONDEXPANSION:
 
 # TODO(not yet ported from automake): install these files: theories etc LICENSE.txt CREDITS.txt INSTALL.md README.md
 
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -19,7 +19,7 @@
 # (transitive) use of variables defined after this file's inclusion in
 # Makefile.coq
 include Makefile.coq.local-early
-.SECONDEXPANSION:
+.SECONDEXANSION:
 
 # TODO(not yet ported from automake): install these files: theories etc LICENSE.txt CREDITS.txt INSTALL.md README.md
 
@@ -29,6 +29,10 @@ CAMERA = python proviola/camera/camera.py
 # the alectryon binary
 ALECTRYON = python3 etc/alectryon/alectryon.py
 ALECTRYON_EXTRAFLAGS ?=
+
+# In the old makefile set up we silence noise coming from coqc by
+# setting TIMING=1. By doing so here we get the same effect.
+TIMING ?= 1
 
 ETAGS ?= etags
 


Makefile.coq.local
```diff
@@ -19,7 +19,7 @@
 # (transitive) use of variables defined after this file's inclusion in
 # Makefile.coq
 include Makefile.coq.local-early
-.SECONDEXPANSION:
+.SECONDEXANSION:
```
JasonGross, line 22: This change should be reverted, it is a misspelling

Makefile.coq.local
```diff
@@ -30,6 +30,10 @@ CAMERA = python proviola/camera/camera.py
 ALECTRYON = python3 etc/alectryon/alectryon.py
 ALECTRYON_EXTRAFLAGS ?=
 
+# In the old makefile set up we silence noise coming from coqc by
+# setting TIMING=1. By doing so here we get the same effect.
```
JasonGross, line 34: It was an accident that the setup silenced noise; the real goal was to get timing files available.  It may be better to just `Redirect` the `Check`s / remove them, so that if new ones are introduced, we notice.

Makefile.coq.local
Alizter: oops, I didn't see that.

Alizter, line 22:
```suggestion
.SECONDEXPANSION:
```

Makefile.coq.local
Alizter: Perhaps this being an accident is an orthogonal issue however? I am proposing that the messages get handled the same way we did before the makefile changes. Redirecting the output ought to be a new issue.

Alizter:
@jdchristensen Will this fix be adequate for now? We can discuss the proper way to handle checks in another issue?

jdchristensen:
It should be very easy to silence the warnings, so we could just settle on one of the methods and then do it.  But if you want to commit this first and then revert it later, I guess that's fine.

Alizter:
I'm merging this but not closing #1487

PR #1495, opened 2021-04-28
Generated subgroups and some cleanup in group theory library

Alizter:
This PR adds subgroups generated by a type family on a group. This can be found in Subgroup.v

We also do some cleanup in Group.v such as standardising the names of the group laws, adding miscellaneous comments and writing down all the group movement lemmas.

Also the index file Groups.v was missing a few files.

@jarlg I am interested in your opinion on the reorganization.


theories/Algebra/Groups/Subgroup.v
```diff
@@ -348,3 +361,104 @@ Proof.
...
+  rewrite <- (negate_involutive h).
+  apply sgt_op.
+  1: assumption.
+  apply sgt_inv.
+  assumption.
+Defined.
+
+
+
+(** Note that this will not automatically be a hprop. Notice that if [X] already contained the unit of the group, then there are at least two different inhabitants of this family at the unit. We can fix this by propositionally truncated. *)
```
mikeshulman, line 429:
```suggestion
(** Note that this will not automatically be a hprop. Notice that if [X] already contained the unit of the group, then there are at least two different inhabitants of this family at the unit. We can fix this by propositionally truncating. *)
```

mikeshulman:
Looks ok to me.

PR #1496, opened 2021-04-28
More ring theory, movement lemmas and cleanup

Alizter:
This builds on top of #1495 so for the time being, I've marked it as draft.

Here is a summary of the changes:
 * Add more ring laws
 * Add `≅` notation for `CRingIsomorphism` as that was quickly getting tiresome to write. This notation lives in `ring_scope`. 
 * Add another constructor for Rings that takes in the underlying abelian group.
 * Image of a ring homomorphism as a ring together with inclusion maps
 * Product ring
 * Define powers of ring elements together with some basic laws
 * Add in ring movement lemmas (for now only for addition) e.g. `rng_moveR_Mr : y = - x + z <~> x + y = z `

### DIFF
--- a/theories/Algebra/Rings/CRing.v
+++ b/theories/Algebra/Rings/CRing.v
@@ -306,8 +306,51 @@ Proof.
   intros [r1 s1] [r2 s2]; apply path_prod; cbn; apply rng_mult_comm.
 Defined.
 
+Infix "×" := cring_product : ring_scope.
+
+Definition cring_product_fst {R S : CRing} : R × S $-> R.
+Proof.
+  snrapply Build_CRingHomomorphism.
+  1: exact fst.
+  repeat split.
+Defined.
+
+Definition cring_product_snd {R S : CRing} : R × S $-> S.
+Proof.
+  snrapply Build_CRingHomomorphism.
+  1: exact snd.
+  repeat split.
+Defined.
+
+Definition cring_product_corec (R S T : CRing)
+  : (R $-> S) -> (R $-> T) -> (R $-> S × T).
+Proof.
+  intros f g.
+  srapply Build_CRingHomomorphism'.
+  1: apply (ab_biprod_corec f g).
+  repeat split.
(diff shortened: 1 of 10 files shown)


mikeshulman:
Could you use a general notation for equivalences/isomorphisms in a (wild) category, instead of introducing a new notation just for ring isomorphisms?

mikeshulman (approved):

Alizter:
I originally avoided using the wildcat library  since I didn't want to impose `≅` as a notation for equivalences. But thinking about it now, it may be possible to have it in a scope similar to how we handle SgOp with `mc_add_scope`.

theories/Algebra/Rings/CRing.v
```diff
@@ -231,3 +332,76 @@ Proof.
...
+  { intros [r1 s1] [r2 s2] [r3 s3].
+    apply path_prod; cbn; apply rng_mult_assoc. }
+  1: intros [r1 s1]; apply path_prod; cbn; apply rng_mult_one_l.
+  1: intros [r1 s1]; apply path_prod; cbn; apply rng_mult_one_r.
+  intros [r1 s1] [r2 s2]; apply path_prod; cbn; apply rng_mult_comm.
+Defined.
+
+Infix "×" := cring_product : ring_scope.
+
+Definition cring_product_corec (R S T : CRing)
```
andreaslyn, line 358: Consider adding the universal property `Hom R S * Hom R T <~> Hom R (S * T)` below this.

andreaslyn (approved):

Alizter:
OK, I've introduced `≅` as a notation for CatEquiv in WildCat/Equiv.v. It can be accessed using `wc_iso_scope`. So if this scope is open, we print `≅` instead of `$<~>`. I've gone ahead and also used it in a few places.

I've also rebased.

theories/Algebra/Rings/CRing.v
Alizter: Done!

PR #1497, opened 2021-04-28
Ideal arithmetic and Chinese remainder theorem

Alizter:
This PR builds on top of #1496 and #1495.

Here is a summary of the PR:

The biggest changes are in Ideal.v which looks like it has been written from scratch:

 * We begin with various lemmas about ring elements being in an ideal
 * We define intersection ideals, sums of ideals, products of ideals
 * We define the ideal generated by a family on a ring (different but analagous to the subgroup generated in #1495)
 * We can define principal ideals (but not what it means to be principal yet)
 * We define equality of ideals, which is just a family of "if and only if"'s
 * We define subsets of ideals, i.e. a family of implications (this is our notion of "divded by" for ideals)
 * Define the colon ideal
 * Define the annihilator of an ideal
 * We define some ideal (unicode) notations. These are convenient to use and we are running out of ascii.
 * Define coprime (comaximal) ideals and properties (e.g. existence of x and y such that x + y = 1).
 * We state and prove every lemmas about ideal arithmetic I could think of / find. There are a lot!
 
Next we do some improvements and cleanup in QuotientRing.v:
 * We add a quotient notation and use previous ring notations to make statements more concise
 * We add convenient induction principles for ring quotients whose hypothesis talk about ring elements.
 * We add a quality-of-life improvement to the first isomorphism theorem 

Finally we have a new file called Algebra/Rings/ChineseRemainder.v in which we prove the chinese remainder theorem. I've spent some time documenting the proof so it is easy to read. Interestingly it uses univalence since we need to reason about equalities in the quotient ring. Since our rings are commutative, we also state and prove the version that uses products of ideals instead of intersections.

### DIFF
--- a/theories/Algebra/Rings/ChineseRemainder.v
+++ b/theories/Algebra/Rings/ChineseRemainder.v
@@ -1,0 +1,149 @@
+Require Import Basics Types WildCat.
+Require Import Algebra.Congruence.
+Require Import Algebra.AbGroups.
+Require Import Algebra.Rings.CRing.
+Require Import Algebra.Rings.Ideal.
+Require Import Algebra.Rings.QuotientRing.
+
+(** * Chinese remainder theorem *)
+
+Import Ideal.Notation.
+Local Open Scope ring_scope.
+
+Section ChineseRemainderTheorem.
+
+  (** We assume [Univalence] in order to work with quotients. We also need it for [Funext] in a few places.*)
+  Context `{Univalence}
+    (** We need two coprime ideals [I] and [J] to state the theorem. We don't introduce the coprimeness assumption as of yet in order to show something slightly stronger. *)
+    {R : CRing} (I J : Ideal R).
+
+  (** We begin with the homomorphism which will show to be a surjection. Using the first isomorphism theorem for rings we can improve this to be the isomorphism we want. *)
+  (** TODO: rename *)
+  (** This is the corecursion of the two quotient maps *)
+  Definition rng_homo_crt : R $-> (R / I) × (R / J).
+  Proof.
+    apply cring_product_corec.
+    1,2: apply rng_quotient_map.
+  Defined.
+
+  (** Since we are working with quotients, we make the following notation to make working with the proof somewhat easier. *)
+  Notation "[ x ]" := (rng_quotient_map _ x).
+
+  (** We then need to prove the following lemma. The hypotheses here can be derived by coprimality of [I] and [J]. But we don't need that here. *)
+  Lemma issurjection_rng_homo_crt' (x y : R)
+    (q1 : rng_homo_crt x = (0, 1)) (q2 : rng_homo_crt y = (1, 0))
+    : IsSurjection rng_homo_crt.
+  Proof.
+    (** In order to show that [rng_homo_crt] is a surjection, we need to show that its propositional truncation of the fiber at any point is contractible. *)
+    intros [a b].
+    revert a; srapply QuotientRing_ind_hprop; intro a.
+    revert b; srapply QuotientRing_ind_hprop; intro b.
+    (** We can think of [a] and [b] as the pair [(a mod I, b mod J)]. We need to show that there merely exists some element in [R] that gets mapped by  [rng_homo_crt] to the pair. *)
+    snrapply Build_Contr; [|intros z; strip_truncations; apply path_ishprop].
+    (** We make this choice and show it maps as desired. *)
+    apply tr; exists (b * x + a * y).
+    (** Finally using some simple ring laws we can show it maps to our pair. *)
+    rewrite rng_homo_plus.
+    rewrite 2 rng_homo_mult.
+    rewrite q1, q2.
+    apply path_prod.
+    + change ([b] * 0 + [a] * 1 = [a] :> R / I).
+      by rewrite rng_mult_one_r, rng_mult_zero_r, rng_plus_zero_l.
+    + change ([b] * 1 + [a] * 0 = [b] :> R / J).
+      by rewrite rng_mult_one_r, rng_mult_zero_r, rng_plus_zero_r.
+  Defined.
+
+  (** Now we show that if [x + y = 1] for [I x] and [J y] then we can satisfy the hypotheses of the previous lemma. *)
+  Section rng_homo_crt_beta.
+
+    Context (x y : R) (ix : I x) (iy : J y) (p : x + y = 1).
+
+    Lemma rng_homo_crt_beta_left : rng_homo_crt x = (0, 1).
+    Proof.
+      apply rng_moveR_Mr in p.
+      rewrite rng_plus_comm in p.
+      apply path_prod; apply qglue.
+      - change (I (-x + 0)).
+        apply ideal_in_negate_plus.
+        1: assumption.
+        apply ideal_in_zero.
+      - change (J (-x + 1)).
+        rewrite rng_plus_comm.
+        by rewrite <- p.
+    Defined.
+
+    Lemma rng_homo_crt_beta_right : rng_homo_crt y = (1, 0).
+    Proof.
+      apply rng_moveR_rM in p.
+      rewrite rng_plus_comm in p.
+      apply path_prod; apply qglue.
+      - change (I (-y + 1)).
+        by rewrite <- p.
+      - change (J (-y + 0)).
+        apply ideal_in_negate_plus.
+        1: assumption.
+        apply ideal_in_zero.
+    Defined.
+
+  End rng_homo_crt_beta.
+
+  (** We can now show the map is surjective from coprimality of [I] and [J]. *)
+  Global Instance issurjection_rng_homo_crt
+    : Coprime I J -> IsSurjection rng_homo_crt.
+  Proof.
+    intros c.
+    (** First we turn the coprimality assumption into an equivalent assumption about the mere existence of two elements of each ideal which sum to one. *)
+    apply equiv_coprime_sum in c.
+    (** Since the goal is a hprop we may strip the truncations. *)
+    strip_truncations.
+    (** Now we can break apart the data of this witness. *)
+    destruct c as [[[x ix] [y jy]] p];
+    change (x + y = 1) in p.
+    (** Now we apply all our previous lemmas *)
+    apply (issurjection_rng_homo_crt' x y).
+    1: exact (rng_homo_crt_beta_left x y ix jy p).
+    exact (rng_homo_crt_beta_right x y ix jy p).
+  Defined.
+
+  (** Now suppose [I] and [J] are coprime. *)
+  Context (c : Coprime I J).
+
+  (** The Chinese Remainder Theorem *)
+  Theorem chinese_remainder : R / (I ∩ J)%ideal ≅ (R / I) × (R / J).
+  Proof.
+    (** We use the first isomorphism theorem. Coq can already infer which map we wish to use, so for clarity we tell it not to do so. *)
+    snrapply rng_first_iso'.
+    1: rapply rng_homo_crt.
+    1: exact _.
+    (** Finally we must show the ideal of this map is the intersection. *)
+    apply ideal_subset_antisymm.
+    - intros r [i j].
+      apply path_prod; apply qglue.
+      1: change (I (- r + 0)).
+      2: change (J (- r + 0)).
+      1,2: rewrite rng_plus_comm.
+      1,2: apply ideal_in_plus_negate.
+      1,3: apply ideal_in_zero.
+      1,2: assumption.
+    - intros i p.
+      apply equiv_path_prod in p.
+      destruct p as [p q].
+      apply ideal_in_negate'.
+      rewrite <- rng_plus_zero_r.
+      (** Here we need to derive the relation from paths in the quotient. This is what requires univalence. *)
+      split.
+      1: exact (related_quotient_paths _ _ _ p).
+      1: exact (related_quotient_paths _ _ _ q).
+  Defined.
+
+  (** We also have the same for products of ideals. *)
+  Theorem chinese_remainder_prod : R / (I ⋅ J)%ideal ≅ (R / I) × (R / J).
+  Proof.
+    etransitivity.
+    { rapply rng_quotient_invar.
+      symmetry.
+      rapply ideal_intersection_is_product. }
+    rapply chinese_remainder.
+  Defined.
+
+End ChineseRemainderTheorem.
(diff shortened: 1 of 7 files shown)


Alizter:
rebased

theories/Algebra/Rings/ChineseRemainder.v
```diff
@@ -0,0 +1,150 @@
...
+
+Section ChineseRemainderTheorem.
+
+  (** We assume [Univalence] in order to work with quotients. We also need it for [Funext] in a few places.*)
+  Context `{Univalence}
+    (** We need two coprime ideals [I] and [J] to state the theorem. We don't introduce the coprimeness assumption as of yet in order to show something slightly stronger. *)
+    {R : CRing} (I J : Ideal R).
+
+  (** We begin with the homomorphism which will show to be a surjection. Using the first isomorphism theorem for rings we can improve this to be the isomorphism we want. *)
+  (** TODO: rename *)
```
Alizter, line 22:
```suggestion
```

theories/Algebra/Rings/ChineseRemainder.v
```diff
@@ -0,0 +1,150 @@
...
+  Theorem chinese_remainder_prod : R / (I ⋅ J)%ideal ≅ (R / I) × (R / J).
+  Proof.
+    etransitivity.
+    { rapply rng_quotient_invar.
+      symmetry.
+      rapply ideal_intersection_is_product. }
+    rapply chinese_remainder.
+  Defined.
+
+End ChineseRemainderTheorem.
```
Alizter, line 150:
```suggestion
End ChineseRemainderTheorem.

```

theories/Algebra/Rings/Ideal.v
```diff
@@ -20,40 +26,153 @@ Record Ideal (R : CRing) := {
...
-  (** The zero ideal is an ideal *)
-  Global Instance isideal_trivial_subgroup
-    : IsIdeal (R:=R) trivial_subgroup.
+  Definition ideal_in_zero : I cring_zero := subgroup_unit I.
+  Definition ideal_in_plus a b : I a -> I b -> I (a + b) := subgroup_op I a b.
+  Definition ideal_in_negate a : I a -> I (- a) := subgroup_inverse I a.
+  Definition ideal_in_plus_negate a b : I a -> I b -> I (a - b)
+    := subgroup_op_inverse I a b.
+
+  Lemma ideal_in_negate' a : I (- a) -> I a.
```
mikeshulman, line 49: Shouldn't this and the next two lemmas be instances of general facts about subgroups?

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+
+(** The annihilator of an ideal. *)
+Definition ideal_annihilator `{Funext} {R : CRing} (I : Ideal R) : Ideal R
+  := ideal_quotient (ideal_zero R) I.
+
+(** *** Ideal notations *)
+
+(** We declare a module for various (unicode) ideal notations. These exist in their own special case, and can be imported and used when needing to reason about ideals. *)
+
+(** TODO: reserve these properly *)
```
mikeshulman, line 324: Do you want to do this before merging?

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+    left.
+    by apply isideal.
+  - apply sgt_in.
+    right.
+    by apply isideal.
+Defined.
+
+(** *** Ideal lemmas *)
+
+(** We use these notations in the rest of this file *)
```
mikeshulman, line 408: Very minor, but "these" is a bit confusing here since this comment is somewhat distant from the notation module.  Is there a reason not to define and import the notation module at the same time?

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+
+  Context {R : CRing}.
+
+  (** Subset relation is antisymmetric *)
+  Lemma ideal_subset_antisymm (I J : Ideal R) : I ⊆ J -> J ⊆ I -> I ↔ J.
+  Proof.
+    intros p q x; split; by revert x.
+  Defined.
+
+  (** The zero ideal is contained by all ideals *)
```
mikeshulman, line 422:
```suggestion
  (** The zero ideal is contained in all ideals *)
```

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+  Proof.
+    etransitivity.
+    2: rapply ideal_sum_self.
+    etransitivity.
+    2: rapply ideal_sum_subset_pres_r.
+    2: rapply ideal_product_comm.
+    apply ideal_product_intersection_sum_subset.
+  Defined.
+
+  (** If the sum of ideals is the whole ring then their product is a subset of their intersection. *)
```
mikeshulman, line 748:
```suggestion
  (** If the sum of ideals is the whole ring then their intersection is a subset of their product. *)
```

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+      induction q as [r x | | ].
+      { destruct x.
+        cbv in p.
+        by apply p. }
+      1: apply ideal_in_zero.
+      by apply ideal_in_plus_negate.
+    Defined.
+
+    (** Ideal quotients partially cancel *)
+    Lemma ideal_quotient_susbset_left_inverse (I J : Ideal R)
```
mikeshulman, line 805:
```suggestion
    Lemma ideal_quotient_subset_left_inverse (I J : Ideal R)
```
However, I'm not sure I would describe this as a "left inverse" or as "canceling", since it is an inequality rather than an equality.

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+    Defined.
+
+    (** The ideal quotient of unit by I is unit *)
+    Lemma ideal_quotient_unit_top (I : Ideal R)
+      : (ideal_unit R :: I) ↔ ideal_unit R.
+    Proof.
+      cbv; split; trivial.
+    Defined.
+
+    (** The ideal quotient by a sum is an intersecton of ideal quotients *)
```
mikeshulman, line 840:
```suggestion
    (** The ideal quotient by a sum is an intersection of ideal quotients *)
```

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+
+    (** Ideals are subsets of their ideal quotients *)
+    Lemma ideal_quotient_subset (I J : Ideal R) : I ⊆ (I :: J).
+    Proof.
+      intros x i r j.
+      rewrite rng_mult_comm.
+      by apply isideal.
+    Defined.
+
+    (** Products are subsets iff the first factor is a subset of the ideal quotient *)
```
mikeshulman, line 787:
```suggestion
    (** The ideal quotient is a right adjoint to the product in the monoidal lattice of ideals. *)
```

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+    intros p.
+    apply ideal_sum_smallest.
+    1: apply ideal_sum_subset_l.
+    transitivity J.
+    1: exact p.
+    apply ideal_sum_subset_r.
+  Defined.
+
+  (** Products left distribute over sums *)
+  Lemma ideal_dist_l (I J K : Ideal R) : I ⋅ (J + K) ↔ I ⋅ J + I ⋅ K.
```
mikeshulman, line 571: On paper, this follows by abstract nonsense from the adjunction `ideal_quotient_subset_prod` (left adjoints preserve colimits).  I suppose we might want to avoid using the latter since it involves funext, but if we weren't worried about that, would it simplify the formalization to derive this lemma that way?

theories/Algebra/Rings/Ideal.v
mikeshulman: This should also follow formally from `ideal_quotient_subset_prod`.  Would that proof be any easier?

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+          rewrite rng_mult_negate_r.
+          by apply ideal_in_plus_negate. }
+      intros x p y k z j; hnf in p.
+      rewrite <- rng_mult_assoc.
+      rewrite (rng_mult_comm y).
+      by apply p, tr, sgt_in, ipn_in.
+    Defined.
+
+    (** Ideal quotients distribute over intersections *)
+    Lemma ideal_quotient_intersection (I J K : Ideal R)
```
mikeshulman, line 889: This should also follow formally from `ideal_quotient_subset_prod`.  But it's hard to imagine that proof being much shorter here.

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +184,728 @@ Proof.
...
+        1: by apply p.
+        by apply q.
+      - rewrite rng_mult_zero_r.
+        apply ideal_in_zero.
+      - rewrite rng_dist_l.
+        rewrite rng_mult_negate_r.
+        by apply ideal_in_plus_negate.
+    Defined.
+
+    Lemma ideal_quotient_product (I J K : Ideal R)
```
mikeshulman, line 866: This should also follow formally from `ideal_quotient_subset_prod` and a lemma that the product of ideals is associative.  I didn't see the latter lemma, is it missing?

theories/Algebra/Rings/QuotientRing.v
```diff
@@ -102,58 +106,88 @@ Section QuotientRing.
...
-Definition Build_CRing' (R : AbGroup)
-  `(Mult R, One R, LeftDistribute R mult (@group_sgop R))
-  (iscomm : @IsCommutativeMonoid R mult one)
-  : CRing
-  := Build_CRing R (@group_sgop R) _ (@group_unit R) _
-       (@group_inverse R) (Build_IsRing _ _ _ _).
+Infix "/" := QuotientRing : ring_scope.
 
-(** The image of a ring homomorphism *)
-Definition rng_image {R S : CRing} (f : R $-> S) : CRing.
```
mikeshulman, line 114: Did this get moved to another file?

mikeshulman (approved):
This looks good!  Sorry to take so long to look at it.

mikeshulman:
i'm glad you aren't actually using the name "colon ideal" -- in general I don't like names derived from notations.  Unfortunately, the term "quotient ideal" isn't much better in my mind: a quotient is a colimit, but this is more like an internal-hom.  I would have been more inclined to call it a [residual ideal](https://en.wikipedia.org/wiki/Residuated_lattice) -- do you think we could get away with using that terminology?

theories/Algebra/Rings/Ideal.v
Alizter: Ah good observation, hadn't noticed that!

theories/Algebra/Rings/Ideal.v
Alizter: I think it would be worth deriving the result this way. Since we don't quite have the machinery of adjunctions ready, I will make a note here.

theories/Algebra/Rings/QuotientRing.v
Alizter: Yes, it was moved to CRing.v

theories/Algebra/Rings/Ideal.v
Alizter: Yes, good catch. I have proven them for the predicate `issubgroup` and derived the versions for `subgroup` and `ideal` later on. The naming convention is now consistently `subgroup_in_X`.

theories/Algebra/Rings/Ideal.v
Alizter: Done.

theories/Algebra/Rings/Ideal.v
Alizter: This was to avoid conflict with the coprime lemmas. I've reordered things and imported immediately now.

theories/Algebra/Rings/Ideal.v
Alizter: OK, I've renamed it `ideal_quotient_product_left`.

theories/Algebra/Rings/Ideal.v
Alizter: I didn't prove associativity because it looked hard and I didn't need it. 0:-)

Alizter: I've added a TODO in the file in case anybody wants to try.

theories/Algebra/Rings/Ideal.v
Alizter: I don't quite see how this follows.

theories/Algebra/Rings/Ideal.v
mikeshulman: Well, if this is easier to prove directly than associativity, we should conversely be able to deduce associativity from it.

theories/Algebra/Rings/Ideal.v
mikeshulman: By adjunction again: `(I :: -)` is a contravariant left adjoint, so it sends coproducts to products.

Alizter:
Regarding the terminology: I didn't know about residual lattices and I like the idea of calling these residual ideals. It seems to be used in constructive algebra. However I am still weary about renaming it. For a lot of algebra, my standard reference for is Dummit & Foote, there they call it a quotient ideal. In algebraic geometry, the term residue gets thrown in for a different reason, which might generate confusion in the future.

At the end of the day, this can be considered a bit of a fringe concept anyhow that I haven't seen much used outside of ring theory. Also there is a similar concept of a fractional ideal generalizes a lot of things here, but at the level of modules.

Perhaps leaving a comment like: `Quotient as in division not colimit` might be appropriate? The reason being in Z we have `(6) : (3) = (2)`.

Alizter:
@mikeshulman Is this good to merge now?

I think it would definitely be interesting to investigate the (closed) monoidal lattice of ideals once we have the appropriate categorical machinery in place.

mikeshulman:
Sure, with a comment about "quotient as in division" I'm fine.

theories/Algebra/Rings/Ideal.v
```diff
@@ -65,7 +164,728 @@ Proof.
...
+  exact (q o p).
+Defined.
+
+Coercion ideal_eq_subset {R : CRing} {I J : Ideal R} : ideal_eq I J -> ideal_subset I J.
+Proof.
+  intros f x; apply f.
+Defined.
+
+(** Quotient (a.k.a colon) ideal *)
+(** Unfortunately, due to truncatedness constraints, we need to assume funext. *)
```
Alizter, line 274:
```suggestion
(** Quotient (a.k.a colon) ideal *)
(** Note that this is quotient as in division rather than a colimit. In fact, the quotient ideal is more like an internal hom as we will see later. *)
(** Unfortunately, due to truncatedness constraints, we need to assume funext. *)
```

Issue #1494, opened 2021-04-27
CI failing due to changes in coq master


Alizter:
Our CI build for coq master no longer works since the makefile targets called in `install_coq.sh` are no longer in use.

I suggest that this is a problem with the CI rather than that script. We should no longer use the coq submodule in the CI as @JasonGross had suggested in #1388.

PR #1498, opened 2021-04-29
Make x and y implicit args in (is)equiv_transport

jdchristensen:
This makes them match `transport`.  Those args were only needed in one spot in the library, so I had to use the `@` notation there.  But many other places are simpler now.

### DIFF
--- a/theories/Truncations/Connectedness.v
+++ b/theories/Truncations/Connectedness.v
@@ -127,11 +127,7 @@ Definition merely_path_is0connected `{Univalence}
            (A : Type) `{IsConnected 0 A} (x y : A)
 : merely (x = y).
 Proof.
-  (** This follows immediately from the previous result. *)
-  rapply center.
-(* Old proof:
   refine ((equiv_path_Tr x y)^-1 (path_contr (tr x) (tr y))).
-*)
 Defined.
 
 Definition is0connected_merely_allpath `{Univalence}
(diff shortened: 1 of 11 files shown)


mikeshulman:
This looks good, but I would be inclined to keep the explicit arguments in BlakersMassey, for readability.

theories/Truncations/Connectedness.v
```diff
@@ -127,7 +127,11 @@ Definition merely_path_is0connected `{Univalence}
            (A : Type) `{IsConnected 0 A} (x y : A)
 : merely (x = y).
 Proof.
+  (** This follows immediately from the previous result. *)
+  rapply center.
+(* Old proof:
```
Alizter, line 132: Is there any point keeping the old proof anymore?

jdchristensen: Sorry, didn't mean this to be part of this commit.  I'll remove it.

theories/Truncations/Connectedness.v
```diff
@@ -127,11 +127,7 @@ Definition merely_path_is0connected `{Univalence}
            (A : Type) `{IsConnected 0 A} (x y : A)
 : merely (x = y).
 Proof.
-  (** This follows immediately from the previous result. *)
-  rapply center.
-(* Old proof:
   refine ((equiv_path_Tr x y)^-1 (path_contr (tr x) (tr y))).
```
Alizter: Why get rid of the new proof?

jdchristensen: Just because it's unrelated to this PR.

Alizter (approved):
LGTM

Alizter:
Don't worry about the "building coq master job", I am currently in the process of fixing it. #1494.

jdchristensen:
I'm counting Mike's comment as approval.

PR #1499, opened 2021-04-29
Reduce universe levels in equiv_induction_inv and dependencies.

jdchristensen:
This slight tweak to the proof reduces the number of universe levels required, for some reason.  In a downstream application, the old proof forced another universe strictly larger than the others.

The changes to `contr_basedequiv` and `contr_basedequiv'` are just "documentation" to the reader, since nothing in the statement forces those universes to be equal.  It only happens in the proof, implicitly.

### DIFF
--- a/theories/Types/Universe.v
+++ b/theories/Types/Universe.v
@@ -489,7 +489,8 @@ Theorem equiv_induction_inv {U : Type} (P : forall V, V <~> U -> Type) :
 Proof.
   intros H0 V.
   apply (equiv_ind (equiv_path V U)).
-  intro p; induction p; apply H0.
+  (* We manually apply [paths_ind_r] to reduce universe levels. *)
+  revert V; rapply paths_ind_r; apply H0.
 Defined.
 
 Definition equiv_induction_inv_comp {U : Type} (P : forall V, V <~> U -> Type)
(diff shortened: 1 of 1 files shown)


jdchristensen:
`equiv_induction` and `equiv_induction'` already have weaker universe requirements, so I don't think any changes are needed to them.  After this change, `contr_based_equiv` and `contr_based_equiv'` (which use `equiv_induction` and `equiv_induction_inv`, respectively) both have `(* u u0 |= u < u0 *)`, which is sensible.  Before this change, the latter had `(* u u0 u1 u2 |= u1 < u = u0 < u2 *)`.

theories/Types/Universe.v
```diff
@@ -499,16 +500,16 @@ Definition equiv_induction_inv_comp {U : Type} (P : forall V, V <~> U -> Type)
 
 (** ** Based equivalence types *)
 
-Global Instance contr_basedequiv {X : Type}
-: Contr {Y : Type & X <~> Y}.
+Global Instance contr_basedequiv@{u +} {X : Type@{u}}
```
mikeshulman, line 503: What does the bare `+` mean?

mikeshulman: That does seem handy.  Is that documented?  I just skimmed through the section on universe polymorphism in the Coq manual and didn't see it, but maybe I missed it.

jdchristensen: At https://coq.inria.fr/refman/addendum/universe-polymorphism.html it says:
```
It is possible to provide only some universe levels and let Coq infer the others by adding a + in the list of bound universe levels:
```
and gives an example.

jdchristensen: An added benefit of the annotation in this case is that in place of two universe variables constrained to be equal (and so always duplicated when passed in), we have one variable u.

Issue #1388, opened 2020-08-27
Our .git folder is huge


Alizter:
### Issue

On the coq Zulip it was noticed that the git history file for the HoTT repo is larger than 600MB, for reference the .git history for the coq repo is 152MB.

This seems to be because we have a bot autoupdating static documentation for every commit to the `gh-pages` branch.

### Proposed solution

1. Move the `gh-pages` branch to a seperate repository.
2. Delete the `gh-pages` branch in this repository, this will delete all the history.
3. Configure the `gh-pages` branch in the new repository to autoupdate for every commit.
4. We could also add a new `gh-pages` branch back but only update documentation on each (coq) release. Though I am not sure how useful this would be.

@JasonGross I believe you know the most about this set up, what do you think?

JasonGross:
Why not just clone with `--single-branch` or similar so that you don't fetch the gh-pages on cloning/fetching?

If that's not sufficient, I think it's fine to move the gh-pages branch to a separate repo, say, HoTT-gh-pages-history or something.  One possibility is that we could have the separate repo include all history, while in this repo we'd blow away the history of gh-pages on each commit.

mikeshulman:
I think it'd be better not to expect casual users to know about something
esoteric like --single-branch.  (-:

jdchristensen:
Wow, --single-branch reduces the .git folder from 625MB to 11MB!  I definitely think we should do something to make things work more efficiently by default.  In the meantime, is there a way to save this space in my current clone, without losing local branches, etc?  I tried
```
  git gc
  # .git is ~1GB here.
  git branch -d -r upstream/gh-pages
  git branch -d -r origin/gh-pages
  git branch -d -r jarlg/gh-pages
  # .git is ~1GB here.
  git gc
  # .git went from 1GB to 2.7GB!
  git reflog expire --expire-unreachable=now
  git gc
  # Still 2.7GB!
```
Not sure how git could blow up the .git folder just because I deleted some remote tracking branches!

jdchristensen:
One more command helped a lot:
```
git gc prune=now
```
Now down to 383MB.

JasonGross:
I was just about to comment with `git gc --prune=now`, but I see you found it first.  Running `du -h .git | sort -h` on my machine tells me that about 239M is in the submodule coq-HoTT, so that's probably a significant chunk of where your space usage is coming from.  If you get rid of tags (things shown by `git tag`) and then prune again, that might get rid of a bit more.

[...]

PR #1502, opened 2021-04-29
Update CI to use docker images

### LINKED ISSUE #1494: CI failing due to changes in coq master

Alizter:
I've updated the CI to use docker images (basically precompiled versions of coq). I've gotten rid of the old build jobs that use the submodule. We only leave the quickdoc, html, dpdgraph and validate together with alyectron.

Note that we are building the library on the following versions:
 * Our current supported version 8.13
 * The latest stable version of coq
 * The dev version of coq

<s> I don't think it makes sense to try the dev version as that can be unstable.</s>

The advantage of doing things like this is that we can build the library in around 4min!

In the future, we want the build jobs to upload their image as an artifact so that the documentation jobs can work directly with that. Something similar was suggested by Jason in https://github.com/HoTT/HoTT/issues/1388#issuecomment-826792147.

closes #1494

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,8 @@ jobs:
           env: { UPDATE_QUICK_DOC: "yes" }
         - label: update html
           env: { UPDATE_HTML: "yes", UPDATE_QUICK_DOC: "" }
-        
+        - label: validate
+          env: { VALIDATE: "yes"   , UPDATE_QUICK_DOC: "" }
     env: ${{ matrix.env }}
 
     # The operating system
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,8 @@ jobs:
           startGroup "Workaround permission issue" # https://github.com/coq-community/docker-coq-action#permissions
           sudo chown -R coq:coq .
           endGroup
-          etc/coq-scripts/github/reportify-coq.sh --errors make TIMED=1 -j2     --output-sync GH_REPORT_ERRORS=1
+          etc/coq-scripts/github/reportify-coq.sh --stdout --errors ${{ matrix.extra_gh_reportify }} etc/coq-scripts/timing/make-pretty-timed.sh -j2 TIMING=1
+          make
     - name: Revert permissions
       # to avoid a warning at cleanup time - https://github.com/coq-community/docker-coq-action#permissions
       if: ${{ always() }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         # We build on our supported version of coq and the master version
-        coq_version: [ '8.13' , 'latest' ]
+        coq_version: [ '8.13' , 'latest' , 'dev' ]
         ocaml_version: [ '4.11-flambda' ]
         
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 ## This is the name of the CI
 name: CI
 
-## These are the tiggers for the CI to run
+## These are the triggers for the CI to run
 on: [ push , pull_request ]
 
 jobs:
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # We build on our supported version of coq and master
+        # We build on our supported version of coq and the master version
         coq_version: [ '8.13' , 'latest' ]
         ocaml_version: [ '4.11-flambda' ]
         
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ jobs:
         # We build on our supported version of coq and the master version
         coq_version: [ '8.13' , 'latest' , 'dev' ]
         ocaml_version: [ '4.11-flambda' ]
+        include:
+        - coq_version: 'dev'
+          extra_gh_reportify: '--warnings'
         
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,8 +30,10 @@ jobs:
         ocaml_version: ${{ matrix.ocaml_version }}
         custom_script: |
           startGroup "Workaround permission issue" # https://github.com/coq-community/docker-coq-action#permissions
-          sudo chown -R coq:coq .
+            sudo chown -R coq:coq .
           endGroup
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
 ## This is the name of the CI
 name: CI
 
-## These are the actions that trigger the CI to run
-on: [push, pull_request]
+## These are the tiggers for the CI to run
```
Alizter, line 4:
```suggestion
## These are the triggers for the CI to run
```

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
+on: [ push , pull_request ]
 
 jobs:
+
   build:
-    # TODO: Eventually include different OS's in the matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        # We build on our supported version of coq and master
```
Alizter, line 13:
```suggestion
        # We build on our supported version of coq and the master version
```

JasonGross:
> I don't think it makes sense to try the dev version as that can be unstable.

No, part of the contract we must uphold to continue being tested on Coq's CI is to test against the tip of Coq master.  So we should test the dev version as well.

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
         include:
-        - env: { UPDATE_HTML: "", UPDATE_QUICK_DOC: "", UPDATE_DEP_GRAPHS: "", BUILD_COQ: "", FORCE_COQ_VERSION: "", VALIDATE: "" }
         - label: update dep graphs
           env: { UPDATE_DEP_GRAPHS: "yes", BUILD_COQ: "yes", UPDATE_QUICK_DOC: "" }
         - label: update quick doc
           env: { UPDATE_QUICK_DOC: "yes" }
         - label: update html
           env: { UPDATE_HTML: "yes", UPDATE_QUICK_DOC: "" }
-        - label: validate
-          env: { VALIDATE: "yes"   , UPDATE_QUICK_DOC: "" }
```
JasonGross, line 23: Why'd you remove the validate target?

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
+    # We use the coq docker so we don't have to build coq
+    - uses: coq-community/docker-coq-action@v1
+      with:
+        coq_version: ${{ matrix.coq_version }}
+        ocaml_version: ${{ matrix.ocaml_version }}
+        custom_script: |
+          startGroup "Workaround permission issue" # https://github.com/coq-community/docker-coq-action#permissions
+          sudo chown -R coq:coq .
+          endGroup
+          etc/coq-scripts/github/reportify-coq.sh --errors make TIMED=1 -j2     --output-sync GH_REPORT_ERRORS=1
```
JasonGross, line 32: We should also pass `--warnings` iff we're building for Coq master.  And I think we don't want `GH_REPORT_ERRORS=1`
```suggestion
          etc/coq-scripts/github/reportify-coq.sh --stdout --errors ${{ matrix.extra_gh_reportify }} etc/coq-scripts/timing/make-pretty-timed.sh -j2 TIMING=1
          make
```

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
 
 jobs:
+
   build:
-    # TODO: Eventually include different OS's in the matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        # We build on our supported version of coq and the master version
+        coq_version: [ '8.13' , 'latest' ]
```
Alizter, line 14:
```suggestion
        coq_version: [ '8.13' , 'latest' , 'dev' ]
```

Alizter:
@JasonGross OK fair enough, I've also added the dev version.

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
+        ocaml_version: [ '4.11-flambda' ]
```
JasonGross: I think this will work?
```suggestion
        coq_version: [ '8.13' , 'latest' ]
        ocaml_version: [ '4.11-flambda' ]
        include:
        - coq_version: '8.13'
          extra_gh_reportify: '--warnings'
```

JasonGross (changes requested):

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
 jobs:
+
   build:
-    # TODO: Eventually include different OS's in the matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        # We build on our supported version of coq and the master version
+        coq_version: [ '8.13' , 'latest' , 'dev' ]
+        ocaml_version: [ '4.11-flambda' ]
```
JasonGross, line 15: I think this will work?
```suggestion
        coq_version: [ '8.13' , 'latest' , 'dev' ]
        ocaml_version: [ '4.11-flambda' ]
        include:
        - coq_version: 'dev'
          extra_gh_reportify: '--warnings'
```

.github/workflows/ci.yml
Alizter: Ah that wasn't intentional.

.github/workflows/ci.yml
```diff
@@ -1,32 +1,53 @@
...
           env: { UPDATE_HTML: "yes", UPDATE_QUICK_DOC: "" }
-        - label: validate
-          env: { VALIDATE: "yes"   , UPDATE_QUICK_DOC: "" }
-        - label: with submodule coq
-          env: { FORCE_COQ_VERSION: ""      , BUILD_COQ: "yes", UPDATE_QUICK_DOC: "" }
-        - label: building coq master
-          env: { FORCE_COQ_VERSION: "master", BUILD_COQ: "yes", UPDATE_QUICK_DOC: "", EXTRA_GH_REPORTIFY: "--warnings" }
-          display-warnings: true
-
+
```
Alizter, line 54:
```suggestion
        - label: validate
          env: { VALIDATE: "yes"   , UPDATE_QUICK_DOC: "" }
```

.github/workflows/ci.yml
Alizter: I've added it back in.

jdchristensen:
This seems like a nice approach!

Alizter:
@JasonGross What exactly does validate do?

JasonGross:
validate runs the .vo files through coqchk.  I like adding it to my jobs as a service to the coqdevs to give coqchk more coverage

JasonGross (approved):

PR #1503, opened 2021-05-01
Syllepsis

Alizter:
Here is Kristina's proof of the syllepsis. It can be merged as is, but it was written without using existing constructs in the library, so it might be worth cleaning it up a little before. Adding some documentation may also help, but I don't fully understand the proof as is. We need to also change the names which aren't very descriptive.

I've created this draft PR so we can have a look. Since it is a result that hasn't been written down in HoTT before. (At least that I am aware of).

### DIFF
--- a/theories/Homotopy/Syllepsis.v
+++ b/theories/Homotopy/Syllepsis.v
@@ -1,0 +1,267 @@
+Require Import Basics Types.
+
+Local Definition C {X} {a : X} {x y u v : a = a} p q :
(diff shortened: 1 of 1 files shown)


jdchristensen:
Has anyone asked Kristina if she's willing to add some comments and choose some names that clarify things a bit?

jdchristensen:
Tagging @kristinas in case she wants to comment.

kristinas:
Yes, I didn't see this until now. I agree the names aren't very good. Any suggestions? Also, what other constructs are in the library that I could make use of?

PR #1507, opened 2021-05-06
Add more backtracking to make_equiv and make_equiv_contr_basedpaths

### LINKED ISSUE #1506: make_equiv getting confused
The following works fine:
```coq
Definition good : pType <~> { X : Type & X}.
make_equiv.
```
But the following fails:
```coq
Definition bad (Y : pType): pType <~> { X : Type & X}.
Fail make_equiv.  (* No applicable tactic. *)
```
Somehow, the extra `pType` in the context causes a problem, and the backtracking search doesn't find a valid proof.  And while this example is contrived, this came up in practice in a deeply nested equivalence that should be easy for `make_equiv` but  is tedious to do manually.  Is there an easy workaround?

jdchristensen:
Replace `first` with `+` in `build_record` and `build_record_with_evars` so that `make_equiv` and `make_equiv_contr_basedpaths` do more backtracking in their search.  Also factor out two `cbn`s, so they aren't undone and redone in some cases.

I checked the timing, and it makes no significant difference for the building of the library.

Closes #1506.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -578,8 +578,7 @@ Ltac multi_assumption :=
 
 (* Build an element of a possibly-nested record type out of hypotheses in the context. *)
 Ltac build_record :=
-  first [ cbn; multi_assumption
-        | cbn; unshelve econstructor; build_record ].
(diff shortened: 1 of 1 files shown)


Alizter (approved):
LGTM

Issue #1452, opened 2021-04-13
Unify `BAut X` with `Type_ O`?


jdchristensen:
`BAut X` is defined to be `{ Z : Type & merely (Z = X) }`.

When `O : Type -> Type` is a `Subuniverse` (so its values are hprops), `Type_ O` is defined to be `{ Z : Type & O Z }` (roughly).

If we fix `X` and define `O Z` to be `merely (Z = X)`, then the two coincide.  Except that `O Z` is in a larger universe than `Z`, which isn't allowed by the definition of `Subuniverse`.

Would it be worth it to add another universe variable to `Subuniverse` saying where the hprops live?  They would infiltrate a lot of other code.  But many facts are true for both `BAut X` and `Type_ O`, so they could be proved only once.  This came up in practice, since `Type_ O` is the object classifier for fibrations with fiber in `O`, but one can't use that to study fibrations whose fiber is merely equal to `X`.

Maybe there are other situations in which one would like to consider a `Subuniverse` where the predicate lands in a different universe?  Or maybe this is rare enough that we should just keep `Type_ O` and `BAut X` separate?  Or maybe there can be `Subuniverse'` and `Type_'` that allow this flexibility, but are not used in the `Modalities` code?

mikeshulman:
What about redefining `BAut X` to be `{ Z : Type & merely (Z <~> X) }`?

jdchristensen:
I meant to mention that as an option as well.  It might be a fair bit of work adjusting all of the existing BAut proofs.  Any idea whether things would generally become simpler or more complicated?  It looks like essentially all results about BAut already assume Univalence, so the change wouldn't force use to use more Univalence (and might let us reduce it?)

mikeshulman:
No, I don't really have a guess.  Probably it would make some things more complicated but others less complicated, and I don't have a guess about which would predominate. It would be worth a try to find out, I think.

jdchristensen:
I agree that it would probably be worth trying to change the definition of `BAut` to use equivalences and see how it goes.

About adding an extra universe variable to `Subuniverse`, I just noticed that `FactorizationSystem` has two extra universe variables for the two predicates on the universe, so there is a precedent for allowing this flexibility.

jdchristensen:
[...]

PR #1508, opened 2021-05-09
Reduce universes in BAut from 19 to 3.

jarlg:
As noted by @Alizter [here](https://github.com/HoTT/HoTT/issues/1452#issuecomment-820810626), `BAut` has 19 universe variables. By proving connectedness more directly it is reduced to 3. This makes reasoning about the universe level of `BAut` easier, and resolved some unification errors I ran into when experimenting.

### DIFF
--- a/theories/Algebra/Aut.v
+++ b/theories/Algebra/Aut.v
@@ -3,21 +3,10 @@ Require Import Basics.
 Require Import Types.
 Require Import Truncations.
 Require Import Algebra.ooGroup.
-
-Local Open Scope path_scope.
+Require Import Spaces.BAut.
 
 (** * Automorphism oo-Groups *)
 
-(** To define the automorphism oo-group [Aut X], we have to construct its classifying space [BAut X]. *)
-
-(** [BAut X] is the type of types that are merely equal to [X]. *)
-Definition BAut@{u v} (X : Type@{u}) : Type@{v}
-  := sig@{v v} (fun Z => merely (paths@{v} Z X)).
-
-Global Instance ispointed_baut {X : Type} : IsPointed (BAut X) := (X; tr 1).
-
-(** Now we can define [Aut X], since [BAut X] is connected by [is0connected_component]. *)
+(** We define [Aut X] using the pointed, connected type [BAut X]. *)
 Definition Aut (X : Type) : ooGroup
   := Build_ooGroup (Build_pType (BAut X) _) _.
-
-(** The type [BAut X] is studied further in [Spaces.BAut] and its subdirectories. *)
--- a/theories/Algebra/Aut.v
+++ b/theories/Algebra/Aut.v
@@ -2,7 +2,6 @@
 Require Import Basics.
 Require Import Types.
 Require Import Truncations.
-Require Import Factorization.
 Require Import Algebra.ooGroup.
 
 Local Open Scope path_scope.
@@ -11,35 +10,13 @@ Local Open Scope path_scope.
 
 (** To define the automorphism oo-group [Aut X], we have to construct its classifying space [BAut X]. *)
 
-(** [BAut X] is the type of types that are merely equivalent to [X]. *)
+(** [BAut X] is the type of types that are merely equal to [X]. *)
 Definition BAut@{u v} (X : Type@{u}) : Type@{v}
   := sig@{v v} (fun Z => merely (paths@{v} Z X)).
 
 Global Instance ispointed_baut {X : Type} : IsPointed (BAut X) := (X; tr 1).
 
-(** Equivalently, [BAut X] is the (-1)-image of the classifying map [1 -> Type] of [X]. *)
-Definition equiv_baut_image_unit X
-: BAut X <~> image (Tr (-1)) (unit_name X).
-Proof.
-  unfold BAut, image; simpl.
-  apply equiv_functor_sigma_id; intros Z; simpl.
-  apply Trunc_functor_equiv; unfold hfiber.
-  refine ((equiv_contr_sigma _)^-1 oE _).
-  apply equiv_path_inverse.
-Defined.
-
-Global Instance isconnected_baut {X : Type}
-  : IsConnected 0 (BAut X).
-Proof.
-  exists (tr (X; tr 1)).
-  rapply Trunc_ind; intros [Z p].
-  strip_truncations.
-  apply (ap tr).
-  rapply path_sigma_hprop.
-  exact p^.
-Defined.
-
-(** Now we can define [Aut X], since [BAut X] is connected. *)
+(** Now we can define [Aut X], since [BAut X] is connected by [is0connected_component]. *)
 Definition Aut (X : Type) : ooGroup
   := Build_ooGroup (Build_pType (BAut X) _) _.
 
--- a/theories/Algebra/Aut.v
+++ b/theories/Algebra/Aut.v
@@ -27,17 +27,20 @@ Proof.
   apply equiv_path_inverse.
 Defined.
 
-(** Now we can define [Aut X], by proving that [BAut X] is connected. *)
-Definition Aut (X : Type) : ooGroup.
+Definition isconnected_baut {X : Type}
+  : IsConnected 0 (BAut X).
 Proof.
-  refine (Build_ooGroup
-            (Build_pType { Z : Type & merely (Z = X) } (X ; tr 1)) _).
-  refine (conn_pointed_type (point _)); try exact _.
-  pose (c := conn_map_compose (Tr (-1))
-                              (factor1 (image (Tr (-1)) (unit_name X)))
-                              (equiv_baut_image_unit X)^-1).
-  refine (conn_map_homotopic _ _ _ _ c); intros []; reflexivity.
+  exists (tr (X; tr 1)).
+  rapply Trunc_ind; intros [Z p].
+  strip_truncations.
+  rapply path_Tr; apply tr.
+  apply (path_sigma' _ p^).
+  apply path_ishprop.
 Defined.
+
+(** Now we can define [Aut X], since [BAut X] is connected. *)
+Definition Aut (X : Type) : ooGroup
+  := Build_ooGroup (Build_pType (BAut X) (X; tr 1)) isconnected_baut.
 
 Definition BAut X : Type := classifying_space (Aut X).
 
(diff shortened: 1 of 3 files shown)


Alizter (approved):
Nice!

theories/Algebra/Aut.v
```diff
@@ -27,18 +27,21 @@ Proof.
...
-  pose (c := conn_map_compose (Tr (-1))
-                              (factor1 (image (Tr (-1)) (unit_name X)))
-                              (equiv_baut_image_unit X)^-1).
-  refine (conn_map_homotopic _ _ _ _ c); intros []; reflexivity.
+  exists (tr (X; tr 1)).
+  rapply Trunc_ind; intros [Z p].
+  strip_truncations.
+  rapply path_Tr; apply tr.
+  apply (path_sigma' _ p^).
+  apply path_ishprop.
```
jdchristensen, line 38: I like the new proof better, independent of the universe issues.  It can be made slightly cleaner:
```coq
  exists (tr (X; tr 1)).
  rapply Trunc_ind; intros [Z p].
  strip_truncations.
  apply (ap tr).
  apply path_sigma_hprop.
  exact p^.
```

theories/Algebra/Aut.v
```diff
@@ -27,18 +27,21 @@ Proof.
...
 Defined.
 
+(** Now we can define [Aut X], since [BAut X] is connected. *)
+Definition Aut (X : Type) : ooGroup
+  := Build_ooGroup (Build_pType (BAut X) (X; tr 1)) isconnected_baut.
+
 Definition BAut X : Type := classifying_space (Aut X).
```
jdchristensen, line 45: It's possible to get `BAut` down to two universe variables.  Since the extra "middle" universe doesn't appear in the hypotheses or conclusion, I think it can't give any extra generality.
```coq
Definition BAut@{u v} X : Type@{v} := classifying_space (Aut@{u v} X).
```

theories/Algebra/Aut.v
```diff
@@ -27,18 +27,21 @@ Proof.
   apply equiv_path_inverse.
 Defined.
 
-(** Now we can define [Aut X], by proving that [BAut X] is connected. *)
-Definition Aut (X : Type) : ooGroup.
+Definition isconnected_baut {X : Type}
+  : IsConnected 0 (BAut X).
```
jdchristensen, line 31: I wonder if this should be a Global Instance?  If someone is using the BAut from the end of this file, then it won't matter, but what if someone is using BAut.BAut?  (I find it confusing that the name is reused...)

Alizter: I think the idea is that nobody should be using BAut.BAut.

jarlg: I don't see any reason to reach for `Algebra.Aut.BAut` in this case. If we had used equivalences in defining `BAut X` there would be a point, since the oo-group definition would require univalence. But in this case there is no reason and this would just be a superfluous instance, as far as I can tell.

jdchristensen: This change accidentally intertwines the two `BAut`s, since in the definition of `Aut X` it uses `BAut X` (referring to `BAut.BAut`) rather than writing out the sigma type like the old code had.  With the new code, if you have `BAut X` in your goal, `cbn` reduces it to `BAut.BAut X`.

Even with that fixed, I find this odd.  Maybe `BAut` should be a local definition in BAut.v (with a different name?), and then right at the end of BAut.v `BAut` could be defined to be the classifying space of `Aut X`.  As it stands, depending on what has been `Require`d at the top of a file, the meaning of `BAut` will be different, which will be confusing to the reader.

I guess the current situation might have to do with dependencies?

jdchristensen: Sorry, ignore my previous comment.  I thought BAut was defined once in BAut.v and again in Aut.v, but now I see that both definitions are in Aut.v, right near each other.  (I was misreading BAut.BAut as meaning a definition from the file BAut.v rather than from the Module BAut in Aut.v.)  So, now my question is, why have this Module at all?  The BAut it defines is only used in a couple of places, and one could just inline the sigma type.

Alizter: @jarlg and I have gotten rid of the two definitions and simplified it to one. The reason there were two before was so BAut could be pointed. We declared an instance of the type BAut being pointed and the proofs that rely on this go through.

theories/Algebra/Aut.v
```diff
@@ -11,10 +11,11 @@ Local Open Scope path_scope.
 
 (** To define the automorphism oo-group [Aut X], we have to construct its classifying space [BAut X]. *)
 
-(** [BAut X] is the type of types that are merely equivalent to [X].  We put this in a module to make it local to this file; at the end of the file we'll redefine [BAut X] for export to be the classifying space of [Aut X] (which will be judgmentally, but not syntactically, equal to this definition.) *)
-Module Import BAut.
-  Definition BAut (X : Type) := { Z : Type & merely (Z = X) }.
-End BAut.
+(** [BAut X] is the type of types that are merely equivalent to [X]. *)
```
jdchristensen, line 14:
```suggestion
(** [BAut X] is the type of types that are merely equal to [X]. *)
```

theories/Algebra/Aut.v
```diff
@@ -27,18 +28,19 @@ Proof.
...
-  pose (c := conn_map_compose (Tr (-1))
-                              (factor1 (image (Tr (-1)) (unit_name X)))
-                              (equiv_baut_image_unit X)^-1).
-  refine (conn_map_homotopic _ _ _ _ c); intros []; reflexivity.
+  exists (tr (X; tr 1)).
+  rapply Trunc_ind; intros [Z p].
+  strip_truncations.
+  apply (ap tr).
+  rapply path_sigma_hprop.
+  exact p^.
```
jdchristensen, line 39: With the new proof, one can generalize this.  Maybe this should go in Truncations/Connectivity.v, near the other results with `is0connected` in their name?  The definition of `Aut` still is able to use this instance.
```coq
Global Instance isconnected_component {X : Type} (x : X)
  : IsConnected 0 { z : X & merely (z = x) }.
Proof.
  exists (tr (x; tr 1)).
  rapply Trunc_ind; intros [z p].
  strip_truncations.
  apply (ap tr).
  rapply path_sigma_hprop.
  exact p^.
Defined.
```

theories/Algebra/Aut.v
```diff
@@ -11,10 +11,11 @@ Local Open Scope path_scope.
...
-  Definition BAut (X : Type) := { Z : Type & merely (Z = X) }.
-End BAut.
+(** [BAut X] is the type of types that are merely equivalent to [X]. *)
+Definition BAut@{u v} (X : Type@{u}) : Type@{v}
+  := sig@{v v} (fun Z => merely (paths@{v} Z X)).
+
+Global Instance ispointed_baut {X : Type} : IsPointed (BAut X) := (X; tr 1).
 
 (** Equivalently, [BAut X] is the (-1)-image of the classifying map [1 -> Type] of [X]. *)
 Definition equiv_baut_image_unit X
```
jdchristensen, line 21: I believe that this was only in this file to support the old proof that `BAut` is 0-connected.  Maybe it should move to Spaces/BAut/* now?

theories/Algebra/Aut.v
jarlg: Good catch.

theories/Truncations/Connectedness.v
```diff
@@ -144,6 +144,18 @@ Proof.
   - apply tr; assumption.
 Defined.
 
+(** The path component of an point [x : X] is connected. *)
```
Alizter, line 147:
```suggestion
(** The path component of a point [x : X] is connected. *)
```

Alizter:
OK having another look at this. Does it make sense to define `BAut` in `Aut.v` anymore? Perhaps we can move the definition to the more obvious `Spaces.BAut` and then import it in `Aut.v` to define the automorphism oogroup?

jarlg:
That works. This means we cannot define `BAut X` as `classifying_space (Aut X)`, like it was previously defined. But we now only work with `BAut X` as a pointed, connected type in `BAut.v`. Not sure if the classifying space definition is desirable? The "new" approach has the added benefit that it's easy to work with `BAut` as just a type, and so experimenting with using equvalences instead of paths is feasible and doesn't require univalence just for the definition.

theories/Spaces/BAut.v
```diff
@@ -36,6 +36,17 @@ Proof.
...
+(** [BAut X] is the (-1)-image of the classifying map [1 -> Type] of [X]. *)
+Definition equiv_baut_image_unit X
+: BAut X <~> image (Tr (-1)) (unit_name X).
+Proof.
+  unfold BAut, image; simpl.
+  apply equiv_functor_sigma_id; intros Z; simpl.
+  apply Trunc_functor_equiv; unfold hfiber.
+  refine ((equiv_contr_sigma _)^-1 oE _).
+  apply equiv_path_inverse.
+Defined.
```
jdchristensen, line 48: It turns out that this is also a more general fact, which should be near `is0connected_component` in Connectivity.v.
```coq
Definition equiv_component_image_unit {X : Type} (x : X)
: { z : X & merely (z = x) } <~> image (Tr (-1)) (unit_name x).
Proof.
  unfold image; simpl.
  apply equiv_functor_sigma_id; intros z; simpl.
  apply Trunc_functor_equiv; unfold hfiber.
  refine ((equiv_contr_sigma _)^-1 oE _).
  apply equiv_path_inverse.
Defined.
```
Sorry for not getting my ideas straight from the start!

jdchristensen (approved):

mikeshulman:
Do I understand correctly that it would still be true that `BAut X` would unify with `classifying_space (Aut X)`, but rather than the former simplying to the latter it would be the other way around?  That doesn't seem like a big deal.

Alizter:
@mikeshulman Yes.

jarlg:
Since `BAut X` is now only a type, whereas `classifying_space (Aut X)` is a pointed type, they can't strictly unify. But you're right @mikeshulman, the latter coerces to the former. (`BAut X` now only has an `IsPointed` instance.)

Actually I'm coming up against statements where I wish `BAut X` was a pointed type so this is turning out a bit annoying. I guess we could either define `BAut X` as a pointed type, or go via `classifying_space` again.

mikeshulman:
Seems reasonable to define it as a pointed type.

jdchristensen:
I see no harm in defining `BAut` to mean the pointed type.

PR #1509, opened 2021-05-10
added extra universe to definition of Subuniverse

Alizter:
Here is an experiment to address @jdchristensen's questions from #1452.

The change in universes doesn't affect too much code. Most changes where taking `Subuniverse@{i}` to `Subuniverse@{i i}`.

I guess the next step is to see if defining `BAut` as `Type_` is worth doing.

### DIFF
--- a/theories/Modalities/Accessible.v
+++ b/theories/Modalities/Accessible.v
@@ -31,7 +31,7 @@ Module Import IsLocal_Internal.
     (forall (i : lgen_indices f), ooExtendableAlong (f i) (fun _ => X)).
 End IsLocal_Internal.
 
-Class IsAccRSU@{a i} (O : Subuniverse@{i}) :=
+Class IsAccRSU@{a i} (O : Subuniverse@{i i}) :=
 {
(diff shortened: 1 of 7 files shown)


jdchristensen:
Nice!  Those changes are pretty mild, so I would say that this looks promising.

jdchristensen:
(Of course, it might happen that we find that some of those "i i" places need to be generalized if `BAut` and `Type_` are going to share some code...)

Alizter:
Yep, I was thinking that also, but I didn't want to generalise them yet since we don't know if we need it.

Alizter:
@jdchristensen Did we want to continue with this?

jdchristensen:
I don't think we should merge this now.  Because it uses `i i` in most places, it doesn't actually generalize `Subuniverse` in a way that would allow code to be shared, I think.  So I think it's better to close this for now, but to keep in mind that if someone wants to unify `Type_` and `BAut` in the future, it probably won't be too hard.

Issue #1511, opened 2021-05-10
things not exported by HoTT.v


jdchristensen:
As a test, I did:
```
make clean
make theories/HoTT.vo
make
```
to see what things in the library were not used by HoTT.vo.  It turns out that there are many things.  I know that some are intentional, but I wonder if some of these have been accidentally left out?  The list is below, sorted alphabetically.  (BTW, building HoTT.vo only takes 54% of the total build time.)

<details><summary>Files built after HoTT.vo</summary>

```
contrib/HoTTBookExercises.v
contrib/HoTTBook.v
theories/Algebra/Groups/FreeGroup/LoopSusp.v
theories/Algebra/Groups/FreeProduct.v
theories/Algebra/Groups/GroupCoeq.v
theories/Algebra/Groups/Presentation.v
theories/Algebra/Universal/Algebra.v
theories/Algebra/Universal/Congruence.v
theories/Algebra/Universal/Homomorphism.v
theories/Algebra/Universal/Operation.v
theories/Algebra/Universal/TermAlgebra.v
theories/Analysis/Locator.v
theories/Categories/Adjoint/Composition/AssociativityLaw.v
theories/Categories/Adjoint/Composition/Core.v
theories/Categories/Adjoint/Composition/IdentityLaws.v
theories/Categories/Adjoint/Composition/LawsTactic.v
theories/Categories/Adjoint/Composition.v
theories/Categories/Adjoint/Core.v
theories/Categories/Adjoint/Dual.v
theories/Categories/Adjoint/Functorial/Core.v
theories/Categories/Adjoint/Functorial/Laws.v
theories/Categories/Adjoint/Functorial/Parts.v
theories/Categories/Adjoint/Functorial.v
theories/Categories/Adjoint/HomCoercions.v
theories/Categories/Adjoint/Hom.v
theories/Categories/Adjoint/Identity.v
theories/Categories/Adjoint/Notations.v
theories/Categories/Adjoint/Paths.v
theories/Categories/Adjoint/Pointwise.v
theories/Categories/Adjoint/UnitCounitCoercions.v
theories/Categories/Adjoint/UnitCounit.v
theories/Categories/Adjoint/UniversalMorphisms/Core.v
theories/Categories/Adjoint/UniversalMorphisms.v
theories/Categories/Adjoint/Utf8.v
theories/Categories/Adjoint.v
theories/Categories/Cat/Core.v
theories/Categories/Category/Core.v
theories/Categories/Category/Dual.v
[...]

PR #1512, opened 2021-05-11
moved Axioms to new directory

Alizter:
as suggested in #1511


jdchristensen:
If we keep those files, this LGTM.  But do we really need them?

Alizter:
For the library, no. But we would like to make it easier for people to use axioms in their own developments. Making them parameterize everything by the axioms is unnecessary IMO.

Also it has the advantage of "just working" without needing to understand how the axioms are actually set up.

jdchristensen:
It's not clear to me that needing to remember that you have to type:
```coq
Require Import HoTT.Axioms.UnivalenceAxiom.
```
is any simpler than remembering that you have to type:
```coq
Axiom univalence_axiom : Univalence.
Global Existing Instance univalence_axiom.
```
which just fits in with what you would do anytime you want to assume something.  And having additional files in the tree just makes it a bit harder to find things you are looking for.  There's also the point about having multiple witnesses of Univalence, but I'm not sure why that would be a problem.  Anyways, if people think these files are handy, that's fine, but I wonder if there is a single example of anyone using them anywhere.

mikeshulman:
It would be nice to know whether anyone is actually using these files.  But it doesn't do any harm to keep them.

mikeshulman:
I suppose we could delete them and see if anyone complains...

jdchristensen:
It doesn't really matter to me.  I'm just in a "spring cleaning" mood!  :-)

SkySkimmer:
I think the idea is that if 2 files want a global axiom they should get the same one, especially if some 3rd file uses both.

Alizter:
A search on GitHub indicates that it is being used by at least a few:
https://github.com/search?l=Coq&p=2&q=UnivalenceAxiom&type=Code

mikeshulman:
Good to know, thanks.  I probably vote for keeping them, then.

jdchristensen:
I agree.

PR #1513, opened 2021-05-12
reorganized _CoqProject

Alizter:
I've written a mini style guide for _CoqProject. Alphabetized the entries and included details on how to maintian it. It should be much easier to search through now.

This relies on #1512


_CoqProject
```diff
@@ -6,400 +6,368 @@
...
-#   # Classes
-#
-
 # IMPORTANT: File paths MUST NOT start with ./, or else the makefile
 # processing will break
 
-#
+# **Notes on Style:**
+# * We organise the files here alphabetically.
+# * We break files up into groups if they are in the same folder / related.
```
jdchristensen, line 14: Since things aren't strictly ordered alphabetically, maybe clarify in this way:
```suggestion
# * We order folders alphabetically, with some exceptions mentioned below.
# * We break files up into groups if they are in the same folder / related.
# * Within each group, we order files alphabetically.
```

Alizter:
I completely missed the Analysis folder!

I've given another look through to make sure things are alphabetically ordered.

Alizter:
@mikeshulman I suspect it is possible to autogenerate the _CoqProject file completely, but probably not to a degree of readability that we have now. Perhaps this is better to do in the long term anyway.

If we switch to dune in the future, replacing coq_makefile, we won't even have to write out our files in _CoqProject since dune will be smart enough to find them. However that is still some way away, since dune doesn't quite support our use case fully.

Regarding the autogeneration of HoTT.v, it shouldn't be too difficult to write a script that handles our index files. The difficult part is that we don't index everything in a folder on purpose, and also there are files like BAut.v that share the same name as a directory. I am unsure if this is something we should spend time doing.

jdchristensen:
I'm wondering if it's better to not try to make this human readable.  We could just use the output of `find . -name '*.v' | sed 's|^\./||' | sort`, for example.  I don't see any reason a human would want to read it, and having to maintain it using a bunch of rules about the ordering and spacing is extra work.

If we do keep it human readable, I think the table of contents section should use exactly the folder names, e.g. `contrib` and `Classes`.

mikeshulman:
I like the `find` proposal.  We really do just want to compile all the `.v` files, right?

Alizter:
@jdchristensen OK I will update this PR accordingly.

Alizter:
OK, so I am willing to along with `find`. I will add some documentation on using it inside the _CoqProject file. However it is a bit annoying if you have the coq-HoTT submodule since find picks those up. We could modify the command to not look through that directory but that makes things even more esoteric. Therefore I have opened #1522

jdchristensen:
LGTM.

Alizter:
Ah I need a clean working directory before I run find. I had some untracked files which got added in too. Fixing.

Alizter:
OK, I don't understand what happened. Will check tomorrow.

mikeshulman:
Hmm, I suppose the presence of untracked files could be a problem.  Maybe the find command should only be run from the CI?

JasonGross:
Use git ls-files when the .git directory exists.  Use find only when .git does not exist (perhaps also use it only if _CoqProject also does not exist).  Here's an example:
https://github.com/coq-community/coq-performance-tests/blob/ae8385b9471409387d0f47f01e38b866ba70bda1/PerformanceExperiments/Makefile#L28-L46

Alizter:
Closing this for now, I have something better coming.

PR #1517, opened 2021-05-12
Cleanups to Factorization.v

jdchristensen:
Broken into two commits, as the second one moves code.  So it's best to view the two commits separately.

The first commit reduces the number of universes for Factorization and PathFactorization, reduces the use of Univalence, and simplifies the proofs of `cancelR_class1` and `cancelL_class2`.  It also makes the proofs of those dual to each other.

The second commit moves `contr_factor` outside of the Section, mainly so that the Univalence hypothesis comes first, but also because it doesn't benefit much from being in the Section.

### DIFF
--- a/theories/Factorization.v
+++ b/theories/Factorization.v
@@ -122,6 +122,19 @@ Record FactorizationSystem@{i j k} :=
 
 Global Existing Instances ishprop_class1 ishprop_class2.
 
+(** The type of factorizations is, as promised, contractible. *)
+Theorem contr_factor `{Univalence} (factsys : FactorizationSystem)
+        {X Y : Type} (f : X -> Y)
+  : Contr (Factorization (@class1 factsys) (@class2 factsys) f).
+Proof.
+  apply contr_inhabited_hprop.
+  - apply hprop_allpath.
+    intros fact fact'.
+    apply path_factorization; try exact _.
+    apply path_factor.
+  - apply factor.
+Defined.
+
 Section FactSys.
 
   Context (factsys : FactorizationSystem).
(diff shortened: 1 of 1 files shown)


theories/Factorization.v
```diff
@@ -11,15 +11,15 @@ Local Open Scope path_scope.
...
+  (** It's important that these are all declared with a single [Context] command, so that the marker [@{ctxi}] refers to the same universe level in all of them. *)
   Context {class1 class2 : forall (X Y : Type@{ctxi}), (X -> Y) -> Type@{ctxi}}
           `{forall (X Y : Type@{ctxi}) (g:X->Y), IsHProp (class1 _ _ g)}
           `{forall (X Y : Type@{ctxi}) (g:X->Y), IsHProp (class2 _ _ g)}
           {A B : Type@{ctxi}} {f : A -> B}.
 
   (** A factorization of [f] into a first factor lying in [class1] and a second factor lying in [class2]. *)
   Record Factorization :=
-    { intermediate : Type ;
+    { intermediate : Type@{ctxi} ;
```
jdchristensen, line 22: Without this annotation, we get a second universe variable u.  But because of the class1 and class2 conditions below, we must have u <= ctxi, so having this second variable doesn't add any generality.  I noticed this because it caused a problem in other code not in this commit, and forced me to add universe annotations.  With this change, I can avoid universe annotations.

theories/Factorization.v
```diff
@@ -113,39 +113,40 @@ Record FactorizationSystem@{i j k} :=
...
-      intros fact fact'.
-      apply path_factorization; try exact _.
-      apply path_factor.
-    - apply factor.
-  Defined.
-
-  (** It follows that the left class is right-cancellable and the right class is left-cancellable. *)
-  Definition cancelR_class1 {X Y Z} (f : X -> Y) (g : Y -> Z)
+  (** The left class is right-cancellable and the right class is left-cancellable. *)
+  Definition cancelR_class1 `{Funext} {X Y Z} (f : X -> Y) (g : Y -> Z)
```
jdchristensen, line 149: This `Funext` argument ends up being the second argument.  Should I create an additional Section just for these two results in order to get `Funext` first, or does that not matter much?

Alternatively, we could add to the definition of a FactorizationSystem that the classes respect homotopy, in which case we wouldn't need Funext here at all.  Would that be reasonable?  Or would it mean that other places that create factorization systems would need to assume Funext?

mikeshulman: Hmm, I suppose that would be analogous to how Subuniverses are assumed to respect equivalence.  I expect most concrete instances of factorization systems are naturally homotopy-invariant without funext, just like most concrete subuniverses.

mikeshulman (approved):

Issue #1516, opened 2021-05-12
Improve CI further


Alizter:
Currently with #1502 our CI builds the library on top of docker images for coq. However we are still currently building coq and the HoTT library from scratch when building documentation.

We should fully stratify our CI so that:

 1. The library is built.
 2. On top of this the documentation is built.
 3. On top of that the documentation is deployed (to a single-branch gh-pages see #1388). I guess however this only should happen on a push.
 4. We can also do some validation steps on the side (which are currently incorrect marked as doc).

Since the CI will be stratified, we can also run a few more jobs at the same time. It may be worth testing Windows and MacOS build instructions. Though this wouldn't really be testing our library, but the installation of coq.

Alizter:
I should add that I have experimented with uploading docker image artefacts but it seems to be very slow. I wonder if we can flag the success of a job before it uploads artefact so that we at least get feedback that the library compiles.

JasonGross:
Let's back up the documentation history somewhere first?  And deploy the non-single-commit version there as well?

JasonGross:
> I should add that I have experimented with uploading docker image artefacts but it seems to be very slow. I wonder if we can flag the success of a job before it uploads artefact so that we at least get feedback that the library compiles.

Maybe we only upload docker images on the master branch and not on PRs?

JasonGross:
> It may be worth testing Windows and MacOS build instructions. Though this wouldn't really be testing our library, but the installation of coq.

I almost have a script that works on all three OSes on GH actions and installs opam and Coq, if we add an opam file to the repo I think we could just add `opam install .` to test the library

Alizter:
I like the idea of using opam. I recently updated our master .opam in extra-dev which we can probably use here.

SkySkimmer:
Can we deduplicate the failing checks reported in the diff? eg
![Screenshot_20210531_165633](https://user-images.githubusercontent.com/2461932/120211448-3bf48a00-c231-11eb-9283-54a7fe87c85a.png)
should have only 1 check failure.

JasonGross:
The docker image automatically installs a matcher that we need to remove by doing something like `echo :: remove matcher:: coq matcher` or something like that (we need to look up the syntax and also the owner of the matcher installed by the docker image)

JasonGross:
[...]

PR #1522, opened 2021-05-15
removed coq-HoTT submodule

### LINKED ISSUE #1456: Remove coq-HoTT submodule
We no longer need to build our own version of coq. We should tell users to just install the appropriate version of coq via opam.

Once coq 8.14 is released, we can relax this assumption to coq-core which will only install coq without the stdlib.

We can also get rid of some of our configure and installing coq scripts. Though I am unsure how much of configure we can remove. (Ideally all of it).

Alizter:
closes #1456

I'm curious, is anybody still using the submodule?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,3 @@
-[submodule "coq-HoTT"]
-	path = coq-HoTT
-	url = https://github.com/coq/coq.git
-	branch = master
 [submodule "etc/coq-scripts"]
 	path = etc/coq-scripts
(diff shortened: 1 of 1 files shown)


Alizter:
OK our CI is one such user. I am marking the PR as draft until we update that. I suppose that will be covered by #1516

jdchristensen:
I use it, but find it awkward (e.g. with worktrees), so I'm fine with getting rid of it.

PR #1523, opened 2021-05-16
n-fold separation

mikeshulman:
For a lex modality O and a type A, the following are equivalent:
- All (n+2)-fold iterated identity types of A are O-modal.
- The O-reflector A -> OA is an n-truncated map.

### DIFF
--- a/theories/Modalities/Lex.v
+++ b/theories/Modalities/Lex.v
@@ -1,6 +1,6 @@
 (* -*- mode: coq; mode: visual-line -*- *)
 Require Import HoTT.Basics HoTT.Types.
-Require Import HFiber Extensions Pullback NullHomotopy Factorization PathAny.
+Require Import HFiber Extensions Pullback NullHomotopy Factorization PathAny Truncations.
 Require Import Modality Accessible Localization Descent Separated.
 
 Local Open Scope path_scope.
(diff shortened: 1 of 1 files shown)


mikeshulman:
I'm not sure where the definition of `nSep` should go (or whether it should be called something different).

And actually, now that I've written this, it seems that lexness may be unnecessary?  I need to go back and look at CORS again.

mikeshulman:
Oh no, I think I was getting confused about what CORS 2.26 says.  Now I'm back to thinking lexness is necessary.

Alizter (approved):
LGTM

theories/Modalities/Lex.v
```diff
@@ -91,6 +91,13 @@ Section LexModality.
     : O (x = y) <~> (to O X x = to O X y)
     := equiv_path_OO O O x y.
 
+  Definition equiv_path_O_to_O {X : Type} (x y : X)
+    : (equiv_path_O x y) o (to O (x = y)) == @ap _ _ (to O X) x y.
```
jdchristensen, line 95: Maybe `equiv_path_O_beta` would fit our naming pattern?

theories/Modalities/Lex.v
```diff
@@ -402,3 +409,31 @@ Proof.
...
+  - apply istruncmap_from_ap; intros x y.
+    pose (i := fst (IHn (x = y)) _).
+    apply istruncmap_mapinO_tr, (mapinO_homotopic _ _ (equiv_path_O_to_O O x y)).
+  - intros x y.
+    apply (snd (IHn (x = y))).
+    pose (i := istruncmap_ap n (to O A) x y).
+    apply mapinO_tr_istruncmap in i.
+    apply istruncmap_mapinO_tr, (mapinO_homotopic _ ((equiv_path_O O x y)^-1 o (@ap _ _ (to O A) x y))).
+    { intros p; apply moveR_equiv_V; symmetry; apply equiv_path_O_to_O. }
+    rapply mapinO_compose.
```
jdchristensen, line 438: If you add
```coq
    pose mapinO_isequiv. (* This speeds up the next line. *)
```
before the `rapply mapinO_compose.`, it is faster.

jdchristensen (approved):

theories/Modalities/Lex.v
mikeshulman: Hmm... it's similar, but `equiv_path_O` isn't an induction or recursion principle.

PR #1524, opened 2021-05-17
Update ObjectClassifier.v and classify maps into BAut.

jarlg:
In the 1st commit, we expand `ObjectClassifier.v` to speak about pointed maps into the universe pointed at some type, and into `Type_ O` pointed at some `O`-local type. In particular, the former corresponds to the type of fiber sequences, which we may consider to be pointed (`equiv_sigma_pfibration`) or not (`equiv_sigma_fibration_p`).

In the 2nd commit, we view `X => merely (X <~> F)` as a subuniverse to deduce facts about maps into `BAut F` from the statements about `Type_ O` in `ObjectClassifier.v`. Both `pType` and `pForall` are made cumulative in order for `equiv_pequiv_postcompose` to work at the end of `BAut.v`.

_Question_: in `Truncations/Connectedness.v` we added `conn_to_trunc_ind`, with the naming indicating that this is an induction principle for (n+1)-connected types into n-types. Any suggestions for a better name?

_Comment_: the interest of #1509 is be to make `BAut F` definitionally `Type_ O` for the subuniverse `X => merely (X = F)`, thus letting us simply specialize results in `ObjectClassifier.v` to `BAut F`. An alternative is to redefine `BAut F` to use equivalences, in which case `BAut F` already would be `Type_ O` definitionally for `subuniverse_merely_equiv`. Having these results will help in determining which approach simplifies things.

### DIFF
--- a/theories/ObjectClassifier.v
+++ b/theories/ObjectClassifier.v
@@ -7,13 +7,12 @@ Local Open Scope pointed_scope.
 
 (** We prove that type families correspond to fibrations [equiv_sigma_fibration] (Theorem 4.8.3) and the projection [pointed_type : pType -> Type] is an object classifier [ispullback_square_objectclassifier] (Theorem 4.8.4). *)
 
-(** We denote the type of all maps into a type [Y] as follows. *)
+(** We denote the type of all maps into a type [Y] as follows, and refer to them "bundles over Y". *)
 Definition Slice (Y : Type@{u}) := { X : Type@{u} & X -> Y }.
 Definition pSlice (Y : pType@{u}) := { X : pType@{u} & X ->* Y }.
 
 Definition sigma_fibration@{u v} {Y : Type@{u}} (P : Y -> Type@{u}) : Slice@{u v} Y
   := (sig@{u u} P; pr1).
-
 
 Definition sigma_fibration_inverse {Y : Type@{u}} (p : Slice Y) : Y -> Type@{u}
   := hfiber p.2.
@@ -61,11 +60,11 @@ Proof.
   - reflexivity.
 Defined.
 
-(** ** Classifying fibrations with specified fiber *)
+(** ** Classifying bundles with specified fiber *)
 
 Local Notation "( X , x )" := (Build_pType X x).
 
-(** Fibrations over [B] with fiber [F] correspond to pointed maps into the universe pointed at [F]. *)
+(** Bundles over [B] with fiber [F] correspond to pointed maps into the universe pointed at [F]. *)
 Proposition equiv_sigma_fibration_p@{u v +} `{Univalence} {Y : pType@{u}} {F : Type@{u}}
   : (Y ->* (Type@{u}, F)) <~> { p : Slice@{u v} Y & hfiber p.2 (point Y) <~> F }.
 Proof.
@@ -104,7 +103,7 @@ Definition equiv_sigma_pfibration@{u v +} `{Univalence} {Y F : pType@{u}}
 
 (** * The classifier for O-local types *)
 
-(** Families of O-local types correspond to fibrations with O-local fibers. *)
+(** Families of O-local types correspond to bundles with O-local fibers. *)
 Theorem equiv_sigma_fibration_O@{u v} `{Univalence} {O : Subuniverse} {Y : Type@{u}}
   : (Y -> Type_@{u v} O) <~> { p : { X : Type@{u} & X -> Y } & MapIn O p.2 }.
 Proof.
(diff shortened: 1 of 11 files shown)


theories/Truncations/Connectedness.v
```diff
@@ -55,6 +55,15 @@ Proof.
...
 
+(** ** Connectedness of path spaces *)
+
+Global Instance isconnected_paths `{Univalence} {n A}
+       `{IsConnected n.+1 A} (x y : A)
+: IsConnected n (x = y).
+Proof.
+  refine (contr_equiv' _ (equiv_path_Tr x y)^-1).
+Defined.
+
```
jarlg, line 66: (This was just moved up.)

theories/Truncations/Connectedness.v
```diff
@@ -76,6 +85,18 @@ Proof.
   apply O_lex_leq_Tr.
 Defined.
 
+(** To prove an [n]-truncated predicate on an (n+1)-connected, pointed type, it's enough to prove it for the basepoint. *)
+Definition conn_to_trunc_ind `{Univalence} (n : trunc_index) {A : pType@{u}} `{IsConnected n.+1 A}
+           (P : A -> Type@{u}) `{forall a, IsTrunc n (P a)} (p0 : P (point A))
+  : forall a, P a.
```
jarlg, line 91: Name suggestions?

theories/ObjectClassifier.v
```diff
@@ -1,118 +1,175 @@
...
+  : (Y ->* (Type_ O, (F; inO)))
+       <~> { p : Fib@{u v} Y & hfiber p.2 (point Y) <~> F }.
+Proof.
+  refine (_ oE equiv_sigma_fibration_Op).
+  refine (_ oE (equiv_sigma_assoc' _ (fun p _ => hfiber p.2 (point Y) <~> F))^-1%equiv).
+  srapply equiv_functor_sigma_id; intro; cbn.
+  refine (_ oE equiv_sigma_symm0 _ _).
+  apply equiv_sigma_contr; intro e.
+  rapply contr_inhabited_hprop.
+  rapply conn_to_trunc_ind@{u v u u u u u u u}.
```
Alizter, line 146: Why the universe annotations here?

Alizter:
LGTM

Perhaps we should find a home for ObjectClassifier.v other than just in `theories/`. I'm not sure where though.

theories/ObjectClassifier.v
jarlg: Otherwise this ends up having 8 universes instead of 3.

theories/Truncations/Connectedness.v
mikeshulman: Isn't this just `conn_point_incl` combined with `conn_map_elim`?

jdchristensen: Yes, it is.  I used this proof instead of that one since it is intuitive.  I think adding this result is a good idea, because I was seeing variations of this proof repeated in various places, so it captures a common idiom.

jdchristensen: (I also don't mind if the proof is replaced with the shorter one.)

mikeshulman: Yes, I certainly agree with adding the result.  I think I prefer the shorter proof, but I won't insist.  How about naming it something that suggests those two inputs, like `conn_point_elim`?

theories/ObjectClassifier.v
```diff
@@ -1,118 +1,175 @@
...
 
-Section FamPow.
-(** We consider Families and Powers over a fixed type [A] *)
-Variable A : Type.
-Definition Fam A:=sig (fun I:Type  => I->A).
-Definition p2f: (A->Type)-> Fam A:=  fun Q:(A->Type) => ( (sig Q) ; @pr1 _ _).
-Definition f2p: Fam A -> (A->Type):=
- fun F => let (I, f) := F in (fun a => (hfiber f a)).
+(** We denote fibrations over a base [Y] as follows. *)
+Definition Fib (Y : Type@{u}) := { X : Type@{u} & X -> Y }.
```
mikeshulman, line 11: I'm not sure I agree with this name.  To me a fibration is a map *definitionally* of the form `pr1` for some type family.  (Or, perhaps, definitionally isomorphic to one such.)  An arbitrary map is *equivalent* to a fibration, but I wouldn't call it one.

mikeshulman:
The diff for the changes to `ObjectClassifier` is one of those extremely unhelpful ones that intersperses removed lines with added ones that are entirely unrelated to them.  Is the old stuff only renamed, or replaced?

Alizter:
On github, the split review option in the cog menu at the top seems to be easier to read.

mikeshulman (approved):
Thanks!  I need to remember that cog.  I wish it were more visible.

theories/ObjectClassifier.v
jdchristensen: We could switch back to `Fam`, or could use something like `Slice`.

theories/Truncations/Connectedness.v
jarlg: That seems better, I'll change it to `conn_point_elim`.

theories/ObjectClassifier.v
jarlg: I find `Slice` entirely unambiguous, so I'll go with that.

Alizter: I like Slice as a name.

jarlg:
I renamed things according to the suggestions, however some of the comments `ObjectClassifier.v` still talk about "fibrations". If desired I could rewrite these to refer to "types over" or "bundles over" some type `Y`, or some better suggestion. Let me know.

Alizter:
@jarlg I think it would be good to update the documentation while we are at it, but if you don't have time just open an issue.

jarlg:
Alright, I'll do it -- was mostly wondering if the terminology was fine, which I think it is.

mikeshulman:
Looks great!

PR #1525, opened 2021-05-17
Autogenerate _CoqProject

Alizter:
We now autogenerate _CoqProject.

If a `.v` file is being tracked by git then it gets added to the _CoqProject. I've also added (commented out) script for untracked files, but I am unsure if we want this.

The generated _CoqProject has the correct arguments, and all the relevant .v files listed alphabetically. There is also a warning to deter users from editing it:

```
###############################################################################
# WARNING: This file is autogenerated by the generate_coqproject.sh script
# found in etc/. It is set to be untracked by git.
###############################################################################
-R theories HoTT
-Q contrib HoTT.Contrib
-arg -noinit
-arg -indices-matter
-arg -native-compiler -arg no


contrib/HoTTBookExercises.v
contrib/HoTTBook.v
theories/Algebra/AbGroups/AbelianGroup.v
theories/Algebra/AbGroups/Abelianization.v
theories/Algebra/AbGroups/AbPullback.v
theories/Algebra/AbGroups/AbPushout.v
theories/Algebra/AbGroups.v
```

The advantages of doing this is that we no longer have to maintain a _CoqProject, hence it has been removed.


.gitignore
```diff
@@ -104,3 +104,6 @@ contrib/*.bak.*
 
 # ignore dune _build directory
 _build/
+
+# ignore _CoqProject since we autogenerate it
+_CoqProject
```
jdchristensen, line 109: Needs newline.

jdchristensen (approved):

.gitignore
Alizter: Why do we need a newline here? There wasn't one before.

Alizter: I guess its a general linux thing to include the newline everywhere. I was only aware of doing it for `.v` files. I'll add one in.

Alizter, line 109:
```suggestion
_CoqProject

```

mikeshulman:
So this means that when someone adds a new .v file, `make` will start compiling the new file as soon as it gets added to git locally, right?  We should also update the section "Creating new files" in STYLE.md.

Alizter:
@mikeshulman Yes that is what it should do. I'll add some changes to STYLE.md like you suggest.

Alizter:
@mikeshulman I've updated STYLE.md

PR #1527, opened 2021-05-18
Added .opam file to library

Alizter:
This lets us use `opam` to build and install the library from the source directory. This will be useful for the CI.

### DIFF
--- a/hott.opam
+++ b/hott.opam
@@ -8,7 +8,7 @@ build: [
 ]
 install: [make "install"]
 depends: [
-  "coq" {>= "8.13.2"}
+  "coq" {>= "8.13~"}
 ]
 authors: ["The HoTT Library Development Team"]
 dev-repo: "git+https://github.com/HoTT/HoTT.git"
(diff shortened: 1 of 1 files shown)


mikeshulman (approved):

hott.opam
```diff
@@ -0,0 +1,26 @@
...
+maintainer: [ "Jason Gross <jgross@mit.edu>" "Ali Caglayan <alizter@gmail.com>" ]
+homepage: "http://homotopytypetheory.org/"
+bug-reports: "https://github.com/HoTT/HoTT/issues"
+license: "BSD-2-Clause"
+build: [
+  [make "-j%{jobs}%"]
+]
+install: [make "install"]
+depends: [
+  "coq" {>= "8.13.2"}
```
JasonGross, line 11: This seems suspicious.  Are you sure it's not 
```suggestion
  "coq" {>= "8.13~"}
```
?

Alizter: Yes, that seems to be more sensible. I didn't know about the tilde syntax, but checking the opam documentation it makes more sense to use.

PR #1528, opened 2021-05-19
Silence warnings

### LINKED ISSUE #1438: Quieting the build
The build is quite noisy, filled with messages like
```
File "./theories/Categories/Functor/Composition/Laws.v", line 43, characters 0-55:
Warning: The default value for hint locality is currently "local" in a
section and "global" otherwise, but is scheduled to change in a future
release. For the time being, adding hints outside of sections without
specifying an explicit locality is therefore deprecated. It is recommended to
use "export" whenever possible. [deprecated-hint-without-locality,deprecated]
```
We should either fix the hints or suppress the warning.

Alizter:
closes #1438

We should create some issues so that we eventually do carry out the changes the warnings are asking about. But first, we need to update to coq 8.14 which has yet to be released.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -13,6 +13,9 @@ Create HintDb rewrite discriminated.
 #[export] Hint Variables Opaque : rewrite.
 Create HintDb typeclass_instances discriminated.
 
+(** We disable the [deprecated-hint-rewrite-without-locality] warning until we update to coq V8.14.0 *)
+Global Set Warnings Append "-deprecated-hint-rewrite-without-locality".
+
 (** ** Type classes *)
 
 (** This command prevents Coq from trying to guess the values of existential variables while doing typeclass resolution.  If you don't know what that means, ignore it. *)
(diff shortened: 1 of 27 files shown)


mikeshulman:
The warnings are still present on this PR?

theories/Basics/Overture.v
```diff
@@ -13,6 +13,12 @@ Create HintDb rewrite discriminated.
 #[export] Hint Variables Opaque : rewrite.
 Create HintDb typeclass_instances discriminated.
 
+(** We disable the [deprecated-instance-without-locality] warning until we update to coq V8.14.0 *)
+Global Set Warnings Append "-deprecated-instance-without-locality".
+
+(** We disable the [deprecated-hint-without-locality] warning until we update to coq V8.14.0 *)
+Global Set Warnings Append "-deprecated-hint-without-locality".
```
JasonGross, line 20:
```suggestion
(** We disable the [deprecated-hint-rewrite-without-locality] warning until we update to coq V8.14.0 *)
Global Set Warnings Append "-deprecated-hint-rewrite-without-locality".
```

theories/Basics/Overture.v
```diff
@@ -13,6 +13,12 @@ Create HintDb rewrite discriminated.
 #[export] Hint Variables Opaque : rewrite.
 Create HintDb typeclass_instances discriminated.
 
+(** We disable the [deprecated-instance-without-locality] warning until we update to coq V8.14.0 *)
+Global Set Warnings Append "-deprecated-instance-without-locality".
```
JasonGross, line 17: I don't think we want this one, as we can already do `Global Instance`

Alizter: I misunderstood the instance at first. I thought it was complaining about not passing an `#[export]` flag. Now I have fixed all the offenders.

Alizter:
@JasonGross @mikeshulman Is this good to merge? It will be useful for future PRs.

mikeshulman (approved):

JasonGross (approved):

mikeshulman:
Looks like it needs rebasing, though.

PR #1529, opened 2021-05-20
Equations and more on isomorphism

andreaslyn:
I have defined equations, moved `IsIsomorphism` to Isomorphism.v and added a some more isomorphism stuff.

I changed the way import of algebra notations work, to make it more consistent with the rest of the library. Now algebra notations are introduced with a `Local Open Scope Algebra_scope`.


theories/Algebra/Universal/Isomorphism.v
```diff
@@ -0,0 +1,278 @@
...
+
+Lemma issig_isomorphism {σ : Signature} (A B : Algebra σ)
+  : SigIsomorphism A B <~> (A ≅ B).
+Proof.
+  issig.
+Defined.
+
+(** Isomorphic algebras are equal [A ≅ B -> A = B]. *)
+
+Corollary isomorphism_to_path `{Univalence}
```
andreaslyn, line 194: I realize that this is not following the naming convention, but I have another definition `path_isomorphism` below. I am open to other ideas regarding the naming of these.

theories/Algebra/Universal/Isomorphism.v
```diff
@@ -0,0 +1,278 @@
...
+    : transport (fun C => forall u, Operation C (σ u))
+        (path_equiv_family (equiv_isomorphism f)) (operations A)
+      = operations B.
+  Proof.
+    funext u.
+    refine (transport_forall_constant _ _ u @ _).
+    now apply path_operations_equiv.
+  Qed.
+
+  Theorem is_isomorphism_to_path
```
andreaslyn, line 77: Here I am using the same naming style as with `isomorphism_to_path` below.

theories/Algebra/Universal/Isomorphism.v
```diff
@@ -0,0 +1,278 @@
...
+(** Equal algebras are isomorphic [A = B -> A ≅ B] *)
+
+Lemma path_to_isomorphism {σ} {A B : Algebra σ} (p : A = B) : A ≅ B.
+Proof.
+  destruct p. exact (Build_Isomorphism (hom_id A)).
+Defined.
+
+(** To find a path between two isomorphisms [F G : A ≅ B], it suffices to find a path between the underlying families of functions. *)
+
+Lemma path_isomorphism `{Funext} {σ : Signature} {A B : Algebra σ}
```
andreaslyn, line 210: I considered naming this `path_isomorphism_path_def` instead, then use `path_isomorphism` for `isomorphism_to_path`.

Alizter:
Maybe `ua_algebra_scope` is a better name?

andreaslyn:
> Maybe `ua_scope` is a better name?

Because using `Algebra_scope` for isomorphism is misleading?

I think that `ua_scope` is also fine. It could be confused with univalence axiom scope, but I do not think we have such a scope.

Alizter:
Another alternative is to do what I did in #1496 and use `wc_iso_scope`. This way, you will need to replace all uses of isomorphism with the wildcat equivalence `$<~>` but the scope will give you the isomorphism notation. Of course, you will have to write down the wildcat of algebras, but I don't think that is too difficult.

andreaslyn:
I am closing this in favor of another PR, where I connect universal algebra to wildcat.

PR #1530, opened 2021-05-21
Change type of Pi to AbGroup for n.+2

### LINKED ISSUE #1160: Change HomotopyGroup_type to be AbGroup when above n=1.
We should change the HomotopyGroup_type so that when n is larger than 1 we have the type of abelian groups. This will allow us to do abelian group things to the higher homotopy groups.

Alizter:
I've changed the definition of HomotopyGroup so that it is a pType at 0, Group at 1 and AbGroup at n.+2. I've adapted the rest of the library accordingly.

In order to satisfy the uniform inheritance condition for coercions, I had to split the definition of HomotopyGroup_type into two parts.

closes #1160

### DIFF
--- a/theories/Homotopy/HomotopyGroup.v
+++ b/theories/Homotopy/HomotopyGroup.v
@@ -1,76 +1,78 @@
-Require Import Basics.
-Require Import Types.
-Require Import Pointed.
-Require Import Algebra.AbGroups.
-Require Import Truncations.
+Require Import Basics Types Pointed WildCat.
+Require Import Algebra.Groups Algebra.AbGroups.
+Require Import Truncations Modalities.ReflectiveSubuniverse.
 Require Import Spaces.Nat.
-Require Import Modalities.ReflectiveSubuniverse.
-Require Import WildCat.
 
 Local Open Scope nat_scope.
-Local Open Scope pointed_scope.
 Local Open Scope path_scope.
+Local Open Scope pointed_scope.
+Local Open Scope wc_iso_scope.
+
+(** We define the type of homotopy groups of positive index seperately so we can get some coercions to behave well. *)
+Definition HomotopyGroup_type_pos (n : nat) : Type :=
+  match n with
+  | 0 => Group
+  | n.+1 => AbGroup
+  end.
 
 (** The type that the nth homotopy group will have. *)
-Definition HomotopyGroup_type (n : nat) : Type
-  := match n with
-      | 0 => pType
-      | n.+1 => Group
-     end.
+Definition HomotopyGroup_type (n : nat) : Type :=
+  match n with
+  | 0 => pType
+  | n.+1 => HomotopyGroup_type_pos n
+  end.
+
+(** Almost homotopy group is a group *)
+Definition HomotopyGroup_type_pos_group (n : nat) : HomotopyGroup_type_pos n -> Group :=
+  match n with
+  | 0 => fun G => G
+  (** This works since [AbGroup]s coerce to [Group]s *)
+  | n.+1 => fun G => G
+  end.
+
+(** If we had defined HomotopyGroup_type in one go, we would have violated the uniform inheritence checker that coq's coercion system uses. *)
+Coercion HomotopyGroup_type_pos_group : HomotopyGroup_type_pos >-> Group.
 
 (* Every homotopy group is, in particular, a pointed type. *)
-Definition HomotopyGroup_type_ptype (n : nat) : HomotopyGroup_type n -> pType
-  := match n return HomotopyGroup_type n -> pType with
-     | 0 => fun X => X
-     (* This works because [ptype_group] is already a coercion. *)
-     | n.+1 => fun G => G
-     end.
+Definition HomotopyGroup_type_ptype (n : nat) : HomotopyGroup_type n -> pType.
+Proof.
+  destruct n.
+  1: exact idmap.
+  (** We need to simplify the type here for the coercion to be inserted. *)
+  cbn; exact idmap.
+Defined.
 
 Coercion HomotopyGroup_type_ptype : HomotopyGroup_type >-> pType.
 
-(** We first define [Pi 1 X], and use this to define [Pi n X].
-  The reason is to make it easier for Coq to see that [Pi (n.+1) X] is
-  definitionally equal to [Pi 1 (iterated_loops n X)] *)
+Create HintDb path_hints.
+Ltac path_auto := eauto with path_hints.
+
+#[export] Hint Immediate idpath : path_hints.
+#[export] Hint Immediate concat_1p : path_hints.
+#[export] Hint Immediate concat_p1 : path_hints.
+#[export] Hint Immediate concat_Vp : path_hints.
+#[export] Hint Immediate concat_pV : path_hints.
+#[export] Hint Immediate concat_p_pp : path_hints.
+#[export] Hint Immediate concat_pp_p : path_hints.
+
+(** We first define [Pi 1 X], and use this to define [Pi n X]. The reason is to make it easier for Coq to see that [Pi (n.+1) X] is definitionally equal to [Pi 1 (iterated_loops n X)] *)
 Definition Pi1 (X : pType) : Group.
 Proof.
-  srapply (Build_Group (Tr 0 (loops X)));
-  repeat split.
+  snrapply (Build_Group (Tr 0 (loops X))); repeat split.
   (** Operation *)
-  + intros x y.
+  { intros x y.
     strip_truncations.
-    exact (tr (x @ y)).
+    exact (tr (x @ y)). }
   (** Unit *)
-  + exact (tr 1).
+  1: exact (tr 1).
   (** Inverse *)
-  + srapply Trunc_rec; intro x.
-    exact (tr x^).
+  { srapply Trunc_rec; intro x.
+    exact (tr x^). }
   (** IsHSet *)
-  + exact _.
-  (** Associativity *)
-  + intros x y z.
-    strip_truncations.
-    cbn; apply ap.
-    apply concat_p_pp.
-  (** Left identity *)
-  + intro x.
-    strip_truncations.
-    cbn; apply ap.
-    apply concat_1p.
-  (** Right identity *)
-  + intro x.
-    strip_truncations.
-    cbn; apply ap.
-    apply concat_p1.
-  (** Left inverse *)
-  + intro x.
-    strip_truncations.
-    apply (ap tr).
-    apply concat_Vp.
-  (** Right inverse *)
-  + intro x.
-    strip_truncations.
-    apply (ap tr).
-    apply concat_pV.
+  1: exact _.
+  (** Axioms *)
+  all: hnf; intros; strip_truncations;
+    nrapply (ap tr); path_auto.
 Defined.
 
 (** Definition of the nth homotopy group *)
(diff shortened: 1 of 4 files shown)


theories/Homotopy/HomotopyGroup.v
```diff
@@ -1,112 +1,109 @@
...
+Create HintDb path_hints.
+Ltac path_auto := eauto with path_hints.
+
+#[export] Hint Immediate idpath : path_hints.
+#[export] Hint Immediate concat_1p : path_hints.
+#[export] Hint Immediate concat_p1 : path_hints.
+#[export] Hint Immediate concat_Vp : path_hints.
+#[export] Hint Immediate concat_pV : path_hints.
+#[export] Hint Immediate concat_p_pp : path_hints.
+#[export] Hint Immediate concat_pp_p : path_hints.
```
Alizter, line 56: TODO: I should move these to PathGroupoids.v

Alizter:
@jdchristensen This should simplify some things in the Hurewicz branch.

jdchristensen:
It's unfortunate that so many proofs need a separate case now.  I'd be interested to see whether this simplifies things in the Hurewicz branch, or means that additional cases need to be handled.  I wonder if there's a clever trick that can somehow make things a bit smoother?  If nobody has any suggestions over the next day or two, then I vote to merge this, but I think it's worth waiting to see if something else is suggested.  E.g., since being abelian is a property, and doesn't affect the group homomorphisms, it could be a typeclass instance applied to some objects of type Group.  That's just a random idea and probably has lots of issues.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -146,123 +143,161 @@ Definition pi_functor (n : nat) {X Y : pType}
...
+  1: exact (ptr_functor 0).
+  destruct n.
+  1: exact pi1_functor.
+  intros f.
+  exact (pi1_functor (iterated_loops_functor n.+1 f)).
 Defined.
 
 Definition pi_functor_succ {X Y : pType} (f : X $-> Y) (n : nat)
-  : pi_functor n.+1 f $== pi_functor 1 (iterated_loops_functor n f).
+  : pi_functor n.+2 f $== pi1_functor (iterated_loops_functor n.+1 f).
```
jdchristensen, line 154: It's unfortunate that this doesn't typecheck anymore with `n.+1` and `n` instead of `n.+2` and `n.+1`, even though it does typecheck if you hardcode `1` and `0`.  It's in fact surprising to me that it *does* typecheck for larger `n`, since the LHS is a group homomorphism between abelian groups, while the RHS is a group homomorphism between groups, but I guess the coercions kick in to make it work.  In the case where you have `n.+1` on the left, I think the problem is that the reduction of the type blocks.

I'm concerned that this kind of thing is going to cause more case separation in the future.  Here, we can't even state what we expect to be true.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -146,123 +143,161 @@ Definition pi_functor (n : nat) {X Y : pType}
...
 Proof.
-  reflexivity.
+  hnf; reflexivity.
```
jdchristensen, line 156:
```suggestion
  (** Using [hnf] here makes this faster. *)
  hnf; reflexivity.
```

Alizter: I'm not sure what the blowup here is. There are definitely too many terms being unfolded by both `cbn` and `simpl`.

theories/Homotopy/HomotopyGroup.v
mikeshulman: More specifically, the LHS has type `pi_functor_type n.+1 X Y` which reduces to `GroupHomomorphism (Pi n.+1 X) (Pi n.+1 Y)`, whereas the RHS has type `GroupHomomorphism (Pi1 (iterated_loops n X)) (Pi1 (iterated_loops n Y))`.  The problem is that the type `Pi n.+1 X` doesn't reduce to `Pi1 (iterated_loops n X)` for general `n`, and it's not clear how to make it do so: when `n=0` it does, whereas when `n>0` it reduces to `Build_AbGroup' (Pi1 (iterated_loops n X)) _`, which *coerces* to `Pi1 (iterated_loops n X)` but isn't syntactically equal to it.

Mathematically, I feel like the "right" solution is to define `Pi1` so that it outputs a `HomotopyGroup_type_pos n`.  Then we could define `Pi` with only a single induction so that `Pi n.+1 X` would reduce to `Pi1 (iterated_loops n X)` for all `n`.  But of course, this would require the input of `Pi1`, and the output of `iterated_loops`, being an "n-fold loopspace" rather than just a pointed type.  So we would have to define the type of n-fold loopspaces, which presumably coerces to `pType`.  Is that a road we want to go down?

jdchristensen: Regarding your first paragraph, while `Build_AbGroup' (Pi1 (iterated_loops n X)) _` is not definitionally equal to `Pi1 (iterated_loops n X)` when n > 0, when you wrap two such things in `GroupHomomorphism`, they do become definitionally equal, because the coercion gets applied and throws away the proof of abelianess.  This is why the new version of `pi_functor_succ` type checks, which surprised me at first.  So, just to emphasize, the two types are definitionally equal when n = 0 and also when n > 0 (i.e. when you substitute n.+1 for n), but not for general n.

I'm not sure about the idea of making `Pi1` accept an n-fold loop space as input, as `Pi1` is a pretty basic thing that should be very easy to understand.

BTW, it's not really an issue that the new version of the above result misses the base case, since the base case is also true by reflexivity.  Come to think of it, it's not even clear why we need this result at all, since it is just reflexivity.  (I guess it could be useful to have because of the speed issue with naively using reflexivity.)  I'm bringing this up more as a sign that there could be further such issues ahead.

As Ali mentioned, I already did something similar to this PR in my WIP on formalizing the Hurewicz theorem.  In [HomotopyGroup2.v](https://github.com/jdchristensen/HoTT/blob/Hurewicz/theories/Hurewicz/HomotopyGroup2.v),
I defined `Pi' n X` which lands in abelian groups when `n >= 2`, and coerces to the type of `Pi n X`.  Then in [Smashing.v](https://github.com/jdchristensen/HoTT/blob/Hurewicz/theories/Hurewicz/Smashing.v) (which has lots of complicated stuff showing that the different operations on iterated loop spaces agree), this new `Pi'` is used because I need to form the abelian group of group homomorphisms into `Pi' n.+2 Z`.  I wonder if having two `Pi`s, with only one of them landing in abelian groups when needed, would make things smoother?  In my approach, I didn't have to change anything in the existing library.  I just added `Pi'`, which was pretty trivial.  (The only cleverness was in choosing the right proof that the group is abelian, when it is.  The more obvious choice made it harder for Coq to see that certain things are definitionally equal.  The HomotopyGroup2.v file has a comment about this.)

jdchristensen: Here's another off-the-cuff idea.  Define a "possibly abelian group" to be a term of `{ G : Group & Unit + IsAbGroup G }`.  Then `Pi n.+1 X` can give the element of `Unit` when `n = 0` and the proof of abelianess when `n > 0`.  But it can make that case distinction *after* supplying the first component of the pair.  And so the coercion to `Group` will compute for all `n`.  I did a rough test and this works.  Not sure how it would fit into the big picture, or whether it's possible to coerce this to `AbGroup` when `n` is large enough.

jdchristensen: We can also pin down the type more precisely, e.g.
```coq
Definition condition (n : nat) (G : Group)
  := match n with
    | 0 => Unit
    | _ => IsAbGroup G
    end.

(** The type that the nth homotopy group will have. *)
Definition HomotopyGroup_type'' (n : nat) : Type
  := match n with
      | 0 => pType
      | S m => { G : Group & condition m G }
     end.
```
Again, `pr1` is a computable coercion to `Group`.  But we could also have a coercion from `{ G : Group & IsAbGroup G }` to `AbGroup`, which would kick in automatically for `Pi n.+2 X` in a context where an abelian group was expected.  (I think we'd have to make `{ G : Group & IsAbGroup G }` into a class for this to work?  Haven't tested this part yet.  Maybe there's even a way to redefine the type `AbGroup` to make this work out definitionally??)

jdchristensen: I think it's possible to get this to all work out smoothly, if we're willing to change our definitions of `Group` and `AbGroup`.  In the code below, think of the existing `Group` type as an `InternalGroup`, and think of `NewGroup` and `NewAbGroup` as what would be the exposed definitions.
```coq
Record GroupWithStructure (S : Group -> Type) := {
  group_object : Group;
  group_structure : S group_object;
}.

Definition NewGroup := GroupWithStructure (fun _ => Unit).

Definition forget {S : Group -> Type} (Gs : GroupWithStructure S) : NewGroup
  := Build_GroupWithStructure _ (group_object _ Gs) tt.

Coercion forget : GroupWithStructure >-> NewGroup.

Definition NewAbGroup := GroupWithStructure (fun G : Group => IsAbGroup G).

Definition condition (n : nat) (G : Group)
  := match n with
    | 0 => Unit
    | _ => IsAbGroup G
    end.

(** The type that the nth homotopy group will have. *)
Definition HomotopyGroup_type'' (n : nat) : Type
  := match n with
      | 0 => pType
      | S m => GroupWithStructure (condition m)
     end.
```

With this, it's easy to define `Pi n x` to have the type shown above.  The type of `Pi 1 X` is then definitionally equal to `NewGroup` while the type of `Pi n.+2 X` is definitionally equal to `NewAbGroup`.  Moreover, the coercion `forget` that treats `Pi n.+1 X` as a `NewGroup` computes for every `n`.  In addition, for a `NewGroup` defined with `tt` explicitly in the last spot (which `Pi` does), `forget` is definitionally the identity map, so it doesn't get in the way.

I haven't checked what happens with homomorphisms and `Pi_functor`, but I don't see an issue.

With appropriate definitions, none of the rest of the library would need to change, as we would still have `Build_Group'` and accessor functions with the same names.

It seems weird to store `tt` along with every group, but by having this unused "slot", we can make the coercion uniform, so it computes.  What do people think?

Alizter:
I always felt like I was missing something when designing the algebra library. I feel as though there are ways to get coq to accept various casts between algebraic structures without causing too much headache. For a while, I kept reading that canonical structures was supposed to help with this, but I never managed to git it to work.

Here is a small example that demonstrates what canonical structures actually does:
```coq
Require Import Basics Types Algebra.Groups Algebra.AbGroups.

Local Open Scope mc_add_scope.

(** G is a group that happens to be commutative. *)
Axiom G : Group.
Axiom commutative_G : Commutative ((.*.) : G -> G -> G).
Global Existing Instance commutative_G.

(** Obviously, coq has no way of knowing that this is an abgroup *)
Fail Check (G : AbGroup).

(** Say we have a lemma about abelian groups *)
Definition ab_is_great (A : AbGroup) (x y : A) : Type
  := x + y = y + x.

(** Coq can't apply this lemma to G because it's not an abelian group. *)
Fail Lemma G_is_great (x y : G) : ab_is_great _ x y.

(** But if we mark the constructor Builld_AbGroup' as a "Canonical Structure" then we can give hints as on how to use G like an abelian group to coq. The commutativity part will be found by typeclasses. *)
Canonical Structure Build_AbGroup'.

(** Now the lemma works! *)
Lemma G_is_great (x y : G) : ab_is_great _ x y.

(** But coq still cannot cast G to an abelian group. *)
Fail Check (G : AbGroup).
```

jdchristensen:
@Alizter If something like that works, it would be very nice.  But the one thing I need doesn't seem to work.  I have
```coq
Definition AbHom' (A : Group) (B : AbGroup) `{Funext} : AbGroup.
```
and even with the canonical structures and definitions you made, `AbHom' H G` fails (where of course `H` is assumed to be a group).  Maybe someone knows how to make groups cast to abelian groups when we have an instance showing that they are abelian?

theories/Homotopy/HomotopyGroup.v
mikeshulman: I actually kind of like that.  We could even take it further by defining a `GroupWithStructure` to be a particular kind of `pTypeWithStructure`, so that `Pi n X` would coerce to a `pType` for every `n`.  Of course that would require changes to the `Pointed` library too.

mikeshulman: But if a `pTypeWithStructure` also knows what its homomorphisms are, it might make the statement of the long exact sequence look more uniform.

jdchristensen: I'm also growing to like my `GroupWithStructure` suggestion, especially now that `AbGroup` has been redefined to be essentially what I proposed in my previous comment.  Also, it occurred to me that it could be useful for other subcategories of groups, e.g. finite groups, finitely generated groups, etc.

@mikeshulman Can you elaborate on "knows what its homomorphisms are"?  How might this be organized?

jdchristensen: It's not clear to me how the `pTypeWithStructure` idea might simplify things, since the underlying `pType` isn't uniform in `n`, so the coercion to `pType` wouldn't compute.  For `n > 0`, the underlying `Group` is uniform in `n`, so the trick above achieve getting it to compute.

Another remark:  a less radical idea is to use `GroupWithStructure` to define the output type of `Pi`, and as the overall library definition of `AbGroup`, but not as the library definition of `Group`.  Then the type of `Pi 1 X` would not literally be `Group`, but would coerce to it.  This would be less invasive, but add in a coercion in that case.  Maybe it should be the way to try things for a first pass.  I'm too busy this week to try it, though.

mikeshulman: Perhaps using displayed categories?  If the predicate `S` is a displayed category over `pType`, then it has a natural notion of homomorphism.

mikeshulman: Uniformity is a good point.  The idea with the coercion sounds worth trying.

Alizter:
@jdchristensen I have spoken with the mathcomp devs also, and it has been on peoples wishlist for coq for a while now. Cyril Cohen has dubbed these "reverse coercions" and there are some discussions going on: https://coq.discourse.group/t/generalizing-coercions-to-infer-typeclass-arguments/1241/4

Alizter:
This will need a rebase. But I also wonder if using fmap will change anything with respect to the separate cases that were forced.

jdchristensen:
I'd still like to try my `GroupWithStructure` idea (or maybe it should be called `GroupWithProperty`), but I'm behind on too many things right now to give it a shot.  If someone else wants to try, that would be great.  Or we could put this on the backburner for a while, if it isn't blocking anything else.

Alizter:
Don't have time to work on this any longer.

PR #1532, opened 2021-05-21
Connect universal algebra to wildcat

andreaslyn:
I have removed the isomorphism results from Homomorphism.v and connected with wildcat. I think the formalisation is nicer now that I am using the wildcat notations.

In the next PR I will reintroduce the isomorphism results in Isomorphism.v.


theories/WildCat/Core.v
```diff
@@ -147,7 +147,7 @@ Definition cat_assoc_opp_strong {A : Type} `{Is1Cat_Strong A}
   := (cat_assoc_strong f g h)^.
 
 Global Instance is1cat_is1cat_strong (A : Type) `{Is1Cat_Strong A}
-  : Is1Cat A.
+  : Is1Cat A | 1000.
```
andreaslyn, line 150: I think it is better with low the priority of this instance, to avoid overriding better alternatives. For example, algebras form a 1cat without assuming funext, but requires funext for strong 1cat, so we want coq to pick the instance of 1cat which does not need funext.

theories/Algebra/Universal/TermAlgebra.v
```diff
@@ -1,20 +1,14 @@
-(** This file defines the term algebra [TermAlgebra], also
-    referred to as the absolutely free algebra.
```
andreaslyn, line 2: Sorry, it was overkill to make the comments in this file into single line.

mikeshulman (approved):
That's great!

spitters:
Looks great. Do you want to fix the warnings? I believe it is recommend to give more explicit (global/local) information, but I haven't at it carefully.

Alizter:
Those warning should go away once #1528 is fixed. I don't think it has anything to do with this PR.

Issue #1520, opened 2021-05-15
Mechanisation of GCH -> AC


dominik-kirst:
Mostly during the HoTT Spring School, Felix Rech and I have ported our Coq mechanisation (cf. [CPP'21 paper](https://popl21.sigplan.org/details/CPP-2021/23/The-Generalised-Continuum-Hypothesis-Implies-the-Axiom-of-Choice-in-Coq)) of the result that the generalised continuum hypothesis (GCH) implies the axiom of choice (AC) to the HoTT library. The code is available [here](https://github.com/dominik-kirst/sierpinski-hott) and @spitters suggested to open an issue to discuss whether this development is of interest for the HoTT library. 

To provide some background, the result was first proven by Sierpinski for classical set theory, which we compared to the natural translation to Coq's type theory in our paper. This translation of set theory into type theory requires extensionality and unique choice, so it appeared worthwhile to compare the existing mechanisation to a version in HoTT.

We assume GCH as `forall X Y : hSet, infinite X -> hinject X Y -> hinject Y (X -> hProp) -> hinject Y X + hinject (X -> hProp) Y`, where `infinite X` denotes the existence of an injective function `nat -> X` and `hinject X Y` denotes the mere existence of an injective function `X -> Y`, and show the formulation of AC as `forall (X Y : hSet) (R : X -> Y -> hProp), (forall x, hexists (R x)) -> hexists (fun f => forall x, R x (f x))`. Already a weaker formulation of GCH (not yielding a case distinction) implies LEM and therefore also propositional resizing, so we freely assume both where necessary.

The development consists of three files:
- `Ordinals.v` contains a basic theory of (constructive) ordinals as developed in the HoTT book. It closes with a construction of the Hartogs number for a given hSet, which is an ordinal of not smaller but still controlled cardinality.
- `Sierpinski.v` contains a proof that any hSet embeds into its Hartogs number using GCH.
- `Results.v` contains the main result GCH -> AC, as well as the implications WO -> AC, GCH -> LEM, and LEM -> propositional resizing.

I'm looking forward to your thoughts and would also be curious to discuss a few more questions I gathered during this project :)

mikeshulman:
I would love to see this incorporated into the library, not just for the result but for the general theory of ordinals that might be useful for other work.  And I'd be interested to hear and discuss your questions too.

dominik-kirst:
> I would love to see this incorporated into the library, not just for the result but for the general theory of ordinals that might be useful for other work.

[...]

PR #1533, opened 2021-05-22
Mechanisation of GCH -> AC

dominik-kirst:
This is the PR announced in #1520 of a mechanisation that the generalised continuum hypothesis implies the axiom of choice (c.f. a related [abstract](https://github.com/dominik-kirst/sierpinski-hott/blob/master/tex/abstract.pdf)). I've slightly restructured the file organisation and tried to conform to the style guidelines as much as possible, apologies in advance for anything that I've missed.

Since I didn't find a folder especially dedicated to set theory, I've created one. It contains the following files:
- `Ordinals.v` consists of basic ordinal theory following the HoTT book. This file is probably most relevant for reuse, so I guess it could be mentioned in `HoTT.v` and `HoTTBook.v`.
- `Hartogs.v` contains a construction of the Hartogs numbers, i.e. given a set yields an ordinal of strictly greater cardinality (in the presence of AC).
- `GCH.v` contains a definition of the generalised continuum hypothesis and a proof that already CH implies LEM.
- `Sierpinski.v` contains a proof of Sierpinski's result that GCH implies AC, using the Hartogs number as an intermediate step.
If there's a better place for these files I'm happy to move them.

I have two questions that should probably be settled before merging:
- The main theorem still requires an instance of `ExcludedMiddle` although the premise GCH implies LEM. Is there a way the formal assumption can be eliminated while still reusing the type class for intermediate steps?
- The lemma `hset_equiv_bij` in `Hartogs.v` states that an equivalence of sets can be obtained from an injective an merely surjective function. Should this be moved to a more general file as I was surprised not to find it in the core library yet?

I'm looking forward to your feedback :)


theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+    srapply power_morph. 3: apply X. intros P.
+    srapply power_morph. 3: apply P.
+    srapply power_inj.
+  Defined.
+
+  Lemma injective_uni_fix :
+    IsInjective uni_fix.
+  Proof.
+    intros X Y. unfold uni_fix. intros H % injective_power_morph; trivial.
+    intros P Q. intros H % injective_power_morph; trivial.
```
Alizter, line 265: This seems to be problematic as H is being used again.

Alizter:
There seems to be some variable clash in Hartogs number. The older version of coq 8.13.2 ignored it, but the newer version isn't allowing it.

dominik-kirst:
> There seems to be some variable clash in Hartogs number. The older version of coq 8.13.2 ignored it, but the newer version isn't allowing it.

Thanks for pointing this out, Ali. So the CI also runs against Coq's master branch?

Alizter:
@dominik-kirst Yes, we have a version of coq that we support, currently v8.13.2 and we support the master version of coq in order to keep feedback with the developers.

If you want to check if your development runs with master, you can create an opam switch that has the master verison of coq installed. But it isn't necessary as the CI will usually tell you if you need to change something.

dominik-kirst:
> If you want to check if your development runs with master, you can create an opam switch that has the master verison of coq installed. But it isn't necessary as the CI will usually tell you if you need to change something.

Okay I'll do that, perhaps the variable clash in `Hartogs.v` is not the only incompatibility.

Alizter:
I will try to submit some more review later this week. But currently whilst skimming I have a few things I would like to pursue, which seem largely orthogonal.
1. Do some general cleanup:
  * Making the proofs a bit more readable (adding more comments), making tactics more concise
  * Annotating definitions with their types
  * Moving things around (there are some things defined in Ordrinals.v for example, which are probably of more general interest).
2. Try to remove prop resizing as much as possible. There are many reasons we might want to do this, one of them being that in the library in general, we avoid use of prop resizing. The uses are restricted to the folder PropResizing.
3. Try to use some more constructive notions, for example for oridinals. Here is some reading if you are interested: https://arxiv.org/abs/2104.02549

I think our main focus should be 1, but I have to say you have done a very good job with this PR, but I would like to help you clean it up a bit more. I currently don't have the time to do this, but I might be more available later this week. For now, I wouldn't worry too much about doing anything, unless anybody has some more specific things they want changed. I'm also interested how you found the STYLE.md guide, whether it was easy to read/use etc. I want to update it a bit and make it simpler to follow.

Issues 2 and 3 are more of a future problem, and may require further investigating so I would not take them too seriously for now. We can definitely have a discussion about what can be done.

Sorry for being so vague with the comment, but I will try to get back to you soon. Good job so far! :-)

mikeshulman:
I don't think there's any need to restrict *uses* of PropResizing to a particular folder.  We could even discuss whether that folder should exist; the statement of the axiom could go in some other file and the two other files would go naturally in Metatheory.

mikeshulman:
Specifically, implementing axioms as typeclasses allows us to track their uses independently of folder organization or namespacing.

mikeshulman:
Using fewer hypotheses and being more constructive are always good goals, but not worth blocking otherwise good work for.

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+
+
+(** * Cardinals *)
+
+Definition Cardinal := Trunc 0 HSet.
+Definition card A `{IsHSet A} : Cardinal
+  := tr (Build_HSet A).
+
+
+Instance le_cardinal `{Univalence} : Le Cardinal
```
mikeshulman, line 17: We generally qualify Instances explicitly with `Global` (the normal case) or `Local` (if desired), so the reader doesn't have to remember or guess which is the default in the current situation.

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
+From HoTT Require Import Basics TruncType ExcludedMiddle abstract_algebra.
+From HoTT Require Import PropResizing.PropResizing.
+
+From HoTT.Sets Require Import Ordinals.
+
+(** This file contains a construction of the Hartogs number. *)
+
+
+
+(** * Cardinals *)
```
mikeshulman, line 10: Maybe this section would be better placed in a file `theories/Sets/Cardinals.v`?

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+      + intros [B small_B] C C_B; cbn in *. apply tr.
+        unshelve eexists (C; _); cbn; auto.
+        revert small_B. srapply Trunc_rec. intros [f isinjective_f]. apply tr.
+        destruct C_B as [b ->].
+        exists (fun '(x; x_b) => f x); cbn.
+        intros [x x_b] [y y_b] fx_fy. apply path_sigma_hprop; cbn.
+        apply (isinjective_f x y). exact fx_fy.
+  Defined.
+
+  Definition power_type (A : Type) : Type
```
mikeshulman, line 100: I think a basic definition like this should be in a more central file rather than buried in a file called Hartogs.

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+    := A -> HProp.
+
+  Notation 𝒫 := power_type.
```
mikeshulman, line 103: I'm not sure (Ali?) but I think we are still trying to restrict unicode notations to `Utf8` files or at least non-global scopes?  Actually, aren't we trying to put all notations in some scope?

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+
+  Definition subtype_inclusion {A} (U V : 𝒫 A)
```
mikeshulman, line 105: I would put "proper" in the name of this somewhere.

theories/Sets/Hartogs.v
```diff
@@ -0,0 +1,370 @@
...
+    := (forall a, U a -> V a) /\ merely (exists a : A, V a /\ ~U a).
+  Coercion subtype_as_type' {X} (Y : 𝒫 X) : Type
+    := { x : X & Y x }.
+
+  Infix "⊂" := subtype_inclusion (at level 50).
```
mikeshulman, line 110: You don't have to agree, but i generally prefer ⊊ for this relation, since some people use ⊂ to mean ⊆.

theories/Sets/Sierpinski.v
```diff
@@ -0,0 +1,451 @@
...
+From HoTT Require Import PropResizing.PropResizing.
+From HoTT Require Import Spaces.Nat.
+From HoTT Require Import Equiv.BiInv.
+From HoTT Require Import HIT.unique_choice.
+
+From HoTT.Sets Require Import Ordinals Hartogs GCH.
+
+Open Scope type.
+
+(* The proof of Sierpinski's theorem given in this file consists of 2 ingredients:
```
mikeshulman, line 11: I don't think the average reader will know that "Sierpinski's theorem" means "GCH implies AC".  In fact, I don't even see this result appearing at all on the first page of Google hits for "Sierpinski's theorem"; most of the hits are about something topological.  How about calling this file something like `GCHImpliesAC`, and being more explicit in the introductory comments about that too?

theories/Sets/Sierpinski.v
```diff
@@ -0,0 +1,451 @@
...
+    split; try apply H. intros x x'. cbn. intros Hy Hy'. rewrite <- Hy' in Hy. now apply Hf.
+  - intros x. exists (f x). apply tr. exists x. reflexivity.
+  - cbn. intros x. destruct iota as [x' H]. now apply Hf.
+  - cbn. intros [y x]. apply path_sigma_hprop. cbn.
+    destruct iota as [x' Hx]. apply Hx.
+Qed.
+
+
+
+(** * Euivalences on infinite sets *)
```
mikeshulman, line 129:
```suggestion
(** * Equivalences on infinite sets *)
```

theories/Sets/Sierpinski.v
```diff
@@ -0,0 +1,451 @@
...
+  GCH -> infinite X -> hinject X (HN (Build_HSet (X -> HProp))).
+Proof.
+  intros gch HX. eapply hinject_trans; try apply tr, inject_power; try apply X.
+  apply (@Sierpinski_step (Build_HSet (X -> HProp)) HN_bound gch).
+  - apply infinite_inject with X; trivial. apply inject_power. apply X.
+  - intros n. cbn. rewrite !powit_shift. eapply path_infinite_power. cbn. now apply infinite_powit.
+  - apply HN_inject.
+Qed.
+
+Theorem Sierpinski (X : HSet) :
```
mikeshulman, line 396: Similarly, I would name this something more descriptive.

theories/Sets/Sierpinski.v
```diff
@@ -0,0 +1,451 @@
...
+  intros gch. eapply hinject_trans with (nat + X).
+  - apply tr. exists inr. intros x y. apply path_sum_inr.
+  - apply Sierpinski'; trivial. exists inl. intros x y. apply path_sum_inl.
+Qed.
+
+End Sierpinski.
+
+(* Usual set-theoretic formulation of AC *)
+
+Definition AC :=
```
mikeshulman, line 408: This should probably go in a separate file like `Sets/Choice.v`.

mikeshulman: Although to match other cases, we should perhaps have a typeclass `Choice` for when people want to assume it, and then this would be something like `Choice_type`.

theories/Sets/Hartogs.v
dominik-kirst: This is actually just needed to make the coercion below work as we inline the definition everywhere else. Would you in general suggest using the name `power_type` more? In any case, I've now moved the definition to the new `Powers.v` file.

theories/Sets/Sierpinski.v
dominik-kirst: Good point, I've followed your suggestions. I found this naming in Smullyan's and Fitting's textbook and got used to it at some point. In the CPP'21 paper we follow their naming for simplicity but acknowledge that a more prominent result goes by the same name, so overloading the name should be avoided where possible.

theories/Sets/Sierpinski.v
dominik-kirst: I've moved the statement and the implication WO -> AC to `AC.v`. I haven't defined the typeclass yet, I guess if AC should formulated as typeclass then so should also GCH?

Also, shouldn't there also be assumed instances like `Choice_type -> Choice`? In the main result `GCH_AC` currently assumes and instance of `ExcludedMiddle` to enable intermediate lemmas to rely on LEM, although the premise GCH implies LEM.

dominik-kirst:
> We generally try to give global definitions names that make sense globally. Names like `hinject` and `sings` and `powit` are not very descriptive when taken out of their immediate context.

I've renamed `powit` to `power_iterated` and made `sings` a local definition. The idea of `hinject` was to signal the truncation of `inject` as `hexists` does of `exists`. So it seems this is not a general naming pattern and I should switch to a more specific name like `merely_inject`?

dominik-kirst:
The new commits also contain the fixes following Mike's comments that I didn't reply explicitly to. Should I resolve these conversations or do you want to double-check? :)

theories/Sets/Hartogs.v
mikeshulman: I don't know, could go either way.

theories/Sets/Hartogs.v
mikeshulman: Unfortunately I think it's not completely possible to restrict a notation to a particular section or module -- declaring a notation modifies the parser globally.  I'd probably go for a dummy scope.

theories/Sets/Sierpinski.v
mikeshulman: If anyone were going to *use* GCH as an assumption in a theorem, then yes, we'd want it to be a typeclass.  I don't know how likely that is, though.

I don't think it's possible to make `Choice_type -> Choice` an instance, is it?

In general, we deal with implications between axioms by proving them in a metatheory file and then just asserting the implication axiomatically for practical use, such as `Univalence_Implies_Funext` in `Types/Universe.v`.

mikeshulman:
Go ahead and resolve the comments that you fixed.

mikeshulman:
I'm not hugely fond of `hexists` myself, although I don't have a better suggestion at the moment.  (Actually I would probably prefer to call that simply `exists`, which it occurs to me might be possible now.)  What about something like `Injection` for `inject` and `InjectsInto` for `hinject`?

theories/Sets/Sierpinski.v
dominik-kirst: I've added a typeclass called `Choice` as suggested, analogous to the treatment of other axioms.

> If anyone were going to _use_ GCH as an assumption in a theorem, then yes, we'd want it to be a typeclass. I don't know how likely that is, though.

Agreed, I've let GCH as is since it only appears as an assumption in the main result. If in a future development another result with GCH as a premise is mechanised, adding a typeclass is a quick change.

> I don't think it's possible to make `Choice_type -> Choice` an instance, is it?

Yeah adding it with `Existing Instance` doesn't help much as long as `Choice_type` is not declared a typeclass itself. But wouldn't `Choice_type -> Choice` make sense as an axiom (also if not declared an instance), similar for LEM? Then the main result `GCH_AC` could be closed without assuming an instance of `ExcludedMiddle`.

> In general, we deal with implications between axioms by proving them in a metatheory file and then just asserting the implication axiomatically for practical use, such as Univalence_Implies_Funext in Types/Universe.v.

Or should I add an axiom stating `GCH -> ExcludedMiddle`?

dominik-kirst:
> I'm not hugely fond of `hexists` myself, although I don't have a better suggestion at the moment. (Actually I would probably prefer to call that simply `exists`, which it occurs to me might be possible now.) What about something like `Injection` for `inject` and `InjectsInto` for `hinject`?

I've renamed to Injection/InjectsInto now.

Alizter:
We have a file on cardinals that overlaps a bit with what you have. Maybe you would like to use that?

https://github.com/HoTT/HoTT/blob/master/theories/Spaces/Card.v

theories/Sets/Sierpinski.v
mikeshulman: I think the problem with `Choice_type -> Choice`, and the reason we used dummy typeclasses for axioms in the first place, is that a hypothesis of `Choice` effectively assumes choice for all universes at once, whereas a hypothesis of `Choice_type` only assumes it for one choice of universes.

dominik-kirst:
> We have a file on cardinals that overlaps a bit with what you have. Maybe you would like to use that?
> 
> https://github.com/HoTT/HoTT/blob/master/theories/Spaces/Card.v

I could do that, but `Card.v` defines the order on cardinals with `isinj` while we use the typeclass `IsInjective`. Which notion do you prefer to be kept?

Alizter:
@dominik-kirst We've been meaning to merge those two notions for a while, but haven't gotten round to doing it. I would prefer `IsInjective`, but that would then probably have to be moved to `HSet.v`.

dominik-kirst:
Okay, the code now uses `Spaced/Card.v` instead of `Sets/Cardinality.v`. I've changed the definition of `leq_card` to use `IsInjective` and appended the definitions and lemmas about cardinality comparisons to the bottom of the file.

theories/Spaces/Card.v
```diff
@@ -142,3 +145,69 @@ Section contents.
...
+
+
+(** * Cardinality comparisons *)
+
+(* We also work with cardinality comparisons directly to avoid unnecessary type truncations via cardinals. *)
+
+Definition Injection X Y :=
+  { f : X -> Y | IsInjective f }.
+
+Lemma Injection_refl X :
```
Alizter, line 158: You can use the `Reflexive` typeclass and write something like:
```coq
Global Instance reflexive_injection : Reflexive Injection.
Proof.
```

that way you can use the `reflexivity` tactic. Same for `Transitive`.

theories/Spaces/Card.v
```diff
@@ -142,3 +145,69 @@ Section contents.
...
+  Injection nat X.
+
+
+
+(** * Sets of same cardinality are isomorphic *)
+
+(* For this result, we say that two sets X and Y have equal cardinality,
+   if there is an injection from X to Y that is also merely surjective. *)
+
+Lemma equiv_hset_bijection {X Y} (f : X -> Y) :
```
mikeshulman, line 200: Doesn't this follow quickly from `isequiv_surj_emb` in `Truncations/Core`?

mikeshulman (approved):

Alizter:
@dominik-kirst Do you want to work on this further or shall we merge this as is?

theories/Spaces/Card.v
dominik-kirst: I've now declared these lemmas as instances of the `Reflexive` and `Transitive` classes. This leaves me with two questions:
- What is the advantage if of the `Transitivity` typeclass? I'd expect it to be possible to rewrite with injections in the context if the goal is an injection, as with the usual setoid rewriting. This doesn't seem to work though (line 63 in `Powers.v`).
- Why does `reflexivity` not work for mere relations? I've added the instance for `InjectsInto` but `reflexivity` on a trivial claim fails (line 244 in `GCHtoAC.v`).

theories/Spaces/Card.v
dominik-kirst: I've added a proof of `equiv_hset_bijection` using `isequiv_surj_emb` which is indeed a bit shorter (maybe it can be done better). However, this approach seems to rely on funext to show that `IsConnected` is a proposition on the way, so I'm not sure if the first version without funext should be replaced.

dominik-kirst:
> @dominik-kirst Do you want to work on this further or shall we merge this as is?

I've just replied to the remaining two issues, other than that I'd be fine with merging :)

Alizter:
Excellent! I will merge when the CI is green.

theories/Sets/Hartogs.v
JasonGross: This is an alias, and does not modify the parser.  I believe notations are automatically local in sections in any case, but you can add `Local` to make that explicit

theories/Spaces/Card.v
mikeshulman: I don't think there's much advantage to `transitivity`, and sometimes there's a disadvantage that it can introduce extra universe variables.

Alizter: We are not using setoid rewriting. I don't exactly remember the reason though. It is something to investigate in the future. It should be noted that our Transitivity class is different to the coq one, which probably causes some headache for the rewriting tactics.

I'll have a look into what is going on with reflexivity though.

Alizter:
By the way, in the future it is better to make a branch for a PR other than master since it makes it easier for others to checkout.

theories/Spaces/Card.v
mikeshulman: I don't understand why that should be necessary.  Can you post your code?

mikeshulman: You shouldn't need funext; use `BuildIsSurjection` in `Truncations/Core`.

mikeshulman: And for the other hypothesis you can use `isembedding_isinj_hset` in `HSet.v`.  (We should unify `isinj` with `IsInjective`.)

theories/Spaces/Card.v
Alizter: OK, so the reflexivity issue is caused by your definition of `InjectsInto` which is `X -> X -> HProp`. This causes reflexivity a headache for some reason. It can easily be fixed by declaring the type of `InjectsInto` to be `Type`. This breaks a few proofs later on, but they can easily be fixed.

I have the changes ready, if you like I can push to your pr?

dominik-kirst: Sure, if you prefer this just go ahead! Or is there an easy fix such that `reflexivity` can handle mere propositions explicitly typed to `HProp`?

And regarding @mikeshulman's comment, should I then maybe remove the `Transitivity` instances again?

theories/Spaces/Card.v
dominik-kirst: I see, thanks for pointing out these two lemmas, I wasn't aware of them! I've now removed the explicit statement of `equiv_hset_bijection` and just use `isequiv_surj_emb` directly.

dominik-kirst:
> By the way, in the future it is better to make a branch for a PR other than master since it makes it easier for others to checkout.

Oh I didn't know this makes a difference, thanks for letting me know :)

theories/Spaces/Card.v
Alizter: Yes, perhaps it is more sensible to remove the Transitivity instances again. The main advantage of using the instances is uniformity, but there is no need to burden yourself if it doesn't work out as well as before.

dominik-kirst: Done.

Alizter:
OK LGTM! Merging when CI is finished.

jdchristensen:
Should the six files in the Sets folder be exported in HoTT.v?

Alizter:
@jdchristensen I think so. I can do it together with the other one.

dominik-kirst:
Should I also add the references in `HoTTBook.v` for the results on ordinal theory?

Alizter:
@dominik-kirst That would be good too!

PR #1534, opened 2021-05-23
Define hExt and characterise trivial extensions

jarlg:
The first commit adds facts about functoriality of `ab_biprod` and lemmas for working with equations in abelian groups, basically ported from groups. While this may seem redundant, I did not easily find a way to specialize the results for groups without breaking things (specifically `CRing`), and this solution was discussed with @Alizter.

The main content is the thrid commit, which defines a type `hExt` of abelian group extensions, characterises paths in `hExt` and proves that the trivial extensions are the split ones. Some of these results (e.g. characterisation of paths) could be done for all group extensions and then specialised for a type of abelian group extensions, however in general I don't think there's much commonality so it makes sense to define these types separately. A more modular approach made inference troublesome when I tried.

This is part of joint work with my advisor, @jdchristensen.

### DIFF
--- a/theories/Algebra/AbGroups/AbExt.v
+++ b/theories/Algebra/AbGroups/AbExt.v
@@ -1,0 +1,225 @@
+Require Import HoTT.Basics HoTT.Types HSet.
+Require Import Groups AbGroups.AbelianGroup WildCat.
+Require Import Homotopy.ExactSequence Pointed.
+
+Local Open Scope pointed_scope.
+Local Open Scope type_scope.
+Local Open Scope mc_add_scope.
+
+(** * Extensions of abelian groups *)
+
+(** The type of abelian group extensions [A -> E -> B]. *)
+Record hExt (B A : AbGroup) : Type := {
+  hext_E : AbGroup;
+  hext_i : A $-> hext_E;
+  hext_p : hext_E $-> B;
+  hext_isembedding_i : IsEmbedding hext_i;
+  hext_issurjection_p : IsSurjection hext_p;
+  hext_isexact : IsExact (Tr (-1)) hext_i hext_p;
+}.
+
+(** Given an extension [A -> E -> B : hExt B A], we coerce it to [E]. *)
+Coercion hext_E : hExt >-> AbGroup.
+
+Global Existing Instances hext_isembedding_i hext_issurjection_p hext_isexact.
+
+Arguments hext_E {B A E} : rename.
+Arguments hext_i {B A E} : rename.
+Arguments hext_p {B A E} : rename.
+
+Arguments Build_hExt {B A} _ _ _ _ _ _.
+  
+Definition issig_hext {B A : AbGroup} : _ <~> hExt B A := ltac:(issig).
+
+(** A more useful organization of [hExt] as a sigma-type. *)
+Definition issig_hext' {B A : AbGroup}
+  : {X : {E : AbGroup & (A $-> E) * (E $-> B)} &
+          (IsEmbedding (fst X.2)
+           * IsSurjection (snd X.2)
+           * IsExact (Tr (-1)) (fst X.2) (snd X.2))}
+      <~> hExt B A.
+Proof. (** [make_equiv] is slow to solve this, so we prove it by hand. *)
+  refine (issig_hext oE _).
+  equiv_via {X : {E : AbGroup & (A $-> E) * (E $-> B)} &
+                  {_ : IsEmbedding (fst X.2) &
+                       {_ : IsConnMap (Tr (-1)) (snd X.2) & IsExact (Tr (-1)) (fst X.2) (snd X.2)}}}.
+  - apply equiv_functor_sigma_id; intro X.
+    exact (equiv_sigma_prod1 _ _ _)^-1%equiv.
+  - refine (_ oE (equiv_sigma_assoc _ _)^-1%equiv).
+    apply equiv_functor_sigma_id; intro E.
+    exact (equiv_sigma_prod _)^-1%equiv.
+Defined.
+
+(** [hExt B A] is pointed by the split sequence [A -> A+B -> B]. *)
+Global Instance ispointed_hext {B A : AbGroup} : IsPointed (hExt B A).
+Proof.
+  rapply (Build_hExt (ab_biprod A B) ab_biprod_inl ab_biprod_pr2).
+  snrapply Build_IsExact.
+  - srapply Build_pHomotopy.
+    + intro x; cbn.
+      apply ab_unit_l.
+    + apply path_ishprop.
+  - intros [[a b] p]; cbn; cbn in p.
+    rapply contr_inhabited_hprop.
+    apply tr.
+    exists a.
+    srapply path_sigma_hprop; cbn.
+    apply (path_prod' idpath).
+    exact (p^ @ left_identity _).
+Defined.
+
+(** Paths in [hExt] corespond to isomorphisms between the [hext_E]s respecting [hext_i] and [hext_i]. *)
+Proposition equiv_path_hext `{Univalence} {B A : AbGroup} (E F : hExt B A)
+  : (E = F :> hExt B A)
+      <~> {phi : GroupIsomorphism E F &
+                 (phi $o hext_i = hext_i)
+                 * (hext_p $o grp_iso_inverse phi = hext_p)}.
+Proof.
+  refine (_ oE equiv_ap issig_hext'^-1 _ _).
+  refine (_ oE (equiv_path_sigma_hprop _ _)^-1).
+  refine (_ oE (equiv_path_sigma _ _ _)^-1).
+  srapply equiv_functor_sigma'.
+  1: exact equiv_path_abgroup^-1%equiv.
+  intro q; lazy beta.
+  snrefine (_ oE equiv_concat_l _ _); only 3: symmetry.
+  2: exact (grp_homo_compose (equiv_path_abgroup^-1 q) hext_i,
+            (grp_homo_compose hext_p (grp_iso_inverse (equiv_path_abgroup^-1 q)))).
+  1: exact (equiv_path_prod _ _)^-1%equiv.
+  refine (transport_prod _ _ @ _).
+  apply path_prod'.
+  - apply transport_abgrouphomomorphism_from_const.
+  - apply transport_abgrouphomomorphism_to_const.
+Defined.
+
+
+(** * Characterization of trivial extensions
+
+    We characterize trivial extensions in [hExt] as those for which [hext_p] splits. *)
+
+(** If [hext_p : E -> B] splits, we get an induced map [fun e => e - s(hext_p s)] from [E] to [ab_kernel hext_p]. *)
+Definition hext_p_split_to_kernel {B A : AbGroup} (E : hExt B A)
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : GroupHomomorphism E (@ab_kernel E B hext_p).
+Proof.
+  snrapply (grp_kernel_corec (G:=E) (A:=E)).
+  - (** We construct the map [fun e => e - s(hext_p e)] by going via [E + E]. *)
+    rapply grp_homo_compose.
+    (** [fun e => (e,-e)] *)
+    1: exact (ab_biprod_rec grp_homo_id ab_homo_negation).
+    (** [fun (e,f) => e + s(hext_p s)] **)
+    refine (ab_biprod_corec grp_homo_id (s $o hext_p)).
+  - intro x; simpl.
+    refine (grp_homo_op hext_p x _ @ _).
+    refine (ap (fun y => hext_p x + y) _ @ right_inverse (hext_p x)).
+    refine (grp_homo_inv _ _ @ ap (-) _ ).
+    apply h.
+Defined.
+
+(** The composite [A -> E -> (ab_kernel hext_p) + B] is [grp_cxfib]. *)
+Lemma hext_p_split_to_kernel_beta {B A : AbGroup} (E : hExt B A)
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : hext_p_split_to_kernel E h $o hext_i == grp_cxfib cx_isexact.
+Proof.
+  intro a.
+  apply path_sigma_hprop; cbn.
+  refine (ap _ _ @ right_identity (hext_i a)).
+  refine (ap (fun x => - s x) _ @ _).
+  1: rapply cx_isexact.
+  exact (ap _ (grp_homo_unit _) @ negate_mon_unit).
+Defined.
+
+(** The induced map [E -> ab_kernel hext_p + B] is an isomorphism. We suffix it with 1 since it is the first composite in the desired isomorphism [E -> A + B]. *)
+Definition hext_p_split_iso1 {B A : AbGroup} (E : hExt B A)
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : GroupIsomorphism E (ab_biprod (@ab_kernel E B hext_p) B).
+Proof.
+  srapply Build_GroupIsomorphism.
+  - refine (ab_biprod_corec _ hext_p).
+    exact (hext_p_split_to_kernel E h).
+  - srapply isequiv_adjointify.
+    + refine (ab_biprod_rec _ s).
+      rapply subgroup_incl.
+    + intros [a b]; simpl.
+      apply path_prod'.
+      * srapply path_sigma_hprop; cbn.
+        refine ((ab_assoc _ _ _)^ @ _).
+        apply ab_cancelL1.
+        refine (ap _ _ @ ab_inv_r _).
+        apply (ap (-)).
+        refine (grp_homo_op (s $o hext_p) _ _ @ _).
+        refine (ap (fun x => x + _) (ap s a.2 @ grp_homo_unit s) @ _).
+        exact (ab_unit_l _ @ ap s (h b)).
+      * refine (grp_homo_op hext_p a.1 (s b) @ _).
+        exact (ap (fun y => y + _) a.2 @ left_identity _ @ h b).
+    + intro e; simpl.
+      by apply ab_moveR_gM.
+Defined.
+
+(** The full isomorphism [E -> A + B]. *)
+Definition hext_p_split_iso {B A : AbGroup} (E : hExt B A)
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : GroupIsomorphism E (ab_biprod A B).
+Proof.
+  etransitivity (ab_biprod (ab_kernel _) B).
+  - exact (hext_p_split_iso1 E h).
+  - srapply (equiv_functor_ab_biprod (grp_iso_inverse _) grp_iso_id).
+    rapply grp_iso_cxfib.
+Defined.
+
+Proposition hext_p_split_beta `{Funext} {B A : AbGroup} (E : hExt B A)
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : hext_p_split_iso E h $o hext_i == ab_biprod_inl.
+Proof.
+  intro a.
+  refine (ap _ (ab_corec_beta _ _ _ _) @ _).
+  refine (ab_biprod_functor_beta _ _ _ _ _ @ _).
+  srefine (ap (fun p => ab_biprod_corec p _ a) _ @ _).
+  1: exact grp_homo_id.
+  - refine (ap (fun g => _ $o g) _ @ _);
+      apply equiv_path_grouphomomorphism; intro.
+    1: apply hext_p_split_to_kernel_beta.
+    apply eissect.
+  - refine (path_prod' idpath _); cbn.
+    rapply cx_isexact.
+Defined.
+
+(** An extension [E] in [hExt B A] is trivial if and only if it splits. *)
+Proposition iff_hext_trivial_split `{Univalence} {B A : AbGroup} (E : hExt B A)
+  : (E = point _ :> hExt B A)
+      <-> {s : GroupHomomorphism B E & hext_p $o s == idmap}.
+Proof.
+  refine (iff_compose (iff_equiv (equiv_path_hext _ _)) _); split.
+  - intros [phi [g h]].
+    exists (grp_homo_compose (grp_iso_inverse phi) ab_biprod_inr).
+    intro x.
+    refine (equiv_path_grouphomomorphism^-1 h (ab_biprod_inr x) @ _); cbn.
+    apply left_identity.
+  - intros [s h].
+    exists (hext_p_split_iso E h).
+    split; apply equiv_path_grouphomomorphism.
+    + apply hext_p_split_beta.
+    + intros [a b]; simpl.
+      rewrite (left_identity b), (right_identity a).
+      refine (grp_homo_op hext_p  _ _ @ _).
+      refine (ap011 (+) _ (h _) @ _).
+      1: rapply cx_isexact.
+      apply ab_unit_l.
+Defined.
+
+
+(** * The set [Ext B A] of abelian group extensions *)
+
+Definition Ext (B A : AbGroup) := Tr 0 (hExt B A).
+
+Global Instance ispointed_ext {B A : AbGroup} : IsPointed (Ext B A) := tr (point _).
+      
+(** An extension [E : hext B A] is trivial in [Ext B A] if and only if [E] merely splits. *)
+Proposition iff_ab_ext_trivial_split `{Univalence} {B A : AbGroup} (E : hExt B A)
+  : (tr E = point (Ext B A))
+      <~> merely { s : GroupHomomorphism B E & hext_p $o s == idmap }.
+Proof.
+  refine (_ oE (equiv_path_Tr _ _)^-1).
+  srapply equiv_iff_hprop;
+    apply Trunc_functor;
+    apply iff_hext_trivial_split.
+Defined.
--- a/theories/Algebra/AbGroups/AbExt.v
+++ b/theories/Algebra/AbGroups/AbExt.v
@@ -9,49 +9,49 @@ Local Open Scope mc_add_scope.
 (** * Extensions of abelian groups *)
 
 (** The type of abelian group extensions [A -> E -> B]. *)
-Record Exts (B A : AbGroup) : Type := {
-  exts_E : AbGroup;
-  exts_i : A $-> exts_E;
-  exts_p : exts_E $-> B;
-  exts_isembedding_i : IsEmbedding exts_i;
-  exts_issurjection_p : IsSurjection exts_p;
-  exts_isexact : IsExact (Tr (-1)) exts_i exts_p;
+Record AbExt (B A : AbGroup) : Type := {
+  abext_E : AbGroup;
+  abext_i : A $-> abext_E;
+  abext_p : abext_E $-> B;
+  abext_isembedding_i : IsEmbedding abext_i;
+  abext_issurjection_p : IsSurjection abext_p;
+  abext_isexact : IsExact (Tr (-1)) abext_i abext_p;
 }.
 
-(** Given an extension [A -> E -> B : Exts B A], we coerce it to [E]. *)
-Coercion exts_E : Exts >-> AbGroup.
+(** Given an extension [A -> E -> B : AbExt B A], we coerce it to [E]. *)
+Coercion abext_E : AbExt >-> AbGroup.
 
-Global Existing Instances exts_isembedding_i exts_issurjection_p exts_isexact.
+Global Existing Instances abext_isembedding_i abext_issurjection_p abext_isexact.
 
-Arguments exts_E {B A E} : rename.
-Arguments exts_i {B A E} : rename.
-Arguments exts_p {B A E} : rename.
+Arguments abext_E {B A E} : rename.
+Arguments abext_i {B A E} : rename.
+Arguments abext_p {B A E} : rename.
 
-Arguments Build_Exts {B A} _ _ _ _ _ _.
+Arguments Build_AbExt {B A} _ _ _ _ _ _.
 
 (** TODO Figure out why printing this term eats memory and seems to never finish. *)
-Local Definition issig_exts_do_not_print {B A : AbGroup} : _ <~> Exts B A := ltac:(issig).
+Local Definition issig_abext_do_not_print {B A : AbGroup} : _ <~> AbExt B A := ltac:(issig).
 
 (** [make_equiv] is slow if used in the context of the next result, so we give the abstract form of the goal here. *)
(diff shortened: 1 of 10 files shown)


theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+Definition issig_hext {B A : AbGroup} : _ <~> hExt B A := ltac:(issig).
+
+(** A more useful organization of [hExt] as a sigma-type. *)
+Definition issig_hext' {B A : AbGroup}
+  : {X : {E : AbGroup & (A $-> E) * (E $-> B)} &
+          (IsEmbedding (fst X.2)
+           * IsSurjection (snd X.2)
+           * IsExact (Tr (-1)) (fst X.2) (snd X.2))}
+      <~> hExt B A.
+Proof. (** [make_equiv] is slow to solve this, so we prove it by hand. *)
```
Alizter, line 41: Does using `issig` make a difference?

jarlg: No; going via `issig_hext` (just above) is also slow. I did experiment a bit with various combinations, but they all took a couple of seconds on my laptop so I figured it best to write it out.

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+
+Global Existing Instances hext_isembedding_i hext_issurjection_p hext_isexact.
+
+Arguments hext_E {B A E} : rename.
+Arguments hext_i {B A E} : rename.
+Arguments hext_p {B A E} : rename.
+
+Arguments Build_hExt {B A} _ _ _ _ _ _.
+  
+Definition issig_hext {B A : AbGroup} : _ <~> hExt B A := ltac:(issig).
```
jdchristensen, line 32: If I do `Print issig_hext.`, my computer runs out of memory (with coqtop being the culprit).  What's going on here?

theories/Algebra/AbGroups/AbExt.v
jdchristensen: If you put
```coq
  snrapply equiv_adjointify.
  - decomposing_intros.
```
right after the first `refine` in the proof, you see that the tactic is decomposing the abelian group into all of its data and axioms, and also decomposing the `IsExact` element into a few pieces.  Not sure if there's a way to prevent that.  `Opaque IsExact` and `Opaque AbGroup` give errors.

theories/Algebra/AbGroups/AbExt.v
jdchristensen: (`Check issig_hext.` is fine.)

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+
+(** The type of abelian group extensions [A -> E -> B]. *)
+Record hExt (B A : AbGroup) : Type := {
+  hext_E : AbGroup;
+  hext_i : A $-> hext_E;
+  hext_p : hext_E $-> B;
+  hext_isembedding_i : IsEmbedding hext_i;
+  hext_issurjection_p : IsSurjection hext_p;
+  hext_isexact : IsExact (Tr (-1)) hext_i hext_p;
+}.
```
jdchristensen, line 19: Can you remind me why this untruncated type of extensions is called `hExt`?

theories/Algebra/AbGroups/AbExt.v
jdchristensen: Here's one way to organize it:
```coq
(** [make_equiv] is slow if used in the context of the next result, so we give the abstract form of the goal here. *)
Definition issig_hext_helper {AG : Type} {P : AG -> Type} {Q : AG -> Type}
           {R : forall E, P E -> Type} {S : forall E, Q E -> Type} {T : forall E, P E -> Q E -> Type}
  : {X : {E : AG & P E * Q E} & R _ (fst X.2) * S _ (snd X.2) * T _ (fst X.2) (snd X.2)}
  <~> {E : AG & {H0 : P E & {H1 : Q E & {_ : R _ H0 & {_ : S _ H1 & T _ H0 H1}}}}}
  := ltac:(make_equiv).

(** A more useful organization of [hExt] as a sigma-type. *)
Definition issig_hext' {B A : AbGroup}
  : {X : {E : AbGroup & (A $-> E) * (E $-> B)} &
          (IsEmbedding (fst X.2)
           * IsSurjection (snd X.2)
           * IsExact (Tr (-1)) (fst X.2) (snd X.2))}
      <~> hExt B A
  := issig_hext oE issig_hext_helper.
```
It's still a mystery to me why printing `issig_hext` is so dangerous.  I guess there should be a comment warning about this.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -259,8 +310,113 @@ Definition ab_biprod_corec {A B X : AbGroup}
...
+(** *** Functoriality of [ab_biprod] *)
+
+Definition functor_ab_biprod {A A' B B' : AbGroup} (f : A $-> A') (g: B $-> B')
+  : ab_biprod A B $-> ab_biprod A' B'
+  := (ab_biprod_corec (f $o ab_biprod_pr1) (g $o ab_biprod_pr2)).
+
+Definition ab_biprod_functor_beta {Z X Y A B : AbGroup} (f0 : Z $-> X) (f1 : Z $-> Y)
+           (g0 : X $-> A) (g1 : Y $-> B)
+  : functor_ab_biprod g0 g1 $o ab_biprod_corec f0 f1
+    == ab_biprod_corec (g0 $o f0) (g1 $o f1).
```
jdchristensen, line 326: This can also be `$==`.

Also, if the projections are simplified like mentioned above, this proof can be replaced with `fun _ => idpath`.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -259,8 +310,113 @@ Definition ab_biprod_corec {A B X : AbGroup}
...
+       + exact (Build_GroupIsomorphism _ _ f _).
+       + exact (Build_GroupIsomorphism _ _ g _). }
+  all: intros [a b]; simpl.
+  all: apply path_prod'.
+  1,3: refine (ap _ (right_identity _) @ _).
+  3,4: refine (ap _ (left_identity _) @ _).
+  1,3: refine (eisretr _ _ @ _).
+  3,4: refine (eissect _ _ @ _).
+  1,3: apply right_identity.
+  all: apply left_identity.
```
jdchristensen, line 350: With the simpler projections, these six lines become:
```coq
  1,2: apply eisretr.
  all: apply eissect.
```

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -270,3 +426,89 @@ Proof.
     1: exact (fun a => -a).
     1-2: exact negate_involutive.
 Defined.
+
+Definition ab_add_homo {A : AbGroup}
+  : ab_biprod A A $-> A
+  := ab_biprod_rec grp_homo_id grp_homo_id.
+
+(** ** Working with equations in groups *)
```
jdchristensen, line 434: I wish we could just reuse the equations for groups...  What did you try and what went wrong?

theories/Classes/interfaces/canonical_names.v
```diff
@@ -241,6 +241,15 @@ Class RightInverse {A} {B} {C} (op : A -> B -> C) (inv : A -> B) (unit : C)
...
   := commutativity: forall x y, f x y = f y x.
 
+Global Instance ishprop_commutative `{Funext} {G : Type}
+       `{IsTrunc 0 G} (m : SgOp G) : IsHProp (Commutative m).
+Proof.
+  apply hprop_allpath.
+  intros x y.
+  funext k l.
+  apply path_ishprop.
+Defined.
```
jdchristensen, line 251: I think this should be replaced by `Hint Unfold Commutative : typeclass_instances.`, since the proof is already found by typeclass search, once `Commutative` is unfolded.  I suspect that other things here should have similar hints added, which would avoid needing to unfold things so often.

theories/Algebra/AbGroups/AbExt.v
jarlg: How about not including `issig_hext`, since the useful version is really `issig_hext'`? Printing `issig_hext` also hangs on my computer, not quite sure how to debug a term I can't inspect...

As for the alternative proof of `issig_hext'`, is that really an improvement? I guess an argument in favor is that it illustrates how to help out `make_equiv`.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -259,8 +310,113 @@ Definition ab_biprod_corec {A B X : AbGroup}
            (f : X $-> A) (g : X $-> B)
   : X $-> ab_biprod A B := grp_prod_corec f g.
 
-(** The negation automorphism of an abelian group *)
-Definition ab_homo_negation { A : AbGroup } : GroupIsomorphism A A.
+Definition ab_corec_beta {X Y A B : AbGroup} (f : X $-> Y) (g0 : Y $-> A) (g1 : Y $-> B)
+  : ab_biprod_corec g0 g1 $o f == ab_biprod_corec (g0 $o f) (g1 $o f)
```
jdchristensen, line 314: I must have forgotten to save this comment:  This can be `$==`, which I think is saying something slightly stronger, even if it unfolds to `==` after coercion.

theories/Homotopy/ExactSequence.v
```diff
@@ -222,6 +222,30 @@ Proof.
...
+  rapply (cancelL_mapinO _ _ pr1).
+Defined.
+
+Definition equiv_cxfib {O : Modality} {F X Y : pType} {i : F ->* X} {f : X ->* Y}
+           `{forall y y' : Y, In O (y = y')} `{MapIn O _ _ i} (ex : IsExact O i f)
+  : F <~>* pfiber f := Build_pEquiv _ _ _ (isequiv_cxfib ex).
+
+Proposition equiv_cxfib_beta {F X Y : pType} {i : F ->* X} {f : X ->* Y}
+            `{forall y y' : Y, In O (y = y')} `{MapIn O _ _ i} (ex : IsExact O i f)
+  : i o* pequiv_inverse (equiv_cxfib ex) == pr1. (** poitned pr1? *)
```
jdchristensen, line 241: Use `o` instead of `o*` unless you upgrade it to a pointed homotopy equivalence.  Typo in comment.  Remove comment?

theories/Algebra/AbGroups/AbExt.v
jdchristensen: Since you needed to type a large term with `equiv_via` anyways, I definitely prefer the proof with `make_equiv`, which makes it very clear that there is nothing to see here.

jdchristensen: If you get rid of `issig_hext` and instead embed `issig` in the proof of `issig_hext'`, then I suspect that printing `issig_hext'` will become a problem.  (I didn't check, though.)  We could keep `issig_hext` but rename it as `issig_hext_do_not_print`, or something like that.  :-)

jarlg: Alright, I'm changing to your version with `Local` definitions for the two helper results `issig_hext_helper` and `issig_hext_do_not_print`. I'll include a TODO about figuring out why this explodes unless someone chimes in with any insight.

theories/Algebra/AbGroups/AbExt.v
jarlg: There's no good justification -- the `h` reminds me that it's a space and not a set. Is there a common notation for a type whose set-truncation is a group G? Open to suggestions.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -38,10 +38,40 @@ Proof.
   nrapply (Build_Group G _ _ _ l).
 Defined.
 
+Definition equiv_abgroup_commutative_group
+  : { G : Group & Commutative (@group_sgop G) } <~> AbGroup
```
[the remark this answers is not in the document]
jarlg: Good catch, I originally had the equivalence the other way. (But this way it corresponds to issig-type equivalences.)

I did try redefining `AbGroup` as you say but got stuck making things in`CRing` compile. Attempting this now again, I got it to work. I'm opening a separate PR for that so we can compare and discuss.

theories/Homotopy/ExactSequence.v
Alizter: It should be easy to add ppr1 for pointed projections.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -38,10 +38,40 @@ Proof.
...
-(** ** Subgroups of abelian groups *)
+Section AbGroupLaws.
+  Context {A : AbGroup} (x y z : A).
+
+  Definition ab_assoc := associativity x y z.
+  Definition ab_unit_l := left_identity x.
+  Definition ab_unit_r := right_identity x.
+  Definition ab_inv_l := left_inverse x.
+  Definition ab_inv_r := right_inverse x.
+  Definition ab_comm := commutativity x y.
```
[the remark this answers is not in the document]
jarlg: I'm not experiencing any breakage, and could have written `apply left_identity` in all the places I've written `ab_unit_l`. So that wasn't the reason for the change.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -38,10 +38,40 @@ Proof.
...
+  : GroupIsomorphism A B <~> (A = B).
+Proof.
+  refine (equiv_ap_inv equiv_abgroup_commutative_group _ _ oE _).
+  refine (equiv_path_sigma_hprop _ _ oE _).
+  exact equiv_path_group.
+Defined.
+
+Definition equiv_path_abgroup_group `{Univalence} {A B : AbGroup}
+  : (A = B :> AbGroup) <~> (A = B :> Group)
+  := equiv_path_group oE equiv_path_abgroup^-1.
```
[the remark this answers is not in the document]
jarlg: Right, thanks.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -167,6 +197,7 @@ Proof.
   apply path_prod; simpl; apply commutativity.
 Defined.
 
+(** These are known to be embeddings, since their versions for [grp_prod] are. *)
```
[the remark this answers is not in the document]
jarlg: Yes, I should make that clearer.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -38,10 +38,40 @@ Proof.
...
+  refine (equiv_ap_inv equiv_abgroup_commutative_group _ _ oE _).
+  refine (equiv_path_sigma_hprop _ _ oE _).
+  exact equiv_path_group.
+Defined.
+
+Definition equiv_path_abgroup_group `{Univalence} {A B : AbGroup}
+  : (A = B :> AbGroup) <~> (A = B :> Group)
+  := equiv_path_group oE equiv_path_abgroup^-1.
+
+(** ** hSubgroups of abelian groups *)
```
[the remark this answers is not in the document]
jarlg: Absolutely.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -193,6 +224,26 @@ Definition ab_biprod_pr1 {A B : AbGroup} : ab_biprod A B $-> A
...
   := ab_biprod_rec grp_homo_const grp_homo_id.
 
+Global Instance issurjection_ab_biprod_pr1 {A B : AbGroup}
+  : IsSurjection (@ab_biprod_pr1 A B).
+Proof.
+  intro a.
+  rapply contr_inhabited_hprop.
+  apply tr.
+  exists (a, mon_unit); cbn.
+  apply ab_unit_r.
```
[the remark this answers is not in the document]
jarlg: Surprised this isn't already in the library, at least I cannot find it. Adding `issurj_retr` to the end of `Truncations/Connectedness.v`.

Good catch on redefining the `_pr1` stuff!

theories/Algebra/AbGroups/AbelianGroup.v
jarlg: Nice!

theories/Classes/interfaces/canonical_names.v
jarlg: Indeed, that seems better.

theories/Algebra/AbGroups/AbelianGroup.v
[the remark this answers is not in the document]
Alizter: If things are breaking that is not good. If we can isolate examples that were known to work before, we should find any commits that may have broken this. The specific movement lemmas for groups and abgroups are not needed, but provide a more rigid interface for use. Using left_identity is completely fine, but I can easily cook up examples where it struggles and the more specific lemmas should be used instead. I think it's better to be redundant than unusable.

theories/Algebra/AbGroups/AbelianGroup.v
jarlg: Actually this isn't saying anything more general. The `isgraph` instance on morphisms between abelian groups (which gives the `$==` notation) is induced from groups, which is again induced from Type (i.e. boils down to `isgraph_arrow`) which gives exactly what's written here. (Try `Print isgraph_grouphomomorphism.` to see what I mean.)

But I'm changing it to `$==`, unless there's an argument against, because in my experience this might make things play more nicely with other results about wild categories.

theories/Algebra/Groups/ShortExactSequence.v
```diff
@@ -8,11 +8,32 @@ Local Open Scope path_scope.
...
+  : GroupHomomorphism A (grp_kernel f)
+  := grp_kernel_corec cx.
+
+Definition grp_cxfib_homotopic {A B C : Group} {i : A $-> B} {f : B $-> C} (cx : IsComplex i f)
+  : cxfib cx == grp_cxfib cx := fun _ => idpath.
+
+Global Instance isequiv_grp_cxfib {A B C : Group} {i : A $-> B} {f : B $-> C}
+           `{IsEmbedding i} (ex : IsExact (Tr (-1)) i f)
+  : IsEquiv (grp_cxfib cx_isexact)
+  := isequiv_homotopic (cxfib cx_isexact) (grp_cxfib_homotopic _).
```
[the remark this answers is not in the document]
jarlg: Nice! I'll go with inlining it.

theories/Homotopy/ExactSequence.v
jarlg: @Alizter read my mind :) I forgot to treat this comment, thanks.

jarlg:
I've pushed the various changes discussed above, and will rebase onto #1535 (or whatever we decide on there) when that has been merged.

theories/Algebra/AbGroups/AbelianGroup.v
jdchristensen: Right, that's what I was hinting at.  Even though it is definitionally the same, it is in some sense syntactically stronger in that it might let you use it with other results.

theories/Homotopy/ExactSequence.v
```diff
@@ -238,7 +238,7 @@ Definition equiv_cxfib {O : Modality} {F X Y : pType} {i : F ->* X} {f : X ->* Y
 
 Proposition equiv_cxfib_beta {F X Y : pType} {i : F ->* X} {f : X ->* Y}
             `{forall y y' : Y, In O (y = y')} `{MapIn O _ _ i} (ex : IsExact O i f)
-  : i o* pequiv_inverse (equiv_cxfib ex) == pr1. (** poitned pr1? *)
+  : i o pequiv_inverse (equiv_cxfib ex) == pfib _.
```
jdchristensen, line 241: Since you switched to a pointed map on the RHS, maybe it would make sense to use `o*` and `==*`?

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+  refine (ap _ (ab_corec_beta _ _ _ _) @ _).
+  refine (ab_biprod_functor_beta _ _ _ _ _ @ _).
+  srefine (ap (fun p => ab_biprod_corec p _ a) _ @ _).
+  1: exact grp_homo_id.
+  - refine (ap (fun g => _ $o g) _ @ _);
+      apply equiv_path_grouphomomorphism; intro.
+    1: apply hext_p_split_to_kernel_beta.
+    apply eissect.
+  - refine (path_prod' idpath _); cbn.
+    rapply cx_isexact.
```
jdchristensen, line 179: The highlighted lines can be replaced with
```coq
  nrapply path_prod'.
  2: rapply cx_isexact.
  refine (ap _ (hext_p_split_to_kernel_beta E h a) @ _).
  apply eissect.
```
and `Funext` is no longer required.

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+Defined.
+
+(** An extension [E] in [hExt B A] is trivial if and only if it splits. *)
+Proposition iff_hext_trivial_split `{Univalence} {B A : AbGroup} (E : hExt B A)
+  : (E = point _ :> hExt B A)
```
jdchristensen, line 184: Maybe `(E = point (hExt B A))` or just `(E = point _)`?

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+    apply (path_prod' idpath).
+    exact (p^ @ left_identity _).
+Defined.
+
+(** Paths in [hExt] corespond to isomorphisms between the [hext_E]s respecting [hext_i] and [hext_i]. *)
+Proposition equiv_path_hext `{Univalence} {B A : AbGroup} (E F : hExt B A)
+  : (E = F :> hExt B A)
+      <~> {phi : GroupIsomorphism E F &
+                 (phi $o hext_i = hext_i)
+                 * (hext_p $o grp_iso_inverse phi = hext_p)}.
```
jdchristensen, line 72: How do things work out if you use homotopies instead of equalities here?

theories/Algebra/AbGroups/AbelianGroup.v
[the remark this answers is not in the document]
jdchristensen: I take it back.  The problem with `left_identity` went away when I retested, so I must have had something messed up, e.g. my `coq-project-filename`, or something like that.

theories/Algebra/AbGroups/AbExt.v
jarlg: Let's do `Exts`, which is also short enough to be used as notation in a paper (as opposed to "Extensions", which would be fine in code but I'd prefer not to write on paper).

Alizter: Maybe I am missing something, but isn't hExt already a set? In which case you can just call it Ext?

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+  - apply transport_abgrouphomomorphism_from_const.
+  - apply transport_abgrouphomomorphism_to_const.
+Defined.
+
+
+(** * Characterization of trivial extensions
+
+    We characterize trivial extensions in [hExt] as those for which [hext_p] splits. *)
+
+(** If [hext_p : E -> B] splits, we get an induced map [fun e => e - s(hext_p s)] from [E] to [ab_kernel hext_p]. *)
```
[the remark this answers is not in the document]
jarlg: Yep, thanks!

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+  1: exact equiv_path_abgroup^-1%equiv.
+  intro q; lazy beta.
+  snrefine (_ oE equiv_concat_l _ _); only 3: symmetry.
+  2: exact (grp_homo_compose (equiv_path_abgroup^-1 q) hext_i,
+            (grp_homo_compose hext_p (grp_iso_inverse (equiv_path_abgroup^-1 q)))).
+  1: exact (equiv_path_prod _ _)^-1%equiv.
+  refine (transport_prod _ _ @ _).
+  apply path_prod'.
+  - apply transport_abgrouphomomorphism_from_const.
+  - apply transport_abgrouphomomorphism_to_const.
```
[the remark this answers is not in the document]
jarlg: Unless you do `symmetry` again after `equiv_functor_sigma'`, the variable `q` ends up being an isomorphism instead of a path and this makes the last two lines (90 and 91) a bit more troublesome. I prefer it this way.

Alizter:
Looks like we need a rebase

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+      {s : GroupHomomorphism B E} (h : hext_p $o s == idmap)
+  : GroupHomomorphism E (@ab_kernel E B hext_p).
+Proof.
+  snrapply (grp_kernel_corec (G:=E) (A:=E)).
+  - (** We construct the map [fun e => e - s(hext_p e)] by going via [E + E]. *)
+    rapply grp_homo_compose.
+    (** [fun e => (e,-e)] *)
+    1: exact (ab_biprod_rec grp_homo_id ab_homo_negation).
+    (** [fun (e,f) => e + s(hext_p s)] **)
+    refine (ab_biprod_corec grp_homo_id (s $o hext_p)).
```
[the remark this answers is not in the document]
jarlg: Nice! I'm leaving the full group instance for another time.

theories/Algebra/AbGroups/AbExt.v
```diff
@@ -0,0 +1,225 @@
...
+    + intros [a b]; simpl.
+      apply path_prod'.
+      * srapply path_sigma_hprop; cbn.
+        refine ((ab_assoc _ _ _)^ @ _).
+        apply ab_cancelL1.
+        refine (ap _ _ @ ab_inv_r _).
+        apply (ap (-)).
+        refine (grp_homo_op (s $o hext_p) _ _ @ _).
+        refine (ap (fun x => x + _) (ap s a.2 @ grp_homo_unit s) @ _).
+        exact (ab_unit_l _ @ ap s (h b)).
```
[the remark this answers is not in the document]
jarlg: Thanks!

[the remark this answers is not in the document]
jdchristensen: I guess `apply (ap ((-) o s))` should work to combine two of the lines.  (Untested.)

theories/Algebra/AbGroups/AbExt.v
jarlg: It makes things slightly more troublesome. The line
```coq
  1: exact (equiv_path_prod _ _)^-1%equiv.
```
would need to be expanded to
```coq
  1: { refine (_ oE (equiv_path_prod _ _)^-1%equiv).
       nrapply equiv_functor_prod'.
       all: apply equiv_path_grouphomomorphism^-1%equiv. }
```
Unless there are arguments otherwise, I'll stick with the current version.

theories/Homotopy/ExactSequence.v
jarlg: It certainly makes sense, but I don't need it and it's not just the matter of adding a few lines (as far as I can tell). The inverse of `cxfib` doesn't compute, so one needs to precompose with `cxfib` (as I've done here). The corresponding pointed thing is to prewhisker with a pointed equivalence, which we don't currently have a lemma for (and which I don't want to spend time proving now).

jarlg:
I've rebased. Note the second to last commit "Mark IsPointed as Transparent". With the new formalization of `AbGroup`, inference has slightly changed and an issue with applying `istrunc_paths` when e.g. `ispointed_group` is part of the goal surfaced. I've seen this before, and one can solve it by unfolding. The general solution however is to mark the typeclass as transparent, I believe.

theories/Types/Sigma.v
```diff
@@ -516,6 +516,10 @@ Proof.
   make_equiv.
 Defined.
 
+Definition equiv_sigma_prod1 (A B C : Type)
```
mikeshulman, line 519: What's "1" about this?

mikeshulman:
Coming a bit late to this party, but don't we usually give things singular names?  `Group` not `Groups`, `Modality` not `Modalities`, etc.  So `Exts` doesn't really fit.  I didn't quite follow what was wrong with `Ext` or `Extension`?

mikeshulman:
(In light of having this theory, possibly the file `theories/Extensions` should be renamed something like `ExtensionAlong`.)

theories/Algebra/AbGroups/AbExt.v
jdchristensen: The reason I suggested it is that later you use this result when proving `iff_hext_trivial_split`, and there both directions simplify.  So overall it's roughly neutral, but homotopies fit better with how the other results in this file are stated.

jdchristensen:
`Ext` is the set-truncation of the "Ext space" `Exts`.  I agree that `Exts` isn't a great name.  I guess `Extension` might be good.  I thought it was used already, but I see that only `ExtensionAlong` is used.

mikeshulman:
If `Extension` seems too ambiguous, we could say `AbExtension`.

theories/Homotopy/ExactSequence.v
jdchristensen: That's fine.  I'm just confused about why you wanted to use a pointed map on the RHS, instead of just `pr1`.

jdchristensen:
Or how about `AbExt` for the type of extensions and `Ext` for the set?  That also matches the filename.

mikeshulman:
That would be okay.  I like including `Ab` in the name in case we later also study extensions of nonabelian groups, say `GpExt`.

theories/Types/Sigma.v
jarlg: I've added one more layer to
```coq
equiv_sigma_prod0 (A B : Type) : {a : A & B} <~> A * B.
```
just above. This isn't needed anymore though, after the `make_equiv` proof of `issig_hext` supplied by @jdchristensen, so I can simply remove it if that's preferable.

mikeshulman:
Ah, I see, so it's just continuing the bad naming of `equiv_sigma_prod0`.
(-:  Once you've written it there's no reason not to include it, but it
would be nice if we could find better names for all these lemmas.

theories/Algebra/AbGroups/AbExt.v
jdchristensen: @jarlg Let me know if you want to make this change, or leave it like it is.  And whether you have any other changes to make.  If not, I'll merge.

jarlg: Go ahead and merge. I'll consider making this change if stronger reasons crop up in the future, but current I'm happy with it as-is. I don't have any other changes to make.

PR #1535, opened 2021-05-24
Redefine AbGroup to be commutative groups.

jarlg:
To prove things in #1534, we had to port lemmas from `Group` for working with equations in `AbGroup`. This can be resolved by coercing elements in `AbGroup` via the underlying `Group` instead of a direct coercion `AbGroup >-> Sortclass` (to the underlying type). However then it seems cleanest to simply define `AbGroup` as commutative groups, and we do this here.

(I definitely recommend split view for the diff!)

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -12,34 +12,18 @@ Local Open Scope mc_add_scope.
 (** Definition of an abelian group *)
 
 Record AbGroup := {
-  abgroup_type : Type;
-  abgroup_sgop : SgOp abgroup_type;
-  abgroup_unit : MonUnit abgroup_type;
-  abgroup_inverse : Negate abgroup_type;
-  abgroup_isabgroup : IsAbGroup abgroup_type;
+  abgroup_group : Group;
+  abgroup_commutative : Commutative (@group_sgop abgroup_group);
 }.
 
-Existing Instances abgroup_sgop abgroup_unit abgroup_inverse abgroup_isabgroup.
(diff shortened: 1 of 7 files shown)


theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -12,35 +12,19 @@ Local Open Scope mc_add_scope.
...
+  abgroup_group : Group;
+  abgroup_commutative : Commutative (@group_sgop abgroup_group);
 }.
 
-Existing Instances abgroup_sgop abgroup_unit abgroup_inverse abgroup_isabgroup.
+Coercion abgroup_group : AbGroup >-> Group.
 
-(** We want abelian groups to be coerced to the underlying type. *)
-Coercion abgroup_type : AbGroup >-> Sortclass.
+Existing Instance abgroup_commutative.
```
Alizter, line 21: Shouldn't this be global

Alizter:
I'm OK with this change. The reason it was originally done this way was to confirm with `IsAbGroup`.

There are of course multiple ways of building up the strucutre of an abelian group, and we are seeing a tension between two of them.

There are ways of resolving this multiple inheritance but I will leave it to a future discussion.

Alizter (approved):
LGTM

theories/Algebra/AbGroups/AbelianGroup.v
jarlg: Yes, thanks!

jdchristensen (approved):
Nice.  Overall that seems to have simplified things, and should also simplify things in #1534.

PR #1536, opened 2021-05-25
Delete quotient

Alizter:


Alizter:
I'm closing this for now since I had some upstream commits but the build is getting stuck. I will need to do some debugging.

PR #1539, opened 2021-05-29
stop using .git to find .v files and instead use find

### LINKED ISSUE #1538: Opam build doesn't work with autogenerated _CoqProject
We are making the assumption that there will be a .git folder that can be used for _CoqProject to be generated, but during opam builds for example, this isn't around. We should therefore implement Jason's suggestion of checking if .git exists first and doing something else if it doesn't. https://github.com/HoTT/HoTT/pull/1513#issuecomment-841803022

Alizter:
closes #1538

### DIFF
--- a/etc/generate_coqproject.sh
+++ b/etc/generate_coqproject.sh
@@ -1,5 +1,5 @@
 ## List tracked .v files
-TRACKED_V_FILES=$(git ls-files "*.v")
+TRACKED_V_FILES=$(find * -type f -name *.v)
 
 ## List untracked .v files
 #UNTRACKED_V_FILES=$(git ls-files --others --exclude-standard "*.v") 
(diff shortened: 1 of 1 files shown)


JasonGross (changes requested):
Please don't do this unconditionally. It makes it impossible to have temporary or work in progress files.  Instead you should check whether or not the .git directory exists, and use find only if it doesn't.

etc/generate_coqproject.sh
```diff
@@ -1,5 +1,5 @@
 ## List tracked .v files
-TRACKED_V_FILES=$(git ls-files "*.v")
+TRACKED_V_FILES=$(find theories/ contrib/ -type f -name *.v)
```
JasonGross, line 2:
```suggestion
if [ -d .git ]; then
  TRACKED_V_FILES="$(git ls-files "*.v")"
else
  echo "Warning: Not a git clone, using find instead" >&2
  TRACKED_V_FILES="$(find theories/ contrib/ -type f -name "*.v")"
fi
```

mikeshulman (approved):

PR #1542, opened 2021-05-31
Reduce universe levels in hfiber_functor_sigma

jdchristensen:
By avoiding `symmetry` and `transitivity`, we get rid of two larger universes in `hfiber_functor_sigma`.  Only a couple of other places are affected by this.  This is useful for WIP where I'm tracking universe usage.

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -34,12 +34,12 @@ Proof.
   refine (equiv_ap_inv issig_abgroup _ _ oE _).
   About equiv_path_sigma_hprop.
   refine (equiv_path_sigma_hprop _ _ oE _).
-  exact equiv_path_group@{v w u u u u u u u u u}.
+  exact equiv_path_group@{v u u u u u u u u u}.
 Defined.
 
 Definition equiv_path_abgroup_group `{Univalence} {A B : AbGroup}
   : (A = B :> AbGroup) <~> (A = B :> Group)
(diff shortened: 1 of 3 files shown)


mikeshulman (approved):

Alizter:
Why do these tactics introduce new universes?

jdchristensen:
For example, `symmetry` is part of the type class
```coq
Class Symmetric {A} (R : Relation A) :=
  symmetry : forall x y, R x y -> R y x.
```
The type class has two universe variables:
```coq
Symmetric@{u u0} = 
fun (A : Type@{u}) (R : Relation@{u u0} A) => forall x y : A, R x y -> R y x
     : forall A : Type@{u}, Relation@{u u0} A -> Type@{max(u,u0)}
```
Since `Equiv` is a relation on Type, we have `A` equal to `Type` here, which means that `u` needs to be a universe variable for a universe containing `Type`, i.e. one level higher.  The proof of this typeclass is `equiv_inverse`, which doesn't need to refer to this larger universe.

SkySkimmer:
Could also be done ala https://github.com/HoTT/HoTT/pull/1358

jdchristensen:
@SkySkimmer It looks like the `symmetry` and `transitivity` tactics are already done like `reflexivity`, so I'm not sure what you are proposing exactly.  If there's a way to eliminate this universe variable globally, that would be great.

SkySkimmer:
Basically every `Reflexivity`/`Symmetry`/`Transitivity` instance which can produce the extra universe needs `Arguments /`

Alizter:
I'll submit a patch that does this later today.

PR #1543, opened 2021-05-31
Fix universe introduced by symmetry and transitivity when working with Equiv

Alizter:
Hopefully this should address the issues raised in #1542


theories/PropResizing/Nat.v
```diff
@@ -194,7 +194,7 @@ Section AssumeStuff.
   Defined.
 
   Definition graph_succ_path_equiv@{}
-    := Eval unfold graph_succ_path_equiv0 in graph_succ_path_equiv0@{u}.
+    := Eval unfold graph_succ_path_equiv0 in graph_succ_path_equiv0@{}.
```
SkySkimmer, line 197: Instead you should rename the `foo0` to `foo` and add the `@{}` to enforce it doesn't get extra universes in the future

Alizter: Alright, I think I got most of them now.

theories/PropResizing/Nat.v
```diff
@@ -42,7 +42,7 @@ Section AssumeStuff.
     := pr2 (pr2 A) x y.
   (** We will need universe annotations in a few more places, but not many. *)
 
-  Definition equiv_path_graph0 (A B : Graph)
+  Definition equiv_path_graph (A B : Graph)
```
SkySkimmer, line 45:
```suggestion
  Definition equiv_path_graph@{} (A B : Graph)
```

theories/PropResizing/Nat.v
```diff
@@ -416,7 +397,7 @@ Section AssumeStuff.
   Normally addition is defined *using* recursion, but here we can
   "cheat" because we know how to add graphs, and then prove that it
   satisfies the recursive equations for addition. *)
-  Definition graph_add0 (A B : Graph) : Graph.
+  Definition graph_add (A B : Graph) : Graph.
```
SkySkimmer, line 400:
```suggestion
  Definition graph_add@{} (A B : Graph) : Graph.
```

theories/PropResizing/Nat.v
```diff
@@ -805,7 +782,7 @@ Section AssumeStuff.
     (** The crucial point that makes it work for arbitrary [X] is to
     prove in one big induction that these types are always
     contractible.  Here is the base case. *)
-    Lemma contr_partial_Nrec_zero0 : Contr (partial_Nrec zero).
+    Lemma contr_partial_Nrec_zero : Contr (partial_Nrec zero).
```
SkySkimmer, line 785:
```suggestion
    Lemma contr_partial_Nrec_zero@{} : Contr (partial_Nrec zero).
```

theories/PropResizing/Nat.v
```diff
@@ -868,7 +843,7 @@ Section AssumeStuff.
     Qed.
 
     (** And here, essentially, is the inductive step. *)
-    Local Definition partial_Nrec_succ0 (n : N)
+    Local Definition partial_Nrec_succ (n : N)
```
SkySkimmer, line 846:
```suggestion
    Local Definition partial_Nrec_succ@{} (n : N)
```

SkySkimmer (dismissed):

Alizter:
@SkySkimmer Thanks!

theories/PropResizing/Nat.v
```diff
@@ -384,17 +376,6 @@ Section AssumeStuff.
       apply path_N; reflexivity.
   Qed.
 
-  (** Sometimes we just need a bigger fish. *)
-  Universe large.
```
SkySkimmer, line 388: CI is not happy because `N_propind` was not the only use of `large`.

jdchristensen (approved):
LGTM.  This fixes the original problem I had, and also reduces universe variables in a number of other places.

Out of curiosity, what was this old code doing:
```coq
  Definition equiv_path_graph@{} := ltac:(first[exact equiv_path_graph0@{u}|exact equiv_path_graph0]).		
```
Was it accommodating different versions of Coq?

jdchristensen:
To answer my own question, I think that some of these changes to Nat.v are unrelated to the change to symmetry and transitivity, and are just getting rid of legacy code.  (Which is fine, of course.  I was just unsure whether there was a connection.)

SkySkimmer:
>Out of curiosity, what was this old code doing:
Was it accommodating different versions of Coq?

Yes, before https://coq.github.io/doc/master/refman/addendum/universe-polymorphism.html#coq:flag.Private-Polymorphic-Universes (https://github.com/coq/coq/pull/8850 Coq 8.10) the proof for `equiv_path_graph` had an extra public universe.

theories/PropResizing/Nat.v
```diff
@@ -868,7 +843,7 @@ Section AssumeStuff.
     Qed.
 
     (** And here, essentially, is the inductive step. *)
-    Local Definition partial_Nrec_succ0 (n : N)
+    Local Definition partial_Nrec_succ@{} (n : N)
```
jdchristensen, line 846:
```suggestion
    Local Definition partial_Nrec_succ0 (n : N)
```

theories/PropResizing/Nat.v
```diff
@@ -939,8 +914,6 @@ Section AssumeStuff.
         refine (equiv_sigma_symm _ oE _).
         exact ((equiv_sigma_contr _)^-1%equiv). }
     Defined.
-    Local Definition partial_Nrec_succ@{}
-      := Eval unfold partial_Nrec_succ0 in partial_Nrec_succ0@{nr nr nr nr}.
```
jdchristensen, line 943: Reinstate these two lines.

theories/PropResizing/Nat.v
SkySkimmer: No, the `0` means this is an intermediate definition meant to be used only through a wrapper to get rid of universes.

theories/PropResizing/Nat.v
```diff
@@ -939,14 +915,12 @@ Section AssumeStuff.
...
```
SkySkimmer, line 943: Please restore this instead of doing it inline later

Alizter:
@jdchristensen I have added the `large` universe back in (the place it was defined really matters). I kept the removal of the legacy code too however. I think it is good to remove legacy code when you touch a file like this, otherwise it will never happen.

We should come back in the future and do a nice cleanup too, perhaps even move these files to MetaTheory since they don't really serve a purpose other than to prove a point.

Alizter:
I will do some squashing before merging

theories/PropResizing/Nat.v
```diff
@@ -387,14 +379,6 @@ Section AssumeStuff.
   (** Sometimes we just need a bigger fish. *)
   Universe large.
 
-  (** Importantly, however, the universe of the motive [P] is not
-      constrained but can be arbitrary. *)
```
jdchristensen, line 391: I think this comment should be preserved, above the definition.

theories/PropResizing/Nat.v
```diff
@@ -366,7 +358,7 @@ Section AssumeStuff.
   Qed.
 
   (** Final Peano axiom: induction *)
```
Alizter, line 360:
```suggestion
  (** Importantly, the universe of the motive [P] is not constrained but can be arbitrary. *)
  (** Final Peano axiom: induction *)
```

theories/PropResizing/Nat.v
Alizter: Ok, i've added it a bit further up

theories/PropResizing/Nat.v
jdchristensen: Picking nits here, but it would make more sense to have these comments in the other order (and probably as one comment):
```coq
  (** Final Peano axiom: induction.  Importantly, the universe of the motive [P] is not constrained but can be arbitrary. *)
```

jdchristensen: @Alizter I think this minor change to the comment is the last thing remaining.  Want me to make the change and merge this?

Alizter: @jdchristensen that would be good thanks!

PR #1544, opened 2021-05-31
Pull HasEquivFuns out of HasEquivs

andreaslyn:
A have made a new class `HasEquivFuns` for the equivalence notation `a $<~> b`, since the laws in `HasEquivs` may require `Funext` or `Univalence`. With this new class we can use the `a $<~> b`, even if an axiom is required for `HasEquivs`.

My first attempt was to only take `CatEquiv'` out of `HasEquivs`, but then the useful coercion `Coercion cate_fun : CatEquiv >-> Hom` got invalid.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -508,12 +508,17 @@ Proof.
   intros []; reflexivity. 
 Defined.
 
-Global Instance hasequivs_group : HasEquivs Group.
+Global Instance hasequivfuns_group : HasEquivFuns Group.
 Proof.
-  unshelve econstructor.
+  srapply Build_HasEquivFuns.
   + exact GroupIsomorphism.
   + exact (fun G H f => IsEquiv f).
   + intros G H f; exact f.
+Defined.
+
+Global Instance hasequivs_group : HasEquivs Group.
+Proof.
+  unshelve econstructor.
   + exact Build_GroupIsomorphism.
   + intros G H; exact grp_iso_inverse.
   + cbn; exact _.
--- a/theories/Algebra/Rings/CRing.v
+++ b/theories/Algebra/Rings/CRing.v
@@ -275,12 +275,17 @@ Proof.
   intros []; reflexivity. 
 Defined.
 
-Global Instance hasequivs_cring : HasEquivs CRing.
+Global Instance hasequivfuns_cring : HasEquivFuns CRing.
 Proof.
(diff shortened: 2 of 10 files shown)


Alizter:
@andreaslyn I am confused why you think the laws in `HasEquivs` might require funext. Have you got an example where this has come up?

mikeshulman:
Yes, the wildcat library was specifically designed to *not* require funext, by using homotopies rather than equalities as higher cells.

andreaslyn:
In order to construct `HasEquivs (Algebra sig)`, I need to show that the inverse of an (infinitary) algebra isomorphism is a homomorphism. I believe this requires funext, because I have defined isomorphism to be a homomorphism `f : forall s, A s -> B s` which is pointwise an equivalence.

However, I just realized that an alternative is to define `IsIsomorphism f` by a homomorphism g, such that `f $o g = Id` and `g $o f = Id`. I think this is actually better, since I avoid a use of funext.

mikeshulman:
I'm still confused.  If `f` is pointwise an equivalence, you can assemble the inverses into a family of maps in the other direction, right?  What goes wrong?

It doesn't seem right to me either to define an isomorphism in terms of *equality* of functions, since in order to *use* that in practice you'll need funext.

andreaslyn:
I was too quick when I wrote `f $o g = Id`. What I meant was `forall s x, f s (g s x) = x`.

About what goes wrong. Let `f : A -> B` be a homomorphism and pointwise equivalence. To show that the inverse `g := f^-1` is a homomorphism I need
```
g (op B b) = op A (fun i => g (b i))
```
where `op X : (I -> X) -> X` is an operation. It suffices to show that
```
op B b = f (op A (fun i => g (b i))) = op B (fun i => f (g (b i)))
```
using that `f` is a homomorphism. Here I need funext for `b = fun i => f (g (b i)`.

mikeshulman:
Ah, interesting!  I wonder whether in the absence of funext an infinitary algebra should be *defined* to be invariant on pointwise paths between inputs.

mikeshulman:
With your `forall s x, f s (g s x) = x` version, I guess you can prove without funext that the category of algebras `HasEquivs`.  But can you construct any particular equivalence therein without funext?

andreaslyn:
You are right. This is a problem!

Yes, I think that your suggestion invariance under pointwise paths of inputs will solve it. I need to look into this.

Thank you for the input!

Issue #1481, opened 2021-04-24
Generalize strip_truncations to strip_reflections?


jdchristensen:
Here's a handy result.  At first, you might think it needs `Funext`, since that is needed to show that `Y -> Z` is O-local when `Z` is O-local.  But by juggling the context, you can avoid `Funext`.

```coq
Lemma O_concat {O : ReflectiveSubuniverse} {A : Type} {a0 a1 a2 : A}
  : O (a0 = a1) -> O (a1 = a2) -> O (a0 = a2).
Proof.
  intro Op.
  rapply O_rec; intro q.
  revert Op; rapply O_rec; intro p.
  exact (to O _ (p @ q)).
Defined.
```

This can be abstracted.  E.g. here's something that just handles the first argument.

```coq
Definition O_rec_to_arrow (O : ReflectiveSubuniverse) {X Y Z : Type} `{In O Z}
           (f : X -> (Y -> Z))
  : (O X) -> (Y -> Z).
Proof.
  intros ox y.
  revert ox.
  apply O_rec.
  intro x.
  exact (f x y).
Defined.

Definition O_rec_to_arrow_beta (O : ReflectiveSubuniverse) {X Y Z : Type} `{In O Z}
           (f : X -> (Y -> Z)) (x : X)
  : (O_rec_to_arrow O f) ((to O X) x) == (f x).
Proof.
  intro y.
  rapply O_rec_beta.
Defined.
```

With this, one can write:

```coq
Lemma O_concat' {O : ReflectiveSubuniverse} {A : Type} {a0 a1 a2 : A}
  : O (a0 = a1) -> O (a1 = a2) -> O (a0 = a2).
Proof.
  rapply O_rec_to_arrow; intro p.
  rapply O_rec; intro q.
  exact (to O _ (p @ q)).
Defined.
```

Even better, maybe we should generalize `strip_truncations` to `strip_reflections`:

```coq
(** A tactic that generalizes strip_truncations. *)
Ltac strip_reflections :=
  (** Search for hypotheses of type [O X] for some [O] such that the goal is [O]-local. *)
  progress repeat match goal with
                    | [ T : _ |- _ ]
                      => revert_opaque T;
                        (** The dependent case requires that [O] be a modality. *)
                        refine (@O_rec _ _ _ _ _ _ _) || refine (@O_ind _ _ _ _ _ _ _);
                        (** Ensure that we didn't generate more than one subgoal, i.e. that the goal was appropriately local. *)
                        [];
[...]

PR #1546, opened 2021-06-01
strip_reflections, strip_modalities tactics

Alizter:
As discussed in #1481

Perhaps it would be better to have a `strip_using` tactic where we provide an induction principle, than all three become special cases of that.

I also wonder why `strip_truncations` is so good at what it does. Replacing it in `Classes/implementations/HProp_lattice.v` for example breaks things. Projective.v also breaks.

### DIFF
--- a/theories/Modalities/Modality.v
+++ b/theories/Modalities/Modality.v
@@ -215,8 +215,8 @@ Ltac strip_modalities :=
     match goal with
     | [ T : _ |- _ ]
       => revert_opaque T;
-        (** The dependent case requires that [O] be a modality. *)
-        refine (@O_ind@{_ _ _} _ _ _ _ _ _ _);
+        (** Handle the non-dependent and dependent cases.  The last case requires that [O] be a modality. *)
+        refine (@O_rec _ _ _ _ _ _ _) || refine (@O_indpaths _ _ _ _ _ _ _ _ _) || refine (@O_ind _ _ _ _ _ _ _);
         (** Ensure that we didn't generate more than one subgoal, i.e. that the goal was appropriately local. *)
         [];
         intro T
--- a/theories/Modalities/Modality.v
+++ b/theories/Modalities/Modality.v
@@ -208,6 +208,19 @@ Proof.
   srapply (isequiv_oD_to_O O O).
 Defined.
 
+(** A tactic that generalizes [strip_reflections] to modalities. *)
+Ltac strip_modalities :=
+  (** Search for hypotheses of type [O X] for some [O] such that the goal is [O]-local. *)
+  progress repeat
+    match goal with
+    | [ T : _ |- _ ]
+      => revert_opaque T;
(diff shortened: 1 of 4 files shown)


theories/Projective.v
```diff
@@ -154,7 +154,7 @@ Section AC_oo_neg1.
     pose (X := Build_HSet (Tr 0 A)).
     pose proof ((equiv_isoprojective_hasochoice _ X)^-1 (AC X)) as P.
     pose proof (P A _ X _ idmap tr _) as F; clear P.
-    strip_truncations.
+    revert F; refine (Trunc_ind@{_ _} _ _); intros F.
```
mikeshulman, line 157: Why this change?  (Sorry if I missed it.)

mikeshulman:
Would it be worth including `O_indpaths` too?

theories/Modalities/Modality.v
```diff
@@ -208,6 +208,19 @@ Proof.
...
 
+(** A tactic that generalizes [strip_reflections] to modalities. *)
+Ltac strip_modalities :=
+  (** Search for hypotheses of type [O X] for some [O] such that the goal is [O]-local. *)
+  progress repeat
+    match goal with
+    | [ T : _ |- _ ]
+      => revert_opaque T;
+        (** The dependent case requires that [O] be a modality. *)
+        refine (@O_ind@{_ _ _} _ _ _ _ _ _ _);
```
jdchristensen, line 219: I wonder if we should try `O_rec` *and* `O_ind` in the modality case?  Maybe there's an advantage to using `O_rec` when it applies?  Also, it would mean that the user can type `strip_modalities` and it would clear reflective subuniverses and modalities, when both are present.

Also:  why the `@{_ _ _}`?

Alizter: I was experimenting with collapsing universes so it could faithfully replace strip_truncations. Couldn't get it to work however.

jdchristensen:
@Alizter Would you like me to make a few changes to finish this off?

Alizter:
@jdchristensen that would be good. I haven't got so much time this week.

theories/Projective.v
jdchristensen: I'm guessing it slipped in by accident.  I'm reversing it.

theories/Modalities/Modality.v
jdchristensen: I'm having strip_reflections to O_rec and O_indpaths, and strip_modalities do those two plus O_ind.

jdchristensen:
I think I've handled the remaining issues.  I tested both new tactics, covering all of the cases.  I'll wait for one "LGTM" before merging.

jdchristensen:
I just added a commit that demonstrates strip_reflections by using it in a few places that had the "revert; apply ...; intro" idiom.  (I also removed a few nearby rewrites.)

jdchristensen:
Merging, but if there is feedback later, I'm happy to make fixes.

PR #1547, opened 2021-06-05
Use git ls-files in a worktree as well

jdchristensen:
In a git worktree, .git is a file, not a directory, so use -e instead of -d to test for it.

### DIFF
--- a/etc/generate_coqproject.sh
+++ b/etc/generate_coqproject.sh
@@ -1,5 +1,5 @@
 ## List tracked .v files
-if [ -d .git ]; then
+if [ -e .git ]; then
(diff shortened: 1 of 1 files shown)


mikeshulman (approved):

Alizter:
Wait I am confused, how is .git not a directory?

jdchristensen:
With git worktrees, you can checkout multiple branches at the same time, in different directories.  All of the directories are based on the same git repo.  The original directory has the .git directory, and the worktrees have a .git file which tells git where to find the .git directory.  My .git directory is around 1GB, so this is a big savings if you want to work with multiple branches at once.

PR #1548, opened 2021-06-05
Export more things via HoTT.vo

jdchristensen:
This tries to address #1511.

With this change, `make theories/HoTT.vo` builds everything except (some) files in the directories `Axioms`, `PropResizing`, `Classes`, `Categories`, `MetaTheory` and `contrib`.  Also, the top-level files `Categories.v`, `Tests.v`, `Utf8.v` are excluded.  The only other file excluded is `Algebra/Groups/FreeGroup/LoopSusp.v`.  Let me know if it should be included.

However, I suspect that there are files in the main library directories that are built for HoTT.vo for dependency reasons, but aren't exported by HoTT.vo.  Does anyone know a quick way to check what is missing?


Alizter (approved):

mikeshulman:
Why is `Algebra/Groups/FreeGroup/LoopSusp.v` excluded?

jdchristensen:
I had a vague memory that someone suggested it not be exported, maybe because of the issues mentioned in this comment:
```coq
(** Note that this is NOT the definition of a free group that is used in the library. The definition in use will be exported in the file Algebra.Groups.FreeGroup. This file merely exists to serve as a demonstration that the free group *could* be defined this way.

This is not a very practical definition because we cannot even show that it is free at the time of writing, and univalence is used in a crucial way which seems overkill for a definition of freegroup. *)
```
But I'm happy to add it if that makes sense.

JasonGross:
> Does anyone know a quick way to check what is missing?

You can look at [the dependency graphs in the wiki](https://hott.github.io/HoTT/dependencies/HoTTCore.svg)

jdchristensen:
@JasonGross I don't think that addresses what I'm asking.  For example, I'd like to know that every file in one of the directories we intend to export in HoTT.vo (e.g. Basics, Types, Modalities, Algebra, etc.) is either directly or indirectly being exported.  The facts that some of the exports are indirect and that the export syntax doesn't match path names makes it slightly tricky, so before coming up with a shell script, I thought I'd ask if there's a quick way to check this.

(By observing make's behaviour, I already know that the dependencies of what are exported include all of the files in those directories, so what I'm asking about involves ignoring dependencies.)

Alternatively, we could just merge the changes above, and if anyone notices a file that is not exported, we fix it then.

mikeshulman:
Regarding free groups, it does seem to me that the LoopSusp file could be useful.  Even if we take the ListQuotient definition as "the" definition of a free group, what the LoopSusp file proves is that the fundamental group of the suspension of A+1 is isomorphic to that free group, which is an important statement.  Thinking of it this way does suggest a different naming scheme, i.e. ListQuotient and LoopSusp should not both define an object called `FreeGroup` if we might potentially want to import both of them.

jdchristensen:
Yes, I agree.  I think that `FreeGroup` in `LoopSusp` should be renamed (along with anything else the conflicts).  As it stands, if we export both from `HoTT.v`, then there is risk of confusion.  How about we leave that as a separate issue (which I am happy to open, but probably won't tackle)?

BTW, I did some poking around in the Basics directory and found these files that are not exported by HoTT.v or Basics.v: Basics/Datatypes, Basics/Decimal, Basics/Hexadecimal, Basics/Logic.  I'm sure there are others in other folders, but just thinking of those:  are they intentionally excluded, or should they be added to Basics.v?

jdchristensen:
Another question:  should every directory Foo have a file Foo.v which gathers together the things that will be exported from Foo?  Like Basics.v.  E.g. should Modalities.v be added?

jdchristensen:
Basics/Hexadecimal, for example, is available after doing `Require Import HoTT`, but just requires a qualified name, e.g. `Hexadecimal.hexidecimal`.  So maybe there's no need to worry about whether everything is exported (possible indirectly) by HoTT.v, since all it affects is whether a longer name is needed.  The changes I made do ensure that everything in the main directories is built, so I think it follows that all those things are available after `Require Import HoTT`, possibly with a longer name.  So maybe it's fine to merge this as it and export more things later if people ask.

mikeshulman:
>  How about we leave that as a separate issue (which I am happy to open, but probably won't tackle)?

Sounds good.

> Basics/Datatypes, Basics/Decimal, Basics/Hexadecimal, Basics/Logic

My guess would be that Datatypes and Logic should be exported, but maybe Decimal and Hexadecimal are only used for printing/parsing numeral notations and don't need to be exported?

> should every directory Foo have a file Foo.v which gathers together the things that will be exported from Foo? Like Basics.v. E.g. should Modalities.v be added?

I think I never created a Modalities.v because it seemed less likely to me that another file would want to import "everything from Modalities" the same way almost everything wants to import "everything from Basics" or "everything from Types".  The latter two are so fundamental that it's not reasonable to expect people to specify exactly what they want from them and what they don't, but for Modalities it seemed better to me to let users specify which of the files they want to import, since they're more likely to be aware of what they're using and what they're not, and being more specific is better for dependency management.  I think the same reasoning applies to directories like Algebra and Spaces.  But I'm open to contrary arguments.

jdchristensen:
Ah, Basics/Overture.v already exports Basics/Datatypes.v and Basics/Logic.v, so they don't need to be separately listed.

I think it'll be too tedious and have little reward to pursue this further.  Maybe we can just merge the changes I made, and see whether anything else comes up in the future.

jdchristensen:
I'm merging this as is.  Further improvements can of course be made in future PRs.

Issue #1545, opened 2021-06-01
Further stratify cells and coherence in wildcat library


Alizter:
I have a few arguments on paper using wildcats that I would like to formalise. Some of these require 3-cells for example, but not much coherence at all. Currently we only have Is21Cat providing these cells, which has a lot more coherence than what I need it for. There is also the question of how to organise the coherence data, but I think for Is21Cat, all or nothing is sufficient.

I am therefore proposing we stratify (at least) the cells of a wildcat as `IsGlob`, `Is1Glob`, `Is2Glob` etc. I know it is possible to do this for a general `n` but I don't really see it as a useful thing to do right now. I know Mike is experimenting with having all cells up to oo with coinductive types, but there are a few more things to be worked out there.

What I am proposing is intending to make the wildcat library a bit more immediately useful.

mikeshulman:
Sure.

PR #1550, opened 2021-06-16
Stratify cells in wildcat

Alizter:
I have gone ahead and split the cells data for wildcats as discussed in #1545. We now have the typeclasses `Is2Graph` and `Is3Graph`.

This also prompted me to go through the wildcat library and do a little cleanup. This included naming things properly etc. Unfortunately, there are example where coq isn't smart enough to guess things anymore. (Especially when mixing `->*` and `$->`). The solution is to probably try to stick to wildcat notation as much as possible.

This was much trickier than anticipated to carry out however. The completely unbundled approach to algebraic hierarchies is not very rigid with respect to changes lower down, and the number of underscores in the code has increased a bit. I can quickly see that this will get out of hand fast, so perhaps we need to bundle some things up in a clever way at some point.

### DIFF
--- a/theories/WildCat/Equiv.v
+++ b/theories/WildCat/Equiv.v
@@ -31,7 +31,7 @@ Class HasEquivs (A : Type) `{Is1Cat A} :=
 (** Since apparently a field of a record can't be the source of a coercion (Coq complains about the uniform inheritance condition, although as officially stated that condition appears to be satisfied), we redefine all the fields of [HasEquivs]. *)
 
 Definition CatEquiv {A} `{HasEquivs A} (a b : A)
-  := @CatEquiv' A _ _ _ _ a b.
+  := @CatEquiv' A _ _ _ _ _ a b.
 
 Notation "a $<~> b" := (CatEquiv a b).
 Infix "≅" := CatEquiv : wc_iso_scope.
@@ -39,7 +39,7 @@ Arguments CatEquiv : simpl never.
 
 Definition cate_fun {A} `{HasEquivs A} {a b : A} (f : a $<~> b)
   : a $-> b
-  := @cate_fun' A _ _ _ _ a b f.
+  := @cate_fun' A _ _ _ _ _ a b f.
 
 Coercion cate_fun : CatEquiv >-> Hom.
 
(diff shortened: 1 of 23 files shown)


theories/WildCat/Equiv.v
```diff
@@ -298,9 +299,11 @@ Global Instance is0functor_core_precomp {A : Type} `{HasEquivs A}
 Proof.
   apply Build_Is0Functor.
   intros f g al; cbn in h.
-  exact (compose_cate_fun f h
-           $@ (al $@R h)
+  (** Why can't coq resolve this? *)
+  refine (compose_cate_fun f h
+           $@ (_ $@R h)
            $@ (compose_cate_fun g h)^$).
+  exact al.
```
Alizter, line 306: Not sure why coq got dumber here.

mikeshulman (approved):

mikeshulman:
What sort of "getting out of hand" do you anticipate?  This seems to me to be a discrete change that's over with when finished, not the start of a lot of further changes that would add additional underscores.

I guess one way to reduce the underscores would be to go the oo-route and just assume that everything is an oo-graph.  I forget whether we decided there are any examples where the whole oo-graph structure is hard to define?

If we don't go that far, we could still collapse at least the definitions of n-graphs into something inductive or recursive.

Alizter:
In the future, we might want to write down a definition of 2/pseudo-functor. Already looking at the Is1Functor instances we can see that having so many unbundled typeclasses around leads to many underscores. It will be even worse if we ever start to write down such definitions. This is what I mean by getting out of hand. But I guess this is something expected when working with higher categories, the amount of data grows very quickly and you really have to do something clever to handle it.

I believe the oo-graph route is the way forward eventually, but there are a few things that need to be worked out there first. I agree that this will reduce the number of underscores we have.

The only example I can think of where it might be hard to write down the oo-graph structure is the "category" whose n-cells are n-transfors (perhaps invertible at some point). I'm not sure this is a blocker for adopting oo-graphs however.

Doing something inductive or recursive may be possible, but it got too tricky for me when I tried this a few months back. I also didn't see the point if we are going to go for oo-graphs anyway.

Anyway, the reason I wanted to do this was so I could push through with the 3x3 lemma for pushouts which has been sitting on my todo problem list for too long. 0:-)

mikeshulman:
Yes, looking back at #1302 I remember now that there were cases where it was at least tricky to correctly define all the higher cells, like pType.  But also that we should be able to cop out in any case by defining the dimensions we care about at the bottom and using ordinary paths above that.  It might be worth experimenting with *just* introducing oo-graphs to replace graphs/2graphs/3graphs, but keeping all the other structures the way they are.

Issue #1555, opened 2021-06-21
Huge slowdown in Classes/implementations/pointwise.v


Alizter:
There is a huge slowdown in `Classes/implementations/pointwise.v`. Using `git bisect` I managed to track down the offending commit: 3a2973ea

The compile time for `Classes/implementations/pointwise.v` at 3a2973ea is `(real: 63.66, user: 63.40, sys: 0.19, mem: 395212 ko)`.

The compile time for `Classes/implementations/pointwise.v` at e819147e (the commit before) is `(real: 4.94, user: 4.75, sys: 0.18, mem: 378256 ko)`.

Something has gone really wrong!

jdchristensen:
I guess it's this:
```coq
Global Hint Unfold Commutative : typeclass_instances.
```
I'm pinging Jarl: @jarlg

jdchristensen:
Really the problem is the automation used in the `try_solve_fun` tactic in pointwise.v.  I have a simple change that greatly speeds up the file (1.5s).  I'll make a PR in a bit.

Alizter:
Yeah, I agree that the automation in Classes can be a bit crude sometimes. This is understandable, since it wasn't written for HoTT but only ported later on.

jarlg:
Thanks, good catch! Changing the `Hint Unfold Commutative` to `Typeclasses Transparent Commutative` gets the build time of `pointwise.v` down to normal, and also shaves off almost 20s over the rest of the library. Unless there are arguments otherwise, maybe we should make that change as well? Of course, fixing/removing the tactic used is good too.

Alizter:
@jarlg That sounds like a good idea! Do you want to do a PR?

jarlg:
The speedup didn't survive #1557, I'm now getting only ~3s improvement with the change, so I'll leave things as they are.

jdchristensen:
If the proposed change has the same benefits as the existing `Hint`, and might prevent slow type checking in the future, maybe it is still a good idea?

PR #1558, opened 2021-06-21
Change Hint Unfold Commutative to Typeclasses Transparent.

jarlg:
The issue in #1555 would have been prevented if I'd made `Commutative` a transparent instance, instead of adding the more intrusive `Hint Unfold`. This is how I should have done it from the start, and should prevent similar issues in the future.

Edit: This saves ~4s, as can be seen here: https://gist.github.com/jarlg/5e75f9195ee9379f84fbb38f55d47653


jdchristensen:
LGTM.

I was curious why so many files were significantly slower, so I ran the timing myself, but I didn't see this in my data.  (The largest increase was 0.08s.)  I also didn't see much of a speedup overall, just 1.17s (0.50%).  Still, it seems worth it to prevent future issues.

PR #1560, opened 2021-06-24
Changing various functors to use fmap + misc wildcat additions

Alizter:
Summary of changes:
 * Changed many definitions/lemmas about `X_functor` to use the wildcat `fmap` instead. This is by no means comprehensive, but a good look at what can be done.
 * Added various lemmas about initial and terminal objects.
 * Added some theory of pointed functors (functors between pointed categories)
 * Cleaned up a few proofs along the way.
 * Finished proof that pType is univalent.
 * Probably some others that I missed.

This work is culminating from another project I am working on.

### DIFF
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -149,15 +149,11 @@ Definition group_loops_functor_idmap {X : pType}
   == group_loops_functor (Id (A:=pType) _).
 Proof.
   intros g.
-  unfold grouphom_fun, grouphom_idmap.
-  assert (p := eisretr (loops_group X) g).
   refine (fmap_id loops _ g @ _).
-  rewrite <- p.
+  rewrite <- (eisretr (loops_group X) g).
+  unfold grouphom_fun, grouphom_idmap.
   rewrite !loops_functor_group.
-  pose (q := pointed_htpy
-    (fmap_id loops X : _ ==* _) ((loops_group X)^-1 g)).
-  rewrite q.
-  reflexivity.
+  exact (ap (loops_group X) (fmap_id loops _ _)^).
 Qed.
 
 (** *** Homomorphic properties *)
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -74,7 +74,7 @@ Definition ooGroupHom (G H : ooGroup)
   := B G ->* B H.
 
 Definition grouphom_fun {G H} (phi : ooGroupHom G H) : G -> H
-  := loops_functor phi.
+  := fmap loops phi.
 
 Coercion grouphom_fun : ooGroupHom >-> Funclass.
 
@@ -95,8 +95,8 @@ Defined.
 (** And this functor "is" the same as the ordinary loop space functor. *)
 Definition loops_functor_group
            {X Y : pType} (f : X ->* Y)
-: loops_functor (group_loops_functor f) o loops_group X
-  == loops_group Y o loops_functor f.
+: fmap loops (group_loops_functor f) o loops_group X
+  == loops_group Y o fmap loops f.
 Proof.
   intros x.
   apply (equiv_inj (equiv_path_sigma_hprop _ _)^-1).
@@ -129,16 +129,16 @@ Definition group_loops_functor_compose
 Proof.
   intros g.
   unfold grouphom_fun, grouphom_compose.
-  refine (pointed_htpy (loops_functor_compose _ _) g @ _).
+  refine (pointed_htpy (fmap_comp loops _ _) g @ _).
   pose (p := eisretr (loops_group X) g).
-  change (loops_functor (group_loops_functor psi)
-            (loops_functor (group_loops_functor phi) g)
-          = loops_functor (group_loops_functor
+  change (fmap loops (group_loops_functor psi)
+            (fmap loops (group_loops_functor phi) g)
+          = fmap loops (group_loops_functor
                                  (pmap_compose psi phi)) g).
   rewrite <- p.
   rewrite !loops_functor_group.
   apply ap.
-  symmetry; apply loops_functor_compose.
+  symmetry; rapply (fmap_comp loops).
 Qed.
 
 Definition grouphom_idmap (G : ooGroup) : ooGroupHom G G
@@ -146,32 +146,19 @@ Definition grouphom_idmap (G : ooGroup) : ooGroupHom G G
 
 Definition group_loops_functor_idmap {X : pType}
 : grouphom_idmap (group_loops X)
-  == group_loops_functor pmap_idmap.
+  == group_loops_functor (Id (A:=pType) _).
 Proof.
   intros g.
   unfold grouphom_fun, grouphom_idmap.
   assert (p := eisretr (loops_group X) g).
+  refine (fmap_id loops _ g @ _).
   rewrite <- p.
   rewrite !loops_functor_group.
-  rewrite (pointed_htpy (loops_functor_idmap _)).
-  simpl. apply ap.
-  rewrite concat_p1, concat_1p, ap_idmap.
+  pose (q := pointed_htpy
+    (fmap_id loops X : _ ==* _) ((loops_group X)^-1 g)).
+  rewrite q.
   reflexivity.
 Qed.
-
-Definition group_loops_2functor
-           {X Y : pType} {phi psi : X ->* Y}
-           (theta : pHomotopy phi psi)
-: (group_loops_functor phi) == (group_loops_functor psi).
-Proof.
-  intros g.
-  assert (p := eisretr (loops_group X) g).
-  rewrite <- p.
-  unfold grouphom_fun.
-  rewrite !loops_functor_group.
-  apply ap.
-  apply loops_2functor; assumption.
-Defined.
 
 (** *** Homomorphic properties *)
 
@@ -307,13 +294,13 @@ Global Instance is0functor_group_to_oogroup : Is0Functor group_to_oogroup.
 Proof.
   snrapply Build_Is0Functor.
   intros G H f.
-  by rapply functor_pclassifyingspace.
+  by rapply (fmap pClassifyingSpace).
 Defined.
 
 Global Instance is1functor_group_to_oogroup : Is1Functor group_to_oogroup.
 Proof.
-  snrapply Build_Is1Functor.
-  1: exact @functor2_pclassifyingspace.
-  1: exact functor_pclassifyingspace_idmap.
-  exact functor_pclassifyingspace_compose.
+  snrapply Build_Is1Functor; hnf; intros.
+  1: by rapply (fmap2 pClassifyingSpace).
+  1: by rapply (fmap_id pClassifyingSpace).
+  by rapply (fmap_comp pClassifyingSpace).
 Defined.
--- a/theories/Homotopy/ExactSequence.v
+++ b/theories/Homotopy/ExactSequence.v
@@ -46,26 +46,25 @@ Defined.
 (** Truncation preserves complexes. *)
 Definition iscomplex_ptr (n : trunc_index)
        {F X Y : pType} (i : F ->* X) (f : X ->* Y) (cx : IsComplex i f)
-  : IsComplex (ptr_functor n i) (ptr_functor n f).
+  : IsComplex (fmap (pTr n) i) (fmap (pTr n) f).
 Proof.
-  refine ((ptr_functor_pmap_compose n i f)^* @* _).
+  refine ((fmap_comp (pTr n) i f)^* @* _).
   refine (_ @* ptr_functor_pconst n).
-  apply ptr_functor_homotopy; assumption.
+  rapply (fmap2 (pTr _)); assumption.
 Defined.
 
 (** Loopspaces preserve complexes. *)
 Definition iscomplex_loops
        {F X Y : pType} (i : F ->* X) (f : X ->* Y) (cx : IsComplex i f)
-  : IsComplex (loops_functor i) (loops_functor f).
+  : IsComplex (fmap loops i) (fmap loops f).
 Proof.
-  refine ((loops_functor_compose f i)^* @* _ @* loops_functor_pconst).
-  apply loops_2functor; assumption.
+  refine ((fmap_comp loops i f)^$ $@ _ $@ fmap_zero_morphism _).
+  rapply (fmap2 loops); assumption.
 Defined.
 
-Definition iscomplex_iterated_loops
-           {F X Y : pType} (i : F ->* X) (f : X ->* Y)
-           (cx : IsComplex i f) (n : nat)
-  : IsComplex (iterated_loops_functor n i) (iterated_loops_functor n f).
+Definition iscomplex_iterated_loops {F X Y : pType}
+  (i : F ->* X) (f : X ->* Y) (cx : IsComplex i f) (n : nat)
+  : IsComplex (fmap (iterated_loops n) i) (fmap (iterated_loops n) f).
 Proof.
   induction n as [|n IHn]; [ exact cx | ].
   apply iscomplex_loops; assumption.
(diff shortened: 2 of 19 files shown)


theories/Algebra/ooGroup.v
```diff
@@ -129,50 +129,37 @@ Definition group_loops_functor_compose
...
   assert (p := eisretr (loops_group X) g).
+  refine (fmap_id loops _ g @ _).
   rewrite <- p.
   rewrite !loops_functor_group.
-  rewrite (pointed_htpy (loops_functor_idmap _)).
-  simpl. apply ap.
-  rewrite concat_p1, concat_1p, ap_idmap.
+  pose (q := pointed_htpy
+    (fmap_id loops X : _ ==* _) ((loops_group X)^-1 g)).
+  rewrite q.
```
Alizter, line 159: rewrite is being dumb here for some reason

theories/Homotopy/HomotopyGroup.v
```diff
@@ -196,26 +184,22 @@ Proof.
...
 Proof.
   destruct n; intros x.
   all:refine ((O_functor_compose 0 _ _ _)^ @ _ @ (O_functor_compose 0 _ _ _)).
   all:apply O_functor_homotopy.
   - reflexivity.
-  - exact (pointed_htpy (unfold_iterated_loops_functor n.+1 f)).
+  - exact (pointed_htpy (unfold_iterated_fmap_loops n.+1 f)).
 Defined.
 
 Definition groupiso_pi_functor (n : nat)
```
Alizter, line 199: This is a common trend in HomotopyGroup.v. Many proofs became much much shorter.

theories/Pointed/Core.v
```diff
@@ -664,8 +664,16 @@ Proof.
...
+  - simpl.
+    refine (_^ @ (concat_p1 _)^).
+    rapply (equiv_moveL_Mp _ _ _)^-1.
+    refine (_ @ (concat_p1 _)^).
+    rewrite transport_paths_FlFr.
+    hott_simpl.
+    rewrite transport_path_universe_equiv_path.
+    rewrite ap_V.
+    apply inv_V.
+Defined.
```
Alizter, line 676: The proof wasn't so bad.

theories/Pointed/Loops.v
```diff
@@ -39,181 +39,167 @@ Definition istrunc_loops {n} (A : pType) `{IsTrunc n.+1 A}
...
+Lemma pequiv_loops_punit : loops pUnit <~>* pUnit.
+Proof.
+  snrapply Build_pEquiv'.
+  { srapply (equiv_adjointify (fun _ => tt) (fun _ => idpath)).
+    1: by intros [].
+    rapply path_contr. }
+  reflexivity.
+Defined. 
+
 (** ** Functoriality of loop spaces *)
```
Alizter, line 51: Many proofs here became much shorter.

theories/Pointed/pSusp.v
```diff
@@ -32,63 +32,47 @@ Definition psusp (X : pType) : pType
 
 (** ** Suspension Functor *)
```
Alizter, line 33: Similar savings here. Hopefully soon we will have a notion of 1-adjunction which should do most of the work here more generally.

theories/Truncations/Core.v
```diff
@@ -93,6 +93,9 @@ Section TruncationModality.
     : Tr@{i} n X -> Tr@{j} n Y
     := O_functor@{k k k} (Tr n) f.
 
+  Global Instance is0functor_Tr : Is0Functor (Tr n)
+    := Build_Is0Functor _ _ _ _ (Tr n) (@Trunc_functor).
+
```
jdchristensen, line 98: In other cases, you got rid of `..._functor`.  Is there a reason to keep `Trunc_functor`?

theories/Truncations/Core.v
```diff
@@ -118,6 +121,14 @@ Section TruncationModality.
     : Tr n (X * Y) <~> Tr n X * Tr n Y
     := equiv_O_prod_cmp (Tr n) X Y.
 
+  Global Instance is1functor_Tr : Is1Functor (Tr n).
+  Proof.
+    apply Build_Is1Functor.
+    - apply @O_functor_homotopy.
+    - apply @Trunc_functor_idmap.
+    - apply @Trunc_functor_compose.
+  Defined.
```
jdchristensen, line 130: Similarly, it other cases you got rid of `..._idmap`, `..._compose`, etc.

theories/WildCat/PointedCat.v
```diff
@@ -14,34 +14,99 @@ Global Existing Instance isterminal_zero_object.
...
+  Proof.
+    refine (_ $@L (compose_V_hh be f)^$ $@ _).
+    refine (cat_assoc_opp _ _ _ $@ _).
+    refine (_ $@L (mor_terminal_unique a _ _)^$ $@ _).
+    exact ((mor_initial_unique _ _ _)^$ $@R _).
   Defined.
 
 End ZeroLaws.
 
+(** We show the last two arguments so that end pointes can easily be specified. We had to do this again, since the section encapsulated the previous attempt. *)
```
jdchristensen, line 57:
```suggestion
(** We make the last two arguments explicit so that end points can easily be specified. We had to do this again, since the section encapsulated the previous attempt. *)
```

jdchristensen (approved):
This is very nice.  Since more typeclass inference is involved with this approach, I was curious whether the compile time for the library is significantly affected.

theories/Algebra/ooGroup.v
mikeshulman: Do you mean it doesn't work to just `rewrite (pointed_htpy ...)`?

theories/Homotopy/ExactSequence.v
```diff
@@ -302,31 +301,33 @@ Definition fiberseq_isexact_purely {F X Y : pType} (i : F ->* X) (f : X ->* Y)
 Definition fiberseq_loops {F X Y} (fs : FiberSeq F X Y)
   : FiberSeq (loops F) (loops X) (loops Y).
 Proof.
-  exists (loops_functor fs.1).
-  refine (_ o*E pequiv_loops_functor fs.2).
-  exact ((pfiber_loops_functor fs.1)^-1* ).
+  (** TODO: doesn't work?! *)
+(*   exists (fmap loops fs.1). *)
```
mikeshulman, line 305: That's weird, what's the error?

theories/Pointed/Core.v
mikeshulman: Great!  Maybe remove the `TODO: finish` above?

mikeshulman (approved):
This is great!  It's really nice to see the wildcat machinery starting to pay off.

theories/Algebra/ooGroup.v
Alizter: Yes

theories/Truncations/Core.v
Alizter: I didn't originally want to touch `Trunc_functor` too much since that is used in a lot of places. I ended up needing a wildcat instance for `Tr` however, so I left it partway converted. I am not sure I want to touch it right now.

theories/Homotopy/ExactSequence.v
Alizter: I'm not sure what exactly is going on, but my guess `exists` isn't doing typeclass search before unifying and therefore getting stuck. I think there is a similar issue with the `rewrite` before.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -196,26 +184,22 @@ Proof.
...
   {X Y : pType} (e : X <~>* Y)
-  : Pi n.+1 X $<~> Pi n.+1 Y.
-Proof.
-  snrapply Build_GroupIsomorphism.
-  1: apply (pi_functor n.+1 e).
-  nrefine (Trunc_functor_isequiv _ _).
-  refine (isequiv_homotopic _ (pequiv_iterated_loops_functor_is_iterated_loops_functor n.+1 e)).
-Defined.
+  : Pi n.+1 X $<~> Pi n.+1 Y
+  := emap (Pi n.+1) e.
```
jdchristensen, line 202: This looks like it is three lines shorter, but it is better than that, since it also removes the need for `pequiv_iterated_loops_functor_is_iterated_loops_functor`.

theories/Algebra/ooGroup.v
jarlg: You could do `exact (ap (loops_group X) (fmap_id loops _ _)^).` instead of the pose and rewrite here.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -243,24 +227,10 @@ Proof.
   1,2: reflexivity.
 Defined.
 
-(** WildCat instances for [Pi] *)
-Global Instance is0functor_pi (n : nat) : Is0Functor (Pi n.+1).
-Proof.
-  constructor. intros X Y f. exact (pi_functor (n.+1) f).
-Defined.
-
-Global Instance is1functor_pi (n : nat) : Is1Functor (Pi n.+1).
```
jdchristensen, line 252: Is there some reason that this reason isn't needed anymore?  Seems like it would be useful to have...

theories/Homotopy/HomotopyGroup.v
```diff
@@ -120,13 +133,13 @@ Definition pi_functor_type_pmap {n X Y}
 Coercion pi_functor_type_pmap : pi_functor_type >-> pForall.
 
 (** For the same reason as for [Pi1] we first define [pi1_functor]. *)
```
jdchristensen, line 135:
```suggestion
(** For the same reason as for [Pi1] we make [Pi1] a functor before making [Pi] a functor. *)
```

theories/Homotopy/HomotopyGroup.v
jdchristensen: Ah, but now it's super easy!  I'll make a commitable suggestion in the right spot.

theories/Homotopy/HomotopyGroup.v
```diff
@@ -142,49 +155,24 @@ Proof.
...
-  - apply O_functor_homotopy; exact p.
-  - apply O_functor_homotopy. refine (iterated_loops_2functor (n.+1) _); exact p.
+  apply Build_Is1Functor.
+  - intros a b f g p.
+    exact (fmap2 (Tr 0 o loops) p).
+  - intros X.
+    exact (fmap_id (Tr 0 o loops) X).
+  - intros a b c f g.
+    exact (fmap_comp (Tr 0 o loops) f g).
 Defined.
```
jdchristensen, line 177:
```suggestion

Global Instance is1functor_pi (n : nat) : Is1Functor (Pi n)
  := ltac:(destruct n; exact _).

```

theories/Homotopy/HomotopyGroup.v
```diff
@@ -142,49 +155,24 @@ Proof.
...
-Definition pi_2functor (n : nat)
-  {X Y : pType} (f g : X ->* Y) (p : f ==* g)
-  : pi_functor n f == pi_functor n g.
+Global Instance is1functor_pi1 : Is1Functor Pi1.
 Proof.
-  destruct n; intros x.
-  - apply O_functor_homotopy; exact p.
-  - apply O_functor_homotopy. refine (iterated_loops_2functor (n.+1) _); exact p.
+  apply Build_Is1Functor.
+  - intros a b f g p.
```
jdchristensen, line 170:
```suggestion
  - intros X Y f g p.
```

theories/Homotopy/HomotopyGroup.v
```diff
@@ -142,49 +155,24 @@ Proof.
...
+    exact (fmap2 (Tr 0 o loops) p).
+  - intros X.
+    exact (fmap_id (Tr 0 o loops) X).
+  - intros a b c f g.
```
jdchristensen, line 174:
```suggestion
  - intros X Y Z f g.
```

theories/Homotopy/HomotopyGroup.v
```diff
@@ -142,49 +155,24 @@ Proof.
...
-
-Definition pi_2functor (n : nat)
-  {X Y : pType} (f g : X ->* Y) (p : f ==* g)
-  : pi_functor n f == pi_functor n g.
+Global Instance is1functor_pi1 : Is1Functor Pi1.
 Proof.
-  destruct n; intros x.
-  - apply O_functor_homotopy; exact p.
-  - apply O_functor_homotopy. refine (iterated_loops_2functor (n.+1) _); exact p.
+  apply Build_Is1Functor.
```
jdchristensen, line 169:
```suggestion
  (** The conditions for [Pi1] to be a 1-functor only involve equalities of maps between groups, which reduce to equalities of maps between types.  Type inference shows that [Tr 0 o loops] is a 1-functor, and so it follows that [Pi1] is a 1-functor. *)
  apply Build_Is1Functor.
```
One can even write:
```coq
  assert (is1f : Is1Functor (Tr 0 o loops)) by exact _.
  apply Build_Is1Functor; intros; apply is1f; assumption.
```
but the `apply` step is slow, so the current proof is better.

Alizter:
Regarding the compile time, the compile step takes around 3min, where as the latest master took around 3:30min so it looks to be faster! It would of course be good to compare on somebodies machine to be sure.

theories/Algebra/ooGroup.v
```diff
@@ -129,50 +129,37 @@ Definition group_loops_functor_compose
...
 Definition group_loops_functor_idmap {X : pType}
 : grouphom_idmap (group_loops X)
-  == group_loops_functor pmap_idmap.
+  == group_loops_functor (Id (A:=pType) _).
 Proof.
   intros g.
   unfold grouphom_fun, grouphom_idmap.
   assert (p := eisretr (loops_group X) g).
+  refine (fmap_id loops _ g @ _).
   rewrite <- p.
```
jdchristensen, line 155:
```suggestion
  refine (fmap_id loops _ g @ _).
  rewrite <- (eisretr (loops_group X) g).
```

theories/Algebra/ooGroup.v
jdchristensen: Jarl beat me to it...
```suggestion
  exact (ap (loops_group X) (fmap_id loops _ _)^).
```

theories/Algebra/ooGroup.v
```diff
@@ -129,50 +129,37 @@ Definition group_loops_functor_compose
...
   rewrite !loops_functor_group.
-  rewrite (pointed_htpy (loops_functor_idmap _)).
-  simpl. apply ap.
-  rewrite concat_p1, concat_1p, ap_idmap.
+  pose (q := pointed_htpy
+    (fmap_id loops X : _ ==* _) ((loops_group X)^-1 g)).
+  rewrite q.
   reflexivity.
```
jdchristensen, line 160: Need to also get rid of the reflexivity.
```suggestion
```

jdchristensen:
In my test, the compile time went from 3m50.77s to 3m53.47s, so it is 1.12% slower.  Not bad.  The file that slowed down the most was Homotopy/WhiteheadsPrinciple.vo, which got +0m00.70s.

theories/Homotopy/WhiteheadsPrinciple.v
```diff
@@ -92,21 +92,21 @@ Proof.
...
     exact (concat_1p _ @ concat_p1 _).
   - intros p k; revert p.
     assert (h3 : forall (y:A) (q:x=y),
-               IsEquiv (pi_functor k.+1 (pmap_from_point (@ap _ _ f x y) q))).
+               IsEquiv (fmap (Pi k.+1) (pmap_from_point (@ap _ _ f x y) q))).
     2:exact (h3 x).
     intros y q. destruct q.
-    nrefine (isequiv_homotopic (pi_functor k.+1 (loops_functor (pmap_from_point f x))) _).
-    2:{ apply pi_2functor; srefine (Build_pHomotopy _ _).
+    refine (isequiv_homotopic (fmap (Pi k.+1) (fmap loops (pmap_from_point f x))) _).
```
jdchristensen, line 104: Changing `nrefine` to `refine` here caused the line to go from instantaneous to ~1s.  The following fixes it.  The same thing might also fix some places where Coq is unable to figure out the Is0Functor instances.
```suggestion
    nrefine (isequiv_homotopic (fmap (Pi k.+1) (Is0Functor:=is0functor_pi _) (fmap loops (Is0Functor:=is0functor_loops) (pmap_from_point f x))) _).
```

If this happens a lot, we could make an `fmap'` that allows us to write `fmap' is0functor_loops ...` instead of `fmap loops (Is0Functor:=is0functor_loops) ...`.

Alizter: The real problem here is that we have no control over which typeclass searches we are doing. We can either do all of them or none of them. Since the IsEquiv search is causing the slow down, I've opted to split this into two lines.

theories/Homotopy/ExactSequence.v
jdchristensen: This is probably a place where passing the `Is0Functor` argument to `fmap` would fix it.  But using `refine` rather than `exists` is better.

theories/Homotopy/WhiteheadsPrinciple.v
jdchristensen: I like the way you handled this.  Resolving this.

Alizter:
Let me clean up the git history a bit.

theories/Homotopy/ExactSequence.v
Alizter: So `exists` is a variant of `split` which is turn a variant of `constructor`. `constructor` like `apply` isn't very good at doing typeclass inference. There is even a warning in the ref manual https://coq.inria.fr/refman/proof-engine/tactics.html?highlight=exists#coq:tacn.constructor

PR #1562, opened 2021-07-06
New CI

Alizter:
Here is a project that I have been working on for a few months now. I have completely rewritten the CI.

The CI is now stratified into 4 stages:
 1. **Build Stage** - Here we build the library in 3 different ways:
    - `opam-build`: Build the library using the `hott.opam` file.
    - `build`: The main build job generating timing data.
    - `quick-build`: A faster build than the others, allowing you to see compilation errors quickly.

Each of these jobs runs on our current supported version of coq, the latest stable version and the dev version. The dev versions of the builds have the possibility of breaking due to the live development of coq. The stable version of coq is slightly behind, allowing us to distinguish between coq errors and the library breaking on the latest version of coq.

We don't actually build and install coq but run everything in the coq-docker containers. This makes things much faster.

The compiled files from the `build` job are uploaded as artifacts for use in the documentation and validation stage.

 2. **Documentation and Validation Stage** - Here we carry out the following jobs:
     - `doc-alectryon`: Build the alectryon html documentation and upload the artifacts.
     - `doc-dep-graph`: Build the dependency graphs and upload the artifacts.
     - `doc-coqdoc`: Build the coqdoc html documentation and upload the artifacts.
     - `doc-proviola`: Build the proviola html documentation and upload the artifacts.
     - `doc-timing`: Build the timing html documentation and upload the artifacts.
     - `coqchk`: Validate the compiled coq files. This step wasn't actually being carried out in the previous CI and has some interesting information. It lists all the axioms used in the library.
     - `install`: Test the installation process

The `doc` jobs will only build on the supported version of coq wheras the `coqchk` and `install` jobs use artifacts from all 3 versions.

  3. **Deployment Stage** - Here we deploy the html documentation and dependency graphs. This is done as a single-commit to the branch gh-pages, squashing the history and keeping the size of the repo down as a result. 

 4. **Clean up Stage** - Here we delete the workspace and documentation artifacts generated.

The CI should also be easier to maintain. For example, the current supported coq version is at the top of `ci.yml` and can be changed easily. The jobs are more modular than the previous bash scripts which means we can add and remove parts as we please in the future. 

---

There are a few things left we can do that I haven't included in this PR.
 * Since we don't rely on the coq submodule any more we can finally remove it. #1456
 * Many of the scripts in `etc/ci` are no longer being used, and may be worth removing.

Closes #1516

PR #1563, opened 2021-07-16
implement colimit as coequalizer

Alizter:
This gets rid of the axiomitized HIT definition of `Colimit` and replaces it with an implementation using coequalizers. Generally related to #1067. There is a slight performance hit in `Colimit_Flattening.v` but this is because the proof there isn't so great. Ideally, one would write a proof with not so many reduction tactics, but this will require some effort. I've done some minimal changes to the proof to get it to go through.

### DIFF
--- a/theories/Colimits/Colimit.v
+++ b/theories/Colimits/Colimit.v
@@ -1,9 +1,9 @@
-Require Import Basics.
-Require Import Types.
+Require Import Basics Types.
 Require Import Diagrams.Diagram.
 Require Import Diagrams.Graph.
 Require Import Diagrams.Cocone.
 Require Import Diagrams.ConstantDiagram.
+Require Import HIT.Coeq.
 
 Local Open Scope path_scope.
 Generalizable All Variables.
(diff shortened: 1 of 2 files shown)


theories/Colimits/Colimit.v
```diff
@@ -35,52 +35,83 @@ Definition cocone_postcompose_inv `{D: Diagram G} {Q X}
...
+*)
 
-  Arguments colim {G D} i x.
+(** A colimit is just the coequializer of the source and target maps of the diagram *)
+Definition Colimit {G : Graph} (D : Diagram G) : Type :=
+  @Coeq
+    {x : sig D & {y : sig D & {f : G x.1 y.1 & D _f f x.2 = y.2}}}
+    (sig D)
+    (fun t => t.1)
+    (fun t => t.2.1)
```
jdchristensen, line 53: The first arg to `Coeq` could instead be `{x : sig D & {j : G & G x.1 j}}`, with the last argument adjusted to compensate.  (This is just getting rid of a contractible part of the first argument.)  Would this make anything else simpler?

jdchristensen:
Does this make the file Colimit_Coequalizer unnecessary?  Or does it at least simplify things there, making some of the equivalences definitional?

LGTM!

theories/Colimits/Colimit.v
Alizter: Ah nice, I wondered the same thing, but didn't see the contractible part. I'll see if it simplifies things.

Alizter:
>Does this make the file Colimit_Coequalizer unnecessary? Or does it at least simplify things there, making some of the equivalences definitional?

I also wondered the same thing but haven't investigated it. I'll take a look.

@jdchristensen you might also be interested in https://github.com/HoTT/HoTT/pull/1562#issuecomment-875056706, there I now run the validate jobs where coq checks the validity of our compiled files. It gives us a list of axioms used in the whole library.

theories/Colimits/Colimit.v
Alizter: OK, that didn't simplify much, but I implemented it anyway. Once you get the induction principle for Colimit, things that depend on that don't really care about the implementation.

Alizter:
Regarding `Colimit_Coequalizer.v`. Having a look at the types, it still seems that they are different. In fact, I don't think that the equivalence comes from the functoriality of Coeq. The target type in the coequalizers for example are inequivalent. In Coeq this is just `A` whereas the coequalizer via colimit this becomes `{X : Bool & if X then B else A}`. That type is just a coproduct `A + B` so we can't really expect `inl` to be an equivalence here.

theories/Colimits/Colimit.v
jdchristensen: @Alizter Did you forget to push this change?  (It's not that important, but you said that you implemented it anyway.)

Alizter: Whoops, I didn't push.

Alizter: I made #1565

PR #1566, opened 2021-08-06
Simplify various proofs involving iterated truncations

jdchristensen:
Several places that deal with iterated truncations can be simplified.  In addition, the equivalence defined by `pequiv_loops_em_em` in EMSpace.v is now more direct, using one edge of a square instead of the composite of the other three.

### DIFF
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -51,7 +51,7 @@ Proof.
 Defined.
 
 Definition nat_to_trunc_index (n : nat) : trunc_index
-  := trunc_index_inc minus_two.+2 n.
+  := (trunc_index_inc minus_two n).+2.
 
 Coercion nat_to_trunc_index : nat >-> trunc_index.
 
--- a/theories/Homotopy/EMSpace.v
+++ b/theories/Homotopy/EMSpace.v
@@ -6,6 +6,7 @@ Require Import Homotopy.Suspension.
 Require Import Homotopy.ClassifyingSpace.
 Require Import Homotopy.HSpace.
 Require Import TruncType.
+Require Import WildCat.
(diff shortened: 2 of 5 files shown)


theories/Basics/Trunc.v
```diff
@@ -51,7 +51,7 @@ Proof.
 Defined.
 
 Definition nat_to_trunc_index (n : nat) : trunc_index
-  := trunc_index_inc minus_two.+2 n.
+  := (trunc_index_inc minus_two n).+2.
```
jdchristensen, line 54: This change makes functions (such as `trunc_index_add`) compute better, meaning that less rewriting is required.

theories/Homotopy/EMSpace.v
```diff
@@ -218,27 +219,16 @@ Section EilenbergMacLane.
...
-    srapply Build_pEquiv'.
-    { notypeclasses refine (Build_Equiv _ _ (Trunc_functor _ tr) _).
-      notypeclasses refine (isequiv_conn_map_ino n.+2 _).
-      1,2: exact _.
-      apply conn_map_O_functor.
-      intro x.
-      notypeclasses refine (isconnected_pred_add n.-2 _ _).
-      rewrite 2 trunc_index_add_succ.
-      (* This has to be done by induction since n.-2.+2 only cancels when n >= 0 i.e. a nat *)
-      assert (p : (n .-2 +2+ n).+2 = (n +2+ n)).
```
jdchristensen, line 234: With the change above, this equality becomes definitional.

Alizter (approved):
LGTM I'm happy with the simplifications that you were able to make!

mikeshulman:
Nice!

PR #1568, opened 2021-08-13
A few minor things about Eilenberg-Mac Lane spaces

jdchristensen:
Adds a few small results about EM spaces, and cleans up a few proofs.

### DIFF
--- a/theories/Pointed/Loops.v
+++ b/theories/Pointed/Loops.v
@@ -365,45 +365,27 @@ Proof.
   apply tr_loops.
 Defined.
 
-(* This is slightly different to 7.2.9 in that we ommit n = -1, which is
-   inhabited hsets are contractible. *)
+(* 7.2.9, with [n] here meaning the same as [n-1] there. Note that [n.-1] in the statement is short for [trunc_index_pred (nat_to_trunc_index n)] which is definitionally equal to [(trunc_index_inc minus_two n).+1]. *)
 Theorem equiv_istrunc_contr_iterated_loops `{Univalence} (n : nat)
-  : forall A, IsTrunc n A <~> forall a : A,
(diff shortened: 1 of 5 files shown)


theories/Pointed/Loops.v
```diff
@@ -365,45 +365,27 @@ Proof.
   apply tr_loops.
 Defined.
 
-(* This is slightly different to 7.2.9 in that we ommit n = -1, which is
-   inhabited hsets are contractible. *)
+(* 7.2.9, with [n] here meaning the same as [n-1] there. Note that [n.-1] in the statement is short for [trunc_index_pred (nat_to_trunc_index n)] which is definitionally equal to [(trunc_index_inc minus_two n).+1]. *)
 Theorem equiv_istrunc_contr_iterated_loops `{Univalence} (n : nat)
-  : forall A, IsTrunc n A <~> forall a : A,
-    Contr (iterated_loops n.+1 (A, a)).
+  : forall A, IsTrunc n.-1 A <~> forall a : A, Contr (iterated_loops n (A, a)).
```
jdchristensen, line 370: By starting this at -1 instead of 0, the base case is just `equiv_hprop_inhabited_contr`.  Also, `n` is definitionally the successor of `n.-1`, because of the recent change to `nat_to_trunc_index` that makes it compute better.  The proof below also contains some simplifications that are separate from these changes, making it shorter and avoiding rewrites.

mikeshulman (approved):
Very nice!

jdchristensen:
I was hoping to prove that every pointed, (n-1)-connected n-type is equivalent to a K(G,n), but didn't manage.  I can define the map and just need that a certain triangle commutes to prove that it is an equivalence...

PR #1571, opened 2021-08-22
[WIP] Adjunctions, limits, colimits in wildcats + possible progress towards 3x3 lemma for pushouts 

Alizter:
Here is a sneak-peak at something I have been working on for quite some time. This PR is currently huge but this is for viewing purposes. I will definitely split it up.

We define adjunctions in a wild category. For the moment, they exist between `Is0Functor`s and they are defined as a natural equivalence `Hom(Fa,b)=Hom(a,Gb)`. I have a few reasons for using this definition, the main one being the coherence needed in the functors F and G is very low, compared with say the unit-counit adjunction. We develop some basic theory of adjuncts, but this part is still incomplete. The main result I would like to finish is `adjunction_postcomp` which says that postcomposition preserves adjunctions.

Now moving onto limits and colimits. We define them as right and left adjoints to the diagonal functor respectively. This definition has the advantage of making the proofs of the following relatively easy: limits commute with right adjoints and colimits commute with left adjoints. There is a question of "Why not define colimits as op limits?", this is quite tricky to work with, and it is easier to duplicate the work for the time being. Something clever needs to be done to set that up, which I am not able to do for now. As a corollary, we get limits commute with limits and colimits commute with colimits.

Note `adjunction_swap_fun01` is left unfinished, this says that the functor `swap_fun01 A B C` is adjoint to `swap_fun01 B A C` where `swap_fun01 A B C : Fun01 A (Fun01 B C) -> Fun01 B (Fun01 A C)`. I just didn't have time to finish that yet, but I don't see any problems.

Next, the work needs to be put in to show that `Type` has colimits of the shape `Span` which is just a span diagram. The existence of this colimit is given by `Pushout`. What I am hoping is that when some of the admitted lemmas are finished, the result about colimits commuting with colimits will definitionally collapse to the 3x3 lemma for pushouts.

Interested parties may be: @mikeshulman @jdchristensen @jarlg

### DIFF
--- a/theories/Categories.v
+++ b/theories/Categories.v
@@ -3,114 +3,114 @@
 
 (** First we give modules to all of the kinds of category theory constructions (corresponding to directories), so that we can refer to them as [Category.foo] or [Functor.foo] after [Require Import Categories.] *)
 (** ** Categories *)
-Require Category.
+Require HoTT.Categories.Category.
 (** ** Functors *)
-Require Functor.
+Require HoTT.Categories.Functor.
 (** ** Natural Transformations *)
-Require NaturalTransformation.
+Require HoTT.Categories.NaturalTransformation.
 (** ** Functor Categories *)
-Require FunctorCategory.
+Require HoTT.Categories.FunctorCategory.
 (** ** Groupoids *)
-Require GroupoidCategory.
+Require HoTT.Categories.GroupoidCategory.
 (** ** Precategory of Groupoids *)
-Require CategoryOfGroupoids.
+Require HoTT.Categories.CategoryOfGroupoids.
 (** ** Discrete Categories *)
-Require DiscreteCategory.
+Require HoTT.Categories.DiscreteCategory.
 (** ** Indiscrete Categories *)
-Require IndiscreteCategory.
+Require HoTT.Categories.IndiscreteCategory.
 (** ** Finite Discrete Categories (natural numbers as categories) *)
-Require NatCategory.
+Require HoTT.Categories.NatCategory.
 (** ** Chain Categories [[n]] *)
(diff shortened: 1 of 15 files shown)


jdchristensen:
This sounds great!  I won't get a chance to provide feedback, but I think these things should be quite useful.  Here's one example that I was thinking about recently:  by `equiv_bg_pi1_adjoint` we know that `Pi 1` is left adjoint to `B` on pointed, connected types.  It follows that `Pi 1` preserves colimits in the category of pointed, connected types.  A special case of this is the van Kampen theorem.  This seems much easier than other proofs of van Kampen (in HoTT, or not).  I think it would be fairly straightforward to do this without any WildCat machinery, but with the machinery it should be even easier.

jarlg (commented):
Nice work! I didn't run or go through any of the proofs, but I commented on the two things that caught my eye. It'd be very cool if "colimits commute" indeed gets us the 3x3 lemma! Since you've already proven the former, is there anything blocking the latter? (Did you try doing it?)

Alizter:
So I have run out of free time to work on this. The 3x3 lemma goes through modulo some morphism extensionality. I'm not super certain these are actually provable in their current state, but hopefully this serves as an indicator that this argument will probably work.

Alizter:
OK, I will factor this up in the coming weeks so that we at least have adjunctions and (co)limits in the library. This should allow for some development of other topics such as abelian categories too.

I am sure that the 3x3 lemma will go through, but it will require me to wrestle with morphism extensionality. I am lacking the time and energy to currently do this however.

Alizter:
I really wish that I could have gotten this to work, but: C'est la vie.

PR #1573, opened 2021-09-13
Adapt w.r.t. coq/coq#14679.

ppedrot:
Supersedes https://github.com/HoTT/HoTT/pull/1572. It is probably backwards compatible but I haven't checked.


JasonGross:
I still think it might be better to make the beta-iota normalization hint a global/export one in Overture.v, ideally with a comment containing an example of a how applying a typeclass hint can lead to a non-normal goal.

ppedrot:
@JasonGross but this breaks some unrelated stuff, see my comment on the other PR. HoTT relies both on terms not being normalized and unfold sometimes normalizing :shrug:.

ppedrot:
@JasonGross are you fine with merging this as-is then?

Alizter:
LGTM so merging

PR #1576, opened 2021-09-22
Small optimization for typeclass search.

Alizter:
Later there will be more typeclass instances floating around which makes these areas slower, therefore we optimize it slightly.

### DIFF
--- a/theories/Pointed/Loops.v
+++ b/theories/Pointed/Loops.v
@@ -123,7 +123,7 @@ Global Instance is0functor_iterated_loops n : Is0Functor (iterated_loops n).
 Proof.
   induction n.
   1: exact _.
-  rapply is0functor_compose.
+  nrapply is0functor_compose; exact _.
 Defined.
 
 Global Instance is1functor_iterated_loops n : Is1Functor (iterated_loops n).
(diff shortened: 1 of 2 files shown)


theories/Pointed/Loops.v
```diff
@@ -123,7 +123,7 @@ Global Instance is0functor_iterated_loops n : Is0Functor (iterated_loops n).
 Proof.
   induction n.
   1: exact _.
-  rapply is0functor_compose.
+  nrapply is0functor_compose; exact _.
```
jdchristensen, line 126: Shouldn't this be essentially the same as the original code?

Alizter: I think the difference is that rapply does typeclass search on each hole, starting from the first, but the latter will generate a goal and then do typeclass search on that. The initial holes are about categories usually, and when there are lot of instances floating around it wastes time.

Alizter: I should note that you won't notice anything at the moment, but my development of #1571 shows this to be useful.

jdchristensen: Ok, LGTM.

PR #1577, opened 2021-09-22
sum functor preserves homotopies 

Alizter:


theories/Types/Sum.v
```diff
@@ -403,6 +403,16 @@ Section FunctorSum.
 
 End FunctorSum.
 
+Definition functor_sum_homotopic {A A' B B' : Type}
+  {f f' : A -> A'} {g g' : B -> B'}
+  : f == f' -> g == g' -> functor_sum f g == functor_sum f' g'.
```
jdchristensen, line 408: Maybe it's better to have `(p : f == f') (q : g == g')` before the colon?  I think it's a bit easier to read, and it means that the caller can use named arguments.

Alizter: I closed this PR in favour of #1578

PR #1579, opened 2021-09-22
argument names for is1functor_compose

Alizter:
This let's us specify them manually without using the full `@` syntax.


jdchristensen:
LGTM.

PR #1582, opened 2021-09-23
misc. additions for yoneda lemma

Alizter:
This depends on #1580

### DIFF
--- a/theories/WildCat/Yoneda.v
+++ b/theories/WildCat/Yoneda.v
@@ -41,6 +41,10 @@ Proof.
     refine (cat_assoc_opp _ _ _).
 Defined.
 
+Definition fun01_hom {A} `{Is01Cat A}
+  : Fun01 (A^op * A) Type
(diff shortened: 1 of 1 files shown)


theories/WildCat/NatTrans.v
```diff
@@ -5,9 +5,13 @@ Require Import WildCat.Square.
 
 (** ** Natural transformations *)
 
-Definition Transformation {A B : Type} `{IsGraph B} (F : A -> B) (G : A -> B)
+Definition Transformation {A : Type} {B : A -> Type} `{forall x, IsGraph (B x)}
+  (F G : forall (x : A), B x)
   := forall (a : A), F a $-> G a.
 
+(** This let's us apply transformations to things. *)
```
jdchristensen, line 12:
```suggestion
(** This lets us apply transformations to things. *)
```
We also discussed adding a comment here about the Identity keyword.

Alizter: This is the same commit as before, hence why I marked this as draft. I will rebase when that gets merged.

PR #1583, opened 2021-09-23
Product functors, swap functor

Alizter:


mikeshulman:
LGTM

theories/WildCat/Prod.v
```diff
@@ -116,3 +117,71 @@ Definition emap11 {A B C : Type} `{HasEquivs A} `{HasEquivs B} `{HasEquivs C}
...
+  : Is0Functor (@snd A B).
+Proof.
+  apply Build_Is0Functor.
+  intros ? ? f; exact (snd f).
+Defined.
+
+(** Swap functor *)
+
+Definition prod_swap {A B : Type} : A * B -> B * A
+  := fun '(a , b) => (b , a).
```
jdchristensen, line 162: I wonder if this should somehow be unified with `equiv_prod_symm`?  (Or if we should at least standardize on `swap` vs `symm` for the naming?)  Also, it probably makes sense for the parts of this that don't involving WildCats to be in Types/Prod.v.

Alizter:
I'm taking the LGTM as confirmation and merging this.

PR #1585, opened 2021-10-14
Update to Coq V8.14

Alizter:
I will release a tag when this is merged.


SkySkimmer:
Something like `dune build coq-core.install && make -f Makefile.dune theories-prelude` should be close to what `coqocaml` did.

Alizter:
This is OK to merge, the CI error is just because the "latest" coq-docker image is still stuck on 8.13.2 so opam is complaining that there is a version mismatch. The rest of the CI is fine tho.

PR #1591, opened 2021-10-25
Limits in wildcats

Alizter:
We define (co)limits in a wild cat and prove that (co)limits commute with (co)limits. Unfortunately, there are a lot of morphism extensionality assumptions dotted throughout and I am not sure how to avoid them.

Some of these assumptions may not actually be attainable. Take for instance `HasMorExt (Fun01 A B)`. Since the 2-cells are incoherent, will this ever be provable?

### DIFF
--- a/theories/WildCat/Limits.v
+++ b/theories/WildCat/Limits.v
@@ -1,0 +1,464 @@
+Require Import Basics Types.
+Require Import WildCat.Core.
+Require Import WildCat.NatTrans.
+Require Import WildCat.FunctorCat.
+Require Import WildCat.Adjoint.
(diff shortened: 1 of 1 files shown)


Alizter:
I was sorting through some of my old notes and I found the motivation for this. I wanted to have a more mature category theory library for doing some more modern algebraic topology a la topos theory. For example, I have written that I wanted to prove the Eckmann-Hilton splitting for which I needed a notion of descent for which I needed some general facts about limits/colimits in a wildcat. I would like to come back to these ideas eventually, but I don't have the time or the motivation to finish it at the moment. Leaving this here to remind my future self.

PR #1595, opened 2021-11-15
Adapt w.r.t. coq/coq#15198.

ppedrot:
Should be backwards compatible.

### DIFF
--- a/theories/PathAny.v
+++ b/theories/PathAny.v
@@ -64,14 +64,14 @@ Proof.
 Defined.
 
 (** This tactic just applies the previous lemma, using a match to figure out the appropriate type families so the user doesn't have to specify them. *)
-Ltac contr_sigsig a c :=
+Ltac contr_sigsig a c0 :=
   match goal with
   | [ |- Contr (@sig (@sig ?A ?B) (fun ab => @sig (@?C ab) (@?D ab))) ] =>
     (* The lemma only applies when C depends only on the first component of ab, so we need to factor it somehow through pr1. *)
     let C' := fresh in
     transparent assert (C' : {C' : A -> Type & forall ab, C' ab.1 = C ab});
     [ eexists; intros ab; reflexivity
-    | nrefine (contr_sigma_sigma A B C'.1 (fun a b c => D (a;b) c) a c);
+    | nrefine (contr_sigma_sigma A B C'.1 (fun a b c => D (a;b) c) a c0);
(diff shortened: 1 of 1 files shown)


mikeshulman:
What's going on here exactly?

ppedrot:
The `uconstr` interpretation in master is faulty, because it does not computes the closure of the term, leading to unbound variable errors when nesting `ltac` quotations inside `uconstr` quotations. coq/coq#15198 fixes this, but it unraveled the fact that HoTT is silently exploiting the bug.

Namely, in the offending tactic the Ltac variable `c` is bound to an argument of the (dynamic) type of a Gallina term (a `constr`, in the implementation jargon). There is an `uconstr` quotation (implicitly introduced by refine) which contains a subterm of the form `fun c => ...`. The (questionable, but I am not here to rewind the past) semantics of Ltac is to substitute bound Gallina variables inside quotations by the value of the corresponding variable in the Ltac closure whenever this maps to an identifier. But `c` is mapped to a `constr`, not an identifier, so it will fail at runtime.

The patch simply prevents this faulty shadowing by giving a different name to the Ltac variable.

ppedrot:
Following a suggestion by @Alizter I've tweaked the PR to remove the offending bound Gallina variable instead of renaming the Ltac one. If you ask me, caring for the precise name of bound Ltac variables is a bit like being a firm believer of the Hungarian notation, but I am not a HoTT dev, so...

mikeshulman:
Wait a minute, you're telling me that the *bound* variable named `c` has an effect *outside its scope*?

ppedrot:
@mikeshulman It's all a matter of passing an identifier as the given name of a bound Gallina variable inside a quotation. In Ltac, if you write `let n := ident:(foo) in refine (fun n => ...)` you'll generate a term `fun foo => ...`. Don't ask, don't tell.

mikeshulman:
So the old HoTT code, which *looks* correct, is technically "exploiting a bug" because Ltac used to be incorrectly implementing its intended but stupid behavior?  And "fixing that bug" causes our correct-looking code to no longer work?

ppedrot:
> Ltac used to be incorrectly implementing its intended but stupid behavior? And "fixing that bug" causes our correct-looking code to no longer work?

Welcome to the Ltac world, mister Anderson. And consider yourself lucky because here at least we know the intended semantics, so we can claim with certainty it was a bug exploit. Most of the time it's just "meh, it's broken, let's fix the relevant scripts and forget about it".

PR #1596, opened 2021-11-16
Adapt w.r.t. coq/coq#15171.

ppedrot:
HoTT was exploiting a weird multigoal behaviour of the exact tactic that was interleaving global evar resolution with goal filling. It is not clear that this can be qualified as a bug, but the net result is quite surprising and it's better for HoTT not to depend on it, especially if we merge coq/coq#15171.

This should be backwards compatible.

### DIFF
--- a/theories/Algebra/Rings/Z.v
+++ b/theories/Algebra/Rings/Z.v
@@ -158,7 +158,7 @@ Lemma cring_catamorphism_fun_negate {R} x
   : cring_catamorphism_fun R (- x) = - cring_catamorphism_fun R x.
 Proof.
   snrapply (groups.preserves_negate _).
-  1-6: exact _.
+  1-6: typeclasses eauto.
   snrapply Build_IsMonoidPreserving.
   1: exact _.
   split.


theories/Algebra/Rings/Z.v
```diff
@@ -158,7 +158,7 @@ Lemma cring_catamorphism_fun_negate {R} x
   : cring_catamorphism_fun R (- x) = - cring_catamorphism_fun R x.
 Proof.
   snrapply (groups.preserves_negate _).
-  1-6: exact _.
+  1-6: typeclasses eauto.
```
SkySkimmer, line 161: I would have done `3,6: exact _.` as the point is that we don't need TC to fill the other goals.

ppedrot: Yeah, I actually started writing this and then thought that the current variant was more similar than the previous one. Not sure which one is better, but they're both correct.

SkySkimmer: When the original is arguably incorrect similarity is not a positive.

ppedrot: Well, typeclass eauto is specifically designed to handle correctly multigoal proofs, so there is that.

PR #1598, opened 2021-12-15
Create dependabot.yml

JasonGross:
This will tell dependabot to create PRs to update our submodules and github actions versions

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,17 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


Alizter:
So this will create a PR for each submodule? Should we do https://github.com/HoTT/HoTT/pull/1590 first?

JasonGross:
```
npm does not support Node.js v10.24.0
You should probably upgrade to a newer version of node as we
can't make any promises that npm will work with this version.
You can find the latest version at https://nodejs.org/
/usr/local/lib/node_modules/npm/lib/npm.js:32
```

Hmmmm

JasonGross:
> So this will create a PR for each submodule? Should we do #1590 first?

Yes, and sure

JasonGross:
Shall we merge this now?  (The ci failure is unrelated)

PR #1605, opened 2021-12-20
Latest version of npm in ci

Alizter:
Let's see if this fixes the ci.


JasonGross:
What about updating node.js rather than npm?

PR #1606, opened 2021-12-20
Workaround for version of node

Alizter:
See https://github.com/nodejs/help/issues/3644

Let's see if this fixes the CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -230,6 +230,11 @@ jobs:
         custom_script: |
           sudo apt-get update
           sudo apt-get install -y ghc cabal-install
+          
+          # See https://github.com/nodejs/help/issues/3644
+          curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash -
+          sudo apt-get install -y nodejs
+          
           sudo apt-get install -y --allow-unauthenticated \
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -230,6 +230,11 @@ jobs:
         custom_script: |
           sudo apt-get update
           sudo apt-get install -y ghc cabal-install
+          
+          # See https://github.com/nodejs/help/issues/3644
+          curl -fsSL https://deb.nodesource.com/setup_current.x | sudo -E bash -
+          sudo apt-get install -y nodejs
+          
           sudo apt-get install -y --allow-unauthenticated \
           libnode-dev node-gyp libssl-dev npm aspcud      \
```
JasonGross, line 239: Do things break if we remove libnode-dev?
```suggestion
          node-gyp libssl-dev npm aspcud      \
```

JasonGross:
If that doesn't work, what about following https://github.com/jeroen/V8/issues/100#issuecomment-708288151 ?

JasonGross:
@Alizter  What's the version that gave the stack overflow rather than the node version error?

Alizter:
@JasonGross I think it is this one: https://github.com/Alizter/HoTT/tree/fix-ci-dpd-graph

Issue #1567, opened 2021-08-06
Remove some things from Set/Ordinals.v from global namespace?


jdchristensen:
For example, `factor1` and `factor2` conflict with the ones defined in Factorization.v.  Maybe they can be made into `Local` definitions?  Probably some other things aren't intended to be global as well.

jdchristensen:
`image` as well.  The issue is that if I `Require Import HoTT` in external work, I now get the versions from Sets/Ordinal.v by default.

Alizter:
I think that the real problem here is that the order things are imported in https://github.com/HoTT/HoTT/blob/master/theories/HoTT.v are wrong. The names we actually want to use should end up last, so folders like `Sets` should be imported earlier on.

We can also add some tests to `Tests.v` that check certain definitions are what we expect them to be. This will catch any name overloads that we don't want in the future.

I'm not sure making things local is the best idea, since someone might want to use the sets `image` more in the future. With my suggestion they can do `Require Import HoTT. Import Sets.LocationOfImage`. This isn't possible when things are made local.

mikeshulman:
I've been thinking for a while that perhaps we really ought to be making more use of modules and namespaces.  Instead of trying to give everything a long globally-unique name and then `Require Import`ing everything, it would probably make more sense in many cases to give things shorter names that make sense in the context of their module (e.g. file) and then `Require` them and refer to them qualified by module name.

jdchristensen:
It probably won't scale to keep everything globally unique all of the time, but it is handy when things are mostly unique, so that you can tell locally what something means without needing to check what was important.  And names written as Module.name can get pretty verbose.

In the case of Ordinals.v, couldn't that file just make use of the general `image` function we already have?  It seems to be redoing something that we already have for a general modality.

jdchristensen:
[...]

PR #1608, opened 2021-12-25
Relation coequalizer

Alizter:
Here is a simpler definition of a coequalizer as a HIT. We then redefine Quotient and Coeq in terms of this. Working with the "untypetheoretic coequalizer" can be a bit awkward sometimes since you might have to pass a sigma type here and there. This mostly removes this requirement.

This is a step towards cleaning up the HIT folder.

### DIFF
--- a/theories/Algebra/AbGroups/Abelianization.v
+++ b/theories/Algebra/AbGroups/Abelianization.v
@@ -1,6 +1,6 @@
 Require Import Basics Types Cubical WildCat.
 Require Import Truncations.
-Require Import HIT.Coeq.
+Require Import Colimits.Coeq.
 Require Import Algebra.Groups.
 Require Import Algebra.AbGroups.AbelianGroup.
 
@@ -97,9 +97,8 @@ Section Abel.
     (x y z : G) : dp_apD (Abel_ind P a c) (ab_comm x y z) = c x y z.
   Proof.
     apply dp_apD_path_transport.
-    refine (apD_compose' tr _ _ @ _).
-    refine (ap _ (Coeq_ind_beta_cglue _ _ _ (x, y, z)) @ _).
-    apply concat_V_pp.
+    refine (apD_compose' tr _ _ @ ap _ _ @ concat_V_pp _ _).
+    rapply Coeq_ind_beta_cglue.
   Defined.
 
   (** We also have a recursion princple. *)
@@ -166,7 +165,7 @@ Section AbelGroup.
       refine (ap _ (associativity _ _ _)). }
     intros b c d.
     revert a.
-    Abel_ind_hprop a; cbn.
+    Abel_ind_hprop a; simpl.
     refine (ap _ (associativity _ _ _) @ _).
     refine (ab_comm _ _ _ @ _).
     refine (ap _ (associativity _ _ _)^).
@@ -178,7 +177,7 @@ Section AbelGroup.
     intros x y.
     Abel_ind_hprop z; revert y.
     Abel_ind_hprop y; revert x.
-    Abel_ind_hprop x; cbn.
+    Abel_ind_hprop x; simpl.
     apply ap, associativity.
   Defined.
 
@@ -192,13 +191,13 @@ Section AbelGroup.
   Global Instance abel_leftidentity : LeftIdentity abel_sgop abel_mon_unit.
   Proof.
     Abel_ind_hprop x.
-    cbn; apply ap, left_identity.
+    simpl; apply ap, left_identity.
   Defined.
 
   Global Instance abel_rightidentity : RightIdentity abel_sgop abel_mon_unit.
   Proof.
     Abel_ind_hprop x.
-    cbn; apply ap, right_identity.
+    simpl; apply ap, right_identity.
   Defined.
 
   (** Hence Abel G is a monoid *)
@@ -211,7 +210,6 @@ Section AbelGroup.
     Abel_ind_hprop y.
     revert x.
     Abel_ind_hprop x.
-    cbn.
     refine ((ap ab (left_identity _))^ @ _).
     refine (_ @ (ap ab (left_identity _))).
     apply ab_comm.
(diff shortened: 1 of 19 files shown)


Alizter:
@mikeshulman ping

mikeshulman:
Sorry, I'm on vacation this week.  d-:

theories/Algebra/AbGroups/Abelianization.v
```diff
@@ -98,8 +98,17 @@ Section Abel.
   Proof.
     apply dp_apD_path_transport.
     refine (apD_compose' tr _ _ @ _).
-    refine (ap _ (Coeq_ind_beta_cglue _ _ _ (x, y, z)) @ _).
-    apply concat_V_pp.
```
mikeshulman, line 102: Why did this proof get more complicated?

mikeshulman:
Are you sure this is better than defining coequalizers to be a particular kind of colimit?

theories/Algebra/AbGroups/Abelianization.v
Alizter: Since Coeq is no longer an inductive type, coq had trouble unifying some terms. I have cleaned this up now and given a shorter proof overall. The trick is to suggest Coeq_ind_beta_cglue as late as possible so that coq can infer various subterms.

Alizter:
We could redefine Coeq to be a particular shaped colimit. But I think this is slightly orthogonal to this PR. The purpose of this PR is really to give a simpler base HIT that we base most other HITs off of. Perhaps in the future I can fix my (co)limits branch of the wildcat library too and give a nice unified approach.

jdchristensen:
Can you explain more the purpose of this PR?  Originally, `Coeq` was defined as a HIT giving the coequalizer of two parallel maps.  Now, `RelCoeq` is defined as a HIT giving the quotient of a type by an equivalence relation.  The two are very similar, so it's not clear why one approach is better than the other.  And if I'm reading the stats correctly, I think this PR adds 60 lines of code.

(Also, I don't like the name `RelCoeq`.  It's a (homotopy) quotient, not a coequalizer, although it's of course also the coequalizer of the projection maps.)

Alizter:
You are correct that the two are very similar. My reason for introducing it is that it is a simpler datatype. Instead of having 3 types and 2 maps, you have 1 type and a family of families over it. There is a general heuristic that "type-theoretic styles lead to simpler proofs".

This is evident (perhaps unsurprisingly) in the definition of quotient, where the new guy avoids the sigma types. The derivation of Coeq from RelCoeq is less chunky than the other way round. I therefore concluded that it makes sense to take that as a base type and derive the others from it. Granted I haven't gone as far as to replace some uses of Coeq with RelCoeq, but it might simplify other proofs.

RelCoeq also has the advantage of making the needed universe number more clear, which is 3. One universe for the type, one for the family and one where the datatype lives. (I am unaware if semantically we know that coequalizers live in the same universe tho @mikeshulman prob has a sketch of it). In any case, this seems to be a good number of universes for Coq to work with. Coeq on the other hand can have 4 universes (even though it is a bit silly). Since RelCoeq is a base HIT, it makes sense to be explicit about it's universes.

Regarding the name, homotopy quotient (`HQuotient`?) sounds fine, I was just a bit worried with confusing it with the truncated version `Quotient`.

jdchristensen:
That explanation makes some sense.  I don't know enough about the various similar constructions that exist in the library to really know what approach is best, but I did notice a few months back that there were redundant things in the library related to quotients, see #1567.  It would be nice to have this organized well.

About the name:  I think `Quotient` makes the most sense, since all of the limits and colimits we define are homotopy limits and colimits.  The truncated one should be `SetQuotient` or something like that, in my opinion.

mikeshulman:
I agree that the adjective "homotopy" isn't necessary, but the word "quotient" has a bit of a connotation of something that is effective in the sense that the thing you quotient by is the "kernel" of the quotient map.  Coequalizer doesn't have that connotation, although I do agree that it's also a bit weird to call something a coequalizer when the given data doesn't look like two parallel maps.

jdchristensen:
Hmm, I don't associate that meaning to "quotient", which to me includes taking the quotient of an algebraic structure by a relation.

mikeshulman:
Well, I said a "bit" of a connotation.  I think the most classical notion of "quotient" is the quotient of an equivalence relation, and some people say "quotient of the equivalence relation generated by" when they want to take the "quotient" of a relation that's not an equivalence relation.  It's certainly true that not everyone does that, though, so maybe we don't need to worry about it.

Alizter:
After thinking about it a bit more, I have come up with an even more convincing argument. As I mentioned before, the datatype for RelCoeq is much simpler than Coeq which also gives us a much easier time characterizing the path type.

As Mike mentioned, RelCoeq A R is A modded out by the reflexive, symmetric, transitive closure of R. Therefore it is easy to see that paths in RelCoeq must consist of signed chains of edges of R. It is not immediately apparent how to characterize Coeq in this way. Basically you can write the following induction principle for the path type:

```coq
Lemma path_relcoeq_ind {A : Type} {R : A -> A -> Type} (a : A)
  (P : forall (b : A), rcoeq a = rcoeq b -> Type)
  (r : P a idpath)
  (e : forall b c (q : rcoeq a = rcoeq b) (s : R b c),
    P b q <~> P c (q @ rcglue s))
  : forall b (q : rcoeq a = rcoeq b), P b q.
```

I'm in the middle of proving this now, but the relevant paper is https://arxiv.org/abs/1901.06022. From this induction principle it should be possible to derive similar ones for Coeq, Pushout etc.

Alizter:
Egbert calls this the "graph quotient" in his thesis btw. Does that seem like a better name? `GraphQuotient`?

@mikeshulman @jdchristensen

mikeshulman:
Hmm, I kind of like that.

jdchristensen:
I like `GraphQuotient` better than `RelCoeq`, but I'd still lean towards `Quotient`.  Still, `GraphQuotient` is fine with me.

Alizter:
Alright, I've renamed it to `GraphQuotient`. I am still hesitant to fully commit to the name `Quotient` but I could be convinced in the future (perhaps we could rename the current `Quotient` `SetQuotient` or something...).

mikeshulman (approved):

PR #1610, opened 2021-12-29
encode-decode lemmas

Alizter:
I had this laying around, probably wouldn't hurt to add to the library.

Technically we could refactor a few proofs in the library to go through this, but I am not sure we would gain much.

### DIFF
--- a/theories/Homotopy/EncodeDecode.v
+++ b/theories/Homotopy/EncodeDecode.v
@@ -1,0 +1,66 @@
+Require Import Basics Types Pointed.
+Require Import Truncations.
+
+(** ** Encode-decode method of characterizing identity types *)
+
+(** See PathAny.v for a related characterization of identity types. *)
+
+Definition encode_decode
+  (A : Type) (a0 : A) (code : A -> Type)
+  (c0 : code a0) (decode : forall x, code x -> a0 = x)
+  (s : forall (c : code a0), decode _ c # c0 = c)
+  (r : decode _ c0 = idpath) (a1 : A)
+  : a0 = a1 <~> code a1.
+Proof.
+  srapply equiv_adjointify.
+  - exact (fun p => p # c0).
+  - apply decode.
+  - intro p.
+    destruct (decode _ p) in p.
+    apply s.
+  - intros [].
+    exact r.
+Defined.
+
+(** Encode-decode for truncated identity-types *)
+Definition encode_decode_trunc n
+  (A : Type) (a0 : A) (code : A -> Type) `{forall a, IsTrunc n (code a)}
+  (c0 : code a0) (decode : forall x, code x -> Tr n (a0 = x))
+  (s : forall (c : code a0), Trunc_rec (fun p => p # c0) (decode _ c) = c)
+  (r : decode _ c0 = tr idpath) (a1 : A)
+  : Tr n (a0 = a1) <~> code a1.
+Proof.
+  srapply equiv_adjointify.
+  - apply (Trunc_rec (fun p => p # c0)).
+  - apply decode.
+  - intro p.
+    pose (decode _ p) as p'.
+    clearbody p'.
+    strip_truncations.
+    destruct p' in p.
+    apply s.
+  - intros p.
+    strip_truncations.
+    destruct p.
+    exact r.
+Defined.
+
+(** Encode-decode for loop spaces *)
+Definition encode_decode_loops
+  (A : pType) (code : pFam A)
+  (decode : forall x, code x -> point A = x)
+  (s : forall (c : code.1 (point A)), decode _ c # code.2 = c)
+  (r : decode _ (pr2 code) = idpath)
+  : loops A <~> code.1 (point A)
+  := encode_decode _ _ code.1 code.2 decode s r _.
+
+(** Encode-decode for truncated loop spaces *)
+Definition encode_decode_trunc_loops n
+  (A : pType) (code : pFam A) `{forall a, IsTrunc n (code a)}
+  (decode : forall x, code x -> Tr n (point A = x))
+  (s : forall (c : code.1 (point A)),
+    Trunc_rec (fun (p : loops A) => p # code.2) (decode _ c) = c)
+  (r : decode _ code.2 = tr idpath)
+  : pTr n (loops A) <~> code (point A)
+  := encode_decode_trunc _ _ _ code.1 code.2 decode s r _.
+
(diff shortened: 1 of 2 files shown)


mikeshulman (approved):

Alizter:
@mikeshulman Do you think any proofs we have could benefit from factoring through this? The only one I can think of that really fits this is the loop circle=Z one. I don't think Blakers-Massey works right?

mikeshulman:
Nah, I wouldn't bother.

theories/Homotopy/EncodeDecode.v
```diff
@@ -0,0 +1,75 @@
...
+
+(** ** Encode-decode method *)
+
+(** Encode-decode for loop spaces *)
+Definition encode_decode_loops
+  (A : Type) (a0 : A) (code : A -> Type)
+  (c0 : code a0) (decode : forall x, code x -> a0 = x)
+  (s : forall (c : code a0), decode _ c # c0 = c)
+  (r : decode _ c0 = idpath)
+  : a0 = a0 <~> code a0.
```
jdchristensen, line 15: Does it work to generalize this to `a0 = a1 <~> code a1`?  Can this result be proved using `equiv_path_from_contr` (or vice versa)?

theories/Homotopy/EncodeDecode.v
```diff
@@ -0,0 +1,75 @@
...
+Proof.
+  assert (t : forall x (p : point A = x), decode x (p # c0) = p).
```
jdchristensen, line 17: I'm surprised that `point A` works here...

jdchristensen: Maybe it would be less confusing to use `a0` instead?

theories/Homotopy/EncodeDecode.v
Alizter: I get to
```
1 goal
A : Type
a0, a1 : A
code : A -> Type
c0 : code a0
decode : forall x : A, code x -> a0 = x
s : forall c : code a0, transport code (decode a0 c) c0 = c
r : decode a0 c0 = 1
x : A
c : code x
transport (fun y : A => code y) (decode x c) c0 = c
```
Which looks like it should be s but that doesn't quite work.

jdchristensen: This works:
```
  refine (equiv_path_from_contr a0 code c0 _ b)^-1%equiv.
  exists (a0; c0); intros [y c].
  destruct (decode y c). (* This line is the trick. *)
  snrapply path_sigma; simpl.
  1: exact (decode a0 c).
  apply s.
```
But your direct proof is probably better.  I also don't see anything really simple in the other direction.  So maybe they should just be proved separately.  But maybe they should be in the same file?  Not sure if the existing PathAny.v file is well-named or in the best location...

jdchristensen:
There's also related work in #1449.

theories/Homotopy/EncodeDecode.v
jdchristensen: Here's how the direct proof looks when the goal is generalized to involve paths instead of loops:
```
Definition encode_decode_loops
  (A : Type) (a0 : A) (code : A -> Type)
  (c0 : code a0) (decode : forall x, code x -> a0 = x)
  (s : forall (c : code a0), decode _ c # c0 = c)
  (r : decode _ c0 = idpath)
  (a1 : A)
  : a0 = a1 <~> code a1.
Proof.
  srapply equiv_adjointify.
  + exact (fun p => p # c0).
  + apply decode.
  + intro p.
    (* Do induction on [decode _ p], but leave the goal unchanged: *)
    destruct (decode _ p) in p.
    apply s.
  + intros [].
    exact r.
Defined.
```
The third subgoal (involving `s`) becomes a bit trickier, but the fourth subgoal becomes easier, without needing `r` to be generalized as `t` at the start.

Alizter: Nice, I've rewritten the proofs in the more general way and derived the pointed case for loop spaces as a special case.

jdchristensen:
This looks good.  The only comment I have left is that I think these results and the results in PathAny.v should be together, as they both give techniques for identifying path types.  I'm not sure why PathAny.v has that name or why it is in the top-level directory.  That said, because most of the new encode-decode results depend on Pointed which depends on PathAny, maybe they can't easily be put together...  In that case, maybe there should be a comment in each file pointing out that a similar result is in the other file?

Alizter:
I can move `encode-decode` into PathAny.v, then the pointed versions into Pointed somewhere and the truncated version in to Truncations/Core.v

jdchristensen:
For the pointed version, I think the only thing you need that is not in Basics is the definition of `loops`.  That could be moved to Overture.v, where `pType` is defined...

jdchristensen:
I'm not sure we're better off with these results split into separate files just so that some of them are near the results in PathAny.v.  It's odd how tricky it is to put things in places where that can be easily found and also satisfy dependency requirements...

mikeshulman:
On balance I think I would prefer leaving these in a single file, given that right now we aren't actually using them for anything.  In general my feeling is that these results are more interesting than useful.  In practice encode-decode seems more useful as an imprecisely-specified methodology than as a single precise theorem, despite that Steve insisted on stating a general theorem for the book.  PathAny, by contrast, is very useful in practice, so it makes sense to have it in a central location with minimal dependencies.

mikeshulman:
But I'm open to being convinced otherwise.

jdchristensen:
I think Mike is right.

Just one minor thing:  this doesn't need to import PathAny.v.  After that, LGTM!

theories/Homotopy/EncodeDecode.v
```diff
@@ -0,0 +1,67 @@
+Require Import Basics Types Pointed.
+Require Import Truncations.
+Require Import PathAny.
```
jdchristensen, line 3:
```suggestion
```

jdchristensen:
Oh, but I do still think comments in both files pointing to the other make sense.

mikeshulman:
By the way, I didn't mean I think Steve was wrong to insist on stating a precise theorem.  Mathematicians like to see precise theorems!  And likewise this PR is good to have.  It's just that it's hard to capture all the uses of the encode-decode "idea" in the statement of a single theorem, and even if you could it would often be more trouble than it's worth to try to contort a particular example in order to apply the theorem rather than just applying the method directly.

Alizter:
Alright, I've added a comment in EncodeDecode.v and PathAny.v pointing to each other.  I've also removed the useless import.

jdchristensen (approved):

PR #1611, opened 2021-12-29
Earring space

Alizter:
The homotopy type of the (Hawaiian) Earring space can be constructed in HoTT. It is also probably possible to give a combinatorial description of it's fundamental group but I didn't get to that.

I learnt about the inverse limit description of the Earring space from https://wildtopology.com/2013/11/23/the-hawaiian-earring/

cc @jdchristensen

### DIFF
--- a/theories/Homotopy/Bouquet.v
+++ b/theories/Homotopy/Bouquet.v
@@ -87,3 +87,7 @@ Section AssumeUnivalence.
   Defined.
 
 End AssumeUnivalence.
+
+Global Instance is0functor_bouquet : Is0Functor Bouquet := _.
+Global Instance is1functor_bouquet : Is1Functor Bouquet := _.
+
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -739,6 +739,25 @@ Proof.
   reflexivity.
 Defined.
 
+Global Instance is1functor_pointify : Is1Functor pointify.
+Proof.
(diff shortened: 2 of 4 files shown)


mikeshulman:
The construction as an inverse limit is a strict limit in the 1-category of topological spaces.  It's not obvious to me that this limit is a homotopy limit; is it?  If not, I don't think we can say the "limit" as constructed in HoTT also deserves the name "earring space".

Note that the earring space is not a CW-complex, so to represent it in HoTT I think we would have to be talking about its *weak* homotopy type...

Alizter:
I think you are correct. This is a development I had from a while back.

I think my thinking from back then was the Earring space is 1-truncated so is completely determined by it's fundamental group.

Now thinking about it, this doesn't even have the correct fundamental group. It is some weird universal-cover-looking object. The Hawaiian earring group should carve out a subgroup of this however.

Now we can take the fundamental group of the Hawaiian earring and deloop it but this wouldn't really be a homotopy type of the Hawaiian earring right? That space doesn't even live in simplicial sets for example so I guess it's not really worth talking about in this manner.

Anyway, something I thought might be worth sharing. I'll close this PR.

mikeshulman:
> Now we can take the fundamental group of the Hawaiian earring and deloop it but this wouldn't really be a homotopy type of the Hawaiian earring right?

It would be a *weak* homotopy type of the earring space, which is all that we can see when representing spaces by simplicial sets.

PR #1616, opened 2022-01-10
remove dependabot

Alizter:
I don't think dependabot is particularly useful in our case. I think it was worthwhile experimenting and I am sure that the feature will improve in the future, but at the moment the inability to disable on forks and the frequency of PRs is annoying.

Since we can count the dependencies it updates on one hand, I don't think it is worth keeping.

cc @JasonGross


JasonGross:
Let's not disable it yet.  I want to try to fix the node issue, and then I think it might be a bit more useful.

JasonGross:
We could also tell it to only check for updates monthly instead of daily

SkySkimmer:
If it's spamming forks then it must be disabled.

Alizter:
What has the node issue got to do with this?

Alizter:
In any case @JasonGross, I am disabling it now for the time being. We always have the git history if we want to resurrect it. I am currently subscribed to a few other peoples forks, which is why I am noticing the influx of PRs that dependabot has been generating. It also seems to me that the maintainers are not particularly interested in solving these issues for the time being, therefore waiting for it to mature seems like the best option.

Alizter:
cc @jdchristensen update your master branch to stop the PR spam

jdchristensen:
Thanks, I have now updated.  I was getting a bit annoyed by all of the messages too.

PR #1618, opened 2022-01-15
Presentation notation

Alizter:
Here is something fun that I have been working on. I've created a notation for group presentations allowing us to quickly write one down without having to do a lot of leg work. The new notation allows us to do things like:
```coq
Definition presentation_q16 : GroupPresentation :=
  ⟨ a , b | grp_pow a 8 , (grp_pow a 4) * -(grp_pow b 2) , a * b * a * -b ⟩.
```

Currently there is only support up to 3 generators, we can add more as needed. I haven't worked out how to get an arbitrary number of binders just yet.

### DIFF
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -110,7 +110,6 @@ Section Reduction.
   Proof.
     intros x y.
     strip_truncations.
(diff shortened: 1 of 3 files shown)


jdchristensen (approved):
Nice!  LGTM.

mikeshulman:
Neat!

PR #1620, opened 2022-01-21
fix warnings

Alizter:
The `Typeclasses Opaque`/`Transparent` commands now need locality attributes. I've defaulted to `#[global]` pretty much everywhere and that seems to work.


mikeshulman (approved):
LGTM.  What's up with CI?

Alizter:
The CI is still trying 8.15, let me merge the update then rebase.

Alizter:
Should be good now. Locality for that command is an 8.15 feature.

jdchristensen:
So we no longer support Coq 8.14?  That's what I have, so I can't build the library anymore.  I'll upgrade, but I thought we would try to maintain backwards compatibility when possible.

Alizter:
@jdchristensen You can still use versions of the library targeted for 8.14 via opam if you want to do external developments. The HoTT library always stays compatible with coq master. I think it would be a lot of work to keep the library compatible across many versions of coq, especially when we have various bugs been fixed between versions.

PR #1621, opened 2022-01-25
Syllepsis round two

kristinas:
Proof of syllepsis and some associated coherences.


Alizter:
@kristinas I've rebased and cleaned up the git history like we discussed.

Alizter:
I am in the process of trying to optimize the terms in this proof. At the moment, the CI is able to compile it. But I cannot even finish the Syllepsis file on my laptop without coq running out of memory. This is obviously not great for people trying to compile the whole library. I suspect that the many uses of `equiv_ind'` are to blame, since they copy the goal whilst guessing the type family every time. I suspect that chaining these isn't great for performance.

kristinas:
I mean, I can try to rework the proof to avoid this, if you think that's a way to go. I would personally prefer to keep the proof conceptually nicer as opposed to computationally nicer, but if we can't get it to compile on people's laptops then that's obviously a problem.

mikeshulman:
If we do manage to optimize it to make it compilable on laptops, I wonder if it would be possible to keep the conceptual proof in a way that would only be compiled by the CI?  We could just comment it out, but then it wouldn't be checked and would probably get out of date with the rest of the library.

mikeshulman:
Maybe a separate makefile target that isn't compiled by default but we can configure the CI to run?

kristinas:
But why is equiv_ind' so expensive as opposed to an ordinary induction? I am thinking of it as just another induction rule.

Alizter:
I'm not sure how much overhead it creates but when using `equiv_ind`, coq has to do a typeclass search for `IsEquiv` which can be costly each time. My suggestion of `equiv_ind'` (which doesn't exist yet but is simple to do) uses the bundled equivalence `Equiv` so coq doesn't need to do the search.

kristinas:
Oh ok, I thought you already did it and it was still a problem. Indeed, I hope that will speed things up.

Alizter:
To be clear, the actual proof of syllepsis runs completely fine. Kristina has in addition shown the following:
 - `eh` The Eckmann-Hilton cell.
 - `ehlnat`, `ehrnat` left and right natuarality of `eh`.
 - `eh_p_pp` that `eh p (q @ r)` can be characterized.
 - `eh_pp_p` that `eh (p @ q) @ r` can be characterized.
 - `eh_V` the syllepsis
 - `eh_V_p_pp` that the syllepsis `eh_V p (q @ r)` can be characterized.

Alizter:
@JasonGross How difficult would it be to set up a makefile that doesn't get used by the user but does by the CI?

Alizter:
If we merge this PR as is, we will have to set a minimum memory requirement. I've split the Syllepsis file into the actual Syllepsis part of interest and the really really big proof.

Currently loading the really really big proof in coqide takes up 5.1GiBs of memory!!

Maybe we can disable this since I am not aware of any immediate uses.

cc @jdchristensen also about this discussion.

jdchristensen:
I haven't looked at this carefully, but I don't remember any issues with the earlier proof.  Is this something that only happens with the current version?

In any case, I think we should figure out what is causing the blow up and find a way to work around it.  In the past, when dealing with equivalences and `Tr`, there were times when making something `Opaque` sped things up a lot.  E.g. #1475, but this happens in other situations as well.  I would guess that a similar trick would help here.  Maybe try some of the ideas in #1475?  Or try making some other things Opaque?

BTW, is there a particular line that can be pinpointed as causing the issue?

Alizter:
@jdchristensen To be clear, the current proof of the syllepsis runs completely fine, no memory issues. Kristina has just shown an extra coherence that explodes to 5GiB. That extra coherence (in the file Syllepsis_2.v) doesn't contain anything other than path-induction and equivalences. I am not sure if making some things Opaque will help, there are just a lot of terms in this case. I mostly wanted to get your attention to the fact that we have a nice enough proof of the Syllepsis now.

jdchristensen:
I'm more concerned about the time than the memory.  On a fast machine, building the entire library (w/o this commit) takes 1m10s using 4 cores, but building just the Syllepsis_2.v file takes 1m3s.

I think there's a lot of low hanging fruit here.  For example, just changing the `Defined` in the big definition to `Qed` reduces the time to 53s and the memory to 1.7GB.

jdchristensen:
The equivalence `lrucancel` is only used in the form `lrucancel^-1`.  Reversing the direction of the equivalence in the definition and then using `lrucancel` in place of `lrucancel^1` reduces the build time by more than 10%.  These two changes are the first two things I tried, so clearly there is lots of room to make this more efficient.  `rlucancel` is also usually used reversed, so it could be flipped, or `rlucanel_inv` could be defined once and then used repeatedly.

My rough work is here:  https://gist.github.com/jdchristensen/540c3c99e8cd4c16aaaaf5156580e3ef

jdchristensen:
Should I go ahead and push some commits to this branch?  I have a few other ideas too.

Alizter:
I think there is some more clean up that can be done such as reusing/distributing lemmas. But I think that should be left till later.

jdchristensen:
I just pushed some commits which bring the compile time for Syllepsis_2.v down to 17s, using around 620MB.  I have more ideas, but am out of time for a day or so.

Alizter:
I'm happy with merging this as is at the moment. @jdchristensen if you have some more improvements down the line, I guess you can open another PR. I am also unsure if we should keep `Syllepsis_2.v` as an extra file.

jdchristensen:
I'm not sure it's worth increasing the build time of the entire library by ~10% for the extra coherence law.  I have a few small things to try, but I don't think they will improve things a lot.  Maybe someone else has an idea?  @kristinas , could it be that by changing the order that that path inductions are done would allow us to simplify the intermediate terms?  Or maybe with the current order, carefully chosen uses of `cbn`, `simpl`, or something more precise could help?  Maybe @JasonGross has an idea?  At least now it is fast enough that it isn't too painful to experiment more.  In any case, let's wait a bit longer before deciding what to do.

Alizter:
@jdchristensen I think that is a sensible place to put it.

Alizter:
Also a more AT question: Are we observing some puffed out version of the 1st stable stem of the spheres. Kristina tells me that the syllepsis also fits into a similar identity that EH does. This seems to me like we have a series: "EH, Syl, Syl^2". Which very much looks like Z, Z_2, Z_2 ... of Pi_{1+n}(S^n). I might be imagining things though.

jdchristensen:
I just pushed more commits that clean things up and speed things up a little more.  I'm not at the same machine, but I estimate that on that machine the time to build Syllepsis_2 will be down to around 13s.  (It's 18s on a 3 year-old machine.)  I'm out of ideas for now, but welcome any other suggestions.

kristinas:
How fast should it be to be considered acceptable?

jdchristensen:
@kristinas I don't think there's a definitive answer to that.  I'm just puzzled that this result appears fairly straightforward and yet is one of the slowest things in the library.  So I suspect that there is still room for improvement.  If the result was a central one that many other things would need to build on, then of course it would be fine for it to take 20s if that was the best possible.  But since this is something that is not likely to be built upon, it's less clear whether we should build it by default.  BTW, the Qed line is the slowest line in the file, which makes it a bit hard to know what to do to improve things.

I should say that what you did is very impressive!  I don't know how you came up with exactly the right formulations and figured out the right order to induct over the variables!

kristinas:
I am now having problems compiling the library: File "./theories/Basics/Overture.v", line 598, characters 0-37:
Error: This command does not support this attribute: global.
[unsupported-attributes,parsing]

make[2]: *** [Makefile.coq:763: theories/Basics/Overture.vo] Error 1
make[1]: *** [Makefile.coq:386: all] Error 2
make: *** [Makefile:21: invoke-coqmakefile] Error 2
My Coq version is 8.14.1

jdchristensen:
@kristinas Yes, somewhat annoyingly, Coq 8.15 is now needed.  (8.14 was only released a few months ago...)

jdchristensen:
Interesting idea about using `let` in `equiv_ind`.  I tried defining
```
Definition equiv_ind' `{IsEquiv A B f} (P : B -> Type)
  : (forall x:A, let fx := f x in P (fx)) -> forall y:B, P y
  := fun g y => transport P (eisretr f y) (g (f^-1 y)).
```
and using this in `equiv_path_ind`, and it didn't change the build time of Syllepsis_2.  (Changing `equiv_ind` to use a `let` makes other parts of the library not compile, so would require adding `cbn zeta` in various places.)

kristinas:
Hmm where does the cost actually come from then? I thought it was the size of the terms.

jdchristensen:
Using `let` might temporarily keep the size down, but `Qed` needs to do some normalization to check the term and might need to reduce the `let`s.  But this gets into territory where I find it hard to predict how Coq will behave, so I'm hoping an expert can provide some advice...

One more idea:  what if the main goal of `eh_V_p_pp_gen` is replaced by a general type family `P` depending on the various arguments, and then a general induction principle is proven for such families.  Then `eh_V_p_pp_gen` would just be an application of this general principle to the specific goal involving `eh_V_gen`.  This would keep the size of the term down during the many inductions that are done.  Not sure if this would work, as the general lemma would probably need to have a very complicated term that is provided to prove the final goal...  Instead, maybe within the proof, one can temporarily abstract away the goal as a general function of the hypotheses?  I tried using `pattern` to do something like this, but didn't get too far.

jdchristensen:
As an experiment, I proved a general induction lemma for families of the complicated form (over 90 arguments to the family), and then applied that lemma where needed.  It brings the time down from 13.5s to 8.3s.  But it's currently very ugly, with the list of arguments repeated four times.  Someone who is better with ltac can probably create the abstract goal automatically from within the original long proof, so all of the arguments don't need to be repeated.  I put it here:  https://gist.github.com/jdchristensen/84ca533b9603e5758467b7d8ac367ae0

kristinas:
This is a good idea. Does this keep the time down because the actual goal is now just P?

jdchristensen:
Yes, I think that's why it speeds things up.  It's surprising to me that it still takes as long as it does, so maybe someone can improve it even further...

The other speed-ups I did made the code easier to read as well, by factoring things out, avoiding inverses, etc.  But this change, as currently written, makes it more complicated because of repetition of lists of variables.  Here's a toy example of what I'd like to be able to do using ltac.

```
Definition test {X : Type} (a b : X) (p q : a = b) : p @ q^ @ p = q @ p^ @ q.
Proof.
  transparent assert (forall (P : forall (b : X) (p q : a = b), Type), P a 1 1 -> forall b p q, P b p q).
```

I'd like to be able to generate that assertion automatically, with the part `P a 1 1` specified by hand or left as an evar.  If the user has to first do `reverse; intros X a` to leave the goal looking like `forall (b : X) (p q : a = b), p @ q^ @ p = q @ p^ @ q`, that's of course fine.  Note that in the real example, there are around 90 dependent variables, not three, and the goal depends on some of them explicitly, but not on some others.

Any experts able to help?

JasonGross:
> Any experts able to help?

Sure, what's the part of this that's tricky?  How I'd write it in pseudocode:
1. find all hypotheses of type `_ = _` which occurs in the goal, and revert it
2. find subterms of the new goal of type `_ = ?x` where `x` is a variable and revert it (or use `revert dependent`?)
3. Write a tactic that takes in a term of the form `forall a b c ..., _` and turns it into `forall a b c ..., Type` (this is the type of `P`)
4. Write a tactic that takes in a term of the form `forall a b c ..., _` and also `P` and returns `forall a b c ..., P ...`
5. Write a tactic that takes in a term of the form `forall a b c ..., _` and also the type of `P` and returns `forall (P : ...), P ... -> forall a b c ..., P ...`

jdchristensen:
In this problem, not all of the variables are paths, so 1 and 2 don't literally apply.  But that's not an issue, since I can about *every* hypothesis in the context except `X` and `a`, so I can do `reverse; intro X a` to get the goal in the form of a big `forall`.  Steps 3 and 4 are the ones I don't know how to do.

JasonGross:
Here's some code, let me know if you want me to explain it anymore, or if there are tweaks you want to make but don't know how to:
```coq
Ltac make_P ty :=
  lazymatch ty with
  | forall a : ?A, ?ty
    => let ty' := fresh in
       constr:(forall a : A,
                  (* bind [ty] in [match] so that we avoid issues such as https://github.com/coq/coq/issues/7299 and similar ones *)
                  match ty return _ (* without [return _], [match] tries two ways to elaborate the branches, which results in exponential blowup on failure *) with
                  | ty'
                    => ltac:(let ty := (eval cbv delta [ty'] in ty') in
                             clear ty';
                             let res := make_P ty in
                             exact res)
                  end)
  | _ => constr:(Type)
  end.

Ltac apply_P ty P :=
  lazymatch ty with
  | forall a : ?A, ?ty
    => let ty' := fresh in
       let P' := fresh in
       constr:(forall a : A,
                  (* bind [ty] in [match] so that we avoid issues such as https://github.com/coq/coq/issues/7299 and similar ones *)
                  match ty, P a return _ (* without [return _], [match] tries two ways to elaborate the branches, which results in exponential blowup on failure *) with
                  | ty', P'
                    => ltac:(let ty := (eval cbv delta [ty'] in ty') in
                             let P := (eval cbv delta [P'] in P') in
                             clear ty' P';
                             let res := apply_P ty P in
                             exact res)
                  end)
  | _ => P
  end.

Ltac apply_P_evars P :=
  lazymatch type of P with
  | forall a, _
    => apply_P_evars open_constr:(P _)
  | _ => P
  end.

Ltac make_P_and_evars ty :=
  let P_ty := make_P ty in
  let P := fresh "P" in
  open_constr:(forall P : P_ty,
                  ltac:(let res := apply_P_evars P in exact res)
                  -> ltac:(let res := apply_P ty P in exact res)).

Definition test {X : Type} (a : X) : forall (b : X) (p q : a = b), True.
Proof.
  let G := match goal with |- ?G => G end in
  let P_ty := make_P G in
  idtac P_ty; (* (forall b : X, a = b -> a = b -> Type) *)
  let T := make_P_and_evars G in
  idtac T. (* (forall P : forall b : X, a = b -> a = b -> Type, P ?b ?p ?q -> forall (b : X) (p q : a = b), P b p q) *)
```

JasonGross:
Note, however, that Ltac is quite bad (performance-wise) with nested binders, so if you're hoping for a performance-gain by using Ltac to generate this term with 90 binders, be prepared that it might actually be slower.  If you want to use Ltac2 instead, it's probably possible to do in a much faster way.

theories/Homotopy/Syllepsis_2.v
```diff
@@ -0,0 +1,321 @@
...
+
+  destruct wry0, wry1, wrz0, wrz1.
+  revert wlx0.
+  snrapply equiv_path_ind_lrucancel.
+  revert wlx1.
+  snrapply equiv_path_ind_lrucancel.
+  destruct wlx2.
+  exact H.  }
+  apply lem.
+  reflexivity.
```
Alizter, line 305: I've never seen this warning before.

Alizter: Just after `reflexivity` the shelf is full of all sorts of goals that coq is able to solve automatically at Qed time. The warning is emitted since you should really be doing this yourself.

Alizter: To be more explicit, replace `reflexivity` with
```coq
  Unshelve.
  all: try reflexivity.
```

mikeshulman: What is the normative principle according to which the user "should" solve goals that Coq is able to solve itself?

theories/Homotopy/Syllepsis_2.v
```diff
@@ -0,0 +1,321 @@
...
+Qed.  (* This line is a bit slow. *)
```
Alizter, line 306: This no longer makes such a big difference on my machine at least. You can leave it as `Defined` if you wish.

theories/Homotopy/Syllepsis_2.v
Alizter: It used to be an error in Coq to leave shelved goals at Qed, but some users of other proof langauges (MTac) wanted something where Coq is able to solve the terms at unification. in https://github.com/coq/coq/pull/9347 this error was changed to a warning to allow this.

I am at a loss for why these are shelved at all however, I expected reflexivity to have solved them.

mikeshulman: That doesn't answer my question.  Why does it even need to be a warning?

Alizter: From reading the discussion, it seems that the actual proof term doesn't have any holes in it. Only that some ltac tactics have asked for an evar and not used them. So this seems to be a bug in the ltac tactic rather than something at Qed time. It appears to NOT be the case that Qed is solving goals as I originally suspected.

jdchristensen:
@JasonGross That works!  It takes 1.8s for the tactics to generate the new intermediate goal, but even though that is a bit slow, overall the file compiles even faster than before.  It's down to 6.7s on a fast machine.  I noticed that the tactics make three passes over the type of the goal.  `make_P G` takes 0.5s.  `make_P G` and `apply_P_evars P` take 0.6s.  And `apply_P G P` takes 1.0s.  Can it be done with one pass? ...  Ah, yes, we don't need `P` applied to lots of evars, but can just use a single evar.  That eliminates a lot of the tactic code and speeds things up even more.  The tactic now takes 0.9s and the file takes 5.9s.

I also found the inner match confusing (the one with a comment about Coq issue 7299), and wonder if it can be avoided?

The last commit is looking pretty good to me.  @JasonGross Can you take a look and see if you can find any further simplifications and/or speed ups?

theories/Homotopy/Syllepsis_2.v
Alizter: This doesn't seem to be a problem with reflexivity actually. Dan's tactic seems to be generating useless evars.

Alizter: OK seems to be fixed in the latest commit.

jdchristensen: Sorry, that was because of the timing tests that were in the second-last commit.  They are gone in the latest commit.

theories/Homotopy/Syllepsis_2.v
jdchristensen: Yes, it doesn't make that much difference now.  I'm also using `assert` instead of `transparent assert`, and that also only saves a bit of time.  I think I'll leave them opaque for now anyways.

Alizter:
I don't know if you also noticed, but stepping through this proof line-by-line is really slow. I take it @jdchristensen you are using emacs, do you have the same issue? I suspect this might be coq being slow to print the goal.

jdchristensen:
I'm using emacs, and I don't notice it being slow.  Most lines feel like they take < 0.5s.  The slowest lines are the *statement* of the big definition (1.3s), then the tactic that creates the assertion (0.9s), and then the Qed (0.8s).  I do suspect that the proof of the assertion could probably be sped up a bit more.

JasonGross:
> Can it be done with one pass? ... Ah, yes, we don't need `P` applied to lots of evars, but can just use a single evar. That eliminates a lot of the tactic code and speeds things up even more. The tactic now takes 0.9s and the file takes 5.9s.

Here's a faster Ltac2 version:
```coq
Require Import Ltac2.Ltac2.

Ltac2 rec replace_under_prod (ty : constr) (final : constr) :=
  match Constr.Unsafe.kind ty with
  | Constr.Unsafe.Prod a ty
    => Constr.Unsafe.make (Constr.Unsafe.Prod a (replace_under_prod ty final))
  | _ => final
  end.

Ltac2 make_P (ty : constr) := replace_under_prod ty 'Type.

Ltac2 rec count_prod (ty : constr) :=
  match Constr.Unsafe.kind ty with
  | Constr.Unsafe.Prod _ ty => Int.add 1 (count_prod ty)
  | _ => 0
  end.

Ltac2 apply_P_on (p : constr) (n : int) :=
  Constr.Unsafe.make
    (Constr.Unsafe.App
       p
       (Array.init
          n
          (fun i => Constr.Unsafe.make (Constr.Unsafe.Rel (Int.sub n i))))).

Ltac2 apply_P (ty : constr) (p : int -> constr) :=
  let n := count_prod ty in
  replace_under_prod ty (apply_P_on (p n) n).

Ltac2 make_P_and_evars (ty : constr) :=
  let p_ty := make_P ty in
  let res := apply_P ty (fun n => Constr.Unsafe.make (Constr.Unsafe.Rel (Int.add n 1))) in
  '(forall P : $p_ty, _ -> $res).

Ltac make_P_and_evars :=
  ltac2:(ty |- Control.refine (fun _ => make_P_and_evars (Option.get (Ltac1.to_constr ty)))).

Set Default Proof Mode "Classic".
Definition test {X : Type} (a : X) : forall (b : X) (p q : a = b), True.
Proof.
  let G := match goal with |- ?G => G end in
  let T := open_constr:(ltac:(make_P_and_evars G)) in
  idtac T. (* (forall P : forall b : X, a = b -> a = b -> Type, ?Goal0 -> forall (b : X) (p q : a = b), P b p q) *)
```

> I also found the inner match confusing (the one with a comment about Coq issue 7299), and wonder if it can be avoided?

We can avoid it if we don't care about preserving binder names.  Here's a way to do it differently:
```coq
Ltac make_P ty :=
  lazymatch ty with
  | forall a : ?A, ?ty
    => let a' := fresh in
       constr:(forall a' : A,
                  ltac:(let ty := constr:(match a' with a => ty end) in
                        let res := make_P ty in
                        exact res))
  | _ => constr:(Type)
  end.

Ltac apply_P ty P :=
  lazymatch ty with
  | forall a : ?A, ?ty
    => let a' := fresh in
       constr:(forall a' : A,
                  ltac:(let ty := constr:(match a' with a => ty end) in
                        let P := uconstr:(P a') in
                        let res := apply_P ty P in
                        exact res))
  | _ => P
  end.
```

theories/Homotopy/Syllepsis.v
```diff
@@ -921,3 +921,305 @@ Proof.
...
+  ltac2:(ty |- Control.refine (fun _ => make_P_and_evars (Option.get (Ltac1.to_constr ty)))).
+
+Ltac generalize_goal :=
+  let lem := fresh "lem" in
+  match goal with |- ?G =>
+                    let T := open_constr:(ltac:(make_P_and_evars G)) in
+                    assert (lem : T) end.
+
+(* Make Ltac1 the default again. *)
+Set Default Proof Mode "Classic".
```
Alizter, line 972: Incidentally, Ltac2 leaking the proof mode was fixed today. https://github.com/coq/coq/pull/15615. So perhaps we can add a comment saying "Remove in Coq 8.16".

jdchristensen:
Wow, Ltac2 is super fast!  The new tactic takes essentially no time.  Thanks @JasonGross !  It's too bad that the tactics are so hard for a newbie to understand.  I guess those are de Bruijn indices?

Syllepsis_2 is down to 4.6s, so I merged it back into Syllepsis.  I left the tactics just before where they are needed, but am happy to move any that might be of general use elsewhere.  Maybe it's best for now to leave them here since they use Ltac2, which we don't use elsewhere?

I'm happy for this to be merged whenever others agree.

Summary of things I learned:
- `equiv_path_ind` encodes a nice idiom that could probably be used in other proofs.  Namely, if you have a path `p` in a type that is *equivalent* to a path type with one end free, you can use `equiv_path_ind` to do induction on the path.
- The new `generalize_goal` tactic can speed up manipulating large goals, with almost no change needed to the structure of the proof.  Maybe we should use it more?
- Ltac2 is fast!  Maybe we should use it more?

mikeshulman:
What exactly does `generalize_goal` do?

I would love to use more Ltac2, but I haven't been able to find documentation that I could use to learn it.  Is it more documented than it used to be?

jdchristensen:
@mikeshulman Here's how I explained it in the last commit:
```
(* The following tactics will be used to make the proof faster, but with only minor modifications, the proof goes through without these tactics. The final tactic [generalize_goal] takes a goal of the form [forall a b c ..., expression] and asserts a new goal [forall P, _ -> forall a b c ..., P a b c ...] which can be used to prove the original goal. Because [expression] has been replaced with a generic function, the proof of the new goal can be more efficient that the proof of the special case, especially when there are around 84 variables. *)
```

Alizter:
@jdchristensen I was wrong about the proof mode comment. Since we use ltac in the rest of the file, we do need to set the proof mode back to "Classic" again. The bug I mentioned before: https://github.com/coq/coq/pull/15615 but doesn't affect us really.

>The new generalize_goal tactic can speed up manipulating large goals, with almost no change needed to the structure of the proof. Maybe we should use it more?

I would be interested in seeing if the generalize_goal tactic can speedup proofs in Cubical.PathCube, DPathCube etc.

>Ltac2 is fast! Maybe we should use it more?

I think this would be a good idea! Since it was designed from the ground up, it learnt from the many mistakes that were made during the original development of Ltac, hence is much better behaved.

@mikeshulman Documentation for Ltac2 is here https://coq.github.io/doc/master/refman/proof-engine/ltac2.html I don't know if it is different from when you last saw it.

Alizter:
Also, the slowness in CoqIDE I was observing earlier was due to https://github.com/coq/coq/issues/10623. Completely a CoqIDE thing hence why nobody using emacs noticed.

mikeshulman:
I want a tutorial, not a reference manual.  And it has to start with basics, not assuming I am already an Ltac1 wizard because I'm not.

Alizter:
@mikeshulman There are no tutorials from the Coq side only a refman AFAIK.

mikeshulman:
So then the answer to my question is no, and my answer to @jdchristensen 

> Ltac2 is fast! Maybe we should use it more?

is that I'm dubious about the idea until there's a way for me (and other users who are even less wizardly than I) to learn it.  I'm okay with calling out to it in a few speed-critical places like this, but as a general matter I value comprehensibility a lot too.

JasonGross:
> I want a tutorial, not a reference manual. And it has to start with basics, not assuming I am already an Ltac1 wizard because I'm not.

Indeed there's no tutorial, only a reference manual, but the reference manual is sort-of tutorial-like, at least a little bit.  Roughly the programming model and syntax of Ltac2 is that of OCaml, though there's some syntactic notations that make some parts of it like Ltac (such as `match goal`).

mikeshulman:
The syntax isn't the hard part, how it interacts with Coq is the hard part.

SkySkimmer:
This is using Ltac2 from the stdlib and therefore Require-ing the stdlib prelude, is that acceptable?

JasonGross:
> I guess those are de Bruijn indices?

Yep!  Not sure if they're 0-indexed or 1-indexed.  Roguhly, the reason to use de Bruijn indices is that it's expensive to set up a context where all the binders are in scope, especially when dealing with evars.  So we go the unsafe route and build the term we want without typechecking it, which means using de Bruijn indices.

theories/Homotopy/Syllepsis.v
```diff
@@ -0,0 +1,1225 @@
...
+  ltac2:(ty |- Control.refine (fun _ => make_P_and_evars (Option.get (Ltac1.to_constr ty)))).
+
+Ltac generalize_goal :=
+  let lem := fresh "lem" in
+  match goal with |- ?G =>
+                    let T := open_constr:(ltac:(make_P_and_evars G)) in
+                    assert (lem : T) end.
+
+(* Make Ltac1 the default again. Only needed while we support Coq <= 8.15. *)
+Set Default Proof Mode "Classic".
```
SkySkimmer, line 972: Nothing about this changes in 8.16.

JasonGross:
> The syntax isn't the hard part, how it interacts with Coq is the hard part.

I'm sure @jferle and @ppedrot will be happy to take suggestions about how Ltac2 should be documented / which parts are under-documented.  I'm also happy to explain any bits of it that are confusing or don't make sense.

SkySkimmer:
>I'm sure @jferle and @ppedrot will be happy to take suggestions about how Ltac2 should be documented / which parts are under-documented. I'm also happy to explain any bits of it that are confusing or don't make sense.

typo, should be @jfehrle

theories/Homotopy/Syllepsis.v
Alizter: Yep sorry that was my mistake.

Alizter:
@mikeshulman I found this tutorial for Ltac2: https://github.com/tchajed/ltac2-tutorial/blob/master/src/ltac2_tutorial.v

mikeshulman:
Haha, yes, I just found that tutorial too.  It's not exactly inspiring, with comments like

> Picking an application and really trying to use Ltac2 for it is likely to uncover more tricks and bug reports necessary to make the language work.

and

> You will need to read the Ltac2 source to figure out how things work. This ranges from fine to totally impossible.

However, it's almost 2 years old, so maybe things are better by now.

jdchristensen:
> This is using Ltac2 from the stdlib and therefore Require-ing the stdlib prelude, is that acceptable?

I don't know.  If people think I should revert to the Ltac1 version of the tactic, just let me know.  The Ltac1 version took 0.9s, so it is not so bad to use it.  And it is shorter and easier to understand than the Ltac2 version.

Alizter:
Regarding dependence on coq stdlib. I don't think it is a big problem to be dependent on Ltac2.Ltac2. AFAIK there is nothing from the stdlib being imported/exported/used here. I could submit a PR upstream that changes compilation of Ltac2 to use `-noinit` which would stop `Coq.Init` from appearing.

Alizter:
Doing this for example shows the issue more clearly:
```coq
Print Libraries.
Require Import Ltac2.Ltac2.
Print Libraries.
```

<details>
<summary>Before</summary>


```
Loaded library files: 
  HoTT.Basics.Notations
  HoTT.Basics.Logic
  HoTT.Basics.Datatypes
  HoTT.Basics.Overture
  HoTT.Basics.Decimal
  HoTT.Basics.Hexadecimal
  HoTT.Basics.Numeral
  HoTT.Basics.Nat
  HoTT.Basics.Tactics
  HoTT.Basics.PathGroupoids
  HoTT.Basics.UniverseLevel
  HoTT.Basics.Contractible
  HoTT.Basics.Equivalences
  HoTT.Basics.Trunc
  HoTT.Basics.Decidable
  HoTT.Basics.Utf8
  HoTT.Basics
  HoTT.Types.Unit
  HoTT.Types.Empty
  HoTT.Types.Paths
  HoTT.Types.Prod
  HoTT.Types.Forall
  HoTT.Types.Arrow
  HoTT.Types.Sigma
  HoTT.Types.Equiv
  HoTT.Types.Bool
  HoTT.Types.Universe
  HoTT.Types.Sum
  HoTT.Types.WType
  HoTT.Types.IWType
  HoTT.Types
```

</details>

<details>
<summary>After</summary>

Loaded library files: 
  HoTT.Basics.Notations
  HoTT.Basics.Logic
  HoTT.Basics.Datatypes
  HoTT.Basics.Overture
  HoTT.Basics.Decimal
  HoTT.Basics.Hexadecimal
  HoTT.Basics.Numeral
  HoTT.Basics.Nat
  HoTT.Basics.Tactics
  HoTT.Basics.PathGroupoids
  HoTT.Basics.UniverseLevel
  HoTT.Basics.Contractible
  HoTT.Basics.Equivalences
  HoTT.Basics.Trunc
  HoTT.Basics.Decidable
  HoTT.Basics.Utf8
  HoTT.Basics
  HoTT.Types.Unit
  HoTT.Types.Empty
  HoTT.Types.Paths
  HoTT.Types.Prod
  HoTT.Types.Forall
  HoTT.Types.Arrow
  HoTT.Types.Sigma
  HoTT.Types.Equiv
  HoTT.Types.Bool
  HoTT.Types.Universe
  HoTT.Types.Sum
  HoTT.Types.WType
  HoTT.Types.IWType
  HoTT.Types
  Coq.Init.Notations
  Coq.Init.Ltac
  Coq.Init.Logic
  Coq.Init.Datatypes
  Coq.Init.Specif
  Coq.Init.Decimal
  Coq.Init.Hexadecimal
  Coq.Init.Number
  Coq.Init.Nat
  Coq.Init.Byte
  Coq.Init.Peano
  Coq.Init.Wf
  Coq.Init.Tactics
  Coq.Init.Tauto
  Coq.Init.Prelude
  Ltac2.Init
  Ltac2.Int
  Ltac2.Message
  Ltac2.Control
  Ltac2.Bool
  Ltac2.Array
  Ltac2.Char
  Ltac2.Constr
  Ltac2.Std
  Ltac2.Env
  Ltac2.List
  Ltac2.Fresh
  Ltac2.Ident
  Ltac2.Ind
  Ltac2.Ltac1
  Ltac2.Option
  Ltac2.Pattern
  Ltac2.Printf
  Ltac2.String
  Ltac2.Notations
  Ltac2.Ltac2

```

</details>

Alizter:
I had a look at making Ltac2 noinit, I think it is doable, but earliest will appear in 8.16.

jfehrle:
> I'm sure @jfehrle and @ppedrot will be happy to take suggestions about how Ltac2 should be documented / which parts are under-documented.

Yes, though I don't expect to make major changes to that chapter soon.  The doc and the software need a lot of work.  But maybe can do some smaller improvements sooner (eager to finish the debugger first, including for Ltac2).  The research and creating examples takes a long time (need an expert to help), plus it usually takes multiple go arounds to get clear & accurate descriptions.

jfehrle:
@Alizter Is "synthetic homotopy" the best label were can come up with?  I have zero idea where to look in the doc or the code for that label.  Is there a "synthetic homotopy" module somewhere I'm not aware of?

Alizter:
@jfehrle Well that's a label we use internally in the HoTT library which this PR is for. It doesn't have anything to do withe the labels in the coq repo.

jfehrle:
Hmm, perhaps it would be clearer with an additional "part: libraries" tag.

kristinas:
This is a great speedup Dan and Jason!

jdchristensen:
Should we add Homotopy/Syllepsis.v and Homotopy/EncodeDecode.v to HoTT.v?  Everything else in the Homotopy folder is included.

Alizter:
@jdchristensen I can submit a PR.

PR #1622, opened 2022-01-26
Adapt to Coq PR #15537: schemes paths_rec & paths_ind have different automatically generated names

herbelin:
This is a consequence of fixing a naming bug in coq/coq#15537. It now requires an explicit `rename` flag in the subsequent calls to `Arguments` for `paths` induction schemes.

A priori, I don't see other consequences of the naming fix (except on printing, where sometimes, especially in `return` clauses, `y` will be displayed `a`).

If ok with the change, this is backward-compatible and can be merged as soon as now.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -242,9 +242,9 @@ Cumulative Inductive paths {A : Type} (a : A) : A -> Type :=
 Arguments idpath {A a} , [A] a.
 
 Scheme paths_ind := Induction for paths Sort Type.
-Arguments paths_ind [A] a P f y p.
+Arguments paths_ind [A] a P f y p : rename.
 Scheme paths_rec := Minimality for paths Sort Type.
-Arguments paths_rec [A] a P f y p.
+Arguments paths_rec [A] a P f y p : rename.
 
 Register idpath as core.identity.refl.
 


theories/Basics/Overture.v
```diff
@@ -673,6 +673,8 @@ Ltac path_via mid :=
 Inductive Empty : Type0 := .
 Register Empty as core.False.type.
 
+Register Empty as core.False.type.
+
```
Alizter, line 677: Why has this happened twice?

theories/Basics/Overture.v
```diff
@@ -701,6 +703,9 @@ Class Asymmetric {A} (R : Relation A) :=
 (** Likewise, we put [Unit] here, instead of in [Unit.v], because [Trunc] uses it. *)
 Inductive Unit : Type0 := tt : Unit.
 
+Register Unit as core.True.type.
+Register tt as core.True.I.
+
```
Alizter, line 708: Also these are done a couple of lines down.

Alizter (changes requested):

theories/Basics/Overture.v
herbelin: Sorry, I messed up the patch by working in a non-fresh archive. Should not be fixed.

Alizter (approved):
LGTM

PR #1624, opened 2022-01-30
Rename various types (including Nat) to adhere to capitalization rule

### LINKED ISSUE #1470: Shall we change the name of `nat` to `Nat`?
Since I am cleaning up the nat library in #1469 it might be a good time to consider doing this.

The proposal is to change the name `nat` to `Nat`. This puts us more in line with our naming convention. What do you think?

Alizter:
I've renamed the following:
 - `option` -> `Option`
 - `sum` -> `Sum`
 - `prod` -> `Prod`
 - `iff` -> `Iff`
 - `list` -> `List`
 - `nat` -> `Nat`
 
Together with their corresponding induction principles.

I've also fixed a typo in `Smash.v`.

These are all in separate commits so we can undo any we don't wish to do. I believe these are all worth doing. Notably, Dan raised concerns that `Nat` related lemmas may be harder to find, but this renaming doesn't affect the lemmas.

Fixes #1470

### DIFF
--- a/theories/Sets/Ordinals.v
+++ b/theories/Sets/Ordinals.v
@@ -136,16 +136,6 @@ Proof.
 Defined.
 
 
-Global Instance transitive_iff
-  : Transitive iff.
-Proof.
-  intros A B C A_B B_C.
-  split.
-  - intros a. apply B_C. apply A_B. exact a.
-  - intros c. apply A_B. apply B_C. exact c.
-Qed.
-
-
 Lemma transitive_Isomorphism
   : forall A B C,
     Isomorphism A B
--- a/STYLE.md
+++ b/STYLE.md
@@ -722,7 +722,7 @@ argument in question.  One way to avoid this is to instead use a
 `Fixpoint` definition, or the tactic `fix`, along with `destruct`.
 There is a tactic `simple_induction` defined in `Overture` whose
 interface is almost the same as `induction` but does this internally,
-although it only works for induction over `nat` and `trunc_index`.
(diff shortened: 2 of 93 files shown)


Alizter:
It seems that my renaming of Nat has been too eager and has changed some comments. Let me fix those.

theories/Sets/Ordinals.v
```diff
@@ -136,16 +136,6 @@ Proof.
...
-Global Instance transitive_iff
-  : Transitive iff.
-Proof.
-  intros A B C A_B B_C.
-  split.
-  - intros a. apply B_C. apply A_B. exact a.
-  - intros c. apply A_B. apply B_C. exact c.
-Qed.
-
-
```
Alizter, line 148: We already have this instance so I got rid of it.

Alizter:
There is also the issue of `paths` which I haven't touched. Should we rename it to `Paths` or more correctly `Path`?

Alizter:
@mikeshulman Any thoughts on this?

mikeshulman (approved):

Alizter:
@jdchristensen Can I get your thoughts on this also?

Alizter:
Rebased

jdchristensen:
I think some minimal compatibility with the Coq stdlib is useful, but for types that are unique to HoTT, we might as well be systematic.

mikeshulman:
We could also keep the old names but make Notations for the capitalized ones that we use everywhere.

PR #1625, opened 2022-02-02
propositional truncation via join construction

Alizter:
While working on the full join construction, I thought that this would be a good addition to our metatheory folder. We define truncations as a private inductive type at the moment, but this shows that having coequalizers is enough.

### DIFF
--- a/theories/Colimits/Sequential.v
+++ b/theories/Colimits/Sequential.v
@@ -16,7 +16,7 @@ Local Open Scope path_scope.
 Notation coe := (transport idmap).
 Notation "a ^+" := (@arr sequence_graph _ _ _ 1 a).
 
-(** Mapping spaces from colimits of sequences can be characterized. *)
+(** Mapping spaces into hprops from colimits of sequences can be characterized. *)
 Lemma equiv_colim_seq_rec `{Funext} (A : Sequence) (P : Type) `{IsHProp P}
   : (Colimit A -> P) <~> (forall n, A n -> P).
 Proof.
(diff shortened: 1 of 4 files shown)


theories/Colimits/Sequential.v
```diff
@@ -16,6 +16,55 @@ Local Open Scope path_scope.
...
+    1: exact (colimp (D:=B) _ _ idpath x).
+    refine (IHi (a i) @ _).
+    refine ((colimp (D:=B) _ _ idpath (a i))^ @ _).
+    refine ((colimp (D:=B) _ _ idpath (a i.+1))^ @ _).
+    exact (colimp (D:=B) _ _ idpath x). }
+  intros n m [] x.
+  rewrite transport_paths_FlFr.
+  rewrite ap_const.
+  rewrite ap_idmap.
+  destruct n; simpl; hott_simpl.
```
Alizter, line 65: I got a bit lazy here, but since this entire term is Qed I don't think it matters.

theories/Colimits/Sequential.v
```diff
@@ -16,6 +16,55 @@ Local Open Scope path_scope.
 Notation coe := (transport idmap).
 Notation "a ^+" := (@arr sequence_graph _ _ _ 1 a).
 
+(** Mapping spaces from colimits of sequences can be characterized. *)
```
Alizter, line 19:
```suggestion
(** Mapping spaces into hprops from colimits of sequences can be characterized. *)
```

jdchristensen:
Nice!  This LGTM.

Issue #1629, opened 2022-02-06
Fix dpdgraph job for CI


Alizter:
There was an issue with the version of node/npm for this job. Fixing that caused a stackoverflow. This issue serves as a reminder to fix that. For now we will disable the job.

JasonGross:
Did increasing stack size with, e.g., `ulimit -s 32768` do anything?

Alizter:
I didn't try. I can have another go.

Alizter:
Didn't work
https://github.com/Alizter/HoTT/runs/5085617836?check_suite_focus=true#step:6:3312

PR #1630, opened 2022-02-06
disable dpdgraph job

Alizter:
For now I haven't been able to fix it. I've opened #1629 and disabled the job for the time being.


JasonGross:
If we're doing this, how about instead we just do `make -k || true`.  That way we get at least some dependency graphs

JasonGross:
That's true.  You can add an extra step that runs `make -k` again so we can see what failed easily.  We could also add a problem matcher to report error annotations.

PR #1631, opened 2022-02-09
More speedups to Syllepsis.v

jdchristensen:
- The main speedup comes from factoring out a subterm of the goal of eh_V_p_pp_gen using a let binding.  The subterm appears four times in the goal.  The statement of this goal was the slowest step in the file.  This change speeds it up, and also speeds up the proof.
- The other let bindings in the goal were removed, since they only appeared once in the goal and slowed things down slightly.
- The induction steps in the proof were reordered, giving some more minor speedups.
- The proof of the corollary eh_V_p_pp was adjusted to speed it up a bit.

Altogether these take about 15% off the build time for the entire file, mostly due to the first change.

### DIFF
--- a/theories/Homotopy/Syllepsis.v
+++ b/theories/Homotopy/Syllepsis.v
@@ -1082,14 +1082,14 @@ Definition eh_V_p_pp_gen {X : Type}
            (H_wlrnat_V_x_yz :
              Wlrnat_V_p_pp H_ehrnat_yz0 H_ehrnat_yz1 H_wlrnat_x_yz H_wlrnat_yz_x wlrnat_V_x_y wlrnat_V_x_z =
                wlrnat_V_x_yz)
-  : let subterm := concat_p_pp x0 y0 z0 @
-                  whiskerR (((rlucancel_inv (ulnat_x0 [-] urnat_y0))^ @ wlrnat_x_y) @
+  : let eh_x_y := concat_p_pp x0 y0 z0 @
+                 whiskerR (((rlucancel_inv (ulnat_x0 [-] urnat_y0))^ @ wlrnat_x_y) @
                             rlucancel_inv (urnat_y1 [-] ulnat_x1)) z0 in
-    whiskerR (concat_p1 _ @@ concat_p1 _) subterm @
-    whiskerR (eh_V_gen (ehlnat_1p_x0) (ehlnat_1p_x2) (ehrnat_p1_yz0) (ehrnat_p1_yz1) wlrnat_V_x_yz) subterm @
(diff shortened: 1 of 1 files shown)


theories/Homotopy/Syllepsis.v
```diff
@@ -1082,29 +1082,36 @@ Definition eh_V_p_pp_gen {X : Type}
...
-    (ehrnat_p1_z0) (ehrnat_p1_z1) (wlrnat_V_x_z) in
-  let eh_V_x_yz := eh_V_gen (ehlnat_1p_x0) (ehlnat_1p_x2)
-    (ehrnat_p1_yz0) (ehrnat_p1_yz1) (wlrnat_V_x_yz) in
-  whiskerR (concat_p1 _ @@ concat_p1 _) _ @ whiskerR eh_V_x_yz _ @ lrucancel 1 @
-  whiskerL _ (Syllepsis.concat_pp_p_p_pp _ _ _)^ @ whiskerL _ (concat_p1 _ @@ concat_p1 _)^ =
-  (eh_p_pp_gen H_urnat_yz0 H_urnat_yz1 H_wlrnat_x_yz [-]
-   lrucancel (whiskerL _ (ap (fun p => whiskerL y1 p) (moveL_V1 _ _ eh_V_x_z)))) [-]
-  (eh_pp_p_gen H_ulnat_yz1 H_ulnat_yz0 H_wlrnat_yz_x [-]
-   lrucancel (whiskerL _ (ap (fun p => whiskerR p z0) (moveL_1V _ _ eh_V_x_y)))).
+  : let subterm := concat_p_pp x0 y0 z0 @
```
mikeshulman, line 1085: I don't suppose we could give this a more informative name than `subterm`?

mikeshulman (approved):

theories/Homotopy/Syllepsis.v
jdchristensen: @kristinas Can you think of a good name for the subterm I isolated that comes up four times?

kristinas: These are really impressive speedups Dan! Thanks for doing this. The subterm, well, it's an application of associativity followed by (essentially) eckmann_hilton, so maybe just call it eh_x_y?

Alizter (approved):

jdchristensen:
> Yes, thank you. How does it compare speed-wise now with the rest of the library?

The whole file takes 6.5s on my machine, while the whole library takes 2m54s (on one core), so Syllepsis.v is 3.7% of the overall build time.

PR #1632, opened 2022-02-10
Make dpdgraph job more lenient and increase stacksize.

Alizter:
Hopefully this should allow the CI to finish. I just noticed also that we haven't been deploying doc for a while so that should be fixed.


Alizter:
Tada

PR #1633, opened 2022-02-10
fix typo in ci causing file-dep-graphs to not be deployed

Alizter:
We have dependency graphs for each file. A typo in the CI file caused these not to be uploaded to the ghpages website.


Alizter:
Merging this since this is a trivial behind-the-scenes fix

PR #1635, opened 2022-02-28
autogen.sh: clean up detritus

artagnon:


Alizter:
The reason we kept this in was a kind of "deprecation period". As it has been a while since the old install method worked, I think it is a good idea to remove this script as it just adds confusion.

Alizter (approved):

PR #1636, opened 2022-02-28
ci.yml:  remove ocaml-version hardcoding in one place

artagnon:


.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
...
+on: [push, pull_request]
 
 # We set the supported coq-version from here. In order to use this environment variable correctly, look at how they are used in the following jobs.
 env:
-  coq-version-supported: '8.15'
-  ocaml-version: '4.11-flambda'
-  deployment-branch: 'gh-pages'
+  coq-version-supported: "8.15"
+  ocaml-version: "4.12-flambda"
+  deployment-branch: "gh-pages"
```
Alizter, line 9: In the Github actions workflow documentation it says single quotes not double quotes should be used: https://docs.github.com/en/actions/learn-github-actions/expressions#literals.

I'm confused as to why this has worked however.

Alizter: In any case, since the documentation is recommending the use of single quotes I would prefer to keep all strings with single quotes. The fact that double quotes works is either a bug in the documentation or the implementation of the github actions runner.

.github/workflows/ci.yml
```diff
@@ -1,12 +1,12 @@
...
-on: [ push , pull_request ]
+on: [push, pull_request]
 
 # We set the supported coq-version from here. In order to use this environment variable correctly, look at how they are used in the following jobs.
 env:
-  coq-version-supported: '8.15'
-  ocaml-version: '4.11-flambda'
-  deployment-branch: 'gh-pages'
+  coq-version-supported: "8.15"
+  ocaml-version: "4.12-flambda"
```
Alizter, line 8: Out of curiosity, why the version bump?

Alizter (commented):
I'm OK with the other white space changes.

.github/workflows/ci.yml
artagnon: I saw #1634, and thought I'd write something for testing with dune; I opened up ci.yml, found the hard-coding, and thought I might as well do a minor version-bump to check that HoTT still works. Side-note: IIUC, Coq doesn't build with OCaml 5.0 multicore yet, due to some bugs.

.github/workflows/ci.yml
artagnon: Okay, this was VSCode's automatic format-on-save. I'll drop the whitespace patch.

.github/workflows/ci.yml
Alizter: The OCaml versions that work are [here](https://hub.docker.com/r/coqorg/coq/) since we are running the docker images.

We shouldn't need to touch the OCaml versions at all when testing dune however.

artagnon: Sorry, my confusion: I thought HoTT had its own init, and a bit of custom ml code to initialize coqc/coqtop with custom_args. Looks like it's pure Coq code, so bumping the version of OCaml is unnecessary.

p.s. That page doesn't have images for everything. For instance, 4.13-flambda works for me.

Alizter: That used to be the case, but I turned it into a pure coq project last year. You only need to pass `-indices-matter` and `-noinit`. This is detailed in the README.md.

artagnon: Ah, okay! I had contributed some exercises to the project about a couple of years ago, so I'm a bit dated. Nice to see `hoqtop` being deprecated :)

Alizter:
The jobs in the CI continue after the initial builds, so the OCaml versions used in the documentation should not be changed as they are irrelevant to the original build and only relevant to the documentation tools. In this case Alectryon needs 4.11.

Alizter:
As it is, I don't see the point of changing the OCaml version here.

artagnon:
I've already force-pushed, and dropped that commit.

.github/workflows/ci.yml
```diff
@@ -181,7 +181,7 @@ jobs:
     - uses: coq-community/docker-coq-action@v1
       with:
         coq_version: ${{ env.coq-version-supported }}
-        ocaml_version: 4.11-flambda
+        ocaml_version: ${{ env.ocaml-version }}
```
Alizter, line 184: What I mean is that it's not clear we want to tie this version to the main OCaml version. This simply exists to get Alectryon to work so it might be better to keep it independent. If it no longer works we can fix it, but this would be easier to fix than wondering why bumping the OCaml version has broken Alectryon, when in fact it's because it doesn't support a newer version of OCaml.

Alizter:
That doesn't leave anything left in this PR, but as I said above, I still think that the whitespace tab changes you did before are fine, I must have over indented this file when I first wrote it.

artagnon:
Sorry for the confusion! Dropping.

PR #1637, opened 2022-03-01
The Baer sum of short exact sequences

jarlg:
We define pullbacks and pushouts of short exact sequences as required for the Baer sum, which we also define. For now we only define the function underlying the Baer sum, but in the future we'd like to use it to make Ext into an abelian group. Apart from the first commit (which can be mostly ignored) these are mostly additions, so the unified view should be fine for reviewing.

In the first commit I renamed `AbExt.v` to `AbSES.v` (not sure why git doesn't realise I only renamed the file), as it is interesting to study the space `AbSES` on its own. Currently `AbExt` is defined at the end of the file, but it might be split out later. In a previous PR the terminology wasn't great (`AbSES` was named `hExt`), but now I've settled on calling the untruncated space `AbSES` and the set-truncation for `Ext`. Accordingly, an extension is an isomorphism class of short exact sequences.

Lastly, I've defined pushouts of short exact sequences using `ab_pushout`, which uses set-truncation. However these pushouts are already truncated, since they are along (-1)-truncated maps. This follows from [this](https://github.com/HoTT/HoTT-Agda/blob/master/theorems/stash/modalities/gbm/PushoutMono.agda) work by Eric Finster in Agda. Just mentioning this, since it is not immediately useful for what I'd like to do.


theories/HIT/epi.v
```diff
@@ -36,6 +39,20 @@ Proof.
     exact (ap pr1 ((contr (g; 1))^ @ contr xy)). }
 Defined.
 
+Definition equiv_isespi_isepi_funext {X Y : Type} (f : X -> Y)
```
Alizter, line 42: This entire file needs a cleanup IMO. Univalence is assumed in this section but this lemma really only needs funext, so it could be pushing down univalence requirements when they are not needed.

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,468 @@
...
+Arguments abses_i {B A} E : rename.
+Arguments abses_p {B A} E : rename.
+Arguments abses_isembedding_i {B A} E : rename.
+Arguments abses_issurjection_p {B A} E : rename.
+Arguments abses_isexact {B A} E : rename.
+
+Arguments Build_AbSES {B A} _ _ _ _ _ _.
+
+(** TODO Figure out why printing this term eats memory and seems to never finish. *)
+Local Definition issig_abses_do_not_print {B A : AbGroup} : _ <~> AbSES B A := ltac:(issig).
```
Alizter, line 38: Is this happening in PG? I will look into it.

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,468 @@
...
+
+(** * Short exact sequences of abelian groups *)
+
+(** A short exact sequence of abelian groups consists of a monomorphism [i : A -> E] and an epimorphism [p : E -> B] such that the image of [i] equals the kernel of [p]. Later we will consider short exact sequences up to isomorphism by 0-truncating the type [AbSES] defined below. An isomorphism class of short exact sequences is called an extension. *)
+
+(** The type of short exact sequences [A -> E -> B] of abelian groups. *)
+Record AbSES (B A : AbGroup) : Type := {
+  abses_E : AbGroup;
+  abses_i : A $-> abses_E;
+  abses_p : abses_E $-> B;
```
Alizter, line 17: For now the name is probably OK, but it is not very descriptive. I wonder if it would be easier to understand with start, middle and end?

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,468 @@
...
+
+(** ** Morphisms of short exact sequences *)
+
+Record AbSESMor {A X B Y : AbGroup} (E : AbSES B A) (F : AbSES Y X) := {
+    absesm1 : A $-> X;
+    absesm2 : (abses_E E) $-> F;
+    absesm3 : B $-> Y;
+    absesm12 : (abses_i _) $o absesm1 == absesm2 $o (abses_i _);
+    absesm23 : (abses_p _) $o absesm2 == absesm3 $o (abses_p _);
+}.
```
Alizter, line 106: The names here aren't great either, but I don't have a better suggestion.

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,468 @@
...
+
+Arguments Build_AbSESMor {_ _ _ _ _ _} _ _ _ _ _.
+
+Arguments absesm1 {_ _ _ _ E F} _.
+Arguments absesm2 {_ _ _ _ E F} _.
+Arguments absesm3 {_ _ _ _ E F} _.
+Arguments absesm12 {_ _ _ _ E F} _.
+Arguments absesm23 {_ _ _ _ E F} _.
```
Alizter, line 114: Why not make E and F implicit for the entire record, and then `Arguments AbSESMor {A X B Y} E F : clear implicits.` instead?

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,468 @@
...
+Proof.
+  snrapply (Build_AbSESMor phi _ grp_homo_id).
+  - exact ab_pushout_inr.
+  - exact ab_pushout_commsq.
+  - rapply ab_pushout_rec_beta_right.
+Defined.
+
+(** *** The universal property of [abses_pushout_morphism]
+
+Any map [f : E -> F] of short exact sequences factors (uniquely) through [abses_pushout E f1]. *)
```
Alizter, line 369: You probably want to end the top comment and start a new one. In the HTML documentation, this entire thing will be a header otherwise.

Alizter (commented):
Some initial comments

theories/HIT/epi.v
jarlg: I agree the file needs a cleanup (and I see that I even have a typo "isespi" there now which I'll fix). In my usage univalence is needed for other reasons (the pushout of an embedding being an embedding).

theories/Algebra/AbGroups/AbSES.v
jarlg: Yes, it's happening in PG. There was some discussion on it in a previous PR: #1534.

theories/Algebra/AbGroups/AbSES.v
jarlg: I'm open to better naming suggestions, but it's short and mirrors what I usually write on paper (i being the inclusion, p being the projection, E being in the middle).

theories/Algebra/AbGroups/AbSES.v
jarlg: Open to suggestions here as well, though I think these names are pretty self-explanatory and not-too-hard to remember.

theories/Algebra/AbGroups/AbSES.v
jarlg: Coq didn't like the `clear implicits` when there are some implicit arguments, but simply doing `Arguments AbSESMor {A X B Y} E F.` seems to work just fine. Thanks!

theories/Algebra/AbGroups/AbSES.v
jarlg: Okay, will do!

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,463 @@
...
+Definition abses_direct_sum `{Univalence} {B A : AbGroup} (E F : AbSES B A)
+  : AbSES (ab_biprod B B) (ab_biprod A A)
+  := Build_AbSES (ab_biprod E F)
+                 (functor_ab_biprod (abses_i E) (abses_i F))
+                 (functor_ab_biprod (abses_p E) (abses_p F))
+                 (functor_ab_biprod_embedding _ _)
+                 (functor_ab_biprod_sujection _ _)
+                 (ab_biprod_exact _ _ _ _).
+
+(** The Baer sum of two short exact sequences is obtained from the pointwise direct sum by pulling back along the diagonal map and pushing forth along the codiagonal. (Changing the order of pulling back and pushing forth produces an isomorphic short exact sequence.) *)
```
Alizter, line 440: Shoudn't it be pushing forth then pulling back given how you've defined it?

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -0,0 +1,463 @@
...
+Definition abses_baer_sum `{Univalence} {B A : AbGroup} (E F : AbSES B A)
+  : AbSES B A
+  := abses_pullback
+       (abses_pushout (abses_direct_sum E F) ab_codiagonal)
+       ab_diagonal.
+
+
+(** * The set [Ext B A] of abelian group extensions *)
+
+Definition Ext (B A : AbGroup) := Tr 0 (AbSES B A).
```
Alizter, line 450: So AbSES is a 1-type, but you will have non-trivial automorphisms of the midpoint E which is why you have to set truncate?

jarlg: Exactly, and this is how Yoneda Ext is defined classically.

Alizter: I wonder if the definition could be changed to use an isomorphism class (0-truncation of AbGroup akin to cardinalities) of groups for E, that way you wouldn't need to truncate at the end. I think it might be unlikely since you would also have to have the maps up to an unspecified is, etc.

jarlg: Not sure how that would work. The truncation here is very much intentional, and entirely analogous to the classical construction. Maybe I should add that after showing that the Baer sum of short exact sequences defined in this PR is unital, commutative, and associative all up to isomorphism, we get an actual abelian group structure on `Ext B A`.

Alizter: If you go down the route of showing the Abelian group structure, then I think it would make sense to define `Ext` directly as an Abelian group.

theories/Algebra/AbGroups/AbSES.v
mikeshulman: I don't think single-letter names should be codified.  Someone else might prefer to write j and q, or f and g, or h and k, etc.

mikeshulman: In `Factorization` we used `intermediate`, `factor1`, and `factor2`.

theories/Algebra/AbGroups/AbSES.v
mikeshulman: I think it's correct to have a 1-type of short exact sequences and define extensions to be its 0-truncation.

theories/Algebra/AbGroups/AbSES.v
jarlg: Alright, I'll figure out something better!

theories/Algebra/AbGroups/AbSES.v
jdchristensen: Just to emphasize, it doesn't work to make `E` and isomorphism class, since you need the maps to a from it.  So what Jarl has done is best.  I think this conversation can be resolved.

Alizter: I agree.

theories/Limits/Pullback.v
```diff
@@ -200,6 +200,17 @@ Proof.
   make_equiv_contr_basedpaths.
 Defined.
 
+(** A version where [g] is pointed, but we unbundle the pointed conidition to avoid importing pointed types. *)
```
jdchristensen, line 203:
```suggestion
(** A version where [g] is pointed, but we unbundle the pointed condition to avoid importing pointed types. *)
```

theories/Algebra/AbGroups/AbSES.v
jarlg: Okay, how about the following? 

```coq
Record AbSES (B A : AbGroup) : Type := {
  middle :  AbGroup;
  inclusion : A $-> middle;
  projection : middle $-> B;
  isembedding_inclusion : IsEmbedding inclusion;
  issurjection_projection : IsSurjection projection;
  isexact_inclusion_projection : IsExact (Tr (-1)) inclusion projection;
}.
```
I originally thought we should qualify the names (i.e. put `abses_` in front), but that's automatic since you write e.g. `AbSES.inclusion`. One often calls the middle group for the extension, but I want to reserve that word for isomorphism classes of short exact sequence, i.e. the elements of `Ext B A`.

theories/Algebra/AbGroups/AbSES.v
jarlg: How about this?
```coq
Record AbSESMor {A X B Y : AbGroup} {E : AbSES B A} {F : AbSES Y X} := {
    component1 : A $-> X;
    component2 : (middle E) $-> F;
    component3 : B $-> Y;
    left_square : (inclusion _) $o component1 == component2 $o (inclusion _);
    right_square : (projection _) $o component2 == component3 $o (projection _);
}.
```
A morphism is a natural transformation between the diagrams defining the short exact sequences, and we often refer to the maps of a natural transformation as the "components." `left_square` and `right_square` is vague out of context, but maybe better than `component12` and `component23`?

Alizter: That's better. If you like you can leave a comment, but I think it is clear.

On another note, what confused me at first is that `F` is being coerced to `middle F`. Perhaps it would be better to be explicit here?

Alizter: If you write `E $-> F` coq will start looking for wild cat instances for `AbSES` which don't exist. It isn't clever enough to insert a coercion and then do typeclass search since that might be a bit expensive. If you however write `middle E $-> F` then coq knows that you are looking for `AbGroup` instances already, so when it typechecks `F` it expects an AbGroup. This is where it can insert the coercion to `AbGroup`.

I don't suggest getting rid of the coercion at all, just writing `middle E $-> middle F` ought to be enough, for clarity of the reader.

theories/Algebra/AbGroups/AbSES.v
jarlg: I've pushed this suggestion, as well as improved naming in `AbSESMorphism`.

Alizter (approved):
LGTM. If you go down the route of defining Ext to be an abelian group directly (I don't know if you want to do it in this PR), then you can get rid of the IsPointed instance, but showing the Baer sum satisfies all the group laws might be a bit tedious. (Perhaps we can automate this somewhat?).

jarlg:
I won't make Ext into an abelian group in this PR, as that still requires quite a bit more work. It seems unlikely that much of it can be automated, as it requires proving non-trivial facts relating pushouts and pullbacks of short exact sequences. Of course, more/better automation for working with equations in groups would be helpful and welcome! Good point about `IsPointed`, I'll keep that in mind.

jarlg:
Ext^n is the n-th right derived functor of Hom. We might define Ext^0 = Hom to make certain formulas look nice, but I agree that considering Ext^0(B,A) to be isomorphisms A -> B could also make sense.

I previously tried working with length-n exact sequences using `Spaces.Finite`, but the terms quickly started exploding. Perhaps `FinSeq` is better, but there's also a relation on length-n sequences that's not being captured by the type. Getting the correct space of length-n sequences for general n runs into the usual coherence problem. Despite that, we can define Ext^n by splicing length-1 extensions and putting in the correct relations. There will be a map from length-n exact sequences to Ext^n, but we won't actually define the latter as a (set-)quotient of the former.

jdchristensen:
Ext^0 is Hom, not isos, and I guess this just reflects the fact that only in the case of n=1 can you identify elements of Ext^n with isomorphism classes of exact sequences.  It's annoying that equality on the type of length n+2 exact sequences doesn't automatically produce the correct relation, but that's how it is...

I'm going to review this PR today, so don't merge yet.

Alizter:
So for Ext^2, does the path type have more equivalences than just the automorphisms of the two centre terms? What do those look like exactly?

jarlg:
The congruence is described in section III.5 "Composition of extensions" in Mac Lane's _Homology_. Essentially, you can replace isomorphic extensions, and move a pullback to a pushout as in equation (5.2).

jdchristensen:
As Jarl said, the short way to say it is that it is the equivalence relation generated by having any map of extensions (not necessarily an isomorphism) but which is the identity map on the two ends.  (This works for all n, not just n=2.)  But Mac Lane also gives a concrete description of specific rules that generate the same equivalence relation.

theories/Algebra/AbGroups/AbPushout.v
```diff
@@ -102,3 +135,35 @@ Proof.
...
 Defined.
+
+Definition path_ab_pushout `{Univalence} {A B C : AbGroup} (f : A $-> B) (g : A $-> C)
+           (bc0 bc1 : ab_biprod B C)
+  : @in_cosetL (ab_biprod B C)
+               (grp_image (ab_biprod_corec f (g $o ab_homo_negation))) bc0 bc1
+               <~> (grp_quotient_map bc0 = grp_quotient_map bc1 :> ab_pushout f g).
+Proof.
+  rapply path_quotient.
+Defined.
```
jdchristensen, line 146: The above seems like a technical result that is hard to parse.  The one place it is used can be replaced with `rapply path_quotient`, so maybe it is better to remove this?

jdchristensen, line 146: The statement is hard to parse.  Can it be made more readable?  I noticed that `(grp_image (ab_biprod_corec f (g $o ab_homo_negation)))` appears a few times in the file, so it could be given a name.  Or maybe the relation itself should be given a name?

jdchristensen (approved):
LGTM!

theories/Algebra/AbGroups/AbSES.v
```diff
@@ -145,59 +174,269 @@ Proof.
...
+      (** Pick a preimage under the quotient map. *)
+      assert (bc : merely (hfiber grp_quotient_map bc'))
+        by apply issurj_class_of.
+      strip_truncations.
+      destruct bc as [[b c] q].
+      (** The E-component of the preimage is in the kernel of [projection E]. *)
+      assert (c_in_kernel : (projection E) c = mon_unit).
+      1: { refine (_ @ p); symmetry.
+           rewrite <- q.
+           refine (ab_pushout_rec_quotient_map_beta _ @ _); cbn.
```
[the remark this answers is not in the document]
jarlg: I agree that the unnecessary reflexivity path should be avoided. What I dislike is that this relies on the magic of `simpl`; if you put `cbn` then the goal blows up and reflexivity doesn't work. The lemma is the most robust solution, as I see it. Is there a way to have the lemma but avoid the unnecessary reflexivity path?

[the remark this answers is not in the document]
Alizter: Perhaps a more concrete solution would be to use `change`? You can `change` your goal to a definitionally equal goal (which coq checks by unifying). This has the advantage of being quicker and more precise than expecting/hoping for a result from simpl or cbn. Though on the other hand, it is extremely unlikely the behaviour of simpl will change dramatically in future versions of Coq.

theories/Algebra/AbGroups/AbPushout.v
jarlg: I'm naming it `ab_pushout_subgroup` and adding a comment about it. This then becomes:
```coq
Definition path_ab_pushout `{Univalence} {A B C : AbGroup} (f : A $-> B) (g : A $-> C)
           (bc0 bc1 : ab_biprod B C)
  : @in_cosetL (ab_biprod B C) (ab_pushout_subgroup f g) bc0 bc1
               <~> (grp_quotient_map bc0 = grp_quotient_map bc1 :> ab_pushout f g).
```
Is that good enough? If we make `GroupHomomorphism A B` into a group itself, then we could write write `grp_image (f - g)` for the subgroup.

theories/Algebra/AbGroups/AbSES.v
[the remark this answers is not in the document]
jdchristensen: Either of `rapply left_identity` or `exact (left_identity _)` work without the `simpl`.  But I prefer the simpl, since it makes it clear to someone stepping through the proof what happened, and it requires no hardcoding like `change` would.

theories/Algebra/AbGroups/AbPushout.v
jdchristensen: Sure, that seems better.  I also like the last suggestion.  In my Hurewicz work, I defined `AbHom` for the abelian group of group homomorphisms between abelian groups.  (Only the target needs to be abelian, and that turns out to be useful.)  It's near the top of https://github.com/jdchristensen/HoTT/blob/Hurewicz/theories/Hurewicz/Smashing.v

theories/Algebra/AbGroups/AbSES.v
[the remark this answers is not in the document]
jarlg: You're right. But we'll stick with `simpl` then.

theories/Algebra/AbGroups/AbPushout.v
jarlg: Thanks for the link. We should do this at some point, but seamlessly enriching the hom so that what I suggested works will require some thought. I'm going with `ab_pushout_subgroup` for now.

Alizter:
Good to merge?

jarlg:
Yep, I have no further changes to make.

Alizter:
Alright, I see the tricky part with defining Ext^n. For Ext^2 we can ask that the splice of SES can move pushouts to pullbacks and vice versa to and fro. For Ext^n you will have to do this for each successive pair of spliced SES which is going to be tricky.

I wonder if it would make sense to define the relation successively, so that pushouts and pullbacks along a map are defined seperately for each n, then for n+1 your n+1-relation needs to ask for these pushouts and pullbacks to move around on splicing. Have you tried something like this to overcome the coherence problem?

If none of that makes any sense I can go into more detail.

jarlg:
From Ext^n you define Ext^{n+1} as pairs (E,F) where E is a length n extension, and F is a length 1 extension (i.e. a short exact sequence), and quotient as you say. So defining Ext^n isn't tricky; what's tricky is to define the (untruncated) space of length n exact sequences ES^n such that Ext^n is the 0-truncation of ES^n. In general there isn't a coherence problem in doing things with extensions, as they live in a set. It's working with the space of length n exact sequences that we don't know how to do, because of the coherence involved.

PR #1638, opened 2022-03-01
Remove uses of instantiate (to fix some warnings)

Alizter:
It has been deprecated in 8.16 (alpha) and has the behaviour of idtac. Therefore it can be safely removed.

### DIFF
--- a/theories/Categories/Adjoint/Composition/Core.v
+++ b/theories/Categories/Adjoint/Composition/Core.v
@@ -63,10 +63,10 @@ Section compose.
             | _ => rewrite <- ?composition_of, unit_counit_equation_2
             | [ A : _ -| _ |- _ = 1%morphism ]
               => (etransitivity; [ | apply (unit_counit_equation_1 A) ];
-                  instantiate; try_associativity_quick f_ap)
+                  try_associativity_quick f_ap)
             | [ A : _ -| _ |- _ = 1%morphism ]
               => (etransitivity; [ | apply (unit_counit_equation_2 A) ];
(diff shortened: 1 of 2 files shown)


mikeshulman (approved):
What did it used to do?

(Ugh, that's one of those sentences that sounds fine said out loud but looks totally wrong when committed to paper.)

Alizter:
When you have existential variables as shelved goals, it used to be the case that a tactic:
```coq
tac.
```
would try to solve those goals if `tac` gave any hints how to do it. This meant that in the middle of semi-coloned tactics
```
tac; tac2; tac3.
```
it would not do that, so you would have to manually `instantiate` those shelved goals.

Now (since at least 8.15 I think?) it does that by default between semi-colons.

JasonGross:
@Alizter the behavior you're describing is about constraint solving (c.f. [`solve_constraints`](https://coq.inria.fr/refman/proofs/writing-proofs/proof-mode.html#coq:tacn.solve_constraints)), and it still occurs only at periods.  Furthermore, the instantiation of evars via `instantiated (num := term)` is still available and not deprecated.

In fact the thing argumentless `instantantiate` used to do was evar normalization, which has become obsolete since EConstr.  It used to be useful when a tactic would instantiate an evar by side effect in one goal, and this instantiation was needed in another goal by a tactic which was evar-sensitive.

Alizter:
@JasonGross Right I got confused. There is a valid use of `instantiate (_ := _)` in Tests.v for example.

PR #1639, opened 2022-03-03
mangle autogenerated names

Alizter:
Here is something that we might want to do.

`Set Mangle Names` gives a mangled version of an autogenerated name in order to check if something relies on an autogenerated name which is fragile behaviour. `Set Mangle Names Light` allows for a human readable version of this where an underscore prefix is added to the autogenerated name to make it slightly different. We could also configure the prefix from `"_"` to something like `"_Name_is_autogenerated_do_not_use_"` to be horrendously verbose. 

Anyway setting this showed that the library was almost completely not relying on autogenerated names. The only place was an arguments command which looked like it was in the wrong place so I moved and fixed it.


Alizter:
Since everything will become a de Bruijn index behind the scenes, mangle names simply pokes all possible alpha conversions. This does end up changing goals as you have observed. I can drop the commit enabling the option, but this was just something I wanted to show off really.

jdchristensen:
I'm impressed that it worked!  I sometimes forget to make things robust to name choices.  We could consider adding the option to one of the CI runs.

mikeshulman:
I like the idea of adding it to CI run but not including it in the regular library.

Alizter:
There is in fact a command line parameter `-mangle-names` which can be passed to coqc so that we only mangle on compilation. 

We could even do:

```coq
Set Mangle Names Light.
Set Mangle Names Prefix "_Please_do_not_rely_on_autogenerated_names_".
```

so that when `-mangle-names` is passed in the makefile, it would give a clear error message and be less confusing than printing a random de bruijn index. However `-mangle-names` would not be passed in `_CoqProject` so it wouldn't affect interactive developments.

mikeshulman:
Sounds good.

Alizter:
So I've changed this line of the makefile:
```
	$(COQBIN)coq_makefile -f _CoqProject -o Makefile.coq -arg -mangle-names -arg "_Please_do_not_rely_on_autogenerated_names_"
```

This means there will be no name mangling during interactive development, and only during compilation. I will push an extra commit to demonstrate what happens when you rely on autogenerated names.

SkySkimmer:
>So I've changed this line of the makefile:

You'll stll see mangled names interactively when looking at names generated in another file.

IMO mangled names should be enabled in a new CI job, not added to an existing one.

SkySkimmer:
>so that when -mangle-names is passed in the makefile, it would give a clear error message and be less confusing than printing a random de bruijn index. 

Unless the error message is "H not found"

Alizter:
In fact it seems to be worse, it looks like before the option wasn't being exported before since I forgot to put a `Global`. Looks like we are relying on autogenerated names all over the place (`IHn` being very common). I will draft this PR for now and possibly fix the autogenerated names in the future.

PR #1646, opened 2022-04-27
Fiber sequences of short exact sequences induced by pullback

jarlg:
We prove that a short exact sequence `B -> E -> C` of abelian groups induces a fiber sequence `AbSES C A -> AbSES E A -> AbSES B A` by pulling back (`isexact_abses_pullback`). Our strategy is to prove the analog of exactness on _path data_ and deduce actual exactness from that. The reason being that path data consists of functions which compute, and this makes the "path algebra" manageable. (Recall that path data between two short exact sequences are morphisms between their middles respecting the inclusions and projections.)

We also show that `loops (AbSES B A)` is the set of group homomorphisms `B -> A` (`loops_abses`). Thus the long exact sequence of homotopy groups associated to the fiber sequence above recovers the usual six-term exact sequence of Ext groups — though we have not yet formalised the details of this.

I've separated the commits into logical chunks, with the 3rd, second to last, and last commits being the most interesting ones. There are mostly additions so the "unified" view ought to be fine for reviewing.

There is one kink I hope we can work out: the `Defined.` on line 248 of `Algebra/AbGroups/AbSES/PullbackFiberSequence.v` takes about 24s on my laptop. Any suggestions for how to speed it up? (If no good suggestions come up, I might try redefining `hfiber_cxfib'` and friends as records. But I'm hoping there's some smaller tweak that can be done.)


theories/Algebra/AbGroups/AbSES/PullbackFiberSequence.v
```diff
@@ -0,0 +1,366 @@
...
+      refine (grp_homo_op _ _ _ @ _).
+      apply grp_moveR_Mg.
+      refine (_ @ (left_inverse _)^).
+      apply qglue.
+      exists b.
+      refine (_ @ (grp_unit_r _)^).
+      exact (negate_involutive _)^.
+Defined.
+
+(** That [abses_pullback_trivial_preimage E F p] pulls back to [F] is immediate [abses_component1_trivial_pullback] and the following map: *)
```
jdchristensen, line 120: Change "immediate" to "immediate from".  Also, maybe this would be a good spot to say that this is proving part of the exactness of Ext^1 at the middle spot?

jdchristensen (commented):
Wow, thanks for this huge contribution!  I skimmed it over, and think it looks good.

I made a few minor comments.  And I couldn't quickly see how to speed up the slow parts, but I might take another look later.

Alizter:
@jarlg If you rebase the CI should (mostly) work.

theories/Algebra/AbGroups/AbSES/Pullback.v
```diff
@@ -94,3 +94,241 @@ Definition abses_component1_trivial_pullback `{Univalence} {A B B' : AbGroup}
...
+  - apply equiv_grp_pullback_compose_r.
+  - intro a.
+    by srapply equiv_path_pullback_hset.
+  - reflexivity.
+Defined.
+
+(** Composition of pullback morphisms pointed with path data. *)
+Definition pmap_abses_pullback_compose {A B0 B1 B2 : AbGroup}
+           (f : B0 $-> B1) (g : B1 $-> B2)
+  : pmap_abses (AbSES B2 A) B0 A.
```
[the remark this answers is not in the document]
jarlg: The action of a general map in  `pmap_abses` on path data uses univalence (as far as I can tell). So the pointedness of an arbitrary composite in`pmap_abses` will be path data which doesn't compute (since it will use univalence). Only for specific maps, such as pullbacks, can I describe the action on path data in a way that computes, and thereby make sure that the path data witnessing pointedness of their composite also computes.

So I agree that composition can be defined more generally, but I don't see how to do it while maintaining control/computability of the witnesses of pointedness.

jarlg:
I defined base point preserving functors and pointed transformations between them, which specialise to the pullback functors and their "path data homotopies" I had earlier. Like Dan suggsted above, `to_pointed` preserves composition in this generality.

theories/Algebra/AbGroups/AbSES/Pullback.v
```diff
@@ -223,39 +210,21 @@ Definition abses_pullback_compose `{Univalence} {A B0 B1 B2 : AbGroup}
 
 Definition abses_pullback_pcompose' `{Funext} {B0 B1 B2 A : AbGroup}
```
jdchristensen, line 211: `Funext` can be removed here.  I didn't check if it can be removed anywhere else.

jdchristensen (approved):
Nice!  Not only is this cleaner, it makes several things faster.  For example, the really slow `Defined` at the end of `hfiber_cxfib'_inhabited` in PullbackFiberSequence.v went from 11.7s to 1.5s on my machine.  It still takes about 15s to build the six files in the AbSES directory, so there's room for improvement, but I don't think that should block merging.

theories/Algebra/AbGroups/AbSES/Pullback.v
jarlg: Thanks! This let me drop `Funext` in `abses_pullback_phomotopic'` as well.

jarlg:
I managed to shave off a couple of seconds in `AbSES/PullbackFiberSequence.v` by splitting out two lemmas and marking `abses_pullback'` opaque for one of them. Not sure why line 179 (`exact (@pr2_cxfib' _ A B C E U).`) isn't instant since it's supplying a term of the right type, but it's pretty fast now.

theories/Algebra/AbGroups/AbSES/PullbackFiberSequence.v
```diff
@@ -149,6 +150,37 @@ Definition hfiber_cxfib' `{Funext} {A B C : AbGroup} (E : AbSES C B)
...
+    with (equiv_path_abses_iso ((iscomplex_abses_pullback' _ _ (iscomplex_abses E)).1 U)).
+  apply (ap equiv_path_abses_iso).
+  rapply path_hom.
+  refine (_ $@R abses_pullback_compose' (inclusion E) (projection E) U);
+    unfold trans_comp.
+  refine (_ $@R abses_pullback_homotopic' (projection E $o inclusion E) grp_homo_const (iscomplex_abses E) U).
+  reflexivity.
+Defined.
+
+(** Making [abses_pullback'] transparent speeds up the following proof. *)
```
jdchristensen, line 168: Should say "opaque".

jdchristensen (commented):
These two files should be listed in AbSES.v:
```
Algebra/AbGroups/AbSES/Ext.v
Algebra/AbGroups/AbSES/PullbackFiberSequence.v
```
I'm working on some speedups.  Can I push them to your branch later today?

jarlg:
Sure. Feel free to add those lines in AbSES.v and correct "transparent" to "opaque". Or I can fix those afterwards. Thanks!

jdchristensen:
I addressed the two recent comments I made, and also sped things up a bit.  I didn't see any easy ways to address the remaining slow parts.  So I think this is ready to merge, once someone checks my two commits.

jarlg:
Looks good to me. Thanks! I also removed two "todo"s I forgot to remove earlier.

Alizter (approved):
Wildcat stuff is fine with me

jarlg:
Alright, thanks @Alizter! I think this is ready to merge then.

PR #1648, opened 2022-04-30
Monoidal cat

patrick-nicodemus:
We define a monoidal category.

### DIFF
--- a/theories/Categories/Monoidal/MonoidalCategory.v
+++ b/theories/Categories/Monoidal/MonoidalCategory.v
@@ -28,7 +28,6 @@ Section MonoidalCategoryConcepts.
   Local Notation "A ⊗ B" := (tensor (Datatypes.pair A B)) (at level 45, left associativity).
 
   Local Open Scope functor_scope.
-  Locate identity.
   Definition right_assoc := (tensor ∘ (Functor.Prod.pair 1 tensor) )%functor.
   Definition left_assoc :=  tensor ∘
                                    (Functor.Prod.pair tensor 1) ∘
--- a/theories/Categories/Monoidal/MonoidalCategory.v
+++ b/theories/Categories/Monoidal/MonoidalCategory.v
@@ -1,13 +1,13 @@
-From HoTT Require Import Basics Basics.Utf8 Basics.Tactics.
-From HoTT Require Import implementations.list.
-From HoTT Require Import Category.Core Category.Prod Category.Morphisms. 
-From HoTT Require Import NatCategory.
-From HoTT Require Import Functor.Core Functor.Identity Functor.Composition.Core Functor.Prod.Core.
-From HoTT Require Import NaturalTransformation.Core NaturalTransformation.Isomorphisms NaturalTransformation.Identity NaturalTransformation.Prod.
-From HoTT Require Import NaturalTransformation.Composition.Core.
-From HoTT Require Import FunctorCategory.Core FunctorCategory.Morphisms.
-From HoTT Require Import ProductLaws.
-From HoTT Require Import Cat.Core.
+Require Import Basics Basics.Utf8 Basics.Tactics.
+Require Import implementations.list.
+Require Import Category.Core Category.Prod Category.Morphisms.
+Require Import NatCategory.
+Require Import Functor.Core Functor.Identity Functor.Composition.Core Functor.Prod.Core.
+Require Import NaturalTransformation.Core NaturalTransformation.Isomorphisms NaturalTransformation.Identity NaturalTransformation.Prod.
+Require Import NaturalTransformation.Composition.Core.
+Require Import FunctorCategory.Core FunctorCategory.Morphisms.
+Require Import ProductLaws.
+Require Import Cat.Core.
 
 Set Universe Polymorphism.
 Set Implicit Arguments.
@@ -68,7 +68,7 @@ Section MonoidalCategoryConcepts.
     apply (morphism_of tensor); split; simpl.
     - exact (alpha_nat_trans (a,_)).
     - exact (Core.identity d).
-  Defined.    
+  Defined.
   Local Definition P4 : a ⊗ (b ⊗ (c ⊗ d)) --> (a ⊗ b) ⊗ (c ⊗ d)
     := alpha_nat_trans (a, (b, (c ⊗ d))).
   Local Definition P5 : (a ⊗ b) ⊗ (c ⊗ d) --> ((a ⊗ b) ⊗ c ) ⊗ d
@@ -82,7 +82,7 @@ Section MonoidalCategoryConcepts.
     - exact (Core.identity a).
     - exact (lambda_nat_trans _).
   Defined.
-  
+
   Local Definition Q2 : (a ⊗ (I ⊗ b)) --> a ⊗ b.
   Proof.
     refine (@Category.Core.compose _ _ ((a ⊗ I) ⊗ b) _ _ _).
@@ -95,7 +95,7 @@ Section MonoidalCategoryConcepts.
   End coherence_laws.
 End MonoidalCategoryConcepts.
 
-Class MonoidalStructure (C : PreCategory) := 
+Class MonoidalStructure (C : PreCategory) :=
   Build_MonoidalStructure {
     tensor : (C × C -> C)%category;
     I : C;
@@ -105,4 +105,3 @@ Class MonoidalStructure (C : PreCategory) :=
     pentagon_eq_holds : forall a b c d : C, pentagon_eq alpha  a b c d;
     triangle_eq_holds : forall a b : C, triangle_eq alpha lambda rho a b;
   }.
-  
--- a/theories/Categories/Monoidal/MonoidalCategory.v
+++ b/theories/Categories/Monoidal/MonoidalCategory.v
@@ -2,7 +2,8 @@ Require Import Basics Basics.Utf8 Basics.Tactics.
 Require Import implementations.list.
 Require Import Category.Core Category.Prod Category.Morphisms.
 Require Import NatCategory.
-Require Import Functor.Core Functor.Identity Functor.Composition.Core Functor.Prod.Core.
+Require Import Functor.Core Functor.Identity Functor.Composition.Core Functor.Prod.Core
+        Functor.Utf8.
 Require Import NaturalTransformation.Core NaturalTransformation.Isomorphisms NaturalTransformation.Identity NaturalTransformation.Prod.
 Require Import NaturalTransformation.Composition.Core.
 Require Import FunctorCategory.Core FunctorCategory.Morphisms.
@@ -14,32 +15,27 @@ Set Implicit Arguments.
 Generalizable All Variables.
 Set Asymmetric Patterns.
 
-Context `{Funext}.
-
-Local Infix "×" := Category.Prod.prod.
+Section MonoidalStructure.
 Local Notation id := Functor.Identity.identity.
 Local Notation "x --> y" := (morphism _ x y).
-Local Infix "∘" := Functor.Composition.Core.compose.
-Local Notation nat_trans_prod := NaturalTransformation.Prod.prod.
 Local Open Scope category_scope.
 Local Open Scope morphism_scope.
-
+Local Notation "g ∘ f" := (Functor.Composition.Core.compose g f).
+Context `{Funext}.
 Section MonoidalCategoryConcepts.
   Variable C : PreCategory.
-  Variable tensor : ((C × C) -> C)%category.
+  Variable tensor : ((C * C) -> C)%category.
   Variable I : C.
-
-  Local Notation "A ⊗ B" := (tensor (A, B)) (at level 45).
-
+  Local Notation "A ⊗ B" := (tensor (Datatypes.pair A B)) (at level 45, left associativity).
   Definition right_assoc := tensor ∘ (Functor.Prod.pair (id C) tensor).
   Definition left_assoc :=  tensor ∘
                                    (Functor.Prod.pair tensor (id C)) ∘
                                    (Associativity.functor _ _ _).
   Definition associator := NaturalIsomorphism right_assoc left_assoc.
   (* Orientation  (A ⊗ B) ⊗ C -> A ⊗ (B ⊗ C) *)
-  Definition pretensor (A : C) := Functor.Prod.Core.induced_snd tensor A.
+  Definition pretensor (A : C) := Core.induced_snd tensor A.
   Definition I_pretensor := pretensor I.
-  Definition posttensor (A : C) := Functor.Prod.Core.induced_fst tensor A.
+  Definition posttensor (A : C) := Core.induced_fst tensor A.
   Definition I_posttensor := posttensor I.
   Definition left_unitor := NaturalIsomorphism I_pretensor (id C).
   Definition right_unitor := NaturalIsomorphism I_posttensor (id C).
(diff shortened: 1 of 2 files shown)


patrick-nicodemus:
@Alizter I am not experienced with git. Do you have a suggestion for how to resolve this issue?

Alizter:
Git merge has issues especially when submodules have been changed as in #1647. I suggest using a rebase, which more can be found here: https://www.atlassian.com/git/tutorials/rewriting-history/git-rebase.

If you do an interactive rebase `git rebase upstream/master` it should also get rid of those merge commits. You may need to do a `git fetch upstream` first though.

patrick-nicodemus:
I gave it another try but I might have screwed it up, I am not confident that I've done the rebase correctly. We will see when the CI runs.
(Does the CI run automatically on a regular basis or does it have to be triggered?)

Alizter:
@patrick-nicodemus Would you like me to force push to your branch directly so that I can clean up the history? I'll also mention what I am doing when I do it.

Alizter:
By the way, since you are a first time contributor, GitHub requests a manual acceptance to run the CI. But after this contribution, the CI should automatically run for every PR.

patrick-nicodemus:
> @patrick-nicodemus Would you like me to force push to your branch directly so that I can clean up the history? I'll also mention what I am doing when I do it.

Yeah, please. You can override anything that's in there, I have my own copies of the files.

I believe I somehow have an incorrect version of the .gitmodules file, I did a rebase on my local machine but the fork hosted on github may have problems with the file. I might do a hard reset of the github repo as well.

Alizter:
@patrick-nicodemus Hi, in order for me to push to your branch, you need to select a check box in the bottom right corner of this PR which says: allow commits from other maintainers or something similar like that.

I have a version of your changes with the commit history cleaned up. Unfortunately, multiple things seem to have gone wrong when you were rebasing/merging which caused this mess to appear. Nothing to worry about however!

Let me know when you have done that then I can push to your branch. :-)

patrick-nicodemus:
image
Hi @Alizter ;
I've checked the given checkbox. I hope this is what was intended. Otherwise let me know and i'll look around for the right one.
I appreciate your patience in dealing with my lack of background with Git.

patrick-nicodemus:
I don't think it would be too bad for me to simply nuke the fork and start over with a new one if this becomes too much of a hassle.

Alizter:
@patrick-nicodemus I force pushed. Let me know if everything looks correct.

Alizter:
By the way, if you want to update your local branch to this one, you will have to do
```sh
git reset --hard origin/monoidal-cat
```
which will reset your commit on your local monoidal-cat branch to the one on your fork (which I force pushed too).

Sorry that this has been so complicated! It is usually not supposed to happen the first time :)

patrick-nicodemus:
@Alizter Thank you. I think the version of the commit that got pushed still has the git annotations in it though, so it's failing the build request. I can edit this on my end, hopefully?

patrick-nicodemus:
> By the way, if you want to update your local branch to this one, you will have to do
> 
> ```shell
> git reset --hard origin/monoidal-cat
> ```
> 
> which will reset your commit on your local monoidal-cat branch to the one on your fork (which I force pushed too).
> 
> Sorry that this has been so complicated! It is usually not supposed to happen the first time :)

Ok I missed this, sorry, I should have waited a minute.

patrick-nicodemus:
> By the way, if you want to update your local branch to this one, you will have to do
> 
> ```shell
> git reset --hard origin/monoidal-cat
> ```
> 
> which will reset your commit on your local monoidal-cat branch to the one on your fork (which I force pushed too).
> 
> Sorry that this has been so complicated! It is usually not supposed to happen the first time :)

So, I didn't read this comment before leaping into action and I did an ordinary push/pull between my local repo and the one hosted here. Is this going to screw things up?

Also, I understand how to enter the various commands you're talking about into my terminal to do a hard reset on my local repo or something like this, but I am a bit confused about how to get similarly fine grained control over the Github fork. I see basic pull and push options on the GUI but I'm clicking around on the website and I don't see the hard reset option. Googling around for "github hard reset" only takes me to git documentation on the CLI, which isn't what I want here.

Alizter:
When you force push a branch to a github repo, you will do a hard reset.

Alizter:
I force pushed the same branch I had again (which ended up dropping your extra commits) so it should all be fixed now.

Alizter:
I've gone ahead and removed the froms and extra whitespace too, at least it the CI will run now.

mikeshulman:
This looks fine to me, although I'm not an expert on the Categories sublibrary.  @JasonGross any thoughts?  I guess one question is whether we want to continue adding to the Categories sublibrary, since we're moving towards using the WildCat sublibrary in the rest of the library.

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+Require Import FunctorCategory.Core FunctorCategory.Morphisms.
+Require Import ProductLaws.
+Require Import Cat.Core.
+
+Set Universe Polymorphism.
+Set Implicit Arguments.
+Generalizable All Variables.
+Set Asymmetric Patterns.
+
+Context `{Funext}.
```
JasonGross, line 17: This makes a global funext axiom, you should instead wrap everything in a section

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+
+Local Infix "×" := Category.Prod.prod.
+Local Notation id := Functor.Identity.identity.
+Local Notation "x --> y" := (morphism _ x y).
+Local Infix "∘" := Functor.Composition.Core.compose.
```
JasonGross, line 22: I think we already have a Utf8 file that defines this?

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+Local Notation nat_trans_prod := NaturalTransformation.Prod.prod.
+Local Open Scope category_scope.
+Local Open Scope morphism_scope.
+
+Section MonoidalCategoryConcepts.
+  Variable C : PreCategory.
+  Variable tensor : ((C × C) -> C)%category.
+  Variable I : C.
+
+  Local Notation "A ⊗ B" := (tensor (A, B)) (at level 45).
```
JasonGross, line 32: It might be good to reserve this infix notation in the file that reserves all the notations

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+End MonoidalCategoryConcepts.
+
+Class MonoidalStructure (C : PreCategory) :=
+  Build_MonoidalStructure {
+    tensor : (C × C -> C)%category;
+    I : C;
+    alpha : associator tensor;
+    lambda : left_unitor tensor I;
+    rho : right_unitor tensor I;
+    pentagon_eq_holds : forall a b c d : C, pentagon_eq alpha  a b c d;
```
JasonGross, line 105:
```suggestion
    pentagon_eq_holds : forall a b c d : C, pentagon_eq alpha a b c d;
```

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+
+  Local Notation "A ⊗ B" := (tensor (A, B)) (at level 45).
+
+  Definition right_assoc := tensor ∘ (Functor.Prod.pair (id C) tensor).
+  Definition left_assoc :=  tensor ∘
+                                   (Functor.Prod.pair tensor (id C)) ∘
+                                   (Associativity.functor _ _ _).
+  Definition associator := NaturalIsomorphism right_assoc left_assoc.
+  (* Orientation  (A ⊗ B) ⊗ C -> A ⊗ (B ⊗ C) *)
+  Definition pretensor (A : C) := Functor.Prod.Core.induced_snd tensor A.
```
JasonGross, line 40: You could just use `induced_snd tensor A`, I think.  Any reason for qualifying it?

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,107 @@
...
+  Definition I_posttensor := posttensor I.
+  Definition left_unitor := NaturalIsomorphism I_pretensor (id C).
+  Definition right_unitor := NaturalIsomorphism I_posttensor (id C).
+
+  Variable alpha : associator.
+  Variable lambda : left_unitor.
+  Variable rho : right_unitor.
+  Notation alpha_nat_trans := ((@morphism_isomorphic
+                                  (C × (C × C) -> C)%category right_assoc left_assoc) alpha).
+  Notation lambda_nat_trans := ((@morphism_isomorphic _ _ _) lambda).
```
JasonGross, line 52: You don't need the extra parentheses here.  Any reason to keep them?

JasonGross (changes requested):
Thanks for the contribution!

The funext needs to be moved to a section, everything else is just a suggestion.

I'm also curious how long this file takes to compile.  I seen to recall monoidal categories being slow when I tried them

patrick-nicodemus:
> This looks fine to me, although I'm not an expert on the Categories sublibrary. @JasonGross any thoughts? I guess one question is whether we want to continue adding to the Categories sublibrary, since we're moving towards using the WildCat sublibrary in the rest of the library.

Can you tell me a little bit about this shift? I will do whatever you folks want, I wasn't aware that such a shift was underway or intended.

I would like to know if you can provide links to some discussion of the purpose of this shift and what the advantages are of the Wildcat library. As much info as possible would be great, I don't want to continue in this direction if at some point in the near future the guidance will be that we don't want to add any more to this sublibrary.

theories/Categories/Monoidal/MonoidalCategory.v
patrick-nicodemus: Because the definition of a monoidal category cannot be given without the funext axiom, do you think it is appropriate to use 
`Require Import Axioms.Funext.`

patrick-nicodemus: I ran into some frustrating technical problems at one point due to code spread across multiple files calling on distinct instances of funext. Having a unique instance of funext would be helpful.

theories/Categories/Monoidal/MonoidalCategory.v
patrick-nicodemus: Maybe you can suggest a good way to do this?

There are some overloaded notations, like \circ can mean morphism composition in a category or functor composition. Which one Coq interprets it as by default depends on the order in which the Utf8 files are required/imported, and/or the order in which the notation scopes are opened in which they're defined. I don't know how to say something like, "Let "\circ" be understood as in Functor.utf8" or "let \times be defined as in this notation scope", so I'm manually writing Local Infix "\circ" := Functor.Composition.Core.compose to make it clear that it this interpretation will take precedence over the one given in Category.Utf8. If I reversed the order in which they were imported I would have the same problem with some other command.

theories/Categories/Monoidal/MonoidalCategory.v
patrick-nicodemus: Given the name scopes, NaturalTransformation.Prod.induced_snd is the default interpretation

patrick-nicodemus:
@JasonGross 
> Thanks for the contribution!
> 
> The funext needs to be moved to a section, everything else is just a suggestion.
> 
> I'm also curious how long this file takes to compile. I seen to recall monoidal categories being slow when I tried them
[MonoidalCategory.v.timing.txt](https://github.com/HoTT/HoTT/files/8694090/MonoidalCategory.v.timing.txt)

theories/Categories/Monoidal/MonoidalCategory.v
JasonGross: > do you think it is appropriate to use `Require Import Axioms.Funext.`

No, you should insert `Section with_funext.` above this line and `End with_funext.` at the bottom of the file

theories/Categories/Monoidal/MonoidalCategory.v
JasonGross: I believe you're looking for `Local Open Scope functor_scope` or `(...)%functor`.  The former says "prefer functor scope after this point", and the latter says "interpret `...` in functor scope".  You can do the same to get notations in other scopes, and should do this instead of redeclaring the notation without any scope.

theories/Categories/Monoidal/MonoidalCategory.v
mikeshulman: I would be surprised if there isn't some way to define a monoidal category without funext.  Why do you say that?

patrick-nicodemus: https://github.com/HoTT/HoTT/blob/master/theories/Categories/FunctorCategory/Morphisms.v
The NaturalIsomorphism type here assumes funext, this is what I used to define the associator and left/right unitors.

patrick-nicodemus: We could just work with object-wise isomorphisms everywhere instead but it seems more elegant to be able to state that the pentagon equation and triangle equation hold in the category of functors and natural transformations.

theories/Categories/Monoidal/MonoidalCategory.v
patrick-nicodemus: Ok, done.

mikeshulman:
I (or someone else) can write more about wildcats later, but one of their advantages is that they avoid funext by more or less working with a setoid equality on morphisms.

theories/Categories/Monoidal/MonoidalCategory.v
JasonGross: More specifically, proving that functors forma category requires funext because the laws about composition only hold extensionally on the path parts of NTs.

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,114 @@
...
+Section MonoidalStructure.
+  Context `{Funext}.
+
+Local Notation "x --> y" := (morphism _ x y).
+
+Section MonoidalCategoryConcepts.
+  Variable C : PreCategory.
+  Variable tensor : ((C * C) -> C)%category.
+  Variable I : C.
+  Locate identity.
```
JasonGross, line 27: This seems leftover and should go
```suggestion
```

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,114 @@
...
+  
+  Local Notation "A ⊗ B" := (tensor (Datatypes.pair A B)) (at level 45, left associativity).
```
JasonGross, line 29:
```suggestion
  Local Notation "A ⊗ B" := (tensor (Datatypes.pair A B)).
```

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,114 @@
...
+
+  Local Open Scope functor_scope.
+  Locate identity.
```
JasonGross, line 32:
```suggestion
```

theories/Categories/Monoidal/MonoidalCategory.v
```diff
@@ -0,0 +1,114 @@
...
+  Locate "1". Check (1%functor _).
```
JasonGross, line 33:
```suggestion
```

JasonGross (approved):
LGTM modulo the removals I've suggested here, and assuming CI passes

Alizter:
Regarding wildcats, these are an "unbundled" approach to category theory. It was noticed a while back that much of category theory can be done without fully using all the data of a category. The most restrictive part of the data of a category is the truncatedness of the hom types. Wild cats allow you to do a lot of category theory by using Coq's typeclass search to assemble data about categories together. This has allowed us to apply categorical results to things that aren't strictly 1-categories.

A classic example is the category `Type` the arrows are functions and higher arrows homotopies. Somehow a wild cat is like an oo-cat without all the higher cells and all the coherence, but only as much as is needed for whatever it is that you are doing.

Regarding monoidal categories, your definition would work fine for hSets and vector spaces etc. But we also have our eyes on other kinds of monoidal categories such as pType. Here restricting the homs to sets would not allow us to say much about pType, but if we had a wild monoidal category, then lemmas about it would apply to both situations equally. FTR we have wild bicategories already in the WildCat/ folder and morally a monoidal category is just a "one object bicategory" (as a delooping). There are some practical formalizations reasons we might not define it this way though, and instead it would be better to have something like your development mixed with the wild category stuff.

mikeshulman:
After a side discussion with Patrick, I think it's fine to include this in the categories library.  It'd also be nice to have monoidal WildCats, but it sounds like for where he wants to go with it, the ordinary categories library may be easier.

Alizter:
I was perhaps a little too quick with my explanation of wildcats there. I should also say that I am completely fine with this PR. 0:-)

Alizter (approved):

PR #1651, opened 2022-05-07
Misc warning cleanups

Alizter:
We clean up various warnings that will be introduced in 8.16.


jdchristensen (commented):
You probably didn't mean to include the hott_log file in the commit.

Alizter:
@jdchristensen Thanks! Fixed.

Alizter:
A note about `#[global]` and `#[export]` typeclass instances:
- `#[global]` allows for commands (in this case the instance) to persist even when the section or module is closed. Essentially, requiring a file will add its instances to the typeclasses database.
- `#[export]` allows for commands to persist when the module or section is closed only when the module itself is imported. That means if you want a typeclass instance to be available, you need to import the module containing them (or another exporting that module) explicitly.

#[export] is preferred since in practice we don't rely on the global function so much. It will also allow us to bundle up instances inside modules that we don't want leaking into a main module. Take for example, the WildCat/Paths.v file which contains some local instances. We could instead put these in a module allowing us to import those instances on demand without polluting anything.

mikeshulman (approved):

PR #1652, opened 2022-05-11
[Exercise 1.13] Solved double negation of excluded middle

HyunggyuJang:
Using intuitionistic De Morgan's law.

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -250,17 +250,11 @@ End Book_1_12.
 (** Exercise 1.13 *)
 
 Section Book_1_13.
-  Lemma Book_1_13_aux: forall A B, ~(A + B) -> ~A * ~B.
-  Proof.
-    intros A B nAorB; split.
-    - intro a; exact (nAorB (inl a)).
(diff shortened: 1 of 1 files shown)


mikeshulman:
This is a perfectly good solution, of course.  But I also like the direct solution, and if we include one I'd like to also include the other.  Would you feel like writing out the direct solution too?  If not, I can try to find the time.

HyunggyuJang:
Like https://github.com/HoTT/HoTT/pull/1652/commits/cfbe48305dc45d7d8ad595d20a6034c6f8af11f6?

HyunggyuJang:
Or is it okay with let the original solution as it is, and add newly added direct one with different name for comparision?

mikeshulman:
Thanks!  I was thinking of keeping them both with different names.

HyunggyuJang:
> Thanks! I was thinking of keeping them both with different names.

Got it!

HyunggyuJang:
I’ll fix alignment issue in a minute.

HyunggyuJang:
Done

mikeshulman (approved):
Looks great, thanks!

PR #1653, opened 2022-05-23
Natural number arithmetic

patrick-nicodemus:
We add some theorems of natural number arithmetic involving pred, S,  + , -, < and <=.
We define a hint database for natural number arithmetic and add some lemmas to it.
We define a "left-handed" <= as an inductive type in the other obvious way, namely:
n <= n
n + 1 <= m  ->  n <= m
and prove it's equivalent to the other one.
This is useful for cases where you want to loop from 0 to m, going from bottom to top; if you induct on the normal definition of <= then you will loop from top to bottom.


theories/Spaces/natural_number_arithmetic.v
```diff
@@ -0,0 +1,794 @@
...
+
+Proposition not_lt_implies_geq {n m : nat} : ~(n < m) -> m <= n.
+Proof.
+  intros not_lt. apply stable. intro not_geq.
+  destruct (@leq_dichot m n).
+  - contradiction not_geq.
+  - contradiction not_lt.
+Defined.
+
+Class ComplementaryProp (A B : Type) :=
```
mikeshulman, line 64: This is a general definition not specific to natural numbers, so I would put it in a separate file.  Perhaps `Basics/Decidable`?

theories/Spaces/natural_number_arithmetic.v
```diff
@@ -0,0 +1,794 @@
...
+  - intros m k. change (n.+1 + (m + k)) with (n + (m + k)).+1.
+    apply (transitive_paths _ _ _ (nat_add_n_Sm _ _)).
+    change (m + k).+1 with (m.+1 + k);
+    change (n.+1 + m) with (n + m).+1.
+    apply (transitive_paths _ _ _ (IHn m.+1 k)).
+    apply (ap (fun zz => zz + k)).
+    apply symmetric_paths, nat_add_n_Sm. 
+Defined.
+
+Fixpoint leq_bool (n m : nat) : Bool :=
```
mikeshulman, line 35: This and some other definitions in this file have names that don't make sense globally, as they don't specify that they involve `nat`.

theories/Spaces/natural_number_arithmetic.v
```diff
@@ -0,0 +1,794 @@
...
+Proposition lt_sub_gt_0 : forall n m : nat, m < n -> 0 < n - m.
+Proof.
+  induction n.
+  - intros m ineq. contradiction (not_lt_n_0 m).
+  - induction m.
+    + simpl. easy.
+    + simpl. intro ineq. apply leq_S_n in ineq. now apply IHn in ineq.
+Defined.
+
+Proposition natminuspluseq : forall n m : nat, n <= m -> (m - n) + n = m.
```
mikeshulman, line 272: Generally we use underscores to separate words in the names of definitions and theorems.

mikeshulman (changes requested):
This looks pretty good to me.  Aside from the points I mentioned specifically, I think a better name for the file would be something like `Spaces/Nat/Arithmetic`.

theories/Basics/Decidable.v
```diff
@@ -14,6 +14,13 @@ Local Open Scope path_scope.
 (** ** Definitions *)
 
 (* NB: This has to come after our definition of [not] (which is in [Overture]), so that it refers to our [not] rather than the one in [Coq.Logic]. *)
+
+Class ComplementaryProp (A B : Type) :=
```
mikeshulman, line 18: Maybe put it at the end of the file?  In this place, it breaks the referent "this" in the preceding comment, which is supposted to refer to the definition of `Decidable` below.

mikeshulman:
Did you force-push the right thing?  Now it looks like the *only* change in this PR is adding `ComplementaryProp`.

theories/Basics/BooleanReflection.v
```diff
@@ -0,0 +1,75 @@
...
+    - intro r; destruct r.
+      + split; done.
+      + split; [ contradiction | discriminate ].
+    - intro H; destruct H as [if1 if2].
+      destruct b.
+      + apply ReflectT, if2; reflexivity.
+      + apply ReflectF; intro p; apply if1 in p; discriminate.
+  Defined.
+
+  Proposition refl_tofalse (A : Type) (b : Bool) (p : reflect A b) : ~ A -> b = false.
```
jdchristensen, line 20: Is "refl" short for "reflect"?  "refl" is used for the reflexivity path in the HoTT book, so I think it would be better to use "reflect" here.

theories/Basics/BooleanReflection.v
```diff
@@ -0,0 +1,75 @@
...
+  Proof.
+    destruct p; done.
+  Defined.
+
+  Proposition refl_totrue (A : Type) (b : Bool) (p : reflect A b) : A -> b = true.
```
jdchristensen, line 25: Maybe be consistent about using A, B, ... vs P, Q, ...?

theories/Basics/BooleanReflection.v
```diff
@@ -0,0 +1,75 @@
...
+  Proof.
+    unfold decBool; destruct dec; constructor; assumption.
+  Defined.
+  
+  Proposition decBoolSum (A B: Type) `{H0 : Decidable A} `{H1 : Decidable B} :
+    decBool (A + B) = orb (decBool A) (decBool B).
+  Proof.
+    unfold decBool.
+    unfold dec, decidable_sum at 1, dec. destruct H0, H1; done.
+  Qed.
```
jdchristensen, line 46: We usually use "Defined" unless there is a good reason not to.

theories/Classes/implementations/assume_rationals.v
```diff
@@ -10,7 +10,7 @@ Parameters (Q : Type@{UQ}) (Qap : Apart@{UQ UQ} Q)
   (Qzero : Zero Q) (Qone : One Q) (Qneg : Negate Q) (Qrecip : DecRecip Q)
   (Qle : Le@{UQ UQ} Q) (Qlt : Lt@{UQ UQ} Q)
   (QtoField : RationalsToField@{UQ UQ UQ UQ} Q)
-  (Qrats : Rationals@{UQ UQ UQ UQ UQ UQ UQ UQ} Q)
+  (Qrats : Rationals@{UQ UQ UQ UQ UQ UQ UQ UQ UQ UQ} Q)
```
jdchristensen, line 13: I'm surprised that *additional* universe variables are required.  Any idea how that happened?

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -184,7 +184,7 @@ Section upper_classes.
   #[export] Existing Instances decfield_ring.
 
   Class FieldCharacteristic@{j} {Aap : Apart@{i j} A} (k : nat) : Type@{j}
-    := field_characteristic : forall n : nat, Nat.lt@{i} 0 n ->
+    := field_characteristic : forall n : nat, Nat.lt 0 n ->
```
jdchristensen, line 187: Can the universe variable `i` be completely eliminated here?  (If not, it should be listed alongside the `j` in `FieldCharacteristic@{j}`.)

theories/Spaces/Nat.v
```diff
@@ -514,7 +514,10 @@ Proof.
 Defined.
 
 (** We define the less-than relation [lt] in terms of [leq] *)
-Definition lt n m : Type := leq (S n) m.
+Definition lt n m : Type0 := leq (S n) m.
+Set Printing Universes.
+Print lt.
```
jdchristensen, line 519: Remove.

theories/Spaces/Nat/Arithmetic.v
```diff
@@ -0,0 +1,1032 @@
...
+  Proposition Sleq_implies_leq (n m : nat) : Sleq n m -> Nat.leq n m.
+  Proof.
+    revert m; induction n.
+    + intros; apply leq_0_n.
+    + intros m H. simpl in H. destruct m; [ destruct H |].
+      apply leq_S_n', IHn; exact H.
+  Defined.
+
+  Check leq_rect.
+  Definition Sleq_rect :
```
jdchristensen, line 56: `:` at start of next line, after two spaces.

theories/Spaces/Nat/Arithmetic.v
```diff
@@ -0,0 +1,1032 @@
...
+  Definition Sleq_rect : 
+      forall (n : nat) (P : forall m : nat, Sleq n m -> Type),
+        P n (Sleq_refl n) ->
+        (forall (m : nat) (l : Sleq n m), P m l -> P (m.+1)%nat (Sleq_S n m l)) ->
+        forall (m : nat) (l : Sleq n m), P m l.
+  Proof. intros n P diag succ m l.
+         change l with (leq_implies_Sleq n m (Sleq_implies_leq n m l)).
+         induction (Sleq_implies_leq n m l).
+         - apply diag.
+         - apply succ, IHl0, leq_implies_Sleq; assumption.
```
jdchristensen, line 65: Incorrect indentation.

theories/Spaces/Nat/Arithmetic.v
```diff
@@ -0,0 +1,1032 @@
...
+  | [ H : ?n > ?n |- _ ] => contradiction (not_lt_n_n n)
+  | [ H : ?n < ?n |- _ ] => contradiction (not_lt_n_n n)
+  end.
+
+Global Hint Resolve not_lt_n_n : nat.
+Global Hint Resolve not_lt_n_0 : nat.
+Global Hint Resolve not_leq_Sn_0 : nat.
+Global Hint Extern 2 => nat_absurd_trivial : nat.
+
+(* Proposition iff_reflect (P : Type) (b : Bool) : (reflect P b) <-> (P <-> b = true). *)
```
jdchristensen, line 85: Lots of commented out stuff?

theories/Spaces/Nat/Arithmetic.v
```diff
@@ -0,0 +1,1032 @@
...
+Proposition add_n_sub_n_eq : forall m n : nat,  m + n - n = m.
+Proof.
+  intros m n; destruct m.
+  - induction n.
+    + reflexivity.
+    + assumption.
+  - induction n.
+    + simpl. rewrite nat_add_comm. reflexivity.
+    + simpl. rewrite <- nat_add_n_Sm. assumption.
+Qed.
```
jdchristensen, line 372: `Defined`, here and elsewhere?

jdchristensen (commented):
I'm not really able to review the overall approach here.  My comments are mainly minor stylistic issues.  (Also, this appears to be WIP.  Maybe mark it as "draft" until it is ready for full review?)

theories/Classes/implementations/assume_rationals.v
jdchristensen: (This comment and the one about `FieldCharacteristic` might be related.  It's possible that by removing the universe variable from `lt`, some universe variables become "freer", which means that in subsequent definitions, they aren't forced to be equal.  And then those subsequent definitions end up with *more* universe parameters.  So it would be worth checking whether some universe variables (like the `i` in `FieldCharacteristic`) can be eliminated.  Semantically, it's not clear why the rationals need any universe variables at all...)

theories/Classes/interfaces/abstract_algebra.v
SkySkimmer: i is the universe of A from the section, section universes are not listed in the @{}

theories/Basics/SProp.v
```diff
@@ -0,0 +1,67 @@
+Require Import Basics Types.Sigma.
+
+Record Box (A : SProp) : Type0 := box { unbox : A }.
```
SkySkimmer, line 3: Unset Primitive Projections around this record to remove the warning

theories/Basics/BooleanReflection.v
patrick-nicodemus: Ok, this is a good point.

theories/Classes/implementations/assume_rationals.v
patrick-nicodemus: This is not clear to me either. My guess is that since this is an interface, the idea is that the underlying structures which implement it could potentially be arbitrarily complex as long as they expose the desired functionality to the user?

theories/Basics/BooleanReflection.v
patrick-nicodemus: Yep, agreed.

patrick-nicodemus:
> I'm not really able to review the overall approach here. My comments are mainly minor stylistic issues. (Also, this appears to be WIP. Maybe mark it as "draft" until it is ready for full review?)

Ok, marked as draft for the time being. I would appreciate you continuing to make comments and suggestions anyway, please let me know when you think it is appropriate for it to be ready for full review. I would rather have small parts integrated if they are useful than try to submit a big monolithic chunk of code and discuss it for a long time.

theories/Basics/SProp.v
patrick-nicodemus: Thank you, done.

theories/Spaces/Nat/Arithmetic.v
```diff
@@ -0,0 +1,950 @@
...
+Require Import Basics.Tactics.
+Require Import Types.Bool.
+Require Import Types.Sum.
+Require Import Basics.Utf8.
+
+Require Import Basics.Nat.
+Require Import Basics.Decidable.
+Require Import BooleanReflection.
+(* This prefix seems necessary to avoid conflict with Coq.Init.Sprop? *)
+Require Import HoTT.SProp.
```
SkySkimmer, line 11: There is no Coq.Init.Sprop, only Coq.Logic.StrictProp

mikeshulman:
> I would rather have small parts integrated if they are useful than try to submit a big monolithic chunk of code and discuss it for a long time.

That's a good idea and would make it easier to review.  But in that case, you should separate out the small parts into separate pull requests.  For instance, it would probably be appropriate to have separate PRs for the boolean stuff and for the sprop stuff.

theories/Spaces/Nat/Arithmetic.v
patrick-nicodemus: Thanks, all theorems in this file have been changed to Defined.
The style guide makes the following statements:

> If the value of something being defined matters, then you must either give an explicit term defining it, or construct it with tactics and end the proof with Defined. But if all that matters is that you have defined something with a given type, you can construct it with tactics and end the proof with Qed. The latter makes the term "opaque" so that it doesn't "compute".
> 
>  If something can be made opaque, it is generally preferable to do so, for performance reasons. However, many things which a traditional type theorist would make opaque cannot be opaque in homotopy type theory. For instance, none of the higher-groupoid structure in PathGroupoids can be made opaque, not even the "coherence laws". If you doubt this, try making some of it opaque and you will find that the "higher coherences" such as pentagon and eckmann_hilton will fail to typecheck.

You said earlier that "We usually use "Defined" unless there is a good reason not to." I think it would be helpful to clarify this point in the style guide. For example,
> The important thing is that when defining a transparent term with tactics, you should restrict yourself to tactics which maintain a high degree of control over the resulting term; ... Even plain rewrite is usually to be avoided in this context: although the terms it produces are at least predictable, they are one big transport (under a different name) whereas a term we would want to reason about ought to be constructed using smaller pieces like ap and concat which we can understand.

Since the proof uses rewrites, I terminated the proof with Qed. 
I discussed this issue the other day with @mikeshulman in the Zulip and he said that the proof terms that arise from running destruct on an equality are less complex than those arising from rewrite, so here I was able to replace all "rewrites" with "destructs" and there is no conflict.

theories/Basics/BooleanReflection.v
patrick-nicodemus: I have removed those theorems in the new commit. I am thinking I will use the theorem names in here.
https://coq.inria.fr/distrib/V8.16.0/stdlib/Coq.ssr.ssrbool.html

theories/Spaces/Nat/Arithmetic.v
patrick-nicodemus: Ok, thank you. I don't know what the error is. "Require SProp" and "Require Import SProp" give errors which I don't understand.

Alizter:
Regarding the proof terms of rewrite. Now that we have 8.16 we could start using the generalized rewriting feature of Coq (formly setoid_rewrite). If we spent some time setting this up we could have nice proof terms when doing rewrites. It would be similar to an attempt at nice rewrite terms by @jdchristensen from a few years ago.

patrick-nicodemus:
@mikeshulman Ok, I got overly excited. SProp and BooleanReflection stuff deleted from the commit so we can just talk about natural number arithmetic stuff, I will open a second PR request soon for the boolean reflection stuff, SProp stuff seems like it might need further discussion.

theories/Spaces/Nat/Arithmetic.v
mikeshulman: I would actually be inclined to make these `Qed`s as you originally had them.  Firstly, because of the rewrites, the resulting proof terms are not going to be very nice to look at; and secondly, they inhabit hprops (equalities in nat), so at least propositionally their definitional value doesn't matter.  It seems unlikely to me that we would ever need these to be transparent.

mikeshulman: In particular, aAs you quoted from the style guide, I think that elsewhere we do often use `Qed` for proofs involving `rewrite`.  I think there are some places where we found some value in making rewrite proofs transparent, but I thought in general we were sticking to Qed for rewrite proofs.

theories/Spaces/Nat/Arithmetic.v
SkySkimmer: Probably because SProp is a keyword ie treated specially by the parser

SkySkimmer:
Is setoid_rewrite considered to have nice proof terms?

Alizter:
@SkySkimmer Since we can give the rewrite instances directly, we can fill it with concatenations and aps depending on the form of the term.

theories/Classes/implementations/assume_rationals.v
jdchristensen: Right, since it's an interface, it should accommodate implementations that don't live in `Type0`.  But I'm still confused why additional universe variables appeared.  The attached screenshot shows how the universe variables changed in `Rationals`.  The new variables are u7 and u8, and in the old version they were always replaced by u6.  Odd.  Not a big deal, but I'm just curious, as I hoped that changing `lt` would *reduce* universe variables elsewhere.

Screenshot from 2022-10-10 09-49-48

theories/Classes/interfaces/abstract_algebra.v
jdchristensen: Ah, sorry, I missed that `i` was defined earlier and is an important universe variable that can't be eliminated.

theories/Classes/implementations/assume_rationals.v
SkySkimmer: What does the change in RationalsToField look like?

theories/Classes/interfaces/abstract_algebra.v
patrick-nicodemus: Is this resolved?

theories/Classes/implementations/assume_rationals.v
jdchristensen: (Even though we didn't get to the bottom of why extra universe variables appear here, I don't think this should hold up this PR.)

patrick-nicodemus:
@mikeshulman Your original "changes requested" applied to an early draft of this PR before I threw out a bunch of independent things like stuff related to boolean reflection or SPRop. Can you take another look at it and update the changes requested?

PR #1657, opened 2022-09-02
Make Contr_internal and some algebra Classes Cumulative

jdchristensen:
The main change here is to specify that `Contr_internal` is a `Cumulative Record`.  With this done, the following things work:

```
Definition test@{u v | u < v} (A : Type@{u}) (C : Contr_internal@{u} A)
  : Contr_internal@{v} A := C.

Definition test2@{u v | u < v} (A : Type@{u}) (C : IsHSet A)
  : IsHSet A := C.

Cumulative Record Test@{+u} := { T : Type@{u}; H : IsHSet T }.
```

However, for general `n`, you don't get definitional cumulativity of `IsTrunc_internal n A`:

```
Fail Definition test3@{u v | u < v} (n : trunc_index) (A : Type@{u}) (C : IsTrunc_internal@{u u} n A)
  : IsTrunc_internal@{v v} n A := C.
```

(Is it possible to make that work?  It would be nice to have.)

The reason Jarl and I are interested in making this change is that it allows `Group` to be cumulative as well.  (The `IsHSet` condition was an issue before this.)  When dealing with things like `BAut` and the type of all short exact sequences, the groups you get can lie in different universes, which means one has to fiddle with a lot of explicit universe variables.  This PR includes changes to `Group` and a few other needed things to allow us to avoid this.

Question:  should I also make this change to everything else in `Classes/interfaces/abstract_algebra.v`?  Or should we just make this change as needed?

(Note that I couldn't find a way to tell Coq to make Records and Classes cumulative by default.)

I timed builds before and after this change, and they were within the noise.  No file increased by more than 0.1s.  I see no logical issues with having `Contr_internal` cumulative, but I'm not an expert on this.

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -5,6 +5,8 @@ Require Export Algebra.Groups.
 Require Import Cubical.
 Require Import WildCat.
 
+Local Set Polymorphic Inductive Cumulativity.
+
 Local Open Scope mc_add_scope.
 
 (** * Abelian groups *)
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -7,6 +7,8 @@ Require Import Pointed.Core.
 Require Import WildCat.
 Require Basics.Utf8.
 
+Local Set Polymorphic Inductive Cumulativity.
+
 Generalizable Variables G H A B C f g.
 
 Declare Scope group_scope.
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -539,7 +539,7 @@ In order to achieve moderate coverage and speedy resolution, we currently follow
 (** A space [A] is contractible if there is a point [x : A] and a (pointwise) homotopy connecting the identity on [A] to the constant map at [x].  Thus an element of [contr A] is a pair whose first component is a point [x] and the second component is a pointwise retraction of [A] to [x]. *)
 
 (** We use the [Contr_internal] record so as not to pollute typeclass search; we only do truncation typeclass search on the [IsTrunc] datatype, usually.  We will define a notation [Contr] which is equivalent to [Contr_internal], but picked up by typeclass search.  However, we must make [Contr_internal] a class so that we pick up typeclasses on [center] and [contr].  However, the only typeclass rule we register is the one that turns it into a [Contr]/[IsTrunc].  Unfortunately, this means that declaring an instance like [Instance contr_foo : Contr foo := { center := bar }.] will fail with mismatched instances/contexts.  Instead, we must iota expand such definitions to get around Coq's deficiencies, and write [Instance contr_foo : Contr foo := let x := {| center := bar |} in x.] *)
-Class Contr_internal (A : Type) := Build_Contr {
+Cumulative Class Contr_internal (A : Type) := Build_Contr {
   center : A ;
   contr : (forall y : A, center = y)
 }.
--- a/theories/Classes/interfaces/abstract_algebra.v
+++ b/theories/Classes/interfaces/abstract_algebra.v
@@ -2,6 +2,8 @@ Require Import Spaces.Nat.
 Require Export HoTT.Classes.interfaces.canonical_names.
 Require Import HProp.
 
+Local Set Polymorphic Inductive Cumulativity.
+
 Generalizable Variables A B f g x y.
 
 (* 


Alizter:
Should we perhaps consider making all inductive types cumulative by default? I'm not aware of any reasons against it.

SkySkimmer:
>However, for general n, you don't get definitional cumulativity of IsTrunc_internal n A:
(Is it possible to make that work? It would be nice to have.)

The fixpoint mentions the universe so the kernel rules for cumulativity would need to be expanded in some unknown way.

Alternatively we could use the inductive version
~~~coq
Inductive IsTrunc (A:Type) : nat -> Type :=
| isTrunc_contr : Contr A -> IsTrunc A -2
| isTrunc_S n : (forall x y : A, IsTrunc (x = y) n) -> IsTrunc A (n.+1)
~~~
(argument order needs to be changed because A must be a parameter and n an index)
I think this works for cumulativity but then it doesn't compute on known n.

SkySkimmer:
>(Note that I couldn't find a way to tell Coq to make Records and Classes cumulative by default.)

`Set Polymorphic Inductive Cumulativity`

>Should we perhaps consider making all inductive types cumulative by default? I'm not aware of any reasons against it.

The change would take some work to deal with universe inference changes. That work is somewhere between trivial (not included) and too much to be worth it (included).

jdchristensen:
I thought I read somewhere that `Polymorphic Inductive Cumulativity` only handles `Inductive` and not `Class`, but I just tested and it does indeed do the trick.  However, as @SkySkimmer suggested, it then affects other parts of the library.  The first build error is in Descent.v, and it's not immediately obvious to me how to fix this.  And Categories/Adjoint/Hom.v gets stuck on the `Record AdjunctionHom :=` line (no error, just spins).  Many other files build, but I also suspect there will be other problems.

jdchristensen:
@SkySkimmer Interesting idea for an inductive version of `IsTrunc`.  I have no idea how that would affect things.

jdchristensen:
As an intermediate step, I used `Local Set Polymorphic Inductive Cumulativity` for a few files where this caused no problem.  Trying this for Overture.v gave a problem in ObjectClassifier.v, so there I just kept the `Cumulative` keyword for `Contr_internal`.  This seems like a simple compromise for now.  If people find other uses of cumulativity, they can try adding the local setting to the relevant files and fix issues that arise.

JasonGross:
> And Categories/Adjoint/Hom.v gets stuck on the `Record AdjunctionHom :=` line (no error, just spins).

If you can minimize this (perhaps with the bug minimizer and Timeout), it's worth reporting as a bug

SkySkimmer:
> And Categories/Adjoint/Hom.v gets stuck on the Record AdjunctionHom := line (no error, just spins).

This should improve with coq 8.16

jdchristensen:
> This should improve with coq 8.16

Maybe someone with a development version can test this?  Just put `Local Set Polymorphic Inductive Cumulativity.` near the top of Basics/Overture.v, and see if `make theories/ObjectClassifier.vo` works.

In any case, I'm happy with the current PR.  I could squash the commits before merging.

Alizter:
@jdchristensen Hopefully 8.16 will be released sometime next week. If you can make an issue detailing the suggestion, we can try it out once we update.

Btw I've fixed the CI, if you rebase it will be green.

jdchristensen:
I squashed the commits and force-pushed.

Alizter (approved):

mikeshulman (approved):

mikeshulman:
If there are places where cumulativity of `IsTrunc` for general `n` would be useful, I think the inductive version would be worth trying.  Totally off the cuff without having tested anything, my guess would be that it would necessitate changes to the proofs of a handful of basic lemmas, and a more elaborate proof that `IsTrunc` is a proposition, but few changes elsewhere in the library since I think we use the basic lemmas more often than unfolding the definition.

jdchristensen:
Currently if your goal is `IsTrunc n.+1 A`, you can just say `intros x y`.  But that would have to change to `apply isTrunc_S`.  Easy, but might happen in lots of places.  In the other direction, to use a hypothesis of type `IsTrunc n.+1 A`, we can currently just apply it to arguments, but with the inductive version, I think we'd have to do induction on the term and match on the `isTrunc_S` case, or something like that (maybe even handling the impossible case too).  I'd say we should only attempt this if there is a real need.  (Of course, that could all be put into a lemma or maybe even a coercion, so it would be easy to use.)

mikeshulman:
My memory is that while you *can* say just `intros x y`, if you do then the goal becomes something uglier like `IsTrunc_internal n A`, so it's better style to `apply isTrunc_S` anyway.  And as you say, the other direction, once in a lemma and then apply that lemma.  You're right that it might require a fair number of these changes throughout the library, but most of them would be straightforward search-and-replace.  I agree it's probably not worth it unless we have a reason, but maybe we can make a note of it somewhere.

Alizter:
Isn't it very rare that we have to explicitly prove a truncation level by hand? Perhaps there are not so many cases.

PR #1659, opened 2022-09-03
Allow O_rec to unfold when passed (tr x)

jdchristensen:
We currently have `O_rec` marked as `simpl never` to avoid having it unfold with `cbn`.  This PR changes it to

```
Arguments O_rec {O} {P Q}%type_scope {Q_inO H H0} f%function_scope !x.
```

which means that `cbn` will unfold it if `x` appears with a constructor.  This means that `cbn` changes `O_rec (O:=Tr n) f (tr x)` to `f x` and `Trunc_functor n f (tr x)` to `tr (f x)`, both of which are convenient.  If the argument is `x` instead of `tr x`, then these are left alone.

No other changes were required in the library.

Five `cbn`s in FreeProduct.v got about 0.1s slower each, but both before and after they produced an unwieldy term, and it worked to just remove them.  After this, the overall build time is about 0.1s faster than before, and no file increased by more than 0.06s.

In #1475, `Arguments O_rec / .` was mentioned, but this breaks the build and unfolds too much in some situations.  Similarly, 

```
Arguments O_rec {O} {P Q}%type_scope {Q_inO H H0} f%function_scope / x.
```

unfolds `O_rec` too eagerly.

This simplifies some WIP.

### DIFF
--- a/theories/Algebra/Groups/FreeProduct.v
+++ b/theories/Algebra/Groups/FreeProduct.v
@@ -322,7 +322,7 @@ Section FreeProduct.
         1,3: apply word_concat_w_ww.
         apply amal_omega_K. } }
     { intros r y h1 h2; revert r.
-      rapply amal_type_ind_hprop; cbn.
+      rapply amal_type_ind_hprop.
       intros z;
       change (amal_eta ((x ++ ((inl h1 :: [inl h2]) ++ y)) ++ z)
         = amal_eta ((x ++ [inl (h1 * h2)] ++ y) ++ z)).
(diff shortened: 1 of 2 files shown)


Alizter (approved):
LGTM

Alizter:
@mikeshulman Did you also want to take a look?

mikeshulman:
Sure, LGTM.  Should we consider doing the same with `O_rec_beta`, `O_indpaths`, etc.?

jdchristensen:
The library builds without other changes if all six have `!x`.  I haven't thought about situations where that would help, but maybe it's a good idea in principle?

mikeshulman:
If it doesn't require any other changes or slow things down, I would be inclined to treat all six the same way, if for no other reason than the principle of least surprise.

jdchristensen:
I'll check the timing and add another commit if all is good.

Alizter:
@jdchristensen How was the timing?

jdchristensen:
0.01s faster for the whole build, and no file changed by more than 0.05s.  (I.e. all basically in the noise.)  Those additional definitions aren't used that much.

PR #1661, opened 2022-09-05
bump supported Coq version to 8.16

Alizter:
The HoTT release for 8.16 was already done. If we merge this then we can start using the new 8.16 features such as:
- Reversible coercions, see #1641
- Better cumulative universes
- `Print Notation` command detailing parsing information about a particular notation.
- Generalized rewriting (formerly setoid_rewrite) now supports relations polymorphic in Type.

See the full Coq changelog: https://coq.inria.fr/refman/changes.html

PR #1662, opened 2022-09-13
remove 8.15 compat

Alizter:
We had some compat code leftover.

### DIFF
--- a/theories/Modalities/Localization.v
+++ b/theories/Modalities/Localization.v
@@ -354,13 +354,9 @@ Proof.
     intros ? T; unfold IsLocal.
     nrefine (istrunc_forall@{a i i}); try assumption.
     intros i.
-    (* coq#14758 compat (coq 8.15) *)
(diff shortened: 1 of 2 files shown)


jdchristensen:
Does this mean it won't build with 8.15 anymore?  Maybe we should keep compatibility until we have a strong reason to not be compatible, as some users might not have upgraded to 8.16.

jdchristensen:
Well, even after #1661, the library should still be usable with 8.15.  I would prefer to delay removing the compatibility code until we have merged something else that prevents the library from being used with 8.15.  8.16 was only released a week ago...

Alizter:
@jdchristensen As you wish, this PR does no harm sitting here.

Alizter:
@jdchristensen Since #1697 is going to use 8.16 features, I think we should merge this.

jdchristensen:
I agree.  LGTM.

Issue #620, opened 2014-11-21
Reduce universes for performance


mikeshulman:
It seems likely that a lot of the slowness of `V.v` is due to its insane number of universe parameters (e.g. `V.function` has over 2000).  We should experiment with adding more universe annotations to reduce this number and see how it affects performance.

spitters:
We have more annotations now. I guess this can be closed?

JasonGross:
No, I don't think we do.  Last I checked, `V.v` was still using around thousands, if not tens of thousands, of universe variables in a single definition.

mattam82:
Just as a remark, I'm experimenting with more agressive minimization to try to lower these numbers, I'll keep you updated. In my experience one thing that can be tried otherwise is to see if you have basic definitions creating many universes for the same l.u.b. and explicitly share them.

mattam82:
I've went back on this. The problem is caused by using Let's in sections. Each Let introduces universes that are rigid so when they unifiy with each other you get two universes plus an equality between them instead of a single universe without additional constraints. Most of the 20000's come from this. This is really a design issue with sections, which would be resolved by "lambda-lifting" section constants immediately and just providing an appropriate shortcut to their instantiated versions while working with them in the section. This would avoid having to do any work at section closing time as well. I'm planning to write a CEP about this, and hopefully have it done by Coq 8.7/9. In the meantime, I suggest avoiding Let's here and using definitions instead.

mikeshulman:
I only see five `Let`s in `V.v`, and replacing them with `Definition` doesn't seem to appreciably change the number of universes or the speed of compiling that file.

spitters:
@mattam82 Did you ever open this CEP?

mattam82:
@spitters nope, I've had no time to look into it. We should discuss this with @SkySkimmer and @ppedrot, maybe they have different ideas on how to tame this problem.

SkySkimmer:
Maybe @ppedrot 's pseudo-monomorphic universe instantiation would work for this? Not sure to what extent V uses univ polymorphism.

ppedrot:
I don't think it'd affect let-bindings in a section, but it may help otherwise. I still have the branch somewhere if you want to try to play with it.

spitters:
Where's the branch?

ppedrot:
[...]

Issue #122, opened 2013-04-11
Fix HIT definitions.


peterlefanulumsdaine:
Rewrite the defined HIT eliminators as described to Yves’ email today, so that Anthony’s inconsistency isn’t available for them.

PR #1665, opened 2022-09-23
Reduce universe variables in V.v

jdchristensen:
Closes: #620 

See issue #620 for a discussion of this.  In the past, some definitions in the file HIT/V.v used tens of thousands of universe variables.  It seems like Coq has to a large degree fixed this:  with Coq 8.16, I randomly checked a number of definitions, and the worst I found were:  `V_func` (219 universe variables), `V_union` (219 universe variables) and `infinity` (64).  (Did I miss any that are worse?)

I found that annotating the definition of just one function, `V_ind`, greatly improved things.  Now those three definitions use 3, 3 and 6 universe variables, respectively.  Also, the build time goes from 1.35s to 0.96s on my system.

I also annotated the definition of `V` a bit more, but there is no change to the meaning.  This just documents and enforces the existing universe variables.

One note:  it's possible to get rid of the universe variable `U'` in the definition of `V`, which would make it land in the successor universe `Type@{U+1}`.  But since we can't use successor universes in other situations, we still need `U'` for most or all of the remaining definitions in the file.  So I left it in `V` as well, for consistency.

### DIFF
--- a/theories/HIT/V.v
+++ b/theories/HIT/V.v
@@ -29,6 +29,7 @@ set (h o (spushl R)) = set (h o (spushr R)).
 Axiom ishset_V : IsHSet V.
 Global Existing Instance ishset_V.
 
+(** The induction principle.  Annotating the universes here greatly reduces the number of universe variables later in the file.  For example, [function] below went from 279 to 3.  If [V_ind] needs to be generalized in the future, check [function] to make sure things haven't exploded again. *)
 Fixpoint V_ind@{U' U u | U < U'} (P : V@{U' U} -> Type@{u})
   (H_0trunc : forall v : V@{U' U}, IsTrunc 0 (P v))
   (H_set : forall (A : Type@{U}) (f : A -> V) (H_f : forall a : A, P (f a)), P (set f))
--- a/theories/HIT/V.v
+++ b/theories/HIT/V.v
@@ -19,7 +19,7 @@ Definition bitotal {A B : Type} (R : A -> B -> HProp) :=
 
 Module Export CumulativeHierarchy.
 
-Private Inductive V : Type@{U'} :=
+Private Inductive V@{U' U | U < U'} : Type@{U'} :=
 | set {A : Type@{U}} (f : A -> V) : V.
 
 Axiom setext : forall {A B : Type} (R : A -> B -> HProp)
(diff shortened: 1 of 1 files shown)


jdchristensen:
Oh, I now see that `function` was explicitly mentioned in #620.  With the change to `V_ind`, `function` went from 279 universe variables to 3.

jdchristensen:
I also just noticed that there was an unneeded `fun _ _ =>` in the definition of `V_ind`.  Removing it speeds up the build time from 0.96s to 0.93s.

BTW, someone should check that the universe annotations in `V_ind` are reasonable.  I partly guessed, and then just checked that the library builds, but it's possible that some greater generality could be useful.

jdchristensen:
Tagging @jledent , who last made relevant changes to `V_ind`.

SkySkimmer:
>I also just noticed that there was an unneeded fun _ _ => in the definition of V_ind. Removing it speeds up the build time from 0.96s to 0.93s.

I believe it's there because otherwise it may be possible to get around the need for the truncation hypotheses.
See also https://github.com/HoTT/HoTT/issues/122#issuecomment-22835748
Because there is a fixpoint in this case I'm not sure if the issue applies though.

jdchristensen:
This might be a holdover from a previous version.  Back in 2014, the end looked like:
```
: forall v : V, P v
:= fix F (v : V) :=
     (match v with
      | set A f => fun _ _ => H_set A f (fun a => F (f a))
     end) H_setext H_0trunc.
```
and in master the whole thing is a Fixpoint, which ends with:
```
: P v
:= (match v with
     | set A f => fun _ _ => H_set A f (fun a => V_ind P H_0trunc H_set H_setext (f a))
    end) H_setext H_0trunc.	
```
This PR uses:
```
: P v
:= (match v with
     | set A f => H_set A f (fun a => V_ind P H_0trunc H_set H_setext (f a))
    end).
```
which might be safe because `H_setext` and `H_0trunc` are used in the recursive call.  But I'm not sure.

I can revert this part, unless someone knows that it is safe.  I shouldn't have mixed different issues in this PR.

jdchristensen:
I did check that things like 
```
Definition V_ind_eq@{U' U u | U < U'} `{Funext} (P : V@{U' U} -> Type@{u})
  (H_0trunc : forall v : V@{U' U}, IsTrunc 0 (P v))
  (H_0trunc' : forall v : V@{U' U}, IsTrunc 0 (P v))
  (H_set : forall (A : Type@{U}) (f : A -> V) (H_f : forall a : A, P (f a)), P (set f))
  (H_setext : forall (A B : Type@{U}) (R : A -> B -> HProp@{U}) (bitot_R : bitotal R)
    (h : SPushout R -> V) (H_h : forall x : SPushout R, P (h x)),
    transport@{U' u} _ (setext R bitot_R h) (H_set A (h o spushl R) (H_h oD spushl R))
      = H_set B (h o spushr R) (H_h oD spushr R) )
  (v : V)
  : V_ind P H_0trunc H_set H_setext v = V_ind P H_0trunc' H_set H_setext v
  := idpath.
```
fail, and similarly with a second `H_setext'` argument, but I don't know if that is enough to avoid the inconsistency.

mikeshulman:
It's great that a single universe annotation improves things that much!  That part of this LGTM.

I don't really know about the `fun _ _ =>`.  If no one else is sure either, maybe leave that change out?  0.03s isn't much.

jdchristensen:
Ok, I have dropped that commit and force pushed.  I also added a comment explaining why the universes are annotated and warning anyone who changes them in the future to be careful.

Alizter (approved):

Alizter:
The dpdgraph error can be ignored. Something needs to be tweaked there.

PR #1666, opened 2022-10-10
Added a hint MapIn (Tr (-1)) f -> IsInjective f as an Instance in the…

patrick-nicodemus:
… typeclasses resolution hint database for IsInjective.

This broke proofs in Sets/Ordinals.v and Sets/Hartogs.v where the automation solved more cases than before; I deleted the extra cases that no longer needed a manual solution.

Total compilation time increased by 6 seconds from 8:42 to 8:48, a 1% increase. Total memory usage unchanged.
Largest time increase in Classes/theory/int_abs.vo, compilation took 2 seconds longer.

### DIFF
--- a/theories/Classes/interfaces/abstract_algebra.v
+++ b/theories/Classes/interfaces/abstract_algebra.v
@@ -349,10 +349,10 @@ Global Instance isinj_idmap A : @IsInjective A A idmap
 
 #[export]
 Hint Unfold IsInjective : typeclass_instances.
-Print mapinO_tr_istruncmap.
+
 #[export]
-  Instance IsInjective_mapinO_tr {A B : Type} (f : A -> B)
-  {p : MapIn (Tr (-1)) f} : IsInjective f :=
+Instance isinjective_mapinO_tr {A B : Type} (f : A -> B)
+{p : MapIn (Tr (-1)) f} : IsInjective f :=
   fun x y pfeq => ap pr1 (@center _ (p (f y) (x; pfeq) (y; idpath))).
   
 Section strong_injective.
--- a/theories/Classes/interfaces/abstract_algebra.v
+++ b/theories/Classes/interfaces/abstract_algebra.v
@@ -349,7 +349,12 @@ Global Instance isinj_idmap A : @IsInjective A A idmap
 
 #[export]
 Hint Unfold IsInjective : typeclass_instances.
-
+Print mapinO_tr_istruncmap.
+#[export]
+  Instance IsInjective_mapinO_tr {A B : Type} (f : A -> B)
+  {p : MapIn (Tr (-1)) f} : IsInjective f :=
+  fun x y pfeq => ap pr1 (@center _ (p (f y) (x; pfeq) (y; idpath))).
+  
 Section strong_injective.
   Context {A B} {Aap : Apart A} {Bap : Apart B} (f : A -> B) .
   Class IsStrongInjective :=
--- a/theories/Classes/interfaces/abstract_algebra.v
+++ b/theories/Classes/interfaces/abstract_algebra.v
@@ -352,8 +352,8 @@ Hint Unfold IsInjective : typeclass_instances.
 
 #[export]
 Instance isinjective_mapinO_tr {A B : Type} (f : A -> B)
-{p : MapIn (Tr (-1)) f} : IsInjective f :=
-  fun x y pfeq => ap pr1 (@center _ (p (f y) (x; pfeq) (y; idpath))).
+  {p : MapIn (Tr (-1)) f} : IsInjective f
+  := fun x y pfeq => ap pr1 (@center _ (p (f y) (x; pfeq) (y; idpath))).
   
 Section strong_injective.
   Context {A B} {Aap : Apart A} {Bap : Apart B} (f : A -> B) .
--- a/theories/Sets/Hartogs.v
+++ b/theories/Sets/Hartogs.v
@@ -37,7 +37,6 @@ Section Hartogs_Number.
     exists carrier relation. srapply (isordinal_simulation pr1).
     - exact _.
     - exact _.
-    - intros B C. apply path_sigma_hprop.
     - constructor.
       + intros a a' a_a'. exact a_a'.
       + intros [B small_B] C C_B; cbn in *. apply tr.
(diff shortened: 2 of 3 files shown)


theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -349,7 +349,12 @@ Global Instance isinj_idmap A : @IsInjective A A idmap
 
 #[export]
 Hint Unfold IsInjective : typeclass_instances.
-
+Print mapinO_tr_istruncmap.
```
jarlg, line 352: This line should be omitted.

jarlg:
Cool, this was indeed what I meant to suggest on Zulip. The change in build time is quite high though, and I'm a bit confused since the times you posted on Zulip indicated a _reduction_ of 12s in build time (first five lines below). Are the numbers in the commit message the correct ones?

```text
   After |  Peak Mem | File Name                                                    |   Before |  Peak Mem ||    Change || Change (mem) | % Change | % Change (mem)
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
8m57.95s | 593360 ko | Total Time / Peak Mem                                        | 9m10.24s | 593248 ko || -0m12.28s ||       112 ko |   -2.23% |         +0.01%
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
0m22.45s | 442064 ko | Classes/implementations/natpair_integers.vo                  | 0m19.13s | 441460 ko || +0m03.32s ||       604 ko |  +17.35% |         +0.13%
0m15.91s | 420236 ko | Algebra/AbGroups/AbSES/Pullback.vo                           | 0m19.21s | 421140 ko || -0m03.30s ||      -904 ko |  -17.17% |         -0.21%
0m19.12s | 593360 ko | Colimits/Colimit_Flattening.vo                               | 0m21.97s | 593248 ko || -0m02.84s ||       112 ko |  -12.97% |         +0.01%
0m03.80s | 363460 ko | Classes/theory/int_abs.vo                                    | 0m01.77s | 361544 ko || +0m02.02s ||      1916 ko | +114.68% |         +0.52%
0m08.77s | 433040 ko | Algebra/AbGroups/AbSES/PullbackFiberSequence.vo              | 0m10.48s | 433100 ko || -0m01.71s ||       -60 ko |  -16.31% |         -0.01%
```

jdchristensen:
It's hard to do timing tests well, especially with cpu cores that change speed as the system temperature changes.  I did some tests, only comparing the build time of the part of the library that depends on the change in this PR, and I have a fast system so I could repeat the tests.  With this PR, things were about 0.2% slower, which I think is fine.  (I also know some ways to speed up natpair_integers, which I'll do in a separate PR.)

jdchristensen:
I think it might be better to make `isinj_embedding` (in HSet.v) into an instance.  Then we'd have a chain of instances: `mapinO_pr1` which gives `MapIn (Tr -1)` for a family of hprops;  `istruncmap_mapinO_tr` which goes from that to `IsTruncMap (-1)` (which is what the notation `IsEmbedding` means); and `isinj_embedding` which goes from `IsEmbedding` to `IsInjective`.

jdchristensen:
Oh, rats, I had one of the steps backwards.  `istruncmap_mapinO_tr` is not an instance, but is instead marked as `Hint Immediate`.  That's because the reverse implication is an instance.  So in the end I think I think the instance in this PR is correct.  Then we have the chain (family of HProps) => MapIn (-1) => IsInjective (the second one being this PR), as well as IsTruncMap (-1) => MapIn (-1).

(As an aside, there's an annoying duplication, with `isinj` in HSet.v and `IsInjective` in abstract_algebra.v stating the same thing, but with the latter as a Class.  It's outside the scope of this PR, but is maybe worth cleaning up.)

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -349,7 +349,12 @@ Global Instance isinj_idmap A : @IsInjective A A idmap
...
+#[export]
+  Instance IsInjective_mapinO_tr {A B : Type} (f : A -> B)
```
jdchristensen, line 354: Our convention in the main part of the library would be to use lower case "isinjective_..."

jdchristensen (approved):

jdchristensen:
LGTM once the minor comments are addressed.

theories/Classes/interfaces/abstract_algebra.v
mikeshulman: Also there seems some spurious indentation here?

mikeshulman:
Ditto.

patrick-nicodemus:
> Cool, this was indeed what I meant to suggest on Zulip. The change in build time is quite high though, and I'm a bit confused since the times you posted on Zulip indicated a _reduction_ of 12s in build time (first five lines below). Are the numbers in the commit message the correct ones?
Yes, @jarlg I don't know what happened there. I just got different results on different times running the same experiment.

theories/Classes/interfaces/abstract_algebra.v
patrick-nicodemus: Yeah, I was just following the indentation suggested  by ProofGeneral which is... sometimes unclear in its rationale.

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -349,10 +349,10 @@ Global Instance isinj_idmap A : @IsInjective A A idmap
 
 #[export]
 Hint Unfold IsInjective : typeclass_instances.
-Print mapinO_tr_istruncmap.
+
 #[export]
-  Instance IsInjective_mapinO_tr {A B : Type} (f : A -> B)
-  {p : MapIn (Tr (-1)) f} : IsInjective f :=
+Instance isinjective_mapinO_tr {A B : Type} (f : A -> B)
+{p : MapIn (Tr (-1)) f} : IsInjective f :=
```
jdchristensen, line 355: Very minor, but the line starting with `{p` should probably have two spaces of indent, and the `:=` should be at the start of the next line (again, indented by two spaces).

patrick-nicodemus: Yep. I am not sure if this is a shared or universal experience but ProofGeneral has a mind of its own when it comes to indentation choices and it's extremely aggressive about enforcing them. Not sure how to disable this feature.

jdchristensen:
LGTM.

PR #1667, opened 2022-10-10
Speed up natpair_integers.v

jdchristensen:
The change from using `try apply _` on all goals in Z_ring to using `exact _` on just the two goals that are solved speeds up building natpair_integers.v by around 5%.  The removal of `compute` in the tactics saves another 10%.  Overall, the build time of the part of the library that depends on these files is reduced by 7%.


Alizter (approved):

mikeshulman (approved):

jdchristensen:
The `ring_with_nat` tactic is still quite slow, but it is certainly convenient!

PR #1669, opened 2022-10-11
Rewrote proofs to eliminate dependence on funext.

patrick-nicodemus:
I was able to rewrite the code to eliminate any reference to Funext. The original proof was written in a way that required 
`\forall m, P m -> n <= m` to be a mere proposition if `P m` is, I changed this to something like `\bigwedge_{m < n} ~ P m` which is logically equivalent but is a mere proposition without funext.

As I mentioned in a comment, decidable propositions have `|P| -> P`, so for any decidable proposition one could could apply the results of this section to the propositional truncation of `P` and then conclude `P` itself from this. If anybody wants I can change the code so the section does not assume that `P` is a proposition.

### DIFF
--- a/theories/BoundedSearch.v
+++ b/theories/BoundedSearch.v
@@ -1,35 +1,65 @@
-Require Import HoTT.Basics HoTT.Types.
+Require Import HoTT.Basics HoTT.Types HoTT.Types.Bool.
 Require Import HoTT.Truncations.
 Require Import HoTT.HProp HoTT.DProp.
+Require Import HoTT.HProp.
 Require Import HoTT.Spaces.Nat.
 
 Section bounded_search.
-
-  Context `{Funext}.
   Context (P : nat -> Type)
-          {P_hprop : forall n, IsHProp (P n)}
           (P_dec : forall n, Decidable (P n))
+          (* Assuming that P is an HProp is only a technical convenience,
+             as since we have already assumed P is decidable, P is logically 
+             equivalent to its propositional truncation, and the least n s.t. P n
+             is equivalently the least n such that |P n|.
+             So the results in this section can be applied to
+             any decidable predicate on nat. *)
+          {P_hprop : forall n, IsHProp (P n)}
+
           (P_inhab : hexists (fun n => P n)).
   
   (** We open type_scope again after nat_scope in order to use the product type notation. *)
   Local Open Scope nat_scope.
   Local Open Scope type_scope.
-  
-  Local Definition minimal (n : nat) : Type := forall m : nat, P m -> n <= m.
-  Local Definition ishprop_minimal (n : nat) : IsHProp (minimal n).
-  Proof.
-    apply _.
-  Qed.
-  Local Definition min_n_Type : Type := { n : nat & ((P n) * (minimal n))%type}.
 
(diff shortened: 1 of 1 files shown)


theories/BoundedSearch.v
```diff
@@ -1,64 +1,94 @@
...
-          {P_hprop : forall n, IsHProp (P n)}
           (P_dec : forall n, Decidable (P n))
+          (* Assuming that P is an HProp is only a technical convenience,
+             as since we have already assumed P is decidable, P is logically 
+             equivalent to its propositional truncation, and the least n s.t. P n
+             is equivalently the least n such that |P n|.
+             So the results in this section can be applied to
+             any decidable predicate on nat. *)
+          {P_hprop : forall n, IsHProp (P n)}
+
```
jdchristensen, line 17: Remove this blank line.

theories/BoundedSearch.v
```diff
@@ -1,64 +1,94 @@
...
   Local Open Scope type_scope.
-  
-  Local Definition minimal (n : nat) : Type := forall m : nat, P m -> n <= m.
-  Local Definition ishprop_minimal (n : nat) : IsHProp (minimal n).
-  Proof.
-    apply _.
-  Qed.
-  Local Definition min_n_Type : Type := { n : nat & ((P n) * (minimal n))%type}.
 
+  Local Definition minimal (n : nat) : Type := forall m : nat, P(m) -> n <= m.
```
jdchristensen, line 24: Reinstate blank line before this line, and also add blank line after this line.  Also, revert to `P m` instead of `P(m)`.

theories/BoundedSearch.v
```diff
@@ -1,64 +1,94 @@
...
+  Fixpoint minimalB (n : nat) : Bool :=
+    match n with
+    | O => true
+    | S n' => (match dec (P n') with
+               | inl _ => false
+               | inr _ => true end) && minimalB n'
+    end.
+
+  Local Definition min_n_Type : Type := { n : nat & ((P n) * (minimalB n = true))%type}.
+  Proposition minimal_equiv (n : nat) : (minimalB n = true) <-> minimal n.
```
jdchristensen, line 34: Blank line before this line.

theories/BoundedSearch.v
```diff
@@ -1,64 +1,94 @@
...
+  Proof.
+    split.
+    + induction n; [ intros ? ? ?; apply leq_0_n |].
+      intro H. simpl in H.
+      destruct dec; [ discriminate | simpl in H ]. apply IHn in H; clear IHn.
+      intros k pk.
+      (* apply (rwP (andP _ _)) in H; destruct H as [ndB m]; apply (elimN (decP _)) in ndB. *)
+      (* apply IHn in m; intros k Pk. *)
+      (* clear IHn. *)
```
jdchristensen, line 43: Remove?

jdchristensen:
Nice observation.  You replaced `forall m, P m -> n <= m` with a logically equivalent type that is an HProp.  It turns out that one can also achieve this by simply adding a propositional truncation around the type `forall m, P m -> n <= m`.  I illustrate this in #1670.

jdchristensen:
@patrick-nicodemus  Didn't mean to steal this!  I just wondered if a simpler approach would work, and was surprised that it did.  It's still your insight that made this work.

PR #1670, opened 2022-10-11
Avoid Funext in BoundedSearch.v

jdchristensen:
By inserting a propositional truncation, can avoid needing Funext.

### DIFF
--- a/theories/Analysis/Locator.v
+++ b/theories/Analysis/Locator.v
@@ -535,7 +535,6 @@ Section locator.
       {
         apply minimal_n_alt_type.
         - apply QQpos_eq.
-        - apply P_isHProp.
         - apply P_dec.
         - apply P_inhab.
       }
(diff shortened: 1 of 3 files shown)


Alizter (approved):

jdchristensen:
This is inspired by #1669.  That PR also changed `induction` to `destruct` and `Qed` to `Defined` in some places.  Should I include those changes here?

jdchristensen:
Is there an advantage of `destruct` over `induction`?  Simpler proof term?

Alizter:
@jdchristensen induction generates induction hypotheses whereas destruct does a simple match. i.e fixpoint + match vs match.

jdchristensen:
Ok, I cleaned up some other things too.

Alizter (approved):

jdchristensen:
I couldn't resist trying @patrick-nicodemus 's idea about eliminating the assumption that P is a family of HProps.  Another nice observation!

patrick-nicodemus:
Yep, I don't mind! Looks good!

mikeshulman:
I think `induction` and `destruct` do generate different proof terms too: doesn't `induction` use the eliminator whereas `destruct` uses a literal `match`?

PR #1674, opened 2022-10-15
Weaken "Decidable" to "Stable" in hypotheses

patrick-nicodemus:
If P is double-negation stable (as, for example, any negated proposition ~P is), then merely P -> P.
We could retain the original theorem under its name, or we could leave it up to typeclass resolution to fill the gap. Not clear which is better.

### DIFF
--- a/theories/Basics/Decidable.v
+++ b/theories/Basics/Decidable.v
@@ -46,6 +46,11 @@ Proof.
   - apply Empty_rect,dn,n.
 Qed.
 
+Global Instance stable_negation P : Stable (~ P).
+Proof.
+  intros nnnp p; contradiction nnnp; intro np; contradiction np; exact p.
+Defined.
+
 (**
   Because [vm_compute] evaluates terms in [PROP] eagerly
   and does not remove dead code we
--- a/theories/Basics/Decidable.v
+++ b/theories/Basics/Decidable.v
@@ -48,7 +48,8 @@ Qed.
 
 Global Instance stable_negation P : Stable (~ P).
 Proof.
-  intros nnnp p; contradiction nnnp; intro np; contradiction np; exact p.
+  intros nnnp p.
+  exact (nnnp (fun np => np p)).
 Defined.
 
 (**
--- a/theories/Truncations/Core.v
+++ b/theories/Truncations/Core.v
@@ -194,7 +194,7 @@ Proof.
   refine (contr_inhabited_hprop _ ma).
 Defined.
 
-(** A stable type is logically equivalent to its (-1)-truncation. *)
+(** A stable type is logically equivalent to its (-1)-truncation. (It follows that this is true for decidable types as well.) *)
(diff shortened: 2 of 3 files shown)


theories/Basics/Decidable.v
```diff
@@ -46,6 +46,11 @@ Proof.
   - apply Empty_rect,dn,n.
 Qed.
 
+Global Instance stable_negation P : Stable (~ P).
+Proof.
+  intros nnnp p; contradiction nnnp; intro np; contradiction np; exact p.
```
jdchristensen, line 51: I'd prefer to have some of those semicolons be periods, to allow someone to step through the proof.  Or simply use `exact (nnnp (fun np => np p))` after the `intros`, like this:
```suggestion
  intros nnnp p.
  exact (nnnp (fun np => np p)).
```

theories/Truncations/Core.v
```diff
@@ -194,15 +194,14 @@ Proof.
...
+Definition merely_inhabited_iff_inhabited_stable {A} {A_stable : Stable A}
   : Tr (-1) A <-> A.
 Proof.
   refine (_, tr).
   intro ma.
-  destruct A_dec as [a | na].
-  - exact a.
-  - exact (Empty_rec (Trunc_rec na ma)).
+  apply stable; intro na.
+  revert ma; apply Trunc_ind; [exact _ | done].
```
jdchristensen, line 204: You could use `rapply` to avoid the `exact _`, and could make the `done` more explicit for the reader:
```suggestion
  revert ma; rapply Trunc_ind; exact na.
```

jdchristensen (approved):
Another nice observation!  Just a couple of optional suggestions.

jdchristensen:
I think that replacing the theorem is the right way to go.  You could change the comment before it to:
```
(** A stable type is logically equivalent to its (-1)-truncation. (It follows that this is true for decidable types as well.) *)
```

jdchristensen:
BTW, if you want to apply my suggested changes, you can "batch" them into one commit to keep the history more concise.  I'll make my previous comment into a suggested change as well.

theories/Truncations/Core.v
```diff
@@ -194,15 +194,14 @@ Proof.
   refine (contr_inhabited_hprop _ ma).
 Defined.
 
-(** A decidable type is logically equivalent to its (-1)-truncation. *)
-Definition merely_inhabited_iff_inhabited_decidable {A} {A_dec : Decidable A}
+(** A stable type is logically equivalent to its (-1)-truncation. *)
```
jdchristensen, line 197:
```suggestion
(** A stable type is logically equivalent to its (-1)-truncation. (It follows that this is true for decidable types as well.) *)
```

patrick-nicodemus:
> BTW, if you want to apply my suggested changes, you can "batch" them into one commit to keep the history more concise. I'll make my previous comment into a suggested change as well.

oh, ok. I will remember that in the future.

Alizter (approved):

PR #1680, opened 2022-11-07
Make products and sums cumulative.

dgb37:
Based on some discussion on the zulip and my own need, sums and products have been made cumulative, with the associated necessary changes.

### DIFF
--- a/theories/Basics/Datatypes.v
+++ b/theories/Basics/Datatypes.v
@@ -20,6 +20,8 @@ Global Set Primitive Projections.
 Global Set Nonrecursive Elimination Schemes.
 Local Unset Elimination Schemes.
 
+Local Set Polymorphic Inductive Cumulativity.
+
 (** [option A] is the extension of [A] with an extra element [None] *)
 
 Inductive option (A : Type) : Type :=
@@ -35,7 +37,7 @@ Register option as core.option.type.
 
 (** [sum A B], written [A + B], is the disjoint sum of [A] and [B] *)
 
-Cumulative Inductive sum@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} :=
+Inductive sum@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} :=
   | inl : A -> sum A B
   | inr : B -> sum A B.
 
@@ -52,7 +54,7 @@ Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
 
-Cumulative Record prod@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} := pair { fst : A ; snd : B }.
+Record prod@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} := pair { fst : A ; snd : B }.
 
 Scheme prod_rect := Induction for prod Sort Type.
 
--- a/theories/Basics/Datatypes.v
+++ b/theories/Basics/Datatypes.v
@@ -52,7 +52,7 @@ Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
 
-Record prod (A B : Type) := pair { fst : A ; snd : B }.
+Cumulative Record prod@{i j} (A : Type@{i}) (B : Type@{j}) := pair { fst : A ; snd : B }.
 
 Scheme prod_rect := Induction for prod Sort Type.
 
--- a/theories/Basics/Datatypes.v
+++ b/theories/Basics/Datatypes.v
@@ -35,7 +35,7 @@ Register option as core.option.type.
 
 (** [sum A B], written [A + B], is the disjoint sum of [A] and [B] *)
 
-Cumulative Inductive sum (A B : Type) : Type :=
+Cumulative Inductive sum@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} :=
   | inl : A -> sum A B
   | inr : B -> sum A B.
 
@@ -52,7 +52,7 @@ Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
 
-Cumulative Record prod (A B : Type) := pair { fst : A ; snd : B }.
+Cumulative Record prod@{i j} (A : Type@{i}) (B : Type@{j}) : Type@{max(i,j)} := pair { fst : A ; snd : B }.
 
 Scheme prod_rect := Induction for prod Sort Type.
 
--- a/theories/Basics/Datatypes.v
+++ b/theories/Basics/Datatypes.v
@@ -52,7 +52,7 @@ Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
 
-Cumulative Record prod@{i j} (A : Type@{i}) (B : Type@{j}) := pair { fst : A ; snd : B }.
+Cumulative Record prod (A B : Type) := pair { fst : A ; snd : B }.
 
 Scheme prod_rect := Induction for prod Sort Type.
 
--- a/theories/Basics/Datatypes.v
+++ b/theories/Basics/Datatypes.v
@@ -35,7 +35,7 @@ Register option as core.option.type.
 
 (** [sum A B], written [A + B], is the disjoint sum of [A] and [B] *)
 
-Inductive sum (A B : Type) : Type :=
+Cumulative Inductive sum (A B : Type) : Type :=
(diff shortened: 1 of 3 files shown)


theories/Basics/Datatypes.v
```diff
@@ -52,7 +52,7 @@ Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
 (** [prod A B], written [A * B], is the product of [A] and [B];
     the pair [pair A B a b] of [a] and [b] is abbreviated [(a,b)] *)
 
-Record prod (A B : Type) := pair { fst : A ; snd : B }.
+Cumulative Record prod (A B : Type) := pair { fst : A ; snd : B }.
```
Alizter, line 55: Could you make the universes here explicit? Should be `(A : Type@{i}) (B : Type@{j})` and `: Type@{max(i,j)`.

theories/Classes/orders/rings.v
```diff
@@ -347,7 +347,7 @@ Section another_pseudo_ring_order.
   pose proof (projected_pseudo_order f P).
   pose proof (projected_strict_ring_order f P).
   apply from_pseudo_ring_order; try apply _.
-  pose proof (pseudo_order_apart : IsApart R1).
+  pose proof (@pseudo_order_apart R1 H0 R1lt pseudo_srorder_strict : IsApart R1).
```
jdchristensen, line 350: I wonder why this needed to be changed.

theories/PropResizing/Nat.v
```diff
@@ -175,9 +175,9 @@ Section AssumeStuff.
       intros x y.
       destruct x as [x|x]; destruct y as [y|y]; cbn.
       + apply e.
-      + split; apply idmap.
-      + split; apply idmap.
-      + split; apply idmap.
+      + refine (pair@{s s} _ _); apply idmap.
+      + refine (pair@{s s} _ _); apply idmap.
+      + refine (pair@{s s} _ _); apply idmap.
```
jdchristensen, line 180: It's unfortunate that the change to cumulativity makes some proofs uglier (e.g. here, but elsewhere as well), and it's not clear why it happens.  Can we make `split` work again?

theories/PropResizing/Nat.v
```diff
@@ -855,21 +857,22 @@ Section AssumeStuff.
              f (succ_seg n mh) = xs (f ((equiv_N_segment n)^-1 (inl mh)))) *
            (xsn = xs (f (refl_seg n))))). }
       { intros [f xsn].
-        apply equiv_functor_prod'.
-        { apply equiv_concat_l.
+        refine (equiv_functor_prod' _ _).
+        { refine (equiv_concat_l _ _).
```
jdchristensen, line 861: Another spot where it's unclear why this change is needed.  (Also, I would guess that changing `apply` to `rapply` should be sufficient.  Same for other similar changes.)

theories/PropResizing/Nat.v
```diff
@@ -906,18 +909,19 @@ Section AssumeStuff.
...
         refine (equiv_functor_sigma_id (fun b:X => equiv_sigma_prod0 _ _) oE _).
         refine (equiv_sigma_symm _ oE _).
         refine (_ oE (equiv_sigma_prod0 _ _)^-1).
-        apply equiv_functor_sigma_id; intros f0.
+        refine (equiv_functor_sigma_id _). intros f0.
         refine (equiv_functor_sigma_id (fun b:X => equiv_sigma_prod0 _ _) oE _).
         refine (equiv_sigma_symm _ oE _).
         exact ((equiv_sigma_contr _)^-1%equiv). }
     Defined.
+    Set Printing Universes.
```
jdchristensen, line 921:
```suggestion
```

jdchristensen:
See #1657 for another PR that adds cumulativity.  Does it work to use `Local Set Polymorphic Inductive Cumulativity` near the top of Datatypes.v?

theories/PropResizing/Nat.v
dgb37: Initially I made the cumulativity covariant -- this was based on a misunderstanding of why what I needed was not working -- and many more changes were needed. For some reason when I removed these changes it did not go through at Qed/Defined. Of course I may have made a mistake: I shall re-investigate.

theories/PropResizing/Nat.v
```diff
@@ -906,18 +909,19 @@ Section AssumeStuff.
...
     Local Definition partial_Nrec_succ@{}
       := Eval unfold partial_Nrec_succ0
-        in partial_Nrec_succ0@{nr nr nr nr}.
+        in partial_Nrec_succ0@{nr nr}.
```
dgb37, line 920: This is needed for some reason. The other changes must have been a hangover from other unnecessary changes, and have therefore been removed.

theories/Classes/orders/rings.v
Alizter: Most likely the universes of the instances are not aligning properly. If you enabling `Printing Universes` it might highlight the issue. I wouldn't be surprised if there are some incorrect universe restrictions dotted around hottclasses since it was written before Coq's universe polymorphism had matured.

Alizter: Ideally we should fix it so that it is not needed at all.

jdchristensen:
I verified that it works to put `Local Set Polymorphic Inductive Cumulativity.` near the top of the Datatypes.v file, and to remove the `Cumulative` keyword on `sum` and prod`, without any additional changes.  So I recommend making this change.

theories/Classes/orders/rings.v
jdchristensen: I verified that the only needed argument is the last one:
```
  pose proof (@pseudo_order_apart _ _ _ pseudo_srorder_strict : IsApart R1).
```
The desired type is `PseudoOrder R1lt`, but this still fails:
```
  pose proof (@pseudo_order_apart _ _ R1lt _ : IsApart R1).
```
However, typeclass inference works if you break it into two steps:
```
  pose proof (_ : PseudoOrder R1lt).
  pose proof (pseudo_order_apart : IsApart R1).
```
I don't see how that can happen, and couldn't figure out how to do see why it is failing.

Unless someone sees a quick fix, I think we should not hold up this PR for this minor issue.

dgb37:
If we add the `Local Set` &c., do we still want explicit universe annotations akin to what @Alizter requested on all the types defined?

jdchristensen:
> If we add the `Local Set` &c., do we still want explicit universe annotations akin to what @Alizter requested on all the types defined?

To me that sounds orthogonal to whether the definitions are cumulative, so I see no need for such changes to be part of this PR.  But maybe @Alizter has good reasons to ask for this?

Alizter:
If it works without the annotations then it is fine. My suggestion was in case the number of variables needed reducing.

jdchristensen:
> If it works without the annotations then it is fine. My suggestion was in case the number of variables needed reducing.

I checked, and the annotations to `sum` don't change the universe variables.  The only slight difference is that without annotations, the output universe is `max(Set,u,u0)`, but that's equivalent.  So maybe it's better to remove the annotations?

(BTW, I didn't realize that it was even allowed to use `max`.  It is not allowed for the output type of a regular `Definition`, where it would often be useful, but now I see that it is sometimes allowed for an input type.  Interesting.)

jdchristensen:
@dgb37 Would you like me to push the last couple of changes needed so we can merge this?

Alizter:
@jdchristensen I think you should just go ahead so we can merge.

jdchristensen:
I will do it in a day or two.

jdchristensen:
Ok, I think this is ready to merge.  In the end, only three lines are changed by this PR.

jdchristensen:
Is the doc-proviola CI failure anything to do with this PR, or is it unrelated?

jdchristensen:
@Alizter @JasonGross Do you know whether the doc-proviola errors are related to this PR?  It seems like almost every file is giving a python error `'NoneType' object is not callable`.  E.g.:
```
CAMERA html/HoTT.Homotopy.CayleyDickson.html
DEBUG:root:Processing: html/HoTT.Homotopy.CayleyDickson.html
Exception in thread Thread-2 (most likely raised during interpreter shutdown):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/github/workspace/proviola/camera/Popen_noblock.py", line 20, in enqueue_output
  File "/usr/lib/python2.7/Queue.py", line 138, in put
  File "/usr/lib/python2.7/threading.py", line 384, in notify
<type 'exceptions.TypeError'>: 'NoneType' object is not callable
proviola-xml/HoTT.Homotopy.BlakersMassey.xml (real: 28.77, user: 3.35, sys: 0.81, mem: 36968 ko)
CAMERA html/HoTT.Homotopy.ClassifyingSpace.html
DEBUG:root:Processing: html/HoTT.Homotopy.ClassifyingSpace.html
proviola-xml/HoTT.Homotopy.CayleyDickson.xml (real: 25.08, user: 0.79, sys: 0.12, mem: 23200 ko)
```

Alizter:
@jdchristensen they seem to be related due to the frequency. My guess is that there is a bit of syntax (probably new) that causes proviola to fail. Its definitely in the new arithmetic file, but will require some work to find. I can take a look later.

jdchristensen:
Weird---despite all of the python errors, the doc-proviola run is now marked as being successful.  This PR only changes 3 lines and doesn't have the new arithmetic file, so I think it should be merged now.  If I don't hear any disagreement, I'll do so.

(Maybe the doc-proviola runs have been generating errors for a while, but somehow suppress the error codes?)

Alizter:
@jdchristensen In that case, it is probably due to the Github runner running out of memory and killing a process. If it is green now then I think it is good to merge.

PR #1684, opened 2022-11-28
Setoidrewrite

patrick-nicodemus:
Created a file WildCat/SetoidRewrite.v with some typeclass instances to allow rewriting in categories.
Gave some example theorems using the rewriting machinery to demonstrate usage.
Some miscellaneous additions in neighboring files to support this develoment.

Warning: This imports Coq.Setoids.Setoid from the standard library. Currently the setoid rewriting machinery requires this to work, it depends on this file explicitly. This imports the whole standard library into the namespace.

All files that import WildCat/SetoidRewrite.v will also recursively import the entire Coq.Init standard library. In the long term it would be good if this could be avoided.


jdchristensen:
The SetoidRewrite.v file has various comments and `Locate` commands that should be removed before review.  Also, I merged the swap/flip change to Bifunctor, so you could rebase.  The dependence on the standard library is probably going to be an issue, though.

Alizter:
@jdchristensen I'm working on making the setoid rewrite part of Coq be registarable so that we can remove our dependence on the stdlib with it. https://github.com/coq/coq/pull/16907

So hopefully the dependence is temporary.

jdchristensen:
@Alizter Oh, very nice!  So I agree that we can ignore this dependence on the standard library for now.

@patrick-nicodemus Let us know when the other minor cleanups to this PR are done, so it can be reviewed.  I wonder if all of the examples should be in a separate file?

mattam82:
Indeed registration seems the best way to avoid relying on the stdlib. We could also refactor a bit so that you only need to depend on the CRelationClasses and CMorphisms, as the vanilla library relies on Prop which makes no sense for HoTT

patrick-nicodemus:
> @Alizter Oh, very nice! So I agree that we can ignore this dependence on the standard library for now.
> 
> @patrick-nicodemus Let us know when the other minor cleanups to this PR are done, so it can be reviewed. I wonder if all of the examples should be in a separate file?

The minor comments have been added. Examples were moved to Tests.v

theories/WildCat/SetoidRewrite.v
```diff
@@ -0,0 +1,224 @@
+(* -*- mode: coq; mode: visual-line -*-  *)
+
+(* This file contains the definition of the Coq stdlib typeclass_inferences database. It must be imported before Basics.Overture. *)
```
jdchristensen, line 3:
```suggestion
(* Init.Tactics contains the definition of the Coq stdlib typeclass_inferences database. It must be imported before Basics.Overture. *)
```
(Otherwise it might be interpreted as the file SetoidRewrite.v.)

theories/WildCat/SetoidRewrite.v
```diff
@@ -0,0 +1,224 @@
...
+  Classes.CRelationClasses.
+Require Import WildCat.Core.
+Import CMorphisms.ProperNotations.
+
+(* From Coq Require Init.Tactics. *)
+(* Require Import Basics WildCat.Core. *)
+(* From Coq Require  *)
+(*   Setoids.Setoid Classes.CMorphisms *)
+(*   Classes.CRelationClasses. *)
+(* Import CMorphisms.ProperNotations. *)
```
jdchristensen, line 16: Remove?

jdchristensen:
I don't think that Tests.v is the right place for examples.  It's intended to test that the build is good, focusing on specific issues.  Maybe put them in a file WildCat/Examples.v, which we don't export in WildCat.v?  Or maybe people think that the Tests.v file is appropriate?  Probably best to wait to see what others think before you make the change.

I don't know how SetoidRewrite works, so I can't really review this properly.  Hopefully someone else like @Alizter can do it.

theories/WildCat/SetoidRewrite.v
patrick-nicodemus: Yeah, sorry. Changing the order of the dependencies can affect how typeclass resolution works and cause inference to succeed or fail so I was making sure it was stable

theories/Basics/Notations.v
```diff
@@ -222,7 +222,7 @@ Reserved Notation "x \ F" (at level 40, left associativity).
 Reserved Notation "x <> y" (at level 70, no associativity).
 Reserved Notation "x ->> y" (at level 99, right associativity, y at level 200).
 Reserved Notation "x -|-> y" (at level 99, right associativity, y at level 200).
-Reserved Notation "x --> y" (at level 99, right associativity, y at level 200).
```
Alizter, line 225: What is the reason for this change?

Alizter (approved):
LGTM, just a minor comment

theories/Basics/Notations.v
patrick-nicodemus: The setoid rewriting machinery already reserves that notation at level 55. If we have it at a different associativity/precedence level then Coq throws an error saying we are trying to redefine the existing notation at a different level

jdchristensen:
@patrick-nicodemus @Alizter I noticed that if WildCat.SetoidRewrite is imported, various other files have typeclass resolution failures.  For example, Pointed/Core.v and Colimits/Pushout.v are the first two to fail.  Is this something that can be fixed, since otherwise it's not clear how useful it will be to have this.  Or maybe the registration procedure that is planned will help with this?

jdchristensen:
I also noticed that SetoidRewrite.v and the tests build fine with the `From Coq Require Init.Tactics.` line removed.  Should I therefore remove it?

Alizter:
@jdchristensen If you can move it to user-contrib for the time being, I think that would be best. There is also a test that should be moved too.

jdchristensen:
I'll see what @patrick-nicodemus says.  Maybe he can move it and the tests to contrib.  And remove the `Require Init.Tactics`, if that makes sense.  (But even with that removed, other files have build failures when SetoidRewrite is Required.)

patrick-nicodemus:
I will look at this shortly and get back to you.

The problem iirc is that Coq does not have a sophisticated way of resolving
name conflicts between hint databases. If you run "Create HintDb xyz" where
xyz already exists and is populated, then the existing library is
overwritten with a new empty library with the same name, so the old library
is deleted.

Coq.init defines a hintdb called "typeclasses", so if and when this
imported into a file it wipes out anything added to it by an earlier
import. The breakage you mention will occur anywhere where
Wildcat.SetoidRewrite is imported after other things that add hints to the
typeclass db.

Ali suggested that if you move the Require Coq.Init to the root of the
whole repo before we define the typeclass database then this conflict will
be resolved "in our favor" by wiping out the typeclass database that
Coq.init defines.

I did not follow up on this suggestion at the time because I thought that
it was annoying enough having the Coq.init name space open in a handful of
files and Ali's suggestion would mean that it becomes open in all files
rather than only the ones downstream of WildCat.SetoidRewrite.

I do not have a strong commitment to that stance and I am willing to
discuss it.

jdchristensen:
Ok, it seems to work if `From Coq Require Init.Tactics.` is kept as the first `Require` of SetoidRewrite, and if SetoidRewrite is the first Require of any file that Requires it, and moreover if any further file Requires that file, it does so first, etc.  This is consistent with what you wrote.  But it's pretty awkward.  At the very least, this needs to be documented at the top of SetoidRewrite.v.  But I'd also be inclined to put it (and its tests) in contrib for now, until this awkwardness is made unnecessary by changes to Coq.

jdchristensen:
Here are possible comments for the top of the file:
```
(* Typeclass instances to allow rewriting in categories.  See Tests.v for some examples. *)

(* This file Requires Coq.Init, and therefore brings in the Coq standard library.  Because of this, this file needs to be the *first* file Require'd in any file that uses it.  Otherwise, the typeclasses hintdb is cleared, breaking typeclass inference.  Moreover, if Foo Requires this file, then Foo must also be the first file Require'd in any file that Requires Foo, and so on. *)
```

mikeshulman:
> Ali suggested that if you move the Require Coq.Init to the root of the whole repo before we define the typeclass database then this conflict will be resolved "in our favor" by wiping out the typeclass database that Coq.init defines. I did not follow up on this suggestion at the time because I thought that it was annoying enough having the Coq.init name space open in a handful of files and Ali's suggestion would mean that it becomes open in all files rather than only the ones downstream of WildCat.SetoidRewrite.

What would be the practical import of having the Coq.init namespace open everywhere?

Alizter:
@mikeshulman we have some conflicting reserved notations, and for users it can be misleading to have it available. The hope is to completely remove ourselves from the Coq standard library, but for various reasons that isn't entirely possible today. Our demand can be a concrete use-case for the devs to support also.

PR #1685, opened 2022-11-28
Monoidal wildcat

patrick-nicodemus:
Defined a monoidal wildcat.

### DIFF
--- a/theories/Algebra/AbGroups/AbSES/BaerSum.v
+++ b/theories/Algebra/AbGroups/AbSES/BaerSum.v
@@ -51,6 +51,7 @@ Defined.
 Global Instance isbifunctor_abses' `{Univalence}
   : IsBifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
 Proof.
+  unshelve nrapply Build_IsBifunctor; try (typeclasses eauto).
   intros ? ? g ? ? f E; cbn.
   apply abses_pushout_pullback_reorder.
 Defined.
--- a/theories/Algebra/AbGroups/AbSES/BaerSum.v
+++ b/theories/Algebra/AbGroups/AbSES/BaerSum.v
@@ -51,7 +51,7 @@ Defined.
 Global Instance isbifunctor_abses' `{Univalence}
   : IsBifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
 Proof.
-  unshelve nrapply Build_IsBifunctor; try (typeclasses eauto).
+  snrapply Build_IsBifunctor; try (typeclasses eauto).
   intros ? ? g ? ? f E; cbn.
   apply abses_pushout_pullback_reorder.
 Defined.
--- a/theories/Algebra/AbGroups/AbSES/BaerSum.v
+++ b/theories/Algebra/AbGroups/AbSES/BaerSum.v
@@ -51,7 +51,7 @@ Defined.
 Global Instance isbifunctor_abses' `{Univalence}
   : IsBifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
 Proof.
-  snrapply Build_IsBifunctor; try (typeclasses eauto).
+  eapply Build_IsBifunctor.
   intros ? ? g ? ? f E; cbn.
   apply abses_pushout_pullback_reorder.
 Defined.
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -1,0 +1,85 @@
+Require Import Basics Types.
+
+Require Import WildCat.Core WildCat.Prod WildCat.Bifunctor
+  WildCat.Equiv WildCat.NatTrans.
+
+Section Monoidal.
+  Context (C : Type).
+  Context `{Is1Cat C}.
+  Context `{HasEquivs C}.
+  Context (tensor : C -> C -> C).
+  Context `{!IsBifunctor tensor}.
+
+  Definition left_assoc : C -> C -> C -> C :=
+    fun a b c => tensor (tensor a b) c.
+  
+  Definition right_assoc : C -> C -> C -> C :=
+    fun a b c => tensor a (tensor b c).
+
+  Let right_assoc' := uncurry (uncurry (right_assoc)).
+  Let left_assoc' := uncurry (uncurry (left_assoc)).
+  
+  #[export] Instance Is0Functor_right_assoc
+    : Is0Functor right_assoc'.
+  Proof.
+    nrapply Build_Is0Functor.
+    intros [[a1 b1] c1] [[a2 b2] c2] [[f g] h];
+      cbn in f, g, h.
+    exact (fmap11 tensor f (fmap11 tensor g h)).
+  Defined.
+
+  #[export] Instance Is0Functor_left_assoc : Is0Functor left_assoc'.
+  Proof.
+    nrapply Build_Is0Functor.
+    intros [[a1 b1] c1] [[a2 b2] c2] [[f g] h];
+      cbn in f, g, h.
+    exact (fmap11 tensor (fmap11 tensor f g) h).
+  Defined.
+
+  (* Left to right is the convention in Mac Lane. *)
+  Class Associator := assoc : NatEquiv right_assoc' left_assoc'.
+
+  Notation "a ⊗ b" := (tensor a b).
+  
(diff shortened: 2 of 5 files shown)


theories/WildCat/Monoidal.v
```diff
@@ -0,0 +1,92 @@
...
+  
+
+Section Monoidal.
+  Context (C : Type).
+  Context `{Is1Cat C}.
+  Context `{HasEquivs C}.
+  Context (tensor : C -> C -> C).
+  Context `{forall a, Is0Functor (tensor a)}.
+  Context `{forall b, Is0Functor (flip tensor b)}.
+  Context `{@IsBifunctor _ _ _ _ _ _ _ _ _ tensor _ _}.
```
jdchristensen, line 13: The last line can be written as
```
  Context `{!IsBifunctor tensor}.
```
But you aren't using the backtick feature for these (which brings in other typeclass variables), so maybe it's better to just give all of these names?
```
  Context {H0 : forall a : C, Is0Functor (tensor a}).
  Context {H1 : forall a : C, Is0Functor ((swap tensor) a)}.
  Context {H2 : IsBifunctor tensor}.
```
But ultimately, I think we should fix up `IsBifunctor` as discussed on zulip to avoid needing to be so verbose.

jdchristensen (commented):
I guess this is still WIP?  If so, and you are just asking for feedback, you should mark it as WIP and explicitly tell us what you'd like feedback on.

patrick-nicodemus:
> I guess this is still WIP? If so, and you are just asking for feedback, you should mark it as WIP and explicitly tell us what you'd like feedback on.
@jdchristensen 
I was intending for this to be a self-contained commit, the two aborted proofs are just todos for future work. I think it would be easier to prove those by setoid rewriting so I omitted them until later on after we get that functionality added, for modularity it seemed easier to break it up into two unrelated branches without one depending on the other. If you prefer not to have unproven statements in the library I'll just delete them and worry about it later, I thought that the statements of the theorems could serve as documentation of the form "We want to prove this"

theories/WildCat/Monoidal.v
patrick-nicodemus: I have edited the definition of IsBifunctor based on your suggestion. I have deleted the other one and simply inlined it.

theories/WildCat/Monoidal.v
```diff
@@ -0,0 +1,85 @@
...
+  
+  Definition right_assoc : C -> C -> C -> C :=
+    fun a b c => tensor a (tensor b c).
+
+  Let right_assoc' := uncurry (uncurry (right_assoc)).
+  Let left_assoc' := uncurry (uncurry (left_assoc)).
+  
+  #[export] Instance Is0Functor_right_assoc
+    : Is0Functor right_assoc'.
+  Proof using Type*.
```
jdchristensen, line 24: We haven't used `Proof using` elsewhere in the library, and IIUC this is the same as the default behaviour.  Is there a reason to use it here (and in the other places)?

theories/WildCat/Monoidal.v
```diff
@@ -0,0 +1,85 @@
...
+  Class RightUnitor := right_unitor : NatEquiv (flip tensor I) idmap.
+
+  Definition TriangleLaw {assoc : Associator}
+    `{LeftUnitor} `{RightUnitor} (a c : C)
+    :=  fmap (tensor a) (left_unitor c)
+    $== fmap (flip tensor c) (right_unitor a) $o assoc (a, I, c).
+               
+  Class MonoidalStructure
+    `{forall a, Is1Functor (tensor a)}
+    `{forall b, Is1Functor (flip tensor b)}
```
jdchristensen, line 62: Don't we need tensor to be a bifunctor?  (If not, then should that assumption be removed from the context at the top?)

jdchristensen:
> I was intending for this to be a self-contained commit, the two aborted proofs are just todos for future work.

Ok, sounds good.  If you can describe your intent in the description part of the PR, that will make things easier to review in the future.

I think this looks good, modulo the minor issues I pointed out.

theories/WildCat/Monoidal.v
patrick-nicodemus: Oh. I didn't realize this was equivalent to the default. I don't have a strong justification other than I saw it recommended in some Coq documentation. I'm fine with omitting this.

theories/WildCat/Monoidal.v
patrick-nicodemus: An `IsBifunctor` is composed of two 0-functors which are joined by a coherence condition. So for a monoidal category we should additionally require that tensoring on both sides is a 1-functor. 

Maybe we should name these concepts `Is0Bifunctor` and `Is1Bifunctor` respectively for readability?

jdchristensen: Ah, sorry, I didn't notice that it had a 1, not a 0.

jdchristensen:
I think all of my suggestions have been taking into account, so LGTM.

theories/Algebra/AbGroups/AbSES/BaerSum.v
```diff
@@ -51,6 +51,7 @@ Defined.
 Global Instance isbifunctor_abses' `{Univalence}
   : IsBifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
 Proof.
+  unshelve nrapply Build_IsBifunctor; try (typeclasses eauto).
```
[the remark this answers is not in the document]
patrick-nicodemus: Actually, it turns out to be much slower. I don't know why that is. I have replaced it with `eapply Build_IsBifunctor.`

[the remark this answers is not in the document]
jdchristensen: So just this one spot was slow?  Sometimes typeclass inference wastes time trying to do something impossible, but I would have thought that `typeclasses eauto` would have had the same issue.  ...  Oh, this seems to be a weird glitch where `(refine Build_IsBifunctor _ ... _)` with one too few underscores is slow to fail, but it is fast with the right number.  So `snrapply' Build_IsBifunctor` also handles this quickly.  (The primed versions try lots of arguments first.)  I don't think the primed versions have ever been used in the library, but this is one place that could use them.  But `eapply` is completely fine as well.

jdchristensen:
Somehow your merge caused a build problem.

theories/WildCat/SetoidRewrite.v
```diff
@@ -114,11 +114,9 @@ Defined.
...
   `{Is1Cat A, Is1Cat B, Is1Cat C}
   (F : A -> B -> C)
-  `{forall a, Is0Functor (F a)}
-  `{forall a, Is1Functor (F a)} 
-  `{forall b, Is0Functor (Forall.flip F b)}
-  `{forall b, Is1Functor (Forall.flip F b)}  
-  `{@IsBifunctor A B C _ _ _ _ _ _ F _ _} 
+  `{!IsBifunctor F}
+  `{forall a, Is1Functor (F a)}
+  `{forall b, Is1Functor (Forall.flip F b)}
```
jdchristensen, line 119: I'm not sure how you are able to access `Forall.flip` without having imported `Forall`.  But it's probably best to add `Types.Forall` to the `Require Import Basics.` line, and then just use `flip` here.

jdchristensen (approved):

theories/WildCat/SetoidRewrite.v
patrick-nicodemus: Ok. I put some of the `Require Import` statements lower down in the file closer to where they are relevant.

jdchristensen:
I have already approved the changes, so no need to request my review again.

jdchristensen:
(And yes, all imports should be at the top of the file.)

patrick-nicodemus:
> I have already approved the changes, so no need to request my review again.

@jdchristensen It was my assumption/impression that if I pushed a new commit, Github would automatically invalidate your review and approval. That's why I pinged you a second time. My apologies if I misunderstood this

jdchristensen:
We don't limit ourselves to github's approval mechanism.  If two people type "LGTM" and no non-trivial issues came up in the meantime, that's good enough.  And if one person says "LGTM" and 48 hours pass with no complaints, that's also good enough.  (See https://github.com/HoTT/Coq-HoTT/blob/master/STYLE.md#1122-approval-of-pull-requests for the full details.)

PR #1694, opened 2023-01-08
Fine-grained Require commands

jdchristensen:
I put a wrapper around coqc to make note of which parts of the build were bottlenecks to parallelization, and then adjusted things to make this better.  In most cases, the adjustment was a slight change to `Require` commands, e.g. only Requiring some parts of Basics instead of all of it.  Overall, this PR allows me to build the whole library 15% faster with -j8 on an eight core machine.

Around 10% of that is just from the change that removes AbSES from AbGroups, since most users of AbGroups don't use AbSES, and AbSES is slow to build.

The other changes are smaller, but accumulate to give the remaining 5%.  I can undo some of them if people don't like something.  I think some of them will give more benefit with even more cores, but I haven't tested that.

### DIFF
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -1081,7 +1081,7 @@ Definition Book_7_2_2 := @HoTT.HSet.ishset_hrel_subpaths.
 (* ================================================== thm:path-truncation *)
 (** Theorem 7.3.12 *)
 
-Definition Book_7_3_12 := @HoTT.Truncations.Core.equiv_path_Tr.
+Definition Book_7_3_12 := @HoTT.Truncations.SeparatedTrunc.equiv_path_Tr.
 
 (* ================================================== lem:truncation-le *)
 (** Lemma 7.3.15 *)
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -57,7 +57,8 @@
 
 *)
 
-From HoTT Require Import HoTT.
+From HoTT Require Import Basics Truncations Idempotents Spaces.Spheres Spaces.No
+     HIT.V HIT.Flattening Homotopy.WhiteheadsPrinciple.
 From HoTT Require Categories.
 From HoTT Require Import
   Metatheory.IntervalImpliesFunext
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -23,11 +23,10 @@
 
 *)
 
-From HoTT Require Import HoTT.
-From HoTT Require Import Spaces.Nat.
-From HoTT Require Import
-  Metatheory.Core
-  Metatheory.FunextVarieties.
+From HoTT Require Import Basics Types HProp HSet Projective
+     TruncType Truncations Modalities.Notnot Modalities.Open Modalities.Closed
+     BoundedSearch Equiv.BiInv Spaces.Nat Spaces.Torus.TorusEquivCircles
+     Metatheory.Core Metatheory.FunextVarieties.
 
 Local Open Scope nat_scope.
 Local Open Scope type_scope.
--- a/theories/Algebra/AbGroups.v
+++ b/theories/Algebra/AbGroups.v
@@ -1,5 +1,7 @@
 (** Theory *)
 
+Require Export HoTT.Algebra.Groups.
+
 Require Export HoTT.Algebra.AbGroups.AbelianGroup.
 Require Export HoTT.Algebra.AbGroups.Abelianization.
 Require Export HoTT.Algebra.AbGroups.AbPullback.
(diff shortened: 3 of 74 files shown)


jdchristensen:
There were three parts of the build that had poor parallelization.  The first section (mainly Basics, Types, Wildcat and Categories), a middle section while the Modalities folder was being built (because many things depend on Truncations.Core, which used to depend on Modalities.Descent), and the end stage (including things that depend on AbGroups, which used to depend on AbSES).  After this PR, the first two are much better (but not perfect), but there is still a section at the end with only a couple of cores being used.  The good news is that the things being built towards the end will probably not be prerequisites for much other work, so other additions will probably not cause the build time to increase, as they will use the idle cores.

contrib/HoTTBook.v
```diff
@@ -57,7 +57,8 @@
 
 *)
 
-From HoTT Require Import HoTT.
+From HoTT Require Import Basics Truncations Idempotents Spaces.Spheres Spaces.No
```
Alizter, line 60: I think that we don't even want to import these right? We already try to fully qualify the names.

contrib/HoTTBookExercises.v
```diff
@@ -23,11 +23,10 @@
 
 *)
 
-From HoTT Require Import HoTT.
-From HoTT Require Import Spaces.Nat.
-From HoTT Require Import
-  Metatheory.Core
-  Metatheory.FunextVarieties.
+From HoTT Require Import Basics Types HProp HSet Projective
```
Alizter, line 26: Same comment about importing here.

theories/Algebra/AbGroups.v
```diff
@@ -1,13 +1,16 @@
...
+
 Require Export HoTT.Algebra.AbGroups.AbelianGroup.
 Require Export HoTT.Algebra.AbGroups.Abelianization.
 Require Export HoTT.Algebra.AbGroups.AbPullback.
 Require Export HoTT.Algebra.AbGroups.AbPushout.
-Require Export HoTT.Algebra.AbGroups.AbSES.
 Require Export HoTT.Algebra.AbGroups.Biproduct.
 Require Export HoTT.Algebra.AbGroups.AbHom.
 
+(* We do not export AbGroups.AbSES here, to allow the build to be more parallel. *)
```
Alizter, line 12: The indexing in this file is quite dumb as it stands. We could probably remove more. We definitely want `AbelianGroup` and Groups`. Perhaps even `BiProduct`. But things like `Abelianization` are quite specialised and not used so often.

The examples below can stay however, especially since Z is such a useful group.

theories/Basics.v
```diff
@@ -7,6 +6,7 @@ Require Export Basics.Equivalences.
 Require Export Basics.Trunc.
 Require Export Basics.Decidable.
 Require Export Basics.Utf8.
+Require Export Basics.Notations.
```
Alizter, line 9: Does this make things faster? It seems a bit suspect since Notations should be loaded before Overture, but then again Overture already imports it.

Alizter (commented):
Wow looking good @jdchristensen! I have left a few comments.

contrib/HoTTBookExercises.v
jdchristensen: Here it looks like the file uses lots of things without fully qualified names.

theories/Algebra/AbGroups.v
jdchristensen: I usually like it if the summary file contains everything from the corresponding directory, with maybe one or two clearly marked exceptions.  Because of this, I was wondering about moving the AbSES directory to be one level higher...

theories/Basics.v
jdchristensen: The reason I moved it lower in the file is that it used to be Exported by Utf8.  I removed that to allow Utf8 to build in parallel with Notations, but then one file much later in the build failed because `+` was interpreted incorrectly.  Moving the Export of Notations to this spot restores the previous behaviour, since it was  Exported by Utf8 at this point.  (These changes are somewhat silly micro-optimizations that I could reverse.)

jdchristensen:
BTW, I didn't mention explicitly that this PR creates a new file `Truncations/SeparatedTrunc.v`, which contains the last bit of what was in `Truncations/Core.v`.  Only that last bit depended on `Modalities.Descent`, so separating it out allowed me to have some other files build in parallel with the Modalities library (just needing `Modalities.ReflectiveSubuniverse` and/or `Modalities.Modality`).

jdchristensen:
@Alizter @jarlg Any thoughts on moving the AbSES directory one level higher?  I'm inclined to do so unless someone objects.  And then I'll merge this, unless someone objects.

jarlg:
Fine by me.

PR #1695, opened 2023-01-15
Setoid rewrite isolate

patrick-nicodemus:
This addresses breakage problems with rewriting by moving SetoidRewrite to the /contrib/ section.


contrib/SetoidRewrite.v
```diff
@@ -1,6 +1,13 @@
...
 (* Init.Tactics contains the definition of the Coq stdlib typeclass_inferences database. It must be imported before Basics.Overture. *)
+
+(** Warning: This imports Coq.Setoids.Setoid from the standard library. Currently the setoid rewriting machinery requires this to work, it depends on this file explicitly. This imports the whole standard library into the namespace.
+
+All files that import WildCat/SetoidRewrite.v will also recursively import the entire Coq.Init standard library.  *)
+
+(** Because of this, this file needs to be the *first* file Require'd in any file that uses it.  Otherwise, the typeclasses hintdb is cleared, breaking typeclass inference.  Moreover, if Foo Requires this file, then Foo must also be the first file Require'd in any file that Requires Foo, and so on. In the long term it would be good if this could be avoided.*)
+
 From Coq Require Init.Tactics.
 Require Import Basics.Overture Basics.Tactics.
```
Alizter, line 12:
```suggestion
From HoTT Require Import Basics.Overture Basics.Tactics.
```

Alizter: For all the requires below.

contrib/SetoidRewrite.v
```diff
@@ -214,3 +221,54 @@ Proof.
   apply IsMonic_HasRetraction in X.
   apply X in eq_Gf_Gg. assumption.
 Defined.
+Section SetoidRewriteTests.
```
jdchristensen, line 224:
```suggestion

Section SetoidRewriteTests.
```

jdchristensen (commented):
Can you also update the comment in WildCat.v?  Something like:
```
(* See also contrib/SetoidRewrite.v for tools that can be used for rewriting in wild categories. *)

File: contrib/SetoidRewrite.v [hunk 3]
```diff
@@ -1,12 +1,19 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 
 (* Init.Tactics contains the definition of the Coq stdlib typeclass_inferences database. It must be imported before Basics.Overture. *)
```
jdchristensen, line 3: Can you add a comment before all other comments which says what the purpose of the file is?

jdchristensen: Also, can you collapse the stretches of consecutive blank lines (later in the file) down to one or two blank lines, as appropriate?

jdchristensen:
@patrick-nicodemus I was experimenting with editing files within github and made a couple of changes, which you'll have to pull.  I couldn't see a way to make the change I requested to WildCat.v, so that's the only thing left for you to do.

jdchristensen:
I noticed that SetoidRewrite.v contains three Propositions near the end that don't involve typeclasses and could maybe find a home in another file in the WildCat directory.

patrick-nicodemus:
> I noticed that SetoidRewrite.v contains three Propositions near the end that don't involve typeclasses and could maybe find a home in another file in the WildCat directory.

They don't mention typeclasses in their statements but in the proof they invoke the rewriting machinery.

I tried rewriting it without setoid rewriting for the time being and it just gets so unattractive that I don't think it's worth it. The second to last line of the proof of `IsEpic_HasSection` uses rewriting to efficiently convert 
`g $o f $o right_inverse $== h $o f $o right_inverse` into `g $==h` by cancelling out the right inverse to f.
Without the setoid rewriting machinery, I am not sure there is a clean way to do it. Something like

```
  assert (t:= (eq_gf_hf $o (symmetry _ _ (cat_assoc (right_inverse) f g)))).
  clear eq_gf_hf.
  
  assert (t1 := (cat_assoc right_inverse f h) $o t).
  clear t.
  assert (t2 := @fmap _ _ _ _ _  (@is0functor_postcomp _ _ _ _ _ _ _ _ g)
                  (f $o right_inverse) (Id b) is_section).
  clear t1.
```

jdchristensen:
> They don't mention typeclasses in their statements but in the proof they invoke the rewriting machinery.

Oh, of course, I should have realized that!  I didn't mean that you should redo the proofs.  I'll merge as soon as the CI is done.  Thanks for this!

jdchristensen:
We should remember to move SetoidRewrite.v back to the WildCat folder once we can make it not bring in all of the Coq standard library.

PR #1696, opened 2023-01-19
fix the notation "(.* x )"

jarlg:
This commit fixes the following bug with the notation `(.* x)`, intended to be short for `fun y => sg_op y x`. Previously, the relevant scopes hadn't been opened, causing the notation to use the function `prod` instead of `sg_op`. This caused issues when used in combination with type_scope, in which `prod` refers to the product of types. One can verify the issue by running `Locate "(.* x )".` right after the notation has been defined on the master branch.

Opening the relevant scopes remedies the issue. (Another solution would be to write `sg_op` explicitly, but this solution is a bit more modular, and is probably the original intention.)

### DIFF
--- a/theories/Classes/interfaces/canonical_names.v
+++ b/theories/Classes/interfaces/canonical_names.v
@@ -102,16 +102,20 @@ Hint Extern 4 (Apart (Pos _)) => apply @sig_apart : typeclass_instances.
 
 (* Notations: *)
 Declare Scope mc_add_scope.
+Open Scope mc_add_scope.
 Infix "+" := sg_op : mc_add_scope.
 Notation "(+)" := sg_op (only parsing) : mc_add_scope.
 Notation "( x +)" := (sg_op x) (only parsing) : mc_add_scope.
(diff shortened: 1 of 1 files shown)


jdchristensen:
I don't think the scope needs to be open when the notation is defined, only when it is used.  For example, this works for me:
```
Open Scope mc_mult_scope.
Locate "(.* x )".
```
But maybe I am misunderstanding.

jarlg:
Those commands return

```coq
Notation "(.* x )" := (fun y => prod y x) : mc_mult_scope (default interpretation) (only parsing)
```
but `prod` should be `sg_op`.

jdchristensen:
Oh, I see.  It looks like the error is that the RHS of the notation uses `*` instead of `sg_op`, etc.  I think correcting that might be the right fix.

jarlg:
That was the other solution I referred to in the last parenthesis of the first message. I changed to that one.

jdchristensen:
Thanks, I like this since it parallels the notations on the lines just before the changes.  LGTM.

Alizter (approved):

PR #1697, opened 2023-01-20
H-space theory

jarlg:
We develop a bunch of H-space theory and related things. Most notably, in `Homtopy/HSpace/Moduli.v` we construct an equivalence between the "moduli type" of (coherent) H-space structures on a pointed type `A` and `A ->* (A ->** A)`, whenever `A` is a left-invertible (coherent) H-space. Left-invertibility means that left multiplication by any element is an equivalence, and coherence means that the left and right identities of the H-space agree at the base point. This formula generalizes a classical formula of Arkowitz--Curjel and Copeland and appears as Theorem 2.27 in this preprint: https://arxiv.org/abs/2301.02636.

Another result worth mentioning is `phomotopy_from_homotopy` and friends from `Homotopy/HSpace/Core.v` which gives a general condition on `A` that yields a map `f == g -> f ==* g` for any two pointed functions `f g : B ->* A`. As far as I know, this fact was first observed by Evan Cavallo when `A` is a homogeneous type. The more general condition we use here was observed by Egbert Rijke.

I've written comments in the files so I hope they are mostly self-explanatory!


jdchristensen:
This is great.  I'll look at it closely over the next day or two.  Since it's fairly substantial, let's not merge it until at least two of us have looked it over.  BTW, the file `HSpace/Core.v` contains substantial changes compared to the original `HSpace.v`, so it needs review as well:  `233 insertions(+), 40 deletions(-)`.

theories/Homotopy/HSpace/Core.v
```diff
@@ -0,0 +1,244 @@
...
+
+Definition equiv_hspace_left_op_pt {X : pType} `{IsHSpace X}
+  : X <~> X := Build_Equiv _ _ (pt *.) _.
+
+(** Any element of an H-space defines a pointed self-map by left multiplication. *)
+Definition pmap_hspace_left_op {X : pType} `{IsHSpace X} (x : X)
+  : X ->* [X, x] := Build_pMap X [X,x] (x *.) (right_identity x).
+
+Definition equiv_hspace_left_op {A : pType} `{IsHSpace A}
+  `{forall a, IsEquiv (a *.)} (a : A) : A <~> A
```
jdchristensen, line 41: Should we define `LeftInvertible a` or something like that to capture this frequent hypothesis?  Also, in this case we only need it for the particular `a`, so I think the assumption should be weakened.  Then we can add `IsEquiv (pt *.)` as a global instance, and then `equiv_hspace_left_op_pt` is no longer needed.  (Same for right multiplication of course.)

jarlg:
Let me summarize the changes to `HSpace.v` (now `HSpace/Core.v`).  Outside the Section `HSpaceProperties`, nothing changed. I moved `equiv_hspace_left_op` and `equiv_hspace_right_op` out of the section and made them more general by replacing the hypothesis that the H-space `A` is connected by the hypothesis that left and right multiplication is an equivalence. They work as before, since we have the instances `isequiv_hspace_left_op` and `isequiv_hspace_right_op`. (Writing this now I realise those should be renamed to include "connected" in the name.) Then I removed the section and simply in-lined the arguments to those instances, which now live under the heading "Connected H-spaces".

Those were the _changes_, the rest are new additions.

theories/Homotopy/HSpace/Core.v
```diff
@@ -0,0 +1,241 @@
...
+Definition equiv_hspace_right_op {A : pType} `{IsHSpace A}
+  `{forall a, IsEquiv (.* a)} (a : A) : A <~> A
+  := Build_Equiv _ _ (.* a) _.
+
+(** We make [(pt *.)] into a pointed equivalence using the right identity. If we used the left identity, we'd get a map that's equal to the identity as a pointed map; but without coherence (see Coherent.v) this is not necessarily the case for this map. *)
+Definition pequiv_hspace_left_op_pt {X : pType} `{IsHSpace X} : X <~>* X
+  := Build_pEquiv' equiv_hspace_left_op_pt (right_identity pt).
+
+Definition pequiv_hspace_left_op_pt_beta {X : pType} `{IsHSpace X}
+  : pmap_hspace_left_op pt ==* pequiv_hspace_left_op_pt.
```
jdchristensen, line 53: `pmap_hspace_left_op pt` and `pequiv_hspace_left_op_pt` are in fact definitionally equal, so this result is not needed.  Moreover, I think the previous Definition should be generalized to left multiplication by any left-invertible element `a`.

theories/Pointed.v
```diff
@@ -6,3 +6,4 @@ Require Export Pointed.pEquiv.
 Require Export Pointed.pTrunc.
 Require Export Pointed.pHomotopy.
 Require Export Pointed.pSusp.
+Require Export Pointed.pSect.
```
jdchristensen, line 9: Add the other new files too?

jdchristensen (commented):
I'll be back later to review the rest.

theories/Homotopy/HSpace.v
```diff
@@ -1,51 +1,4 @@
...
-    := Build_Equiv _ _ _ (isequiv_hspace_left_op a).
-
-  Definition equiv_hspace_right_op (a : A) : A <~> A
-    := Build_Equiv _ _ _ (isequiv_hspace_right_op a).
-
-End HSpaceProperties.
+Require Export HSpace.Core.
+Require Export HSpace.Coherent.
+Require Export HSpace.Pointwise.
+(* Require Export HSpace.Moduli. *)
```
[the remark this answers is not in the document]
jarlg: My reasoning is that `Moduli.v` does not contain any results the average user of `HSpace` would need, so you'd only import it if you're specifically interested in it.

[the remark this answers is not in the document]
jdchristensen: I think we should still Export it.  If a user wants to minimize their dependencies, they should choose exactly which files to import.  I'll push this change.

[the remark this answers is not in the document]
jarlg: Sure.

jdchristensen:
I won't make any more changes for a while.

theories/Homotopy/HSpace/Core.v
jarlg: I'm not sure where to place that class, if we make it. (For the record, there's `LeftInverse` from `canonical_names.v` which doesn't do what we want.)

I made the corrections you suggested and just pushed them.

theories/Pointed/pModality.v
```diff
@@ -0,0 +1,23 @@
+Require Import Basics Types ReflectiveSubuniverse Pointed.Core.
+
+Local Open Scope pointed_scope.
+
+Local Notation "[ X , x ]" := (Build_pType X x).
```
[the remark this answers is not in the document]
Alizter: old comment nevermind.

theories/Homotopy/HSpace/Core.v
```diff
@@ -0,0 +1,227 @@
+Require Export Classes.interfaces.abstract_algebra.
+Require Import Pointed.
+
+Local Open Scope pointed_scope.
+Local Open Scope trunc_scope.
+Local Open Scope mc_mult_scope.
+
+(** * H-spaces *)
+
+(** A (noncoherent) H-space [X] is a pointed type with a binary operation for which the base point is a both a left and a right identity. (See Coherent.v for the notion of a coherent H-space.) We say [X] is left-invertible if left multiplication by any element is an equivalence, and dually for right-invertible. *)
```
Alizter, line 10: Is the definition of a H-space from AT the coherent or non-coherent one? I always thought it was the latter, and I couldn't in the past find any literature about the "coherent" one. @jdchristensen do you know?

theories/Homotopy/HSpace/Moduli.v
```diff
@@ -0,0 +1,103 @@
...
+  Pointed Homotopy.EvaluationFibration Tactics.EvalIn.
+
+Local Open Scope pointed_scope.
+Local Open Scope mc_mult_scope.
+
+(** * The moduli type of H-space structures *)
+
+(** When [A] is a left-invertible H-space, we construct an equivalence between the ("moduli") type of H-space structures on [A] and the type [A ->* (A ->** A)]. By the smash-hom adjunction for pointed types, due to Floris van Doorn in HoTT, the latter is also equivalent to the type [Smash A A ->* A].
+
+This equivalence generalizes a formula of Arkowitz--Curjel and Copeland for spaces, and appears as Theorem 2.27 in https://arxiv.org/abs/2301.02636v1. *)
```
Alizter, line 11:
```suggestion
This equivalence generalizes a formula of Arkowitz--Curjel and Copeland for spaces, and appears as Theorem 2.27 in https://arxiv.org/abs/2301.02636v1 *)
```

Alizter: Otherwise, the link is wrong.

theories/Pointed/Core.v
```diff
@@ -12,11 +12,14 @@ Generalizable Variables A B f.
 
 (** ** Pointed Types *)
 
+Notation "'pt'" := (point _) : pointed_scope.
+Notation "[ X , x ]" := (Build_pType X x) : pointed_scope.
```
Alizter, line 16: I introduce a similar notation in https://github.com/HoTT/Coq-HoTT/blob/bbfd27086db10b60ce767320c61d690d6ae2a488/theories/Pointed/Loops.v#L330

Could you replace that with your one?

theories/Truncations/Core.v
```diff
@@ -3,6 +3,8 @@
 Require Import Basics Types WildCat.Core WildCat.Universe.
 Require Import TruncType.
 Require Import Modalities.Modality.
+(* Users of this file almost always want to be able to write [Tr n] for both a [Modality] and a [ReflectiveSubuniverse], so they want the coercion [modality_to_reflective_subuniverse]: *)
+Require Export (coercions) Modalities.Modality.
```
Alizter, line 7: This is an 8.16 feature I believe. Should we also merge my 8.16 compat (which mostly removes warnings) since we are really dropping 8.15.

Alizter (approved):
Looking good. Some very nice results. :)

theories/Homotopy/HSpace/Core.v
jarlg: For H-spaces in AT, the base point is a _strict_ two-sided unit, which we can't express in HoTT. If you interpret the HoTT notion of noncoherent H-space into spaces, you get _homotopies_ witnessing the two unit laws. In the coherent case, you also get a homotopy between the two paths from `pt * pt`  to `pt`. This means that an H-space in AT is a coherent H-space in the sense of HoTT, with the coherence simply being a constant homotopy. I don't think the two notions are equivalent in general though, but maybe for certain nice (e.g. well-pointed) spaces. You need some theorem saying that the HoTT notion can be strictified to the AT notion.

theories/Homotopy/HSpace/Moduli.v
jarlg: Thanks! I guess I should improve my emacs to let me click links :)

theories/Pointed/Core.v
jarlg: Done!

theories/Homotopy/HSpace/Core.v
jdchristensen: Nice, those changes simplified things.  And you already had the global instance I mentioned.  Not sure where "LeftInvertible" and "RightInvertible" should go.  Maybe alongside similar things in canonical_names.v?  Does @Alizter have an opinion?

Also, about this part of the file:  maybe it would work to only have `pmap_...` and `pequiv_...` without `equiv_...`?  There is a coercion from `pEquiv` to `Equiv`.

Alizter: My opinion on canonical_names is to move away from having everything in there and put them in more logical places. Since I don't know what those are at the moment, I suggest we keep them here.

theories/Homotopy/HSpace/Core.v
jdchristensen: To summarize, the coherent one in HoTT is the one that should match the classical notion.  It's also better behaved in various ways as we illustrate in [https://arxiv.org/abs/2301.02636](https://arxiv.org/abs/2301.02636).  But the two notions are logically equivalent:  every non-coherent H-space can be upgraded to a coherent one.  However, the type of coherent H-space structures differs from the type of non-coherent H-space structure.

theories/Pointed/Core.v
jdchristensen: Nice find!

theories/Homotopy/HSpace/Core.v
Alizter: I guess what we should do is have some generic properties subsuming the 0-truncated and homotopy versions defined in Algebra somewhere and use them for both things. Not something worth doing in this PR I think though.

jdchristensen:
I ran my script, and for some reason it found 17 Requires it could remove.  Some aren't related to this PR (and I'm not sure why they arose since the last time I ran it), but some are, so I pushed it.  I'm not sure if it's because of this, but the current version of this PR builds faster than the original version, and also faster than the master branch before this PR!  (This is with -j8, so is most likely due to better parallelism.)

jdchristensen:
I simplified a bunch of things.  One observation is that `pequiv_hspace_left_op` is already a proof that an H-space is homogeneous.  I suggest that we move the three Definitions about paths between H-space structures below the part about homogeneous types, so that all of the material about multiplication being an equivalence is together.  Sound good?

jdchristensen (approved):
This looks great!  I'm all done now.  If anyone else wants to look it over, let me know.  Otherwise, I'll merge soon.

jdchristensen:
I'll also wait to see what Jarl thinks about moving the material on paths, and about whether we should introduce `LeftInvertible a`.  What comes up more often is `forall (a : A), IsEquiv (a *.)`, so maybe that's what we should find a name for.  `IsLIHSpace A` and `IsLICohHSpace A` could include that and the rest of the H-space structure.  Or we could have a separate `IsLeftInvertible A`, or something like that.  Or we could leave it all like it is.  But it's going to come up even more in WIP.  Oh, I just realized I already written something about this:
```
Class LeftInvertible {A} {B} {C} (op : A -> B -> C)
  := left_invertible : forall x:A, IsEquiv (op x).

Definition IsLeftInvertibleCohHSpace (A : pType) := { H : IsHSpace A & prod (IsCoherent A) (LeftInvertible hspace_op) }.

Definition issig_isleftinvertiblecohhspace (A : pType)
  : { hspace_op : SgOp A
    & { hspace_left_identity : LeftIdentity hspace_op pt
    & { hspace_right_identity : RightIdentity hspace_op pt
    & { hspace_coh : hspace_left_identity pt = hspace_right_identity pt
    & LeftInvertible hspace_op } } } }
      <~> IsLeftInvertibleCohHSpace A.
Proof.
  unfold IsLeftInvertibleCohHSpace, IsCohHSpace, IsCoherent, LeftInvertible.
  make_equiv.
Defined.
```

jarlg:
How about moving the material on paths to the beginning of `Moduli.v`? I think that's fairly natural.

I agree that there should be a `LeftInvertible` class like that, and I think a current natural home is right after `RightInverse` in `canonical_names.v`. As for `IsLeftInvertibleCohHSpace`, I'm partial to making it a record type, since then you just add a single `Global Existing Instances IsHSpace IsCoherent IsLeftInvertible` and be on your way. (The `issig` remains essentially the same.)

I'll begin making these changes, but let me know if you have any further thoughts.

jdchristensen:
Yes, that all sounds good to me.

jarlg:
One thing that I didn't anticipate is that Coq isn't automatically able to solve a goal of type `IsEquiv (op x)` by applying `left_invertible`. One solution is to add `Hint Immediate left_invertible : typeclass_instances`. Is there another solution to prefer?

Another thing is that I can't get my record solution above to do what I want. Specifically, Coq doesn't understand that a left-invertible H-space is an H-space...

```coq
Record IsLInvHSpace (A : pType) := {
    islinvhspace_ishspace : IsHSpace A;
    islinvhspace_leftinvertible : LeftInvertible (.*.);
  }.
#[export] Existing Instances islinvhspace_ishspace islinvhspace_leftinvertible.

Definition test (A : pType) `{IsLInvHSpace A} : IsHSpace A.
Proof.
  Fail exact _.
  apply islinvhspace_ishspace.
  assumption.
Defined.
```
Of course, one can implement the instances but why doesn't Coq apply the hypothesis `IsLinvHSpace A` in the typeclass search? Maybe @Alizter has an idea?

Alizter:
@jarlg assumptions of `IsLInvHSpace` won't get picked up as they are not classes. You can make it a class and it should work.

jdchristensen:
Looking at the analogous thing with coherence, we have:
```
Class IsCoherent (X : pType) `{IsHSpace X} :=
  iscoherent : left_identity _ = right_identity _.

Record IsCohHSpace (A : pType) := {
    ishspace_cohhspace : IsHSpace A;
    iscoherent_cohhspace : IsCoherent A;
  }.
```
If you have `IsCoherent A` as a hypothesis, then Coq also automatically knows that `A` is an H-space.  But if you have `IsCohHSpace A` as a hypothesis, Coq gets stuck.  So maybe the first way will work for left-invertibility?  (I don't quite understand @Alizter 's comment.)

Alizter:
Hmm maybe I didn't quite understand the situation then.

jarlg:
Ali's comment is to replace `Record` by `Class`, which works as I want. The same should work for `IsCohHSpace`.

jdchristensen:
Oh, good, that's easy.

jarlg:
But making `IsLInvHSpace` a class breaks other things, notably the `make_equiv` in `Coherent.v`. I guess there's a typeclass cycle or something. I need more time to think about how to structure these various notions of H-spaces. Even just the addition of `LeftInvertible` is making certain files spin (e.g. `EMSpace.v`).

I'd like to go merge this now and think about left-invertibility in the next PR.

jdchristensen:
Ok, I agree, let's hold off on that.  Since there are a lot of changes (especially with some of my commits), I'll wait ~12 hours longer before merging, in case anyone else want to chime in.

jdchristensen:
I found a few more things to clean-up, including renaming the phomotopy* things to have more consistent names.  @jarlg , can you check the last commit before I merge?

jarlg:
Thanks! Those are good corrections and clarifications. LGTM

PR #1698, opened 2023-01-24
Use pt and [X, x] notations, psusp, pjoin, G, etc.

jdchristensen:
Mainly replaces `(point _)` with `pt` and `Build_pType (foo) (bar)` with `[foo, bar]`, but also makes use of `psusp`, the coercion of a group to a pointed type, and a new `pjoin`.

Note that `[X, x]` was chosen as the notation for a pointed type rather than `(X, x)` because the latter conflicts with ordered pairs.  The former conflicts with functor categories, but those are used less often than ordered pairs, and generally in different contexts.  Another nice thing is that there are often many nested parentheses, and the square brackets make it easier to see the matching pairs.  Also, this notation often has fewer parentheses than when using`Build_pType`.

### DIFF
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -41,9 +41,8 @@ Definition group_loops (X : pType)
 Proof.
   (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4256 *)
   pose (x0 := point X);
-  pose (BG := (Build_pType
-               { x:X & merely (x = point X) }
-               (exist (fun x:X => merely (x = point X)) x0 (tr 1)))).
+  pose (BG := [{ x:X & merely (x = point X) },
+               exist (fun x:X => merely (x = point X)) x0 (tr 1)]).
   (** Using [cut] prevents Coq from looking for these facts with typeclass search, which is slow and (for some reason) introduces scads of extra universes. *)
   cut (IsConnected 0 BG).
   { exact (Build_ooGroup BG). }
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -40,9 +40,8 @@ Definition group_loops (X : pType)
   : ooGroup.
 Proof.
   (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4256 *)
-  pose (x0 := point X);
-  pose (BG := [{ x:X & merely (x = point X) },
-               exist (fun x:X => merely (x = point X)) x0 (tr 1)]).
+  pose (BG := [{ x:X & merely (x = pt) },
+               exist (fun x:X => merely (x = pt)) pt (tr 1)]).
   (** Using [cut] prevents Coq from looking for these facts with typeclass search, which is slow and (for some reason) introduces scads of extra universes. *)
   cut (IsConnected 0 BG).
   { exact (Build_ooGroup BG). }
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -48,7 +48,7 @@ Proof.
   cut (IsConnected 0 BG).
   { exact (Build_ooGroup BG). }
   cut (IsSurjection (unit_name (point BG))).
-  { intros; refine (conn_pointed_type (point _)). }
+  { intros; refine (conn_pointed_type pt). }
   apply BuildIsSurjection; simpl; intros [x p].
   strip_truncations; apply tr; exists tt.
   apply path_sigma_hprop; simpl.
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -39,7 +39,6 @@ Coercion group_type : ooGroup >-> Sortclass.
 Definition group_loops (X : pType)
   : ooGroup.
 Proof.
-  (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4256 *)
   pose (BG := [{ x:X & merely (x = pt) },
                exist (fun x:X => merely (x = pt)) pt (tr 1)]).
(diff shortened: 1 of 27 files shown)


theories/Algebra/ooGroup.v
```diff
@@ -41,14 +41,13 @@ Definition group_loops (X : pType)
 Proof.
   (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4256 *)
   pose (x0 := point X);
-  pose (BG := (Build_pType
-               { x:X & merely (x = point X) }
-               (exist (fun x:X => merely (x = point X)) x0 (tr 1)))).
+  pose (BG := [{ x:X & merely (x = point X) },
```
Alizter, line 44: Can you replace `point X` with pt here?

theories/Pointed/Core.v
```diff
@@ -31,7 +31,7 @@ Global Instance ispointed_prod `{IsPointed A, IsPointed B} : IsPointed (A * B)
   := (point A, point B).
 
 (** We override the notation for products in pointed_scope *)
-Notation "X * Y" := (Build_pType (X * Y) ispointed_prod) : pointed_scope.
+Notation "X * Y" := ([(X * Y), ispointed_prod]) : pointed_scope.
```
Alizter, line 34: Do we need the brackets around the product here?

Alizter (approved):
Looking good

theories/Algebra/ooGroup.v
jdchristensen: Yes, that works.  And in fact I can get rid of the `pose (x0 := point X)` and replace `x0` with `pt` as well.

There are also a few places later in the file that use `(point X)`.  In general, I only replaced `(point _)` with `pt`, assuming that if someone explicitly wrote the type, it was either to help the reader or to help Coq.  But if you think I should replace `(point X)` with `pt` in this file I will.

theories/Pointed/Core.v
jdchristensen: Nope, good catch.  I'll fix it.

theories/Algebra/ooGroup.v
Alizter: As long as it is immediately obvious whose `pt` it is, I think it is fine to replace. If we are talking about two pTypes `A` and `B` then it might not be the best.

jdchristensen: I now see that the odd construction with `x0` that was used here was to work around a bug in Coq 8.5.  No workaround is needed now, so I also removed the comment about the bug.

theories/Algebra/ooGroup.v
```diff
@@ -39,7 +39,6 @@ Coercion group_type : ooGroup >-> Sortclass.
 Definition group_loops (X : pType)
   : ooGroup.
 Proof.
-  (* Work around https://coq.inria.fr/bugs/show_bug.cgi?id=4256 *)
```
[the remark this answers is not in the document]
jdchristensen: I don't think I'll mix that into this PR, since it isn't about pointedness.  Feel free to do it.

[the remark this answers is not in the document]
Alizter: OK, in that case can you add the work around comment back in?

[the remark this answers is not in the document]
jdchristensen: Interestingly, the entire proof can be replaced by
```
  Require Import Homotopy.Cover.
  exact (Build_ooGroup (pcomp X pt) _).
```
and it only uses 2 universes instead of 12.  But then some other parts would need to change a bit, since `pcomp` defines the component using `tr x = tr pt` instead of `merely (x = pt)`.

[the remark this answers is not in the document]
Alizter: I can try it in another PR if you like.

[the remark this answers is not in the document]
jdchristensen: As far as I can see, the comment about the Coq bug was only referring to the line `pose (x0 := point X)`.  Since that line is gone, I think it is correct to remove the comment.  I think the remaining comment about `cut` covers the other workarounds.

If you are able to clean this up in another PR, that would be great.

BTW, the reason that `pcomp` uses a different definition of component is that `pcomp`'s version is a special case of `pfiber`, which means that you can use all of the general purpose results about fibers.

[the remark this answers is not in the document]
Alizter: I tried thinking about the functoriality of pfiber, since this is slightly different to the functoriality of hfiber (just with a fixed point). Am I correct in thinking that it is a functor from (2 -> pType) -> pType? I couldn't work out what composed commutative squares become in the target however.

jdchristensen:
Are we good to merge this?  Can you give it a quick check @jarlg ?

Alizter (approved):
I'm happy with merging this.

jarlg:
I'll give it a quick check, just give me a moment.

theories/Homotopy/Bouquet.v
```diff
@@ -22,7 +22,7 @@ Section AssumeUnivalence.
   Proof.
     rapply isconnected_susp.
     rapply contr_inhabited_hprop.
-    apply tr; exact (point _).
+    apply tr; exact pt.
```
jarlg, line 25:
```suggestion
    exact (tr pt).
```

theories/ObjectClassifier.v
```diff
@@ -39,7 +39,7 @@ Definition equiv_sigma_fibration `{Univalence} {Y : Type@{u}}
 
 (** We construct the universal square for the object classifier. *)
 Local Definition topmap {A : Type} (P : A -> Type) (e : sig P) : pType
-  := Build_pType (P e.1) e.2.
+  := [(P e.1), e.2].
```
jarlg, line 42:
```suggestion
  := [P e.1, e.2].
```

jarlg (changes requested):
Only two stylistic changes. Looks good to me!

PR #1699, opened 2023-01-24
Fix obligations warning

Alizter:


jdchristensen:
I don't see a warning for Basics/Tactics.v.  Does it only happen with certain versions of Coq?  I was hoping that this would get rid of
```
*** Warning: in file theories/Basics/Notations.v, declared ML module ltac_plugin has not been found!
*** Warning: in file theories/Basics/Overture.v, declared ML module number_string_notation_plugin has not been found!
```

Alizter:
@jdchristensen That won't be rid of for a while since it is a bug in coqdep (some nonsense about legacy plugin loading method + noinit). The warning appears on master. If you look in changed files, github no longer warns about it like it does for other PRs.

PR #1700, opened 2023-01-24
Shorten proofs in ooGroup.v

Alizter:
I've also added some convenience lemmas to Sigma.

### DIFF
--- a/theories/Algebra/ooGroup.v
+++ b/theories/Algebra/ooGroup.v
@@ -1,7 +1,9 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 Require Import Pointed.
 Require Import Homotopy.ClassifyingSpace.
+Require Import Homotopy.Cover.
 Require Import Algebra.Groups.
+Require Import HFiber.
 Require Import WildCat.
 
 Local Open Scope trunc_scope.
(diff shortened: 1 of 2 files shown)


theories/Algebra/ooGroup.v
```diff
@@ -80,13 +65,14 @@ Definition group_loops_functor
            {X Y : pType} (f : X ->* Y)
 : ooGroupHom (group_loops X) (group_loops Y).
```
jarlg, line 66: This statement is a special case of the fact that any pointed function can be restricted to the (O-)components of the respective base points. We should just add that general statement to `Homotopy.Cover` and give a one-line proof here.

Here's a candidate for the general result:
```coq
Definition pfunctor_O_pcover `{O : ReflectiveSubuniverse} {X Y : pType}
  (f : X ->* Y) : O_pcover O X pt ->* O_pcover O Y pt.
Proof.
  snrapply Build_pMap.
  - intros [x q].
    exists (f x).
    transitivity (O_functor O f (pto O [X, pt] x)).
    { cbn. symmetry.
      exact (O_rec_beta _ _). }
    refine (ap (O_functor O f) q @ _).
    refine (O_rec_beta _ _ @ _).
    exact (ap _ (point_eq f)).
  - cbn. snrapply path_sigma'.
    1: apply point_eq.
    refine (transport_paths_Fl _ _ @ _).
    apply moveR_Vp.
    refine (_ @ (concat_p1 _)^).
    apply moveR_Vp.
    exact (concat_1p _).
Defined.
```
then you can do:
```coq
Definition group_loops_functor
           {X Y : pType} (f : X ->* Y)
  : ooGroupHom (group_loops X) (group_loops Y)
  := pfunctor_O_pcover f.
```

jarlg: What we should probably do is define `pto_O_natural` in `pModality.v` (this would also require the equivalent of `pTr` for a general RSU). Then `pfunctor_O_pcover` above would be a simple application of `pfunctor_pfiber`.

I can PR that and some more stuff about covers later today.

jdchristensen: @jarlg I have lots of stuff formalized about n-connected covers that I should show you when I get a chance.  I never committed it because I suspect that much of it generalizes to O-covers for any modality (or RSU?) O, but I haven't had time to work that out.

jdchristensen:
Currently, `BAut X` is defined using `merely (Z = X)`, instead of `tr Z = tr X`.  If the definition was changed, then maybe some results there would be special cases of things in Cover.v, or could be generalized to Cover.v.  I didn't address this at all in #1702.  @jarlg did some work on BAut.v last year, so he probably has thoughts on whether this would be reasonable.  It might mean that Univalence is needed again in `baut_reduce` ...   Anyways, just mentioning it in case it makes sense to make changes there while changing ooGroup in a similar way.

jarlg:
Unifying `BAut` and `comp`, which is what you're suggesting, sounds like a good idea to me. I'd have to spend a bit more time to be sure.

I remember alternative definitions of `BAut` being discussed in #1452. I think the suggestion you're making here is somewhat orthogonal though, since we could define `O Z` to be `tr Z = tr X` (if `Subuniverse` allowed that). So if `O Z` could live in a universe above `Z`, we'd unify `comp Type X`, `BAut X`, and `Type_ O`. That'd be neat!

jdchristensen:
`comp` and `BAut` can be defined in either the `merely (_ = _)` style or the `tr _ = tr _` style.  But if we want to unify with `O_cover`, we need to use the second style.  Sometimes minor changes like this can make a big difference when formalizing, but I can't tell in this case if that would be true (or even whether the net benefit would be positive or negative).  And you need Univalence to swap between the two styles.  An intermediate step is to keep definitions as they are, but prove equivalences (e.g. that `BAut X` is equivalent to the fibre of `tr` over `tr X`), and use those equivalences as needed (as we do to convert records to sigma-types in order to apply lemmas about sigma types).

One positive note is that a component of a type is 0-connected with either definition, without using univalence.  (This makes it seem odd that we need univalence to prove that they are equivalent.)  More generally, for any modality `O`, the `O`-connected cover is in fact O-connected.  (Again, without univalence.)

jdchristensen:
BTW, the natural map (without univalence) goes in this direction: `merely (x = y) -> (tr x = tr y)`.  So if you are trying to map out of a component (or prove something about a component), then the `merely` definition will be better.  And if you are mapping in, the `tr` definition will be better.

Alizter:
I don't really want to continue working on this. I would be happy for anybody else to take over.

PR #1702, opened 2023-01-26
BAut

jdchristensen:
While browsing, I noticed that the `baut_reduce` tactic could be used in a few more places if the proof of its helper lemma avoids univalence.

### DIFF
--- a/theories/Truncations/Connectedness.v
+++ b/theories/Truncations/Connectedness.v
@@ -26,10 +26,6 @@ A handy benchmark: under our indexing, the map [S1 -> 1] is 0-connected but not 
 
 One reason for our choice is that this way, the n-truncated and n-connected maps are the modal and modally-connected maps for the n-truncation modality.  Many of the basic lemmas about connected maps are in fact true for any modality, and can be found in [Modality.v].  Thus, here we consider mainly properties that involve the interaction of connectedness at different truncation levels. *)
 
-
-Section Extensions.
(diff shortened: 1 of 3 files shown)


jdchristensen:
Note that it looks like I removed Univalence from a couple of results in Connectedness.v, but it is assumed in the Section, so all I am doing is removing a redundant hypothesis in those cases.  Where I really did remove univalence is in the proof of `merely_path_baut`, which I also generalized to `merely_path_component`.

Alizter:
Maybe we should get rid of that section to make it clearer what is and isn't using univalence?

jdchristensen:
It turns out that one thing in that Section (`conn_pointed_type`) didn't need univalence, so I removed the section.  Then, while I was changing things there, I fixed some of the indentation.

theories/Truncations/Connectedness.v
```diff
@@ -98,15 +93,13 @@ Proof.
   exact (p # p0).
 Defined.
 
-End Extensions.
-
 #[export] Hint Immediate conn_point_incl : typeclass_instances.
```
jarlg, line 96: This hint should be moved up slightly to just after `conn_point_elim`. (It was previously placed just after the section which is now gone.)

jarlg:
LGTM, modulo the one comment I made.

Alizter (approved):

PR #1704, opened 2023-01-27
Cover.v: the O-connected cover is O-connected

jdchristensen:


jdchristensen:
@jarlg pointed out that the typeclass instance I proposed is redundant, since Coq can already infer it, so instead I just included a comment to mention that this fact is true for modalities.

jarlg (approved):

PR #1705, opened 2023-01-27
Universe: simplify some things; adapt Sequential

jdchristensen:
`equiv_path` is just a special case of `equiv_transport`, and `isequiv_path` is therefore not needed.  (It's a special case of `isequiv_transport`.)  Also, `equiv_path_V` and a few other things have simpler proofs.  I'm sure more could be done, but these are just things I noticed in passing.


jdchristensen:
I checked that the build time didn't change (within the usual variability).

Alizter (approved):

jarlg (approved):

PR #1707, opened 2023-01-29
Only Require Truncations.Core in Classes library

jdchristensen:
The Classes library doesn't use anything in Truncations/Connectedness.v or Truncations/SeparatedTrunc.v, so don't import or export any of those.  Other files that were getting some of those via Classes now need to explicitly require them.

`-j1` build time improves slightly.  `-j8` build time doesn't change significantly, but looser dependencies will help if we get a smarter build system or when using more than 8 cores.

(Also, a redundant hypothesis removed from `Qpos_enumerable`.)

### DIFF
--- a/theories/Algebra/AbGroups/Abelianization.v
+++ b/theories/Algebra/AbGroups/Abelianization.v
@@ -1,6 +1,7 @@
 Require Import Cubical WildCat.
 Require Import Colimits.Coeq.
 Require Import Algebra.AbGroups.AbelianGroup.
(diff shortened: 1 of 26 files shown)


Alizter (approved):

jdchristensen:
(Note that Truncations/Connectedness exports Modality which export ReflectiveSubuniverse.  That's why many of the changes involve requiring one of those two.)

jarlg (approved):

PR #1708, opened 2023-02-06
Make pFam into a Record to improve polymorphism of pEquiv

jdchristensen:
I finally figured out how to make `pEquiv` have irrelevant universe variables (the best kind):  I had to make `pFam` into a Record.  This has the added benefit of getting rid of the third universe variable.  So now `pEquiv` has just two universe variables, for the two types.

This allows me to revert part of #1706.  `pequiv_compose` can be defined using the WildCat composition, and ends up with fully flexible universe variables:
```
pequiv_compose@{u u0 u1 u2 u3} = 
fun (A : pType@{u}) (B : pType@{u0}) (C : pType@{u1}) (f : pEquiv@{u2 u2} A B)
  (g : pEquiv@{u2 u2} B C) => compose_cate@{u2 u2 u2 u2 u3} g f
     : forall (A : pType@{u}) (B : pType@{u0}) (C : pType@{u1})
         (_ : pEquiv@{u2 u2} A B) (_ : pEquiv@{u2 u2} B C), 
       pEquiv@{u2 u2} A C
(* u u0 u1 u2 u3 |= u2 < u3
                    u <= u2
                    u0 <= u2
                    u1 <= u2 *)
```
No constraints between `u`, `u0` and `u1`. Unfortunately, it does get two extra universe variables, `u2` and `u3` above, that it doesn't have if you define it directly.  But I think it'll be useful to have it match the WildCat definition.

And with this, `test2` in #1703 passes.  And it works in the applications I'm working on.

Almost all of the changes are just to adjust to `pFam` being a Record.

### DIFF
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -35,15 +35,12 @@ Global Instance ispointed_prod `{IsPointed A, IsPointed B} : IsPointed (A * B)
 (** We override the notation for products in pointed_scope *)
 Notation "X * Y" := ([X * Y, ispointed_prod]) : pointed_scope.
 
-(** A pointed type family consists of a type family over a pointed type and a section of that family at the basepoint. By making this a Record, it has one fewer universe variable, and is cumulative. *)
-Record pFam (A : pType) := { pfam_pr1 : A -> Type; dpoint : pfam_pr1 (point A)}.
+(** A pointed type family consists of a type family over a pointed type and a section of that family at the basepoint. By making this a Record, it has one fewer universe variable, and is cumulative. We declare [pfam_pr1] to be a coercion [pFam >-> Funclass]. *)
+Record pFam (A : pType) := { pfam_pr1 :> A -> Type; dpoint : pfam_pr1 (point A)}.
 
 Arguments Build_pFam {A} _ _.
 Arguments pfam_pr1 {A} P : rename.
 Arguments dpoint {A} P : rename.
-
-(** We make [pfam_pr1] a coercion. *)
-Coercion pfam_pr1 : pFam >-> Funclass.
 
 (** The constant pointed family *)
 Definition pfam_const {A : pType} (B : pType) : pFam A
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -35,23 +35,23 @@ Global Instance ispointed_prod `{IsPointed A, IsPointed B} : IsPointed (A * B)
 (** We override the notation for products in pointed_scope *)
 Notation "X * Y" := ([X * Y, ispointed_prod]) : pointed_scope.
 
-(** A pointed type family consists of a type family over a pointed type and a section of that family at the basepoint. *)
-Definition pFam (A : pType) := {P : A -> Type & P (point A)}.
+(** A pointed type family consists of a type family over a pointed type and a section of that family at the basepoint. By making this a Record, it has one fewer universe variable, and is cumulative. *)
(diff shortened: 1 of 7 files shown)


jarlg (approved):
That's great! I like that it also simplified `loops_type`. Happy to merge this.

Alizter:
Would it be useful to also have the same `[F, p]` notation for pFam too? Coq should be able to distinguish the difference from the type of F. Maybe not in this PR.

theories/Pointed/Core.v
```diff
@@ -35,23 +35,23 @@ Global Instance ispointed_prod `{IsPointed A, IsPointed B} : IsPointed (A * B)
...
-(** We make the first projection of a [pFam] a coercion. *)
-Definition pfam_pr1 {A : pType} (P : pFam A) : A -> Type := pr1 P.
-Coercion pfam_pr1 : pFam >-> Funclass.
+Arguments Build_pFam {A} _ _.
+Arguments pfam_pr1 {A} P : rename.
+Arguments dpoint {A} P : rename.
 
-(** We call the section of a pointed type family at the basepoint a [dpoint]. *)
-Definition dpoint {A : pType} (P : pFam A) : P (point A) := pr2 P.
+(** We make [pfam_pr1] a coercion. *)
```
Alizter, line 45: We can make it a coercion directly by replacing the `:` in pfam_pr1 with `:>`.

theories/Pointed/Loops.v
```diff
@@ -326,18 +326,16 @@ Proof.
...
 Defined.
 
-(** In the following lemmas we have used universe annotations explicitly as without them, coq cannot guess the universe levels correctly. Defining pointed maps as a special case of pForall has the side effect of raising the universe level since pForall requires a bigger universe for the type family. Hopefully in the future, coq's "universe guessing" will be smarter and we can drop the annotations here. *)
-
-(* We can convert between families of loops in a type and loops in Type at that type. *)
-Definition loops_type@{i j k} `{Univalence} (A : Type@{i})
-  : pEquiv@{j j k} (loops@{j} (Build_pType@{j} Type@{i} A))
-          [A <~> A, equiv_idmap].
+(* We can convert between loops in a type and loops in [Type] at that type. *)
+Definition loops_type `{Univalence} (A : Type@{i})
```
Alizter, line 330: I am very happy to see this.

Alizter (approved):

jdchristensen:
BTW, it's actually pretty easy to change sigma types to be cumulative, which would also probably solve the pEquiv issue without changing pFam into a record.  But changing pFam into a record also has the benefit of getting rid of a universe variable, which I think is worth it.  I'll create a separate PR making sigma types (and more) cumulative.  I'll merge this one after I think about the [F, p] notation.

jdchristensen:
> Would it be useful to also have the same `[F, p]` notation for pFam too? Coq should be able to distinguish the difference from the type of F.

I tried 
```
Notation "[ P , x ]" := (Build_pFam P x) : pointed_scope.
```
but then Coq doesn't like situations in which `[X, x]` is used to create a `pType`.  So it seems like a different notation would be needed.   I tried a bunch of options, but none of them worked, e.g. `[ P ; x ]`, `[  P ,, x ]`, `[ P, x ]'`.  `[[ P , x ]]` would work, but places like `intros [a|[]]` would need a space inserted, so I'm not sure if it is worth it.

Alizter:
@jdchristensen In that case, I don't think it is worth it. We only use it in a handful of places anyway. Thanks for trying it out though.

PR #1709, opened 2023-02-06
Make sig and everything else in Overture cumulative

jdchristensen:
The first commit changes `sig` to be a Cumulative Record.  This loosens universe constraints. This means that sometimes Coq isn't forced to equate universe variables, which means that the user needs to supply them if they want to keep the number small.  This happens in two places, in ObjectClassifier.v and PropResizing/Nat.v.   The other effect is that if the user doesn't manage universe variables, there are sometimes more of them.  For example, `equiv_path_group` goes from having 10 universe variables to 16.  Luckily, in the places where they were listed explicitly, it worked to just delete them.  In addition, one minor change was needed for some reason in `Metatheory/UnivalenceVarieties.v`.  In total, only four files besides Overture.v needed adjustments.

The second commit makes everything else in Overture.v cumulative, and no other changes were needed to the library.  This was surprising to me, as there are many Records/Inductives/Classes in there that weren't already marked as being Cumulative!

It's a bit unfortunate that making things cumulative can *increase* the number of universe variables, but I've seen enough cases where cumulativity was very helpful to think that this is worth it.

### DIFF
--- a/theories/ObjectClassifier.v
+++ b/theories/ObjectClassifier.v
@@ -104,8 +104,8 @@ Definition equiv_sigma_pfibration@{u v +} `{Univalence} {Y F : pType@{u}}
 Theorem equiv_sigma_fibration_O@{u v} `{Univalence} {O : Subuniverse} {Y : Type@{u}}
   : (Y -> Type_@{u v} O) <~> { p : { X : Type@{u} & X -> Y } & MapIn O p.2 }.
 Proof.
-  refine (_ oE (equiv_sig_coind _ _)^-1).
-  apply (equiv_functor_sigma' equiv_sigma_fibration); intro P; cbn.
+  refine (_ oE (equiv_sig_coind@{u v u v v v u} _ _)^-1).
+  apply (equiv_functor_sigma'@{v u v v v v} equiv_sigma_fibration@{u v}); intro P; cbn.
   rapply equiv_forall_inO_mapinO_pr1.
 Defined.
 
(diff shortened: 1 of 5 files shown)


jdchristensen:
I have also looked into making everything in the entire library Cumulative, and very little breaks.  But the things that do break seem a bit tricky to fix, so I'll make a separate PR after this is merged which shows what happens.

theories/ObjectClassifier.v
```diff
@@ -104,8 +104,8 @@ Definition equiv_sigma_pfibration@{u v +} `{Univalence} {Y F : pType@{u}}
 Theorem equiv_sigma_fibration_O@{u v} `{Univalence} {O : Subuniverse} {Y : Type@{u}}
   : (Y -> Type_@{u v} O) <~> { p : { X : Type@{u} & X -> Y } & MapIn O p.2 }.
 Proof.
-  refine (_ oE (equiv_sig_coind _ _)^-1).
-  apply (equiv_functor_sigma' equiv_sigma_fibration); intro P; cbn.
+  refine (_ oE (equiv_sig_coind@{u v u v v v u} _ _)^-1).
```
Alizter, line 107: Why does `equiv_sig_coind` have so many universes, perhaps we should manually specify them?

jdchristensen: This PR doesn't change the number of universe variables here.  I took a look, and they are actually reasonable.  There are three primary universe variables for the three instances of `Type` in `sig_coind_uncurried`, and then some that are needed to play the role of `max(i,j)` for various `i` and `j`.  It might be possible to get rid of one of them, but I'm not 100% sure it wouldn't confuse Coq to do so.  It's really too bad that we can't use algebraic universes in more places instead of having to introduce extra universe variables.

Alizter: Very well. There are plans to use a more powerful universe polymorphism/minimization algorithm in the future that can handle arbitrary algebraic terms, however currently the biggest issue is an engineering bottleneck with the performance.

jdchristensen: @Alizter Should I go ahead and commit this as is?

Alizter (approved):

PR #1711, opened 2023-02-07
Reduce universes in [equiv_path_Tr] and [equiv_path_group]

jarlg:
We reduce the number of universes in `equiv_path_Tr` from 10 to 2. This has various positive downstream effects, such as reducing the number of universes in `conn_point_elim` from 9 to 2.

We also create an alternative version of `equiv_path_group` with only 2 universes, as opposed to 16. This has a positive effect on `equiv_path_abgroup`, etc. I also did a pass through the current proof of `equiv_path_group'` and cleaned up some universes.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -306,10 +306,9 @@ Proof.
       1: exact _.
       intros [[op' [unit' [neg' ax']]] eq].
       apply path_sigma_hprop; cbn.
-      (* We really need to fix https://github.com/HoTT/HoTT/issues/976 *)
       refine (@ap _ _ (fun x : { oun :
-        { oo : SgOp G | { u : MonUnit G | Negate G}}
-        | @IsGroup G oun.1 oun.2.1 oun.2.2}
+        { oo : SgOp G & { u : MonUnit G & Negate G}}
+        & @IsGroup G oun.1 oun.2.1 oun.2.2}
         => (x.1.1 ; x.1.2.1 ; x.1.2.2 ; x.2))
         ((op;unit;neg);ax) ((op';unit';neg');ax') _).
       apply path_sigma_hprop; cbn.
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -290,50 +290,58 @@ Proof.
 Defined.
 
 (** Under univalence, equality of groups is equivalent to isomorphism of groups. *)
-Definition equiv_path_group {U : Univalence} {G H : Group}
+Definition equiv_path_group' {U : Univalence} {G H : Group@{u}}
   : GroupIsomorphism G H <~> G = H.
 Proof.
-  refine (equiv_compose'
-    (B := sig (fun f : G <~> H => IsMonoidPreserving f)) _ _).
-  { revert G H; apply (equiv_path_issig_contr issig_group).
+  refine (equiv_compose'@{u u v}
+    (B := sig@{u u} (fun f : Equiv@{u u} G H => IsMonoidPreserving@{u u} f)) _ _).
+  { revert G H; apply (equiv_path_issig_contr@{v v v v v} issig_group@{u v}).
     + intros [G [? [? [? ?]]]].
-      exists 1%equiv.
-      exact _.
+      exists equiv_idmap@{u}.
+      exact id_monoid_morphism@{u u u}.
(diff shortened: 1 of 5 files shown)


jarlg:
Strange, the build is genuinely failing... Let me have a look and push a fix.

Alizter (approved):
Nice!

jdchristensen:
By tracing up the call chain a bit higher, one can make `equiv_path_OO` have only one universe variable, simplifying a lot.  Should I push this?

About `equiv_path_group'`, the proof is quite ugly now.  Does it work to just let Coq use universe variables freely, and then control them all at the end when making the version with fewer universe variables?

One other comment:  while messing with universe variable, if it's easy to explicitly add the desired constraints (like `foo@{u v | u < v}`), then I think we should do so.  It's both documentation, and insurance that a later version of Coq doesn't add constraints we don't expect.

jarlg:
Sure, feel free to push any improvements. Noted about constraints.

The proof of `equiv_path_group'` was already quite ugly... but I think what you suggest would work too.

theories/Algebra/Groups/Group.v
```diff
@@ -290,51 +290,59 @@ Proof.
...
+      apply path_sigma_hprop@{u u u}; cbn.
       (* We really need to fix https://github.com/HoTT/HoTT/issues/976 *)
       refine (@ap _ _ (fun x : { oun :
         { oo : SgOp G | { u : MonUnit G | Negate G}}
         | @IsGroup G oun.1 oun.2.1 oun.2.2}
         => (x.1.1 ; x.1.2.1 ; x.1.2.2 ; x.2))
-        ((op;unit;neg);ax) ((op';unit';neg');ax') _).
-      apply path_sigma_hprop; cbn.
-      srefine (path_sigma' _ _ _).
+                ((op;unit;neg);ax) ((op';unit';neg');ax') _).
```
jdchristensen, line 314: BTW, does anyone know what the comment is about?  That issue is closed.  Is it just about wanting to use `&` instead of `|` in the sigma type?

jdchristensen:
I pushed two commits, and I think it's good to go now.  The overall diff is now small.  The technique of defining things without worrying about universes and later cleaning them up is a good one.  And if the cleaned-up version is too limiting in some situation, the original is still there to use.
@Alizter , can you give another thumbs up once you've quickly checked this again?

jdchristensen (approved):

theories/Algebra/Groups/Group.v
Alizter: Yes, at the time I was unable to use the correct notations, but that seems to be fixed now.

theories/Truncations/SeparatedTrunc.v
```diff
@@ -47,7 +47,7 @@ Definition path_Tr {n A} (x y : A)
   : Tr n (x = y) -> (tr x = tr y :> Tr n.+1 A)
   := path_OO (Tr n.+1) (Tr n) x y.
 
-Definition equiv_path_Tr `{Univalence} {n A} (x y : A)
```
Alizter, line 50: You probably don't need to change this any longer.

Alizter (approved):

jdchristensen:
BTW, here's the trick I used to figure out the universe variables in linear time.  I first did:
```
Definition equiv_path_group@{u v +} {U : Univalence} {G H : Group@{u}}
  : GroupIsomorphism G H <~> (paths@{v} G H)
  := equiv_path_group'@{_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _}.
```
Note the `+` in the first line.  So Coq accepts this, with tons of universe variables.  Then I try replacing each underscore in turn with `u` or `v`, one at a time.  Then I remove the `+`.

theories/Truncations/SeparatedTrunc.v
jdchristensen: I know, but I like it slightly better with the type of `A` explicit.  OTOH, it's now inconsistent with `path_Tr`.  I'm happy either way.

theories/Algebra/Groups/Group.v
jdchristensen: Ok, I removed the comment and changed | to & in three places.  There may be other things that could be cleaned up here, but I think we can leave that for another time.

jdchristensen:
@jarlg I'll let you merge when you are happy with it.

jarlg:
Thanks!

PR #1712, opened 2023-02-09
Ext1(Z/n, A) is isomorphic to A/n, and the contravariant six-term exact sequence

jarlg:
Much of the content is joint work with @jdchristensen and @JacobEnder.

The main result of this PR is the construction of the contravariant six-term exact sequence of Ext groups related to a short exact sequence `A -> E -> B` and a "coefficient group" `G`. As an application, we show that `Ext Z/n A` is isomorphic to `A/n` under the assumption that multiplication by `n` defines an embedding `Z -> Z`. (This is of course true for any nonzero `n`, but that's not currently in the library.) To prove these things, we had to clean up universes in `AbSES/` and various other places. This had the benefit of speeding up certain things, so that no line in `AbSES/` is among the slowest 20 lines in the library any more! However, I think there may be some slowdowns caused by the newly enabled cumulativity (at least, some of the slower lines seem even worse in my brief testing), and this PR builds as fast as master on my computer. I've made `IsExact` cumulative, as that was convenient for me, and this didn't cause any noticable slowdown in my testing.

Along the way we define another version of the integers, `Z1`, as the free group on one generator and show that it is commutative and has the correct universal property. We also define projective abelian groups, and show that `Z1` is projective. There are other additions as well, but I hope the commits are self-explanatory.


jarlg:
Hm, it builds fine on my machine, but the CI is getting 17 universes in `ext_cyclic_ab'` as opposed to the 16 I'm getting locally. I'll see if I need to update Coq or something.

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -0,0 +1,223 @@
...
+                o* (pequiv_groupisomorphism (equiv_Z1_hom A))^-1*).
+    1: by apply phomotopy_homotopy_hset.
+    apply ext_cyclic_exact.
+Defined.
+
+(** A version with three universe variables. *)
+Definition ext_cyclic_ab@{u v w | u < v, v < w} `{Univalence}
+  (n : nat) `{emb : IsEmbedding (Z1_mul_nat n)} {A : AbGroup@{u}}
+  : ab_cokernel (ab_mul_nat n) $<~> ab_ext (cyclic' n) A
+  := ext_cyclic_ab'@{u v w u u u u u u u u u u u u u} n (A:=A).
```
Alizter, line 223:
```suggestion
  := ltac:(try
    ( exact (ext_cyclic_ab'@{u v w u u u u u u u u u u u u u} n (A:=A))
    || exact (ext_cyclic_ab'@{u v w u u u u u u u u u u u u u u} n (A:=A)))).
```
This is a way to get cross-version compat.

Alizter: Should probably also include a comment alongside.

jarlg: That's a cool (albeit hacky) suggestion! I think I'd rather see if I can find the culprit line where the two Coq versions are behaving differently, and see if I can fix it at the core. But if we do go with this, I'll be sure to include a comment.

Alizter: You should also check Coq 8.17+rc1 to see if it is a recent master change or was in 8.17. If it is in master, then open an issue so it can be fixed / looked at.

Alizter: We could also add 8.17 to the CI at some point. I'll look into it.

theories/Algebra/AbGroups/AbProjective.v
```diff
@@ -0,0 +1,30 @@
+Require Import Basics Types AbelianGroup AbPullback
+  WildCat Limits.Pullback ReflectiveSubuniverse.
+
+(** * Projective abelian groups *)
+
+(** We define projective abelian groups and show that [P] is projective if and only if every epimorphism [A -> P] merely splits. *)
+
+(** An abelian group [P] is projective if for any map [P -> B] and epimorphism [A -> B], there merely exists a lift [P -> A] making the triangle commute. *)
```
Alizter, line 8: Can you include some ascii art:
```
>>>
       A
     ^ |
  l /  |
       | e
  /    |
       V
P ---> B
   f
<<<
```

theories/Algebra/AbSES/Core.v
```diff
@@ -646,3 +646,72 @@ Proof.
...
+  - snrapply equiv_quotient_abgroup_ump.
+    exists g.
+    intros e; rapply Trunc_rec; intros [a p].
+    refine (ap _ p^ @ _).
+    rapply cx_isexact.
+  - apply isequiv_surj_emb.
+    1: rapply cancelR_conn_map.
+    apply isembedding_isinj_hset.
+    srapply Quotient_ind_hprop; intro x.
+    srapply Quotient_ind_hprop; intro y.
```
Alizter, line 734: Is this where funext comes in? You can use the trick of introducing x, then apply Quotient induction on y, and then reverting x and then inducting, to avoid the funext.

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -0,0 +1,223 @@
...
+  As an application, we use the six-term exact sequence to show that [Ext Z/n A] is isomorphic to [A/n], for nonzero natural numbers [n]. (See [ext_cyclic_ab].) *)
+
+(** Exactness of [0 -> ab_hom B G -> ab_hom E G] follows from the rightmost map being an embedding, which is a consequence of [isembedding_precompose_surjection_hset] from Truncations.Connectedness. *)
+
+Definition isexact_abses_sixterm_i `{Funext}
+  {B A G : AbGroup} (E : AbSES B A)
+  : IsExact (Tr (-1))
+      (pconst : pUnit ->* ab_hom B G)
+      (fmap10 (A:=Group^op) ab_hom (projection E) G).
+Abort.
```
Alizter, line 17: Unfinished?

Alizter:
Some minor comments over an initial inspection. I will do an interactive read over the weekend.

theories/Algebra/AbSES/SixTerm.v
jarlg: Which Coq version is the CI using? (It sounds like 8.17+rc1 from your comment, but just to be sure.)

Alizter: It is using the master branch, which is a bit passed 8.17 now.

theories/Algebra/AbSES/SixTerm.v
jarlg: I'll get to it in a future PR (or this one, depending on how fast things go), but I thought it worth to just include the statement as it's a bit tedious to write out... I guess I should add a short comment about this.

theories/Algebra/AbSES/SixTerm.v
jarlg: I think the CI should use the latest stable version of Coq, which is 8.16.1. Is it easy to make that change? I don't think one should have to build Coq on your own in order to be in tune with the CI.

jdchristensen: The CI builds with several versions of Coq.  This helps ensure that the library works with more than one version, so that we don't have to force every developer to use exactly the same version.  The CI does this so that you *don't* need to build Coq on your own, and can help find issues that can be worked around.  A hacky solution is not a big deal, as we drop support for older versions frequently and can then remove the hack later.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -0,0 +1,80 @@
+Require Import Spaces.Nat Spaces.Int WildCat
+  AbelianGroup AbHom Centralizer AbGroups.Z AbProjective
```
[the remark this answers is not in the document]
jarlg: I placed `ab_mul` in `AbGroups.Z` which is used in this file. (`ab_mul` is the multiplication by an integer on a given abelian group.)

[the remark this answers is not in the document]
jarlg: And to answer your second question, I use this definition of the integers since it easily lets me prove the desired universal property and therefore projectivity. I found the existing integers difficult to work with — we have that `Z` is the initial ring, but we don't have the universal property as a group! I didn't see a short way of proving it either.

(One could use that `Z` is the loop space of the circle, from which the universal property easily follows, but it seems a bit weird to _use_ that fact for these results.)

theories/Algebra/AbSES/SixTerm.v
jarlg: Ah, I wasn't aware. Thanks for explaining. I also found where in the CI it says the version now...

theories/Algebra/AbGroups/Cyclic.v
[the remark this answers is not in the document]
jarlg: Then we need to import `Spaces.Int` in `AbelianGroup.v`. Not a biggie, but just to mention it. At some point we probably want the fact that `ab_mul` defines a homomorphism `Z -> (ab_hom A A)`, which does naturally live in `AbGroup.Z`.

theories/Homotopy/ExactSequence.v
```diff
@@ -21,11 +21,10 @@ Definition IsComplex {F X Y} (i : F ->* X) (f : X ->* Y)
 
 (** This induces a map from the domain of [i] to the fiber of [f]. *)
 Definition cxfib {F X Y : pType} {i : F ->* X} {f : X ->* Y}
-           (cx : IsComplex i f)
-  : F ->* pfiber f.
+  (cx : IsComplex i f) : F ->* pfiber f.
```
jdchristensen, line 24: Except in very simple cases, I prefer having the `:` start a new line (indented two spaces).  This is also in the style guide:  https://github.com/HoTT/Coq-HoTT/blob/master/STYLE.md#195-placement-of-arguments-and-types

theories/Homotopy/ExactSequence.v
```diff
@@ -34,28 +33,26 @@ Defined.
 
 (** ...whose composite with the projection [pfib : pfiber i -> X] is [i].  *)
 Definition pfib_cxfib {F X Y : pType} {i : F ->* X} {f : X ->* Y}
-           (cx : IsComplex i f)
-  : pfib f o* cxfib cx ==* i.
+  (cx : IsComplex i f) : pfib f o* cxfib cx ==* i.
```
jdchristensen, line 36: Here too.

theories/Homotopy/ExactSequence.v
```diff
@@ -68,35 +65,29 @@ Definition iscomplex_iterated_loops {F X Y : pType}
...
 
 (** Passage across homotopies preserves complexes. *)
 Definition iscomplex_homotopic_i {F X Y : pType}
-           {i i' : F ->* X} (ii : i' ==* i) (f : X ->* Y) (cx : IsComplex i f)
-  : IsComplex i' f.
-Proof.
-  exact (pmap_postwhisker f ii @* cx).
-Defined.
+  {i i' : F ->* X} (ii : i' ==* i) (f : X ->* Y) (cx : IsComplex i f)
+  : IsComplex i' f := pmap_postwhisker f ii @* cx.
```
jdchristensen, line 73: We usually put `:=` on a new line too, after two spaces.

theories/Homotopy/ExactSequence.v
```diff
@@ -68,35 +65,29 @@ Definition iscomplex_iterated_loops {F X Y : pType}
...
 
 Definition iscomplex_homotopic_f {F X Y : pType}
-           (i : F ->* X) {f f' : X ->* Y} (ff : f' ==* f) (cx : IsComplex i f)
-  : IsComplex i f'.
-Proof.
-  exact (pmap_prewhisker i ff @* cx).
-Defined.
+  (i : F ->* X) {f f' : X ->* Y} (ff : f' ==* f) (cx : IsComplex i f)
+  : IsComplex i f' := pmap_prewhisker i ff @* cx.
```
jdchristensen, line 77: Here too.

theories/Homotopy/ExactSequence.v
```diff
@@ -218,13 +207,11 @@ Defined.
...
-           (g : F' <~>* F) (h : X' <~>* X) (k : Y' <~>* Y) 
-           (p : Square g h i' i)
-           (q : Square h k f' f)
-           `{IsExact n F X Y i f}
-  : IsExact n i' f'.
+  {i : F ->* X} {i' : F' ->* X'}
+  {f : X ->* Y} {f' : X' ->* Y'}
+  (g : F' <~>* F) (h : X' <~>* X) (k : Y' <~>* Y)
+  (p : Square g h i' i) (q : Square h k f' f)
+  `{IsExact n F X Y i f} : IsExact n i' f'.
```
jdchristensen, line 254: Here.

theories/Homotopy/ExactSequence.v
```diff
@@ -398,13 +382,13 @@ Defined.
 
 (** The connecting maps for the long exact sequence of loop spaces, defined as an extension to a fiber sequence. *)
 Definition connect_fiberseq {F X Y} (i : F ->* X) (f : X ->* Y)
-           `{IsExact purely F X Y i f}
-  : FiberSeq (loops Y) F X.
+  `{IsExact purely F X Y i f} : FiberSeq (loops Y) F X.
```
jdchristensen, line 435: Here.

theories/Homotopy/ExactSequence.v
```diff
@@ -308,28 +294,27 @@ Proof.
 Defined.
 
 Definition pequiv_cxfib {F X Y : pType} {i : F ->* X} {f : X ->* Y}
-           `{IsExact purely F X Y i f}
-  : F <~>* pfiber f := Build_pEquiv _ _ (cxfib cx_isexact) _.
+  `{IsExact purely F X Y i f} : F <~>* pfiber f
```
jdchristensen, line 347: Here and 4 lines down.

theories/Homotopy/ExactSequence.v
```diff
@@ -287,8 +274,7 @@ Defined.
 Definition FiberSeq (F X Y : pType) := { f : X ->* Y & F <~>* pfiber f }.
 
 Definition i_fiberseq {F X Y} (fs : FiberSeq F X Y)
-  : F ->* X
-  := pfib fs.1 o* fs.2.
+  : F ->* X := pfib fs.1 o* fs.2.
```
jdchristensen, line 327: Here.

jdchristensen (commented):
Since there's a lot to do, I'm going to review things one commit at a time.  It might take me a few days.

theories/Algebra/AbSES/SixTerm.v
Alizter: Yeah I should work out a way for the CI to immediately display the version. This is currently some hackery from a few years ago, I'm sure the situation has improved now.

jdchristensen:
> However, I think there may be some slowdowns caused by the newly enabled cumulativity (at least, some of the slower lines seem even worse in my brief testing), and this PR builds as fast as master on my computer.

Can you say more?  I checked that making `pFam` and `sig` Cumulative didn't change the overall build time, but I don't think I checked the per file build time.  BTW, the extra flexibility of `sig` being cumulative is causing me to have to specify universe variables in a couple of places.  So far, not a big deal, but a bit unfortunate, since the goal was to make things involving universes easier.

theories/Algebra/AbSES/Core.v
jarlg: This is the only step which really uses funext, yes. But I think the proof would need to change (e.g. not use `isembedding_isinj_hset`) to avoid it.

theories/Homotopy/ExactSequence.v
jarlg: Thanks, and sorry I'd missed those guidelines. I'll make these changes.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -9,34 +9,39 @@ Require Import Spaces.Nat Spaces.Int WildCat
 (** We can define the integers as the free group on one generator, which we denote [Z1] below. Results from Centralizer.v and Groups.FreeGroup let us show that [Z1] is abelian. *)
 
 (** We define [Z] as the free group with a single generator. *)
-Definition Z1 := FreeGroup Unit.
+Definition Z1@{u v | u < v} := FreeGroup@{u v} Unit.
 Definition Z1_gen : Z1 := freegroup_in tt. (* The generator *)
 
-Definition Z1_rec {G : Group} (g : G) : Z1 $-> G
-  := FreeGroup_rec _ _ (unit_name g).
+(* If we use [FreeGroup_rec] here then we can avoid [Funext], but we don't have the computation rule for [FreeGroup_rec] yet. *)
```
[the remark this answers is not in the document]
jdchristensen: I don't see it...

theories/Algebra/Groups/Subgroup.v
```diff
@@ -441,48 +439,46 @@ Arguments sgt_in {G X g}.
 Arguments sgt_unit {G X}.
 Arguments sgt_op {G X g h}.
 
+(** Note that [subgroup_generated_type] will not automatically land in [HProp]. For example, if [X] already "contains" the unit of the group, then there are at least two different inhabitants of this family at the unit (given by [sgt_unit] and [sgt_in group_unit _]). Therefore, in [subgroup_generated] below propositionally truncate. *)
```
jdchristensen, line 442:
```suggestion
(** Note that [subgroup_generated_type] will not automatically land in [HProp]. For example, if [X] already "contains" the unit of the group, then there are at least two different inhabitants of this family at the unit (given by [sgt_unit] and [sgt_in group_unit _]). Therefore, we propositionally truncate in [subgroup_generated] below. *)
```

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -424,3 +444,68 @@ Proof.
...
+    - exact (i s).
+    - exact (s; idpath).
+  Defined.
+
+  (* We record the computation rule that [to_subgroup_generated] satisfies. *)
+  Local Definition to_subgroup_generated_beta (s : S)
+    : to_subgroup_generated (i s) = subgroup_generated_gen_incl (i s) (s; idpath)
+    := isfreegroupon_rec_beta _ _.
+
+  (* It follows [to_subgroup_generated] is a section of the inclusion map from [G] to [F_S]. *)
```
jdchristensen, line 475:
```suggestion
  (* It follows that [to_subgroup_generated] is a section of the inclusion map from [G] to [F_S]. *)
```

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -0,0 +1,80 @@
...
+  snrapply Build_GroupIsomorphism'.
+  - refine (_ oE (equiv_freegroup_rec A Unit)).
+    symmetry. exact (Build_Equiv _ _ (Unit_rec A) _).
+  - intros f g. cbn. reflexivity.
+Defined.
+
+(** Next we show that [equiv_Z1_hom] is natural. *)
+
+Definition nat_to_Z1 : nat -> Z1 := fun n => grp_pow Z1_gen n.
+Definition Z1_mul_nat (n : nat) : GroupHomomorphism Z1 Z1
```
jdchristensen: Add a blank line.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -0,0 +1,80 @@
...
+Lemma equiv_Z1_hom `{Funext} (A : AbGroup)
+  : GroupIsomorphism (ab_hom ab_Z1 A) A.
+Proof.
+  snrapply Build_GroupIsomorphism'.
+  - refine (_ oE (equiv_freegroup_rec A Unit)).
+    symmetry. exact (Build_Equiv _ _ (Unit_rec A) _).
+  - intros f g. cbn. reflexivity.
+Defined.
+
+(** Next we show that [equiv_Z1_hom] is natural. *)
```
jdchristensen, line 40: Remove this comment.  The next couple of items aren't immediately about naturality (despite having "nat" in their names", and there is another comment at the right spot.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -0,0 +1,80 @@
...
+  induction n as [|n H].
+  1: easy.
+  refine (grp_pow_homo _ _ _ @ _); simpl.
+  by rewrite grp_unit_r.
+Defined.
+
+(** Naturality of [equiv_Z1_hom]. *)
+Lemma ab_mul_precomp `{Funext} (A : AbGroup) (n : Int)
+  : equiv_Z1_hom A o (fun f => grp_homo_compose f (ab_mul n))
+    == ab_mul n o equiv_Z1_hom A.
```
jdchristensen, line 59: Maybe it would be better to add `f` as an explicit argument, rather than stating this as a homotopy?  (But then we see that it is just a repeat of `ab_mul_homo`, so maybe it can just be deleted?)

theories/Spaces/Nat/Core.v
```diff
@@ -739,3 +739,11 @@ Proof.
   apply not_leq_Sn_0.
 Defined.
 
+Definition pos_to_nat : Pos -> nat.
```
jdchristensen, line 742: Since `Pos` is more specialized than `nat`, maybe this function should be in `Pos.Core`?  It should be possible to work with the natural numbers without importing `Pos`.

theories/Algebra/AbGroups/Cyclic.v
jarlg: Good catch, it's clearly redundant and not even used... I'll remove it.

theories/Spaces/Nat/Core.v
jarlg: I guess that's better. I'll have to import `Spaces.Nat.Core` in `Pos.Core` but that works fine.

jdchristensen:
A small tweak to FreeGroup.v makes free groups have only one universe variable.  Therefore, the same is true for `Z1`.  We are then able to remove some of the annotations added to Cyclic.v without increasing the number of universe variables.  (There is still room for improvement in nearby places, but that can be another PR.)

theories/Algebra/AbSES/SixTerm.v
jdchristensen: BTW, I agree with @jarlg 's earlier comment that if there is a simple way to change something in an earlier definition to make the number of universes constant (and maybe smaller at the same time), then that would be great.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -236,6 +236,18 @@ Proof.
   assumption.
 Defined.
 
+Definition ab_mul_nat_homo (n : nat)
+  {A B : AbGroup} (f : GroupHomomorphism A B)
+  : f o ab_mul_nat n == ab_mul_nat n o f.
```
jdchristensen, line 241: The proof can be replaced with `:= grp_pow_homo f n.`, so not sure it is worth keeping this.  If we do keep it, we should keep the argument order the same as `grp_pow_homo`.

Alizter:
I've added an 8.17 job https://github.com/HoTT/Coq-HoTT/pull/1714. If you rebase this PR we can see when the universe issues were introduced.

Alizter:
Ah but now the error message has changed. If you run into the universe issue again, you can at least see when the problem occurs now.

jarlg:
That's strange! That does look a bit a bug..? I suspect the message might not be about the real issue, though.

For the record, here is the error:

```
 # The following term contains unresolved implicit arguments:
 #   (forall (H : Univalence) (n : nat) (emb : IsEmbedding (Z1_mul_nat n))
 #    (A : AbGroup), ab_cokernel (ab_mul_nat n) $<~> ab_ext (cyclic' n) A)
 # More precisely: 
 # - ?A: Cannot infer the implicit parameter A of ab_cokernel whose type is
 #   "AbGroup" in
 #   environment:
 #   H : Univalence
 #   n : nat
 #   emb : IsEmbedding (Z1_mul_nat n)
 #   A : AbGroup
```

theories/Algebra/AbGroups/AbelianGroup.v
jarlg: I kept it made those simplifications. Thanks!

theories/Algebra/Groups/Group.v
```diff
@@ -229,12 +229,14 @@ Proof.
...
 
-Definition grp_iso_inverse {G H : Group}
+Definition grp_iso_inverse@{u} {G H : Group@{u}}
   : GroupIsomorphism G H -> GroupIsomorphism H G.
 Proof.
   intros [f e].
   srapply Build_GroupIsomorphism.
-  - srapply (Build_GroupHomomorphism f^-1).
+  - snrapply (Build_GroupHomomorphism f^-1).
+    2: apply invert_monoid_morphism@{u u u}; exact _.
```
jdchristensen, line 242: Needing to do this is a symptom of a problem elsewhere.  In Classes/theory/groups.v, there is `Section compose_mor` that introduces three types `A`, `B`, `C` and therefore three universe variables.  The results in the section depend on all three universe variables even if they only involve one or two of the types.  Is this a bug in Coq?  Is there a trick to work around it?  At the very least, we can remove the Section.  It's especially annoying because when there are completely free universe variables like this, Coq of course doesn't unify them with any existing universe variables, so they propagate and multiply.  This probably comes up in many places in the library that use Sections...

jdchristensen: In this case, it was easy enough to just split the Section into three, so I did so.

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -0,0 +1,227 @@
...
+    1: exact (abses_pushout_ext (abses_from_inclusion (Z1_mul_nat n))
+                o* (pequiv_groupisomorphism (equiv_Z1_hom A))^-1*).
+    1: by apply phomotopy_homotopy_hset.
+    apply ext_cyclic_exact.
+Defined.
+
+(** A version with three universe variables. *)
+Definition ext_cyclic_ab@{u v w | u < v, v < w} `{Univalence}
+  (n : nat) `{emb : IsEmbedding (Z1_mul_nat n)} {A : AbGroup@{u}}
+  : ab_cokernel (ab_mul_nat n) $<~> ab_ext (cyclic' n) A
```
Alizter, line 224: I guess let's explicitly pass the group to ab_cokernel. I think the `ltac:` is messing with Coq's interference. It was probably a bug that it worked before.

Alizter:
I'll investigate fixing the error on more recent Coq.

jdchristensen:
I just push a commit with `(A := A)` added; let's see if it works.

jdchristensen:
My change did fix the 8.17 builds, and I'm out of time, probably for the rest of the day.

Alizter:
It didn't fix the builds, but the issue is really about universe variables. There are a lot of uses of `@{u v +}` above which make it tricky to see if universes can be reduced from that definition. The reason Coq doesn't use the `A : AbGroup` in the context is because it has the wrong universe level. If you stick a `Set Printing Universes` before the error message, you should get a better message in the CI.

jdchristensen:
I made a guess at the solution, but we can force-push it away if it is wrong.

jdchristensen:
It didn't work, so I deleted that commit.  Does anyone have 8.17 installed to experiment with?

theories/Algebra/AbSES/Pushout.v
```diff
@@ -141,6 +141,45 @@ Definition abses_directsum_distributive_pushouts `{Univalence}
...
+  exists ab_pushout_inr; intro x.
+  nrapply ab_pushout_commsq.
+Defined.
+
+(** *** Properties of pushouts of maps *)
+
+(** The pushout of an epimorphism is an epimorphism. *)
+Global Instance ab_pushout_surjection_inr `{Univalence} {A B C : AbGroup}
+  (f : A $-> B) (g : A $-> C) `{S : IsSurjection f}
+  : IsSurjection (ab_pushout_inr (f:=f) (g:=g)).
```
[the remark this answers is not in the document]
jarlg: It should!

jdchristensen:
I'm finished my review and it LGTM.  I also carefully checked the timing, and there are no serious regressions and many improvements.  In fact, despite all of the new material, the entire library builds almost 5% faster with `-j8`.  I think this is mainly due to the universe simplifications, but other things contributed as well.

I think the only issue left is the 8.17 build.  Maybe someone else can get that working?

Alizter:
@jdchristensen I'll work on getting 8.17 building.

jarlg:
Thanks for all the improvements! Once 8.17 is released and available through opam, I'll be happy to debug the extra universes in `SixTerm.v` more carefully. (I think it's related to the `ap tr` in the fourth line of the proof of `_iii`. It introduces an extra universe which normally would get minimized away, but for some reason isn't in this case. That's on 8.16.1 to be clear.)

I also just realised the results in `SixTerm` should be called contr**a** instead of contr. I'll push that, and then we can merge if everyone is happy.

Alizter:
If you do:
```
opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev
```
you can get access to the "development" and release candidate versions of coq. You can then do `opam install coq.8.17.+rc1` and play around with that.

jarlg:
Ah, that's neat. I'll give it a shot then. (The package is `coq.8.17+rc1`, without the last period... :))

jarlg:
I managed to get rid of the `try` and to get 3 universes in `ab_cyclic_ext` without much fuss. There were two similar spots where `ap tr` was applied, which introduces a new "free" universe which seemed to behave differently between Coq 8.16.1 and 8.17. Not sure what to make of it; changing it to `ap tr@{v}` made it behave uniformly across versions.

jdchristensen:
> If you do ... you can get access to the "development" and release candidate versions of coq. You can then do `opam install coq.8.17+rc1` and play around with that.

Oh, thanks.  I searched a lot and didn't find any instructions for doing that.  Presumably you should also make a different opam "switch" so that you still have the released version of Coq available too?  Or does opam let you install multiple versions in one switch?

jdchristensen (approved):

jarlg:
@Alizter , mind if I make a wiki page where I add those lines about installing a dev version of Coq and using the flambda optimizer? And if you're happy then I'd like to go ahead and merge this.

jdchristensen:
Oh, I also hadn't seen the part of INSTALL.md that says how to install a development version of Coq with opam.  Glad to see it is there.  It might also be good to mention that using flambda with Coq is beneficial.  In my tests, the build speed of the library improves by 8-10%.

I think this is ready to merge as soon as Ali says we're good.

(BTW, should we drop the 8.17 build, since we also have the latest and dev builds?)

Alizter:
@jdchristensen I merged 8.17 in the main repoi, so that commit will disappear when we merge. Currently we have:
```
8.16.1 supported
8.16.1 latest
8.17+rc1
dev ~ 8.18+alpha
```
When 8.17 is released (soon), latest will bump to 8.17, supported will stay the same and then we can remove our own 8.17 job. The issue is that it has been a few months since the 8.17 feature freeze so I can't be sure if something is broken in master or in 8.17.

Alizter:
Feel free to make PRs about documentation. One thing I haven't tried, but should be in theory possible, is having opam switches for every version of Coq you care about, then there is a way to set up Dune so that it compiles both versions at the same time without affecting eachother. The analogous thing is done for OCaml, which allows some developers to build multiple versions of their library at the same time.

Alizter:
@jarlg I added a commit defining FreeGroup_rec_beta.

jarlg:
Thanks! I'll see about using it for `Z1`.

Alizter (approved):
Looks good!

PR #1716, opened 2023-02-16
Makefile.coq.local: avoid character classes in etags regexes

jdchristensen:
I've been wondering for a while why my TAGS file has no matches in it, and today I figured out that my version of `etags` (shipped with emacs) doesn't support character classes (like `[:alnum:]` and `[:space:]`).  So I replaced those with naive things, and it works for me now.

### DIFF
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -277,8 +277,8 @@ dot-file-dep-graphs: $(ALL_DOTFILES)
 TAGS_FILES = $(ALL_VFILES)
 TAGS : $(TAGS_FILES)
 	$(ETAGS) --language=none \
-	-r '/^[[:space:]]*\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Let\|Inductive\|CoInductive\|Proposition\)[[:space:]]+\([[:alnum:]_'\'']+\)/\2/' \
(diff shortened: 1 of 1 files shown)


Alizter (approved):

jdchristensen:
@Alizter Until now, this repo has always used "merge commits" when merging PRs, which include an extra commit by the merging person.  I personally think that that clutters the history, but it also adds additional info, so I guess it was chosen for a reason.  Are we switching away from that, or have your last few merges just been done without the extra commit by accident?  Should we change how we do it?

Alizter:
@jdchristensen Previously GitHub simply didn't have the option to change how you merge, and there were work arounds in the past like repos using bots to push directly to master. I prefer having no merge commits since it makes the history cleaner, and it becomes easier to do things like bisecting and reverting.

The green merge button has an drop down menu on the right which lets you choose the merging strategy. I always do "rebase and merge".

I think it would be sensible to do this from now on, and I guess we should update the relevant documentation.

There are also options to squash commits, but I think we are generally quite tidy when we make PRs so I don't think that is worth using.

SkySkimmer:
> I prefer having no merge commits since it makes the history cleaner, and it becomes easier to do things like bisecting and reverting.

I disagree on every point

mikeshulman:
I prefer merges also.

Alizter:
Very well, I will stop rebasing and merging and continue to make merge commits. I don't feel that strongly either way.

PR #1717, opened 2023-02-17
Reduce universes in [isgraph_op] and [is0functor_compose]

jarlg:
By breaking up two sections and supplying precise arguments, the number of universes in `isgraph_op` and `is0functor_compose` are reduced. Previously these would generate free universes when inferred, e.g. by typeclass search. This caused issues for work in progress, and these changes make things manageable. (We also add universe annotations to `isgraph_type`, otherwise it had 3 universes with the constraint that two of them were equal.)

### DIFF
--- a/theories/Algebra/AbSES/Ext.v
+++ b/theories/Algebra/AbSES/Ext.v
@@ -22,8 +22,7 @@ Definition Ext' (B A : AbGroup@{u}) := Tr 0 (AbSES' B A).
 
 Global Instance isbifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
   : IsBifunctor@{v v w u u v v} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
-  := isbifunctor_compose@{v v w w v v u u u u u u u u v v}
-       _ _ (P:=isbifunctor_abses'@{w u v v}).
+  := isbifunctor_compose _ _ (P:=isbifunctor_abses').
 
 (** [Ext B A] is an abelian group for any [A B : AbGroup]. The proof of commutativity is a bit faster if we separate out the proof that [Ext B A] is a group. *)
 Definition grp_ext `{Univalence} (A B : AbGroup@{u}) : Group.
--- a/theories/Algebra/AbSES/SixTerm.v
+++ b/theories/Algebra/AbSES/SixTerm.v
@@ -160,7 +160,7 @@ Local Definition isexact_ext_cyclic_ab_iii@{u v w | u < v, v < w} `{Univalence}
   : IsExact (Tr (-1))
       (fmap10 (A:=Group^op) ab_hom (Z1_mul_nat n) A)
       (abses_pushout_ext (abses_from_inclusion (Z1_mul_nat n)))
-  := isexact_ext_contra_sixterm_iii@{u v u w}
+  := isexact_ext_contra_sixterm_iii
        (abses_from_inclusion (Z1_mul_nat n)).
 
 (** We show exactness of [A -> A -> Ext Z/n A] where the first map is multiplication by [n], but considered in universe [v]. *)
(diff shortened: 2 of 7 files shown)


jdchristensen (approved):
Nice!

jdchristensen:
Does anyone know why Coq sometimes generates universe variables along with a constraint that some of them are equal?  If the user didn't specify those variables, shouldn't the redundant ones be minimized away?

Also, if there are universe variables in a Section, is it a bug that a definition that doesn't depend on them still has those universe variables present?  This is even worse that the previous issue, as those variables multiply, since they are completely unconstrained.

SkySkimmer:
>Does anyone know why Coq sometimes generates universe variables along with a constraint that some of them are equal? If the user didn't specify those variables, shouldn't the redundant ones be minimized away?

This is probably something with rigid universes not getting minimized. It can be considered a bug IMO.

>Also, if there are universe variables in a Section, is it a bug that a definition that doesn't depend on them still has those universe variables present?

This is the intended behaviour, mostly because checking the definition in a subset of the current universes is not easy to implement.

jdchristensen:
> > Does anyone know why Coq sometimes generates universe variables along with a constraint that some of them are equal? If the user didn't specify those variables, shouldn't the redundant ones be minimized away?
> 
> This is probably something with rigid universes not getting minimized. It can be considered a bug IMO.

Is this a known issue, or should I try to make a small test case?  (I see it regularly, but it's often in complicated situations that might be hard to isolate.  I can try to use the minimizer if needed.)

> > Also, if there are universe variables in a Section, is it a bug that a definition that doesn't depend on them still has those universe variables present?
> 
> This is the intended behaviour, mostly because checking the definition in a subset of the current universes is not easy to implement.

It seems like it shouldn't be hard to notice when a definition makes no use at all of an ambient universe variable.

jdchristensen:
BTW, these questions shouldn't affect the merging of this PR, if someone wants to check it over.

Alizter (approved):

Alizter:
Good job!

jdchristensen:
@SkySkimmer Actually, the example in this PR works in Coq without the HoTT library:

```coq
Global Set Universe Polymorphism.

Class IsGraph (A : Type) :=
{
  Hom : forall (a : A), forall (b : A), Type
}.

Global Instance isgraph_type : IsGraph Type
  := Build_IsGraph Type (fun a b => (forall x : a, b)).

Set Printing Universes.
Print isgraph_type.
(*
isgraph_type@{u u0 u1} = 
{| Hom := fun a b : Type@{u1} => forall _ : a, b |}
     : IsGraph@{u u0} Type@{u0}
(* u u0 u1 |= u0 < u
              u0 = u1 *)
*)
```
This is with 8.16.1.

SkySkimmer:
>Is this a known issue, or should I try to make a small test case? (I see it regularly, but it's often in complicated situations that might be hard to isolate. I can try to use the minimizer if needed.)

for instance
~~~coq
Polymorphic Definition foo :=
  let U1 := Type in
  let U2 := Type in
  let u1_le_u2 := forall (T1:U1) (P:U2 -> Set), P T1 in
  let u2_le_u1 := forall (T2:U2) (P:U1 -> Set), P T2 in
  nat.
Set Printing Universes. Set Printing All.
Print foo.
~~~

>It seems like it shouldn't be hard to notice when a definition makes no use at all of an ambient universe variable.

~~~coq
Set Universe Polymorphism.
Section S.
  Universes u v w.
  Constraint u < v.
  Constraint v < w.

  Definition foo@{|} : Type@{u} -> Type@{w} := fun x => x.
End S.
~~~
To avoid having `foo` depend on `v` we need to generate the universe graph where `v` is removed but the transitive constraint between `u` and `w` is still present. This is not trivial to code.

jdchristensen:
> To avoid having `foo` depend on `v` we need to generate the universe graph where `v` is removed but the transitive constraint between `u` and `w` is still present. This is not trivial to code.

Ah, so the point is that requiring `u < w` is not sufficient;  we want `w` to be at least two levels above `u`, and currently that isn't easy to express using only `u` and `w`.  (I don't think that constraints involving `u+1` are allowed.)  Ok, so the lesson is to be careful about having a Section that contains things that don't use all of the universes.

SkySkimmer:
No, the point is that producing the constraint `u < w` is harder than it looks

PR #1724, opened 2023-02-26
Use universe constraints

### LINKED ISSUE #1570: Use explicit universe constraints
Apparently our hacks like
```
Definition Type2@{i j} := Eval hnf in let gt := (Type1@{j} : Type@{i}) in Type@{i}.
```
can now be replaced with
```
Definition Type2@{i j | Set < j, j < i} := Type@{i}.
```

jdchristensen:
This simplifies UniverseLevel.v and FunextVarieties.v by making use of universe constraints.  It also gets rid of `Type1`, etc., since they can be easily defined as needed.  I decided to move UniverseLevel.v to contrib, since I don't think it will be needed in the library but that's easy to reverse if we do need it in the future.  The library builds after each commit.

Closes #1570.


jdchristensen:
Does something have to be done to make proviola happy when a new file is put in contrib?  @JasonGross / @Alizter 
```
proviola-html/HoTT.Contrib.UniverseLevel.html (real: 0.01, user: 0.01, sys: 0.00, mem: 8828 ko)
Exception in thread Thread-2 (most likely raised during interpreter shutdown):
  File "/usr/lib/python2.7/threading.py", line 801, in __bootstrap_inner
  File "/usr/lib/python2.7/threading.py", line 754, in run
  File "/github/workspace/proviola/camera/Popen_noblock.py", line 20, in enqueue_output
  File "/usr/lib/python2.7/Queue.py", line 138, in put
  File "/usr/lib/python2.7/threading.py", line 384, in notify
<type 'exceptions.TypeError'>: 'NoneType' object is not callable
```

Alizter:
@jdchristensen It doesn't get any special treatment. Sometimes the proviola jobs fail since the runner shuts down subproccesses. If you restart the job and it continues then we will need to have a closer look.

jdchristensen:
Restarting just that one job didn't work (which I what I assume you tried), so I restarted all of them.

jdchristensen:
Now I'm seeing lots of doc-proviola errors in unrelated files, so I think it has nothing to do with this PR.

jdchristensen:
CI says that the doc-proviola job was successful, but if you look at the logs it is full of errors like the one I displayed above.  So something seems to be wrong with the error reporting.

Alizter:
@jdchristensen those errors are because the coq syntax is too new (proviola is not really maintained) or because the jobs run out of memory. It can be safely ignored, the jobs that succeeded will produce the correct doc.

PR #1728, opened 2023-03-02
generate_coqproject.sh: get rid of warning when building with dune

jdchristensen:
I made a few cleanups to etc/generate_coqproject.sh.  It now detects when dune is being
used, and doesn't spit out a warning, since it is expected that we are in a subdirectory.


jdchristensen:
@Alizter Does this minor change make sense?

Alizter (approved):
Yes, sorry I forgot to review before.

PR #1739, opened 2023-07-03
Explicitly use : Type0 more

SkySkimmer:
Currently "Unset Universe Minimization ToSet" is somewhat buggy and these inductives still get minimized to Set automatically, but if that's fixed we need this patch to avoid putting them in floating universes.

Alternatively we could enable minimization to set in these files.


mikeshulman (approved):

jdchristensen (approved):
Looks fine to me.  I won't commit yet in case @Alizter has any comments.

PR #1746, opened 2023-09-02
Miscellaneous cleanups

jdchristensen:
This PR contains a large number of unrelated cleanups to the library.  Each commit is independent, and the library builds after each commit.  I recommend reviewing things one commit at a time.


theories/Basics/Overture.v
```diff
@@ -48,15 +48,12 @@ Definition Relation (A : Type) := A -> A -> Type.
...
   reflexivity : forall x : A, R x x.
-Arguments reflexivity {A R} _.
 
 Class Symmetric {A} (R : Relation A) :=
   symmetry : forall x y, R x y -> R y x.
-Arguments symmetry {A R _ _}.
 
 Class Transitive {A} (R : Relation A) :=
   transitivity : forall x y z, R x y -> R y z -> R x z.
-Arguments transitivity {A R _ _ _}.
```
Alizter, line 59: Why are these redundant? Are they changed elsewhere?

Alizter (approved):
LGTM overall. I am unsure about the names in the first commit but it doesn't really matter to me. I trust you are more familiar with the terminology of sections and retractions.

theories/Basics/Overture.v
jdchristensen: Yes, just a few lines further down in the file, the `Arguments` commands are given again, with `/` as well.

jdchristensen:
About `sect` vs `retr`, you can compare the arguments to `Build_IsEquiv` to see that the names were reversed before.

theories/Basics/PathGroupoids.v
```diff
@@ -1087,15 +1087,14 @@ Proof.
...
   h @@ 1 = whiskerR h 1 :> (p @ 1 = q @ 1)
-  :=
-  match h with idpath => 1 end.
+  := idpath.
 
 Definition concat2_1p {A : Type} {x y : A} {p q : x = y} (h : p = q) :
   1 @@ h = whiskerL 1 h :> (1 @ p = 1 @ q)
-  :=
-  match h with idpath => 1 end.
+  := idpath.
```
jdchristensen, line 1097: Should I just delete these?  Not only are they just repeating the definitions of `whiskerR` and `whiskerL`, they are only special cases of the definitions.  In general, the `1` can be the identity path on any path, not just the identity path on the identity path.

Alizter: I was thinking about this. We could delete them, but we could also add a `#[deprecated]` attribute, see https://coq.inria.fr/distrib/current/refman/using/libraries/writing.html?highlight=deprecated#coq:attr.deprecated.

We would leave them there in case they are there for uniformity and people go searching for them. However if we deprecate them then people will get warned not to use them.

Alizter: Related to this, there are a whole plethora of lemmas that are mostly unused scattered throughout the library. I don't know of a good way to clean these up in a way that won't upset users of the library (of which there more than 0).

jdchristensen: I don't mind unused lemmas, as long as they might be useful.  But these particular lemmas are just repeating the definition, and not even with full generality, so they are pretty useless.  I think we should either delete them, or deprecate them and then delete them in a few months.

Alizter: In that case I would suggest removing them completely.

jdchristensen: Done.

theories/WildCat/Core.v
```diff
@@ -402,6 +402,13 @@ Proof.
...
 
+Definition gpd_moveR_M1 {A : Type} `{Is1Gpd A} {x y : A} {p q : x $-> y}
+  (r : Id _ $== p^$ $o q) : p $== q.
+Proof.
+  refine (_ $@ (cat_assoc _ _ _)^$ $@ ((gpd_isretr p) $@R q) $@ (cat_idl q)).
+  exact ((cat_idr p)^$ $@ (p $@L r)).
+Defined.
+
 Definition gpd_moveR_1M {A : Type} `{Is1Gpd A} {x y : A} {p q : x $-> y}
   (r : p $o q^$ $== Id _) : p $== q.
```
jarlg, line 413: Your `gpd_moveR_M1` looks good, but `gpd_moveR_1M` is incorrect. Here's a correct implementation:

```coq
Definition gpd_moveR_1M {A : Type} `{Is1Gpd A} {x y : A} {p q : x $-> y}
  (r : Id _ $== q $o p^$) : p $== q.
Proof.
  refine ((cat_idl p)^$ $@ _ $@ cat_idr q).
  refine (_ $@ cat_assoc _ _ _ $@ (q $@L (gpd_issect p)^$)^$).
  exact (r $@R p).
Defined.
```
You can drop this in; `gpd_moveR_1M` isn't in use. (GitHub isn't letting me suggest these changes...)

jarlg (commented):
LGTM, I only noticed an existing error in `WildCat.Core` and suggested a correction you can drop in.

About the naming in `Spaces.Nat` (using a `nat_` prefix or not), I agree that leaving a comment is good for future discussion.

theories/WildCat/Core.v
jdchristensen: Nice catch!

PR #1747, opened 2023-09-03
Fix instance cycle in Homotopy.HSpace.Core

jarlg:
I'd previously (inadvertently) created an instance cycle in `Homotopy.HSpace.Core` by making both `ishomogeneous_hspace` and `ishspace_homogeneous` instances. By making one of these a definition we avoid the cycle. I opted to make the passage from a left-invertible H-space to a homogeneous structure an instance because it forgets data. We could make the other direction a `Hint Immediate` (or something similar), but there's currently nothing that would benefit from it.


Alizter:
It would be nice if we had a tool in Coq that could analyse hint databases and determine if there are cycles. If only I had lots of free time.

Alizter (approved):

PR #1751, opened 2023-09-03
bump to coq lang 0.8

Alizter:
This bumps our version of the dune `coq lang` to 0.8. Requiring Dune 3.8 or more. This is the more feature complete version of Coq dune support allowing us to compose with installed Coq libraries. Not that we need to.

Marking as draft for now as it will require other devs to update their Dune version.


Alizter:
@jdchristensen @jarlg are you currently using Dune to build? Would this PR be disruptive if merged now?

jarlg:
If all I need is Dune 3.8 or more, then that's fine by me.

jdchristensen:
I'm currently using Dune 3.7.0, but I assume that it's easy to upgrade to 3.8, right?  If so I'll try that today.

Alizter:
Currently 3.10 is out.

PR #1752, opened 2023-09-03
Be more explicit about universes and remove Check

Alizter:
The `Check` statements here are not needed if we annotate explicitly. They are also noisy.


jdchristensen:
LGTM.  You might use `m` instead of `r` for the fifth variable in the last change, so that it matches the earlier name.

(I think we should also get rid of other noise produced by the build if possible.)

Alizter:
Actually, I've opted to do this in #1753

PR #1753, opened 2023-09-03
chore: move Tests.v to test/ folder

Alizter:
These won't be built when doing `dune build` and can be very noisy. Instead these will be built when doing `dune test` which we do in the CI. The opam jobs should catch any failing tests. The makefile will build all of them.

We further improve performance by splitting the tests file into multiple separate files.

We also add a README detailing some conventions we can use on tests.

@jdchristensen This also makes the main build silent.


test/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+
+## Running tests
+
+To run the tests, simply run
+```
+dune test
+```
+from the root of the repository (or here).
+
```
jdchristensen, line 31:
````suggestion
# Tests

Sometimes there are properties of the library that we would like to test
without polluting it with examples and noisy output. Such tests are collected
here.

We also collect regression tests from various issues and PRs in a
subdirectory called `bugs/`.

## Adding a test

To add a test, create a new .v file in a subdirectory of this directory.

When testing properties of the library, use the same directory
structure. For instance, tests about suspensions should be in
`test/Homotopy/Suspension.v`.

Place regression tests in the subdirectory `bugs', using file name `github<issue number>.v`
or `github<PR number>.v`. For example, if you are adding a test for issue #123, the
file name should be `test/bugs/github123.v`.

## Running tests

To run the tests, simply run
```
dune test
```
````

**Review by jdchristensen — APPROVED** [review]

File: test/README.md (hunk 1 above)
Comment (jdchristensen) [reply test/README.md:31]: Not sure why github doesn't display the suggestion better, but I just rephrased a bunch of things.

**Comment by jdchristensen** [comment]
This is great.  Is there a way to run the tests with `make` as well?  If so, the instructions should be added to the README.md.

Separately, I think we should also link to this README.md from STYLE.md.

File: test/README.md (hunk 1 above)
Comment (Alizter) [reply test/README.md]: Can you push it as a commit to this branch? The suggestion isn't commitable since it is stale now.

**Comment by Alizter** [comment]
@jdchristensen I've updated our _CoqProject script to also include the tests, I don't see an easy way to split these up currently. This means that they will always be run with `make`. Not worse than before however.

I've also updated the STYLE.md file to point to `test/README.md`.

File: test/Modalities/Localization.v [hunk 2]
```diff
@@ -0,0 +1,12 @@
...
+(** [ExtendableAlong_Over] takes 5 universe parameters:
+    - size of A
+    - size of B
+    - size of C
+    - size of D
+    - size of result (>= A,B,C,D) *)
+Check ExtendableAlong_Over@{a b c d m}.
+
+(** Universe parameters are the same as for [ExtendableAlong_Over]. *)
+Check ooExtendableAlong_Over@{a b c d m}.
```
jdchristensen, line 12: I would say to drop this file.  The first one is true because of the annotations, and the second one is clear from the definition.  We have lots and lots of definitions that use universes and I don't think we need to check them like this.

As a separate issue, the constraint that m be at least as large as the others could be added to the definition:
```coq
Fixpoint ExtendableAlong_Over@{a b c d m | a <= m, b <= m, c <= m, d <= m}
```

jdchristensen (approved):

Alizter:
I've annotated (oo)ExtendableAlong_Over	and made sure that `dune build` avoids `test/`.

jdchristensen:
LGTM

jdchristensen:
It seems like `dune build' doesn't work from a subfolder anymore.  I'm pretty sure it used to work, right?

Alizter:
Which folder are you trying it from? It appears to work for me.

jdchristensen:
If I try it within the WildCat folder, it doesn't build anything, but gives no errors.

jdchristensen:
It works within the theories folder, or one up from that.

Alizter:
That's very strange, are you sure it isn't building anything? One way you can test this is by doing `dune clean` and then `dune build --cache disabled` within your `Wildcat` folder.

Also which version of Dune are you using?

I am using 3.10 and I can build everything like you describe just fine.

jdchristensen:
Oh, I see what is happening.  If I run `dune build` within the `WildCat` folder, it only tries to build things in that folder.  Is that how it is supposed to be?  (I'm using 3.10 too.)

Alizter:
@jdchristensen Yes that's what it is supposed to do. You can build outside the folder with `dune build ../..`.

jdchristensen:
Thanks.  Is dune also supposed to rebuild the emacs TAGS file?  (Sorry to clutter this PR...)

Alizter:
@jdchristensen I did not add such a rule, but I can add one if you would like.

jdchristensen:
@Alizter That would be great!

Alizter:
If you want a continuous build with the new tags rule #1754 you can do something like `dune build @default TAGS -w` which will build the `@default` alias (in dune aliases are like sets of targets) together with the new `TAGS` target, all in `-w` watch mode. In watch mode the build will restart as soon as there are source changed.

jdchristensen:
> That's where I am overriding `@default` to mean everything but tests.

It seems like `@default` does include the tests:
```
$ dune build @default --display=short
      coqdep theories/.HoTT.theory.d    
        coqc theories/Idempotents.{glob,vo}
        coqc test/Idempotents.{glob,vo}
...
```
But just `dune build --display=short` doesn't build them.  I'm asking about this because I'd like to build the TAGS and everything else except the tests, but `dune build @default TAGS` also builds the tests.

Alizter:
@jdchristensen Ah thats my miatake. TAGS is pulling in the tests too. I can fix that tomorrow.

Alizter:
@jdchristensen Actually that had nothing to do with it. The reason is as follows: `@default` is a recursive alias meaning that it is really every default alias in each subdirectory. When we run `dune build` we are actually doing `dune build @@default` which is the non-recursive kind, doing the default alias only in the current directory. Perhaps a better way to write the command I mentioned above would be the following:
```
dune build theories/ contrib/ TAGS -w
```
That is clearer to understand and you don't need to know about aliases. You can read this as: Build all the targets in theories/, contrib/ and the target TAGS all in watch mode.

PR #1754, opened 2023-09-04
add emacs TAGS support to dune

Alizter:
@jdchristensen Both `make` and `dune` will share the arguments to `etags` which now live in `etc/emacs/etags-args`. In both systems, the file is inlined to pass the arguments to `etags`.

We add a dune rule that builds (and promotes) the TAGS file. Simply do:

```
dune build TAGS
```

to get up to date tags. This is not included in the `@default` alias, so it won't be built by `dune build`.

I've tested it and both `make` and `dune` are able to produce the `TAGS` file now.

### DIFF
--- a/dune
+++ b/dune
@@ -46,3 +46,15 @@
  (deps
   (alias_rec contrib/all)
   (alias_rec theories/all)))
+
+; Tags for emacs
+
+(rule
+ (target TAGS)
+ (alias emacs)
+ (mode promote)
+ (deps
+  %{bin:etags}
+  (glob_files_rec ./*.v))
+ (action
+  (run etags --language=none %{read:etc/emacs/etags-args} %{deps})))
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -139,6 +139,12 @@ $ make
 You can speed up the build by passing `-jN` where `N` is the number of parallel
 recipes `make` will execute.
 
+You can also use `dune` to build the library.
+
+```shell
+$ dune build
+```
+
 ## 3.4. Installing the library using git
 
 We don't recommend you install the library using the repository and instead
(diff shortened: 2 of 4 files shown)


dune
```diff
@@ -46,3 +46,15 @@
  (deps
   (alias_rec contrib/all)
   (alias_rec theories/all)))
+
+; Tags for emacs
+
+(rule
+ (target TAGS)
+ (alias emacs)
+ (mode promote)
```
Alizter, line 55: This means that TAGS won't just live in `_build/default/` but get automatically promoted to the source directory.

jdchristensen:
LGTM.  Thanks!

jdchristensen:
Sorry, I didn't test this property and it doesn't work.  dune is quoting the contents of the etc/emacs/etags-args file instead of using it verbatim.  The beginning of the command looks like:
```
/usr/bin/etags --language=none '-r '\''/^[ \t]*\(\(Local\|
```
when it should look like
```
/usr/bin/etags --language=none -r '/^[ \t]*\(\(Local\|
```

Also, it would be best to limit this to .v files in the theories and contrib folders.  No need to index files in test or in any other random folders (e.g. I have coq-scripts unpacked here).

Finally, this tries to extract tags from the etags binary itself, since it is listed as one of the dependencies.

Alizter:
@jdchristensen That's a shame, I'll get that fixed ASAP. Thanks for letting me know.

jdchristensen:
I guess one easy fix to the first problem is to make a shell script that contains something like
```
/usr/bin/etags --language=none -r '/^[ \t]*\(\(Local\| .... ' "$@"
```
(i.e. it has all of the hand-coded regular expressions embedded, and has "$@" at the end to take in the list of files to read).  Then both dune and make could run this script.

Alizter:
@jdchristensen I opened #1764

PR #1755, opened 2023-09-05
Makefile.coq.local: find: add missing parentheses and add test folder

jdchristensen:
I believe this find command is meant to handle renamed or removed files or directories, since in normal situations `.vo` and `.glob` files are correctly deleted by `make clean`.  But the implicit `and` between `-name \*.timing`  and `-delete` binds more tightly that the `or`s marked with `-o`, so only timing files are deleted by this command.  The parentheses fix this.

Also, now that we build things in the `test` folder, I added that as well.

### DIFF
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -263,4 +263,4 @@ clean::
 	$(SHOW)"RM *.DOT"
 	$(HIDE)rm -f $(ALL_DOTFILES)
 	rm -f $(EXTRA_CLEANFILES)
-	find contrib theories -name \*.vo -o -name \*.glob -o -name \*.timing -delete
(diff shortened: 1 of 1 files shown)


jdchristensen:
Yes, I have tested it.  I was building some old versions of the library, where the AbSES folder was moved, and .vo files were getting left behind after `make clean`.  After updating the find command like this, the problem goes away.  It's also documented in the find man page:

> Please  note that -a when specified implicitly (for example by two tests appearing without an explicit operator between them) or explicitly has  higher  precedence  than  -o.   This means that find . -name afile -o -name bfile -print will never print afile.

Alizter (approved):
Excellent, LGTM!

jdchristensen:
Merging, since this is a minimal bug fix.

PR #1756, opened 2023-09-12
dune: build _CoqProject by default

jdchristensen:
I noticed _CoqProject was not getting built by dune.  Is this the right way to fix that?


Alizter (approved):
Yes, that is correct.

PR #1758, opened 2023-09-18
HITs: fix inconsistencies from several HITs

jdchristensen:
I noticed that Coq was considering some functions defined using `Join_rec` to be definitionally equal when they weren't.  I traced this back to `GraphQuotient_ind` not using the idiom that makes Coq believe the supporting arguments are actually required.  I looked around and found other cases where this was a problem, and fixed all but one.

The one case I didn't fix was Spaces/No/Core.v.  There are a bunch of suspicious `destruct`s in there that probably need similar guarding, but it was too complicated for me to quickly fix things here.  I'll open an issue about this.  Edit: It's #1759.

### DIFF
--- a/test/bugs/github1758.v
+++ b/test/bugs/github1758.v
@@ -1,6 +1,8 @@
 From HoTT Require Import Basics.Overture HIT.Interval HIT.Flattening Colimits.GraphQuotient
           Spaces.Torus.Torus Cubical.
 
+(* Test that various higher inductive types are defined correctly.  If they are defined in the most naive way, two uses of the induction principle that are definitionally equal on the point constructors will be considered definitionally equal, which is inconsistent.  There is an idiom that must be used in order to force Coq to regard the supplementary data as being required as well.  See, for example, Colimits/GraphQuotient.v for the idiom. *)
+
 Fail Definition test_interval (P : interval -> Type) (a : P zero) (b : P one)
   (p p' : seg # a = b) :
   interval_ind P a b p = interval_ind P a b p'
--- a/test/bugs/github1758.v
+++ b/test/bugs/github1758.v
@@ -1,0 +1,42 @@
+From HoTT Require Import Basics.Overture HIT.Interval HIT.Flattening Colimits.GraphQuotient
+          Spaces.Torus.Torus Cubical.
+
+Fail Definition test_interval (P : interval -> Type) (a : P zero) (b : P one)
+  (p p' : seg # a = b) :
+  interval_ind P a b p = interval_ind P a b p'
+  := 1.
+
+Fail Definition test_wtil {A B f g C D} (Q : Wtil A B f g C D -> Type)
+  (cct' : forall a x, Q (cct a x))
+  (ppt' : forall b y, (ppt b y) # (cct' (f b) y) = cct' (g b) (D b y))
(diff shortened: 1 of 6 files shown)


mikeshulman (approved):

Alizter:
8.18 which was released two weeks ago added some extra kernel support for private inductive types as found in the changelog:
- https://coq.github.io/doc/v8.18/refman/changes.html#changes-in-8-18-0
When we bump our minimal version to 8.18, we should use the `#[private(matching)]` attribute instead. Detailed here:
- https://coq.github.io/doc/v8.18/refman/language/core/variants.html#coq:attr.private(matching)

SkySkimmer:
>When we bump our minimal version to 8.18, we should use the #[private(matching)] attribute instead. Detailed here:

Legacy attributes are not deprecated, we're just not adding new ones.

jdchristensen:
I checked that with Coq 8.18, we still need to use the idiom in order to prevent inconsistencies.  I added a comment to the file containing the tests.

Alizter (approved):
I think this is a good motivation for defining other HITs in terms of some core ones. It will lessen the likelihood of introducing inconsistencies.

I don't think there is much Coq can do about this as private inductive types are a hacky technique to allow us to write HITs. So a garbage-in-garbage-out type situation.

In an ideal world we would have developed a language that we can compile into basic HITs but we would lose the ability to use Coq's inductive types. This is analogous to working with W-types rather than Coq's inductives. If you've ever tried it, its not really convenient to work with. We have similar problems in the universal algebra code.

I think having first-class support in Coq for writing higher inductive types would be the way to go, but that wouldn't fix issues of consistency since they would fall on Coq. And since we will probably be the only user of such HIT definitions it wouldn't be an easy feature to support or check. Not to mention there are all sorts of issues with the computation rules of HITs which AFAIK are not even resolved in theory.

So in conclusion lowering the amount of HITs we have to define lowers the chance of introduced inconsistencies. We have mostly done this anyway for the HITs we use using GraphQuotient. Interested to hear what others think.

jdchristensen:
I agree that it would be good to have fewer HITs in the library.  However, in some cases it might be hard or impossible to get rid of existing HITs.  E.g., I'm not sure how the complicated HITs in Spaces/No/Core.v could be replaced by more basic HITs...

Alizter:
(There is also a whole tangential conversation about using a language/framework that actually has HITs supported like cubical agda, but from what I remember, it was an open question if that supported book HoTT the way we like to claim here.)

Alizter:
> I agree that it would be good to have fewer HITs in the library. However, in some cases it might be hard or impossible to get rid of existing HITs. E.g., I'm not sure how the complicated HITs in Spaces/No/Core.v could be replaced by more basic HITs...

IINM removing all path constructors this issue is still present. I believe it is about defining an inductive-inductive(-inductive) type in Coq. (Similtaneous definition of an inductive type and two inductive families over it). Which AFAIK is not possible (or rather known) to reduce to indexed W-types without axiom K.

Adding inductive-inductive types to Coq is a fine feature request that doesn't have too much support at the moment:
- https://github.com/coq/coq/pull/12464

It would strengthen the logic of Coq so there are all sorts of other considerations one would need.

PR #1760, opened 2023-09-18
chore: fix 8.18 warnings

Alizter:
This will have to wait till we update to 8.18. In the meantime we can ignore these warnings.


jdchristensen:
Looks like we tackled the same thing at the same time.  But the way I did it remains compatible with 8.17.

Alizter:
@jdchristensen Nice!

PR #1762, opened 2023-09-18
dune: make Coq warnings fatal when building locally

Alizter:
When building with the `dev` profile, we make all Coq warnings fatal.

The `dev` profile is selected by default when building locally. Only the opam package CI job builds dune with the `release` profile where the warnings will not be fatal.

We need to think about this some more, since warnings might be introduced outside of our control which would make development annoying. I'll probably end up using another profile like `warn-error` so that `dune build --profile warn-error` would be strict about warnings.

### DIFF
--- a/theories/dune
+++ b/theories/dune
@@ -10,5 +10,10 @@
  (modules :standard)
  (flags -noinit -indices-matter -color on))
 
-; TODO: Tests
-; Prob need to move tests into tests folder
+; On the dev profile, we want to use the -w +all flag to make sure
+; that all warnings are fatal. 
+
+(env
+ (dev
+  (coq
+   (flags -w +all))))


Alizter:
cc @jdchristensen how do you feel about this?

jdchristensen:
@Alizter I'm not sure.  I guess it depends how often this would cause an issue.  It might be annoying if I want to test a newer version of Coq, and have to deal with warnings.  I guess warnings are called warnings because they aren't meant to be fatal errors, so maybe this isn't the best way to go?  In general, with so little development on this repo, I'd like to have the procedures as light-weight as possible.

jdchristensen:
@Alizter BTW, why does the CI on this PR show double the number of checks compared to usual?

Alizter:
@jdchristensen Looks like I accidentally created a branch in this repo rather than my fork.

Alizter:
I'll close this anyway as it doesn't really gain us much. It was useful for me when fixing the warnings but that isn't worth any potential difficulty it introduced.

PR #1763, opened 2023-09-18
Adjustments to support Coq 8.18.0

jdchristensen:
Allow `Pushout_ind` (and therefore various other things) to compute; loosen some universe variables; change `Let ... Qed` to `Local Definition ... Qed` to truly hide the contents (and make some proof terms much smaller); silence two warnings about minimization to Set.

Each commit can be read independently.

Everything builds fine with Coq 8.17 as well.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -93,7 +93,7 @@ Section IsEquivHomotopic.
   Let sect := (fun a:A => (ap f^-1 (h a))^ @ eissect f a).
 
   (* We prove the triangle identity with rewrite tactics.  Since we lose control over the proof term that way, we make the result opaque with "Qed". *)
-  Let adj (a : A) : retr (g a) = ap g (sect a).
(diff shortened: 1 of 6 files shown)


Alizter:
I like your version better since it gets rid of a dubious technique with the opaque lets. Coq added proper support for that use case with the new `#[clearbody]` attribute, but obviously that make 8.17 incompat.

Alizter (approved):

jdchristensen:
Thanks, I'll merge.  In general, I'm going to lean towards faster merging for straightforward changes, since it's getting harder to get two approvals.  I'm always happy to make further changes after merging if anyone notices things that can be improved.

Alizter:
@jdchristensen I think the 2 approvals requirement is a bit hopeful for the time being. I don't mind asking for 1 as long as we continue to review like we have been doing.

mikeshulman:
Yeah, if we really only have 2 active contributors, asking for 2 approvals on each PR is a bit much.  (-:

PR #1764, opened 2023-09-18
fix(emacs): only use vfiles in etags action

Alizter:
@jdchristensen This should work better.

### DIFF
--- a/dune
+++ b/dune
@@ -55,7 +55,8 @@
  (alias emacs)
  (mode promote)
  (deps
-  %{bin:etags}
-  (glob_files_rec ./*.v))
+  (glob_files_rec theories/*.v)
+  (glob_files_rec contrib/*.v))
  (action
-  (run etags --language=none %{read:etc/emacs/etags-args} %{deps})))
+  (system
+   "%{bin:etags} --language=none %{read-lines:etc/emacs/etags-args} %{deps}")))


dune
```diff
@@ -56,6 +56,8 @@
  (mode promote)
  (deps
   %{bin:etags}
-  (glob_files_rec ./*.v))
+  (:vfiles
```
Alizter, line 59: This lets you create a binder for a subset of your deps specification. In this case all the .v files we are globbing.

jdchristensen:
This still has the extra quoting around the regular expression arguments.  This is how I see the command
```
rm -f TAGS _build/default/TAGS; dune build --verbose --cache=disabled -f --always-show-command-line TAGS
```
and in a correctly built TAGS file, `grep functor_join TAGS` should show output.

dune
jdchristensen: Nice, that fixes the second and third issues I mentioned.

Alizter:
@jdchristensen The command that dune outputs with `--verbose` and `--always-show-command-line` isn't always accurate, especially when it comes to quoting. I will play around with etags myself to see what might be going wrong.

Alizter:
The reason I say that is because I wrote my own `etags` executable in order to echo out the arguments being passed:
```
Argument 1: ../install/default/bin/etags
Argument 2: --language=none
Argument 3: -r '/^[ \t]*\(\(Local\|Global\|Cumulative\|NonCumulative\|Monomorphic\|Polymorphic\|Private\)[ \t]+\)*\(Axiom\|Theorem\|Class\|Instance\|Let\|Ltac\|Definition\|Lemma\|Record\|Remark\|Structure\|Fixpoint\|Fact\|Corollary\|Inductive\|CoInductive\|Proposition\)[ \t]+\([a-zA-Z0-9_'\'']+\)/\4/'
Argument 4: -r '/^[ \t]*\([a-zA-Z0-9_'\'']+\)[ \t]*:/\1/'
Argument 5: theories/Algebra/AbGroups.v
Argument 6: theories/Algebra/AbGroups/AbHom.v
Argument 7: theories/Algebra/AbGroups/AbProjective.v
Argument 8: theories/Algebra/AbGroups/AbPullback.v
Argument 9: theories/Algebra/AbGroups/AbPushout.v
```
It looks to me as the regex is being passed fine. The quotation you see is an artefact of how dune serialises its internal commands to shell commands.

jdchristensen:
No, what etags is getting as "Argument 3" is supposed to be two arguments.  The quoting that dune inserts combines them into one argument.  But maybe just inserting line breaks in the etags-args file would fix it?

Alizter:
OK I've managed to fix it. The issue is that we were running the `etags` program directly essentially using the unix `exec`. The shell scripting magic that was being done is actually quite essential before being passed into etags. Therefore switching the action to use the `(system)` shell fixes the issues. I've checked by hand and dune and make agree here.

jdchristensen:
Also, Argument 3 has some single quote characters in it that should be there.

Alizter:
I've also removed the `%{bin:etags}` from the deps and put it directly in the `(system)` action. Dune will automatically add the dependency there so we don't need to be explicit about it. This also allows us to avoid having to bind a name to a subset of the deps and go back to using `%{deps}`.;

jdchristensen:
This seems to be working, but I don't see how.  There are still extra single quote characters being passed to etags, which should mess it up.  It must have some internal code that deletes them.  But I think it would be cleaner to not rely on this behaviour.  Can dune literally put the contents of a file on the command line?  If not, maybe it's cleaner to put the etags command and its args into a script?

jdchristensen:
I can try the script method if dune can't create the right command line.

Alizter:
@jdchristensen The removal of the quote characters is standard shell expansion. This is the same behaviour that the makefile relies on. Feel free to remove any shell behaviours if they are not to your liking. I don't mind keeping them in for something like TAGS however.

jdchristensen:
@Alizter I'm going to PR a different method that I think is easier to understand.  You can let me know what you think.

PR #1765, opened 2023-09-18
create etc/emacs/run-etags.sh to make argument processing clear

jdchristensen:
@Alizter Let me know what you think about this approach.  I think it's very easy to understand what is going on if it needs to be updated in the future.

### DIFF
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -31,8 +31,6 @@ ALECTRYON_EXTRAFLAGS ?=
 # setting TIMING=1. By doing so here we get the same effect.
 TIMING ?= 1
 
-ETAGS ?= etags
-
 CATEGORY_VFILES = $(filter theories/Categories%.v, $(VFILES))
 CORE_VFILES = $(filter-out $(CATEGORY_VFILES),$(filter theories/%.v, $(VFILES)))
 CONTRIB_VFILES = $(filter contrib/%.v, $(VFILES))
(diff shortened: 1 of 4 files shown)


Alizter:
@jdchristensen Are you missing a shell script?

Makefile.coq.local
```diff
@@ -242,7 +240,7 @@ dot-file-dep-graphs: $(ALL_DOTFILES)
 # The TAGS file
 TAGS_FILES = $(ALL_VFILES)
 TAGS : $(TAGS_FILES)
-	$(ETAGS) --language=none $(shell cat etc/emacs/etags-args) $^
+	./etc/emacs/run-etags.sh $^
```
Alizter, line 243: i.e. where did this come from?

jdchristensen: Whoops, fixed.

dune
```diff
@@ -55,7 +55,9 @@
  (alias emacs)
  (mode promote)
  (deps
-  %{bin:etags}
-  (glob_files_rec ./*.v))
+  etc/emacs/run-etags.sh
```
Alizter, line 58: I would keep the `etags` executable here since it is needed for the script.

Alizter (approved):
LGTM, modulo my comment.

Alizter:
Did you check both make and dune agree btw?

dune
jdchristensen: Right!

jdchristensen:
Yes, they agree, modulo sorting (since the .v files get passed in in a different order).

Alizter:
Great!

PR #1768, opened 2023-09-30
Associativity of Join, and lots more

jdchristensen:
This is a big PR, so I'll leave it open for a while and would appreciate it if a few people could help review it.  It's best to review the commits independently, and maybe people can help by reviewing a subset of the commits.  Maybe @JasonGross could take a look at the 11th commit, "Add refine_lhs family of tactics, and demo usage"?

The first two commits don't need review, as they just move Join.v to Join/Core.v and add the meta-file Join.v.

The goal is to prove the associativity of Join, but along the way, many other changes were made.  For the associativity, the idea is to use the Yoneda lemma for 0-groupoid-valued functors.  This leads to a long argument, because of the need to set up the wild category structures, but it has the advantage that each piece is reasonably straightforward.  For most pieces, the top cell can be handled by induction, and for the remaining two or three, the path algebra is not bad.  It's long-winded, but every direct approach I tried got bogged down in path algebra, with no obvious way to do induction.

I also set up the same framework in the two-variable case and use it to prove symmetry of the join, even though that has a direct proof.  I think this provides a good model for future work using this technique.  For example, the ideas here should also work for the 3x3 lemma, the associativity of smash, and similar tasks.

I checked the timing.  The single-threaded build takes ~1.5s longer, and the two new files take about 1.5s, so that makes sense.  No existing file changed significantly in speed.  The `-j8` build speed is unchanged.


theories/Homotopy/CayleyDickson.v
```diff
@@ -300,7 +300,7 @@ Section ImaginaroidHSpace.
   Global Instance cd_op_left_identity
     : LeftIdentity cd_op pt.
   Proof.
-    srapply Join_ind.
+    srapply Join_ind_dp.
```
Alizter, line 303: I wonder if we are actually doing anyone a favour by introducing `DPath` and friends. I used to try to use it as much as possible in the hopes that it would prove useful, but it almost always is too finicky to work with. The only proof that uses the cubical machinery in a serious way is the proof of the equivalence of the Torus and the produce of circles.

I'm interested to know how you find it. I would support reducing our reliance on it as that doesn't appear to gain us much outside of Torus. (Even that can be replaced by a few clever applications of J).

There is potential in the cubical machinary in the realm of wildcats however. If we generalized a lot of stuff to n-cubes in wildcats we could get some useful results about descent and maybe other pasting diagrams. Descent would have nice consequences, for example the Hilton-Milnor splitting.

Just some thoughts.

theories/Types/Paths.v
```diff
@@ -87,6 +87,29 @@ Proof.
   exact ((concat_1p q)^ @ (concat_p1 (1 @ q))^).
 Defined.
 
+(** Variants of the above that do the most common rearranging. We could add similar variants for the others as needed. *)
```
Alizter, line 90: If its more common than the original lemma, we could consider swapping the name.

theories/Homotopy/Join/Core.v
```diff
@@ -111,6 +112,258 @@ Section Join.
 
 End Join.
 
+(** * We now prove many things about [Join_rec], for example, that it is an equivalence of 0-groupoids from the [JoinRecData] that we define next.  The framework we use is a bit elaborate, but it parallels the framework used in TriJoin.v, where careful organization is essential. *)
+
+Record JoinRecData {A B P : Type} := {
```
Alizter, line 117: Is this basically a "join algebra"?

theories/Basics/Tactics.v
```diff
@@ -421,6 +421,66 @@ Tactic Notation "snrapply" uconstr(term)
 Tactic Notation "snrapply'" uconstr(term)
   := do_with_holes' ltac:(fun x => snrefine x) term.
 
+(** Variants, for manipulating one side of an equation without rewriting. *)
+
+(** It would be great if instead of these 32 notations, we could have just four tacticals [lhs], [lhs_V], [rhs] and [rhs_V] that could be used as [lhs rapply lemma], [rhs nrefine (lemma _ _ arg _)], etc.  The former would expand to [refine (_ @ _); [rapply lemma|]], for example. *)
```
Alizter, line 426: Tricky stuff. At this point people use ltac2 or write a tactic. Try asking on Coq zulip, someone might know what to do in this situation.

theories/Homotopy/CayleyDickson.v
jdchristensen: I've always found the `DPath` stuff hard to follow, even though I expect in principle that it should simplify some things.  In this PR, I don't use it, and I created `Join_ind` to refer to the `transport` version and renamed the `DPath` version `Join_ind_dp`.  (This matches the `Susp_ind` and `Susp_ind_dp` naming.)  I wonder if there's a Yoneda proof that the Torus and the product of two circles are equivalent that is less technical...

theories/Types/Paths.v
jdchristensen: The originals are still used quite a bit.  They can be used to `rewrite` deep in an expression, while the new ones are only useful when the `transport` is at the top level.

theories/Homotopy/Join/Core.v
jdchristensen: Maybe?  I don't know what a join algebra is...

theories/Basics/Tactics.v
jdchristensen: I'm ok with the verbose approach, but if someone else can find a way to compress it down to four tacticals, that would be great...

jdchristensen:
I should have mentioned that the real meat is in the 9th commit ("Join/Core: recursion equivalence for join; join_sym via Yoneda") and the last two commits ("Join/TriJoin: 3 variable induction/recursion principles" and "Join/JoinAssoc: associativity of Join using Yoneda lemma").  Those three can be reviewed as a unit.

theories/Basics/Tactics.v
JasonGross: Do you want
```coq
Tactic Notation "lhs" tactic3(tac) := notypeclasses refine (_ @ _); [ tac | ].
```
?

theories/Homotopy/Join/Core.v
Alizter: I made the name up, but I'm thinking this looks alot like an algebra over a monad T or a T-module. You can disregard this as I haven't had time to think it through though.

theories/Basics/Tactics.v
Alizter: @JasonGross Not quite. I think its more:
```
Tactic Notation "lhs" <tactical>(tac) uconstr(term) := tac (term _ @ _).
```
But there is no good way to pass tacticals like `refine` or `exact` around AFAIK.

JasonGross: I don't think the outer `refine` needs to be modified.  Are there any cases where `notypeclasses refine (_ @ _)` isn't adequate? And I don't think `lhs exact` should be `exact (term @ _)`, that'll never succeed.

jdchristensen: @JasonGross That almost does the trick, but there are some puzzling issues.  I tried making the change, and then replacing things like `rapply_lhs foo` with `lhs rapply foo`, and they all worked.  But in some situations the *next* tactic now fails, and I can't figure out why.  With `Set Printing Implicit` I can see that your tactic produces slightly different goals than mine, but that still doesn't explain why the next things fail. They can be made to work with further annotations, but I'd prefer not to have to make the fixes.  I'll push a commit that shows the issues in Spheres.v.

BTW, I had tried something very similar.  With your definition, `lhs rapply lemma` seems to put all of `rapply lemma` into the `tac` argument.  I was trying to do it with two arguments and it didn't work.

jdchristensen: @JasonGross I also tried your definition with the `notypeclasses` removed, since that would more closely match my definition.  But it didn't help.

jdchristensen: @JasonGross So far I only tried converting two files, Join/Core.v and Spheres.v.  There were no issues with the former, and just two with the latter.

jdchristensen: I just saw @Alizter's comment, and @JasonGross is right;  we want the supplied tactic to apply to one of the subgoals of `_ @ _`, not outside.

jdchristensen: @JasonGross I've now converted TriJoin.v, the only other file in which I tried using my tactics, and there were no issues.  So maybe Jason's versions are good enough, and the two places where more annotation are needed are just random fluctuation.

jdchristensen: The difference comes down to `refine (term @ _)` vs `refine (_ @ _); [ refine term | ]` not behaving the same.  Both succeed, but somehow the latter one leaves the goal in a state that makes further unification problems a bit too tricky for Coq, even when they look pretty obvious to me.  `cbn` doesn't help.  Nevertheless, I think the advantage of having four general purpose tacticals makes Jason's approach worth it.  We'll be able to apply more complicated things to one side of an equation, like `lhs (refine (_ @ _); [rapply foo; rapply bar])`, just to make up something, not just the four things that my version hard codes.  So unless I hear another good idea, I'll make things work with Jason's approach and will probably squash the last few commits.

jdchristensen: @JasonGross Ah, it works to just embed `tac` using `ltac:(tac)`.  I'll do this and squash the last few commits.  Thanks for the pointer that led to solving this!

jdchristensen: @JasonGross Ah, it works to just embed `tac` using `ltac:(tac)`.  I'll do this and squash the last few commits.  Thanks for the pointer that led to solving this!

Alizter:
I've remembered why we use Cubical in the Torus proof now, and its solely for managing the path algebra that crops up. To demonstrate how tricky things are, I don't know of anybody who has written our the induction principle for the torus in full using only paths. In contrast, the use of cubical constructions actually allows us to write an induction principle down and also reason about it in a straightforward way.

My last 3x3 lemma attempt (which is attempt no. 6 for me) used something similar to what you have here, although I didn't have the ZeroGpd version of yoneda hence I was getting bogged down by funext and trying to write down the higher paths for "pushout algebras". I'm glad to see this can actually pull through!

I tried to take your ideas and apply them to the Torus being equivalent to the produce of circles. This is actually quite different from the join, since we don't have so much symmetry to exploit. In the join case, and probably many others, we can use the twist and swap maps that you defined to get places. When it comes to the Torus, you still have to prove:
```
(Circle * Circle) -> P $<~> TorusRecData P        (natural in P).
```
I even tried defining `CircleRecData` and noticed that `CircleRecData (CircleRecData P) <~> TorusRecData` but couldn't prove it yet.

Anyway, the take away from this is that this technique has some limitations, but overall is a much cleaner and practical way than anything that has come before! I look forward to the results about pushouts and smash products.

When this is merged, I'll PR my experiments with the torus so they can be looked at. It might be that I've missed something.

jdchristensen:
> I look forward to the results about pushouts and smash products.

I'm not necessarily planning to do these myself...

By the way, one of the tricks hidden in the formalization is the use Mike's notion of an injective and essentially surjective map of 0-groupoids, instead of directly proving that there is a bi-invertible map of 0-groupoids.  I originally did it as a bi-invertible map, and it's fine, but I was able to delete 10 or 20 lines by using Mike's notion.  I never have to prove that `trijoin_rec` is a map of 0-groupoids.  Similarly, it also falls out from the WildCat library that it is natural, because its inverse is natural.

Alizter:
> I'm not necessarily planning to do these myself...

I didn't mean that you would! Just that it is clearer how to go about them now.

jdchristensen:
Does anyone want to look at this more before it is merged?  I'm happy to wait if someone speaks up, but will merge Tuesday if I don't hear anything.

Alizter (approved):
Overall looks very good to me. I like these new tacticals we have now. Have you got any corollaries of join associativity that you want to prove now? Joins of spheres and the Hopf fibrartion (if this is still the state-of-the-art) are some that come to mind.

It would be nice if you could fill in lemma 8.5.9 of the HoTT book with join assoc too.

mikeshulman:
This sounds really neat!  Unfortunately I don't have time to look at it right now, and probably won't for a while, so you shouldn't wait on me.

Alizter:
@jdchristensen I'll make a note of the missing label and try fixing it soon. It might be fixed upstream just that we need to run the hott book scripts again.

Excellent, I'll post my WIP torus PR at some point, and also have a stab at joining spheres.

jdchristensen:
@Alizter I looked in the book source, and there is no \label.

Alizter:
@jdchristensen I opened an issue.

jdchristensen:
@Alizter About spheres, I think the main ingredient is showing that `Join Bool X <~> Susp X`.

PR #1769, opened 2023-10-03
two proofs of join bool being equivalent to suspension

Alizter:
Here are two proofs that `Join Bool A <~> Susp A`.

The first is using a direct adjointified equivalence. However some rewriting is needed to prove the cancellation. Not easy to do with just path algebra either.

The second proof is more abstract, takes longer to setup but is arguably better. Using 0gpd yoneda we can show that the recursion data of both things are naturally equivalent.

This should let us show that joins of spheres are spheres now.

### DIFF
--- a/theories/Homotopy/Suspension.v
+++ b/theories/Homotopy/Suspension.v
@@ -526,3 +526,173 @@ Proof.
   apply (concat (concat_p1 _)).
   apply ap, allpath_p0.
 Defined.
+
+(** Susp rec data *)
+Record SuspRecData {A P} := {
+  srd_n : P;
+  srd_s : P;
+  srd_m : A -> srd_n = srd_s
+}.
+
+Arguments SuspRecData : clear implicits.
+Arguments Build_SuspRecData {A P}%type_scope (srd_n srd_s srd_m)%function_scope.
+
+Definition susp_rec {A P} (f : SuspRecData A P)
+  : Susp A -> P
+  := Susp_rec (srd_n f) (srd_s f) (srd_m f).
+
+Definition susp_rec_beta_merid {A P} (f : SuspRecData A P) (x : A)
+  : ap (susp_rec f) (merid x) = srd_m f x
+  := Susp_rec_beta_merid _.
+
+Definition susprecdata_fun {A P Q} (g : P -> Q) (f : SuspRecData A P)
+  : SuspRecData A Q.
+Proof.
+  srapply Build_SuspRecData.
+  - exact (g (srd_n f)).
+  - exact (g (srd_s f)).
+  - exact (fun x => ap g (srd_m f x)).
+Defined.
+
+Definition susprecdata_susp A : SuspRecData A (Susp A)
+  := Build_SuspRecData North South merid.
+
+Definition susp_rec_inv {A P} (f : Susp A -> P)
+  : SuspRecData A P
+  := susprecdata_fun f (susprecdata_susp A).
+
+Record SuspRecPath {A P : Type} {f g : SuspRecData A P} := {
+  srp_n : srd_n f = srd_n g;
+  srp_s : srd_s f = srd_s g;
+  srp_m : forall x, PathSquare (srd_m f x) (srd_m g x) srp_n srp_s;
+}.
+
+Arguments SuspRecPath {A P} f g.
+
+Definition susp_rec_beta {A P} (f : SuspRecData A P)
+  : SuspRecPath (susp_rec_inv (susp_rec f)) f.
+Proof.
+  srapply Build_SuspRecPath.
+  - reflexivity.
+  - reflexivity.
+  - intros a.
+    apply sq_G1.
+    apply Susp_rec_beta_merid.
+Defined.
+
+Definition isinj_susp_rec_inv {A P} {f g : Susp A -> P}
+  (h : SuspRecPath (susp_rec_inv f) (susp_rec_inv g))
+  : f == g.
+Proof.
+  srapply Susp_ind.
+  - exact (srp_n h).
+  - exact (srp_s h).
+  - intros x.
+    rapply transport_paths_FlFr'.
+    apply sq_path.
+    apply (srp_m h).
+Defined.
+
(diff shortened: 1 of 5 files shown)


jdchristensen:
The path algebra for the direct proof is not bad if you use some of the new things I introduced, like `lhs`, `transport_paths_FFlr'`, `Join_ind_FFlr`, and `equiv_p1_1q`.  See below.  I separated out the definitions of the functions, as that makes the goal easier to read.  (BTW, I don't think it's a problem to use `rewrite` when that makes things easier, as long as we aren't going to need to reason about the proof.)  I haven't look at the rest of this PR yet.

```coq
Definition join_to_susp {A : Type} : Join Bool A -> Susp A.
Proof.
  srapply Join_rec.
    + intros [|].
      * exact North.
      * exact South.
    + intros a.
      exact South.
    + intros [|] a.
      * exact (merid a).
      * reflexivity.
Defined.

Definition susp_to_join {A : Type} : Susp A -> Join Bool A.
Proof.
  srapply Susp_rec.
    + exact (joinl true).
    + exact (joinl false).
    + intros a.
      exact (jglue _ a @ (jglue _ a)^).
Defined.

(** This is a direct proof that the join of bool with a type is equivalent to a suspension. *)
Theorem equiv_join_susp {A : Type} : Join Bool A <~> Susp A.
Proof.
  snrapply (equiv_adjointify join_to_susp susp_to_join).
  - snrapply Susp_ind.
    1,2: reflexivity.
    intros a.
    apply (transport_paths_FFlr' (f:=susp_to_join)).
    apply equiv_p1_1q.
    lhs nrapply (ap _ _); [nrapply Susp_rec_beta_merid | ].
    lhs nrapply (ap_pp _ _ (jglue false a)^).
    lhs refine (_ @@ _).
    1: lhs nrapply ap_V; nrapply (ap inverse).
    1, 2: nrapply Join_rec_beta_jglue.
    apply concat_p1.
  - srapply (Join_ind_FFlr join_to_susp); cbn beta.
    1: intros [|]; reflexivity.
    1: intros a; apply jglue.
    intros b a; cbn beta.
    lhs refine (ap _ _ @@ 1).
    1: nrapply Join_rec_beta_jglue.
    destruct b.
    all: rhs nrapply concat_1p.
    + lhs refine (_ @@ 1); [nrapply Susp_rec_beta_merid | ].
      apply concat_pV_p.
    + apply concat_1p.
Defined.
```

jdchristensen:
In Suspension.v, Mike already proved `issurjinj_Susp_ind_inv`, which gives an equivalence of 0-groupoids between "Susp ind data" and the dependent function type (when combined with `isequiv_0gpd_issurjinj`).  In the case where the type family is constant, this should give a slightly more complicated recursion equivalence, but one that is probably good enough for the 2nd proof in JoinSusp.v.  So I would see if you can remove most or all of the changes to Suspension.v.

In JoinSusp.v, if you want to keep the second proof, I think you can make it a bit shorter by using `isequiv_0gpd_issurjinj`.  There is a choice about which direction to apply it to, so you should see which turns out to be easier.  But either choice should be a bit easier than the current proof.  For example, you'll only need to show that one of the two functions respects paths.

I was originally going to say that I recommend dropping the second proof, but if you want to make the above changes, the second proof might end up being ~80 lines long, so maybe worth having?  I'm also fine with omitting it.

Alizter:
@jdchristensen I noticed there were similar results in Suspension, but I haven't yet spent time reading/understanding them. I was a little weary since it was talking about `ind_data` but that might just be a naming choice. I'll spend a bit more time on this later this week, I also managed to prove joins of spheres are spheres using this. Weirdly enough, the most difficult thing about that proof was rewriting the sum of natural numbers.

theories/Homotopy/Suspension.v
```diff
@@ -526,3 +526,173 @@ Proof.
   apply (concat (concat_p1 _)).
   apply ap, allpath_p0.
 Defined.
+
+(** Susp rec data *)
```
jdchristensen, line 530: If these additions end up being kept, then it would be good to have a high-level comment here explaining what is happening in the next part.

theories/Homotopy/Suspension.v
```diff
@@ -526,3 +526,173 @@ Proof.
...
+Record SuspRecData {A P} := {
+  srd_n : P;
+  srd_s : P;
+  srd_m : A -> srd_n = srd_s
+}.
+
+Arguments SuspRecData : clear implicits.
+Arguments Build_SuspRecData {A P}%type_scope (srd_n srd_s srd_m)%function_scope.
```
jdchristensen, line 538: Does it make sense for `srd_n` and `srd_s` to be in `function_scope`?

theories/Homotopy/Suspension.v
```diff
@@ -526,3 +526,173 @@ Proof.
...
+  - intros f.
+    exists (susp_rec f).
+    apply susp_rec_beta.
+  - intros f g h.
+    apply isinj_susp_rec_inv.
+    exact h.
+Defined.
+
+Definition susp_rec_natequiv A : NatEquiv (susprecdata_0gpd_fun A) (opyon_0gpd (Susp A))
+  := natequiv_inverse _ _ (susp_rec_inv_natequiv A).
```
jdchristensen, line 698: The file should have a newline character at the end.  Otherwise, when someone adds to the file, git will say that line 698 changed.

theories/Homotopy/Suspension.v
Alizter: Not really.

jdchristensen:
@Alizter Is it ok if I PR the direct proof?  (I've also made other changes to my version.)  About the equivalence with spheres, did you prove `equiv_functor_susp`?  That's the missing bit I need for a slick proof with no index rewriting.

Alizter:
@jdchristensen Feel free to open another PR if you wish. I'll push the rest of my additions here, I don't think I will have much time to improve it yet.

Alizter:
I'm not particularly bothered with including the second proof. I think the one you made @jdchristensen is good enough for our purposes. I'm pretty sure they are equivalent proofs too when you take turns them into closed terms.

PR #1771, opened 2023-10-05
Susp is a 1-functor; Join Bool A <~> Susp A; etc.

jdchristensen:
The first commit proves that `Susp` is a 1-functor (and does a couple of other cleanups to Suspension.v).

The second commit proves that `Join Bool A` is equivalent to `Susp A`, using @Alizter 's direct argument, with some reorganization.  It also shows that iterated joins of `Bool` give spheres.

The third commit contains misc cleanups and isn't really related.

### DIFF
--- a/theories/Spaces/Spheres.v
+++ b/theories/Spaces/Spheres.v
@@ -24,12 +24,8 @@ Fixpoint Sphere (n : trunc_index)
        | n'.+1 => Susp (Sphere n')
      end.
 
-(** ** Pointed sphere for non-negative dimensions *)
-Fixpoint psphere (n : nat) : pType
-  := match n with
-      | O => psusp Empty
-      | S n' => psusp (psphere n')
-     end.
+(** ** Pointed sphere for non-negative dimensions.  Note that [n.-1] is the predecessor as a [trunc_index]. *)
+Definition psphere (n : nat) : pType := psusp (Sphere n.-1).
 
 Arguments Sphere : simpl never.
 Arguments psphere : simpl never.
--- a/theories/Spaces/Spheres.v
+++ b/theories/Spaces/Spheres.v
@@ -24,8 +24,8 @@ Fixpoint Sphere (n : trunc_index)
        | n'.+1 => Susp (Sphere n')
      end.
 
-(** ** Pointed sphere for non-negative dimensions.  Note that [n.-1] is the predecessor as a [trunc_index]. *)
-Definition psphere (n : nat) : pType := psusp (Sphere n.-1).
+(** ** Pointed sphere for non-negative dimensions. *)
+Definition psphere (n : nat) : pType := [Sphere n, _].
 
 Arguments Sphere : simpl never.
 Arguments psphere : simpl never.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -232,12 +232,9 @@ Global Instance isequiv_contr_contr {A B : Type}
                   (fun x => path_contr _ _)
                   (fun x => path_contr _ _).
 
-Lemma equiv_contr_contr {A B : Type} `{Contr A} `{Contr B}
-  : (A <~> B).
-Proof.
-  apply equiv_adjointify with (fun _ => center B) (fun _ => center A);
-  intros ?; apply contr.
-Defined.
+Definition equiv_contr_contr {A B : Type} `{Contr A} `{Contr B}
+  : (A <~> B)
+  := Build_Equiv _ _ (fun _ => center B) _.
 
 (** The projection from the sum of a family of contractible types is an equivalence. *)
 Global Instance isequiv_pr1 {A : Type} (P : A -> Type) `{forall x, Contr (P x)}
--- a/theories/Colimits/Pushout.v
+++ b/theories/Colimits/Pushout.v
@@ -91,9 +91,21 @@ Definition pushout_unrec {A B C P} (f : A -> B) (g : A -> C)
   : {psh : (B -> P) * (C -> P) &
            forall a, fst psh (f a) = snd psh (g a)}.
 Proof.
-  exists (h o pushl , h o pushr).
+  exists (h o pushl, h o pushr).
   intros a; cbn.
   exact (ap h (pglue a)).
+Defined.
+
+Definition pushout_rec_unrec {A B C} (f : A -> B) (g : A -> C) P
+  (e : Pushout f g -> P)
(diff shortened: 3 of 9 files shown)


theories/Spaces/Spheres.v
```diff
@@ -24,12 +24,8 @@ Fixpoint Sphere (n : trunc_index)
...
      end.
 
-(** ** Pointed sphere for non-negative dimensions *)
-Fixpoint psphere (n : nat) : pType
-  := match n with
-      | O => psusp Empty
-      | S n' => psusp (psphere n')
-     end.
+(** ** Pointed sphere for non-negative dimensions.  Note that [n.-1] is the predecessor as a [trunc_index]. *)
+Definition psphere (n : nat) : pType := psusp (Sphere n.-1).
```
Alizter, line 28: Won't this mean `psphere` will no longer decompose as pointed suspensions?

jdchristensen: `psusp` has type `Definition psusp (X : Type) : pType`, so it was already throwing away the pointedness of its argument.  So `psusp (psphere n)` is definitionally equal to `psusp (Sphere n)`.  It's true that the new definition won't automatically unfold to `psusp` of a `psphere`, but I'm not sure why that would be needed.

Alizter:
You can merge the definitions of join power easily. I couldn't push so here is a patch:
```diff
diff --git a/theories/Homotopy/Join/Core.v b/theories/Homotopy/Join/Core.v
index 273f987a..d3a5b9f2 100644
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -782,18 +782,11 @@ End JoinTrunc.
 (** Iterated Join powers of a type. *)
 Section JoinPower.

-  (** The join of [n.+1] copies of a type. *)
-  Fixpoint Join_power (A : Type) (n : nat) : Type :=
-    match n with
-    | 0%nat => A
-    | m.+1%nat => Join A (Join_power A m)
-    end.
-
   (** The join of [n] copies of a type. *)
-  Fixpoint Join_power' (A : Type) (n : nat) : Type :=
+  Fixpoint Join_power (A : Type) (n : nat) : Type :=
     match n with
     | 0%nat => Empty
-    | m.+1%nat => Join A (Join_power' A m)
+    | m.+1%nat => Join A (Join_power A m)
     end.

 End JoinPower.
diff --git a/theories/Homotopy/Join/JoinSusp.v b/theories/Homotopy/Join/JoinSusp.v
index af14fd42..c168b102 100644
--- a/theories/Homotopy/Join/JoinSusp.v
+++ b/theories/Homotopy/Join/JoinSusp.v
@@ -52,7 +52,8 @@ Definition equiv_join_susp (A : Type) : Join Bool A <~> Susp A
   := Build_Equiv _ _ (join_to_susp A) _.

 (** It follows that the iterated join of [Bool] gives a sphere. *)
-Definition equiv_join_power_bool_sphere (n : nat): Join_power' Bool n <~> Sphere (n.-1).
+Definition equiv_join_power_bool_sphere (n : nat)
+  : Join_power Bool n <~> Sphere (n.-1).
 Proof.
   induction n.
   - reflexivity.
diff --git a/theories/Metatheory/PropTrunc.v b/theories/Metatheory/PropTrunc.v
index cec2f4a4..2115d10c 100644
--- a/theories/Metatheory/PropTrunc.v
+++ b/theories/Metatheory/PropTrunc.v
@@ -13,7 +13,7 @@ Local Open Scope nat_scope.
 Definition Join_seq (A : Type) : Sequence.
 Proof.
   srapply Build_Sequence.
-  1: exact (Join_power A).
+  1: exact (fun n => Join_power A n.+1).
   intros n.
   exact joinr.
 Defined.
@@ -22,7 +22,12 @@ Defined.
 Definition PropTrunc A : Type := Colimit (Join_seq A).

 (** The constructor is given by the colimit constructor. *)
-Definition ptr_in {A} : A -> PropTrunc A := colim (D:=Join_seq A) 0.
+Definition ptr_in {A} : A -> PropTrunc A.
+Proof.
+  intros a.
+  apply (colim (D:=Join_seq A) 0).
+  exact (joinl a).
+Defined.

 (** The sequential colimit of this sequence is the propositional truncation. *)

@@ -33,10 +38,13 @@ Proof.
   refine (_ oE equiv_colim_seq_rec _ P).
   srapply equiv_iff_hprop.
   { intros h.
-    exact (h 0). }
+    intros a.
+    exact (h 0 (joinl a)). }
   intros f.
   induction n.
-  - exact f.
+  - srapply Join_rec.
+    2,3: contradiction.
+    exact f.
   - cbn.
     srapply Join_rec.
     1,2: assumption.
@@ -52,9 +60,7 @@ Proof.
   rapply (equiv_PropTrunc_rec _ _)^-1.
   intros x.
   srapply contr_colim_seq_into_prop.
-  - intros n.
-    destruct n.
-    1: exact x.
+  - intros n.
     exact (joinl x).
   - intros n.
     rapply jglue.

```

Alizter:
Here are some more:
```coq
(** * Join with Empty *)
Lemma join_empty A : Join A Empty <~> A.
Proof.
  snrapply equiv_adjointify.
  - srapply (Join_rec idmap); contradiction.
  - exact joinl.
  - reflexivity.
  - snrapply Join_ind; [reflexivity| |]; contradiction.
Defined.
```

jdchristensen:
> You can merge the definitions of join power easily.

I think it's convenient to have both.  The old one has `Join_power 0 A = A` definitionally, which is handy sometimes.  And the new one starts at `Empty`, which is convenient to make the base case of some inductions trivial.

I will add the other things you suggested.

jdchristensen:
Thanks, I added those things.  Is this ready for merge?

jdchristensen:
I added one more change:  by including the (-1)-spheres in the statement of `equiv_join_sphere`, it's both more general and has a slightly simpler proof (no rewriting of arithmetic needed).  This is also the natural statement, as it says that the join of the unit sphere in R^n with the unit sphere in R^m is the unit sphere in R^{n+m}.

Alizter:
@jdchristensen Which inductions are you talking about where having `JoinPower A 0 = A` makes it easier? The only one I know about is our propositional truncation construction which AFAIK actually becomes simpler. With the other version of JoinPower.

jdchristensen:
@Alizter One step of that induction becomes slightly easier, but several other places become more awkward, so I thought that that was a good example of why having `JoinPower A 0 = A` can be useful.  E.g. an extra `Join_rec` was required.  In any case, with the use of `nat_iter`, it's now a single line that people can use if it's useful.

PR #1772, opened 2023-10-06
Use nat_iter for iteration

jdchristensen:
I realized that we had many places that were repeating the idiom of iterating a function, such as `iterated_loops`, `Join_power`, `grp_pow`, etc.  So I replaced them with `nat_iter`.  This has the benefit that there are general lemmas proved about `nat_iter` that we can apply in those situations.  I also made various clean-ups while I was touching some files.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -4,6 +4,7 @@ Require Export Classes.interfaces.abstract_algebra.
 Require Export Classes.theory.groups.
 Require Import Pointed.Core.
 Require Import WildCat.
+Require Import Spaces.Nat.Core.
 
 Local Set Polymorphic Inductive Cumulativity.
 
@@ -463,11 +464,7 @@ End GroupMovement.
 
 (** Power operation *)
 
-Fixpoint grp_pow {G : Group} (g : G) (n : nat) : G :=
-  match n with
-  | 0%nat => mon_unit
-  | m.+1%nat => g * grp_pow g m
-  end.
+Definition grp_pow {G : Group} (g : G) (n : nat) : G := nat_iter n (g *.) mon_unit.
 
 (** Any homomorphism respects [grp_pow]. *)
 Lemma grp_pow_homo {G H : Group} (f : GroupHomomorphism G H)
--- a/theories/Algebra/Rings/CRing.v
+++ b/theories/Algebra/Rings/CRing.v
@@ -1,5 +1,5 @@
 Require Import WildCat.
-Require Import Spaces.Nat.
+Require Import Spaces.Nat.Core.
 Require Export Classes.interfaces.abstract_algebra.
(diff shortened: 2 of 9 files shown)


Alizter (approved):
Nice!

Alizter:
Maybe my discomfort with having `Join_power` start from `A` would be alleviated if we renamed `Join_power` to `iterated_join`. The unit being `Empty` makes me think of it less like a multiplicative product anyway.

jdchristensen:
@Alizter Is the last commit roughly what you meant?

Alizter:
@jdchristensen Actually I meant retiring the term "join power" completely since I don't think it is particularly useful. Calling both of them iterated join is better IMO. You could even call the first `iterated_join_nonempty`. But if you feel this is good enough now, then I agree and you can merge.;

jdchristensen:
@Alizter The one that starts at `Empty` is commonly known as the join power and is written in exponential notation: A<sup>*n</sup>.  So I'd like to keep the name `join_power` for that one.  I don't care too much what the other one is named, so maybe it's best to leave it like it is now.

jdchristensen:
One annoying thing with this PR is that `cbn` or `unfold` will now unfold `iterated_loops n.+1 A` to `loops (nat_iter n loops A)`.  `simpl` unfolds it to `loops (iterated_loops n A)`, which is much nicer (and surprisingly clever).  Also, unification produces the uglier form, and that's harder to control.  Is there a directive we could add to get Coq to prefer the nicer form?

One idea is to use notations instead of definitions, e.g.
```coq
Notation iterated_loops n A := (nat_iter n%nat loops A).
Notation join_power A n := (nat_iter n%nat (Join A) (Empty : Type)).
```
Then things work roughly as you'd expect, except that `unfold` no longer works, and you can't ask Coq for the argument types, etc.  Maybe a more serious issue is that terms involving these are all a bit larger, and we've seen cases where having `fix`'s unfolded in the term can slow things down.  So I'd prefer to not go this route.

The situation with the current set-up isn't too hard to deal with, so if there is no simple fix, I'm ok with leaving it as is.

Alizter:
This is a common request in Coq, and I don't know of any silver bullet unfortunately. If you are having terms like those that are unfolding all the time, you could define some custom ltac whilst working.

Something like:
```
  change (nat_iter ?n loops ?A) with (iterated_loops n A). 
```
will refold all your iterated_loops. I used similar things when I want to eta-reduce functions:
```
  change (fun x => ?f x) with f.
```
You can define some ltac or ltac notations to quickly apply these kinds of helpers.

jdchristensen:
The `nat_iter` approach is good because it is concise, and also because it allows us to prove lemmas like `nat_iter_succ_r` once, and then use them for everything defined using `nat_iter`.  So far the issues about unfolding have been very minor, so it's probably best to try the `nat_iter` approach for longer and see how it goes.

PR #1773, opened 2023-10-07
make sure No uses HIT idiom

Alizter:
We add a test that the "dependent cut" differing in the induction principle for No does not offer any opportunity to equate unequal things.

I've also made explicit use of the induction hypotheses when defining the induction principle for No to make sure that they are all being used.

This does not prove the consistency of assuming No however, just merely that an obvious inconsistency doesn't hold.

cc @mikeshulman if you have time to check.


test/bugs/github1759.v
```diff
@@ -0,0 +1,164 @@
+Require Import Basics Types.
```
jdchristensen, line 1:
```suggestion
From HoTT Require Import Basics Types.
```

theories/Spaces/No/Core.v
```diff
@@ -278,7 +278,9 @@ Finally, for conceptual isolation, and so as not to depend on the particular imp
 
     Definition No_ind (x : GenNo) : A x.
     Proof.
+      revert dle dlt ishprop_le ishprop_lt dcut dpath dle_lr dlt_l dlt_r.
       destruct x as [x xno].
+      intros dle dlt ishprop_le ishprop_lt dcut dpath dle_lr dlt_l dlt_r.
```
jdchristensen, line 283:
```suggestion
      destruct x as [x xno].
```
The `foobar` test correctly fails without the change to `No_ind`, so I don't think these lines are needed.  Perhaps because `No_ind` is defined using a `Local` helper function, the HIT idiom has already been applied?

test/bugs/github1759.v
```diff
@@ -0,0 +1,164 @@
...
+
+  Definition foo : forall x, A x
+    := No_ind A dle dlt dcut dpath dle_lr dlt_l dlt_r.
+ 
+  Definition bar : forall x, A x
+    := No_ind A dle dlt dcut' dpath dle_lr' dlt_l' dlt_r'.
+      
+  Fail Definition foobar
+    : forall x, foo x = bar x
+    := 1.
```
jdchristensen, line 162: I'm not able to judge whether this is the right way to test whether `No_ind` is safe, but it looks promising.  Maybe @mikeshulman is able to look at this and verify?

test/bugs/github1759.v
jdchristensen: I committed this so we can get a clean build.

test/bugs/github1759.v
```diff
@@ -0,0 +1,164 @@
+From HoTT Require Import Basics Types.
+Require Import Spaces.No.
```
jdchristensen, line 2:
```suggestion
From HoTT Require Import Spaces.No.
```
Oops, forgot this line.

theories/Spaces/No/Core.v
Alizter: They are not needed, but I added them in to be certain. It's really up to the destruct tactic which context it decides to do  induction over. You can make this more explicit by forcing it to happen over a function, which is essentially what the match statements you wrote are doing.

test/bugs/github1759.v
Alizter: I just realised that this is not correct. Clearly 1 is not going to be well types in a dependent product!

test/bugs/github1759.v
```diff
@@ -0,0 +1,164 @@
...
+
+  Definition foo : forall x, A x
+    := No_ind A dle dlt dcut dpath dle_lr dlt_l dlt_r.
+ 
+  Definition bar : forall x, A x
+    := No_ind A dle dlt dcut' dpath dle_lr' dlt_l' dlt_r'.
+      
+  Fail Definition foobar
+    : forall x, foo x = bar x
+    := 1.
```
Alizter, line 162:
```suggestion
    := fun _ => 1.
```

theories/Spaces/No/Core.v
Alizter: I'll remove them anyhow and just keep the test.

Alizter:
@jdchristensen I've reverted the unneeded change to the induction principle and just left the test so we can be empirically sure that the inconsistency you suspected does not hold. Also `dcut` is the only constructor we need be worried about since the paths and relations on No are all propositions, so its not inconsistent for their elements to be equal AFAIK.

jdchristensen:
I see that the file has one `Axiom` (not including `No_Empty_admitted`) and five things admitted, a couple in definitions that aren't `Local`.  Couldn't any of these be the source of an inconsistency?  I don't really understand the structure of the file well enough to know.  About hprop arguments, do we have to be worried about someone getting something out without providing those arguments at all?  Or getting a definitional equality when it should be propositional?  (None of this is crucial since the library doesn't build on `No`.  All I'm saying is that it's not clear to me where the possible issues are.  It would be nice if there was a more general construction that we could understand in isolation that this was an instance of, like how we base many colimits on graph quotients.)

mikeshulman:
Just a note to say that I've seen this, but don't have time to look at it right now.  When I do, I don't know how easy it will be to page the `No` stuff back into my head, but I'll try.

Alizter:
Thanks @mikeshulman no rush. We can leave this open as long as needs be.

jdchristensen:
How about we merge this PR, and leave the issue open?

mikeshulman:
I got lost in the discussion; what exactly does this PR do?  Does it just check that a certain possible mistake doesn't actually occur?

Alizter:
@mikeshulman Yes.

mikeshulman:
Right, I remember the general issue.  So the current state is that this doesn't seem to be a problem with `No`, and this PR verifies that with a test, but `No` is complicated so we should think about whether something else is broken?  Anyway, yes, fine to merge this PR.

jdchristensen:
@mikeshulman Right, since `No` is complicated, it's not clear if this test is sufficient to know that all is good, or if there might be another subtle issue.  But I think this test at least helps, so should be merged.  I'm going to rebase and adjust, and I'll also add a comment to the top of the file saying this.

jdchristensen:
@Alizter @mikeshulman I rebased this, and then I added a second test, which shows that you can provide two different `ishprop_le` arguments and get definitionally equal results.  I'm guessing this is not inconsistent, but it seems "wrong".  If `No_ind` is using its arguments, then this shouldn't work.  I wonder in @Alizter's original idea of reverting and introing things will change this?

mikeshulman:
Hmm, semantically that seems subtle and tied up with the meaning of things like SProp.  I agree it's not the intended behavior.

Alizter:
@jdchristensen I think reverting and introing will fix that issue. OTOH theoretically I don't believe strict hprops cause any inconsistencies. Though, I haven't really read up on this recently enough.

jdchristensen:
I just added a third test that also shows bad behaviour, with `dlt_r`.  Is that an hprop?

jdchristensen:
(I'll stop pushing to @Alizter 's fork now, in case he wants to try his revert/intro trick.)

Alizter:
@jdchristensen I won't try it today since I am in the middle of another formalization so feel free to keep pushing.

jdchristensen:
To make the subtlety worse, we can only prove that `IsHProp` is an h-prop using `Funext`, and we aren't assuming `Funext` in the test.  So it's not clear whether `No_ind` should even produce propositionally equal outputs in that test.

I'll try @Alizter 's trick ow.

jdchristensen:
That trick didn't help.  Let's leave this open longer in case anyone else wants to experiment.

mikeshulman:
It's not inconsistent to *have* strict hprops, but I don't know of any univalent models in which *all* hprops are strict, although I think the last conversation I had about this concluded that in any infinity-topos we can make a model where every hprop is *equivalent* to a strict one.  I'm not sure whether that would be enough or not for this behavior.

jdchristensen:
It turns out we just needed to do @Alizter 's suggestion in `No_ind_internal` instead of `No_ind`.  With the last commit, if you `Print No_ind_internal`, you see that it is using the usual HIT idiom now.  This trick is already there in `No_ind_internal_step`, but the tests correctly fail with it removed there.  I left it for now, but could remove it.

theories/Spaces/No/Core.v
```diff
@@ -276,7 +277,10 @@ Finally, for conceptual isolation, and so as not to depend on the particular imp
           {struct xno}
     : A (Build_No x xno).
     Proof.
+      (* We use [revert] and [intros] as a way to ensure that the definition depends on all of the variables in the context. *)
+      revert ishprop_le ishprop_lt dpath dle_lr dlt_l dlt_r.
       destruct xno.
```
Alizter, line 282: Alternatively we can do `induction in ishprop_le, ishprop_lt, dpath, dle_lr, dlt_l, dlt_r |- *.`

SkySkimmer: wouldn't induction want t use `is_surreal_rect` which isn't and shouldn't be generated?

jdchristensen: I didn't try that, but now I realized that there's no reason to destruct `xno` in `No_ind_internal`.  The tests still correctly fail.

Alizter: Yes @SkySkimmer, you're probably right. I didn't try this.

jdchristensen: I tried it and @SkySkimmer is right, it doesn't work.

BTW, with revert/intros done twice, the whole file takes twice as long to build.  But with the latest commit, with the second destruct removed completely, we're back to the original speed.

jdchristensen:
As far as I can see, `No_ind_internal` is the only place that a match/destruct is used on a private inductive, and the tests show that that place is protected, so maybe we are good now?  As a bonus, the file got a bit simpler rather than more complicated.

Alizter:
@jdchristensen LGTM. There are still questions of consistency regarding assuming No, but I think they are out of scope of the original issue.

mikeshulman:
Great!

jdchristensen:
It's nice to have this solved.  @Alizter , I'll let you merge.

Issue #1733, opened 2023-03-04
Generate coqdoc using Dune


Alizter:
We should be able to use the `@doc` target to generate the html files. They will appear in `theories/HoTT.html/` and `user-contrib/HoTT.Contrib.html/`. We could also setup Dune rules for running alectryon on these too.

PR #1774, opened 2023-10-07
Fix coqdoc section titles

jdchristensen:
I realized that I was using things like `(** * text here *)` incorrectly,  causing headings to be much too long in the generated html documentation, so I fixed my mistakes.  While I was at it, I found a few other places to fix.

Is there a way to preview the coqdoc html documentation before merging?

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -44,7 +44,7 @@
   - [1.9. Formatting](#19-formatting)
     - [1.9.1. Location of commands](#191-location-of-commands)
     - [1.9.2. Indentation](#192-indentation)
-    - [1.9.3. Line lengths](#193-line-lengths)
+    - [1.9.3. Line lengths and comments](#193-line-lengths-and-comments)
     - [1.9.4. Tactic scripts](#194-tactic-scripts)
     - [1.9.5. Placement of Arguments and types](#195-placement-of-arguments-and-types)
   - [1.10. Implicit Arguments](#110-implicit-arguments)
@@ -906,14 +906,32 @@ excessive churn in the diff.  If you wish, you can submit a separate
 pull request or commit for the re-indentation, labeled as "only
 whitespace changes" so that no one bothers reading the diff carefully.
 
-### 1.9.3. Line lengths
+### 1.9.3. Line lengths and comments
 
 Lines of code should be of limited width; try to restrict yourself to
 not much more than 70 characters.  Remember that when Coq code is
 often edited in split-screen so that the screen width is cut in half,
 and that not everyone's screen is as wide as yours.
 
-Text in comments, on the other hand, should not contain hard newlines.
+[coqdoc](https://coq.inria.fr/refman/using/tools/coqdoc.html) is used to produce
+a browsable [view of the library](https://hott.github.io/Coq-HoTT/coqdoc-html/toc.html).
+coqdoc treats comments specially, so comments should follow the
+conventions described on the coqdoc page.  The most important ones are
+that Coq expressions within comments are surrounded by square brackets,
+and that headings are indicated with comments of the form
+
+```coq
+(** * This is a top-level section *)
+(** ** This is a subsection *)
+(** *** This is a sub-subsection *)
+```
+
+Section titles should be less than 80 characters, on one line, and not
+end in a period.  Other comments are generally written using the style
+`(** This is a comment. *)` or `(* This is a comment. *)`, with the
+latter generally used for inline comments during a proof.
+
+Text in comments should not contain hard newlines.
 Putting hard newlines in text makes it extremely ugly when viewed in a
 window that is narrower than the width to which you filled it.  If
 editing in Emacs, turn off `auto-fill-mode` and turn on
--- a/Makefile.coq.local-early
+++ b/Makefile.coq.local-early
@@ -6,4 +6,4 @@
 # Makefile.coq.local currently contains additional targets that need
 # access to the variables in Makefile.coq
 COQDOCEXTRAFLAGS?=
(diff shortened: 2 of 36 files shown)


Alizter:
@jdchristensen no preview unfortunately, however the built artefacts can be downloaded when the job is finished.

jdchristensen:
@Alizter Can I build the docs locally (easily)?

Alizter:
@jdchristensen Yes, if you want to build with Dune you can do `dune build @doc`. It will probably complain about `odoc` but just ignore that. You will find the html files in `_build/default/theories/HoTT.html/`.

jdchristensen:
I added instructions on comments to STYLE.md.  Despite the instructions in that file, I manually updated the TOC at the top, since I don't have doctoc installed.  Hope I did it right...

jdchristensen:
I silenced all of the warnings coqdoc was generating, which were in fact valid warnings.  Many of the pages look better now.

The only other issue I noticed is that some Category library comments use unicode, and the charset in the html files (both the toc and the individual files) isn't set to utf, so the unicode symbols appear garbled.  Is it easy to change the generated html to use unicode?

Alizter:
@jdchristensen Don't worry about the TOC too much, we can fix it when we come across it.

Alizter:
Regarding unicode, I think it might be possible but I'll have to look into it later. Please open an issue about it.

STYLE.md
```diff
@@ -906,14 +906,32 @@ excessive churn in the diff.  If you wish, you can submit a separate
...
+
+```coq
+(** * This is a top-level section *)
+(** ** This is a subsection *)
+(** *** This is a sub-subsection *)
+```
+
+Section titles should be less than 80 characters, on one line, and not
+end in a period.  Other comments are generally written using the style
+`(** This is a comment. *)` or `(* This is a comment. *)`, with the
```
Alizter, line 931: IINM the former is included in the doc but the latter is not.

jdchristensen: Oh, you are right.  That's controlled by `--parse-comments`.  I would prefer it if all comments were shown, as the inline comments are often quite useful.

jdchristensen:
> Regarding unicode, I think it might be possible but I'll have to look into it later. Please open an issue about it.

The docs say that you need to pass one of `-utf8` or `--utf8`.  In Makefile.coq.local-early, I see that `-utf8` is specified.  I can't see where dune is calling `coqdoc`, though.

jdchristensen:
> @jdchristensen Don't worry about the TOC too much, we can fix it when we come across it.

I tried viewing this PR's STYLE.md on github, and the TOC link worked fine, so all is good.

Alizter:
@jdchristensen To pass flags to Dune (>=3.8) you need to use the `coqdoc_flags` field of the `coq.theory` stanza.

Alizter:
FTR, there is this extension on VSCode that I was using the get the TOC. I checked myself and your version is fine.

STYLE.md
Alizter: On the other hand, sometimes its useful to have comments that don't show up in the doc. For example `TODO` etc. If we have documentation then we ought to be using `(** *)`.

jdchristensen:
> @jdchristensen To pass flags to Dune (>=3.8) you need to use the `coqdoc_flags` field of the `coq.theory` stanza.

I have verified that `make` correctly passes `-utf8`, so the html docs are good as produced by the CI.  I don't understand your instructions above.  Do you want to explain more or just push a separate fix of this yourself?

Alizter:
@jdchristensen something like this:
```
--- a/theories/dune
+++ b/theories/dune
@@ -8,7 +8,8 @@
  (name HoTT)
  (package coq-hott)
  (modules :standard)
- (flags -noinit -indices-matter -color on))
+ (flags -noinit -indices-matter -color on)
+ (coqdoc_flags :standard --utf8))
```
You'll have to do the same for contrib and tests if you want to pass the same args (prob not).

jdchristensen:
Currently `make` passes `-interpolate -utf8 --no-externals $(COQDOCEXTRAFLAGS)` (so it allows extra flags from the environment).

Alizter:
@jdchristensen AFAICT we don't use that anywhere in the CI or anywhere else.

Alizter:
Actually here is the same thing in Dune:
```diff
diff --git a/theories/dune b/theories/dune
index e033640d..b249275d 100644
--- a/theories/dune
+++ b/theories/dune
@@ -8,7 +8,8 @@
  (name HoTT)
  (package coq-hott)
  (modules :standard)
- (flags -noinit -indices-matter -color on))
+ (flags -noinit -indices-matter -color on)
+ (coqdoc_flags -interpolate -utf8 --no-externals %{env:COQDOCEXTRAFLAGS=}))

 ; TODO: Tests
```

Alizter:
@jdchristensen Sorry if I was not clear earlier. I think you can remove `:standard` and just do the change I've mentioned above this comment.

jdchristensen:
> %{env:COQDOCEXTRAFLAGS=}

This causes: `coqdoc: : no such file`.  With the change I just pushed, utf8 works properly.  For some reason, the produced files seem quite different from what make produces, e.g. the order of things in the toc is different, and lots more.  But I don't really feel like investigating further...

Alizter:
@jdchristensen That's completely fine. We don't use the coqdoc feature of Dune. I opened an issue about it #1733 but didn't do anything further. You can think of it as another way to check doc locally. But the CI is the main builder of doc.

STYLE.md
jdchristensen: I searched, and we have lots of important comments written with `(* comment *)`, so I added `--parse-comments`.

jdchristensen:
I think I'm done now!

Alizter (approved):

PR #1778, opened 2023-10-08
wildcat: show Universe is a 2-cat

Alizter:
Here is the 2-category structure on Type. It will be used for when we talk about 3-cells in Type.

### DIFF
--- a/theories/WildCat/Universe.v
+++ b/theories/WildCat/Universe.v
@@ -1,8 +1,8 @@
-Require Import Basics.Overture Basics.Tactics Basics.Equivalences Types.Equiv.
-Require Import WildCat.Core.
-Require Import WildCat.Equiv.
+Require Import Basics.Overture Basics.Tactics Basics.Equivalences Basics.PathGroupoids.
+Require Import Types.Equiv.
+Require Import WildCat.Core WildCat.Equiv WildCat.NatTrans WildCat.TwoOneCat.
 
-(** ** The category of types *)
+(** ** The (1-)category of types *)
 
 Global Instance isgraph_type@{u v} : IsGraph@{v u} Type@{u}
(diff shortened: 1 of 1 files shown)


theories/WildCat/Universe.v
```diff
@@ -1,6 +1,8 @@
-Require Import Basics.Overture Basics.Tactics Basics.Equivalences Types.Equiv.
+Require Import Basics Types.Equiv.
```
jdchristensen, line 1: I know that in general we like to simply require all of `Basics`, but for the `WildCat` library I'd like to keep the dependencies minimal, since many other files require all of `WildCat`, and I'd like the early parts of the build process to parallelize as much as possible.

theories/WildCat/Universe.v
```diff
@@ -112,3 +114,62 @@ Proof.
   intros f x.
   by destruct (f x).
 Defined.
+
```
jdchristensen, line 117:
```suggestion

(** ** The 2-category of types *)

```

jdchristensen (approved):
Nice!

jdchristensen:
In TwoOneCat.v, can you change `(** *** Coherence *)` to `(** Coherence *)`?  The earlier comments aren't headings, so there's no reason for this one to be.  The file is very short.

theories/WildCat/Universe.v
Alizter: Yes, that was sloppy on my part. I'll try to cut it down.

jdchristensen:
@Alizter , I think this is ready to go, modulo fixing the import of Basics and adding the heading.  (I already made the change I suggested in TwoOneCat.v.)

Alizter:
@jdchristensen Sorry that took so long. Your comments should be addressed now.

jdchristensen (approved):

PR #1779, opened 2023-10-09
Spaces.No.Core: speed up the two slowest lines

jdchristensen:
Some simple changes make the two slowest lines instantaneous.  I have no idea why one spot requires univalence to be passed in.  Never seen that before.  I tried various phrasings of that line, but all were slow unless the `ua` argument was given.

### DIFF
--- a/theories/Spaces/No/Core.v
+++ b/theories/Spaces/No/Core.v
@@ -921,7 +921,9 @@ Section NoCodes.
   Proof.
     revert x y.
     refine (No_ind_hprop _ _); intros L R ? xL xR xcut xHL xHR.
-    refine (@No_ind_hprop _ _ _ _).
+    nrefine (No_ind_hprop _ _).
+    (* Coq can find this proof by typeclass search, but helping it out makes this faster. *)
+    { intro x; nrapply istrunc_prod; nrapply istrunc_arrow; exact _. }
     intros L' R' ? yL yR ycut yHL yHR. split.
     - intros x_le_y.
       rewrite le'_cut in x_le_y.
@@ -1008,7 +1010,7 @@ Global Instance insort_maxsort {L R : Type}
 (** Furthermore, every other kind of surreal number *embeds* into the maximal ones.  So the other kinds of surreal numbers are really just subsets of the usual set of surreal numbers; but I don't know of a good way to define them except as their own HIITs. *)
 
 Section RaiseSort.
-  Context `{Univalence} `{S : OptionSort}.
+  Context `{ua: Univalence} `{S : OptionSort}.
 
   Definition No_raise : GenNo S -> No.
   Proof.
@@ -1065,7 +1067,7 @@ Section RaiseSort.
       strip_truncations.
       destruct sh as [[l sh]|[r sh]].
       + apply lt_l with l.
-        apply IHL0, (@No_decode_le _).
+        apply IHL0, (@No_decode_le ua). (* Need to pass in [Univalence] to make this fast. *)
         rewrite p; exact sh.
       + apply lt_r with r.
         apply IHR, No_decode_le.


theories/Spaces/No/Core.v
```diff
@@ -1065,7 +1067,7 @@ Section RaiseSort.
       strip_truncations.
       destruct sh as [[l sh]|[r sh]].
       + apply lt_l with l.
-        apply IHL0, (@No_decode_le _).
+        apply IHL0, (@No_decode_le ua). (* Need to pass in [Univalence] to make this fast. *)
         rewrite p; exact sh.
       + apply lt_r with r.
         apply IHR, No_decode_le.
```
jdchristensen, line 1073: At this spot with the same lemma (and many other spots), Coq finds Univalence quickly.

Alizter: That is absolutely bizarre. Could it be that Coq is searching the typeclasses database for any instances that produce univalence? I wonder if we have some false promises lying around like being able to produce the term of any type which could cause typeclasses to spin out.

JasonGross: There might also be a universe issue?  If `Set Typeclasses Debug` doesn't give enough information (i.e., if the tc log is relatively short but spends a bunch of time producing no output), you can also do `Set Typeclasses Debug Verbosity 2` and get all the things it runs unification/conversion on before deciding to discard things that don't match.

Alizter: Also `Set Typeclasses Depth 3` or another short number can let you understand what Coq is spending its time searching for more concisely.

JasonGross: > i.e., if the tc log is relatively short but spends a bunch of time producing no output

Also btw, if you want an automated way to search for this sort of thing, you can pipe the output of coqc through https://github.com/JasonGross/coq-scripts/blob/master/timing/insert-timings.sh (I think we have this as a submodule here in `etc/coq-scripts/timing/insert-timings.sh`?) and then sort by timing.

theories/Spaces/No/Core.v
```diff
@@ -921,7 +921,9 @@ Section NoCodes.
   Proof.
     revert x y.
     refine (No_ind_hprop _ _); intros L R ? xL xR xcut xHL xHR.
-    refine (@No_ind_hprop _ _ _ _).
+    nrefine (No_ind_hprop _ _).
+    (* Coq can find this proof by typeclass search, but helping it out makes this faster. *)
+    { intro x; nrapply istrunc_prod; nrapply istrunc_arrow; exact _. }
```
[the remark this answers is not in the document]
Alizter: @JasonGross Those kinds of typeclasses speedups are really valuable for us but obviously its very difficult to know all the tricks. Would you be able to open an issue explaining some tricks? The kinds of classes that we use quite often (and often with issue) are truncatedness and the wildcat ones. Our mathclasses code also has a lot of these issues, but that's less of a priority rn.

theories/Spaces/No/Core.v
jdchristensen: I set all of those options, but the log doesn't seem to explain what is going on.  For the record, the line only takes 0.27s, but other similar lines are essentially 0s, so something odd is going on.  There is a line in the middle which takes 0.22s, which is probably the culprit.  The log is here:

```
.002406129: bash etc/generate_coqproject.sh
.009827722: coq_makefile -f _CoqProject -o Makefile.coq
.004980057: make --no-print-directory -f Makefile.coq theories/Spaces/No/Core.vo
.137271849: COQC theories/Spaces/No/Core.v
1.174671528: Debug:
.002857674: Calling typeclass resolution with flags: depth = 3,unique = false,fail = true
.002448413: Debug:
.002302356: Starting resolution with 1 goal(s) under focus and 0 shelved goal(s) in only_classes mode, with depth limit 3
.002810360: Debug: Launching resolution fixpoint on 1 goals: ?X557 : Univalence
.003071873: Debug:
.002171515: Calling fixpoint on : 1 initial goals, 0 stuck goals and 0 non-stuck failures kept with no progress made in this run.
.003264044: Stuck:
.003003146: Failed:
.002262708: Initial: Goal 1 evar: ?X557 status: initial
.002284505: Debug: considering goal 1 of status initial
.221802276: Debug: 1: looking for Univalence without backtracking
.002825297: Debug: 1.1: exact ua on Univalence, 0 subgoal(s)
.002807688: Debug: 1.1: after exact ua finished, 0 goals are shelved and unsolved ( )
.002901274: Debug: Goal 1 has a success, continuing resolution
.002410983: Debug:
.002749356: Calling fixpoint on : 0 initial goals, 0 stuck goals and 0 non-stuck failures kept with progress made in this run.
.002535823: Stuck:
.001930026: Failed:
.002245766: Initial:
.003273406: Debug: Result goals after fixpoint: 0 goals:
.002940105: Debug: after eauto_tac_stuck: 0 goals:
.003224044: Debug: The tactic trace is: simple refine ua;shelve_goals
.002936066: Debug:
.002890889: Finished resolution with a complete solution.
.003152352: Old typeclass evars not concerned by this resolution =
.001812222: 
.002338079: Shelf =
.002727169: Debug: New typeclass evars are:
```

theories/Spaces/No/Core.v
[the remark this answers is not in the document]
jdchristensen: Yes, it would be great if some things were just faster by default, thanks to some clever settings.  I don't understand these options well enough to play with this myself.

theories/Spaces/No/Core.v
JasonGross: If that doesn't reveal what's going on, you can try `Set Debug "tactic-unification,unification". as well, but this will be too verbose to make sense of if we haven't already narrowed down what unification problem is slow.

jdchristensen: I compared the output for the slow line with the output for the similar line three lines further down, and they are identical except for the name of the evar.

theories/Spaces/No/Core.v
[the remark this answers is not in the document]
JasonGross: I would do something like
```
#[export] Hint Constants Opaque: typeclass_instances. (* all unfoldable constants have to be declared as such explicitly*)
#[export] Hint Mode IsTrunc ! ! : typeclass_instances. (* if the level and type both have some known structure, we can proceed *)
#[export] Hint Mode IsTrunc - + : typeclass_instances. (* if the type is fully known we can use tc to infer the level *)
#[export] Hint Mode IsHProp ! : typeclass_instances.
#[export] Hint Mode IsHSet ! : typeclass_instances.
#[export] Hint Mode Contr ! : typeclass_instances.
#[export] Hint Mode IsEquiv - - + : typeclass_instances. (* only infer equivalence if the function is known; - - ! would also be fine if we want to infer IsEquiv id without needing the types to be known *)
#[export] Hint Mode Equiv + + : typeclass_instances. (* only infer Equiv if both types are fully known *)
```
and then see what breaks and why, and adjust from there.

------

If you describe what typeclasses are used in wildcats / mathclasses and how the information flow works, I might be able to guess a good starting point.

theories/Spaces/No/Core.v
jdchristensen: I did have `Verbosity 2` set in that log.  I had:
```coq
        Set Typeclasses Debug.
        Set Typeclasses Debug Verbosity 2.
        Set Typeclasses Depth 3.
        apply (@No_decode_le _).
```
I added `Set Debug "tactic-unification,unification".` and while the output is much longer, the slowest step is between the same two lines as before:
```
.003235886: Debug: considering goal 1 of status initial
.166363132: Debug: 1: looking for Univalence without backtracking
```
I'll attach the full file too: [ua-slow-verbose.txt](https://github.com/HoTT/Coq-HoTT/files/12849441/ua-slow-verbose.txt)

JasonGross: Hmm, suspicious.  I don't suppose `Set Printing All. Set Printing Universes.` gives anything interesting?

Summoning @ppedrot , do you know of any hotspots that would cause a .16 second lag between
```
Debug: considering goal 1 of status initial
```
and
```
Debug: 1: looking for Univalence without backtracking
```
in typeclass resolution?

JasonGross: Wait, how big is the context here @jdchristensen ?  Does `Time evar (ua2 : Univalence).` also take .16s?

theories/Spaces/No/Core.v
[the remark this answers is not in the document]
jdchristensen: `IsHProp`, `IsHSet` and `Contr` are notations, so I had to exclude them.  And we already set `IsTrunc` to be `Opaque` in typeclass search, so those lines are probably not needed in general.

I added the rest right before the slow spot here (again, not very slow, 0.15s) and it changed to 0.007s, as fast as my version.  In other words, with the above, I can just write `refine (No_ind_hprop _ _).` and it's as fast as my version.  Nice!  I checked, and it's just the `Hint Constants Opaque` that affects the speed here.

However, I'm not sure how practical it will be to add `Hint Constants Opaque` to the library in general.  We'd then need to add a *lot* of hints about unfolding throughout.  If someone wants to add `Hint Constants Opaque` to Overture.v and then go through and add `Hint Mode` as needed to make the library build, that would be an interesting experiment.

theories/Spaces/No/Core.v
jdchristensen: > Time evar (ua2 : Univalence).

That takes 0s.

jdchristensen: The context has 23 things in it, and (at least as displayed by default), none of them is particularly large.

jdchristensen: > I don't suppose `Set Printing All. Set Printing Universes.` gives anything interesting?

I didn't see anything that caught my eye in either the context or the goal with those set.  Both are still relatively compact.  And there's really no visible difference with the next goal a few lines down.  Also note that `Univalence` is in `Set`, in case that matters.

theories/Spaces/No/Core.v
```diff
@@ -1065,7 +1067,7 @@ Section RaiseSort.
       strip_truncations.
       destruct sh as [[l sh]|[r sh]].
       + apply lt_l with l.
-        apply IHL0, (@No_decode_le _).
+        apply IHL0, (@No_decode_le ua). (* Need to pass in [Univalence] to make this fast. *)
```
SkySkimmer, line 1070: Maybe the unification of `apply` does something faster when the univalence argument is not an evar?
Is refine faster than apply here?

jdchristensen: @SkySkimmer No, `refine` takes the same amount of time.  But here are some strange observations.  The last two are really mysterious to me!
```coq
        (* nrefine is "notypeclasses refine" *)
        Time apply (No_decode_le). (* slow *)
        Time apply (@No_decode_le). (* fails, unable to find Univalence *)
        Time apply (@No_decode_le ua). (* fast *)
        Time refine (@No_decode_le ua _ _ _ _). (* fast *)
        Time refine (@No_decode_le _ _ _ _ _). (* slow *)
        Time nrefine (@No_decode_le ua _ _ _ _). (* slow! *)
        Time nrefine (@No_decode_le _ _ _ _ _). (* slow, and leaves Univalence as a goal! *)
```
cc @JasonGross

jdchristensen: Oh, and with the last one, it's still slow to solve the isolated Univalence goal:
```coq
        Unshelve.
        Time 2: exact _.
```

SkySkimmer: It's also slow if you use `typeclasses eauto` instead of `exact _`?

jdchristensen: @SkySkimmer Yes, it's the same.  BTW, everything labelled "slow" takes essentially the same amount of time, around 0.28s on my machine, and every labelled "fast" is 0s or very close to 0s.  It's weird that the last one breaks the problem into two steps, and both steps take 0.28s.

jdchristensen: In the situation where all I'm trying to do is solve the goal `Univalence` using `exact _`, if I `clear` just this one thing from the context:
```coq
sh : le' {{ fun l : L => No_raise (xL l) | (fun r : R => No_raise (xR r)) // rxcut }} (No_raise (xL0 l))
```
then it becomes fast.  Strangely, an identical thing is in the context for the next goal, which is fast without manually supplying `ua`.

jdchristensen:
Thanks for all the ideas here.  The slowdown with univalence seems to be a quirk that hasn't bothered us elsewhere, so I'm fine with just hardcoding the solution that works here.  The slowdown with typeclass search involving truncation is a broader problem that could be tackled another time.  I'll open an issue.  For now, I'll just merge this.

theories/Spaces/No/Core.v
SkySkimmer: The problem seems to be in the generation of the "local" hint database corresponding to the local context. It reduces in an inefficient way just to find out that it's not a class.

SkySkimmer: https://github.com/coq/coq/pull/18145

theories/Spaces/No/Core.v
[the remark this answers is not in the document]
JasonGross: > then go through and add `Hint Mode` as needed to make the library build

It's not `Hint Mode` but `Hint Transparent` that needs to be added.  Roughly the rule of thumb is that you need to mark transparent any definition whose truncatedness you want tc to infer, but for which no trunc instance is declared.  (For example, if you want to use instances of `IsTrunc _ { a : A & ... }` interchangably with instances of `IsTrunc _ (hfiber ...)`).  There are a couple of exceptions, though I expect them to be quite rare for `IsTrunc`.  One example is that if you prove `forall (A : Type) (ls : list A), IsHProp (nil = ls)` and then want to automatically infer `forall (A B : Type) (f : A -> B) (ls : list B), IsHProp (map f nil = ls)`, you'll need to declare `map` transparent.  I wouldn't be all that surprised if none of the truncation instances broke.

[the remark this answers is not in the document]
jdchristensen: Thanks, @JasonGross!  I'll copy this to the issue as well.

PR #1781, opened 2023-10-10
Declare universe variables in Context commands

jdchristensen:
I searched the library and only found one more place where a universe variable was used in a Context command without being declared.  I also found an out-of-date comment, which only was relevant if the universe variable in question was not declared.


Alizter (approved):

Alizter:
Feel free to merge.

PR #1783, opened 2023-10-14
Naturality of join_assoc and trijoin_twist

jdchristensen:
This proves that `trijoin_twist` is natural.  Combined with the naturality of `join_sym`, this gives the naturality of `join_assoc`.  Direct proofs seemed hard, but by phrasing things using a new `functor_trijoin` defined using `trijoin_rec` it was possible to make this fairly clean, using naturality of `trijoin_rec` and `trijoinrecdata_fun_twist`.  Along the way, I removed rarely used arguments to `is1natural_natequiv` and `is1natural_nattrans`, and added `isnat_natequiv`, which is needed because another ambiguous coercion doesn't work.


Alizter (approved):
I've had a quick look and looks good to me.

PR #1784, opened 2023-10-17
Join: prove unitors, triangle law, and hexagon law

jdchristensen:
The first commit adds faithfulness of the Yoneda embedding, which is used in the next parts.  It also adds duals for the 0gpd results, which I hadn't added before.

The next two commits make progress towards showing that the join gives a symmetric monoidal structure.  I think all that is missing is the pentagon.

### DIFF
--- a/theories/Homotopy/Join/JoinAssoc.v
+++ b/theories/Homotopy/Join/JoinAssoc.v
@@ -158,7 +158,7 @@ Corollary join_join_power A n m
   : Join (join_power A n) (join_power A m) <~> join_power A (n + m)%nat.
 Proof.
   induction n as [|n IHn].
-  1: exact (equiv_join_empty' _).
+  1: exact (equiv_join_empty_left _).
   simpl. refine (_ oE (join_assoc _ _ _)^-1%equiv).
   exact (equiv_functor_join equiv_idmap IHn).
 Defined.
--- a/theories/Homotopy/Join/JoinAssoc.v
+++ b/theories/Homotopy/Join/JoinAssoc.v
@@ -153,6 +153,8 @@ Proof.
   apply equiv_join_sym.
 Defined.
 
+Arguments join_assoc : simpl never.
+
 (** As a consequence, we get associativity of powers. *)
 Corollary join_join_power A n m
   : Join (join_power A n) (join_power A m) <~> join_power A (n + m)%nat.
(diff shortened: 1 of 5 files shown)


theories/WildCat.v
```diff
@@ -11,6 +11,7 @@ Require Export WildCat.Yoneda.
 Require Export WildCat.Square.
 Require Export WildCat.PointedCat.
 Require Export WildCat.Bifunctor.
+Require Export WildCat.Monoidal.
```
Alizter, line 14: Completely forgot we had this! Looks like there are some incomplete (aborted) results in there still.

theories/Homotopy/Join/JoinAssoc.v
```diff
@@ -229,3 +231,185 @@ Proof.
...
+  lhs nrapply (functor_join_compose idmap _ idmap _).
+  lhs_V nrapply join_trianglelaw'.
+  unfold join_assoc; cbn.
+  apply join_sym_nat.
+Defined.
+
+(** ** The hexagon axiom *)
+
+(** This describes the transformation on [TriJoinRecData] corresponding to precomposition with [functor_join idmap (join_sym C B)], as in the next result. *)
+Definition trijoinrecdata_id_sym {A B C P} (f : TriJoinRecData A B C P)
```
Alizter, line 297: This can be defined ealier in TriJoin right?

Alizter (approved):
Great stuff!

Alizter:
@jdchristensen Will you prove the pentagon next? I had a look and it seems we will probably need a `QuadrupleJoin` helper with its own induction principle like we do for Tri. It might be possible to get by with just `trijoin_ind` but maybe there is a pattern for an n-fold join we can exploit here. It would be useful for higher coherences if we ever need it. But honestly I don't think we will ever right down a 2-monoidal category anytime soon, tricats are just too complicated.

theories/Homotopy/Join/JoinAssoc.v
jdchristensen: It could go in TriJoin, but I think it fits better in JoinAssoc.  `functor_join idmap (join_sym _ _)` is like the twist equivalence, but in the second and third variables, instead of the first and second.  So I think it fits in the file where the twist equivalence is studied.  It's also a fairly specific result, so is probably best kept close to where it is used.

jdchristensen:
> Will you try to prove the pentagon next?

@Alizter I'd like to have it, to complete the story, but I'm not sure how to handle the quadruple join.  The triple-join induction principle was a lot of work, and while I figured out ways to avoid much of the path algebra, there were still a few cases that needed a 3-cell to be filled.  I don't think it would be fun to figure out a 4-cell.  I'll think about whether there's a way to push through the pentagon without developing the general theory of quadruple joins...

PR #1785, opened 2023-10-17
Don't Export all of Classes.interfaces.abstract_algebra

jdchristensen:
Classes.interfaces.abstract_algebra exports Classes.interfaces.canonical names, which has a lot of things it.  I didn't want to change that export, since I'd have to fix a lot in the Classes library.

Instead, I adjusted the few files that exported abstract_algebra to either not do so, or do so very selectively.  This avoids putting all of the names and notations from Classes into the HoTT namespace.  For example, the Classes `Join` was conflicting with the `Join` we have, and the Classes `Join` was the winner.  Lots of files touched, but mainly because they need `Basics` and `Types` added in (and sometimes `Truncations.Core`), since canonical_names exports those.


theories/Algebra/Groups/Group.v
```diff
@@ -1,10 +1,32 @@
...
 Require Export Classes.theory.groups.
 Require Import Pointed.Core.
 Require Import WildCat.
 Require Import Spaces.Nat.Core.
+Require Import Truncations.Core.
+
+(** We repeat these notations from canonical_names, to avoid having all Classes notations in our name space.  todo: Is there a way to do "Require Export Classes.interfaces.canonical_names." and limit to just these notations? *)
+Infix "*" := sg_op : mc_mult_scope.
+Infix "+" := sg_op : mc_add_scope.
+Notation "- x" := (negate x) : mc_scope.
```
jdchristensen, line 25: Instead of repeating these, can I do something like
```coq
Require Export Classes.interfaces.canonical_names ("*", "-x").
```
Nothing I tried worked.

jdchristensen: (I know that
```coq
Require Export (notations) Classes.interfaces.canonical_names.
```
works, but I want to limit the exports.)

Alizter: One idea is to put these notations in their own module and export that instead.

jdchristensen: Thanks, that works.  I decided to make one module with most of the binary-operation notations, rather than getting more fine-grained than that.

Alizter (approved):
This seems better to me.

PR #1789, opened 2023-10-20
Speed up Colimit_Flattening

jdchristensen:
On my machine, the time goes from 3.6s to 1.3s.  There are only two slow lines, both marked with TODO, but they are the tricky kind:  a Defined (0.4s), and the closing of the Section (0.2s).  Normally I would just remove the section, but if I simply remove it and leave the Context command, then the file becomes slower overall.  It might work to copy the Context to every definition, but I didn't think that was worth it for 0.2s.

### DIFF
--- a/theories/Colimits/Colimit_Flattening.v
+++ b/theories/Colimits/Colimit_Flattening.v
@@ -114,7 +114,7 @@ Section Flattening.
       refine (_ @ concat_1p _).
       refine (concat_p_pp _ _ _ @ _).
       refine (_ @@ 1).
-      match goal with |- ap _ ?X @ _ = _ => set (p:=X) end.
+      match goal with |- ap _ ?X @ _ = _ => set (p := X) end.
       assert (r : transport_E'_V g x y = p^).
       { subst p.
         exact (moveL_transport_V_V E' _ _ _ _)^. }
@@ -158,7 +158,7 @@ Section Flattening.
       refine (_ @ concat_pp_p _ _ _).
       match goal with |- _ = (ap ?ff ?pp1 @ ?pp2) @ ?pp3
         => set (p1 := pp1) end.
-      assert (p1 = ap (transport E' (colimp i j g x)^)
+      assert (p1eq : p1 = ap (transport E' (colimp i j g x)^)
         (transport_pV E' (colimp i j g x) y)^).
       { subst p1; clear.
         etransitivity.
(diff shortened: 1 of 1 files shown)


jdchristensen:
The second commit is only indentation, so look at the other two commits separately.

Alizter (approved):

Alizter:
I wonder what we are spending most of the time on?

jdchristensen:
> I wonder what we are spending most of the time on?

One of the proofs is still quite long, with lots of path algebra.  To make it faster, one would probably need to abstract away some of the situation into one or more helper lemmas.  But these simpler changes at least make a big dent in the time.

JasonGross:
> but if I simply remove it and leave the Context command, then the file becomes slower overall

This might be because `Context` outside a `Section` actually means `Axiom`...

jdchristensen:
> This might be because `Context` outside a `Section` actually means `Axiom`...

@JasonGross Because of that, removing the `Section` does not have the semantics we want, but it's still very surprising to me that it makes the file process more slowly.  Do you have some intuition or insight about why having a few axioms would cause a file to be slower than have a `Section` with `Context`?

Alizter:
@jdchristensen It's been a while since I thought about this kind of thing, but my intuition is that axioms and qeds are generally slower for the Coq kernel to handle since it needs to backtrack on its reduction strategy.

When you close a section, the terms in the section go through a process known internally as "cooking" which makes sure that the correct context variables are abstracted as lambdas in order to make a closed term. Notably the difference between the non-section term is the fact that the axioms are now variables.

So I would guess that when working with an axiom, it is somehow treated differently than working with a variable hence the difference in speed.

That's what I imagine is going on, but I could be totally wrong. @SkySkimmer might be able to correct me or offer an insight.

JasonGross:
> > This might be because `Context` outside a `Section` actually means `Axiom`...
> 
> @JasonGross Because of that, removing the `Section` does not have the semantics we want, but it's still very surprising to me that it makes the file process more slowly. Do you have some intuition or insight about why having a few axioms would cause a file to be slower than have a `Section` with `Context`?

Two guesses:
1. Context variables have a single universe assigned to them for the whole section, while axioms can be universe polymorphic.  So maybe removing the section added a bunch of universe variables everywhere
2. If any of the context variables got introduced as typeclasses maybe they persisted past the close of the original section and slowed down later typeclass search

JasonGross:
Also, @Alizter I don't see how your description could be accurate.  The closest thing to that I can imagine happening is that maybe some later tactic or unification heuristic is extremely sensitive to how many arguments some application node has, and so having your constant be applied to, e.g., only 5 things instead of 7 because the first two became axioms rather than arguments made some heuristic take a wrong path.

jdchristensen:
@JasonGross I should have pointed out that the entire file was wrapped in that section, so there is no later material that could get slower when the section is removed.  That's what's puzzling to me.  Since the `End` of the section took a bit of time, it seemed to me like removing the section should only help things...

PR #1790, opened 2023-10-20
ReflectiveSubuniverse: workaround no longer needed

jdchristensen:
Remove a workaround for Coq <= 8.11.

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -903,10 +903,7 @@ Section Reflective_Subuniverse.
                `{In O A} {B_inO : In O {x:A & B x}} (x : A)
     : In O (B x).
     Proof.
-      refine (inO_equiv_inO _ (hfiber_fibration x B)^-1).
-      (** TODO: Why doesn't Coq find this instance? *)
-      (* This is #12571 and it is only needed for Coq <= 8.11 *)
-      all: refine (inO_hfiber pr1 x); assumption.
+      exact (inO_equiv_inO _ (hfiber_fibration x B)^-1).
     Defined.
 
     #[local]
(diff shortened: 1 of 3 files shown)


mikeshulman:
So is Coq bug 4533 not actually fixed?

Alizter:
> Comment author: @mattam82
> 
> Both bugs are now fixed with the new strategy of keyed unification.

Are we using keyed unification? Do we want to be using keyed unification?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -903,10 +903,7 @@ Section Reflective_Subuniverse.
                `{In O A} {B_inO : In O {x:A & B x}} (x : A)
     : In O (B x).
     Proof.
-      refine (inO_equiv_inO _ (hfiber_fibration x B)^-1).
-      (** TODO: Why doesn't Coq find this instance? *)
-      (* This is #12571 and it is only needed for Coq <= 8.11 *)
-      all: refine (inO_hfiber pr1 x); assumption.
+      exact (inO_equiv_inO _ (hfiber_fibration x B)^-1).
```
Alizter, line 906: You could probably avoid proof mode all together here.

jdchristensen: Right!  Done.

jdchristensen:
I'm not sure whether the bug that was "fixed" was similar but not identical to this issue, or whether it was not actually fixed.

jdchristensen:
Tagging @JasonGross , since the bug in question ([Coq Bug 4533](https://github.com/coq/coq/issues/4533)) was based on comments he made.

jdchristensen:
@JasonGross Any thoughts on this one?  If I don't hear anything soon, I'll just go ahead and merge this.

jdchristensen:
> I don't have thoughts; can we extract a standalone example with the bug minimizer and open a new issue on coq?

@JasonGross I'm on the road, and my laptop doesn't have python2.  Would it be easy for you to generate the standalone example?  I just pushed a change to Localization.v which makes it clear exactly which rewrite is the issue and exactly which term gets unfolded unnecessarily.

BTW, @Alizter, we are using Keyed Unification already.

I'm going to merge this since I don't think we're going to solve it in this PR.

PR #1791, opened 2023-10-23
Reduce universe variables, mainly in WildCat

jdchristensen:
Because of [Coq issue 10513 ](https://github.com/coq/coq/issues/10513), unused `Context` variables introduce floating universe variables.  Floating universe variables are particularly annoying because they accumulate, as they are never forced to be equal to other variables.  The very small changes in the first commit cause an out-of-tree definition to go from having 623 universe variables to having 5 universe variables.

The second commit adds a tactic `simple_path_induction` which is to path induction as `simple_induction` is to induction over the naturals.  It avoids introducing universe variables in the induction.  I use this in Join/Core, as an example.  It turns out to not have been the cause of the blow up, but I think it's good to have.  I put this in Basics/Tactics, and moved `simple_induction` there as well.  (Overture.v should be the first file someone reads, and it has a lot of technical things that might be confusing for a beginning, so I'm keeping my eyes open for things that can be moved or removed.)

[Edit: `simple_path_induction` removed since `destruct p` accomplishes the same thing.]

### DIFF
--- a/theories/Basics/Tactics.v
+++ b/theories/Basics/Tactics.v
@@ -15,14 +15,6 @@ Ltac simple_induction n n' IH :=
   intros [| n'];
   [ clear IH | specialize (IH n') ].
 
-(** This should be equivalent to [induction p], except that it may leave additional things from the context reverted.  Using this avoids introducing extra universe variables. *)
-Ltac simple_path_induction p :=
-  match type of p with (_ = ?x) =>
-    move p at top; (* After the next line, we want the first two variables to be [x] and then [p]. *)
-    generalize dependent x;
-    refine (fun x p => match p with 1 => _ end);
-    clear x p end.
-
 (** Debugging tactics to show the goal during evaluation. *)
 
 Ltac show_goal := match goal with [ |- ?T ] => idtac T end.
--- a/theories/Basics/Tactics.v
+++ b/theories/Basics/Tactics.v
@@ -1,8 +1,27 @@
 Require Import Basics.Overture.
 
-
 (** TODO: Clean up *)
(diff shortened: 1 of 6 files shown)


theories/Basics/Tactics.v
```diff
@@ -1,8 +1,27 @@
...
+
+(** The following tactic is designed to be more or less interchangeable with [induction n as [ | n' IH ]] whenever [n] is a [nat] or a [trunc_index].  The difference is that it produces proof terms involving [fix] explicitly rather than [nat_ind] or [trunc_index_ind], and therefore does not introduce higher universe parameters. *)
+Ltac simple_induction n n' IH :=
+  generalize dependent n;
+  fix IH 1;
+  intros [| n'];
+  [ clear IH | specialize (IH n') ].
+
+(** This should be equivalent to [induction p], except that it may leave additional things from the context reverted.  Using this avoids introducing extra universe variables. *)
+Ltac simple_path_induction p :=
```
SkySkimmer, line 19: When is this better than destruct? Is it ever better than destruct?

jdchristensen: Oh, thanks, you're right.  I didn't realize that that was a difference between `destruct` and `induction`.  I've reversed that part.

Alizter (approved):

Alizter:
A further improvement would be to assert the universe levels in the tests by doing `Check foo@{u1 u2}`. That way we can be sure we didn't cause a universe explosion when refactoring in the future.

jdchristensen:
@Alizter Good idea.  Done.  I'll merge after the CI completes.

PR #1794, opened 2023-11-05
Basics/Overture: clarify things related to internal_paths_rew

jdchristensen:
The first commit:
- Explains more clearly why `paths_rew` and `paths_rew_r` are defined, even though they are never used.  It also renames them and makes them `Local`.
- It makes some arguments to `internal_paths_rew` (and `rew_r`) implicit.  Now these look just like `transport`, except for the argument order.
- I also added a remark that `internal_paths_rew` is in fact definitionally equal to `transport`.  It's too bad that it's `internal_paths_rew_r` that arises from the default direction of `rewrite`...
- I removed `paths_rect_r`, which is never used.  The comment suggests that in the past it was needed for `rewrite <-`?

The second commit just moves the `internal_paths_rew` stuff to be right near `transport`, and fixes some other whitespace issues.  So it's best to look at the first commit in isolation to see the substantial changes clearly.

I considered adding
```coq
Ltac paths_rew_to_transport :=
  repeat match goal with |- context [ internal_paths_rew ?P ?u ?p ] =>
                           change (internal_paths_rew P u p) with (transport P p u) end.
```
to Tactics.v, but since it's `rew_r` that is more common, this probably isn't that useful.

(It sure would be nice if we could register `transport` as the lemma that rewrite should use in one direction, and `fun ... => transport P p^ u` to be what to use in the other direction.  I may open a Coq issue suggesting this.)

### DIFF
--- a/test/bugs/github1794.v
+++ b/test/bugs/github1794.v
@@ -1,0 +1,6 @@
+From HoTT Require Import Basics.Overture.
+
+(** When [rewrite] is first used, it defines helper lemmas.  If the first use is in a Section that has universe variables, then these lemmas get unnecessary universe variables.  Overture uses [rewrite] outside of a section to ensure that they have the expected number of universe variables, and we test that here. *)
+
+Check internal_paths_rew@{u1 u2}.
+Check internal_paths_rew_r@{u1 u2}.
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -328,6 +328,7 @@ Arguments transport {A}%type_scope P%function_scope {x y} p%path_scope u : simpl
 Notation "p # x" := (transport _ p x) (only parsing) : path_scope.
 
 (** The first time [rewrite] is used in each direction, it creates transport lemmas called [internal_paths_rew] and [internal_paths_rew_r].  See ../Tactics.v for how these compare to [transport].  We use [rewrite] here to trigger the creation of these lemmas.  This ensures that they are defined outside of sections, so they are not unnecessarily polymorphic.  The lemmas below are not used in the library. *)
+(** TODO: If Coq PR#18299 is merged (possibly in Coq 8.20), then we can instead register wrappers for [transport] to be used for rewriting.  See the comment by Dan Christensen in that PR for how to do this.  Then the tactics [internal_paths_rew_to_transport] and [rewrite_to_transport] can be removed from ../Tactics.v. *)
 Local Lemma define_internal_paths_rew A x y P (u : P x) (H : x = y :> A) : P y.
 Proof. rewrite <- H. exact u. Defined.
 
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -223,12 +223,15 @@ Register paths_rect as core.identity.ind.
 Notation "x = y :> A" := (@paths A x y) : type_scope.
 Notation "x = y" := (x = y :>_) : type_scope.
 
-(** Ensure [internal_paths_rew] and [internal_paths_rew_r] are defined outside sections, so they are not unnecessarily polymorphic. *)
-Lemma paths_rew A a y P (X : P a) (H : a = y :> A) : P y.
-Proof. rewrite <- H. exact X. Defined.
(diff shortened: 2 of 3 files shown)


jdchristensen:
The third commit adds a test for this.  I also verified that `internal_paths_rew` gets free universe variables if the first use of rewrite is in a Section with additional universe variables, so we do still need this trick in Overture.v.

jdchristensen:
We can also replace `internal_paths_rew_r P u p` with `transport P p^ u` propositionally, using `rewrite`, so maybe it is worth having a tactic that does both replacements.  That's what the fourth commit does.

Alizter (approved):
Overall, LGTM. I'm unsure how useful the tactic is, but I guess it cannot hurt to have it around if such a situation arises. Generally, if we need to reason about a proof that used rewrite, we would redo it using path algebra.

SkySkimmer:
Would it work to do `Definition internal_paths_rew_r P u p := transport P p^ u` (then `unfold internal_paths_rew_r` turns it into `transport`) or does it need to be generated by rewrite for some reason?

jdchristensen:
> Would it work to do `Definition internal_paths_rew_r P u p := transport P p^ u` (then `unfold internal_paths_rew_r` turns it into `transport`) or does it need to be generated by rewrite for some reason?

@SkySkimmer I tried that, but then `rewrite` generates a fresh `internal_paths_rew_r0`, it seems.

jdchristensen:
> Overall, LGTM. I'm unsure how useful the tactic is, but I guess it cannot hurt to have it around if such a situation arises. Generally, if we need to reason about a proof that used rewrite, we would redo it using path algebra.

@Alizter Part of the point of my exploration of this is that `rewrite` really "is" just `transport`.  If this could be made definitionally true, then we'd immediately have all of the lemmas about transport available to us, so we'd be able to reason about proofs made using `rewrite`.

Alizter:
One idea we've had previously to achieve this is to use setoid rewrite. This is a little tricky to do in a prelude agnostic way however.

jdchristensen:
@SkySkimmer Here's an example, where I cut-and-paste the definition of `internals_paths_rew_r`, and it is still redefined:
```coq
Definition internal_paths_rew_r :=
fun (A : Type) (a y : A) (P : A -> Type) (HC : P y) (X : a = y) =>
match X in (_ = y0) return (P y0 -> P a) with
| 1 => idmap
end HC.

Lemma define_internal_paths_rew_r A x y P (u : P y) (H : x = y :> A) : P x.
Proof. rewrite -> H. exact u. Defined.

Print define_internal_paths_rew_r.
(* fun (A : Type) (x y : A) (P : A -> Type) (u : P y) (H : x = y) =>
   internal_paths_rew_r0 A x y (fun x0 : A => P x0) u H
   : forall (A : Type) (x y : A) (P : A -> Type), P y -> x = y -> P x *)
```
Note the `0` in `internal_paths_rew_r0`.

Do you know a work-around for this?

jdchristensen:
@SkySkimmer [Your PR](https://github.com/coq/coq/pull/18299) sounds like it would be just what we need.  Can you give an example of how it would be used?

If that PR is merged, then I'll update this one to add a comment saying that most of what is here can be replaced with the new registration mechanism once the next Coq version is released.

SkySkimmer:
It would look like
~~~coq
Definition paths_rew := ...
Register Scheme paths_rew as rew for eq.
~~~
(not sure if it's `as rew` or `as somethingelse`, would have to check)

jdchristensen:
I tested that with this PR, one can do the following in Basics/Overture.v in the Coq-HoTT library:
```coq
Definition paths_rew {A : Type} {x y : A} (P : A -> Type) (u : P x) (p : x = y)
  : P y
  := transport P p u.

Register Scheme paths_rew as rew for paths.

Definition paths_rew_r {A : Type} {x y : A} (P : A -> Type) (u : P y) (p : x = y)
  : P x
  := transport P p^ u.

Register Scheme paths_rew_r as rew_r for paths.
```
The library builds with no changes, and after unfolding the wrappers, you get `transport`.  Very nice!

I had originally tried this version:
```coq
Register Scheme transport as rew for paths.

Definition transport_V {A : Type} {x y : A} (P : A -> Type) (p : x = y) (u : P y) 
  : P x
  := transport P p^ u.

Register Scheme transport_V as rew_r for paths.
```
The only difference is the order of the arguments `u` and `p`, making them match `transport`.  But then the goals that `rewrite` generates are in a different order, which makes `rewrite 2 lemma` and `rewrite lemma by tactic` not work well.  The core of the HoTT library builds with just one minor change, but the Categories and Classes parts of the library use these variants of rewrite in a number of places, so it's probably best to use the wrappers with the arguments in the other order.

jdchristensen:
The Coq PR is scheduled for Coq 8.20, so I added a TODO comment about it, and think we should go ahead and merge this PR now.  I'll do so unless I hear any objections.

PR #1795, opened 2023-12-28
2-cat of pointed types

Alizter:
We show the (2,1)-category structure for pointed types. In order to do this easily, we introduce a new tactic called pelim which allows easy path induction on the pointed equation of various pointed structures. This is particularly useful in removing funext from various pType coherences.

Various places seemed to break from these changes, but could easily be fixed by replacing pointed notations with wildcat notations. This perhaps hints at a larger refactoring that should be done.

My motivation for showing these coherences is the need for 1-functoriality of pre/postcomposition for some unrelated work. Therefore it made sense to quickly generalize to the full 2-category while I was at it.

### DIFF
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -261,8 +261,8 @@ Global Instance phomotopy_symmetric {A B} : Symmetric (@pHomotopy A B).
 Proof.
   intros f g p.
   snrefine (Build_pHomotopy _ _); cbn.
-  - intros x; exact ((p x)^).
-  - refine (inverse2 (dpoint_eq p) @ inv_pV _ _).
+  1: intros x; exact ((p x)^).
+  by pelim p f g.
 Defined.
 
 Notation "p ^*" := (phomotopy_symmetric _ _ p) : pointed_scope.
@@ -285,10 +285,8 @@ Definition pmap_postwhisker {A B C : pType} {f g : A ->* B}
   : h o* f ==* h o* g.
 Proof.
   snrefine (Build_pHomotopy _ _); cbn.
-  - exact (fun x => ap h (p x)).
-  - nrefine (ap _ (dpoint_eq p) @ ap_pp _ _ _ @ whiskerL _ (ap_V _ _) @ _^).
-    nrefine (whiskerL _ (inv_pp _ _) @ concat_pp_p _ _ _ @ _).
-    exact (whiskerL _ (concat_p_Vp _ _)).
+  1: exact (fun x => ap h (p x)).
+  by pelim p f g h. 
 Defined.
 
 Definition pmap_prewhisker {A B C : pType} (f : A ->* B)
@@ -296,10 +294,8 @@ Definition pmap_prewhisker {A B C : pType} (f : A ->* B)
   : g o* f ==* h o* f.
 Proof.
   snrefine (Build_pHomotopy _ _); cbn.
-  - exact (fun x => p (f x)).
-  - apply moveL_pV.
-    nrefine (concat_p_pp _ _ _ @ whiskerR (concat_Ap p (point_eq f))^ _ @ _).
-    exact (concat_pp_p _ _ _ @ whiskerL _ (point_htpy p)).
+  1: exact (fun x => p (f x)).
+  by pelim f p g h.
 Defined.
 
 (** ** 1-categorical properties of [pType]. *)
@@ -311,10 +307,7 @@ Definition pmap_compose_assoc {A B C D : pType} (h : C ->* D)
 Proof.
   snrapply Build_pHomotopy.
   1: reflexivity.
-  pointed_reduce_pmap f.
-  pointed_reduce_pmap g.
-  pointed_reduce_pmap h.
-  reflexivity.
+  by pelim f g h.
 Defined.
 
 (** precomposition of identity pointed map *)
@@ -323,8 +316,7 @@ Definition pmap_precompose_idmap {A B : pType} (f : A ->* B)
 Proof.
   snrapply Build_pHomotopy.
   1: reflexivity.
-  symmetry.
-  apply concat_pp_V.
+  by pelim f.
 Defined.
 
 (** postcomposition of identity pointed map *)
@@ -333,10 +325,7 @@ Definition pmap_postcompose_idmap {A B : pType} (f : A ->* B)
 Proof.
   snrapply Build_pHomotopy.
   1: reflexivity.
-  symmetry.
-  apply moveR_pV.
-  simpl.
-  apply equiv_p1_1q, ap_idmap.
+  by pelim f.
 Defined.
 
 (** ** Funext for pointed types and direct consequences. *)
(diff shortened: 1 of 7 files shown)


jdchristensen:
Nice!  I'll take a close look tomorrow.  A couple of questions for now.  First, how does `pelim` compare with `pointed_reduce` and `pointed_reduce_pmap`?  Can we unify them?  Second, I'm surprised that just adding the (2,1)-category structure made various pointed notations like `<~>*` need to be replaced with wildcat notations.  Do you have any idea why this happens?  One issue with wildcat notations is that you don't always know which wildcat Coq is choosing to interpret your notation with...

Alizter:
> Nice! I'll take a close look tomorrow. A couple of questions for now. First, how does `pelim` compare with `pointed_reduce` and `pointed_reduce_pmap`? Can we unify them? 

I originally tried to modify `pointed_reduce` to avoid using `path_induction` and instead use `pelim`, however it changed too much and pelim is a little slower due to the reduction I needed to do in the tactic.

Also it is not entirely trivial the order in which to do `pelim` on the hypotheses. If you look at the functoriality of the assocciators in the 2-category definition, you will see that the pelim order for similar proofs vary subtly in their order. I don't think it will be entirely trivial to get tactics to choose this order.

I think having the manual control is a little better. I would however like to improve the performance of `pelim` before we start using it more widely since it does more unfolding than it probably needs. The main idea comes from recognising that point equations typically have an application on the LHS and when you do a `generalize dependent` of the pointed structure applied to its basepoint you get a hypothesis that almost always fits `paths_ind_r`.

Thinking about it, using right-handed path induction consistently should lead to a far superior `pointed_elim`, therefore it might be worth investigating swapping the regular path induction for the right handed side at some point.

>Second, I'm surprised that just adding the (2,1)-category structure made various pointed notations like `<~>*` need to be replaced with wildcat notations. Do you have any idea why this happens?

I was also surprised, but I think this is because extra groupoid instances throw off Coq's unification. This is apparent in the fact that a few proofs had to be changed from `srapply` to `snrapply` followed by `exact`. I'm not entirely sure this is the full picture however.

> One issue with wildcat notations is that you don't always know which wildcat Coq is choosing to interpret your notation with...

Yes that might be true. I think the better option would be to redefine `->*` as a specific form of `$->`, but I haven't checked if this is fully sensible just yet. (Especially if it changes universe numbers).

Alizter:
I've just pushed another commit simplifying a few other proofs, but we can undo those if we don't feel they are necessary.

jdchristensen:
Ok, I'm reviewing now, so hold off on pushing anything else for a bit.

theories/Pointed/Core.v
```diff
@@ -765,13 +751,75 @@ Proof.
...
 (** pForall is a 1-groupoid *)
-Global Instance is1gpd_pforall `{Funext} (A : pType) (P : pFam A) : Is1Gpd (pForall A P).
+Global Instance is1gpd_pforall (A : pType) (P : pFam A) : Is1Gpd (pForall A P) | 10.
 Proof.
   econstructor.
   + intros ? ? p. exact (phomotopy_compose_pV p).
   + intros ? ? p. exact (phomotopy_compose_Vp p).
 Defined.
 
+Global Instance is3graph_ptype : Is3Graph pType := fun f g => _.
```
jdchristensen, line 761: It was not obvious what the underscore is replaced with here, so maybe it's better to be explicit?

theories/Pointed/Core.v
```diff
@@ -575,46 +564,38 @@ Definition phomotopy_compose_p1 {A : pType} {P : pFam A} {f g : pForall A P}
...
+  1: intro x; apply concat_Vp.
+  by pelim p f g.
 Defined.
 
 Definition equiv_phomotopy_concat_l `{Funext} {A B : pType}
   (f g h : A ->* B) (K : g ==* f)
   : f ==* h <~> g ==* h.
 Proof.
-  refine ((equiv_path_pforall _ _)^-1%equiv oE _ oE equiv_path_pforall _ _).
+refine ((equiv_path_pforall _ _)^-1%equiv oE _ oE equiv_path_pforall _ _).
```
Alizter, line 598: This is a mistake, I will undo.

theories/Pointed/Core.v
```diff
@@ -268,21 +285,17 @@ Definition pmap_postwhisker {A B C : pType} {f g : A ->* B}
   : h o* f ==* h o* g.
 Proof.
   snrefine (Build_pHomotopy _ _); cbn.
-  - exact (fun x => ap h (p x)).
-  - nrefine (ap _ (dpoint_eq p) @ ap_pp _ _ _ @ whiskerL _ (ap_V _ _) @ _^).
-    nrefine (whiskerL _ (inv_pp _ _) @ concat_pp_p _ _ _ @ _).
-    exact (whiskerL _ (concat_p_Vp _ _)).
+  1: exact (fun x => ap h (p x)).
+  by pelim p f g h.
```
jdchristensen, line 289:
```suggestion
  by pelim p f g h.
```
(Remove trailing whitespace.)

theories/Pointed/Core.v
```diff
@@ -525,91 +531,71 @@ Notation "'ppforall'  x .. y , P"
...
+  1: intro x; apply concat_Vp.
+  by pelim p f g.
 Defined.
 
 Definition equiv_phomotopy_concat_l `{Funext} {A B : pType}
   (f g h : A ->* B) (K : g ==* f)
   : f ==* h <~> g ==* h.
 Proof.
-  refine ((equiv_path_pforall _ _)^-1%equiv oE _ oE equiv_path_pforall _ _).
+refine ((equiv_path_pforall _ _)^-1%equiv oE _ oE equiv_path_pforall _ _).
```
jdchristensen, line 598:
```suggestion
  refine ((equiv_path_pforall _ _)^-1%equiv oE _ oE equiv_path_pforall _ _).
```

theories/Pointed/Core.v
```diff
@@ -765,13 +751,75 @@ Proof.
...
   + intros ? ? p. exact (phomotopy_compose_pV p).
   + intros ? ? p. exact (phomotopy_compose_Vp p).
 Defined.
 
+Global Instance is3graph_ptype : Is3Graph pType := fun f g => _.
+
+Global Instance is21cat_ptype : Is21Cat pType.
+Proof.
+  unshelve econstructor.
+  1: exact _.
```
jdchristensen, line 766:
```suggestion
  - exact _.
```
(May as well unify style.)

theories/Pointed/Core.v
```diff
@@ -765,13 +751,75 @@ Proof.
...
+  - intros f g h p; nrapply Build_Is1Functor.
```
jdchristensen, line 767: The variable names are misleading here.  `f`, `g` and `h` are pointed types, and `p` is a pointed function.  Can you fix these and check the rest?

jdchristensen (approved):
LGTM!  Feel free to merge once the minor things I "pointed" out are fixed.

jdchristensen:
The new `pelim` works very well!  I see that it plays a slightly different role than `pointed_reduce`.  `pointed_reduce` is more ambitious, trying to replace a pointed function with a strictly pointed function, while `pelim` is designed to handle things once the functions aren't needed any more, and only pointedness conditions remain.  This is why the order of the arguments to `pelim` is important.

theories/Pointed/Core.v
Alizter: Since there is a graph structure on pForall, Coq is able to use the instances from before. I think rather than being explicit, I will add a comment mentioning what it is.

theories/Pointed/Core.v
Alizter: I'll try to do a whitespace trim of my own in case any slipped in.

Alizter:
@jdchristensen I can say that I didn't design `pelim` on purpose, it was a happy accident. I was simply trying to remove some funext from the precomposition lemmas and noticed that this pattern I was doing kept working so I turned it into a tactic. I'm glad it worked out in the end, and I hope it can be more widely useful. Especially in the cases where `pointed_reduce` leaves a poorly reduced goal.

theories/Pointed/Core.v
Alizter: Actually, I've just replaced it with the explicit instance.

jdchristensen:
> and I hope it can be more widely useful.

I may use it today after this is merged!

theories/Pointed/Core.v
```diff
@@ -186,6 +186,23 @@ Ltac pointed_reduce_pmap f
     | _ ->* ?Y => let p := fresh in destruct Y as [Y ?], f as [f p]; cbn in *; destruct p; cbn
     end.
 
+(** General tactic to reduce any pointed equations of a pForall *)
```
jdchristensen, line 189:
```suggestion
(** A general tactic to replace pointedness paths in a pForall with reflexivity.  It clears the function, so can only be applied once the function itself is not longer needed. *)
```

theories/Pointed/Core.v
```diff
@@ -186,6 +186,23 @@ Ltac pointed_reduce_pmap f
...
+Ltac pelim q :=
+  destruct q as [q ?];
+  unfold pointed_fun, point_htpy in *;    
+  cbn;
+  generalize dependent (q pt);
+  rapply paths_ind_r;
```
jdchristensen, line 195:
```suggestion
  nrapply paths_ind_r;
```
Does it work with `nrapply`?  That might be safer.

Alizter:
@jdchristensen All comments addressed. I'll merge once the CI is green.

theories/Pointed/Core.v
jdchristensen: And maybe use `try clear q`?  That might allow the tactic to be used in a few situations when the function remains in the goal?

Alizter:
@jdchristensen I've made the `clear`ing a `try`. It doesn't affect anything but it might be useful for the reasons you've stated, although I haven't encountered such a situation yet.

jdchristensen:
I think it's ok if you merge now that one build has succeeded.

jdchristensen:
Did you do a squash merge by accident?  Just a reminder that that's not our usual procedure.

Alizter:
@jdchristensen Yes. I forgot that we preferred not to do this. I've gone ahead and forbid them from the repository settings so now only merge is available.

PR #1799, opened 2024-01-02
1-category of successor structures

Alizter:
One of the nice things about pType is the potential to define the wildcat structure "all the way up". We don't actually do this, but because pHomotopy is defined as pForall we get all the higher morphisms "for free".

I was interested in working out what exactly makes a "forall" in a category tick. From what I could find, its not really a commonly studied phenomenon. Typically in category theory, such "forall" looking sets of dependent functions carry an extra structure of being internal. This is a strong requirement that makes it impossible to replicate in various categories (keep Spectra in mind). However the lesson to learn from pType is that we almost never care about ppForall but only pForall which has nothing to do with the internal (dependent) hom.

Therefore I was interested in seeing if we can push this idea in other categories, notably the category of spectra. That would allow us to have a very nice category to work with.

I believe this is possible but it will require a bit more work on understanding the exact definitions of family of spectra over a spectrum and how a hypothetical "sForall" might be constructed. The route of paramaterized spectra appears to somewhat be a red-herring and doesn't really gain us much.

Anyway, to give an initial proof-of-concept of the wider application of the pType-pforall technique, I decided to choose a structure which is not entirely trivial but still simple to work with. The toy example I chose in the end is the category of successor strucutres.

I don't believe that this work will lead to a deeper understanding of the theory of successor structures, however it does demonstrate that the forall-technique can be generalised.

The basic recipe is as follows:
- Start with a category C.
- Define a "displayed" version of the objects of C. Notably this should be a family over C.
- Define the homotopy family.
- Define foralls and higher homotopies between them using the family. 

Interesting things to note:

- My pelim tactic from before has an analogue here.
- The structure of SuccStr has a non-trivial endomap which can be "displayed" appropriately.
- We get a full 1-cat structure and 2-cat with some more work.

I would like to eventually try this technique on spectra once we work out what the correct way to have displayed spectra is. These are not parameterized spectra as the base space must also be a spectrum. It should probably consist of pFam for each pType in spextra and some how a displayed pointed map which I have yet to formulate.

I think that this work so far is harmless so I would advocate including it. It might be worth considering a way to axiomatize what a "forall" is in a category, but I think even just defining how to "display" things will be quite non-trivial.

cc @jdchristensen and @mikeshulman (who might be interested in the displayed part).

### DIFF
--- a/theories/Homotopy/SuccessorStructure.v
+++ b/theories/Homotopy/SuccessorStructure.v
@@ -1,17 +1,16 @@
 Require Import Basics.
 Require Import Spaces.Int.
 Require Import Spaces.Finite.Fin.
+Require Import WildCat.
 
 (** * Successor Structures. *)
 
 (** A successor structure is just a type with a endofunctor on it, called 'successor'. Typical examples include either the integers or natural numbers with the successor (or predecessor) operation. *)
 
 Record SuccStr : Type := {
-   ss_carrier : Type ;
+   ss_carrier :> Type ;
    ss_succ : ss_carrier -> ss_carrier ;
 }.
-
-Coercion ss_carrier : SuccStr >-> Sortclass.
 
 Declare Scope succ_scope.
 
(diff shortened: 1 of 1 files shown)


theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
 Notation "'N3'" := (Stratified (+N) 3) : succ_scope.
 Notation "'Z3'" := (Stratified (+Z) 3) : succ_scope.
 
+(** ** Category of successor structures *)
+
+(** Insipired by the construction of the wildcat structure on pType, we can give SuccStr a wildcat structure in a similar manner (all the way up). *)
```
jdchristensen, line 76:
```suggestion
(** Inspired by the construction of the wildcat structure on pType, we can give SuccStr a wildcat structure in a similar manner (all the way up). *)
```

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+
+Record ssFam (A : SuccStr) := {
+  ss_pr1 :> A -> Type;
```
jdchristensen, line 79: `ss_pr1` seems like an odd name for this, but I see that `pFam` uses the same convention.  What about something like `ss_fam` or `ssfam`?  (Even if in your head you think of the sigma-type, `pr2` would then be the analog of this map, not `pr1`.)  OTOH, I guess this is the displayed version of the first projection, so maybe that's the motivation.  Still, I'd lean towards another notation.

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+  dss_succ {x} : ss_pr1 x -> ss_pr1 (x.+1);
```
jdchristensen, line 80: I guess the `d` is short for `dependent` or `displayed`?  Maybe `ss_fam_succ` would be more clear?

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+}.
+  
+Arguments ss_pr1 {_ _} _.
+Arguments dss_succ {_ _ _}.
+
+Record ssForall {A : SuccStr} (B : ssFam A) := {
+  ss_fun :> forall x, B x;
+  ss_fun_pt : forall x, ss_fun x.+1 = dss_succ (ss_fun x);
```
jdchristensen, line 88: Sorry to dwell on naming, but `pt` seems wrong.  Maybe `ss_fun_nat`?  Or `ss_fun_succ`?

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+
+Arguments ss_fun {_ _} _ _.
+Arguments ss_fun_pt {_ _} _ _.
+
+Definition ssfam_const {A : SuccStr} (B : SuccStr) : ssFam A.
+Proof.
+  snrapply Build_ssFam.
+  1: exact (fun _ => B).
+  intros x; exact ss_succ.
+Defined.
```
jdchristensen, line 99: Maybe a one-liner is clearer?  ` := Build_ssFam _ (fun _ => B) (fun _ => ss_succ).`

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+  : ssFam A.
+Proof.
+  snrapply Build_ssFam.
+  1: exact (fun x => f x = g x).
+  cbn; intros x p.
+  refine (ss_fun_pt f x @ ap dss_succ p @ (ss_fun_pt g x)^).
+Defined.
+
+Definition ssHomotopy {A : SuccStr} {P : ssFam A} (f g : ssForall P) : Type
+  := ssForall (ssfam_sshomotopy f g).
```
jdchristensen, line 111: Maybe omit the `: Type` to save a universe?

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+      match goal with
+      | y : Y |- _ => sselim_elim eq (f y)
+      | g : ?Z -> Y |- _ =>
+        match goal with
+        | z : Z |- _ => sselim_elim eq (f (g z))
+        end
+      end
+    | _ => idtac "sselim: no hyp found"; fail
+    end
+  | _ => idtac "sselim: no eq found"; fail
```
jdchristensen, line 160: In proof general, these messages are hidden by the "Tactic failure" message.  But it works if you move the message to the `fail` tactic: `fail "sselim: no eq found"`.

theories/Homotopy/SuccessorStructure.v
```diff
@@ -72,3 +71,190 @@ Defined.
...
+    snrapply Build_Is0Functor.
+    intros f h p.
+    snrapply Build_ssForall.
+    + intro x.
+      exact (ap g (p x)).
+    + intro x; cbn.
+      sselim p f h.
+      destruct (p x).
+      clear p.
+      simpl.
```
jdchristensen, line 233:
```suggestion
      destruct (p x); clear p; simpl.
```

jdchristensen (approved):
I think it's fine to include this as another example of creating a wild category.  I made some minor suggestions, and maybe it's worth waiting a day to see if @mikeshulman has any comments?

mikeshulman:
I don't have time to look at this in detail, but it does look interesting!  I'll be interested to see how far it can be generalized.

Regarding the question of how to define displayed spectra and other things, I wonder whether https://arxiv.org/abs/2311.18781 could be relevant?  In that paper we formulated a type theory with a generic "display" operation that acts on (among other things) any type defined in the empty context to compute a notion of "displayed elements" of that type.

theories/Homotopy/SuccessorStructure.v
Alizter: Yes, it is supposed to be the displayed point. A similar convention is used in pFam with dpoint.

Alizter: I think I will keep the name as it is a displayed version of succ.

theories/Homotopy/SuccessorStructure.v
Alizter: Yes, that's totally me copying from pType here and being lazy. I'll fix right away.

theories/Homotopy/SuccessorStructure.v
Alizter: I'm using coq-lsp and both are reported. I didn't know "fail" took arguments, so I'll change it to that.

Alizter:
@jdchristensen Thanks for taking a look. I've addressed all your comments.

Alizter:
> Regarding the question of how to define displayed spectra and other things, I wonder whether [arxiv.org/abs/2311.18781](https://arxiv.org/abs/2311.18781) could be relevant? In that paper we formulated a type theory with a generic "display" operation that acts on (among other things) any type defined in the empty context to compute a notion of "displayed elements" of that type.

Yes! I definitely think this is relevant. I hope to sit down at some point and see if I can turn the crank on displayifying some things using the work you and Astra did. It definitely did inspire me to think more in a displayed way.

jdchristensen:
LGTM!  I'll let you merge when you are ready.  BTW, it's better to push an additional commit rather than squashing your changes into one commit, as otherwise the reviewer has no way to see how you addressed the comments.  In this case, it was all minor stuff, so it doesn't matter too much, but I'm just mentioning it for the future.

Alizter:
@jdchristensen sure that makes sense. By the way, you can see the diffenece by clicking on "force push" in the commit history.

PR #1800, opened 2024-01-08
πₙ(Sⁿ) = Z, and other cleanups

jdchristensen:
There are three commits and it's probably best to look at them separately.  The library builds after each commit.

1) Proves that πₙ(Sⁿ) = Z.  It turns out that very little was needed to prove this, since the n = 2 case follows from the `licata_finster` lemma we already had, since S¹ is a coherent H-space.  The higher cases follow from Freudenthal.

2) Improves `pi_prod` in HomotopyGroup.v in a few independent ways.  First, it is now proven for all n, not just n > 0.  Second, the map underlying the equivalence is now the natural one, so there is no need to manually prove that it is a group homomorphism for n > 0.  Third, the proof is much faster.  (The `Defined.` line used to be a bit slow.)

3) This commit contains lots of minor cleanups to things I read while working on the first two.  It also adds some lemmas that I didn't end up using, but which are probably worth having.  But it's definitely better to review this separately, since it touches a lot.

After review, I'll push one more commit that renames Pi1S1.v to PinSn.v, but I've left that out for now to make it easier to review.  That's the only reason this is marked as a draft.


theories/Homotopy/HomotopyGroup.v
```diff
@@ -111,24 +114,7 @@ Proof.
...
-
-(* Every such map is, in particular, a pointed map. *)
-Definition pi_functor_type_pmap {n X Y}
-  : pi_functor_type n X Y -> Pi n X ->* Pi n Y
-  := match n return pi_functor_type n X Y -> (Pi n X ->* Pi n Y) with
-     | 0 => fun f => f
-     (* This works because [pmap_GroupHomomorphism] is already a coercion. *)
-     | n.+1 => fun f => f
-     end.
-Coercion pi_functor_type_pmap : pi_functor_type >-> pForall.
```
jdchristensen, line 129: These lines were removed because this Coercion didn't work.  The problem is that `pi_functor_type` isn't syntactically the same as the output type of `fmap (Pi n) f`, so the coercion wouldn't kick it.  I tried a few things and concluded that it's probably not possible.  That's why I introduced `pPi` as a functor directly landing in pointed types.

jdchristensen: (The coercion *did* work for `fmap (Pi n.+1) f`, but in that case the existing coercion from groups to pointed types also works.)

theories/Pointed/Core.v
```diff
@@ -251,6 +251,19 @@ Proof.
   exact (equiv_path_prod (fst f _, snd f _) (dpoint P, dpoint Q)).
 Defined.
 
+Definition functor_pprod {A A' B B' : pType} (f : A ->* A') (g : B ->* B')
```
Alizter, line 254: We have a notion of bifunctor, but I'm not sure if it would be useful to use here.

Alizter (approved):
Great! I think I've proven Pi_n S^n at least twice before, but I never cleaned it up and PRed it. I believe the thing that annoyed me  was the Pi functor coercion, which it seems you've fixed in a sensible way.

theories/Homotopy/HomotopyGroup.v
Alizter: I think this is more trouble to get working than it is worth in the end. I think if you really want to be precise with this functor in practice, its better to use an explicit version anyway.

Alizter:
@jdchristensen Feel free to push your renaming commit, I'm happy with the changes. It's good to have this result finally. Hopefully Pi3S2 is also doable since we should have most of the tools in place.

jdchristensen:
I added two commits with minor changes, and then a third that renamed Pi1S1.v.

PR #1801, opened 2024-01-08
CayleyDickson: some cleanups, remove funext

Alizter:
I've being playing around with the proofs in Cayley-Dickson and here are some speedups and cleanups.

Notably funext is removed by using the old intro-revert trick when eliminating multiple hyps of a HIT.

### DIFF
--- a/theories/Homotopy/CayleyDickson.v
+++ b/theories/Homotopy/CayleyDickson.v
@@ -307,8 +307,8 @@ Section ImaginaroidHSpace.
     snrapply Join_ind_FFlr.
     1,2: exact (fun _ => ap _ (hspace_left_identity _)).
     intros a b.
-    refine (whiskerR _ _ @ _).
-    { refine (ap _ (ap_idmap _) @ _).
+    lhs nrapply whiskerR.
+    { lhs refine (ap _ (ap_idmap _)).
       exact (Join_rec_beta_jglue
         (fun c => joinl (pt * c))
         (fun d => joinr (conj pt * d))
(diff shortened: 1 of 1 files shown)


jdchristensen (approved):

jdchristensen:
Nice!

theories/Homotopy/CayleyDickson.v
```diff
@@ -301,37 +304,32 @@ Section ImaginaroidHSpace.
...
-      simpl; intro b; apply ap.
-      srapply hspace_left_identity. }
+    snrapply Join_ind_FFlr.
+    1,2: exact (fun _ => ap _ (hspace_left_identity _)).
     intros a b.
-     simpl. apply sq_dp^-1.
-    rewrite ap_idmap, Join_rec_beta_jglue.
-    apply join_natsq_v.
+    refine (whiskerR _ _ @ _).
+    { refine (ap _ (ap_idmap _) @ _).
```
jdchristensen, line 311: These two lines are good candidates for `lhs nrapply whiskerR` and `lhs refine (ap _ (ap_idmap _))`.  There are two near the end of the diff as well.  I especially like the first case because it removes the parentheses and underscores.

jdchristensen: For some reason, `lhs apply foo` doesn't work, so you need to use `nrapply`.  In any case, whether to use the `lhs` tactical is entirely up to you.

Alizter: I tried using it below, but it caused some blow ups with the printing of the goal. I wish I had the patience to investigate why the performance is so poor here.

Alizter:
@jdchristensen I've used lhs a few times, but it didn't interact well in cd_op_right_identity due to the extreme blow up of term size causing coq to lag. The reason is because we are using rewrite in the operation. At some point we should try to do this cleanly as the rewrite terms are really bad for performance when unfolded.

theories/Homotopy/CayleyDickson.v
jdchristensen: No worries, sometimes it doesn't work as well.

PR #1802, opened 2024-01-08
Remove a few "Require" commands that aren't needed

jdchristensen:
Also improves the comments in the coqcstriprequires.py script.


jdchristensen:
This is really minor, so I'm just going to merge it.

PR #1803, opened 2024-01-09
Add pmap_from_psphere_iterated_loops, and needed lemmas

jdchristensen:
Prove that `(psphere n ->** X) <~>* iterated_loops n X` using various minor lemmas spread in a bunch of files.


Alizter (approved):
Nice! Unfortunately it looks like we can't avoid funext due to the equivalence of function spaces. Not that it probably really matters since it is in the equations for the equivalence.

PR #1804, opened 2024-01-09
wildcat: binary products

Alizter:
We define what it means for a wild cat to have binary products and then show Type, pType and Group to have binary products. This notion of product is only 1-coherent as it will not be the correct notion of a product in a bicategory. But this doesn't matter for now as we haven't even defined a full bicategory yet.

AbGroup will be left out for now as a notion of biproduct will follow soon.

I've also defined binary coproducts as products in the opposite category and shown Type, pType and Group as examples of categories with binary coproducts.

### DIFF
--- a/theories/WildCat.v
+++ b/theories/WildCat.v
@@ -13,6 +13,7 @@ Require Export WildCat.PointedCat.
 Require Export WildCat.Bifunctor.
 Require Export WildCat.Monoidal.
 Require Export WildCat.Products.
+Require Export WildCat.Coproducts.
 
 (* See also contrib/SetoidRewrite.v for tools that can be used for rewriting in wild categories. *)
 
--- a/theories/WildCat.v
+++ b/theories/WildCat.v
@@ -12,6 +12,7 @@ Require Export WildCat.Square.
 Require Export WildCat.PointedCat.
 Require Export WildCat.Bifunctor.
 Require Export WildCat.Monoidal.
+Require Export WildCat.Products.
 
 (* See also contrib/SetoidRewrite.v for tools that can be used for rewriting in wild categories. *)
 
--- a/theories/WildCat/Products.v
+++ b/theories/WildCat/Products.v
@@ -171,6 +171,8 @@ Class HasBinaryProducts (A : Type) `{Is1Cat A} := {
 (** *** Product functor *)
 
 (** Binary products are functorial in each argument. *)
+
+(* TODO: jdchristensen: Another approach to handling functoriality on the right is to first prove that products are symmetric. Then functoriality on the right follows from functoriality on the left... *)
 Global Instance is0functor_cat_prod_l {A : Type}
   `{HasBinaryProducts A} y
   : Is0Functor (fun x => cat_prod x y).
--- a/theories/WildCat/Products.v
+++ b/theories/WildCat/Products.v
@@ -3,6 +3,7 @@ Require Import WildCat.Core WildCat.ZeroGroupoid WildCat.Equiv WildCat.Yoneda Wi
 
 (** * Categories with products *)
 
+(* TODO: jdchristensen: There's a more general result, that given any two maps A -> B and A -> C of 0-groupoids, there's a map A -> prop_0gpd B C of 0-groupoids. Then this lemma is that general result applied to is0functor_yon_0gpd of the two projections. *)
 Definition cat_prod_corec_inv {A : Type} `{Is1Cat A}
   (xy x y z : A) (pr1 : xy $-> x) (pr2 : xy $-> y)
   : yon_0gpd xy z $-> prod_0gpd (yon_0gpd x z) (yon_0gpd y z).
@@ -16,7 +17,7 @@ Proof.
     + exact (pr2 $@L p).
 Defined.
 
-(* A binary product of two objects of a category is an object of the category with a give pair of projections together with an induced map being an equivalence. *) 
+(* A binary product of two objects of a category is an object of the category with a pair of projections such that the induced map is an equivalence. *) 
 Class BinaryProduct {A : Type} `{Is1Cat A} {x y : A} := Build_BinaryProduct' {
   cat_prod : A;
   cat_pr1 : cat_prod $-> x;
@@ -28,7 +29,7 @@ Class BinaryProduct {A : Type} `{Is1Cat A} {x y : A} := Build_BinaryProduct' {
 Arguments BinaryProduct {A _ _ _ _} x y.
 Arguments cat_prod {A _ _ _ _} x y {_}.
 
-(** This is a convencience wrapper for building BinaryProducts *)
+(** This is a convenience wrapper for building BinaryProducts *)
 Definition Build_BinaryProduct {A : Type} `{Is1Cat A} {x y : A}
   (cat_prod : A) (cat_pr1 : cat_prod $-> x) (cat_pr2 : cat_prod $-> y)
   (cat_prod_corec : forall z : A, (z $-> x) -> (z $-> y) -> (z $-> cat_prod))
@@ -50,50 +51,47 @@ Proof.
     by apply cat_prod_pr_eta.
 Defined.
 
-(** The product of two objects is unique up to equivalence. *)
-
-
-Definition cat_equiv_cat_prod_corec_inv {A : Type}
-  {x y z : A} `{BinaryProduct _ x y}
-  : (yon_0gpd (cat_prod x y) z) $<~> prod_0gpd (yon_0gpd x z) (yon_0gpd y z).
-Proof.
-  srapply Build_CatEquiv.
-Defined.
-
 Section Lemmata.
 
-  Context {A : Type} {x y z : A} `{BinaryProduct _ x y}.
+  Context {A : Type} {x y : A} `{BinaryProduct _ x y}.
 
-  Definition cat_equiv_cat_prod_corec
+  Definition cate_cat_prod_corec_inv {z : A}
+    : (yon_0gpd (cat_prod x y) z) $<~> prod_0gpd (yon_0gpd x z) (yon_0gpd y z).
+  Proof.
+    srapply Build_CatEquiv.
+  Defined.
+
+  Definition cate_cat_prod_corec {z : A}
     : prod_0gpd (yon_0gpd x z) (yon_0gpd y z) $<~> (yon_0gpd (cat_prod x y) z)
-    := cat_equiv_cat_prod_corec_inv^-1$.
+    := cate_cat_prod_corec_inv^-1$.
 
-  Definition cat_prod_corec : (z $-> x) -> (z $-> y) -> (z $-> cat_prod x y).
+  Definition cat_prod_corec {z : A}
+    : (z $-> x) -> (z $-> y) -> (z $-> cat_prod x y).
   Proof.
     intros f g.
-    apply cat_equiv_cat_prod_corec.
+    apply cate_cat_prod_corec.
     exact (f, g).
   Defined.
     
   (** Applying the first projection after a map pairing gives the first map. *) 
-  Lemma cat_prod_beta_pr1 (f : z $-> x) (g : z $-> y)
+  Lemma cat_prod_beta_pr1 {z : A} (f : z $-> x) (g : z $-> y)
     : cat_pr1 $o cat_prod_corec f g $== f.
   Proof.
-    exact (fst (cate_isretr cat_equiv_cat_prod_corec_inv (f, g))).
+    exact (fst (cate_isretr cate_cat_prod_corec_inv (f, g))).
   Defined.
 
   (** Applying the second projection after a map pairing gives the second map. *)
-  Lemma cat_prod_beta_pr2 (f : z $-> x) (g : z $-> y)
+  Lemma cat_prod_beta_pr2 {z : A} (f : z $-> x) (g : z $-> y)
     : cat_pr2 $o cat_prod_corec f g $== g.
   Proof.
-    exact (snd (cate_isretr cat_equiv_cat_prod_corec_inv (f, g))).
+    exact (snd (cate_isretr cate_cat_prod_corec_inv (f, g))).
   Defined.
 
   (** The pairing map is the unique map that makes the following diagram commute. *)
-  Lemma cat_prod_eta (f : z $-> cat_prod x y)
+  Lemma cat_prod_eta {z : A} (f : z $-> cat_prod x y)
     : cat_prod_corec (cat_pr1 $o f) (cat_pr2 $o f) $== f.
   Proof.
-    exact (cate_issect cat_equiv_cat_prod_corec_inv f).
+    exact (cate_issect cate_cat_prod_corec_inv f).
   Defined.
 
   (* TODO: decompose and move *)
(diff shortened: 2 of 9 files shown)


jdchristensen:
Would it be better to first define what it means for two objects `x` and `y` in a wild category `A` to have a product, and to then say that `A` has products if all pairs of objects have a product?  Then, things could be proved about any particular binary product even if you don't know that the entire category has them.

The definition of products involves a lot of data.  Would it work to instead say that the functor `A^op -> ZeroGroupoid` sending `z` to `(z $-> x) * (z $-> y)` (product of 0-groupoids) is representable?  In other words, there is a product object `x $* y` and a natural equivalence of 0-groupoids `(z $-> x) * (z $-> y) $<~> (z $-> x $* y)`.  This avoids redundancy, and might turn out to be a well-behaved notion.  All of the data you specify would be derivable, and there could be a constructor that takes data like you listed and provides the natural equivalence.  We'd also be able to leverage the work on 0-groupoids and use results like `isequiv_0gpd_issurjinj` to help construct equivalences of 0-groupoids.  This point of view also makes it clear that the product is unique.  But maybe there are issues I'm not seeing?

About naming, should we prefix more of the names with `cat_`?  We could use `cat_prod` instead of `product`, so it is only one character longer.

There are many spelling mistakes in the comments.  Can you double-check them?

Alizter:
> Would it be better to first define what it means for two objects `x` and `y` in a wild category `A` to have a product, and to then say that `A` has products if all pairs of objects have a product? Then, things could be proved about any particular binary product even if you don't know that the entire category has them.

That sounds like a good idea, I'll try to experiment with it. Do you have any categories in mind where binary products aren't available for all objects?
 
> The definition of products involves a lot of data. Would it work to instead say that the functor `A^op -> ZeroGroupoid` sending `z` to `(z $-> x) * (z $-> y)` (product of 0-groupoids) is representable? In other words, there is a product object `x $* y` and a natural equivalence of 0-groupoids `(z $-> x) * (z $-> y) $<~> (z $-> x $* y)`. This avoids redundancy, and might turn out to be a well-behaved notion. All of the data you specify would be derivable, and there could be a constructor that takes data like you listed and provides the natural equivalence. We'd also be able to leverage the work on 0-groupoids and use results like `isequiv_0gpd_issurjinj` to help construct equivalences of 0-groupoids. This point of view also makes it clear that the product is unique. But maybe there are issues I'm not seeing?

Very interesting idea! I'll try it out!

> About naming, should we prefix more of the names with `cat_`? We could use `cat_prod` instead of `product`, so it is only one character longer.

That seems like a sensible idea.
 
> There are many spelling mistakes in the comments. Can you double-check them?

Yes, sorry about that. I need to reenable my spell checker.

jdchristensen:
> Do you have any categories in mind where binary products aren't available for all objects?

I mainly made that suggestion since I thought it would lead to a better factorization of the code, but there are examples.  E.g. in a poset (regarded as a category), products are meets, which sometimes exist and sometimes don't.  In the category of manifolds with boundary, manifolds that happen to not have a boundary have products, but manifolds with non-empty boundary don't.

jdchristensen:
About my second suggestions (using representability), an intermediate amount of data might be better to work with.  Say that a product of `x` and `y` is an object `xy` and projection maps `xy $-> x` and `xy $-> y` such that the induced map `(z $-> xy) -> (z $-> x) * (z $-> y)` is an equivalence of 0-groupoids for each `z`.  The advantage of this is that there is no need to require naturality, since this map will automatically be natural.

Alizter:
This seems to be working but I will need some more time to finish it. Your suggestion of factoring products for a specific pair of objects also led to more composable code. Another class of examples of such categories I thought of is the sum of a category with products to one without. Obviously, only a subclass of the category has products in that case.

I like that we are reflecting limits from groupoids into wild categories. I'm hopeful we can do a similar strategy for more general limits and colimits. That way things like the 3x3 lemma for pushouts could be reduced to the 3x3 lemma for pullbacks in groupoids.

Alizter:
@jdchristensen I've managed to change the definition to the one you suggested. It does in fact contain the exact same data, but it took a while for me to get it working since I had to really understand the terms I was looking at in the proofs. In the end, I managed to make all the lemmas fall out. Rather than forcing users to go through building a groupoid equivalence, I created a nicer constructor for building products that contains the equivalent data.

I've also fixed some typos and naming choices.

Alizter:
I've added a definition of binary coproduct as a product in the opposite category. Many of the properties and lemmas follow definitionally from the product case. The functoriality of the coproduct was so straight forward, but with the correct lemmas about the interaction of functors and opposite categories, should make it possible. I've however left this for future work, as the functoriality of the coproduct is not as useful compared to the product.

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
+Require Import Basics EquivGpd Types.Prod.
+Require Import WildCat.Core WildCat.ZeroGroupoid WildCat.Equiv WildCat.Yoneda WildCat.Universe WildCat.NatTrans WildCat.Opposite.
+
+(** * Categories with products *)
+
+Definition cat_prod_corec_inv {A : Type} `{Is1Cat A}
+  (xy x y z : A) (pr1 : xy $-> x) (pr2 : xy $-> y)
+  : yon_0gpd xy z $-> prod_0gpd (yon_0gpd x z) (yon_0gpd y z).
```
jdchristensen, line 8: There's a more general result, that given any two maps `A -> B` and `A -> C` of 0-groupoids, there's a map `A -> prop_0gpd B C` of 0-groupoids.  Then this lemma is that general result applied to `is0functor_yon_0gpd` of the two projections.

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
...
+  snrapply Build_Morphism_0Gpd.
+  - simpl; intros f.
+    exact (pr1 $o f, pr2 $o f).
+  - snrapply Build_Is0Functor.
+    simpl; intros f g p; split.
+    + exact (pr1 $@L p).
+    + exact (pr2 $@L p).
+Defined.
+
+(* A binary product of two objects of a category is an object of the category with a give pair of projections together with an induced map being an equivalence. *)
```
jdchristensen, line 19:
```suggestion
(* A binary product of two objects of a category is an object of the category with a pair of projections such that the induced map is an equivalence. *) 
```

theories/WildCat/Products.v
jdchristensen: Maybe `xy`, `x` and `y` should be implicit, and `z` should be at the end?

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
...
+    + apply cat_prod_beta_pr1.
+    + apply cat_prod_beta_pr2.
+  - intros f g [p q].
+    by apply cat_prod_pr_eta.
+Defined.
+
+(** The product of two objects is unique up to equivalence. *)
+
+
+Definition cat_equiv_cat_prod_corec_inv {A : Type}
```
jdchristensen, line 56: Should we shorten `cat_equiv` to `cate`, like we do elsewhere?

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
...
+  - intros [f g].
+    exists (cat_prod_corec z f g).
+    split.
+    + apply cat_prod_beta_pr1.
+    + apply cat_prod_beta_pr2.
+  - intros f g [p q].
+    by apply cat_prod_pr_eta.
+Defined.
+
+(** The product of two objects is unique up to equivalence. *)
```
jdchristensen, line 53: I'm not seeing where this is proved.

theories/WildCat/Products.v
jdchristensen: This result could also go in the Section, since it uses exactly the same context.

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
...
+      exact (cat_prod_beta_pr1 _ _)^$.
+    + refine (cat_prod_beta_pr2 _ _ $@ _).
+      refine (_ $@ cat_assoc _ _ _).
+      refine (_ $@ ((cat_prod_beta_pr2 _ _)^$ $@R _)).
+      exact (cat_prod_beta_pr2 _ _)^$.
+Defined.
+
+Global Instance is0functor_cat_prod_r {A : Type}
+  `{HasBinaryProducts A} x
+  : Is0Functor (fun y => cat_prod x y).
```
jdchristensen, line 220: Another approach to handling functoriality on the right is to first prove that products are symmetric.  Then functoriality on the right follows from functoriality on the left...

theories/WildCat/Universe.v
```diff
@@ -1,5 +1,5 @@
 Require Import Basics.Overture Basics.Tactics Basics.Equivalences Basics.PathGroupoids.
-Require Import Types.Equiv.
+Require Import Types.Equiv Types.Prod.
```
jdchristensen, line 2: This change can be reverted, I think.

theories/WildCat/Coproduct.v
```diff
@@ -0,0 +1,129 @@
...
+  exact cat_pr1.
+Defined.
+
+Definition cat_inr {A : Type} `{Is1Cat A} {x y : A} `{!BinaryCoproduct x y} : y $-> cat_coprod x y.
+Proof.
+  change (cat_prod (x : A^op) y $-> y).
+  exact cat_pr2.
+Defined.
+
+Definition Build_BinaryProduct {A : Type} `{Is1Cat A} {x y : A}
```
jdchristensen, line 28: `Co` missing.
```suggestion
Definition Build_BinaryCoproduct {A : Type} `{Is1Cat A} {x y : A}
```

theories/WildCat.v
```diff
@@ -12,6 +12,7 @@ Require Export WildCat.Square.
 Require Export WildCat.PointedCat.
 Require Export WildCat.Bifunctor.
 Require Export WildCat.Monoidal.
+Require Export WildCat.Products.
```
jdchristensen, line 15: And Coproducts.

jdchristensen (approved):
I like how that turned out!

theories/WildCat/Products.v
Alizter: It doesn't quite use the same section in one of the following proofs so I had to generalize it early.

theories/WildCat/Products.v
Alizter: It seems I've accidentally deleted something, let me try and find it.

Alizter:
@jdchristensen I'm proving coproducts in a few categories as we speak, so I'll push those soon and address your review comments.

theories/WildCat/Products.v
Alizter: Actually I think this comment was meant to say the morphsim out of a product is unqiue. I will remove it as it doesn't add anything.

theories/WildCat/Products.v
```diff
@@ -0,0 +1,295 @@
...
+  cat_pr1 : cat_prod $-> x;
+  cat_pr2 : cat_prod $-> y;
+  cat_isequiv_cat_prod_corec_inv
+    :: forall z, CatIsEquiv (cat_prod_corec_inv cat_prod x y z cat_pr1 cat_pr2);  
+}.
+
+Arguments BinaryProduct {A _ _ _ _} x y.
+Arguments cat_prod {A _ _ _ _} x y {_}.
+
+(** This is a convencience wrapper for building BinaryProducts *)
```
jdchristensen, line 31:
```suggestion
(** This is a convenience wrapper for building BinaryProducts *)
```

theories/WildCat/Products.v
Alizter: I'll leave a todo for the general result.

theories/WildCat/Products.v
Alizter: I've updated the names.

theories/WildCat/Coproduct.v
```diff
@@ -0,0 +1,129 @@
...
+      cat_inl
+      cat_inr
+      cat_coprod_rec
+      cat_coprod_beta_inl
+      cat_coprod_beta_inr
+      cat_coprod_in_eta).
+
+Section Lemmata.
+
+  Context {A : Type} {x y z : A} `{BinaryCoproduct _ x y}.
```
jdchristensen, line 47: Maybe `z` should be removed from the `Context`, and put at the start of most of the results below?  This would mean that the previous result would be general enough (I think), and it would also be `z` in what might be a better order in the arguments.

theories/WildCat/Products.v
Alizter: Whoops, fixed.

theories/WildCat/Coproduct.v
Alizter: I'll have a go.

theories/WildCat/Products.v
Alizter: I'll leave a comment about this as I don't think I'll have time to try it out.

Alizter:
@jdchristensen I've shown that Type, pType and Group have coproducts also. Let me know if there are any more categories you can think of.

We probably need to do a little more work before showing that CRing has coproducts due to the need for a tensor product of rings.

AbGroup will follow this later. I plan to define a notion of binary biproduct where a category has biproducts if it has binary product and coproduct with the canonical map between them being an equivalence. I'll need to do some more experimentation on that however.

Alizter:
Once we have biproducts following this PR, we should be able to define semi-additive and additive categories. Together with some more simple (co)limits we can generalize most of our homological algebra work to an arbitrary category.

jdchristensen:
This looks great!

Alizter:
@jdchristensen I'm looking at possibly generalising all of this for arbitrary limits for which I will need to describe limits in ZeroGroupoid. I've done this by defining limits as the hom from a constant diagram to a diagram in the presheaf category.

It seems at some point I have to require the 1-functoriality of the `carrier` map of a ZeroGroupoid. Do you think this is true? I'm having trouble defining how it should act on 2-cells. I feel like we are missing this level of coherence.

If this is not possible, then I will need to rethink how to define a general notion of limit.

jdchristensen:
> It seems at some point I have to require the 1-functoriality of the `carrier` map of a ZeroGroupoid. Do you think this is true? I'm having trouble defining how it should act on 2-cells. I feel like we are missing this level of coherence.

No, that's not true.  If G is the 0-groupoid with underlying type Unit and one (identity) morphism, H is the 0-groupoid with underlying type Bool with morphisms in both directions between true and false, then there is a 2-cell between the into inclusion functors G → H (as maps of 0-groupoids), but there is no 2-cell in Type between those maps.

I haven't thought about why this would be needed to define limits.  But I did think about defining an equalizer of two parallel functors f, g : G → H of 0-groupoids, and when defining the 1-cells in the equalizer, there is a 2-cell in H that needs filling.  It could be done using the identity type, but that feels a bit unnatural.  So for general limits, we might need to assume that our ambient wild category has enough higher cells for us to construct the appropriate limit of the Yoneda functors.  Not sure.

Things are easier for products since there are no morphisms in the diagram.

Alizter:
Some other ideas I've been floating around: We might be able to define a displayed ZeroGroupoid and maybe replicate the "forall" construction from pType. I feel like this would be too good to be true however.

Alizter:
Merging this now, I'll play around some more with general limits / colimits. If we manage to get something working, it could be very useful in replacing all our limit / colimit theories for Type with something more generic and useful.

Alizter:
@jdchristensen The reason you might need `carrier` to be a 1-functor is that a diagram of groupoids is `D^op -> ZeroGroupoid`, however a type based limit takes functors of the form `D^op -> Type`. The underlying assumption is that carrier is not only a functor, but also limit preserving, which I would believe to be true.

PR #1805, opened 2024-01-09
Reduce universe variables and other clean-ups

jdchristensen:
When dealing with types defined to land in `Set` (aka `Type0`), such as `abgroup_Z`, `Pos`, `Int` and `Fin n`, it makes sense to use `Local Set Universe Minimization ToSet.` as otherwise free universe variables are created that are not matched when using the results.  By doing this and being careful to use `simple_induction`, many universe variables are eliminated.  For example, `abgroup_Z` goes from 25 to 0, `fsucc_mod` goes from 9 to 0, and `stratified_succ` goes from 17 to 1.

There are other improvements mixed in, such as making use of `nat_iter`.   I also cleaned up various comments and fixed indentation.

### DIFF
--- a/theories/Spaces/Int/LoopExp.v
+++ b/theories/Spaces/Int/LoopExp.v
@@ -26,6 +26,8 @@ Definition loopexp {A : Type} {x : A} (p : x = x) (z : Int) : (x = x)
        | pos n => loopexp_pos p n
      end.
 
+(** TODO: One can also define [loopexp] as [int_iter (equiv_concat_r p x) z idpath].  This has slightly different computational behaviour, e.g., it sends [1 : int] to [1 @ p] rather than [p].  But with this definition, some of the results below become special cases of results in Int.Equiv, and others could be generalized to results belonging in Int.Equiv.  It's probably worth investigating this. *)
+
 Lemma loopexp_pos_inv {A : Type} {x : A} (p : x = x) (n : Pos)
   : loopexp_pos p^ n = (loopexp_pos p n)^.
 Proof.
@@ -63,28 +65,6 @@ Proof.
     apply ap.
     apply ap_loopexp_pos.
   + apply ap_loopexp_pos.
-Qed.
-
-Lemma int_add_succ_l a b : int_succ a + b = int_succ (a + b).
-Proof.
-  rewrite <- int_add_assoc, (int_add_comm 1 b).
-  apply int_add_assoc.
-Qed.
-
-Lemma int_add_succ_r a b : a + int_succ b = int_succ (a + b).
-Proof.
-  apply int_add_assoc.
-Qed.
-
-Lemma int_add_pred_l a b : int_pred a + b = int_pred (a + b).
-Proof.
-  rewrite <- int_add_assoc, (int_add_comm (-1) b).
-  apply int_add_assoc.
-Qed.
-
-Lemma int_add_pred_r a b : a + int_pred b = int_pred (a + b).
-Proof.
-  apply int_add_assoc.
 Qed.
 
 Lemma loopexp_pos_concat {A : Type} {x : A} (p : x = x) (a : Pos)
--- a/theories/Algebra/AbGroups/Z.v
+++ b/theories/Algebra/AbGroups/Z.v
@@ -8,7 +8,11 @@ Require Import Algebra.AbGroups.AbelianGroup.
 
 Local Open Scope int_scope.
 
-Definition abgroup_Z : AbGroup.
+Section MinimizationToSet.
+
+Local Set Universe Minimization ToSet.
+
+Definition abgroup_Z@{} : AbGroup@{Set}.
 Proof.
   snrapply Build_AbGroup.
   - refine (Build_Group Int int_add 0 int_negation _); repeat split.
@@ -19,6 +23,8 @@ Proof.
     + exact int_add_negation_r.
   - exact int_add_comm.
 Defined.
+
+End MinimizationToSet.
 
 (** We can multiply by [n : Int] in any abelian group. *)
 Definition ab_mul (n : Int) {A : AbGroup} : GroupHomomorphism A A.
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -252,6 +252,7 @@ Definition pequiv_groupisomorphism {A B : Group}
   := fun f => Build_pEquiv _ _ f _.
 
 Coercion equiv_groupisomorphism : GroupIsomorphism >-> Equiv.
+Coercion pequiv_groupisomorphism : GroupIsomorphism >-> pEquiv.
 
 Definition equiv_path_groupisomorphism `{F : Funext} {G H : Group}
   (f g : GroupIsomorphism G H)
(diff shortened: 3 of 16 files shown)


theories/Spaces/Int/LoopExp.v
```diff
@@ -65,28 +67,6 @@ Proof.
...
-Lemma int_add_pred_l a b : int_pred a + b = int_pred (a + b).
-Proof.
-  rewrite <- int_add_assoc, (int_add_comm (-1) b).
-  apply int_add_assoc.
-Qed.
-
-Lemma int_add_pred_r a b : a + int_pred b = int_pred (a + b).
-Proof.
-  apply int_add_assoc.
-Qed.
```
jdchristensen, line 88: The section above was just moved to Int/Spec.v, where it fits better.

Alizter (approved):
Nice! Did we ever experiment with using minimization to set globally?

SkySkimmer:
We would probably want to change the argument order so that the type argument can be a parameter. Otherwise its universe has more restricted cumulativity AFAICT.

mikeshulman:
Ugh.  Maybe we could make a notation that puts the arguments back in the right order.

It would be nice if there were a way to indicate which arguments are indices and which are parameters that didn't require all the parameters to come before all the indices.

jdchristensen:
@SkySkimmer Is there any chance that in the future Coq could have cumulative fixpoints?

SkySkimmer:
I haven't seen any proposed rules of how that would work.

jdchristensen:
I tried the inductive definition of `IsTrunc`, and Coq accepts it in the modified form below, but it seems quite awkward to work with.
```coq
Cumulative Inductive IsTrunc_internal (A : Type@{u}) : trunc_index -> Type@{u} :=
| istrunc_iscontr : Contr_internal A -> IsTrunc_internal A minus_two
| istrunc_succ : forall {n:trunc_index}, (forall x y:A, IsTrunc_internal (x = y) n) -> IsTrunc_internal A (trunc_S n).

Existing Class IsTrunc_internal.

Notation IsTrunc n A := (IsTrunc_internal A n).
```
It's not just that you need to apply constructors, but that you need to go backwards.  For example, we define `Contr A` to be `IsTrunc (-2) A`, but then one needs to prove that `Contr A` is equivalent to `Contr_internal A`.  And when given a hypothesis of the form `IsTrunc n.+1 A`, when you do induction on it, you need to handle the `istrunc_iscontr` constructor somehow.  The following helps:
```coq
Scheme IsTrunc_internal_ind := Induction for IsTrunc_internal Sort Type.
Scheme IsTrunc_internal_rec := Minimality for IsTrunc_internal Sort Type.
Definition IsTrunc_internal_rect := IsTrunc_internal_ind.

Definition IsTrunc_unfolded (n : trunc_index) (A : Type)
  := match n with
    | minus_two => Contr_internal A
    | n.+1 => forall x y : A, IsTrunc n (x = y)
    end.

Definition istrunc_unfold (n : trunc_index) (A : Type)
  : IsTrunc n A -> IsTrunc_unfolded n A.
Proof.
  intros [istr|k istr]; exact istr.
Defined.

Definition isequiv_istrunc_unfold (n : trunc_index) (A : Type)
  : IsEquiv (istrunc_unfold n A).
Proof.
  simple refine (Build_IsEquiv _ _ (istrunc_unfold n A) _ _ _ _).
  - destruct n; constructor; assumption.
  - destruct n; reflexivity.
  - intros [istr|k istr]; reflexivity.
  - intros [istr|k istr]; reflexivity.
Defined.

Definition equiv_istrunc_unfold (n : trunc_index) (A : Type)
  := Build_Equiv _ _ _  (isequiv_istrunc_unfold n A).
```
I also put primes on the constructors of `Contr_internal` and did:
```coq
Definition center (A : Type) {H : Contr A} : A := @center' A (istrunc_unfold _ _ H).
Definition contr {A : Type} {H : Contr A} (y : A) : center A = y := @contr' A (istrunc_unfold _ _ H) y.
```
But even with these, it's annoying to work with.  E.g. the last result in Contractible.v needs to use that `ap` of an equivalence is an equivalence, but that currently comes later.  It wouldn't be hard to flip the dependency between Contractible.v and Equivalences.v to handle this, but I suspect that we'd continue to find it annoying to work with.

mikeshulman:
We could also rid of `Contr_internal` and incorporate it in the definition of `IsTrunc`:

```coq
Cumulative Inductive IsTrunc_internal (A : Type@{u}) : trunc_index -> Type@{u} :=
| istrunc_iscontr : (center : A) -> (forall y : A, center = y) -> IsTrunc_internal A minus_two
| istrunc_succ : forall {n:trunc_index}, (forall x y:A, IsTrunc_internal (x = y) n) -> IsTrunc_internal A (trunc_S n).
```

Then there would only need to be one definition of `center` and `contr`.

> For example, we define `Contr A` to be `IsTrunc (-2) A`, but then one needs to prove that `Contr A` is equivalent to `Contr_internal A`.  And when given a hypothesis of the form `IsTrunc n.+1 A`, when you do induction on it, you need to handle the `istrunc_iscontr` constructor somehow.

Sure, but don't we only need to do that once?  Once we've proven the relevant equivalences, we can just apply those lemmas and tactics.

In general, I've come more and more to believe in the principle that whenever possible types should be canonical, not things that compute definitionally, because this makes it easier to match against them.  For instance, mightn't an inductive `IsTrunc` simplify some of the problems with typeclass inference?  We have the comment that

```
We prefer to reason about [IsTrunc (S n) A] rather than [IsTrunc n (@paths A a b)]
```

but if `IsTrunc` is canonical there would be no difficulty with this: `IsTrunc (S n) A` doesn't reduce to anything, we just have to apply a tactic or lemma to use it (e.g. `istrunc_intro x y` rather than `intros x y`).

jdchristensen:
@mikeshulman You're probably right that with appropriate tactics, using the inductive definition would be fairly painless.  But it would probably require many small changes throughout the entire library to do the conversion, so it might be a bit of work.

Another observation is that with my sample code above, we essentially have *both* definitions.  `IsTrunc` is the inductive one, and `IsTrunc_unfolded` is essentially the fixpoint one, except that it goes via `IsTrunc` in the inductive step.  Theorems would be stated using `IsTrunc` and proved using `IsTrunc_unfolded`, and we'd be frequently converting back and forth.  Except for the benefit of cumulativity, it would seem cleaner to just use one definition.  One of the things I've learned is that if you can get things to work out definitionally, it really simplifies things in many cases!

mikeshulman:
It seems to me that the only time we'd actually need to transfer explicitly across that equivalence would be in proving that `IsTrunc` is a proposition (and we might be able to just do that by hand without introducing `IsTrunc_internal` at all).  Once we know that, it seems to me that in further proofs we'd only need to use the constructors of `IsTrunc` and their inverses.

Alizter:
@jdchristensen Are you happy with this being merged.

jdchristensen:
No, let's not merge this yet.  If the WIP on making IsTrunc inductive pans out, then the last commit here will need to be undone.  Also, I plan to see if I can make similar improvements to Nat.

Alizter:
I wonder if we can shadow `induction` everywhere and define it as `simple_induction`.

Alizter (approved):
LGTM. Thank you very much for all this careful work!

PR #1806, opened 2024-01-10
Make IsTrunc an inductive type

jdchristensen:
This is based on discussion in PR #1805 .  With the changes shown here, around 225 of the 550 files in the library build.  Every file changed in this PR builds, including all of Basics, Types, HProp.v, HSet.v, TruncType.v, etc.  I haven't got to Truncations/* or Modalities/* yet, but I feel like at this point all of the non-trivial changes are done and it will just be a matter of inserting the constructors of `IsTrunc` and their inverses (via `equiv_istrunc_unfold` and/or `istrunc_paths`) in many places.  It's a little annoying having to do this, so that's why I stopped at this point to see what people think.  Also, there might be things I should change in the foundations, and it would be good to hear such suggestions before doing the rest of the work.

Two benefits I see:  `IsTrunc` is now cumulative, and several things involving typeclasses that cluttered up Overture.v can simply be deleted.

The downsides:  having to convert frequently.

I'm on the fence about whether this is a win, so I'd be particularly interested in feedback on this.

### DIFF
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -17,7 +17,7 @@ Definition path_contr `{Contr A} (x y : A) : x = y
 (** Any space of paths in a contractible space is contractible. *)
 Global Instance contr_paths_contr `{Contr A} (x y : A) : Contr (x = y) | 10000.
 Proof.
-  exists (path_contr x y).
+  apply (Build_Contr _ (path_contr x y)).
   intro r; destruct r; apply concat_Vp.
 Defined.
 
@@ -37,7 +37,7 @@ Arguments path_basedpaths {X x y} p : simpl nomatch.
 
 Global Instance contr_basedpaths {X : Type} (x : X) : Contr {y : X & x = y} | 100.
 Proof.
-  exists (x;1).
+  apply (Build_Contr _ (x;1)).
   intros [y p]; apply path_basedpaths.
 Defined.
 
@@ -105,17 +105,6 @@ Definition contr_retract {X Y : Type} `{Contr X}
 Definition contr_change_center {A : Type} (a : A) `{Contr A}
   : Contr A.
 Proof.
-  exists a.
+  apply (Build_Contr _ a).
   intros; apply path_contr.
-Defined.
-
-(** If a type is contractible, then so is its type of contractions. *)
-Global Instance contr_contr `{Funext} (A : Type) `{contrA : Contr A}
-  : Contr (Contr A) | 100.
-Proof.
-  exists contrA; intros [a2 c2].
-  destruct (contr a2).
-  apply (ap (Build_Contr _ (center A))).
-  apply path_forall; intros x.
-  apply path2_contr.
 Defined.
--- a/theories/Basics/Contractible.v
+++ b/theories/Basics/Contractible.v
@@ -108,3 +108,18 @@ Proof.
   apply (Build_Contr _ a).
   intros; apply path_contr.
 Defined.
+
+(** The automatically generated induction principle for [IsTrunc_internal] produces two goals, so we define a custom induction principle for [Contr] that only produces the expected goal. *)
+Definition Contr_ind@{u v|} (A : Type@{u}) (P : Contr A -> Type@{v})
+  (H : forall (center : A) (contr : forall y, center = y), P (Build_Contr A center contr))
+  (C : Contr A)
+  : P C
+  := match C as C0 in IsTrunc n _ return
+          (match n as n0 return IsTrunc n0 _ -> Type@{v} with
+           | minus_two => fun c0 => P c0
+           | trunc_S k => fun _ => Unit
+           end C0)
+    with
+    | Build_Contr center contr => H center contr
+    | istrunc_S _ _ => tt
+    end.
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -499,12 +499,7 @@ In order to achieve moderate coverage and speedy resolution, we currently follow
 (** A space [A] is contractible if there is a point [x : A] and a (pointwise) homotopy connecting the identity on [A] to the constant map at [x].  Thus an element of [contr A] is a pair whose first component is a point [x] and the second component is a pointwise retraction of [A] to [x]. *)
 
 (** We use the [Contr_internal] record so as not to pollute typeclass search; we only do truncation typeclass search on the [IsTrunc] datatype, usually.  We will define a notation [Contr] which is equivalent to [Contr_internal], but picked up by typeclass search.  However, we must make [Contr_internal] a class so that we pick up typeclasses on [center] and [contr].  However, the only typeclass rule we register is the one that turns it into a [Contr]/[IsTrunc].  Unfortunately, this means that declaring an instance like [Instance contr_foo : Contr foo := { center := bar }.] will fail with mismatched instances/contexts.  Instead, we must iota expand such definitions to get around Coq's deficiencies, and write [Instance contr_foo : Contr foo := let x := {| center := bar |} in x.] *)
-Class Contr_internal (A : Type) := Build_Contr {
-  center : A ;
-  contr : (forall y : A, center = y)
-}.
-
-Arguments center A {_}.
+(** TODO: update comment. *)
 
 (** *** Truncation levels *)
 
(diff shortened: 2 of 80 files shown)


theories/Basics/Contractible.v
```diff
@@ -105,17 +105,6 @@ Definition contr_retract {X Y : Type} `{Contr X}
...
-(** If a type is contractible, then so is its type of contractions. *)
-Global Instance contr_contr `{Funext} (A : Type) `{contrA : Contr A}
-  : Contr (Contr A) | 100.
-Proof.
-  exists contrA; intros [a2 c2].
-  destruct (contr a2).
-  apply (ap (Build_Contr _ (center A))).
-  apply path_forall; intros x.
-  apply path2_contr.
-Defined.
```
jdchristensen, line 121: This was moved to Trunc.v and generalized to `contr_istrunc`.

theories/Basics/Overture.v
```diff
@@ -541,58 +536,60 @@ Local Open Scope trunc_scope.
...
+  : IsTrunc n A -> IsTrunc_unfolded n A.
+Proof.
+  intros [center contr|k istrunc].
+  - exact (center; contr).
+  - exact istrunc.
+Defined.
 
-Existing Class IsTrunc_internal.
+Definition isequiv_istrunc_unfold (n : trunc_index) (A : Type)
+  : IsEquiv (istrunc_unfold n A).
```
jdchristensen, line 568: This could be moved to Trunc.v, since that's the only place it is used.

theories/Basics/Tactics.v
```diff
@@ -733,12 +733,6 @@ Ltac issig :=
 
 (** We show how the tactic works in a couple of examples. *)
 
-Definition issig_contr (A : Type)
-  : {x : A & forall y, x = y} <~> Contr A.
-Proof.
-  issig.
-Defined.
```
jdchristensen, line 740: This example no longer works, and since it was never used, I removed it.

theories/Basics/Trunc.v
```diff
@@ -435,29 +445,31 @@ Proof.
   - rapply contr_inhabited_hprop.
 Defined.
 
-Corollary equiv_contr_hprop (A : Type) `{Funext} `{IsHProp A}
-  : Contr A <~> A.
-Proof.
-  exact (equiv_iff_hprop_uncurried (iff_contr_hprop A)).
-Defined.
```
jdchristensen, line 442: Moved to the end of the file.

theories/Basics/Overture.v
jdchristensen: (It's now also used in ReflectiveSubuniverse.  It could still move to Trunc.v, but maybe it's best in Overture.v.)

jdchristensen:
I added a coercion from `IsTrunc` to `Funclass`.  Now, when using a proof of `IsTrunc`, you can treat it like a function, which should slightly reduce the number of places to be fixed.  But it's much more common to need to *produce* an element of `IsTrunc`, and for then you still need to use one of the constructors.  I don't know if there's any way to change that.

I had to do something with [n = -2], so I produce the contracting homotopy.

This coercion doesn't satisfy the uniform inheritance property, but it does work in practice.

Question:  before this PR, given `C : Contr A` and a goal of `A`, `apply C` would work, with Coq inserting the call to `center`.  I can't figure out why that worked before, but it no longer works now.  Is there a way to make it work again?

jdchristensen:
The second new commit finishes off Modalities/*, Truncations/* and HIT/*, getting us more than half-way.

theories/Basics/Trunc.v
```diff
@@ -435,29 +445,42 @@ Proof.
   - rapply contr_inhabited_hprop.
 Defined.
 
-Corollary equiv_contr_hprop (A : Type) `{Funext} `{IsHProp A}
-  : Contr A <~> A.
```
jdchristensen, line 439: Moved to end of file.

theories/Basics/Trunc.v
```diff
@@ -435,29 +445,42 @@ Proof.
...
+Definition contr_forall `{Funext} `{P : A -> Type} `{forall a, Contr (P a)}
+  : Contr (forall a, P a).
 Proof.
-  exact (equiv_iff_hprop_uncurried (iff_contr_hprop A)).
+  apply (Build_Contr _ (fun a => center (P a))).
+  intro f.  apply path_forall.  intro a.  apply contr.
+Defined.
+
+Global Instance istrunc_forall `{Funext} `{P : A -> Type} `{forall a, IsTrunc n (P a)}
+  : IsTrunc n (forall a, P a) | 100.
```
jdchristensen, line 458: These two moved from Types/Forall, since they are useful here.  Also, the first one `contr_forall` is no longer an instance, since it was redundant.

theories/Types/Unit.v
```diff
@@ -4,6 +4,8 @@
 Require Import Basics.Overture Basics.Equivalences.
 Local Open Scope path_scope.
 
+Local Set Universe Minimization ToSet.
```
jdchristensen, line 7: This was done because otherwise the universe variable in `contr_unit` was free (since `IsTrunc` is now cumulative), which caused a problem in another file (Nullification.v, I think).

Alizter:
I did something similar with the definition of leq for nat and looking back I don't know if it was a good idea. Using inductive types can make the proofs more rigid to reduction, but also means we lose out on a lot of easy simplificiations.

However I suppose that a lesson to be learnt is that proof by reduction can be expensive, in that sense it is cheaper for the proof assistant to have an explicit sequence of terms. It also saves us from having to recompute the same kinds of things repeatedly.

The benefit of this approach is cumulativity and I think that is a good gain, since unnecessary universes cause a lot of performance issues for us.

Also, since being truncated is a hprop, we really shouldn't concern ourselves too much with how the proofs look. If we ever for some reason need to reason about a proof, we can always equate it to a nicer one. For this reason, having something that gains us efficiency is desirable.

Another benefit of this approach is that it will work better with the typeclasses mechanism. It is a huge burden to have to do reduction when resolving typeclasses.

These are my thoughts on this issue, I am therefore happy to proceed with this.

theories/Basics/Trunc.v
```diff
@@ -321,15 +329,15 @@ Definition istrunc_hset {n} {A} `{IsHSet A}
 
 (** Equivalence preserves truncation (this is, of course, trivial with univalence).  This is not an [Instance] because it causes infinite loops. *)
 Definition istrunc_isequiv_istrunc A {B} (f : A -> B)
-  `{IsTrunc n A} `{IsEquiv A B f}
+  `{H : IsTrunc n A} `{IsEquiv A B f}
```
mikeshulman, line 332: Is this name `H` used in the new proof?

theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -1049,8 +1049,7 @@ Section Reflective_Subuniverse.
       generalize dependent A; simple_induction n n IH; intros A ?.
       - (** We have to be slightly clever here: the actual definition of [Contr] involves a sigma, which [O] is not generally closed under, but fortunately we have [equiv_contr_inhabited_allpath]. *)
         refine (inO_equiv_inO _ equiv_contr_inhabited_allpath^-1).
-      - change (In O (forall x y:A, IsTrunc n (x=y))).
-        exact _.
+      - refine (inO_equiv_inO _ (equiv_istrunc_unfold n.+1 A)^-1).
```
mikeshulman, line 1052: Haha!  I totally didn't expect that `equiv_istrunc_unfold` would have any more use after `ishprop_istrunc`, but I was wrong!

theories/TruncType.v
```diff
@@ -106,7 +107,7 @@ Section TruncType.
 
   Global Instance isset_HProp : IsHSet HProp := _.
 
-  Global Instance istrunc_sig_istrunc : forall n, IsTrunc n.+1 (sig (IsTrunc n)) | 0.
+  Global Instance istrunc_sig_istrunc : forall n, IsTrunc n.+1 (sig (fun A => IsTrunc n A)) | 0.
```
mikeshulman, line 110: As long as the eta-expansion is necessary, wouldn't it be even clearer to write `{A:Type & IsTrunc n A}` rather than `sig (fun A => IsTrunc n A)`?

theories/Categories/Category/Sum.v
```diff
@@ -1,5 +1,6 @@
 (** * The coproduct of categories *)
 Require Export Category.Core.
+Require Import Types.Empty.
```
mikeshulman, line 3: What is this for?

mikeshulman (commented):
Thanks for giving this a try!  I guess I'm on the fence too.

I feel like a smarter `exists` tactic ought to be able to notice that `Contr A` can only be built by one constructor, and if we could "coerce from Funclass to something else" then `intros` ought to work for proving other `IsTrunc`s, but that's just wishing.

Do we have any sense for how much mileage we get out of cumulativity?

theories/Basics/Trunc.v
jdchristensen: It was during an intermediate version, but isn't now, so I'll remove it here.

theories/TruncType.v
jdchristensen: Right, that's better.  BTW, things like `IsTrunc` and `IsHProp` no longer work as type families without the eta-expansion.  I tried defining `IsHProp` to be `fun A => IsTrunc (-1) A`, but that caused `IsHProp` to no longer be regarded as part of the `IsTrunc_internal` typeclass.

theories/Categories/Category/Sum.v
jdchristensen: The proof that `sum` is a `PreCategory` produces some goals of the form `IsHSet Empty`.  I guess that when this computed to something of the form `forall x y : Empty, ...`, Coq was able to prove it using `intros []`.  I can avoid requiring `Types.Empty` by adding `apply istrunc_S` in the `repeat` command, and that's probably better, so I'll switch it.

jdchristensen:
With the latest commit, the library builds.  I did have to use "Admitted" for two results where universe variables are causing issues, and I didn't have time to investigate.  One is `hartogs_number_no_injection` in Sets/Hartogs.v, and the other is `Book_4_6_iii` in HoTTBookExercises.v.  Both probably just need a few universe annotations.  If anyone wants to help, that would be great.  I'm tagging @dominik-kirst, who may be able to figure out what universes things should live in in that result.

jdchristensen:
> Do we have any sense for how much mileage we get out of cumulativity?

I'm not sure yet.  So far, it has caused problems in two places, but there's a chance that if we now tighten up some universe variables in earlier proofs, those problems would go away.

I do like that Overture.v has gotten simpler, without those various typeclass hints.  And most of the changes to the rest of the library were just adding constructors, which isn't a big deal.

I just tried setting minimization to set globally, and around 300 files build, with just 6 failures, which is a bit better than before this PR.  I haven't looked into how hard it would be to fix the errors.

jdchristensen:
I did some timing tests, and this PR speeds things up by about 6% for both -j1 and -j8 builds.  And that's without any tuning for the new definition of `IsTrunc`, so I would guess that there is some low hanging fruit that would speed things up even more.  It's also without reducing universe variables as should be possible with this PR.  This speedup, combined with the cleanups to Overture.v, make me think it's best to go ahead with this PR, assuming that the two glitches with universe variables are easy to fix.

This PR improved a few proofs while repairing them, and I checked that those files didn't contribute significantly to the speedup.  In fact, the files that improved the most were generally files that were not changed at all, or which had very minor changes.  For example, the build time of Analysis/Locator.v is 49% of what it was before, even though it was not changed by this PR.

mikeshulman:
That seems like a good argument!

theories/Basics/Overture.v
```diff
@@ -541,58 +536,84 @@ Local Open Scope trunc_scope.
...
+| istrunc_S : forall {n:trunc_index}, (forall x y:A, IsTrunc_internal (x = y) n) -> IsTrunc_internal A (trunc_S n).
+
+Existing Class IsTrunc_internal.
 
-Fixpoint IsTrunc_internal (n : trunc_index) (A : Type@{u}) : Type@{u} :=
-  match n with
-    | minus_two => Contr_internal A
-    | n'.+1 => forall (x y : A), IsTrunc_internal n' (x = y)
-  end.
+Notation IsTrunc n A := (IsTrunc_internal A n).
```
JasonGross, line 547: I think if you use non-uniform inductive parameters you can do away with this notation and do 
```
Cumulative Inductive IsTrunc (n : trunc_index) (A : Type@{u}) Type@{u} :=
| Build_Contr : forall (center : A) (contr : forall y, center = y), IsTrunc_internal A minus_two
| istrunc_S : forall {n:trunc_index}, (forall x y:A, IsTrunc_internal (x = y) n) -> IsTrunc_internal A (trunc_S n).
```
I'm not completely sure though

Alizter: I don't think this is possible since under the hood Coq will not treat trunc_index as an index (non-uniform parameter) and therefore it won't be allowed to vary the type of the constructors.

This is the same for the Vector type in Coq which has to take the type first and then the nat.

jdchristensen: I fixed a few typos, giving
```coq
Cumulative Inductive IsTrunc (n : trunc_index) (A : Type@{u}) : Type@{u} :=
| Build_Contr : forall (center : A) (contr : forall y, center = y), IsTrunc minus_two A
| istrunc_S : forall {n:trunc_index}, (forall x y:A, IsTrunc n (x = y)) -> IsTrunc (trunc_S n) A.
```
but get `Unable to unify "IsTrunc minus_two A" with "IsTrunc n A".`

JasonGross: Ah, too bad

jdchristensen:
Rebased.  All that is left is the universe issue in Hartogs.

jdchristensen:
Ok, just a single universe annotation was needed to fix the last issue.  @dominik-kirst , if you have a chance, could you confirm that when defining `hartogs_number'`, it is reasonable to assume that the set `B` is in the same universe as `A`?
```coq
  Definition hartogs_number' : Ordinal.
  Proof.
    set (carrier := {B : Ordinal@{A _} & card B <= card A}).
```

jdchristensen:
With the inductive definition of `IsTrunc_internal`, where `Contr A` is defined to be `IsTrunc_internal A (-2)`, in this situation
```coq
Definition test {A : Type} (H : Contr A) (P : Contr A -> Type) : P H.
  destruct H.
```
Coq generates two goals, one for each constructor of `IsTrunc_internal`, even though only the first constructor makes sense:
```coq
 P (Build_Contr A center contr)
 P (istrunc_S A i)
```
This causes problems for both the `issig` tactic and the `make_equiv` tactic if there are `Contr` terms around.  Is there a way to make Coq smarter about this?  (It might also be nice if it was smarter in the successor case as well.)

Note that when there is a *goal* of the form `Contr A`, `econstructor` knows that it has to apply the first constructor.  But our custom `ntc_constructor` gets confused because it uses `destruct` internally.  (`make_equiv` uses `elim` internally, giving the same problem.)

mikeshulman:
Could we just write a replacement for `destruct` that does the right thing?

jdchristensen:
> Could we just write a replacement for `destruct` that does the right thing?

`destruct`, `elim` and `induction` all face this issue (and maybe others).  Is there one thing we can change that will fix all of them?

 Also, when we are trying to prove something of the form `forall (n : trunc_index) (t : IsTrunc_internal A n),  P n t` (such as `isequiv_istrunc_unfold`) then we do want the existing induction principle.  It's not clear to me how the tactic will know which we want.

I feel like there must be a standard solution to this, since it also comes up with things like vectors.

Alizter:
@jdchristensen Vectors are unfortunately very poorly behaved and hardly used in Coq. Mainly for these reasons. FTR induction is capable of using different induction principles.

Alizter:
cc @SkySkimmer @JasonGross Any ideas how to fix pattern matching on indexed inductives with destruct?

mikeshulman:
Well, one thing that could in principle be done is to convert the index to a non-uniform parameter.  That would also fix the ordering issue without an extra notation.
```coq
Inductive IsTrunc (n : trunc_index) (A : Type@{u}) : Type@{u} :=
| Build_Contr : forall (center : A) (contr : forall y, center = y), (minus_two = n) -> IsTrunc n A
| istrunc_S : forall {k:trunc_index}, (forall x y:A, IsTrunc k (x = y)) -> (trunc_S k = n) -> IsTrunc n A.
```
Then `destruct` would still produce two goals in the objectionable cases, but one of them would be trivial with `discriminate`.

SkySkimmer:
You could use inversion instead of destruct if you don't care about the term it generates. If you do care I'm not sure what you expect destruct to do.
I don't think there's any point using induction pn IsTrunc with non-variable index.
I don't know what elim does that destruct doesn't.

jdchristensen:
I tried the version of `IsTrunc` with `n` as a parameter, and having to deal with the equalities is awkward.  Moreover, `destruct` still generates two goals, and my hope is to make things like `issig` and `make_equiv` work automatically, so this change wouldn't help with that.

I tried `inversion`, and maybe it's just user error, but it still generated two goals, and in neither case was the term `H` in the goal destructed.

I also tried replacing `elim` with `destruct` in `decomposing_intros` (part of `make_equiv`), and `make_equiv` no longer works.  So I guess we'd need custom versions of both tactics if we went that route.

For now, I've pushed a change that at least makes `make_equiv` work.  (`issig` is a touch harder, and we can fix it later if it is needed.)  `Contr_ind` is located in Equivalences.v since it uses `equiv_ind`, and since that is where `make_equiv` is.  The `match` to determine the family `P` greatly speeds this up.  Is this a reasonable compromise for now?  While I'd like to have a more general solution, it's not like we'll be doing induction on elements of `IsTrunc` very often.

Alizter:
@jdchristensen You might need dependent inversion.

jdchristensen:
> @jdchristensen You might need dependent inversion. Mike's idea is basically inversion but the tactic goes a step further and tries to eliminate the index equalities.

I had tried `dependent inversion` and it didn't work either.  Anyways, I don't think we want ugly proof terms.

As an aside, I just replaced the `equiv_ind` proof of `Contr_ind` with a direct proof using `match`, since it seems better to prove something that fundamental with a proof that is closer to the metal.

I'm happy to merge this as is, unless anyone wants to look at it further.

Alizter (approved):
LGTM, feel free to merge.

mikeshulman:
Sure, that seems fine to me for now.

PR #1807, opened 2024-01-12
remove workaround for coq/coq#8994

### LINKED ISSUE #1541: Remove workarounds for coq/coq#8994
coq/coq#12975 / coq/coq#9711 has now been fixed, so we can finally remove workarounds. I am only aware of https://github.com/HoTT/HoTT/blob/d49e8b11e212b188e9d3d49115ddc8aef8f351e8/theories/WildCat/Equiv.v#L46-L48
at the moment, so it would be good to note any others here.

Once we update to 8.14 we will no longer need to do the workarounds.

Alizter:
- fixes #1541


jdchristensen:
What about the other fields that still have primed versions?

Alizter:
Actually that's a good point. I've only addressed one. I'll draft this for now and make sure I've got them all.

jdchristensen:
(Just to be clear, I meant the other fields of the same Class.)

Alizter:
I tried removing the primes in the other fields and got the wildcat library working, however Coq gets stuck on the `Defined` of one of the proofs of ExactSequence. Specifically `isexact_connect_R`.

Not sure what is going on there.

jdchristensen:
I guess the issue is the one explained in a comment:  `Since apparently a field of a record can't be the source of a coercion ...`.

I think it's fine to merge this to at least get rid of one redefinition.

Alizter:
I also had a go at removing InO_internal from ReflectiveSubuniverse but I was not successful.

Alizter:
Rebasing this breaks `DisplayedEquiv`. @gio256 could you have a look at what is wrong? I wasn't able to fix it.

PR #1810, opened 2024-01-12
use Quotient in ua library

Alizter:
The use of quotient here was easy to replace. In a few places, cbn is replaced with simpl as Quotient interacts poorly with cbn, probably because of simpl never.


jdchristensen (approved):
LGTM.  IIRC, the goal is to remove HIT.quotient, right?

Alizter:
@jdchristensen yes. There are some universe issues with the newer quotient in the Classes library, whcih is why I started inspecting the universe levels closely.

PR #1812, opened 2024-01-13
add universe constraints in GraphQuotient

Alizter:
GraphQuotient takes 3 universe levels. The first is the level of the type we are quotienting. The second is the level of the relation we are quotienting by. Finally, the third is the level the quotient type will live in.

Before we didn't enforce that the first and second universe levels are less than or equal to the third. We add those constraints in.

@jdchristensen This mostly compiles, getting stuck on Ordinals.v. I've tried to understand what is going on and I imagine it is because we are defining a small image which is no longer possible with the new universe restrictions on the quotient type. If this cannot be resolved, it will probably mean that we need to pull in the join construction, or do something clever with prop resizing which we do have.

Therefore I have marked it as draft if you want to take a look. If not we can keep it around until we understand and fix the issue.

### DIFF
--- a/theories/Sets/Ordinals.v
+++ b/theories/Sets/Ordinals.v
@@ -742,7 +742,7 @@ Qed.
 
 (** * Ordinal limit *)
 
-(** We can use PropResizing to resize the image of a function to whatever universe we want. It should be noted that using the join construction we can also do this without PropResizing, but since we already assume it here, we use it. *)
+(** We can use PropResizing to resize the image of a function to whatever universe we want. *)
 Definition image@{i j |} `{PropResizing} {A : Type@{i}} {B : HSet@{j}} (f : A -> B) : Type@{i}
   := Quotient@{i i i} (fun a a' : A => resize_hprop@{j i} (f a = f a')).
 
--- a/theories/Sets/Ordinals.v
+++ b/theories/Sets/Ordinals.v
@@ -742,15 +742,16 @@ Qed.
 
 (** * Ordinal limit *)
 
-Definition image@{i j} {A : Type@{i}} {B : HSet@{j}} (f : A -> B) : Type@{i}
-  := Quotient (fun a a' : A => f a = f a').
+(** We can use PropResizing to resize the image of a function to whatever universe we want. It should be noted that using the join construction we can also do this without PropResizing, but since we already assume it here, we use it. *)
+Definition image@{i j |} `{PropResizing} {A : Type@{i}} {B : HSet@{j}} (f : A -> B) : Type@{i}
+  := Quotient@{i i i} (fun a a' : A => resize_hprop@{j i} (f a = f a')).
 
-Definition factor1 {A} {B : HSet} (f : A -> B)
+Definition factor1 `{PropResizing} {A} {B : HSet} (f : A -> B)
   : A -> image f
   := Quotient.class_of _.
 
-Lemma image_ind_prop {A} {B : HSet} (f : A -> B)
-  (P : image f -> Type) `{forall x, IsHProp (P x)}
+Lemma image_ind_prop@{i j k|} `{PropResizing} {A : Type@{i}} {B : HSet@{j}} (f : A -> B)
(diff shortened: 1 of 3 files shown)


jdchristensen:
I probably won't have a chance to look into this in detail, but I do think it's a bug in Ordinals if it was creating a type in a universe lower than the relation.

BTW, I don't think the Join construction gets around this.  IIRC, the Join construction uses propositional resizing.  So the `JoinConstruction` Section in Seperated.v should have `PropResizing` in the Context.

jdchristensen:
> BTW, I don't think the Join construction gets around this. IIRC, the Join construction uses propositional resizing. So the `JoinConstruction` Section in Seperated.v should have `PropResizing` in the Context.

Actually, I don't think the join construction uses propositional resizing.  In any case, I think what will be needed is simply some careful universe management, or maybe just an extra constraint on universe variables in Ordinals.v.

Alizter:
> > BTW, I don't think the Join construction gets around this. IIRC, the Join construction uses propositional resizing. So the `JoinConstruction` Section in Seperated.v should have `PropResizing` in the Context.
> 
> Actually, I don't think the join construction uses propositional resizing. In any case, I think what will be needed is simply some careful universe management, or maybe just an extra constraint on universe variables in Ordinals.v.

Yeah I thought that was a bit suspect. I haven't had the time to think about this further, but I don't think we are in any rush to get it right just yet.

Alizter:
@jdchristensen Using prop resizing, I've made a small image for Ordinals.v. I also noted that the join construction would give this, but I don't think its worth using here since we already have prop resizing.

This should be good to go now.

theories/Sets/Ordinals.v
```diff
@@ -742,15 +742,16 @@ Qed.
 
 (** * Ordinal limit *)
 
-Definition image@{i j} {A : Type@{i}} {B : HSet@{j}} (f : A -> B) : Type@{i}
-  := Quotient (fun a a' : A => f a = f a').
+(** We can use PropResizing to resize the image of a function to whatever universe we want. It should be noted that using the join construction we can also do this without PropResizing, but since we already assume it here, we use it. *)
```
jdchristensen, line 745: I don't think the join construction would help here.  To apply it, you need to know that the codomain `B` is locally small, and for that, you need to use `PropResizing`.

jdchristensen (commented):
Thanks for locating and solving the issue in Ordinals.v!  Can I push a change to GraphQuotient that uses a Section to avoid the need for most universe annotations and some arguments?

Alizter:
@jdchristensen Yes, no problem.

Alizter:
By the way, as I suspected, we can get an inconsistency before but the argument is a little convoluted. There is a discussion on Zulip: https://hott.zulipchat.com/#narrow/stream/228519-general/topic/Inconsistency.20of.20HITs.20that.20land.20in.20any.20universe

jdchristensen:
About the changes to Ordinals, I wonder if we should add `Quotient'` to Quotient.v which takes `PropResizing` as an assumption and uses that to allow the result to land in the universe of the type `A`.  We could then define `Quotient'_rec`, `Quotient'_ind_hprop`, etc.  Then Ordinals.v could use this instead of Quotient and wouldn't need to deal with `equiv_resize_hprop`.  And it's probably useful for other set-quotients as well.

And maybe the image factorization could also move to Quotient.v.

(I also would prefer these to be called `SetQuotient` and `set_image`, or something like that.)

You just merged this PR, but we could still make these changes in another PR.

Alizter:
@jdchristensen Those suggestions sound fine to me. I've been reluctant to use propositional resizing in more places, but I suppose I don't see any reason why we shouldn't have an alternative set quotient. Maybe it would help to put it in a separate file and call it `SmallSetQuotient` or something.

jdchristensen:
Regarding my suggestion to move the image factorization into Quotient.v, it turns out that it was already there.  PR #1874 changes Ordinals.v to use this.

PR #1813, opened 2024-01-13
symmetry and associativity of products

Alizter:
We prove symmetry and associativity of binary products.

I had a look at extending the result for colimits, but I couldn't get some things to hold definitionally so I will need to take a deeper look at some later point.

I also don't quite see how this will help with the left/right functoriality of the product. @jdchristensen if you have some more comments on this that would be useful. Although, I will probably do it in a follow up PR.

### DIFF
--- a/theories/WildCat/Products.v
+++ b/theories/WildCat/Products.v
@@ -168,6 +168,121 @@ Class HasBinaryProducts (A : Type) `{Is1Cat A} := {
   binary_products :: forall x y : A, BinaryProduct x y;
 }.
 
+(** *** Symmetry of products *)
+
+Section Symmetry.
+
+  (** The requirement of having all binary products can be weakened further to having specific binary products, but it is not clear this is a useful generality. *)
+  Context {A : Type} `{HasEquivs A} `{!HasBinaryProducts A}.
+
+  Definition cat_prod_swap (x y : A) : cat_prod x y $-> cat_prod y x
+    := cat_prod_corec cat_pr2 cat_pr1.
+
+  Lemma cat_prod_swap_cat_prod_swap (x y : A)
+    : cat_prod_swap x y $o cat_prod_swap y x $== Id _.
+  Proof.
+    unfold cat_prod_swap.
+    apply cat_prod_pr_eta.
+    - refine ((cat_assoc _ _ _)^$ $@ _).
+      refine (cat_prod_beta_pr1 _ _ $@R _ $@ _).
+      refine (cat_prod_beta_pr2 _ _ $@ _).
+      exact (cat_idr _)^$.
+    - refine ((cat_assoc _ _ _)^$ $@ _).
+      refine (cat_prod_beta_pr2 _ _ $@R _ $@ _).
+      refine (cat_prod_beta_pr1 _ _ $@ _).
+      exact (cat_idr _)^$.
+  Defined.
+
+  Lemma cate_prod_swap (x y : A)
+    : cat_prod x y $<~> cat_prod y x.
+  Proof.
+    snrapply cate_adjointify.
+    1,2: apply cat_prod_swap.
+    all: apply cat_prod_swap_cat_prod_swap.
+  Defined. 
+
+End Symmetry.
+
+(** *** Associativity of products *)
+
+Section Associativity.
+
+  Context {A : Type} `{HasEquivs A} `{!HasBinaryProducts A}.
+
+  Definition cat_prod_assoc (x y z : A)
+    : cat_prod x (cat_prod y z) $-> cat_prod (cat_prod x y) z
+    := cat_prod_corec (cat_prod_corec cat_pr1 (cat_pr1 $o cat_pr2)) (cat_pr2 $o cat_pr2).
+
+  Definition cat_prod_assoc_inv (x y z : A)
+    : cat_prod (cat_prod x y) z $-> cat_prod x (cat_prod y z)
+    := cat_prod_corec (cat_pr1 $o cat_pr1) (cat_prod_corec (cat_pr2 $o cat_pr1) cat_pr2).
+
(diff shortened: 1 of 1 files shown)


theories/WildCat/Products.v
```diff
@@ -168,6 +168,121 @@ Class HasBinaryProducts (A : Type) `{Is1Cat A} := {
...
+    - apply cat_prod_assoc.
+    - apply cat_prod_assoc_inv.
+    - apply cat_prod_assoc_cat_prod_assoc_inv.
+    - apply cat_prod_assoc_inv_cat_prod_assoc.
+  Defined.
+
+End Associativity.
+
+(** *** Unit of products *)
+
```
jdchristensen, line 285:
```suggestion
```

theories/WildCat/Products.v
```diff
@@ -168,6 +168,121 @@ Class HasBinaryProducts (A : Type) `{Is1Cat A} := {
...
+        refine (cat_prod_beta_pr2 _ _ $@ _).
+        exact (_ $@L (cat_idr _)^$).
+      + refine ((cat_assoc _ _ _)^$ $@ _).
+        refine (cat_prod_beta_pr2 _ _ $@R _ $@ _).
+        refine (cat_prod_beta_pr2 _ _ $@ _).
+        exact (_ $@L (cat_idr _)^$).
+  Defined.
+
+  Lemma cate_prod_assoc (x y z : A)
+    : cat_prod x (cat_prod y z) $<~> cat_prod (cat_prod x y) z.
```
jdchristensen, line 273: The slick way to prove associativity of a binary operation once you have symmetry is to prove the twist equivalence, `A * (B * C) <~> B * (A * C)`.  The reason this is easier is that you don't have to define a separate inverse map, and you only have to show that one composite is the identity.  Then associativity follows from this and symmetry.  (You can see how in my proof that `Join` is associative.)

theories/WildCat/Products.v
```diff
@@ -168,6 +168,121 @@ Class HasBinaryProducts (A : Type) `{Is1Cat A} := {
   binary_products :: forall x y : A, BinaryProduct x y;
 }.
 
+(** *** Symmetry of products *)
+
+Section Symmetry.
+
+  (** The requirement of having all binary products can be weakened further to having specific binary products, but it is not clear this is a useful generality. *)
```
jdchristensen, line 175: I guess what you mean to say here is that one only needs the two products in the statement.  In fact, if you just have a single product of `x` and `y`, you can prove that it is also the product of `y` and `x`.  So that's a pretty clean statement.  Then the commutativity of products when you do have all products follows from the uniqueness of the product (which would be a nice thing to have anyways).  So I'd be inclined to organize it that way. You are essentially embedding the proof of the uniqueness in the proof you have given.

jdchristensen:
> I also don't quite see how this will help with the left/right functoriality of the product.

Well, if you already have functoriality of the product in the left variable, you could define functoriality in the right variable like this:  given x, y, y', f: y → y', you get x * y → y * x → y' * x → x * y', where the first and last maps are the symmetry, and the middle map is the functoriality in the left.  Then the symmetry is natural by definition, and the 1-functoriality should be easy to prove.  But I'm not sure I like this idea, as the induced map is a bit more complicated than needed.

An intermediate idea is to define the 0-functoriality by hand, prove that symmetry is natural with respect to that, and then use symmetry to prove the 1-functoriality.  Also maybe not worth it.  A third option is to prove bifunctoriality from the start, and then to deduce functoriality in each variable.  (For Join, we prove bifunctoriality.)

I'm throwing these ideas out there because the proofs of functoriality were longer than I expected, but it's fine if none of these ideas sound good and you just leave it as it is.

theories/WildCat/Products.v
Alizter: I wonder if we can generalise the twist argument to any symmetric monoidal category, but for that we would need associativity to begin with. I'm thinking there might be some common structure between these bifunctors allowing the twist argument.

theories/WildCat/Products.v
Alizter: The problem is that the witness of binary products will not be definitionally the same. So even if we only require `BinaryProduct x y` then the statement about `cat_prod` being symmetric would not apply to the instances of a `HasBinaryProducts` but only a single one and its swapped version.

theories/WildCat/Products.v
Alizter: Anyway, I've used the twist argument, and it is slick!

Alizter:
>I'm throwing these ideas out there because the proofs of functoriality were longer than I expected, but it's fine if none of these ideas sound good and you just leave it as it is.

Yes, it is a bit long, but at least the terms are obvious. I think if we use symmetry in the definition, then it will become harder to reason about the terms in other proofs.

The new twist associativity proof for instance uses the fmap value of these functors and only works because it is cat_prod_corec as we expect.

theories/WildCat/Products.v
jdchristensen: I probably wasn't clear enough.  First you prove that if z, p1 : z -> x and p2 : z -> y form a product of x and y (so the appropriate map is an equivalence), then z, p2 : z -> y, p1 : z -> x form a product of y and x.

Second, you prove that if z, p1, p2 and z', p1', p2' are both products of x and y, then there is an equivalence from z to z' commuting with the projections.

Finally, we assume that we have all binary products, and let x and y be objects.  Let z, p1, p2 be a binary product of x and y, and let w, q1, q2 be a binary product of y and x.  By the first result, z, p2, p1 is also a product of y and x, so by the second result, there is an equivalence from z to w, as desired.  And that equivalence should turn out to be definitionally equal to the one you've defined.  The amount of work should be about the same, but you also get results one and two as well, which are independently useful.

PR #1816, opened 2024-01-16
Hartogs: improve speed, reorganize

jdchristensen:
There are two commits and they need to be read independently because the second one moves many things around.

The first one reduces the build time of Hartogs.v by 47%.

The second one moves things from inside the Hartogs Section to outside of it when they don't depend on the fixed type `A` and the universe variable `A` that are in the context.  Some are moved earlier in the file, and some are moved to Ordinals.v.  While editing, I changed lots of whitespace to match our conventions and made other minor fixes.

### DIFF
--- a/theories/Sets/Hartogs.v
+++ b/theories/Sets/Hartogs.v
@@ -6,9 +6,61 @@ From HoTT.Sets Require Import Ordinals Powers.
 
 (** This file contains a construction of the Hartogs number. *)
 
+(** We begin with some results about changing the universe of a power set using propositional resizing. *)
+
+Definition power_inj `{PropResizing} {C : Type@{i}} (p : C -> HProp@{j})
+  : C -> HProp@{k}.
+Proof.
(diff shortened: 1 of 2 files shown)


theories/Sets/Hartogs.v
```diff
@@ -6,37 +6,79 @@ From HoTT.Sets Require Import Ordinals Powers.
...
   Definition hartogs_number' : Ordinal.
   Proof.
     set (carrier := {B : Ordinal@{A _} & card B <= card A}).
     set (relation := fun (B C : carrier) => B.1 < C.1).
-
-    exists carrier relation. srapply (isordinal_simulation pr1).
-    - exact _.
-    - exact _.
+    exists carrier relation. snrapply (isordinal_simulation pr1).
+    1, 2, 3, 4: exact _.
```
Alizter, line 80:
```suggestion
    1-4: exact _.
```

Alizter (approved):

PR #1817, opened 2024-01-17
Adapt to coq/coq#18352 (ltacX_common_plugin)

SkySkimmer:


Alizter:
Is there a way to make this backwards compatible?

theories/Basics/Notations.v
```diff
@@ -9,6 +9,7 @@ Reserved Notation "'exists' x .. y , p"
 Local Set Warnings Append "-skip-spaces-curly".
 
 (** ML Tactic Notations *)
+Declare ML Module "ltacX_common_plugin".
 Declare ML Module "ltac_plugin".
```
SkySkimmer, line 13:
```suggestion
Require Coq.Init.Ltac.
```
would work I think.

SkySkimmer:
In the end we decided to do the more backwards compatible https://github.com/coq/coq/pull/18528 instead.

PR #1819, opened 2024-01-18
Truncatedness of pointed maps and pForall

jdchristensen:
This proves Buchholtz-van Doorn-Rijke, Theorem 4.2 and related things.  The first commit has the meat, and can be reviewed on its own.  The second commit just contains minor cleanups that I noticed while reading things.

### DIFF
--- a/theories/Modalities/Modality.v
+++ b/theories/Modalities/Modality.v
@@ -160,6 +160,15 @@ Proof.
   rapply Build_Modality'.
 Defined.
 
+(** When combined with [isequiv_oD_to_O], this yields Theorem 7.7.7 in the book. *)
+Definition isequiv_oD_to_O_modality
+           `{Funext} (O : Modality) {A : Type}
+           (B : O A -> Type) `{forall a, In O (B a)}
+  : IsEquiv (fun (h : forall oa, B oa) => h oD to O A).
+Proof.
+  srapply (isequiv_oD_to_O O O).
+Defined.
+
 (** Of course, modalities have dependent eliminators. *)
 Definition O_ind {O : Subuniverse} {A : Type} `{ReflectsD O O A}
   := @OO_ind O O A _ _.
@@ -197,15 +206,6 @@ Proof.
     + rapply O_ind'.
     + rapply O_ind_beta'.
     + rapply inO_paths'.
-Defined.
-
-(** When combined with [isequiv_oD_to_O], this yields Theorem 7.7.7 in the book. *)
-Definition isequiv_oD_to_O_modality
-           `{Funext} (O : Modality) {A : Type}
-           (B : O A -> Type) `{forall a, In O (B a)}
-  : IsEquiv (fun (h : forall oa, B oa) => h oD to O A).
-Proof.
-  srapply (isequiv_oD_to_O O O).
 Defined.
 
 (** A tactic that extends [strip_reflections] to modalities. It handles non-dependent elimination for reflective subuniverses and dependent elimination for modalities. [strip_truncations] does the same for truncations, but introduces fewer universe variables, so tends to work better when removing truncations. *)
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -443,7 +443,7 @@ Definition pointed_fam {A : pType} (B : A -> pType) : pFam A
 Definition point_pforall {A : pType} (B : A -> pType) : pForall A (pointed_fam B)
   := Build_pForall A (pointed_fam B) (fun x => point (B x)) 1.
 
-(** The pointed type of pointed maps. For dependent pointed maps we need a family of pointed types, not just a family of types with a point over the basepoint of [A]. *)
+(** The pointed type of dependent pointed maps. Note that we need a family of pointed types, not just a family of types with a point over the basepoint of [A]. *)
 Definition ppForall (A : pType) (B : A -> pType) : pType
   := [pForall A (pointed_fam B), point_pforall B].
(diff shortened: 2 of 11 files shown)


theories/Modalities/Modality.v
```diff
@@ -160,6 +160,15 @@ Proof.
   rapply Build_Modality'.
 Defined.
 
+(** When combined with [isequiv_oD_to_O], this yields Theorem 7.7.7 in the book. *)
+Definition isequiv_oD_to_O_modality
```
jdchristensen, line 164: This was simply moved earlier so that the "this" in the comment made sense.

theories/Pointed/Core.v
```diff
@@ -71,6 +72,9 @@ Definition Build_pMap (A B : pType) (f : A -> B) (p : f (point A) = point B)
   : A ->* B
   := Build_pForall A (pfam_const B) f p.
 
+(** The [&] tells Coq to use the context to infer the later arguments (in this case, all of them). *)
+Arguments Build_pMap & _ _ _ _.
```
jdchristensen, line 76: This lets us write `(Build_pMap _ _ ...)` in some cases where Coq insisted that the arguments be filled in.

theories/Pointed/Core.v
```diff
@@ -445,6 +455,11 @@ Notation "'ppforall'  x .. y , P"
 Definition pconst {A B : pType} : A ->* B
   := point_pforall (fun _ => B).
 
+Definition ppMap (A B : pType) : pType
+  := [A ->* B, pconst].
```
jdchristensen, line 459: The definition of `ppMap` was changed so that the underlying unpointed type is `A ->* B` rather than a large term involving `pfam_const`.

theories/Pointed/pTrunc.v
```diff
@@ -135,21 +125,11 @@ Definition ptr_loops_eq `{Univalence} (n : trunc_index) (A : pType)
   : pTr n (loops A) = loops (pTr n.+1 A) :> pType
   := path_ptype (ptr_loops n A).
 
-Definition pequiv_ptr_functor {X Y : pType} n
-  : X <~>* Y -> pTr n X <~>* pTr n Y.
```
jdchristensen, line 139: We had this result twice.

theories/Pointed/Core.v
```diff
@@ -235,6 +239,16 @@ Definition issig_pequiv' (A B : pType)
...
   := ltac:(make_equiv).
 
+(** pForall can also be described as a type of extensions. *)
+Definition equiv_extension_along_pforall `{Funext} {A : pType} (P : pFam A)
+  : ExtensionAlong (unit_name (point A)) P (unit_name (dpoint P)) <~> pForall A P.
+Proof.
+  unfold ExtensionAlong.
+  refine (issig_pforall A P oE _).
+  apply equiv_functor_sigma_id; intro s.
+  symmetry; apply equiv_unit_rec.
```
Alizter, line 249: I checked, and it's probably not possible to remove funext from `equiv_unit_rec` so the funext will have to stay.

theories/Pointed/pTrunc.v
```diff
@@ -1,10 +1,12 @@
-Require Import Basics Types WildCat Truncations.Core Truncations.SeparatedTrunc
+Require Import Basics Types WildCat Truncations
   Pointed.Core Pointed.pEquiv Pointed.Loops Pointed.pModality.
 
 Local Open Scope pointed_scope.
 
 (** * Truncations of pointed types *)
 
+(** TODO: Many things here can be generalized to any modality or any reflective subuniverse, and could be moved to pModality.v *)
```
Alizter, line 8: Out of curiosity, do we have a use for them outside of truncations?

theories/Pointed/pTrunc.v
```diff
@@ -168,3 +148,29 @@ Proof.
     reflexivity.
 Defined.
 
+(** ** Truncatedness of [pForall] and [pMap] *)
+
+(** Buchholtz-van Doorn-Rijke, Theorem 4.2:  Let [j >= -1] and [n >= -2].  When [X] is [j]-connected and [Y] is a pointed family of [j+k+1]-truncated types, the type of pointed sections is [n]-truncated.  We formalize it with [j] replaced with a trunc index [m], and so there is a shift compared to the informal statement. This version also allows [n] to be one smaller than BvDR allow. *)
+Definition istrunc_pforall `{Univalence} {m n : trunc_index}
```
Alizter, line 154: Not that it particularly matters, but perhaps calling this a lemma or proposition would be more useful to the reader. I don't think it warrants the status of a theorem however.

Alizter: Also for the following results.

theories/Pointed/Core.v
Alizter: Perhaps it would be interesting to note that it is a special case of ppForall?

Alizter (approved):
LGTM, I left some comments.

theories/Pointed/pTrunc.v
jdchristensen: In homotopy theory, people often use localization at a prime in the pointed world, so I expect that in the future we'll have uses for these things for other RSUs.

theories/Pointed/pTrunc.v
jdchristensen: I usually just use "Definition" rather than trying to rank the importance of things.  And Definition is more flexible, since you can use the := style.

PR #1820, opened 2024-01-18
BlakersMassey: slight speedup

jdchristensen:
This PR slightly speeds up one line near the end of the file.

It turns out that the line that was labelled "slow" is instantaneous.  I don't know what fixed that, but it's not this PR.  In fact, the whole file only takes 0.7s to process, and no line is slower than 0.1s.


Alizter:
Did the new IsTrunc improve it I wonder?

Alizter (approved):

jdchristensen:
> > It turns out that the line that was labelled "slow" is instantaneous.

> Did the new IsTrunc improve it I wonder?

I tested with a commit from September 2023, and that line (as well as the whole file) were fast then as well.  So I don't know what fixed it.

PR #1822, opened 2024-01-20
ReflectiveSubuniverse: add a couple of things and simplify one proof

jdchristensen:
This contains very minor changes.  The main reason I'm making this an independent PR is that I have another question about typeclass resolution.  I'll add a comment beside the relevant spot in the code, and will attach the verbose typeclasses debug log here.
[debug.log](https://github.com/HoTT/Coq-HoTT/files/13996119/debug.log)

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -298,7 +298,7 @@ Section Reflective_Subuniverse.
   Defined.
 
   (** If [T] is in the subuniverse, then [to O T] is an equivalence. *)
-  Definition isequiv_to_O_inO (T : Type) `{In O T} : IsEquiv (to O T).
+  Global Instance isequiv_to_O_inO (T : Type) `{In O T} : IsEquiv (to O T).
   Proof.
     pose (g := O_rec idmap : O T -> T).
     refine (isequiv_adjointify (to O T) g _ _).
@@ -309,7 +309,6 @@ Section Reflective_Subuniverse.
     - intros x.
       apply O_rec_beta.
   Defined.
-  Global Existing Instance isequiv_to_O_inO.
 
   Definition equiv_to_O (T : Type) `{In O T} : T <~> O T
     := Build_Equiv T (O T) (to O T) _.
@@ -509,6 +508,11 @@ Section Reflective_Subuniverse.
       - symmetry. exact (O_rec_beta (g o f) x).
     Defined.
 
+    (** In particular, we have: *)
+    Definition O_rec_postcompose_to_O {A B : Type} (f : A -> B) `{In O B}
+    : to O B o O_rec f == O_functor f
+    := O_rec_postcompose f (to O B).
+
   End Functor.
 
   Section Replete.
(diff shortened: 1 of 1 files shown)


theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -575,19 +579,26 @@ Section Reflective_Subuniverse.
...
     : IsEquiv (O_rec (O := O) f).
     Proof.
-      apply isequiv_O_inverts.
-      nrefine (cancelR_isequiv (O_functor (to O A))); [ exact _ | ].
-      nrefine (isequiv_homotopic (O_functor (O_rec f o to O A))
-                                (O_functor_compose _ _)).
-      refine (isequiv_homotopic (O_functor f)
-               (O_functor_homotopy _ _ (fun x => (O_rec_beta f x)^))).
+      srapply (cancelL_isequiv (to O B)).
+      1,2: exact _. (* Why is this line needed? *)
```
jdchristensen, line 583: `srapply (cancelL_isequiv (to O B)).` generates three goals:  
```
?X259 : (PreReflects O B)
?X261 : (IsEquiv (to O B))
?X262 : (IsEquiv (to O B o O_rec f))
```
The third needs a proof (line 584), but the first two are instances.  And the debug.log file shows that typeclass search finds the instances for those two, and then fails for the third.  Then for some reason I'm left with all three goals, instead of just the last one.  Can someone explain?  Earlier in the file (e.g. on line 566) the same goals are solved and stay solved.

jdchristensen: BTW, exactly the same thing happens with `simple refine (cancelL_isequiv (to O B)).`, so this has nothing to do with the `srapply` tactic.  I have also attached the less verbose version of the log, in case that is already enough to see the issue.
[debug1.txt](https://github.com/HoTT/Coq-HoTT/files/13998066/debug1.txt)

jdchristensen: One difference I noticed is that in the cases where something like this succeeds, the debug log has all of the top-level goals numbered as "1:".  But in this case, where two solvable goals are left unsolved, the top-level goals are numbered "1:", "2:" and "3:".  Maybe this indicates that Coq thinks that they are dependent in some sense, so if 3 can't be solved, it backtracks and doesn't solve 1 and 2?

And some more experimenting found that if I provide the implicit argument `C`, the problem goes away.  But `C` should be clear from the provided argument `to O B`.

In any case, I'd like to merge this, but even after merging, if anyone can explain the behaviour to me, that would be great.

Alizter: I think you've correctly identified the dependency behaviour but perhaps @SkySkimmer can confirm this is the case.

Alizter (approved):

theories/Modalities/ReflectiveSubuniverse.v
SkySkimmer: IDK

jdchristensen: It can be simplified a bit further:
```coq
Class IsEquiv {A B : Type} (f : A -> B) := {}.

Class Reflects (B : Type) := {}.

Definition to_O (B : Type) `{Reflects B} : B -> B.
Admitted.

Global Instance isequiv_to_O (B : Type) `{Reflects B} : IsEquiv (to_O B).
Admitted.

Definition cancelL_isequiv {A B C} (g : B -> C) {f : A -> B}
  `{IsEquiv B C g} `{IsEquiv A C (fun a => g (f a))}
  : IsEquiv f.
Admitted.

Definition isequiv_O_rec_O_inverts
       {A B : Type} `{Reflects B} (f : A -> B)
  : IsEquiv f.
Proof.
  Typeclasses eauto := debug.
  simple refine (cancelL_isequiv (to_O B)).
  1, 2: exact _. (* Coq finds these instances but then ignores them. *)
Abort.
```

Alizter: @jdchristensen I think that would be worth reporting.

jdchristensen: In https://github.com/coq/coq/issues/18526, it was explained that when there are dependencies between typeclass goals, earlier solutions are thrown away if later goals can't be automatically solved.  I can see how this might be reasonable in some situations, but also think the behaviour I expected could be good.  I also figured out that the reason that the version with `(C:=O B)` works is that unification solves the early typeclass goal, so it isn't part of the typeclass search.  As an experiment, I tried redefining `to O` to take an RSU `O` instead of a subuniverse, so it no longer requires typeclass search.  (I renamed the current `to` to `to'`, and had to add primes whenever we had `to O` with `O` a subuniverse.)  That allowed me to remove the annotation `(C:=O B)` here, but had no other effects.  I thought that the library might build faster, but it was unchanged, and I thought that maybe some goals would be solved automatically, but none were.  So I won't make a PR for that.

PR #1824, opened 2024-01-20
Spaces/Torus/*: simplify and try to remove Funext

jdchristensen:
The first commit simplifies `loops_torus`.  The second just records that the loops commute.  These two commits can go in.

The third commit attempts to remove Funext from TorusEquivCircles.v, using the standard trick.  This goes fine, except for two things, which are closed with `Admitted.`  The first involves some "cube-algebra" that would now be different because Funext is gone.  I'm hoping it'll be a bit simpler, but I'm not good at working with cubes, so I was hoping @Alizter might take a look.

The second `Admitted` happens because a later `Defined` fails with `Error: Case analysis on private inductive GraphQuotient.GraphQuotient.GraphQuotient`.  I'm not sure what is up with that.

(It would also be possible to prove this equivalence using the 0-groupoid Yoneda lemma.  It would be a bit verbose, but I think it would be straightforward and would build quickly.)

I'll mark this as WIP for now.  If we can't solve those problems, I'll PR just the first two commits, along with removing the `Section` in TorusEquivCircles, to make it faster.

### DIFF
--- a/theories/Spaces/Torus/Torus.v
+++ b/theories/Spaces/Torus/Torus.v
@@ -85,6 +85,10 @@ Admitted.
 (** The torus is pointed. *)
 Global Instance ispointed_torus : IsPointed Torus := tbase.
 
+(** The loops commute. *)
+Definition loops_commute_torus : loop_a @ loop_b = loop_b @ loop_a
+  := equiv_sq_path^-1 surf.
+
 (* TODO:
 (* We ought to be able to prove the computation rules all at the same time *)
 (* This gives me the idea of writing all our computation rules as a
--- a/theories/Spaces/Torus/TorusEquivCircles.v
+++ b/theories/Spaces/Torus/TorusEquivCircles.v
@@ -5,11 +5,6 @@ Require Import Spaces.Circle Spaces.Torus.Torus.
 (* In this file we:
     - Prove that the Torus is equivalent to the product of two circles
  *)
-Section TorusEquivCircle.
-
-  (* We use function extensionality when defining the
-     map from circles to torus *)
-  Context `{Funext}.
 
   (* Here is a cube filler for help with circle recursion into the torus *)
   Definition c2t_square_and_cube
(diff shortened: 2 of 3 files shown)


Alizter:
I think that using the ZeroGroupoid yoneda lemma would be a welcome improvement. If it let's us avoid the square path algebra, then we can go as far as to remove the PathSquare constructor of Torus and replace it with the usual 2-path. This would be difficult as stating the dependent eliminator and it's computation rule is quite involved, however you might be able to get away with a non-dependent eliminator and a uniqueness statement.

jdchristensen:
According to the CI, the second `Defined.` works fine with Coq 8.17.  This is the one that has `Case analysis on private inductive` with Coq 8.18.  I wonder what different.

jdchristensen:
> I think that using the ZeroGroupoid yoneda lemma would be a welcome improvement.

I agree, but I won't have time to tackle it.  Do you think a quick fix to the cube algebra is possible in the meantime?  If not, I'll make a separate PR with just the first two commits, and the Section removed.

Alizter:
> According to the CI, the second `Defined.` works fine with Coq 8.17. This is the one that has `Case analysis on private inductive` with Coq 8.18. I wonder what causes the difference.

If you search https://coq.inria.fr/refman/changes.html for `private` you will see there was a kernel bug where private inductive types were leaking outside of their module. I think this probably caused the issue. I don't think it would have led to an inconsistency, since somebody would have to know about this bug and purposefully exploit it and derive some falsehood. So unlikely to have affected us in practice.

Alizter:
> > I think that using the ZeroGroupoid yoneda lemma would be a welcome improvement.
> 
> I agree, but I won't have time to tackle it. Do you think a quick fix to the cube algebra is possible in the meantime? If not, I'll make a separate PR with just the first two commits, and the Section removed.

I had a go at swapping the arguments before and I got lost with the cube algebra. The argument presented here is based off of Licata and Brunerie's paper here: https://dlicata.wescreates.wesleyan.edu/pubs/lb15cubicalsynth/lb15cubicalsynth.pdf

Kristina had another proof that was a more technical argument but avoided cubes. I had since spoken to her and she told me that there are some major simplifications that can be done to her argument.

I'll try to have another go and see if I can adapt the cube algebra. The complicated existential variables you see here is a way of coaxing coq's elaborator to suggest certain paths. At the time, the unification wasn't strong enough to infer this.

I'll also try to have a stab at a ZeroGroupoid yoneda argument. I'm not sure it will allow us to avoid all of the algebra, however it should simplify the equivalence.

Alizter:
The algebra we are trying to rewrite when we get rid of funext is very sensitive to unfolding and can cause Coq to run out of memory quite quickly. I've tried tacking it but ultimately didn't get anywhere.

jdchristensen:
Ok, closing this.  I left a comment in the code pointing to this PR in case anyone wants to try in the future.

PR #1825, opened 2024-01-21
Torus improvements, keeping Funext

jdchristensen:
This is the variant of PR #1824 that doesn't try to remove Funext, but still makes the other improvements plus a few additional changes.  The commit titles are self-explanatory.  You should deselect the indentation commit to see the real changes.


Alizter (approved):
Nice!

jdchristensen:
The changes here didn't significantly speed up TorusEquivCircle.   It's still one of the slowest files in the library with one of the slowest lines.  But it's not that bad, so isn't worth putting too much effort into.

PR #1826, opened 2024-01-21
all cube concatenations

Alizter:
I noticed that these were missing and had a poor definition. This should make it easier to understand and work with.


jdchristensen:
So the previous comment about it being slow with tactics no longer applies?  That's great if it's true.

Would it make sense to only have the shared arguments in the context, and the arguments specific to each type of concatenation as part of the statements?  I know Coq will drop unused arguments, but this might make things more clear to the reader.

Alizter:
In the previous proof we were eliminating 2 cubes, whereas here we eliminate 1 and a few paths, hence the speedup.

I've made sure that the section only contains common arguments now.

jdchristensen (approved):

PR #1828, opened 2024-01-23
coproduct symmetry and associativity

Alizter:
Follows definitionally from corresponding result for products. I had to tweak some classes to be definitional for this to work. I also added an extra test about the opposite category.

Unfortunately, I couldn't work out how to deal with the functoriality yet as opposite functors become quite confusing to work with.

### DIFF
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -2,9 +2,10 @@ From HoTT Require Import Basics WildCat.Core WildCat.Opposite.
 
 (* Opposites are (almost) definitionally involutive. *)
 
-Definition test1 A `{Is01Cat A} : A = (A^op)^op := 1.
-Definition test2 A `{x : Is01Cat A} : x = is01cat_op (A := A^op) := 1.
-Definition test3 A `{x : Is2Graph A} : x = is2graph_op (A := A^op) := 1.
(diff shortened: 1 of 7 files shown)


test/WildCat/Opposite.v
```diff
@@ -2,9 +2,10 @@ From HoTT Require Import Basics WildCat.Core WildCat.Opposite.
 
 (* Opposites are (almost) definitionally involutive. *)
 
-Definition test1 A `{Is01Cat A} : A = (A^op)^op := 1.
-Definition test2 A `{x : Is01Cat A} : x = is01cat_op (A := A^op) := 1.
-Definition test3 A `{x : Is2Graph A} : x = is2graph_op (A := A^op) := 1.
+Definition test1 A : A = (A^op)^op := 1.
```
jdchristensen, line 5: This isn't really testing anything, since it also works with a single `^op`.  But maybe it's fine, as the beginning of the pattern.  To make it clear that we are just checking equality of the underlying types, we could add `:> Type` before the `:=`.

jdchristensen (approved):

test/WildCat/Opposite.v
Alizter:
```suggestion
Definition test1 A : A = (A^op)^op :> Type := 1.
```

PR #1830, opened 2024-01-26
Minor updates to Pointed/*, ReflectiveSubuniverse

jdchristensen:
The only non-trivial thing is an improvement to the universes in the result `istrunc_pmap` that I added recently.  When I went to use it in other code, I found that Coq had made some sub-optimal choices that meant I couldn't use it.

### DIFF
--- a/theories/Modalities/ReflectiveSubuniverse.v
+++ b/theories/Modalities/ReflectiveSubuniverse.v
@@ -162,13 +162,12 @@ Class Reflects@{i} (O : Subuniverse@{i}) (T : Type@{i})
 Arguments extendable_to_O O {T _ _ Q Q_inO}.
 
 (** Here's a modified version that applies to types in possibly-smaller universes without collapsing those universes to [i]. *)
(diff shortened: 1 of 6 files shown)


Alizter:
LGTM but I'm confused about the adjunction proof failing. Could you explain? I seem to remember we agreed to keep that proof as a demo rather than use Wildcat/Adjunct.v. We could probably move it to contrib in that case.

Alizter:
Actually rereading this, it seems fine as I forgot about the earlier fail.

Alizter (approved):

jdchristensen:
> LGTM but I'm confused about the adjunction proof failing. Could you explain?

It's the proof of naturality of one of the proofs of the adjunction that fails.  It always failed, but now it fails a bit earlier since something changed in the library.  This proof of naturality is commented out, so I only noticed this when I tried uncommenting it.

The context here is that the later proof of the adjunction has rigid universe variables that makes it useless in my application.  So I was experimenting with the other proof.  In a future PR, I'm probably going to make some changes here.

PR #1831, opened 2024-01-28
A search and replace script

jdchristensen:
I modified etc/coqcstriprequires.py to a new script etc/coqcreplace.py that does arbitrary search-and-replaces in .v files, testing each replacement and only keeping it if the file compiles and does so within a specified timeout (as a function of the previous duration).  The script is in the second-last commit below.  The earlier commits contains examples of its usage:

1) This is a manual commit, changing "simple notypeclasses refine" to "snrefine", to make the next commit work better.

2) This commit adds "n" before "rapply" and "refine" in 13 places.  The script found hundreds of places where this replacement worked, but to prevent unnecessary churn, I limited it to places where the time improved a bit.  Then I manually tested the replacements it chose, and trimmed it down to these 13.  They are all very small improvements, which means that we aren't currently wasting any appreciable time on typeclass search.

3, 4, 5) Remove unneeded uses of Univalence and Funext.  There were surprisingly many of these.  I took care to not remove a Funext hypothesis that would be replaced by a Univalence hypothesis from a Context line, when the result only needed Funext.

6) A manual simplification of Univalence/Funext in ReflectiveSubuniverses.v.

7) Change `Require Import Foo` to `Require Import Foo.Core` for many `Foo`, including `Truncations`, `WildCat`, `Spaces.Nat`, etc.  This slightly improves the parallel build speed.

8) Change one import of `Groups` to `Groups.Group`.  The script checked the whole library, and also for AbGroups, but this was the only one that worked.

9) The script.

10) Backport of some changes to the script to coqcstriprequires.py.

The commits can be read independently, which might make reviewing this easier.

### DIFF
--- a/etc/coqcreplace.py
+++ b/etc/coqcreplace.py
@@ -1,0 +1,228 @@
+#!/usr/bin/python3
+
+# A wrapper for coqc that will try to replace a regular expression
+# with a replacement string.  The results need to be inspected by
+# hand, as it will often be too eager.  See the "Limitations" section
+# below.
+
+# The patterns are specified using environment variables, using python's
+# regular expression syntax, e.g.:
+#   export COQC_SEARCH='(^|[^n])(refine|rapply)'
+#   export COQC_REPLACE='\1n\2'
+# This will replace an occurence of 'refine' or 'rapply' that is either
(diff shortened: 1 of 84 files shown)


Alizter (approved):
LGTM!

PR #1837, opened 2024-02-08
WildCat/Induced: various minor simplifications

jdchristensen:
E.g. don't eta-expand function calls; don't do unneeded "cbn in *"; use notypeclasses when possible.


jdchristensen:
@gio256 Thanks for pointing this file out to me.

Alizter:
I forgot about this file too. IIRC it is used in the functor category.

Alizter (approved):

Issue #1833, opened 2024-01-30
Flattening lemma for GraphQuotient


Alizter:
It would be useful to have the flattening lemma for GraphQuotient as it should allow us to derive similar flattening lemmas for other HITs we define in terms of GraphQuotients.

One thing I've never understood with the flattening lemma is how to present the data. In the book we use univalence to construct a map `F : Q -> Type` and then use those parts to construct a quotient equivalent to `sig F`. I was thinking about if it was possible to start with a map `F : Q -> Type`, but there doesn't seem to be a way to get the Q_rec data out of this. I guess this is one of the peculiarities with HITs is that they don't really have an eta rule.

Without the flattening lemma, we cannot describe the total space of the Hopf fibration and relate it to a join. Being able to do this will let us compute Pi3S2 and construct isomorphisms PinS2=PinS3.

jdchristensen:
> Without the flattening lemma, we cannot describe the total space of the Hopf fibration and relate it to a join.

We have HIT/Flattening.v which seems to handle coequalizers, and Colimits/Colimit_Flattening.v which seems to handle arbitrary colimits.  Then Colimits/Colimit_Pushout_Flattening.v which deduces the flattening lemma for pushouts from the one for general colimits.  The first two seem redundant, and I don't know why both exist.  In any case, all we need are pushouts for the Hopf fibration, so I think we should be good, right?

Alizter:
@jdchristensen The first is from 10 years ago and was from when Coeq was the HIT that was used to define others. The second is for the Type valued diagram Colimit library and I adapted that to use Coeq IIRC. I'm suggesteing we state the flattening lemma for GraphQuotient and from that all flattening lemmas can be derived.

Also, the first flattening lemma proof we have uses a HIT for the sigma type of the family, and AFAIK this is an unnecessary step as most HIT describe their own flattening lemma HIT by proving the induction principle it should have.

Therefore having the flattening lemma for GraphQuotients will be useful to derive them for all of them.

I believe a higher level idea of what is happening is that we are producing "lax colimits" in the Grothendieck category. Therefore it might be possible to leverage some idea of "displayed colimit" and show that Type can turn displayed colimits into colimits in the Grothendieck category. This should be a "descent property" that is found in all oo-topoi and probably other categories like pType, Spectra and 1-categories.

[...]

PR #1840, opened 2024-02-11
flattening for pushouts + total space of Hopf construction

Alizter:
We use the colimits library to prove flattening for pushouts. The proof is a bit annoying as it doesn't compute well.

We use this to show that the Hopf construction has total space equivalent to a join. Because of the uncomputable nature of the flattening lemma we are unable to show that it preserves the base point.

This should be enough for something else I am working on. Corollaries like the Hopf fibrations for the spheres would ideally come once we sort out the computability.

I tried proving the flattening lemma directly for GraphQuotient and Pushout but I was not able to get around the difficult path algebra easily. The fact that this is done in the Colimits library is good and we should use it, however it is not as easy as I thought to transfer across equivalences in this case. Maybe I am missing something obvious?

### DIFF
--- a/theories/Colimits/Colimit_Pushout_Flattening.v
+++ b/theories/Colimits/Colimit_Pushout_Flattening.v
@@ -10,7 +10,7 @@ Require Import Colimits.Colimit_Flattening.
 
 (** * Pushout case *)
 
-(** We show the flattening lemma in the case of the pushout. *)
+(** We show the flattening lemma in the case of the pushout. This pushout is defined as the colimit of a span and is not the pushout that apepars elsehwere in the library. The flattening lemma here however can be used to derive the flattening lemma there and this is done in Colimits/PushoutFlattening.v. *)
 
 Section POCase.
   Context `{Univalence} {A B C} {f: A -> B} {g: A -> C}.
--- a/theories/Colimits/Colimit_Pushout_Flattening.v
+++ b/theories/Colimits/Colimit_Pushout_Flattening.v
@@ -10,7 +10,7 @@ Require Import Colimits.Colimit_Flattening.
 
 (** * Pushout case *)
 
-(** We show the flattening lemma in the case of the pushout. This pushout is defined as the colimit of a span and is not the pushout that apepars elsehwere in the library. The flattening lemma here however can be used to derive the flattening lemma there and this is done in Colimits/PushoutFlattening.v. *)
+(** We show the flattening lemma in the case of the pushout. This pushout is defined as the colimit of a span and is not the pushout that appears elsewhere in the library. The flattening lemma here however can be used to derive the flattening lemma there and this is done in Colimits/PushoutFlattening.v. *)
 
 Section POCase.
   Context `{Univalence} {A B C} {f: A -> B} {g: A -> C}.
--- a/theories/Colimits/Colimit_Pushout_Flattening.v
+++ b/theories/Colimits/Colimit_Pushout_Flattening.v
@@ -49,31 +49,31 @@ Section POCase.
   Definition PO_flattening
     : PO (functor_sigma f f0) (functor_sigma g g0) <~> exists x, POCase_P x.
   Proof.
-    assert (PO (functor_sigma f f0) (functor_sigma g g0)
-            = Colimit (diagram_sigma POCase_E)). {
+    transitivity (Colimit (diagram_sigma POCase_E)).
+    { apply equiv_path.
       unfold PO; apply ap.
       srapply path_diagram; cbn.
       - intros [|[]]; cbn. all: reflexivity.
       - intros [[]|[]] [[]|[]] [] x; cbn in *.
         all: reflexivity. }
-    rewrite X; clear X.
     transitivity (exists x, E' (span f g) POCase_E POCase_HE x).
     - apply flattening_lemma.
     - apply equiv_functor_sigma_id.
       intro x.
-      assert (E' (span f g) POCase_E POCase_HE x = POCase_P x). {
-        unfold E', POCase_P, PO_rec.
-        f_ap. srapply path_cocone.
-        - intros [[]|[]] y; cbn.
-          1: apply path_universe_uncurried; apply g0.
-          all: reflexivity.
-        - intros [[]|[]] [[]|[]] []; cbn.
-          + intro y; simpl; hott_simpl.
-            unfold path_universe.
-            rewrite <- path_universe_V_uncurried.
-            refine (path_universe_compose (f0 y)^-1 (g0 y))^. 
-          + intros; apply concat_Vp. }
-      rewrite X. reflexivity.
+      apply equiv_path.
+      unfold E', POCase_P, PO_rec.
+      f_ap. srapply path_cocone.
+      + intros [[]|[]] y; cbn.
+        1: apply path_universe_uncurried; apply g0.
+        all: reflexivity.
+      + intros [[]|[]] [[]|[]] []; cbn.
+        * intro y. simpl.
+          rhs nrapply concat_1p.
+          unfold path_universe.
+          lhs nrapply (ap (fun x => x @ _) _^).
+          1: nrapply path_universe_V_uncurried.
+          exact (path_universe_compose (f0 y)^-1 (g0 y))^. 
+        * intros; apply concat_Vp.
   Defined.
 
 End POCase.
--- a/theories/Colimits/PushoutFlattening.v
+++ b/theories/Colimits/PushoutFlattening.v
@@ -18,81 +18,6 @@ Section Flattening.
     exact _.
   Defined.
 
-  Lemma flattening_coh
-    : forall x, pushout_flattening_fam x <~> POCase_P A0 B0 C0 f0 g0 (equiv_pushout_PO x).
-  Proof.
-    (** Proving that the type families are equivalent takes some work but is mostly trivial. We don't know how to transport over families of equivalences yet, so we use univalence to turn it into a transport over a family of paths. *)
-    snrapply Pushout_ind.
-    1,2: hnf; reflexivity.
-    intros a.
-    hnf.
-    lhs nrapply transport_idmap_ap.
-    unshelve erewrite ap_homotopic.
-    2: {
-      intros w.
-      rapply path_universe_uncurried.
-      apply equiv_path_universe. }
-    rewrite 2 transport_pp.
-    rewrite transport_path_universe_uncurried.
-    rewrite transport_path_universe_V.
-    apply moveR_equiv_V.
-    change (?x = path_universe_uncurried 1) with (x = path_universe equiv_idmap).
-    rewrite path_universe_1.
-    lhs nrapply (transport_idmap_ap _ _ _)^.
-    rewrite transport_paths_FlFr.
-    rewrite Pushout_rec_beta_pglue.
-    rewrite concat_pp_p.
-    apply moveR_Vp.
-    rewrite concat_p1.
-    rewrite ap_compose.
-    rewrite Pushout_rec_beta_pglue.
-    unfold popp'.
-    simpl.
-    rewrite 2 concat_1p.
-    change(path_universe 1%equiv @
-      ap (Colimit_Pushout_Flattening.POCase_P A0 B0 C0 f0 g0)
-      (popp a) = path_universe (g0 a oE (f0 a)^-1)).
-    rewrite PO_rec_beta_pp.
-    rewrite path_universe_1.
-    rewrite concat_1p.
-    reflexivity.
-  Defined.
-
-  Lemma flattening_coh_2
-    : forall x, pushout_flattening_fam x <~> POCase_P A0 B0 C0 f0 g0 (equiv_pushout_PO x).
-  Proof.
-    intros x.
-    snrapply Build_Equiv.
-    - revert x.
-      snrapply Pushout_ind.
-      1,2: intros x; exact idmap.
-      intros x.
-      rapply dpath_arrow.
-      intros y.
-      lhs nrapply transport_compose.
-      lhs nrapply transport2.
-      1: apply Pushout_rec_beta_pglue.
-      unfold popp'.
-      simpl.
-      rewrite 2 concat_1p.
-      lhs nrapply (transport_compose idmap (POCase_P A0 B0 C0 f0 g0) (popp x)).
-      lhs nrapply ap011.
-      1: reflexivity.
-      1: nrapply PO_rec_beta_pp.
-      rhs nrapply (transport_idmap_ap _ (pglue x) _).
-      lhs nrapply transport_path_universe_uncurried.
-      rhs nrapply ap011.
-      2: reflexivity.
-      2: apply Pushout_rec_beta_pglue.
-      rhs nrapply transport_path_universe_uncurried.
-      reflexivity.
-    - revert x.
-      snrapply Pushout_ind.
-      1,2: intro; exact (isequiv_idmap _).
-      intros a.
-      apply path_ishprop.
-  Defined.
-
   Definition pushout_flattening
     : Pushout (functor_sigma f f0) (functor_sigma g g0)
       <~> exists x, pushout_flattening_fam x.
@@ -102,7 +27,9 @@ Section Flattening.
     symmetry.
     snrapply equiv_functor_sigma'.
     1: apply equiv_pushout_PO.
-    apply flattening_coh_2.
+    intro x.
+    apply equiv_path.
+    nrapply Pushout_rec_PO_rec.
   Defined.
 
 End Flattening.
--- a/theories/Colimits/PushoutFlattening.v
+++ b/theories/Colimits/PushoutFlattening.v
@@ -2,7 +2,7 @@ Require Import Basics Types Colimits.Pushout Colimits.Colimit_Pushout Colimits.C
 
 (** * Flattening for pushouts *)
 
-(** We derive flattening for pushouts using the flattening lemma for colimits. Most of the work has already been done. What is left is to transport the result along the appropriate equivalences. *)
+(** We derive flattening for pushouts using the flattening lemma for colimits. Most of the work has already been done in Colimits/Colimit_Pushout_Flattening.v. The pushout there is defined as a colimit of a span whereas the pushout we use elsewhere is defined as a coequalizer. What is left is to transport the result along the appropriate equivalences. *)
 
 Section Flattening.
   Context `{Univalence} {A B C} {f : A -> B} {g : A -> C}
--- a/theories/Colimits/PushoutFlattening.v
+++ b/theories/Colimits/PushoutFlattening.v
@@ -1,0 +1,67 @@
+Require Import Basics Types Colimits.Pushout Colimits.Colimit_Pushout Colimits.Colimit_Pushout_Flattening.
+
+(** * Flattening for pushouts *)
+
+(** We derrive flattening for pushouts using the flattening lemma for colimits. Most of the work has already been done, what is left is to transport the result along the appropriate equivalences. *)
+
+Section Flattening.
+  Context `{Univalence} {A B C} {f : A -> B} {g : A -> C}.
+
+  Context (A0 : A -> Type) (B0 : B -> Type) (C0 : C -> Type)
+          (f0 : forall x, A0 x <~> B0 (f x)) (g0 : forall x, A0 x <~> C0 (g x)).
+
+  Definition POCase_P : Pushout f g -> Type.
+  Proof.
+    nrefine (Pushout_rec Type B0 C0 _).
+    cbn; intro x.
+    snrapply path_universe.
+    1: exact ((g0 x) oE (f0 x)^-1%equiv).
+    exact _.
+  Defined.
+
+  Definition PO_flattening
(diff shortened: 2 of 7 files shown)


theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,30 @@ Proof.
   nrefine (pequiv_O_inverts k (loop_susp_unit X)).
   rapply freudenthal_hspace.
 Defined.
+
+(** *** Total space of the Hopf fibration *)
```
Alizter, line 109:
```suggestion
(** *** Total space of the Hopf construction *)
```

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,30 @@ Proof.
...
+    (fun x => Build_Equiv _ _ (fun y => sg_op x y) (H1 x)))^-1%equiv).
+  snrapply equiv_pushout.
+  - refine (equiv_sigma_prod0 _ _ oE _ oE equiv_sigma_symm0 _ _).
+    snrapply equiv_functor_sigma_id.
+    intros x.
+    snrapply Build_Equiv.
+    + exact (.* x).
+    + exact _.
+  - rapply equiv_sigma_unit_ind.
+  - rapply equiv_sigma_unit_ind.
```
jarlg, line 128: I think `equiv_contr_sigma` should be used here, and that `equiv_sigma_unit_ind` can be removed.

theories/Homotopy/Hopf.v
```diff
@@ -21,7 +22,8 @@ Definition hopf_construction `{Univalence} (X : pType)
 Proof.
   srapply Build_pFam.
   - apply (Susp_rec (Y:=Type) X X).
-    exact (fun x => path_universe (x *.)).
+    (** In order to use the flattening lemma for colimits to show that the total space is a join, we need for this equivalence to be a composition with the inverted identity equivalence so that the fiber is definitionally equivalent to the flattening lemma sigma type. This doesn't change anything elsewhere, but saves us having to rewrite an IsEquiv witness. *)
+    exact (fun x => path_universe ((x *.) o equiv_idmap^-1)).
```
jdchristensen, line 26: `equiv_idmap^-1` is definitionally `idmap`, and composition with the identity map is definitionally the other map, so maybe this isn't needed?

theories/Colimits/PushoutFlattening.v
```diff
@@ -0,0 +1,73 @@
+Require Import Basics Types Colimits.Pushout Colimits.Colimit_Pushout Colimits.Colimit_Pushout_Flattening.
+
+(** * Flattening for pushouts *)
+
+(** We derrive flattening for pushouts using the flattening lemma for colimits. Most of the work has already been done, what is left is to transport the result along the appropriate equivalences. *)
```
jdchristensen, line 5:
```suggestion
(** We derive flattening for pushouts using the flattening lemma for colimits. Most of the work has already been done. What is left is to transport the result along the appropriate equivalences. *)
```

jdchristensen (commented):
I'll take a closer look later.

jdchristensen:
There's some relevant discussion about different flattening lemmas here:  https://github.com/HoTT/Coq-HoTT/issues/1833

It would be nice if some redundancies could be removed, but that can be future work.

jdchristensen:
I think the top of PushoutFlattening.v and Colimit_Pushout_Flattening.v should have comments referring to the other file, and explaining what is different between them:  the first is for a pushout defined directly as a coequalizer, and the second is for a pushout defined as a special case of Colimit (which is itself defined as a coequalizer).

BTW, do the problems with computability go right back to Colimit_Flattening.v, or do they arise from something in this PR?

theories/Homotopy/Hopf.v
Alizter: I'll try it out, I wasn't sure at the time and just wanted to be literally what it needs to be.

Alizter:
> BTW, do the problems with computability go right back to Colimit_Flattening.v, or do they arise from something in this PR?

@jdchristensen I think it stems from this PR, though I haven't really computed with the other flattening lemma yet. Upon inspection, I don't see any reason why it should struggle, although there are a lot more terms to unfold compared to proving it directly.

> There's some relevant discussion about different flattening lemmas here: #1833
> 
> It would be nice if some redundancies could be removed, but that can be future work.

I initially experimented with deriving the flattening lemma for coequalizers using the statement of one for GraphQuotient but it wasn't entirely obvious how to proceed. The main issue is adapting the family being flattened as even very similar types have long arguments to show that their families using univalence agree.

theories/Colimits/PushoutFlattening.v
```diff
@@ -0,0 +1,73 @@
...
+    cbn; intro x.
+    snrapply path_universe.
+    1: exact ((g0 x) o (f0 x)^-1).
+    exact _.
+  Defined.
+
+  Lemma flattening_coh
+    : forall x, pushout_flattening_fam x <~> POCase_P A0 B0 C0 f0 g0 (equiv_pushout_PO x).
+  Proof.
+    (** Proving that the type families are equivalent takes some work but is mostly trivial. We don't know how to transport over families of equivlences yet, so we use univalence to turn it into a transportation over a family of paths. *)
```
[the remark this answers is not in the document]
Alizter: As a side question, do you think we can characterise transport over a family of equivalences without univalence?

Alizter:
OK the main performance problems come from Colimit_Pushout_Flattening.PO_flattening I've managed to make a shorter proof of PushoutFlattening.pushout_coh that seems to compute ok. There is basically a very large path induction in PO_flattening which is never going to compute in any reasonable case due to the sheer size of the terms involved. I'll see if I can tweak this to be more direct.

theories/Colimits/PushoutFlattening.v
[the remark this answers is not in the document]
jdchristensen: @Alizter Can you tell me the type of the lemma you are thinking of?  BTW, it's probably a good idea to factor this out.  It might reduce the size of the term and improve performance.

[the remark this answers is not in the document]
Alizter: I'm thinking about `transport (fun x => A x <~> B x) p y`. Via univalence this should be similar to `transport_paths` though I am not sure if univalence is necessary to show this.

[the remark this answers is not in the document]
Alizter: Also I don't think this will be drastic at improving performance. In my experiments, it turns out just getting the flattening lemma for colimits to compute on a constructor is a non-trivial task for Coq and I quickly run out of memory. I swapped the direction of the Hopf equivalence so that we check base points on the inverse. This allowed some big maps to stay uninverted and lands us to comparing pairs of a sigma type. The flattening_coh falls into a secondary component which is good, but even just comparing the primary ones needs Coq to unfold the flattening lemma and we get stuck.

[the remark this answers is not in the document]
jdchristensen: > I'm thinking about `transport (fun x => A x <~> B x) p y`.

Various results of that form can be proved just using path induction.  Only one of them uses Funext.  I was surprised about the first one, which proves that two functions are equal without using Funext.

```coq
(** This is an improvement to [transport_arrow], which only shows that the functions are homotopic.  Even without funext, we can prove that these functions are equal. *)
Definition transport_arrow' {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 -> C x1)
  : transport (fun x => B x -> C x) p f = transport _ p o f o transport _ p^.
Proof.
  destruct p; auto.
Defined.

(** This is like [transport_arrow], but for a family of equivalence types. It just shows that the functions are homotopic. *)
Definition transport_equiv {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1) (y : B x2)
  : (transport (fun x => B x <~> C x) p f) y  =  p # (f (p^ # y)).
Proof.
  destruct p; auto.
Defined.

(** A version that shows that the underlying functions are equal. *)
Definition transport_equiv' {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^) :> (B x2 -> C x2).
Proof.
  destruct p; auto.
Defined.

(** A version that shows that the equivalences are equal.  Here we do need [Funext], for [path_equiv]. *)
Definition transport_equiv'' `{Funext} {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^).
Proof.
  apply path_equiv.
  destruct p; auto.
Defined.
```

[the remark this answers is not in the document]
Alizter: Yes, I've also noticed the first one before. I think because we typically use funext straight after it gets merged. I'll see about putting these in the library and whether it ends up being useful in the proofs.

jdchristensen:
If you add these two results to the end of Colimit_Pushout.v (still inside the Section):
```coq
  Definition equiv_pushout_PO_beta_pglue (a : A)
    : ap equiv_pushout_PO (pglue a) = popp a.
  Proof.
    cbn.
    refine (_ @ _).
    1: nrapply Pushout_rec_beta_pglue.
    unfold popp'; cbn.
    rewrite 2 concat_1p.
    reflexivity.
  Defined.

  Definition Pushout_rec_PO_rec (P : Type) (pushb : B -> P) (pushc : C -> P)
    (pusha : forall a : A, pushb (f a) = pushc (g a))
    : Pushout_rec P pushb pushc pusha == PO_rec P pushb pushc pusha o equiv_pushout_PO.
  Proof.
    snrapply Pushout_ind.
    1, 2: reflexivity.
    intro a; cbn beta.
    nrapply transport_paths_FlFr'; apply equiv_p1_1q.
    lhs exact (Pushout_rec_beta_pglue P pushb pushc pusha a).
    symmetry.
    lhs nrapply (ap_compose equiv_pushout_PO _ (pglue a)).
    lhs nrapply (ap _ (equiv_pushout_PO_beta_pglue a)).
    nrapply PO_rec_beta_pp.
  Defined.
```
then the proof of `flattening_coh` is easy:
```coq
  Lemma flattening_coh
    : forall x, pushout_flattening_fam x <~> POCase_P A0 B0 C0 f0 g0 (equiv_pushout_PO x).
  Proof.
    intro x.
    apply equiv_path.
    nrapply Pushout_rec_PO_rec.
  Defined.
```

jdchristensen:
The two new results don't require Univalence.  They require Funext, but only because `equiv_pushout_PO` does.  And we're only using the map underlying this equivalence, which doesn't need Funext.  So it would be possible to work with this underlying map instead and avoid Funext if you want.

Alizter:
@jdchristensen I've just pushed my changes from earlier today. I haven't had time to address your later comments just yet.

Alizter:
@jdchristensen Thanks for your suggestion, it seems to work well! I've removed the old coherence results accordingly now as the new proof is much nicer.

theories/Colimits/PushoutFlattening.v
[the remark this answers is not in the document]
Alizter: I've made an issue for adding those lemmas, it seems they won't be needed just yet.

theories/Colimits/Colimit_Pushout_Flattening.v
```diff
@@ -49,31 +49,46 @@ Section POCase.
...
       - intros [|[]]; cbn. all: reflexivity.
       - intros [[]|[]] [[]|[]] [] x; cbn in *.
-        all: reflexivity. }
-    rewrite X; clear X.
+        all: reflexivity. } 
     transitivity (exists x, E' (span f g) POCase_E POCase_HE x).
     - apply flattening_lemma.
     - apply equiv_functor_sigma_id.
       intro x.
-      assert (E' (span f g) POCase_E POCase_HE x = POCase_P x). {
```
jdchristensen, line 64: Could the problem with computation have been because `assert` was used instead of `transparent assert`?  In fact, in this case all you have to do in replace the `assert` line with `apply equiv_path.`, and then the existing part in `{ ... }` finishes the proof.  Since that is shorter, maybe it's worth seeing if it behaves as well as your longer proof?

Alizter: I don't think assert makes the proof opaque. Only something like `abstract` would have such an effect. Using assert is the same as `pose proof` or `pose` followed by `clearbody` AFAIK. You would probably see this when doing `Show Proof`.

Alizter: It didn't make much difference. I'm still having a lot of trouble unfolding `unfold Colimit_Pushout_Flattening.PO_flattening.`.

theories/Homotopy/Hopf.v
```diff
@@ -21,7 +22,8 @@ Definition hopf_construction `{Univalence} (X : pType)
 Proof.
   srapply Build_pFam.
   - apply (Susp_rec (Y:=Type) X X).
-    exact (fun x => path_universe (x *.)).
+    (** In order to use the flattening lemma for colimits to show that the total space is a join, we need for this equivalence to be a composition with the inverted identity equivalence so that the fiber is definitionally equivalent to the flattening lemma sigma type. This doesn't change anything elsewhere, but saves us having to rewrite an IsEquiv witness. *)
+    exact (fun x => path_universe ((x *.) o equiv_idmap)).
```
jdchristensen, line 26: I now see why this helps later.  It would be nice if we could instead account for this later, but it doesn't look simple to do, so I'm ok with doing it here.  But then I think we should add "^-1%equiv" to make it match what really arises.  (Maybe it also helps to change "o" to "oE"?)

(For the record, the issue is that the proof that the composite is an equivalence is different.)

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,27 @@ Proof.
...
+
+(** The total space of the Hopf construction on [Susp X] is the join of [X] with itself. Note that we need both left and right multiplication to be equivalences. This is true when [X] is a 0-connected H-space for example. *)
+(* TODO: Show that this is a pointed equivalence. We cannot yet do this as we cannot compute with the flattening lemma due to the massive size of the proof. *)
+Definition equiv_hopf_total_join `{Univalence} (X : pType)
+  `{IsHSpace X} `{forall a, IsEquiv (a *.)} `{forall a, IsEquiv (.* a)}
+  : psigma (hopf_construction X) <~> pjoin X X.
+Proof.
+  snrefine (_ oE (pushout_flattening (f:=const_tt X) (g:=const_tt X) _
+    (Unit_ind (pointed_type X)) (Unit_ind (pointed_type X)) (fun _ => equiv_idmap)
+    (fun x => Build_Equiv _ _ (fun y => sg_op x y) (H1 x)))^-1%equiv).
```
jdchristensen, line 119:
```suggestion
    (fun x => Build_Equiv _ _ (x *.) (H1 x)))^-1%equiv).
```

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,27 @@ Proof.
...
+  snrapply equiv_pushout.
+  - refine (equiv_sigma_prod0 _ _ oE _ oE equiv_sigma_symm0 _ _).
```
jdchristensen, line 121:
```suggestion
  - cbn. refine (equiv_sigma_prod0 _ _ oE _ oE equiv_sigma_symm0 _ _).
```

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,27 @@ Proof.
...
+    snrapply equiv_functor_sigma_id.
+    intros x.
+    snrapply Build_Equiv.
+    + exact (.* x).
+    + exact _.
```
jdchristensen, line 126:
```suggestion
    exact (Build_Equiv _ _ (.* x) _).
```

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,27 @@ Proof.
...
+  - exact (equiv_contr_sigma (Unit_ind (pointed_type X))).
+  - exact (equiv_contr_sigma (Unit_ind (pointed_type X))).
+  - hnf; reflexivity.
+  - hnf; reflexivity.
```
jdchristensen, line 130:
```suggestion
  - reflexivity.
  - reflexivity.
```

theories/Colimits/Colimit_Pushout_Flattening.v
jdchristensen: `assert` is definitely opaque.  That's why we have our own `transparent assert` in the library.  Unless there's a particular reason you like your proof better, I think the original proof (using `equiv_path` instead of `assert`) might be better.

theories/Homotopy/Hopf.v
jdchristensen: (Or, if you want `hnf`, then use a period instead of a semicolon.  Or is it there because it speeds something up?)

theories/Colimits/Colimit_Pushout_Flattening.v
Alizter: I mean that `assert` is opaque when you are in proof mode, but the term that is constructed at the end is still transparent. I've gone ahead and swapped all asserts and rewrites with the better alternatives anyway.

theories/Homotopy/Hopf.v
Alizter: `reflexivity` doesn't seem to unfold those terms and `hnf` is a cheap way to expose the head for it.

Alizter: Or so I thought. It seems to work now, but I was under the impression it didn't work before. Weird.

theories/Homotopy/Hopf.v
Alizter: I can't change `(x *.)` to a bundled equivalence so `oE` doesn't seem to be an option. Unless I pull out the bundled version which is a little verbose.

Alizter:
I've added some comments about the differences between the two flattening lemmas and addressed some of Dan's recent comments.

jdchristensen:
About computing with PO_flattening, I think we'd like a lemma of the form
```coq
  Definition PO_flattening_beta_pol
    : PO_flattening o pol == functor_sigma pol (fun _ => idmap).
```
This typechecks, and I was hoping it was definitionally true.  I tried with one of the proofs of PO_flattening, and it wasn't definitional, but maybe with a different proof it will be.  Certainly there's no fundamental reason why the underlying map of PO_flattening isn't the obvious thing, so I'm a bit puzzled here.

theories/Colimits/Colimit_Pushout_Flattening.v
```diff
@@ -10,7 +10,7 @@ Require Import Colimits.Colimit_Flattening.
 
 (** * Pushout case *)
 
-(** We show the flattening lemma in the case of the pushout. *)
+(** We show the flattening lemma in the case of the pushout. This pushout is defined as the colimit of a span and is not the pushout that apepars elsehwere in the library. The flattening lemma here however can be used to derive the flattening lemma there and this is done in Colimits/PushoutFlattening.v. *)
```
jdchristensen, line 13:
```suggestion
(** We show the flattening lemma in the case of the pushout. This pushout is defined as the colimit of a span and is not the pushout that appears elsehwere in the library. The flattening lemma here however can be used to derive the flattening lemma there and this is done in Colimits/PushoutFlattening.v. *)
```

theories/Homotopy/Hopf.v
```diff
@@ -103,3 +105,25 @@ Proof.
   nrefine (pequiv_O_inverts k (loop_susp_unit X)).
   rapply freudenthal_hspace.
 Defined.
+
+(** *** Total space of the Hopf construction *)
```
jdchristensen, line 109: Maybe this should go right after the definition of the Hopf construction?

theories/Colimits/Colimit_Pushout_Flattening.v
jdchristensen: Oh, right, so that's not the issue.  Anyways, like I wrote elsewhere, it should be possible to have a proof of PO_flattening that computes well...

theories/Colimits/Colimit_Pushout_Flattening.v
Alizter: Sorry, fixed.

theories/Homotopy/Hopf.v
Alizter: Done.

Alizter: I also updated some of the comments.

Alizter:
@jdchristensen It seems that there are various heavy terms in PO_flattening that cause Coq to struggle when unfolding. I agree that having a computation lemma would be good. I believe the obstruction to that falling out definitionally however stems from Colimit_Flattening.v.

By the way, I've added the Hopf entries to HoTTBook.v.

Alizter:
Another useful consequence of this equivalence computing is that we can describe the Hopf map directly from the join to the suspension (without univalence or H-space multiplication being an equivalence) and show it is homotopic to the projection composed with the equivalence (also pointed). This would also let us show the fiber of that map is X.

jdchristensen (approved):
This looks great!  I think it's good to merge this now, and to think about the computation issues in a future PR.

PR #1845, opened 2024-02-12
pinch map of suspension

Alizter:
We define the pinch map of a suspension and show that it computes accordingly with the wedge projections.


theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+Proof.
+  refine (ap_compose (fmap psusp _) _ _ @ _).
+  refine (ap (ap _) (Susp_rec_beta_merid _) @ _).
+  rapply Susp_rec_beta_merid.
+Defined.
+
+(** Doing wedge projections on the pinch map gives the identity. *)
+Definition wedge_pr1_psusp_pinch {X}
+  : wedge_pr1 $o psusp_pinch X ==* Id (psusp X).
+Proof.
```
jdchristensen, line 218: There are various was to reorganize the long proofs below, but I think the way it is might be simplest.  One idea is to separately prove that `wedge_pr1 o wedge_inl` is the identity and `wedge_pr1 o wedge_inr` is constant (the reverse for `pr2`), and then use those facts in the proof.  It probably won't make things shorter, but maybe those facts are handy to know.  Also, the path `wglue @ ap wedge_inr _ @ wglue^` earlier is `fmap loops wedge_inr _`, and using that and functoriality, one can get a more conceptual proof.  Still, it's probably not shorter, so the current way might be best.

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
   - destruct p; exact pmap_idmap.
   - exact pconst.
 Defined.
+
+(** ** The pinch map on the suspension *)
+
+(** Given a suspension, there is a natural map from the suspension to the wedge of the suspension with itself. This is known as the pinch map.
+
+This is the image to keep in mind:
+>>
```
jdchristensen, line 170: I think it should start with `<<` and end with `>>`.

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+   \   |   /                       /|\
+    \  |  /                       / | \
+     \ | /                       /  |  \
+      \|/                       /   |   \
+       *                       *merid(x)^*
+                                \   |   /
+                                 \  |  /
+                                  \ | /
+                                   \|/
+                                    *
```
jdchristensen, line 191: Maybe say that the image is misleading.  Since we aren't dealing with a reduced suspension, we have extra `merid pt` paths.  And we also need to deal with `wglue`.  (By the way, if you wedge the south pole of the left summand with the north pole of the right summand, you can get a cleaner picture that doesn't need `merid pt` and doesn't need `X` to be pointed.  But I think this approach will fit better with what you are doing on co-H-spaces.)

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+Proof.
+  snrapply Build_pHomotopy.
+  - snrapply Susp_ind_FlFr.
+    + reflexivity.
+    + exact (merid pt).
+    + intros x.
+      rhs nrapply concat_1p.
+      rhs nrapply ap_idmap.
+      apply moveR_pM.
+      change (?t = merid x @ (merid pt)^) with (t = loop_susp_unit X x).
```
jdchristensen, line 227:
```suggestion
      change (?t = _) with (t = loop_susp_unit X x).
```

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+      lhs nrapply (ap_compose (psusp_pinch X)).
+      lhs nrapply (ap _ (psusp_pinch_beta_merid x)).
+      lhs nrapply ap_pp.
+      lhs nrapply (ap (fun x => _ @ x) (ap_V _ _)).
+      apply moveR_pV.
+      nrefine (_ @ whiskerL _ (wedge_rec_beta_wglue _ _)^).
```
jdchristensen, line 233:
```suggestion
      rhs nrapply (whiskerL _ (wedge_rec_beta_wglue _ _)).
```

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+      lhs nrapply ap_pp.
+      lhs nrapply (ap (fun x => _ @ x) _).
+      { lhs nrapply (ap_compose _ _ _)^.
```
jdchristensen, line 236: This is one of those spots where the tactic let's you avoid counting underscores:
```suggestion
      { lhs_V nrapply ap_compose.
```

theories/Homotopy/Wedge.v
```diff
@@ -155,3 +161,119 @@ Proof.
...
+        apply ap_const. }
+      lhs nrapply concat_p1.
+      lhs nrapply ap_pp.
+      lhs nrapply (ap (fun x => _ @ x) (wedge_rec_beta_wglue _ _)).
+      f_ap.
+      lhs nrapply (ap_compose wedge_inl _ _)^.
```
jdchristensen, line 242:
```suggestion
      lhs_V nrapply (ap_compose wedge_inl).
```
(Similar things for the second proof as well.)

jdchristensen (approved):
LGTM once the minor changes are done.

jdchristensen:
I also noticed the "necessary" is spelled wrong twice in unchanged comments.  Maybe you can fix those while you are editing this file?

theories/Homotopy/Wedge.v
Alizter: Ah I missed the V versions of those! These are very useful tactics by the way.

theories/Homotopy/Wedge.v
Alizter: I've added those lemmas in but I don't think it really helps in the proofs.

theories/Homotopy/Wedge.v
Alizter: I added a note that the picture isn't literally what is going on. I also noticed that I labelled the bottom suspension as `(merid x)^` wheen it should be `(merid x)`. I agree that the function could be cleaner but as you guessed keeping things pointed helps forget some of these unnecessary details.

Alizter:
@jdchristensen I've addressed your comments now.

theories/Homotopy/Wedge.v
```diff
@@ -51,7 +57,7 @@ Definition wedge_incl {X Y : pType} : X \/ Y -> X * Y
   := pprod_corec _ wedge_pr1 wedge_pr2.
 
 (** 1-universal property of wedge. *)
-(** TODO: remove rewrites. For some reason pelim is not able to immediately abstract the goal so some shuffling around is necessery. *)
+(** TODO: remove rewrites. For some reason pelim is not able to immediately abstract the goal so some shuffling around is neccessery. *)
```
jdchristensen, line 60: Still two typos in that word!  :-)
```suggestion
(** TODO: remove rewrites. For some reason pelim is not able to immediately abstract the goal so some shuffling around is necessary. *)
```

jdchristensen (approved):

theories/Homotopy/Wedge.v
Alizter: Ahh really sorry about this. I've fixed it now.

PR #1846, opened 2024-02-12
put hor \/ notation in module

Alizter:
This notation is a bit too global and conflicts with wedge sum notations we use later. We fix this by putting it behind a module and import it when it is used. (It isn't used enough to justify it being global).

I also don't think it should be in type_scope but I don't really want to spend time thinking about that right now.

### DIFF
--- a/theories/Sets/AC.v
+++ b/theories/Sets/AC.v
@@ -4,6 +4,7 @@ From HoTT Require Import Spaces.Card.
 
 From HoTT.Sets Require Import Ordinals.
 
+Local Open Scope hprop_scope.
 
 (** * Set-theoretic formulation of the axiom of choice (AC) *)
 
--- a/theories/Sets/GCH.v
+++ b/theories/Sets/GCH.v
@@ -2,7 +2,8 @@ From HoTT Require Import TruncType abstract_algebra.
 From HoTT Require Import PropResizing.PropResizing.
 From HoTT Require Import Spaces.Nat.Core Spaces.Card.
 
-Open Scope type.
+Local Open Scope type.
+Local Open Scope hprop_scope.
 
 
 (** * Formulation of GCH *)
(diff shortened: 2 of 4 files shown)


jdchristensen:
If we had an idea which scope it should be in, that might be a better solution, as it is what scopes are designed to handle.  But type_scope does make some sense for both of them, since we use it for binary operations on types...  I don't suppose it would work to use lowercase v (the letter) as the notation for wedges?  That's common in ascii, but IIRC Coq needs there to be a non-letter symbol in a notation.

mikeshulman:
Really?  I thought we defined `'o'` for function composition.

Alizter:
I think we would be able to make a `v` notation but obviously using `v` might be trickier then. What if I create a scope called `hlogic` and stick it in there rather than type_scope?

mikeshulman:
I feel like in an ideal world, a proof assistant ought to be able to support using the same notation `\/` in different places with different meanings, and I think `\/` would be preferable to `v`.  I don't remember exactly how Coq's scopes work, but in principle that seems to me like the right way to handle it.

Alizter:
@mikeshulman That is possible to some extent with the `Arguments` command however sometimes Coq is a little too eagar to insert a coercion from pType to Type which invokes type_scope and forces hor. This caused some statements I wrote about the wedge to not type check correctly when importing Truncations.Core.

mikeshulman:
I guess the mistake is that Coq uses scopes for two different things.  On the one hand you can "open" or "close" a scope to inform it what notations you want to be available manually, but on the other it also automatically invokes certain scopes at certain places in a term.

Alizter:
Its also possible to filter out notations when importing modules in Coq using "import filters". However doing this every time Truncations.Core is used seems like more work than just hiding the notation for hor to begin with.

jdchristensen:
I'm not sure why I had a vague memory about needing a symbol in a notation, but clearly that's wrong.  But, as Ali points out, `v` isn't a great choice because we may use that letter in other ways.

Adding a new scope seems best.  The library creates tons of them in various files.  (54, according to my count.)  So we could make one more. We already have `dprop_scope` and `dhprop_scope`, so we could add `hprop_scope`, for example.

Alizter:
I've made a scope called `hprop_scope` and stuck it in there. This should at least make both notations usable for the time being.

jdchristensen (approved):

PR #1849, opened 2024-02-12
various improvements to wedge

Alizter:
These are some quality of life improvements for working with wedges. They consist of:
- <s>Adding explicit universe levels for wedges. Not sure if they need to be generalised further. I've opted for 2 levels for the source and target. I can't tell if having 3 will be useful at this point.</s> Universe levels are generalized correctly now.
- Generalising wedge_rec to target any type (in any universe) and naming that wedge_rec'. Then redefining wedge_rec as a special case of wedge_rec'.
- Using wild category notations for various arrows and homotopies so that fmap and friends work better.
- Add a computation rule for the wedge inclusion.

### DIFF
--- a/theories/Homotopy/Wedge.v
+++ b/theories/Homotopy/Wedge.v
@@ -8,8 +8,8 @@ Require Import Homotopy.Suspension.
 
 Local Open Scope pointed_scope.
 
-Definition Wedge (X Y : pType) : pType
-  := [Pushout (fun _ : Unit => point X) (fun _ => point Y), pushl (point X)].
+Definition Wedge@{u v} (X Y : pType@{u}) : pType@{v}
+  := [Pushout@{u u v v} (fun _ : Unit => point X) (fun _ => point Y), pushl (point X)].
 
 Notation "X \/ Y" := (Wedge X Y) : pointed_scope.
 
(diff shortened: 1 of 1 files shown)


theories/Homotopy/Wedge.v
```diff
@@ -29,17 +29,26 @@ Proof.
 Defined.
 
 Definition wglue {X Y : pType}
-  : pushl (point X) = (pushr (point Y) : X \/ Y) := pglue tt.
+  : pushl (point X) = (pushr (point Y)) :> (X \/ Y) := pglue tt.
```
Alizter, line 32: This might need #1846

theories/Homotopy/Wedge.v
```diff
@@ -8,8 +8,8 @@ Require Import Homotopy.Suspension.
 
 Local Open Scope pointed_scope.
 
-Definition Wedge (X Y : pType) : pType
-  := [Pushout (fun _ : Unit => point X) (fun _ => point Y), pushl (point X)].
+Definition Wedge@{u v} (X Y : pType@{u}) : pType@{v}
+  := [Pushout@{u u v v} (fun _ : Unit => point X) (fun _ => point Y), pushl (point X)].
```
jdchristensen, line 12: I think it usually works better to let `X` and `Y` live in independent universes.  This makes it easy for Coq to choose the universe variables without inadvertently adding unintended constraints.  In this case, you can achieve it by removing all annotations, except for using `Pushout@{Set _ _ _}`.    (Or you can add `Local Set Universe Minimization ToSet.`)

(Also, I see on the Coq roadmap that support for `max` and `+1` is planned, which will simplify many situations like this, making the output variable unnecessary.)

theories/Homotopy/Wedge.v
```diff
@@ -29,17 +29,26 @@ Proof.
 Defined.
 
 Definition wglue {X Y : pType}
-  : pushl (point X) = (pushr (point Y) : X \/ Y) := pglue tt.
+  : pushl (point X) = (pushr (point Y)) :> (X \/ Y) := pglue tt.
 
-Definition wedge_rec {X Y Z : pType} (f : X $-> Z) (g : Y $-> Z)
+(** Wedge recursion into an unpointed type. We define this one separately as the universe variables need to be more general. *)
+Definition wedge_rec'@{i j v} {X Y : pType@{i}} {Z : Type@{v}}
```
jdchristensen, line 35: I'm guessing you can then just remove most other universe annotations.

jdchristensen (approved):

jdchristensen:
LGTM.

Issue #1841, opened 2024-02-11
Add transport lemmas for families of equivalences


Alizter:
> I'm thinking about `transport (fun x => A x <~> B x) p y`.

Various results of that form can be proved just using path induction.  Only one of them uses Funext.  I was surprised about the first one, which proves that two functions are equal without using Funext.

```coq
(** This is an improvement to [transport_arrow].  That result only shows that the functions are homotopic, but even without funext, we can prove that these functions are equal. *)
Definition transport_arrow' {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 -> C x1)
  : transport (fun x => B x -> C x) p f = transport _ p o f o transport _ p^.
Proof.
  destruct p; auto.
Defined.

(** This is like [transport_arrow], but for a family of equivalence types. It just shows that the functions are homotopic. *)
Definition transport_equiv {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1) (y : B x2)
  : (transport (fun x => B x <~> C x) p f) y  =  p # (f (p^ # y)).
Proof.
  destruct p; auto.
Defined.

(** A version that shows that the underlying functions are equal. *)
Definition transport_equiv' {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^) :> (B x2 -> C x2).
Proof.
  destruct p; auto.
Defined.

(** A version that shows that the equivalences are equal.  Here we do need [Funext], for [path_equiv]. *)
Definition transport_equiv'' `{Funext} {A : Type} {B C : A -> Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^).
Proof.
  apply path_equiv.
  destruct p; auto.
Defined.
```

_Originally posted by @jdchristensen in https://github.com/HoTT/Coq-HoTT/pull/1840#discussion_r1485660385_

PR #1850, opened 2024-02-12
characterize fiber of loop-susp counit

Alizter:
We characterize the fiber of the loop-susp counit. We can't make it a pointed equivalence just yet as we are missing pointedness of the Hopf construction. Interestingly we have to twist the first component when we try to show the sigma types are equivalent.

I've also gone ahead and added the lemmas in #1841 as we use one of these.

### DIFF
--- a/theories/Homotopy/Hopf.v
+++ b/theories/Homotopy/Hopf.v
@@ -146,12 +146,12 @@ Proof.
   funext q.
   simpl.
   lhs rapply (transport_equiv (merid p) _ q).
+  simpl.
   rewrite transport_paths_Fl.
   rewrite ap_V.
   rewrite inv_V.
   rewrite Susp_rec_beta_merid.
   rewrite transport_idmap_ap.
-  simpl.
   rewrite ap_compose.
   rewrite functor_susp_beta_merid.
   rewrite Susp_rec_beta_merid.
--- a/theories/Homotopy/Hopf.v
+++ b/theories/Homotopy/Hopf.v
@@ -1,6 +1,6 @@
 Require Import Types Basics Pointed Truncations.
 Require Import HSpace Suspension ExactSequence HomotopyGroup.
-Require Import WildCat.Core WildCat.Universe  WildCat.Equiv Modalities.ReflectiveSubuniverse Modalities.Descent.
+Require Import WildCat.Core WildCat.Universe WildCat.Equiv Modalities.ReflectiveSubuniverse Modalities.Descent.
 Require Import HSet Spaces.Nat.Core.
 Require Import Homotopy.Join Colimits.Pushout Colimits.PushoutFlattening.
 
@@ -129,16 +129,8 @@ Proof.
   rapply freudenthal_hspace.
 Defined.
 
-(** A version that shows that the underlying functions are equal. *)
-Definition transport_equiv' {A : Type} {B C : A -> Type}
-  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
-  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^) :> (B x2 -> C x2).
-Proof.
-  destruct p; auto.
-Defined.
-
-(** We can use the Hopf construction to show that the fiber of the loop-susp counit is equivalent to the join of the loop spaces. *)
-(* TODO: use explicit path algebra once [equiv_hopf_total_join] is computable and make this poitned. *)
+(** Since [loops X] is an H-space, the Hopf construction provides a map [Join (loops X) (loops X) -> Susp (loops X)].  We show that this map is equivalent to the fiber of [loop_susp_counit X : Susp (loops X) -> X] over the base point, up to the automorphism of [Susp (loops X)] induced by inverting loops. *)
+(* TODO: use explicit path algebra once [equiv_hopf_total_join] is computable and make this pointed. *)
 Definition equiv_pfiber_loops_susp_counit_join `{Univalence} (X : pType)
   : pfiber (loop_susp_counit X) <~> pjoin (loops X) (loops X).
 Proof.
@@ -152,9 +144,8 @@ Proof.
   intros p.
   nrapply path_equiv.
   funext q.
-  unfold emap.
   simpl.
-  lhs rapply transport_equiv.
+  lhs rapply (transport_equiv (merid p) _ q).
   rewrite transport_paths_Fl.
   rewrite ap_V.
   rewrite inv_V.
@@ -165,6 +156,16 @@ Proof.
   rewrite functor_susp_beta_merid.
   rewrite Susp_rec_beta_merid.
   rewrite transport_path_universe.
-  apply moveR_Vp.
-  reflexivity.
+  apply concat_V_pp.
+Defined.
+
+(** As a corollary we get 2n-connectivity of [loop_susp_counit X] for an n-connected [X]. *)
+Global Instance conn_map_loop_susp_counit `{Univalence}
+  {n : trunc_index} (X : pType) `{IsConnected n.+2 X}
+  : IsConnMap (n.+1 +2+ n.+1)%nat (loop_susp_counit X).
+Proof.
+  rapply (conn_point_elim (-1)).
+  1: exact (isconnected_pred_add' n 0 _).
+  nrapply (isconnected_equiv' _ _ (equiv_pfiber_loops_susp_counit_join X)^-1).
+  rapply isconnected_join.
 Defined.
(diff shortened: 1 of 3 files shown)


Alizter:
I've been a bit lazy by using rewrite in the proof however I am not too concerned about it for the time being as it won't compute due to the Hopf equivalence anyway. I'll add a note to use explicit path algebra when that business is sorted out.

theories/Homotopy/Hopf.v
```diff
@@ -1,6 +1,6 @@
 Require Import Types Basics Pointed Truncations.
 Require Import HSpace Suspension ExactSequence HomotopyGroup.
-Require Import WildCat.Core Modalities.ReflectiveSubuniverse Modalities.Descent.
+Require Import WildCat.Core WildCat.Universe  WildCat.Equiv Modalities.ReflectiveSubuniverse Modalities.Descent.
```
jdchristensen, line 3:
```suggestion
Require Import WildCat.Core WildCat.Universe WildCat.Equiv Modalities.ReflectiveSubuniverse Modalities.Descent.
```

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,43 @@ Proof.
...
   rapply freudenthal_hspace.
 Defined.
+
+(** A version that shows that the underlying functions are equal. *)
+Definition transport_equiv' {A : Type} {B C : A -> Type}
+  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
+  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^) :> (B x2 -> C x2).
+Proof.
+  destruct p; auto.
+Defined.
```
jdchristensen, line 138: This is in the other file.
```suggestion
```

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,43 @@ Proof.
...
+
+(** We can use the Hopf construction to show that the fiber of the loop-susp counit is equivalent to the join of the loop spaces. *)
+(* TODO: use explicit path algebra once [equiv_hopf_total_join] is computable and make this poitned. *)
```
jdchristensen, line 141:
```suggestion
(* TODO: use explicit path algebra once [equiv_hopf_total_join] is computable and make this pointed. *)
```

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,43 @@ Proof.
...
+  funext q.
+  unfold emap.
+  simpl.
+  lhs rapply transport_equiv.
+  rewrite transport_paths_Fl.
+  rewrite ap_V.
+  rewrite inv_V.
+  rewrite Susp_rec_beta_merid.
+  rewrite transport_idmap_ap.
+  simpl.
```
jdchristensen, line 163: I would move this `simpl` to right after `lhs rapply transport_equiv.`, since the term it clears appears then.

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,43 @@ Proof.
...
+  rewrite ap_compose.
+  rewrite functor_susp_beta_merid.
+  rewrite Susp_rec_beta_merid.
+  rewrite transport_path_universe.
+  apply moveR_Vp.
+  reflexivity.
```
jdchristensen, line 169:
```suggestion
  apply concat_V_pp.
```

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,43 @@ Proof.
...
+
+(** A version that shows that the underlying functions are equal. *)
+Definition transport_equiv' {A : Type} {B C : A -> Type}
+  {x1 x2 : A} (p : x1 = x2) (f : B x1 <~> C x1)
+  : transport (fun x => B x <~> C x) p f = (equiv_transport _ p) oE f oE (equiv_transport _ p^) :> (B x2 -> C x2).
+Proof.
+  destruct p; auto.
+Defined.
+
+(** We can use the Hopf construction to show that the fiber of the loop-susp counit is equivalent to the join of the loop spaces. *)
```
jdchristensen, line 140: Can you expand on the comment?  Maybe something like:
```coq
(** Since [loops X] is an H-space, the Hopf construction provides a map [Join (loops X) (loops X) -> Susp (loops X)].  We show that this map is equivalent to the fiber of [loop_susp_counit X : Susp (loops X) -> X] over the base point, up to the automorphism of [Susp (loops X)] induced by inverting loops. *)
```

jdchristensen (approved):
This is great to have!  Are you planning on adding a corollary that says that if X is n-connected, then the loop-susp conunit is roughly 2n-connected?

theories/Homotopy/Hopf.v
Alizter: Oops that was left in.

theories/Homotopy/Hopf.v
Alizter: Unfortunately that stops `lhs rapply transport_equiv` from working. Let me see if it just needs some parameters.

Alizter: I was not able to adapt it.

Alizter:
@jdchristensen I can add that corollary too.

theories/Homotopy/Hopf.v
jdchristensen: It works for me:
```coq
Definition equiv_pfiber_loops_susp_counit_join `{Univalence} (X : pType)
  : pfiber (loop_susp_counit X) <~> pjoin (loops X) (loops X).
Proof.
  snrefine (equiv_hopf_total_join (loops X) oE _).
  1: rapply ishspace_loops.
  1,2: exact _.
  snrapply equiv_functor_sigma'.
  1: exact (emap psusp (equiv_path_inverse _ _)).
  snrapply Susp_ind; hnf.
  1,2: reflexivity.
  intros p.
  nrapply path_equiv.
  funext q.
  unfold emap.
  simpl.
  lhs rapply transport_equiv.
  simpl.
  rewrite transport_paths_Fl.
  rewrite ap_V.
  rewrite inv_V.
  rewrite Susp_rec_beta_merid.
  rewrite transport_idmap_ap.
  rewrite ap_compose.
  rewrite functor_susp_beta_merid.
  rewrite Susp_rec_beta_merid.
  rewrite transport_path_universe.
  apply concat_V_pp.
Defined.
```

jdchristensen: I think you tried to move the other `simpl`.

Alizter: Oh yes you're right.

Alizter:
FTR do we know what the fibers and cofibers of the unit and counit are? I'm having trouble finding these results in AT textbooks. Freudenthal obviously says that we have an 2n-connected fiber for the unit, but I'm not sure if this is a smash or a join, I'm a little confused about the connectedness indexing. I don't recall ever coming across those characterisations. I have even less idea about the cofibers.

jdchristensen:
> I wrote down the corollary but I only managed to do it for n.+2 connected types. Do you think it can be strengthened to n.+1?

If X is (-1)-connected, the corollary would say that the fibre is (-2)-connected.  Since every type is (-2)-connected, this is trivially true.  So you can extend it one step down, but you might have to break the proof into two cases.

I don't know the answers to the other questions off the top of my head.

jdchristensen:
BTW, while you are adding those results about transport, here's another one I had lying around:
```coq
(** This is an improvement to [transport_arrow_toconst].  That result shows that the functions are homotopic, but even without funext, we can prove that these functions are equal. *)
Definition transport_arrow_toconst' {A : Type} {B : A -> Type} {C : Type}
  {x1 x2 : A} (p : x1 = x2) (f : B x1 -> C)
  : transport (fun x => B x -> C) p f = f o transport B p^.
Proof.
  destruct p; auto.
Defined.
```
I assume that a similar thing can be done for `fromconst`.

Alizter:
@jdchristensen Thanks that works. I've also added the other transport lemma.

theories/Homotopy/Hopf.v
```diff
@@ -128,3 +128,46 @@ Proof.
...
+Defined.
+
+(** As a corollary we get 2n-connectivity of [loop_susp_counit X] for an n-connected [X]. *)
+Global Instance conn_map_loop_susp_counit `{Univalence}
+  {n : trunc_index} (X : pType) `{IsConnected n.+1 X}
+  : IsConnMap (n +2+ n) (loop_susp_counit X).
+Proof.
+  destruct n.
+  - intro x; hnf; exact _.
+  - rapply (conn_point_elim (-1)).
```
jdchristensen, line 169: Maybe use `nrapply` here to save a tiny bit of time?  Then you need an `exact _` two lines down.

jdchristensen:
This looks good to me.  I restarted one job that failed with a download glitch, and it worked.

PR #1851, opened 2024-02-14
Flattening lemma for GraphQuotient

Alizter:
In this PR we state and prove the flattening lemma for [GraphQuotient] which grants #1833. This is much more direct than other flattening lemmas we have in the library and might lead to better computation.

I'm working on deriving flattening lemmas for coequalizers and pushouts after this. Lot's of exciting stuff after that!

I took the time to explain some of the ideas that go into this proof as it can be quite instructive. Hopefully it reads OK and will be useful to future eyes. I wanted to emphasise how its quite an important consequence of univalence as it makes theory inside a higher topos happen.

@mikeshulman I've also added you as a reviewer so that you can give me feedback about the comments I've written. If you are too busy however that's no problem.

### DIFF
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -175,12 +175,12 @@ Section Flattening.
     apply moveR_Vp.
     rewrite (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
     hott_simpl.
-    rewrite concat_pp_p.
+    lhs nrapply concat_pp_p.
     apply moveR_Mp.
-    rewrite concat_Vp.
+    rhs nrapply concat_Vp.
     apply moveR_pM.
-    rewrite concat_1p.
-    rapply ap_V.
+    rhs nrapply concat_1p.
+    nrapply ap_V.
   Defined.
 
   (** Now that we've shown that [sig DGraphQuotient] acts like a [GraphQuotient] of [sig F] by an appropriate relation, we can use this to prove the flattening lemma. The maps back and forth are very easy so this could almost be a formal consequence of the induction principle. *) 
@@ -209,17 +209,16 @@ Section Flattening.
       simpl.
       lhs nrapply transport_paths_FFlr.
       rewrite GraphQuotient_rec_beta_gqglue.
-      rewrite flatten_rec_beta_gqglue.
-      rewrite concat_p1.
-      apply concat_Vp.
+      refine ((_ @@ 1) @ concat_Vp _).
+      lhs nrapply concat_p1.
+      apply inverse2.
+      nrapply flatten_rec_beta_gqglue.
     - snrapply flatten_ind.
       1: reflexivity.
       intros a b r x.
-      lhs nrapply (transport_paths_FFlr (g := GraphQuotient_rec _ _) _).
+      nrapply (transport_paths_FFlr' (g := GraphQuotient_rec _ _)); apply equiv_p1_1q.
       rewrite flatten_rec_beta_gqglue.
-      rewrite GraphQuotient_rec_beta_gqglue.
-      rewrite concat_p1.
-      apply concat_Vp.
+      exact (GraphQuotient_rec_beta_gqglue _ _ (a; x) (b; e a b r x) (r; 1)).
   Defined.
 
 End Flattening.
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -65,29 +65,21 @@ Proof.
   rapply GraphQuotient_ind_beta_gqglue.
 Defined.
 
-(** ** Flattening lemma *)
+(** ** The flattening lemma *)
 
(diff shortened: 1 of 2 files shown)


jdchristensen:
Glad to see this!  Is it ok if I push a change that mainly updates comments?  There are some that I think add more clutter than help which I'd like to condense or delete, and it's a bit tedious to suggest many changes via the github ui.

Alizter:
@jdchristensen Feel free to push to this PR!

jdchristensen (approved):
Very nice!

jdchristensen:
The failing check just looks like a network glitch again.  I'll restart it when the CI lets me.

jdchristensen:
Now it's an error about missing artifacts.  We can ignore it.

Alizter:
@jdchristensen I've rerun the jobs when it finishes I will merge.

jdchristensen:
Oh, I had one other comment.  Do you think the flattening lemma should be put in a separate file, GraphQuotient_Flattening.v?  This might help with dependencies, so only files that use flattening will need the extra dependencies.

theories/Colimits/GraphQuotient.v
```diff
@@ -58,3 +64,162 @@ Proof.
...
+    : ap (flatten_rec Qgq Qgqglue) (flatten_gqglue s x) = Qgqglue a b s x.
+  Proof.
+    lhs nrapply ap_sig_rec_path_sigma; cbn.
+    rewrite GraphQuotient_ind_beta_gqglue.
+    apply moveR_pM.
+    apply moveL_pM.
+    rewrite 3 concat_pp_p.
+    apply moveR_Vp.
+    rewrite (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
+    hott_simpl.
```
jdchristensen, line 177: Is it hard to replace the `hott_simpl` with something more direct?  It is a bit slow and also makes the Defined slow because it makes the proof term large.

theories/Colimits/GraphQuotient.v
```diff
@@ -58,3 +64,162 @@ Proof.
...
+    rewrite concat_pp_p.
+    apply moveR_Mp.
+    rewrite concat_Vp.
+    apply moveR_pM.
+    rewrite concat_1p.
+    rapply ap_V.
```
jdchristensen, line 183: These rewrites are all easy to remove.
```suggestion
    lhs nrapply concat_pp_p.
    apply moveR_Mp.
    rhs nrapply concat_Vp.
    apply moveR_pM.
    rhs nrapply concat_1p.
    nrapply ap_V.
```

theories/Colimits/GraphQuotient.v
```diff
@@ -58,3 +64,162 @@ Proof.
...
+      rewrite concat_p1.
+      apply concat_Vp.
+    - snrapply flatten_ind.
+      1: reflexivity.
+      intros a b r x.
+      lhs nrapply (transport_paths_FFlr (g := GraphQuotient_rec _ _) _).
+      rewrite flatten_rec_beta_gqglue.
+      rewrite GraphQuotient_rec_beta_gqglue.
+      rewrite concat_p1.
+      apply concat_Vp.
```
jdchristensen, line 222: This can also be sped up a bit:
```suggestion
      refine ((_ @@ 1) @ concat_Vp _).
      lhs nrapply concat_p1.
      apply inverse2.
      nrapply flatten_rec_beta_gqglue.
    - snrapply flatten_ind.
      1: reflexivity.
      intros a b r x.
      nrapply (transport_paths_FFlr' (g := GraphQuotient_rec _ _)); apply equiv_p1_1q.
      rewrite flatten_rec_beta_gqglue.
      exact (GraphQuotient_rec_beta_gqglue _ _ (a; x) (b; e a b r x) (r; 1)).
```

jdchristensen:
> Oh, I had one other comment. Do you think the flattening lemma should be put in a separate file, GraphQuotient_Flattening.v? This might help with dependencies, so only files that use flattening will need the extra dependencies.

BTW, I'm not sure why this folder uses underscores in names.  Besides the Classes library, the only other files with an underscore are two files in HIT.

Alizter:
> > Oh, I had one other comment. Do you think the flattening lemma should be put in a separate file, GraphQuotient_Flattening.v? This might help with dependencies, so only files that use flattening will need the extra dependencies.
> 
> BTW, I'm not sure why this folder uses underscores in names. Besides the Classes library, the only other files with an underscore are two files in HIT.

Underscores are a historical artefact so I generally don't introduce them. There are times when they are needed, i.e. when naming files which have punctuation in their title.

theories/Colimits/GraphQuotient.v
Alizter: I'll see about doing that, just laziness on my part.

Alizter: I'll do it in a follow up PR so I don't have to pull the changes here again.

PR #1853, opened 2024-02-14
transport011: transport over 2 1-paths and lemmas about them

Alizter:
We've probably worked with this in disguise a few times. Having a name and lemmas about it will make reasoning much easier.

Afterwards I'll submit a PR and show how it is useful for deriving functoriality results for GraphQuotient.

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -801,6 +801,47 @@ Proof.
   destruct p, q; reflexivity.
 Defined.
 
+(** Transporting along 2 1-dimensional paths *)
+Definition transport011 {A B} (P : A -> B -> Type) {x1 x2 : A} {y1 y2 : B}
+  (p : x1 = x2) (q : y1 = y2) (z : P x1 y1)
+  : P x2 y2
+  := transport (fun x => P x y2) p (transport (fun y => P x1 y) q z).
+
+Definition transport011_pp {A B} (P : A -> B -> Type) {x1 x2 x3 : A} {y1 y2 y3 : B}
+  (p1 : x1 = x2) (p2 : x2 = x3) (q1 : y1 = y2) (q2 : y2 = y3) (z : P x1 y1)
+  : transport011 P (p1 @ p2) (q1 @ q2) z
+    = transport011 P p2 q2 (transport011 P p1 q1 z).
+Proof.
+  destruct p1, p2, q1, q2; reflexivity.
+Defined.
+
+Definition transport011_compose {A B A' B'} (P : A -> B -> Type) (f : A' -> A) (g : B' -> B)
+  {x1 x2 : A'} {y1 y2 : B'} (p : x1 = x2) (q : y1 = y2) (z : P (f x1) (g y1))
+  : transport011 (fun x y => P (f x) (g y)) p q z
+   = transport011 P (ap f p) (ap g q) z.
+Proof.
+  destruct p, q; reflexivity.
+Defined.
+
+Definition transport011_natsq {A B X} (P : A -> B -> Type) {f : A -> X} {g : B -> X}
+  {a1 a2 : A} {b1 b2 : B} {x1 : P a1 b1} {x2 : P a2 b2} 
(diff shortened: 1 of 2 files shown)


theories/Basics/PathGroupoids.v
```diff
@@ -801,6 +801,47 @@ Proof.
...
+Definition transport011_natsq {A B X} (P : A -> B -> Type) {f : A -> X} {g : B -> X}
+  {a1 a2 : A} {b1 b2 : B} {x1 : P a1 b1} {x2 : P a2 b2} 
+  (h : forall a b, P a b -> f a = g b) (p : a1 = a2) (q : b1 = b2)  
+  (r : transport011 P p q x1 = x2)
+  : ap f p = h a1 b1 x1 @ ap g q @ (h a2 b2 x2)^.
+Proof.
+  destruct p, q, r.
+  apply moveL_pV.
+  apply concat_1p_p1.
+Defined.
```
jdchristensen, line 835: This result looks unusual, and I wouldn't call it `natsq`.  I think it's mixing a couple of things.  It's combining the naturality of the next result (specialized to the type family `f a = g b`) with a computation rule in the case of paths.  I also don't like the contractible pair of `r` and `x2`.  Will it work for you if we replace it with:
```coq
Definition transport011_paths {A B X} (f : A -> X) (g : B -> X)
  {a1 a2 : A} {b1 b2 : B} (p : a1 = a2) (q : b1 = b2)
  (r : f a1 = g b1)
  : transport011 (fun a b => f a = g b) p q r = (ap f p)^ @ r @ ap g q.
Proof.
  destruct p, q; cbn.
  symmetry.
  lhs nrapply concat_p1.
  apply concat_1p.
Defined.
```

theories/Basics/PathGroupoids.v
```diff
@@ -801,6 +801,47 @@ Proof.
...
+
+Definition ap_transport011 {A B} {P : A -> B -> Type} {Q : A -> B -> Type}
+  {x1 x2 : A} {y1 y2 : B} (p : x1 = x2) (q : y1 = y2)
+  (f : forall x y, P x y -> Q x y) (z : P x1 y1)
+  : f x2 y2 (transport011 P p q z) = transport011 Q p q (f x1 y1 z).
+Proof.
+  destruct p, q; reflexivity.
+Defined.
```
jdchristensen, line 843: I'm not sure why `ap_transport` later in the file isn't called `transport_nat`, since it is a naturality result, but given that naming, the naming here makes sense.  I suggest that this move before the result about paths, since it is more general, and that it be polished up a bit, like this:
```coq
(** Naturality of [transport011]. *)
Definition ap_transport011 {A B} {P Q : A -> B -> Type}
  {a1 a2 : A} {b1 b2 : B} (p : a1 = a2) (q : b1 = b2)  
  (f : forall {a b}, P a b -> Q a b) (x : P a1 b1)
  : f (transport011 P p q x) = transport011 Q p q (f x).
Proof.
  destruct p, q; reflexivity.
Defined.
```

jdchristensen (approved):
LGTM after a few fixes.

theories/Basics/PathGroupoids.v
Alizter: For context, this appears when showing GraphQuotient is a 2-functor. The source category I think is the Grothendieck category of relations.

My goals look like:
image

If I destruct `q` then I have an `ap` on a `transport011` so I suppose that is the lemma I need really but I'm not sure. I'll remove this lemma for now and add it back in the PR for the functoriality that way we can suss out the correct lemmas.

theories/Basics/PathGroupoids.v
Alizter: Thanks, I've added that in.

theories/Basics/PathGroupoids.v
jdchristensen: It might be fine as a local lemma that handles what you need at that point.  Or maybe we can break it into conceptual pieces that combine to handle it.

PR #1854, opened 2024-02-14
remove rewrite

Alizter:
@jdchristensen This is a followup to #1851, it removes all the rewrites in the proof. The reason I used rewrite in the first place was because it was convenient and we seldom have to reason about the computation rules themselves (+ laziness).

I rewrote it using explicit path algebra so it can be reasoned about in the future if we ever have to. Another advantage is that Coq won't be slow when it tries to unfold these terms.

### DIFF
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -168,7 +168,10 @@ Section Flattening.
     : ap (flatten_rec Qgq Qgqglue) (flatten_gqglue s x) = Qgqglue a b s x.
   Proof.
     lhs nrapply ap_sig_rec_path_sigma; cbn.
-    rewrite GraphQuotient_ind_beta_gqglue.
+    lhs nrapply (ap (fun x => x @ _)).
+    { nrapply ap.
+      nrapply (ap01 (fun x => ap10 x _)).
+      nrapply GraphQuotient_ind_beta_gqglue. }
     apply moveR_pM.
     apply moveL_pM.
     lhs nrapply concat_pp_p.
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -171,14 +171,27 @@ Section Flattening.
     rewrite GraphQuotient_ind_beta_gqglue.
     apply moveR_pM.
     apply moveL_pM.
-    rewrite 3 concat_pp_p.
+    lhs nrapply concat_pp_p.
+    lhs nrapply concat_pp_p.
+    lhs nrapply concat_pp_p.
     apply moveR_Vp.
-    rewrite (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
-    hott_simpl.
+    lhs nrapply ap.
+    1: nrapply ap.
+    1: nrapply (ap (fun x => x @ _)).
+    1: nrapply (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
+    lhs nrapply ap.
+    1: nrapply ap.
+    1: lhs nrapply concat_pp_p.
+    1: nrapply concat_pp_p.
+    lhs nrapply ap.
+    1: nrapply concat_V_pp.
+    lhs nrapply concat_V_pp.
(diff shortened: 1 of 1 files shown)


theories/Colimits/GraphQuotient.v
```diff
@@ -171,14 +171,27 @@ Section Flattening.
     rewrite GraphQuotient_ind_beta_gqglue.
```
Alizter, line 171: Looks like I missed this one, I'll try to fix.

Alizter:
@jdchristensen Am I remembering correctly that you once wrote a `rewrite` tactic that was able to do these kinds of manipulations? Would it be useful to have in the main library or does it obfuscate the construction of the terms too much?

jdchristensen:
I do have a rewrite tactic that basically tries to generalize one side of an equation and use `ap` to transform it.  The terms are slightly better than what rewrite produces (it's as if a `transport_paths*` lemma has been applied to them), but still not great.  When possible, the newer `lhs` and `rhs` tactics are better.  My tactic is also more fragile than `rewrite`, and probably slower.

BTW, I don't object to using `rewrite` in general.  The only times I push to replace `rewrite` are when (a) it slows things down; (b) it interferes with reasoning about a proof (and note that it doesn't always interfere; sometimes the `internal_paths_rew` terms disappear); (c) other tactics work equally well.  But when changing something deep inside a term, sometimes `rewrite` is the best choice.  It's also sometimes faster than manual path algebra.

Also note that if https://github.com/coq/coq/pull/18299 is merged, then our rewrite terms will at least be expressed in terms of transport.

theories/Colimits/GraphQuotient.v
```diff
@@ -168,17 +168,33 @@ Section Flattening.
     : ap (flatten_rec Qgq Qgqglue) (flatten_gqglue s x) = Qgqglue a b s x.
   Proof.
     lhs nrapply ap_sig_rec_path_sigma; cbn.
-    rewrite GraphQuotient_ind_beta_gqglue.
+    lhs nrapply (ap (fun x => x @ _)).
+    { nrapply ap.
+      nrapply (ap01 (fun x => ap10 x _)).
+      nrapply GraphQuotient_ind_beta_gqglue. }
```
jdchristensen, line 174: This is a case where I would be fine with the `rewrite`, as long as it isn't causing trouble.  But now that you have worked out an alternative, keeping it is probably best for the future.

theories/Colimits/GraphQuotient.v
```diff
@@ -168,17 +168,33 @@ Section Flattening.
...
     apply moveR_pM.
     apply moveL_pM.
-    rewrite 3 concat_pp_p.
+    lhs nrapply concat_pp_p.
+    lhs nrapply concat_pp_p.
+    lhs nrapply concat_pp_p.
```
jdchristensen, line 179:
```suggestion
    do 3 lhs nrapply concat_pp_p.
```

theories/Colimits/GraphQuotient.v
```diff
@@ -168,17 +168,33 @@ Section Flattening.
...
+    lhs nrapply ap.
+    1: nrapply ap.
+    1: nrapply (ap (fun x => x @ _)).
+    1: nrapply (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
+    lhs nrapply ap.
+    1: nrapply ap.
+    1: lhs nrapply concat_pp_p.
+    1: nrapply concat_pp_p.
+    lhs nrapply ap.
+    1: nrapply concat_V_pp.
```
jdchristensen, line 190: Using `lhs nrapply ap` for whiskering confused me, so I suggest this.
```suggestion
    lhs refine (1 @@ (1 @@ (_ @@ 1))).
    1: nrapply (ap10_dpath_arrow DGraphQuotient (fun _ => Q) (gqglue s)).
    lhs refine (1 @@ (1 @@ _)).
    { lhs nrapply concat_pp_p.
      nrapply concat_pp_p. }
    lhs nrapply (1 @@ concat_V_pp _ _).
```

jdchristensen (approved):
Thanks!

PR #1856, opened 2024-02-14
functoriality of GraphQuotient

Alizter:
We prove the standard functoriality results for GraphQuotient. These can be used later to help derive the flattening lemmas for Coeq and Pushout.

### DIFF
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -232,3 +232,129 @@ Section Flattening.
   Defined.
 
 End Flattening.
+
+(** ** Functoriality of graph quotients *)
+
+Lemma functor_gq {A B : Type} (f : A -> B)
+  {R : A -> A -> Type} {S : B -> B -> Type} (e : forall a b, R a b -> S (f a) (f b))
+  : GraphQuotient R -> GraphQuotient S.
+Proof.
+  snrapply GraphQuotient_rec.
+  1: exact (fun x => gq (f x)).
+  intros a b r.
+  apply gqglue.
+  apply e.
+  exact r.
+Defined.
+
+Lemma functor_gq_idmap {A : Type} {R : A -> A -> Type}
+  : functor_gq (A:=A) (B:=A) (S:=R) idmap (fun a b r => r) == idmap.
+Proof.
+  snrapply GraphQuotient_ind.
+  1: reflexivity.
+  intros a b r.
+  lhs nrapply (transport_paths_FlFr (gqglue r)).
+  apply moveR_Mp.
+  rhs nrapply concat_p1.
+  rhs nrapply inv_Vp.
+  rhs nrapply concat_1p.
+  rhs nrapply GraphQuotient_rec_beta_gqglue.
+  apply ap_idmap.
+Defined.
+
+Lemma functor_gq_compose {A B C : Type} (f : A -> B) (g : B -> C)
+  {R : A -> A -> Type} {S : B -> B -> Type} {T : C -> C -> Type}
+  (e : forall a b, R a b -> S (f a) (f b)) (e' : forall a b, S a b -> T (g a) (g b))
+  : functor_gq g e' o (functor_gq f e) == functor_gq (g o f) (fun a b r => e' _ _ (e _ _ r)).
+Proof.
+  snrapply GraphQuotient_ind.
+  1: reflexivity.
+  intros a b r.
+  simpl.
+  lhs nrapply (transport_paths_FlFr (gqglue r)).
+  apply moveR_Mp.
+  rhs nrapply concat_p1.
+  rhs nrapply inv_Vp.
+  rhs nrapply concat_1p.
+  lhs nrapply GraphQuotient_rec_beta_gqglue.
+  rhs nrapply (ap_compose (functor_gq f e) (functor_gq g e') (gqglue r)).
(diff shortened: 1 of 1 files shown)


theories/Colimits/GraphQuotient.v
```diff
@@ -222,3 +222,129 @@ Section Flattening.
...
+  exact r.
+Defined.
+
+Lemma functor_gq_idmap {A : Type} {R : A -> A -> Type}
+  : functor_gq (A:=A) (B:=A) (S:=R) idmap (fun a b r => r) == idmap.
+Proof.
+  snrapply GraphQuotient_ind.
+  1: reflexivity.
+  intros a b r.
+  lhs nrapply (transport_paths_FlFr (gqglue r)).
```
jdchristensen, line 246: You should consider `transport_FlFr'` (and `transport_FFlr'`) which simplify the path algebra slightly.  They can often be combined with `apply equiv_p1_1q` or `apply equiv_1p_q1`.

(This is also one of those places where we need a `Flr` version, to avoid the `ap_idmap` at the end, but I'm not sure how many of these lemmas it's worth generating!)

Alizter: I think its fine to generate those lemmas, but I would perhaps prove them in terms of each other so that we can better understand how they themselves compute. There are a few places where this can be necessary.

theories/Colimits/GraphQuotient.v
```diff
@@ -222,3 +222,129 @@ Section Flattening.
...
+    apply ap.
+    apply p.
+  - intros a b r.
+    simpl.
+    lhs nrapply (transport_paths_FlFr (gqglue r)).  
+    lhs nrapply concat_pp_p.
+    apply moveR_Vp.
+    lhs nrapply ap.
+    1: apply GraphQuotient_rec_beta_gqglue.
+    rhs nrapply (ap (fun x => x @ _)).
```
jdchristensen, line 307: I'd prefer something like `lhs nrefine (1 @@ _).` and `lhs nrefine (_ @@ 1).` for the first and third lines.

theories/Colimits/GraphQuotient.v
Alizter: I've tried using the `'` versions but I don't quite see how it simplifies the path algebra so I have left it.

theories/Colimits/GraphQuotient.v
```diff
@@ -222,3 +222,129 @@ Section Flattening.
...
+    lhs nrapply (transport_paths_FlFr (gqglue r)).  
+    lhs nrapply concat_pp_p.
+    apply moveR_Vp.
+    lhs nrefine (1 @@ _).
+    1: apply GraphQuotient_rec_beta_gqglue.
+    rhs nrefine (_ @@ 1).
+    2: apply GraphQuotient_rec_beta_gqglue.
+    apply moveR_pM.
+    rapply (transport011_helper S).
+    apply q.
```
jdchristensen, line 311: This can be done with the basic lemmas about transport011.  E.g. here is one way (with some rewrites that are easy to remove):
```suggestion
    destruct (q a b r).
    rewrite ap_transport011.
    rewrite transport011_paths.
    lhs nrapply concat_p_pp.
    exact (concat_p_Vp _ _ @@ 1).
```

theories/Colimits/GraphQuotient.v
```diff
@@ -222,3 +222,129 @@ Section Flattening.
...
+  snrapply GraphQuotient_ind.
+  1: reflexivity.
+  intros a b r.
+  lhs nrapply (transport_paths_FlFr (gqglue r)).
+  apply moveR_Mp.
+  rhs nrapply concat_p1.
+  rhs nrapply inv_Vp.
+  rhs nrapply concat_1p.
+  rhs nrapply GraphQuotient_rec_beta_gqglue.
+  apply ap_idmap.
```
jdchristensen, line 252: 7 lines become 4.  Same for most places that FlFr and FFlr are used.
```suggestion
  nrapply (transport_paths_FlFr' (gqglue r)).
  apply equiv_p1_1q.
  rhs nrapply ap_idmap.
  apply GraphQuotient_rec_beta_gqglue.
```

jdchristensen (approved):

theories/Colimits/GraphQuotient.v
Alizter: Oh I see now. I wasn't supplying the path so I was getting confused why it was generating more hyps.

Alizter: Wait actually that wasn't the problem. I was using `lhs`. :facepalm:

Alizter:
@jdchristensen I've simplified some more proofs now.

theories/Colimits/GraphQuotient.v
```diff
@@ -307,18 +287,18 @@ Proof.
...
     apply p.
-  - intros a b r.
-    simpl.
-    lhs nrapply (transport_paths_FlFr (gqglue r)).  
-    lhs nrapply concat_pp_p.
-    apply moveR_Vp.
-    lhs nrefine (1 @@ _).
-    1: apply GraphQuotient_rec_beta_gqglue.
-    rhs nrefine (_ @@ 1).
+  - intros a b s.
```
Alizter, line 290: Seems that this didn't actually simplify anything just shuffled things around. The new proof is not so bad though.

Issue #1855, opened 2024-02-14
Split HIT files into Core and lemmas


Alizter:
We should split the various HIT files like GraphQuotient, Coeq, Pushout into separate files like we do for Join which will make the compilation faster.

              @jdchristensen What about splitting the various HITs into files like we do for Join? That way we can have the core definition and various results like functionality or flattening in separate files? In any case, I would suggest doing the splitting later when we have the flattening lemmas and we can do it all in one go.

_Originally posted by @Alizter in https://github.com/HoTT/Coq-HoTT/issues/1851#issuecomment-1944425556_

Alizter:
@jdchristensen Yes that is what I meant. I'm fine with splitting the flattening lemmas into separate files too. One thing to keep in mind is that the flattening lemams are now much lighter so it might not be as big of an issue as before with the colimits library. I think we will have to see once those PRs are merged.

PR #1857, opened 2024-02-14
flattening lemma for Coeq, Pushout and improvements to Hopf

Alizter:
The first two commits in this PR are about deriving the flattening lemma for Coeq from the flattening lemma for GraphQuotient, and then the flattening lemma for Pushout out of the one for Coeq.

An immediate application is that the flattening lemma is now computable in reasonable time allowing us to improve the equivalence for the total space of the Hopf fibration.

I also delete PushoutFlattening.v as this was really about deriving the flattening lemma from the (slow) colimits library. Now that we have a more direct result, things are simpler.

There is the question of splitting things up as discussed in #1855, however I will leave this to a future PR so this one can more easily be reviewed.


theories/Homotopy/Hopf.v
```diff
@@ -20,31 +20,42 @@ Definition hopf_construction `{Univalence} (X : pType)
...
-  - exact (equiv_contr_sigma (Unit_ind (pointed_type X))).
-  - exact (equiv_contr_sigma (Unit_ind (pointed_type X))).
-  - reflexivity.
-  - reflexivity.
+  snrapply Build_pEquiv'.
+  { etransitivity.
+    1: etransitivity.
+    2: exact (equiv_pushout_flatten (f := const_tt X) (g := const_tt X)
+      (fun _ : Unit => X) (fun _ : Unit => X) (fun x => Build_Equiv _ _ (x *.) (H1 x))).
+    - snrapply equiv_functor_sigma_id.
```
Alizter, line 39: This should a triviality, I'll see if I can make things hold definitionally. Notice that above I removed the superfluous composition which seems to be an artefact of the other flattening lemma.

jdchristensen:
This sounds great!  I'll take a look at it tomorrow, I hope.  You are on a roll, and it's a full time job keeping up!

Alizter:
This is the last of what I have that's finished for now. It will probably be some time before I manage another sprint.

jdchristensen:
Is it ok if I make some commits to simplify a bit of path algebra?

jdchristensen:
Do you think that Colimit_Flattening.v will also follow fairly easily from flattening for coequalizers?

Alizter:
@jdchristensen feel free to push to this PR.

I think flattening for colimits will follow easily too, however I don't have a need for it just yet.

jdchristensen:
Ok, I'm working on this, so don't push anything for a while.

jdchristensen:
@Alizter I just pushed two commits, but I'll have more later, so please don't push anything.

The first commit just contains minor simplifications to the path algebra in Coeq.v.

The second one simplifies things a lot in Pushout.v.  I noticed that if we define Pushout_rec using Coeq_rec (instead of defining it using Pushout_ind which uses Coeq_ind), the first equivalence you had becomes a definitional equality.  Then, to make the Hopf proof go through, I needed to make the parallel change to Susp_rec.

jdchristensen:
Ok, I'm done, and this LGTM.  Nice work.  My last commit just adjusts a few comments.

jdchristensen (approved):

Alizter:
Great!

PR #1858, opened 2024-02-15
Miscellaneous things related to recent changes

jdchristensen:
Minor changes.  The commit messages give a good summary.

The commits are independent, but don't interfere, so can be read together or apart.

### DIFF
--- a/theories/Colimits/GraphQuotient.v
+++ b/theories/Colimits/GraphQuotient.v
@@ -303,33 +303,31 @@ Defined.
 
 (** ** Equivalence of graph quotients *)
 
-Lemma equiv_functor_gq {A B : Type} (f : A <~> B)
-  (R : A -> A -> Type) (S : B -> B -> Type) (e : forall a b, R a b <~> S (f a) (f b))
-  : GraphQuotient R <~> GraphQuotient S.
+Global Instance isequiv_functor_gq {A B : Type} (f : A -> B) `{IsEquiv _ _ f}
+  {R : A -> A -> Type} {S : B -> B -> Type} (e : forall a b, R a b -> S (f a) (f b))
+  `{forall a b, IsEquiv (e a b)}
+  : IsEquiv (functor_gq f e).
 Proof.
-  srapply equiv_adjointify.
-  - exact (functor_gq f e).
+  srapply isequiv_adjointify.
   - nrapply (functor_gq f^-1).
-    intros a b r.
-    apply e.
-    exact (transport011 S (eisretr f a)^ (eisretr f b)^ r).
+    intros a b s.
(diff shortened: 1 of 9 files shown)


theories/Colimits/GraphQuotient.v
```diff
@@ -303,33 +303,31 @@ Defined.
...
 Proof.
-  srapply equiv_adjointify.
-  - exact (functor_gq f e).
+  srapply isequiv_adjointify.
   - nrapply (functor_gq f^-1).
-    intros a b r.
-    apply e.
-    exact (transport011 S (eisretr f a)^ (eisretr f b)^ r).
+    intros a b s.
+    apply (e _ _)^-1.
```
Alizter, line 314: This is a funny quirk of `apply` where it can reverse the direction to apply it in reverse. I think it might be due to the symmetric instance but I'm not sure.

theories/Colimits/Pushout.v
```diff
@@ -167,15 +169,15 @@ Proof.
...
-  (i : p $@ (f' $@L t) == (u $@R f) $@ p')
-  (j : q $@ (g' $@L t) == (v $@R g) $@ q')
-  : functor_pushout h k l p q $== functor_pushout h' k' l' p' q'.
+  {A B C : Type} {f : A -> B} {g : A -> C}
+  {A' B' C' : Type} {f' : A' -> B'} {g' : A' -> C'}
+  {h h' : A -> A'} {k k' : B -> B'} {l l' : C -> C'}
+  {p : k o f == f' o h} {q : l o g == g' o h}
+  {p' : k' o f == f' o h'} {q' : l' o g == g' o h'}
+  (t : h == h') (u : k == k') (v : l == l')
+  (i : forall a, p a @ (ap f') (t a) = u (f a) @ p' a)
```
Alizter, line 178: I think I originally did this so we can use square operations but I guess it didn't end up being useful in the end.

Alizter (approved):

theories/Colimits/GraphQuotient.v
jdchristensen: Yeah, I've never figured out why that sometimes works.  Here I had to fill it in because of the other changes (and it's clearer this way in any case).

theories/Colimits/Pushout.v
jdchristensen: Right, since that was the only spot WildCat was used, I thought it was worth avoiding it.  BTW, with these changes the dependencies are fairly light, so I no longer think it's necessary to split out the flattening lemmas.  Pushout.v has a total of 30 dependencies (including transitive dependencies), which is not bad.

PR #1859, opened 2024-02-16
Equivalences for displayed wild categories

gio256:
Changes to WildCat/Displayed.v include:
- One fix -- `dcat_assoc_opp` and `dcat_assoc_opp_strong` both took the type family `D` as an unnecessary argument.
- Monomorphisms and epimorphisms.
- Changes to arguments. All functor methods now expect both the underlying functor and the displayed functor.
- An unrelated proof that was missing (`is1catstrong_sigma`).

The additions in WildCat/DisplayedEquiv.v closely follow [WildCat/Equiv.v](https://github.com/HoTT/Coq-HoTT/blob/3002e0d82dca445ef3072e12c11f8a36936aa3d8/theories/WildCat/Equiv.v) up through univalence.

There is still more work to do to make this useful (namely proving that univalence of the base category and univalence of the displayed category implies univalence of the total category). I'm happy to wait and PR all at once, but it seems simpler to first review everything that follows existing proofs.

### DIFF
--- a/theories/WildCat/Displayed.v
+++ b/theories/WildCat/Displayed.v
@@ -55,6 +55,14 @@ Proof.
   apply DId.
 Defined.
 
+Global Instance reflexive_DHom {A} {D : A -> Type} `{Is01DCat A D} {a : A}
+  : Reflexive (DHom (Id a))
+  := fun a' => DId a'.
+
+Global Instance reflexive_DGpdHom {A} {D : A -> Type} `{Is0DGpd A D} {a : A}
+  : Reflexive (DGpdHom (Id a))
+  := fun a' => DId a'.
+
 (** A displayed 0-functor [F'] over a 0-functor [F] acts on displayed objects and 1-cells and satisfies no axioms. *)
 Class Is0DFunctor {A : Type} {B : Type}
   {DA : A -> Type} `{IsDGraph A DA} {DB : B -> Type} `{IsDGraph B DB}
@@ -105,7 +113,7 @@ Global Existing Instance is0dgpd_dhom.
 Global Existing Instance is0dfunctor_postcomp.
 Global Existing Instance is0dfunctor_precomp.
 
-Definition dcat_assoc_opp {A : Type} (D : A -> Type) `{Is1DCat A D}
+Definition dcat_assoc_opp {A : Type} {D : A -> Type} `{Is1DCat A D}
   {a b c d : A}  {f : a $-> b} {g : b $-> c} {h : c $-> d}
   {a' : D a} {b' : D b} {c' : D c} {d' : D d}
   (f' : DHom f a' b') (g' : DHom g b' c') (h' : DHom h c' d')
@@ -137,6 +145,22 @@ Definition dcat_comp2 {A : Type} {D : A -> Type} `{Is1DCat A D} {a b c : A}
   (p' : DHom p f' g') (q' : DHom q h' k')
   : DHom (p $@@ q) (h' $o' f') (k' $o' g')
   :=  (k' $@L' p') $o' (q' $@R' f').
+
+Notation "q $@@' p" := (dcat_comp2 q p).
+
+(** Monomorphisms and epimorphisms. *)
+
+Definition DMonic {A} {D : A -> Type} `{Is1DCat A D} {b c : A}
+  {f : b $-> c} {mon : Monic f} {b' : D b} {c' : D c} (f' : DHom f b' c')
+  := forall (a : A) (g h : a $-> b) (p : f $o g $== f $o h) (a' : D a)
+      (g' : DHom g a' b') (h' : DHom h a' b'),
+      DGpdHom p (f' $o' g') (f' $o' h') -> DGpdHom (mon a g h p) g' h'.
+
+Definition DEpic {A} {D : A -> Type} `{Is1DCat A D} {a b : A}
+  {f : a $-> b} {epi : Epic f} {a' : D a} {b' : D b} (f' : DHom f a' b')
+  := forall (c : A) (g h : b $-> c) (p : g $o f $== h $o f) (c' : D c)
+      (g' : DHom g b' c') (h' : DHom h b' c'),
+      DGpdHom p (g' $o' f') (h' $o' f') -> DGpdHom (epi c g h p) g' h'.
 
 Global Instance isgraph_sigma {A : Type} (D : A -> Type) `{IsDGraph A D}
   : IsGraph (sig D).
@@ -262,7 +286,7 @@ Global Existing Instance is0dgpd_dhom_strong.
 Global Existing Instance is0dfunctor_postcomp_strong.
 Global Existing Instance is0dfunctor_precomp_strong.
 
-Definition dcat_assoc_opp_strong {A : Type} (D : A -> Type) `{Is1DCat_Strong A D}
+Definition dcat_assoc_opp_strong {A : Type} {D : A -> Type} `{Is1DCat_Strong A D}
   {a b c d : A}  {f : a $-> b} {g : b $-> c} {h : c $-> d}
   {a' : D a} {b' : D b} {c' : D c} {d' : D d}
   (f' : DHom f a' b') (g' : DHom g b' c') (h' : DHom h c' d')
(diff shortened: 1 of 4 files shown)


theories/WildCat/Displayed.v
```diff
@@ -138,6 +146,22 @@ Definition dcat_comp2 {A : Type} {D : A -> Type} `{Is1DCat A D} {a b c : A}
...
 
+Notation "q $@@' p" := (dcat_comp2 q p).
+
+(** Monomorphisms and epimorphisms. *)
+
+Definition DMonic {A} {D : A -> Type} `{Is1DCat A D} {b c : A}
+  {f : b $-> c} {mon : Monic f} {b' : D b} {c' : D c} (f' : DHom f b' c')
+  := forall (a : A) (g h : a $-> b) (p : f $o g $== f $o h) (a' : D a)
+      (g' : DHom g a' b') (h' : DHom h a' b'),
+      DGpdHom p (f' $o' g') (f' $o' h') -> DGpdHom (mon a g h p) g' h'.
```
Alizter, line 157: I think this is the correct notion of displayed monic as it characterises monic maps in the Grothendieck category. However I'm thinking about the perhaps more general situation where we might know `g $== h` in addition, so the base map might need not be monic. I'm not sure this is a useful generalisation however.

Alizter:
I'll do a closer review later today. Thanks for the work!

theories/WildCat/Displayed.v
gio256: I think this is what the 1lab calls [weak monos](https://1lab.dev/Cat.Displayed.Morphism.html#weak-monos), although I don't think I would consider them strictly weaker. I didn't include them as I'm also not sure of their usefulness, but they could be worth adding.

Alizter: I think we shouldn't complicate the theory we have too much until we have a use-case for them. No need to change anything here.

theories/WildCat/DisplayedEquiv.v
```diff
@@ -0,0 +1,504 @@
...
+  : DGpdHom (compose_cate_idr f) (dcate_fun (f' $oE' id_dcate a'))
+    (dcate_fun f').
+Proof.
+  refine (compose_dcate_fun f' _ $@' _ $@' dcat_idr (dcate_fun f')).
+  apply (_ $@L' dcate_buildequiv_fun _).
+Defined.
+
+(** Some more convenient equalities for equivalences. The naming scheme is similar to [PathGroupoids.v].*)
+
+Definition dcompose_V_hh {A} {D : A -> Type} `{DHasEquivs A D}
```
gio256, line 256: The naming here is slightly confusing as we also have e.g. `compose_dcate`, but I'm not sure of the best place to indicate that these operations are displayed. Maybe `composed_V_hh`?

Alizter: I think we should rename compose_dcat to dcompose_dcat and then talk about dcompose lemmas. composed reads differently and I don't think it is a good name.

jdchristensen: Maybe the best convention is to always put the "d" or "D" right at the left of the identifier (and only once), e.g. `dcompose_cate`, `dcompose_V_hh`, etc?

jdchristensen:
Thanks for this PR!  I skimmed over it and think it looks good.  But maybe someone more familiar with displayed categories also should approve?

theories/WildCat/Displayed.v
Alizter: I've opened a discussion over at the 1lab and have asked them what they think about the notion: https://github.com/plt-amy/1lab/discussions/360

theories/WildCat/DisplayedEquiv.v
Alizter: Yes, that seems like a better idea. In that case my renaming suggestion should be `compose_dcat -> dcompose_cat` since there is no dcompose for 1-cats that it could be confused with.

gio256: Would class names become `DIs1Cat` or `IsD1Cat`? I suppose we could also say that `D` is for dependent and `d` is for displayed, `dIs1Cat`, `dHom`, `dis01cat_hom`, `dreflexive_hom`.

Alizter: I think `d` should always be about "displayed" in this sense as the "dependent" versions of categories are just the sections of the displayed categories. I think it would be too confusing to have both names around. For class names, `IsD1Cat` is what I would expect.

jdchristensen: When I said that "d" or "D" should come first, I didn't consider all cases.  I think that "Is" and "Build_" are examples of things that should stay first, and then "d" or "D", and then the rest of the name.  There might be other things that have a stronger "left" priority than "d"/"D" as well.  As for capitalization, I think we agreed that it should match the next letter, e.g. "DHom", "dcompose".  What about when the next character is a number?  Then maybe look to the next letter for capitalization, like "D1Cat"?

Alizter:
Would you be able to create a file called `DUniverse` which has the displayed versions of `WildCat/Universe.v`? We don't have any examples of displayed cats just yet AFAIK so it might be instructive to have a "toy" example too. This example will be quite instructive as we will see the theory in action.

There are a few special cases of interest:
- Sections over the unit type would be the category of pointed types.
- Sections over the empty type would be a singleton, since `Empty -> Type` is unique.
- Sections over `nat` would correspond to some sort of sequence of spaces.
- I believe that the Grothendieck category will be the category of types again. I'm not entirely sure however since I know this doesn't work for spectra, but I guess there we don't have a topos.

Anyway, this doesn't have to be in this PR, but it might help to introduce DUniverse in another PR and perhaps show that it has displayed equivalences etc.

gio256:
I have a toy example with [pType](https://github.com/gio256/Coq-HoTT/blob/040efd94ff484a4c225b922d126060c55fe61475/theories/WildCat/pType.v) but haven't defined any others yet. I will work on Universe and include any other examples that come up.

jdchristensen:
@Alizter Can you explain what you have in mind for the universe?  What category would it be displayed over?  I'm imagining lots of examples of categories displayed over the universe, e.g. pType, Group, n-Type, etc.  And maybe iterates, e.g. AbGroup displayed over Group.

One thing that's still not clear is whether this framework saves us much.  It would be nice to be able to create a wild category really easily by presenting it as a displayed category, but I've seen a draft of pType done this way, and it doesn't seem much shorter than the direct definition.

Alizter:
@jdchristensen For the universe I had in mind Types and fibrations over those types (type families). The family for the displayed cat would be `fun x => x -> Type`. The displayed morphisms would be fiberwise morphsisms etc. IINM this should correspond to fiberwise homotopy theory from "classical" homotopy theory.

Alizter:
Here are some more examples I just thought of:

- There is a displayed category of relations, basically corresponding to the family `fun A => A -> A -> Type`. DMorhpisms in this category would be a function on the underlying types that preserves the relation. The Grothendieck category would be a category of Types with relations. There is a functor from Type to this category which takes a Type and gives it `=` as a relation. There is also a functor from the Grothendieck category back to Type via GraphQuotient. IINM passing from the Grothendieck category to Type and back is homotopic to the identity functor. Whereas the other composition is the reflexive-symmetric-transitive closure of a relation. Having this setup might allow us to easier state various naturality properties about GraphQuotient.
- On Type there is a displayed category on the identity map with morphsisms given by `IsEquiv`. Composition is given by the composition for the witness. Identity is always a given. Since this is a hprop (for Type at anyway), the other coherences are trivial. The grothendieck category should be the category of Types and morhpsisms would be equivalences, so I guess this is the core of Type. We could probably play the same game with any predicate on functions like those from a factorization system. (IsTrunc, IsConnected) etc.
- I can think of a few examples in homological algebra: If Ring is the category of rings, then the category of left modules can be constructed as a displayed category over Ring^op. Sections of this displayed category is the usual R-mod. The change-of-rings functor corresponds to certain maps between sections.
- I haven't had time to think it through, but it might be possible to do some G-equivariant homotopy theory in a displayed setting. This isn't straightforward as I know here it is easy to be naive.

There is some "presheaf" looking stuff going on with displayed categories the more I think of it. For example, the displayed category of fibrations over Type I mentioned before has a section over the unit type. This is sort of like the fact that ooGpd are oo-presheaves over the point. I didn't think it through enough, but I thought I would mention it.

Alizter (approved):
I had a look through and it looks fine to me. I can't remember if we discussed this already, but do you have some specific goals in mind?

gio256:
Sorry to push after reviews, but the proof that the total category is univalent turned out to be surprisingly simple once I factored `cat_equiv_path` correctly.

It would be good to get another set of eyes on the above commit, though.

jdchristensen:
The new commit is very clean.  Nice!

Alizter:
Thanks!

PR #1860, opened 2024-02-19
join construction

Alizter:
Here is a formalization of the basic results in the join construction. We can now construct essentially small homotopy images of maps and define the propositional truncation from simpler HITs.

This was quite tricky to formalize as I had to constantly wrestle with the universe constraints, but I managed to get somewhere in the end.

I've left this as a draft for now as it is missing the theory about image factorizations, embeddings etc. I'm looking to get some feedback on what I have so far before I continue.

### DIFF
--- a/theories/Homotopy/JoinConstruction.v
+++ b/theories/Homotopy/JoinConstruction.v
@@ -1,0 +1,324 @@
+Require Import Basics Types.
+Require Import Limits.Pullback Colimits.Pushout Diagrams.Diagram Diagrams.Sequence Colimits.Colimit Colimits.Sequential.
+Require Import Join.Core.
+Require Import NullHomotopy.
+
+(** * The Join Construction *)
+
+(** ** Propositional Truncation *)
+
+(** Instead of using the propositional truncation defined in Truncations.Core, we instead give a simpler definition here out of simple HITs. This way we can break dependencies and also manage universe levels better. *)
+(** TODO: this should be used in Truncations.Core instead of the other definition. *)
+
+Definition merely@{i j} (A : Type@{i}) : Type@{j}.
+Proof.
+  (** The propositional truncation of a type will be the infinite join power, or the colimit of the sequence of the nth join power. First we define this sequence. *)
+  transparent assert (s : Sequence@{j j j}).
+  { snrapply Build_Sequence.
+    - exact (iterated_join A).
+    - intros n.
+      apply pushr. }
+  (** Then we define the colimit of this sequence. *)
+  exact (Colimit s).
+Defined.
+
+Definition merely_in@{i j} {A : Type@{i}} (x : A) : merely A.
+Proof.
+  snrapply colim.
+  1: exact O.
+  exact x.
+Defined.
+
+(** A sequence of null-homotopic maps has a contractible colimit. This is already proven in Sequential.v but we state the hypotheses a little differently here.  *)
+Lemma contr_seq_colimit_nullhomotopic `{Funext} (s : Sequence) (x : s O)
+  (is_null : forall n : nat, NullHomotopy (@arr (sequence_graph) s n n.+1%nat idpath))
+  : Contr (Colimit s).
+Proof.
+  snrapply contr_colim_seq_into_prop.
+  - intros n.
+    destruct n.
+    + exact x.
+    + exact (is_null n).1.
+  - intros n y.
+    symmetry.
+    exact ((is_null n).2 y).
+Defined.
+
+Definition merely_rec@{i j k} (A : Type@{i}) (P : Type@{j}) `{IsHProp P}
+  : (A -> P) -> merely@{i k} A -> P.
+Proof.
+  intros f.
+  apply Colimit_rec@{i k k k k k k}.
+  snrapply Cocone.Build_Cocone.
+  2: intros ? ? ? ?; nrapply path_ishprop; exact _.
+  simpl.
+  intros n.
+  induction n.
+  1: exact f.
+  snrapply Join_rec.
+  - exact f.
+  - exact IHn.
+  - intros ? ?; nrapply path_ishprop; exact _.
+Defined.
+
+(* TODO: move *)
+Lemma nullhomotopy_joinr (A B : Type) (x : A) : NullHomotopy (@joinr A B).
+Proof.
+  exists (joinl x).
+  intros y.
+  symmetry.
+  apply jglue.
+Defined.
+
+(* TODO: move *)
+Lemma nullhomotopy_joinl (A B : Type) (y : B) : NullHomotopy (@joinl A B).
+Proof.
+  exists (joinr y).
+  intros x.
+  apply jglue.
+Defined.
+
+Global Instance ishprop_merely@{i j} `{Funext} (A : Type@{i})
+  : IsHProp (merely@{i j} A).
+Proof.
+  apply hprop_inhabited_contr.
+  rapply merely_rec.
+  intros x.
+  apply contr_seq_colimit_nullhomotopic.
+  - exact x.
+  - intros m.
+    simpl.
+    apply nullhomotopy_joinr.
+    exact x.
+Defined.
+
+(** We can construct the homotopy image of a map [f : A -> B] using this definition of propositional truncation, which we will later show to be essentially small. *)
+Definition himage@{i j} {A : Type@{i}} {B : Type@{j}} (f : A -> B) : Type@{j}
+  := {y : B & merely@{j j} (hfiber f y)}.
+
+(** ** Essentially Small and Locally Small Types *)
+
+(** A type in a universe [v] is essentially small, with respect to a strictly smaller universe [u], if there is a type in the universe [u] that is equivalent to it. *)
+Definition IsEssentiallySmall@{u v | u < v} (A : Type@{v})
+  := {B : Type@{u} & A <~> B}.
+
+(** A type is locally small if all of its path types are essentially small. *)
+Definition IsLocallySmall@{u v | u < v} (A : Type@{v})
+  := forall x y : A, IsEssentiallySmall@{u v} (x = y).
+
+(** Under univalence, being essentially small is a proposition. *)
+Global Instance ishprop_isessentiallysmall@{u v} `{Univalence} (A : Type@{v})
+  : IsHProp (IsEssentiallySmall@{u v} A).
+Proof.
+  apply hprop_allpath.
+  intros [X e] [X' e'].
+  snrapply path_sigma.
+  - apply path_universe_uncurried.
+    exact (e' oE e^-1).
+  - apply path_equiv.
+    lhs nrapply (transport_equiv' (path_universe_uncurried (e' oE e^-1)) e).
+    funext x; simpl.
+    rewrite transport_const.
+    rewrite transport_path_universe.
+    apply ap, eissect.
+Defined.
+
+(** Therefore, so is being locally small. *)
+Global Instance ishprop_islocallysmall@{u v} `{Univalence} (A : Type@{v})
+  : IsHProp (IsEssentiallySmall@{u v} A) := _.
+
+(** A sigma type is essentially small if both of its types are essentially small. *)
+Definition isessentiallysmall_sigma@{u v k | u <= v, v < k}
+  `{Funext} (A : Type@{u}) (P : A -> Type@{v})
+  (ies_A : IsEssentiallySmall@{u k} A)
+  (ies_P : forall x, IsEssentiallySmall@{v k} (P x))
+  : IsEssentiallySmall@{v k} {x : A & P x}.
+Proof.
+  eexists.
+  nrapply (equiv_functor_sigma'@{u v _ _ k k} ies_A.2).
+  nrapply (equiv_ind@{u v k} ies_A.2^-1%equiv).
+  1: exact _.
+  intros x.
+  nrefine (equiv_path@{v k} _ _ _ oE _).
+  { apply ap.
+    symmetry.
+    apply eisretr. }
+  exact (ies_P ((ies_A.2)^-1%equiv x)).2.
+Defined.
+
+(** Every small type is trivially essentially small *)
+Definition isessentiallysmall_small@{u v} (A : Type@{u})
+  : IsEssentiallySmall@{u v} A.
+Proof.
+  exists A.
+  exact equiv_idmap.
+Defined.
+
+(** The join of two essentially small types is essentially small. *)
+Definition isessentiallysmall_join@{u1 u2 v k} (A : Type@{u1}) (B : Type@{u2})
+  (ies_A : IsEssentiallySmall@{v k} A) (ies_B : IsEssentiallySmall@{v k} B)
+  : IsEssentiallySmall@{v k} (Join@{u1 u2 v} A B).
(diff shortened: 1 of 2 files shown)


theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+  snrapply Cocone.Build_Cocone.
+  2: intros ? ? ? ?; nrapply path_ishprop; exact _.
+  simpl.
+  intros n.
+  induction n.
+  1: exact f.
+  snrapply Join_rec.
+  - exact f.
+  - exact IHn.
+  - intros ? ?; nrapply path_ishprop; exact _.
```
jdchristensen, line 61: This part could also use `equiv_into_hprop` from Join.Core.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+    simpl.
+    apply nullhomotopy_joinr.
+    exact x.
+Defined.
+
+(** We can construct the homotopy image of a map [f : A -> B] using this definition of propositional truncation, which we will later show to be essentially small. *)
+Definition himage@{i j} {A : Type@{i}} {B : Type@{j}} (f : A -> B) : Type@{j}
+  := {y : B & merely@{j j} (hfiber f y)}.
+
+(** ** Essentially Small and Locally Small Types *)
```
jdchristensen, line 99: This might be a good time to merge the file https://github.com/jdchristensen/NonAccessible/blob/main/Smallness.v,
which proves a lot of things about Smallness (but relies on the join construction stated as an axiom).  We might want a folder with (a) a short file with the main definitions, like the beginning of Smallness.v, (b) a file containing this new material, (c) the rest of Smallness.v.  Or maybe some other arrangement.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+
+(** A type in a universe [v] is essentially small, with respect to a strictly smaller universe [u], if there is a type in the universe [u] that is equivalent to it. *)
+Definition IsEssentiallySmall@{u v | u < v} (A : Type@{v})
```
jdchristensen, line 102: I think it's better to not constrain u and v.  There's no semantic reason to assume that A is in a larger universe.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+(** ** Fiber-wise Joins of Maps *)
+
+(** The fiber-wise join of two maps is the generalization of the join of two spaces, which can be thought of as the fiber-wise join of maps [A -> 1] and [B -> 1]. The fiber-wise join of two maps [f : A -> X] and [g : B -> X] is the pushout of the projections of the pullback of [f] and [g]. *)
+Definition FiberwiseJoin@{a b x k}
+  {A : Type@{a}} {B : Type@{b}} (X : Type@{x}) (f : A -> X) (g : B -> X)
+  : Type@{k}.
+Proof.
+  nrapply Pushout@{k k k k}.
+  - exact (pullback_pr1@{_ _ _ k} (f := f) (g := g)).
+  - exact (pullback_pr2@{_ _ _ k}(f := f) (g := g)).
```
jdchristensen, line 222:
```suggestion
  - exact (pullback_pr2@{_ _ _ k} (f := f) (g := g)).
```

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+  - exact x.
+  - intros m.
+    simpl.
+    apply nullhomotopy_joinr.
+    exact x.
+Defined.
+
+(** We can construct the homotopy image of a map [f : A -> B] using this definition of propositional truncation, which we will later show to be essentially small. *)
+Definition himage@{i j} {A : Type@{i}} {B : Type@{j}} (f : A -> B) : Type@{j}
+  := {y : B & merely@{j j} (hfiber f y)}.
```
jdchristensen, line 97: A priori, the image of a map from A : Type@{i} to B : Type@{j} should live in Type@{max(i,j)}.  Here you constrain i <= j, but I don't think that makes sense for what could be a general definition.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+Require Import NullHomotopy.
+
+(** * The Join Construction *)
+
+(** ** Propositional Truncation *)
+
+(** Instead of using the propositional truncation defined in Truncations.Core, we instead give a simpler definition here out of simple HITs. This way we can break dependencies and also manage universe levels better. *)
+(** TODO: this should be used in Truncations.Core instead of the other definition. *)
+
+Definition merely@{i j} (A : Type@{i}) : Type@{j}.
```
jdchristensen, line 13: Shouldn't this just take one universe variable, with the output universe the same as the input universe?

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+  - exact (fiberwise_join_power_map@{u v k} f).
+  - simpl; intros n ? p; destruct p.
+    apply fiberwise_join_power_incl_comm.
+Defined.
+
+(** Here is our main theorem, it says that for any map [f : A -> X] from a small type [A] into a locally small type [X], image is an essentially small type. *)
+Theorem isessentiallysmall_infinite_fiberwise_join_power@{u v k | u <= v, v < k}
+  `{Funext} {A : Type@{u}} {X : Type@{v}} (f : A -> X)
+  (ils_X : IsLocallySmall@{v k} X)
+  : IsEssentiallySmall@{v k} (himage@{u v} f).
```
jdchristensen, line 309: I'm not sure why u and v are independent here.  Shouldn't the constraint on how small A is (u) by the same as the constraint on how small the path types of B are (v)?

theories/Homotopy/JoinConstruction.v
jdchristensen: I'm also not sure why this is needed.  Can't you just use (Tr -1) from Truncations.Core?  It's separately interesting that pushouts and sequential colimits suffice for defining propositional truncation, but for the goal of proving that appropriate images are small, it isn't needed.  I'd be inclined to make this part a separate file, maybe in the metatheory folder.  But I could be convinced otherwise.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+    exact _. }
+  1,2: exact _.
+Defined.
+
+(** ** Fiber-wise Joins of Maps *)
+
+(** The fiber-wise join of two maps is the generalization of the join of two spaces, which can be thought of as the fiber-wise join of maps [A -> 1] and [B -> 1]. The fiber-wise join of two maps [f : A -> X] and [g : B -> X] is the pushout of the projections of the pullback of [f] and [g]. *)
+Definition FiberwiseJoin@{a b x k}
+  {A : Type@{a}} {B : Type@{b}} (X : Type@{x}) (f : A -> X) (g : B -> X)
+  : Type@{k}.
```
jdchristensen, line 218: I had to work a lot with fiberwise joins in work on projective spaces and tangent spaces (no public repo yet), and it was much cleaner there to work with type families instead of maps.  Then the fiberwise join of P : X -> Type and Q : X -> Type is simply the pointwise join:  fun x => Join (P x) (Q x).  This gets rid of the things you had to deal with involving bundling of data.  OTOH, it would mean having to take sigma types at the end.  I think it's worth thinking about, though.

jdchristensen:
It's great to see this!  I've given it a quick look over, and can review again later, maybe after some restructuring.

One meta-comment is that we can't really trust this until the universe issues with GraphQuotient are worked out, since they put the result in the wrong universe, and the colimits defined here use that construction.  I don't think there will be a problem, but technically speaking we can't consider this complete until that is resolved.

Another comment is that the parts of this file that are general facts about joins should probably go into the Join folder, in new or existing files.

theories/Homotopy/JoinConstruction.v
```diff
@@ -0,0 +1,324 @@
...
+(** Therefore, so is being locally small. *)
+Global Instance ishprop_islocallysmall@{u v} `{Univalence} (A : Type@{v})
+  : IsHProp (IsEssentiallySmall@{u v} A) := _.
+
+(** A sigma type is essentially small if both of its types are essentially small. *)
+Definition isessentiallysmall_sigma@{u v k | u <= v, v < k}
+  `{Funext} (A : Type@{u}) (P : A -> Type@{v})
+  (ies_A : IsEssentiallySmall@{u k} A)
+  (ies_P : forall x, IsEssentiallySmall@{v k} (P x))
+  : IsEssentiallySmall@{v k} {x : A & P x}.
```
jdchristensen, line 135: This something wrong here.  You start with A in universe u, and also assume that A is u-small, which is automatic.  Similarly, P lands in universe v, but you also assume that it is v small.  Since u <= v, this result is then a tautology, and can be proved with `exact (_; equiv_idmap).`

theories/Homotopy/JoinConstruction.v
jdchristensen: Also, this suffers from the same problems as the result about sigma types.  It's a tautology and can be proved with `exact (_; equiv_idmap).`.  And if you fix the universe variables, then the proof below won't go through, since it requires X to be small on the second line.

Alizter:
> One meta-comment is that we can't really trust this until the universe issues with GraphQuotient are worked out, since they put the result in the wrong universe, and the colimits defined here use that construction. I don't think there will be a problem, but technically speaking we can't consider this complete until that is resolved.

FTR, when developing this file locally, it is based on top of my other patches, that includes the universe levels for GraphQuotient being fixed. So I am a little bit more certain the universes are correct here.

> Another comment is that the parts of this file that are general facts about joins should probably go into the Join folder, in new or existing files.

Yes, this is something that I need to do. One question I have about organisation is do we want the define the propositional truncation this way? I kept everything in one file to keep it mostly self contained, but ideally we would use the join construction in Seperated.v, I will have to think about how to organize things.

theories/Homotopy/JoinConstruction.v
Alizter: That sounds like a good idea. I didn't want to diverge too much from the setup in Egbert's thesis so I wasn't courageous enough to try defining fiberwise joins too differently. I think this is the better thing to do, as it means we can reuse stuff we have for Joins for the fiberwise version.

theories/Homotopy/JoinConstruction.v
Alizter: I can take the stuff that doesn't rely on the join construction first and rebase the work here on that. I'll do that in another PR.

theories/Homotopy/JoinConstruction.v
Alizter: Yes, that makes more sense.

theories/Homotopy/JoinConstruction.v
Alizter: My original thinking was that I wanted to make propositional truncations definable before we touch ReflectiveSubuniverse/ since Seperated.v would use it. I guess this isn't really necessary in the end. I'll move it to a metatheory file.

Alizter:
It seems that 2 years ago I wrote a file called `Metatheory/PropTrunc.v` with the construction of propositional truncation using the join construction. I'd completely forgotten about that, so I don't need to do that.

jdchristensen:
I've updated https://github.com/jdchristensen/NonAccessible to work with the current library.  Part of the file Smallness.v from that repo has now been merged into Coq-HoTT via #1867 , and I've updated the remaining part of Smallness.v to work with the new naming and other conventions.  The part of Smallness.v that remains is the part that depends on the join construction, so that's why I'm mentioning this in this PR.

Alizter:
I've tried to update this PR using the new smallness stuff, but for whatever reason I cannot get the smallness of joins to typecheck. The statement is:
```coq
Global Instance issmall_join@{u k l big+ | k < big, u <= l, k <= l}
  (A : Type@{k}) (B : Type@{k})
  (sA : IsSmall@{u k} A) (sB : IsSmall@{u k} B)
  : IsSmall@{u k} (Join@{k k k} A B).
Proof.
```
But if you observe the term that fills it:
```coq
  Check( 
fun   (A : Type@{k}) (B : Type@{k})
  (sA : IsSmall@{u k} A) (sB : IsSmall@{u k} B)
=>   
    Build_IsSmall@{u k}
      (Join@{k k k} A B)
      (Join@{u u u} (smalltype@{u k} A) (smalltype@{u k} B))
      (equiv_functor_join@{u k u k l big}
      (equiv_smalltype@{u k} A : Equiv@{u k} _ _)
      (equiv_smalltype@{u k} B : Equiv@{u k} _ _)
    
  )).
```
It forces the universes `u` and `k` to be equal. I have no idea why it is being collapsed however.

FTR the universe `big` comes from the wildcat `Type` which stems from how we defined `Join_rec`. My intention for `l` was to be `max(u,k)` since I didn't want to impose any relations between them.

jdchristensen:
I can see why it is happening, but can't find a simple solution.

The first thing is that Join@{u u l} is not definitionally equal to Join@{u u u}.  And equiv_functor_join produces an equivalence from Join@{k k l} to Join@{u u l}.  So this only matches the goal of an equivalence from Join@{k k k} to Join@{u u u} when both k and u equal l (and therefore equal each other).

I tried a few tweaks to make equiv_functor_join and isequiv_functor_join more flexible, but they didn't work.  It should be possible to decouple the third universe in Join@{u1 u2 u3} from the universe level of the equivalence.

Alizter:
> I tried a few tweaks to make equiv_functor_join and isequiv_functor_join more flexible, but they didn't work. It should be possible to decouple the third universe in Join@{u1 u2 u3} from the universe level of the equivalence.

I managed to get it to work by making the universes in the functoriality of join more flexible.

jdchristensen:
> I managed to get it to work by making the universes in the functoriality of join more flexible.

Great!  BTW, this kind of problem should go away with algebraic universes, since that third universe variable wouldn't even be there in that case.

Alizter:
I've pushed my partial adaptation of the join construction to the new smallness stuff. Unfortunately I couldn't finish the proof about smallness of colimits since I found the number of universe variables involved far too overwhelming to keep track of. It's likely I've introduced some mistakes in this push, so we will need to be careful of the universe levels and check they make sense.

theories/Homotopy/JoinConstruction.v
jdchristensen: This was done, but I guess I'll leave this unresolved until the commented out `IsEssentiallySmall` is removed.

jdchristensen:
> I've pushed my partial adaptation of the join construction to the new smallness stuff. Unfortunately I couldn't finish the proof about smallness of colimits since I found the number of universe variables involved far too overwhelming to keep track of. It's likely I've introduced some mistakes in this push, so we will need to be careful of the universe levels and check they make sense.

It's too bad that the changes to equiv_functor_join required so many universe annotations.  Do you think it's worth waiting for algebraic universes, which I expect will simplify a lot of things here?

Alizter:
@jdchristensen It's possible it can already be simplified, but I spent a lot of time fighting Coq this time around to get it to work. I can see about removing some annotations when I get some more time.

PR #1862, opened 2024-02-19
[8.19] fix % in Arguments warnings

Alizter:
This draft PR is for when we bump the minimal version of Coq to 8.19. It fixes a warning where `%` was changed to `%_` and hence `%` in `Arguments` is now deprecated. In the mean time, I will just disable this warning in the CI.

For the record (because I totally forgot this earlier):
- `%` means open scope deeply in the argument (in 8.20).
- `%_` means open scope shallowly in the argument (previous behaviour of `%`).

### DIFF
--- a/theories/Algebra/Universal/Algebra.v
+++ b/theories/Algebra/Universal/Algebra.v
@@ -116,7 +116,7 @@ Proof.
   exact q.
 Defined.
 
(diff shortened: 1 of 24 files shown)


jdchristensen:
I didn't check carefully, but this seems to be a straightforward seach-and-replace, so LGTM.

PR #1863, opened 2024-02-19
surpress argument-scope-delimiter warning

Alizter:
This suppresses the `argument-scope-delimiter` warning which we cannot fix until Coq 8.19. A PR fixing it is ready in #1862. In the meantime we can safely ignore this warning.


Alizter:
Merging this as it is a simple tweak to our warnings.

PR #1864, opened 2024-02-19
[8.18] bump minimal version to Coq 8.18

Alizter:
This is a draft for now as we need to decide whether or not to do this yet.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ concurrency:
 
 # We set the supported coq-version from here. In order to use this environment variable correctly, look at how they are used in the following jobs.
 env:
-  coq-version-supported: '8.17'
+  coq-version-supported: '8.18'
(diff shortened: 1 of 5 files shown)


Alizter:
@jdchristensen Are you still on 8.17 for your private projects?

jdchristensen:
I use 8.18.0 currently.  With 8.19 out, I think it's good if we support both 8.18 and 8.19, dropping 8.17.
With this PR, will the CI check both 8.18 and 8.19?

Alizter:
We already check 8.19 and Coq master. 8.19 is checked by "latest" and master by "dev". "Supported" is 8.17 however it also means we miss 8.18 in the CI. Supporting more than 3 Coq versions is technically difficult as things are typically deprecated and removed over a 2 release cycle.

This PR will mean we are checking 8.18, 8.19 and Coq master.

jdchristensen (approved):

PR #1867, opened 2024-02-19
Smallness

Alizter:
@jdchristensen This is the stuff from Smallness.v that doesn't rely on the join construction. When we finish the join construction, we can think about adding the other results in.

TODO:
- [ ] check all the comments and see if they make sense.
- [x] resolve comments about naming convention. Should we pick `islocally_small` or `islocallysmall`? I think the latter.
- [x] Should `IsLocallySmall` be an inductive type? Experience from `IsTrunc` suggests that this might be a good idea. (Edit: Probably not)
- [x] Make `IsLocallySmall` and `IsSmall` typeclasses?
- [x] Make IsSmall cumulative?


theories/Smallness.v
```diff
@@ -0,0 +1,240 @@
...
+  : IsSmall@{i j} X.
+Proof.
+  nrapply issmall_equiv_issmall.
+  - exact (equiv_fibration_replacement f)^-1%equiv.
+  - apply sigma_closed_issmall; assumption.
+Defined.
+
+(** Propositional resizing says that every (-1)-truncated type is small. *)
+Definition issmall_hprop@{i j | } `{PropResizing} (X : Type@{j}) (T : IsTrunc (-1) X)
+  : IsSmall@{i j} X.
```
jdchristensen, line 95: This is essentially the definition of `PropResizing`.  I wonder if we should make this an Axiom that depends on `PropResizing`, so we don't have duplicated concepts.  I'll expand on this idea in a top-level comment.

jdchristensen:
Here's an idea for how to organize things.  Rename the folder PropResizing to Smallness or some more general name.  Have Smallness/Core.v give just the most basic definitions, e.g. things that can go through with minimal dependencies (e.g. a subset of Basics.Overture Basics.Tactics Basics.Trunc, which are what PropResizing.v depends on).  Change the contents of PropResizing to use `IsSmall` instead of the separate axioms currently there.  (This will of course require trivial changes to other places in the library that use those axioms.)  Have new files Smallness/JoinConstruction.v which contains the join construction, and Smallness/Smallness.v (?) which contains most of what is currently in the file Smallness.v from my repo.

What do you think, @Alizter?

(I'm also hoping we can reduce the number of files in the theories folder, putting things in appropriate subfolders.)

Alizter:
@jdchristensen That sounds sensible, I will try to do that.

Alizter:
I had a go at turning IsLocallySmall into an inductive type but I am having trouble proving it is a hprop. I can get an inversion based proof to the end, but too many universes get introduced. I will hold off on that for now.

jdchristensen:
> Here's an idea for how to organize things.

(I won't repeat it all.  See above.)

A further expansion of this idea would be to have a top-level folder `theories/Universes` which would include things about propositional resizing and smallness, as discussed above, but would also contain the top-level files ObjectClassifier.v, DProp.v, HProp.v, HSet.v, TruncType.v, and any other files that work with universe levels and/or create subuniverses.  (And then maybe contrib/UniverseLevel.v could be put there too, even though it isn't needed in practice.)

Other related files are Spaces/BAut.v and Spaces/BAut/*, as well as Modalities/ReflectiveSubuniverse.v.  The latter should stay where it is, but I'm not sure about the BAut material.  It might make sense to have it in Universes/ as well, so it is alongside ObjectClassifier.v.  Or it could stay in Spaces/.

Of course, Types/Universe.v is related, but would remain where it is.

Alizter:
I've pushed some commits moving things around like you've suggested. I'll continue to work on Smallness when have the time.

jdchristensen:
Thanks, I think this is good.  I've updated STYLE.md to reflect the current situation, and also wrote it assuming that Smallness.v would be moved there.  Should we do that now?

Alizter:
Yes, I will move `Smallness.v` there, but first I wanted to move out the core file. I had an attempt at doing this a few months ago, but I got stuck updating some other files to do with prop resizing. I can't find that attempt now, but I will have a fresh go at this.

jdchristensen:
When you move the PropResizing stuff, can you update STYLE.md to stay in sync?

jdchristensen:
BTW, one of my students will be using universes and propresizing a fair bit, so it might be nice to get at least part of this PR merged soon.

Alizter:
@jdchristensen In that case, I'll see if I can make some progress with this.

Alizter:
I've pushed a commit using `IsSmall` in `PropResizing`, being careful to keep the same universe variables since it can be very sensitive later in the library. However, two results are now homeless:
- `issmall_inhabited_issmall`
- `islocally_small_trunc`

I don't really know where to put these.

Alizter:
I still haven't updated STYLE.md but I've ran out of time today.

theories/PropResizing/PropResizing.v
```diff
@@ -12,11 +15,62 @@ Existing Class PropResizing.
...
-Axiom equiv_resize_hprop : forall `{PropResizing} (A : Type@{i}) `{IsHProp A},
-    A <~> resize_hprop A.
+(** Propositional resizing says that every (-1)-truncated type is small. *)
+Axiom issmall_hprop@{i j | } : forall `{PropResizing} (X : Type@{j})
+  (T : IsHProp X), IsSmall@{i j} X.
+
+(** TODO: inline and remove *)
+Definition resize_hprop@{j i | } `{PropResizing} (A : Type@{j}) `{IsHProp A}
+  : Type@{i}
+  := smalltype@{i j} (issmall_hprop@{i j} A _).
```
jdchristensen, line 25: I'm not sure about inlining.  But if `smalltype : IsSmall A -> Type` was turned into a coercion, then it  would be pretty easy to inline.  Alternatively, if `IsSmall` was a typeclass, then something like `smalltype A` could find the instance and therefore find the small type.

(Aside: I'd prefer to have the universe variables in the same order `i j`; would that cause much trouble elsewhere?)

theories/PropResizing/PropResizing.v
```diff
@@ -12,11 +15,62 @@ Existing Class PropResizing.
...
+
+(** TODO: inline and remove *)
+Definition equiv_resize_hprop@{j i | } `{PropResizing} (A : Type@{j}) `{IsHProp A}
+  : A <~> resize_hprop A
+  := (equiv_smalltype@{i j} (issmall_hprop@{i j} A _))^-1%equiv.
```
jdchristensen, line 30: Similar comments about inlining, and universe variable order.  Also, would it cause much trouble elsewhere to have the equivalence be in the other order?

jdchristensen:
About the overall structure:  I'm not super happy with some of the results from Smallness being at the end of PropResizing, just because they need that assumption.  They don't really fit there.  Here are two solutions:

1. Make a small file, `Universes/IsSmall`, which simply defines `IsSmall`.  Then both Smallness.v and PropResizing.v can Require it, and Smallness.v can Require PropResizing.v.  This also keeps the dependencies of PropResizing minimal.  (Or, `IsSmall` could be defined in Overture.v, for example.)
2. Merge PropResizing.v into Smallness.v.  Once you have `IsSmall`, the contents of PropResizing.v is really only 5 or 10 lines.  This reduces the number of files, but has the disadvantages that users of PropResizing will depend on a lot more of the library .

theories/PropResizing/PropResizing.v
Alizter: I tried that order first, but it causes a lot of issues in Hartogs.v which I wasn't able to get to the bottom of.

Alizter:
By the way, I had a go at making `IsLocallySmall` into an inductive type, but it makes showing it is a hprop a real pain so I don't think there is any gain.

Alizter:
I've moved `IsSmall` to `Overture.v` and reduced deps in `PropResizing`. Also moved those lemmas back from `PropResizing` to `Smallness.v`.

theories/Universes/Smallness.v
```diff
@@ -190,13 +190,15 @@ Proof.
...
+(** TODO: move? *)
 (** Sends a trunc_index [n] to the natural number [n+2]. *)
 Fixpoint trunc_index_to_nat (n : trunc_index) : nat
   := match n with
     | minus_two => 0%nat
     | n'.+1 => (trunc_index_to_nat n').+1
     end.
 
+(** TODO: move? *)
 Notation "n ..+2" := (trunc_index_to_nat n) (at level 2) : trunc_scope.
```
jdchristensen: The first one should probably go into Basics/Trunc.v.  But maybe it's not common enough to warrant a Notation defined in that file?  In fact, since it's only used once in Smallness.v, maybe we should just get rid of the notation completely?

Alizter: Yeah, I think the function can stay, but the notation doesn't really save any work.

Alizter:
How do we feel about making `IsLocallySmall` and `IsSmall` typeclasses? I feel like it would be a foot-gun as having fine control over the universe variables matters.

jdchristensen:
> How do we feel about making `IsLocallySmall` and `IsSmall` typeclasses? I feel like it would be a foot-gun as having fine control over the universe variables matters.

I was just going to post a comment about how it might work out well.  I was thinking that the universes might be fine.  For example, if there is an instance saying that `A` is small, we'd still have to write `smalltype A` to get the small type, and so we'd be able to annotate it as `smalltype@{i j} A` when needed, which would trigger typeclass search for `IsSmall@{i j} A` with the correct universe variables.

Also, issmall_hprop would become an instance, so in a context where we have `PropResizing` and `IsHProp A`, we'd just be able to write `smalltype@{i j} A` to obtain the small type, letting us get rid of the `resize_hprop` wrapper (and similarly for the `equiv_resize_hprop` wrapper).

I can imagine other instances too, e.g. that `BAut X` is small.  And some of the results in Smallness.v would make sense as instances.

Alizter:
Very well, I will experiment with `IsSmall` first since `IsLocallySmall` will need some extra work/thinking to unfold during typeclass search.

jdchristensen:
And without those wrappers, PropResizing.v becomes very short, and could maybe go into Overture.v.  (I'm trying to remove junk from Overture.v, but having some of the basic definitions and axioms there is not unreasonable.)

Alizter:
Now, which lemmas should be instances?

Alizter:
PropResizing still relies on some stuff from Basics that is not found in Overture.v, but I will see if the wrappers can be removed.

Alizter:
I've removed the two wrappers from `PropResizing` only `ishrprop_resize_hprop` remains which prevents `PropSizing.v` from being inlined in `Overture.v`.

jdchristensen:
I think the move to making IsSmall a typeclass went well!

> I feel like `ishprop_resize_hprop` should be a general result about truncation levels being preserved by small types.

Oh, right, that's a good idea.  Then it's not specific to having PropResizing around.

Alizter:
I've moved the `PropResizing` axiom to `Overture.v`. Looks like it worked as expected.

Alizter:
I tried adding some more instances, but the later developments are very sensitive to additions. For instance, the `sigma_closed_issmall` lemma changes something in Ordinals.v and its not clear why.

theories/PropResizing/Nat.v
```diff
@@ -3,7 +3,7 @@
 Require Import Basics.
 Require Import Types.
 Require Import HProp.
-Require Import PropResizing.PropResizing Smallness.
+Require Import Smallness.
```
jdchristensen, line 6: We should probably use Universes.Smallness everywhere that Smallness is Required.

jdchristensen:
I think it's ok to not make other things instances for now.  If it looks like it'll be useful, someone can do so and fix problems that arise.

Alizter:
@jdchristensen Shall I take a look at making `IsLocallySmall` a typeclass?

jdchristensen:
> Shall I take a look at making `IsLocallySmall` a typeclass?

Sure.  If it's easy and works well, that might be handy.  E.g. universes are 1-locally small.  But if it causes an issue, we can skip it.

STYLE.md
```diff
@@ -27,7 +27,7 @@
     - [3.6. Truncation](#36-truncation)
     - [3.7. Coercions and Existing Instances](#37-coercions-and-existing-instances)
   - [4. Axioms](#4-axioms)
-    - [4.1. Univalence and function extensionality](#41-univalence-and-function-extensionality)
+    - [4.1. Univalence, function extensionality and propsitional resizing](#41-univalence-function-extensionality-and-propsitional-resizing)
```
jdchristensen, line 30:
```suggestion
    - [4.1. Univalence, function extensionality and propositional resizing](#41-univalence-function-extensionality-and-propositional-resizing)
```

STYLE.md
```diff
@@ -514,15 +520,17 @@ for `x, y : A`.)
 
 ## 4. Axioms ##
 
-### 4.1. Univalence and function extensionality ###
+### 4.1. Univalence, function extensionality and propsitional resizing ###
```
jdchristensen, line 523:
```suggestion
### 4.1. Univalence, function extensionality and propositional resizing ###
```

Alizter:
@jdchristensen I don't think it helped, but I've made it a typeclass and declared some instances.

jdchristensen:
> @jdchristensen I don't think it helped, but I've made it a typeclass and declared some instances.

@Alizter Did you forget to push?

Alizter:
I missed that I needed to pull before.

jdchristensen:
Great!  I think the main things left are to decide what to do about PropResizing/\*, Spaces/Universe.v, Spaces/BAut.v and Spaces/BAut/\*.  Probably the contents of PropResizing/ should just go directly in Universes/, without a subfolder?  About the rest, I wrote:

> Maybe Spaces/Universe.v should move to Universes/Automorphisms.v, or something like that? It depends on BAut.v and BAut/Rigid.v, so this would suggest that maybe those should move to the new Universes/ folder. It might still make sense to leave the Spaces/BAut/ folder in the current location with the other files in contains, since those are about specific spaces? Or maybe everything should move to Universes/?

Alizter:
I think the contents of `PropResizing` should go into `Metatheory` since it is about alternative constructions of nat and truncation using prop resizing.

Spaces/Universe seems an ideal candidate for `Universes/` I will move that there. Renaming it `Automorphisms.v` seems fine.

I suppose `BAut/` and `BAut.v` can be moved to universes.

Alizter:
I'm still unsure about `BAut/`. It contains many spaces so it seems kind of wrong to move that. I think we should perhaps just move `BAut.v` and then explode the contents of `BAut` since these are all spaces.

jdchristensen:
Coq is giving me some weird behaviour with the `issmall_hprop` instance.  For some reason, even when I specify the universe variables in `smalltype@{i j}`, it generates fresh ones if it has used the instance.  Here's an example:
```coq
Require Import Basics.Overture.
Definition test@{i j|} `{PropResizing} (P : Type@{j}) `{IsHProp P} : Type@{i}.
Proof.
  exact (smalltype@{i j} P).
  Fail Defined.
  (* Universe HoTT.PropResizing.ImpredicativeTruncation.1105 is unbound. *)
  Restart.
  pose (T := smalltype@{i j} P).
  (* T := smalltype@{HoTT.PropResizing.ImpredicativeTruncation.1093
                    HoTT.PropResizing.ImpredicativeTruncation.1094} P *)
  exact T.
  Fail Defined.
  (* Universe HoTT.PropResizing.ImpredicativeTruncation.1107 is unbound. *)
  Restart.
  pose (T := @smalltype@{i j} P (issmall_hprop@{i j} P _)).
  (* T := smalltype@{i j} P : Type@{i} *)
  exact T.
Defined.
```
This is with 8.19.1.  If I write `smalltype@{i j}`, it should be looking for a proof of `IsSmall@{i j}`, and so it should apply `issmall_hprop@{i j}`.  But for some reason it doesn't.

Alizter:
@jdchristensen Does `Set Universe Minimization ToSet` help?

Alizter:
I get the same failures on 8.20. I'm not really sure why fresh universes are introduced.

jdchristensen:
> Does `Set Universe Minimization ToSet` help?

No, it doesn't.

> I get the same failures on 8.20. I'm not really sure why fresh universes are introduced.

Is this a bug in Coq?  I'm trying to make a minimal version in vanilla Coq, but am having trouble.  During typeclass search, Coq is failing to use hypotheses in the local context to resolve goals.

theories/Colimits/Quotient.v
```diff
@@ -395,6 +396,9 @@ Definition quotient_kernel_factor_small@{a a' b ab | a < a', a <= ab, b <= ab}
   : exists (C : Type@{a}) (e : A -> C) (m : C -> B),
       IsHSet C * IsSurjection e * IsEmbedding m * (f = m o e).
 Proof.
-  exact (quotient_kernel_factor_general@{a a a a' b ab ab}
-           f (fun a b => resize_hprop@{b a} (f a = f b)) (fun x y => equiv_resize_hprop _)).
+  nrefine (quotient_kernel_factor_general@{a a a a' b ab ab}
+           f (fun a b => smalltype@{a b} (f a = f b))
+           (fun x y => (equiv_smalltype _)^-1%equiv)).
+  - exact _.
+  - rapply issmall_hprop@{a b}.
```
jdchristensen, line 403: I'm guessing that this had to be made long-winded because of the same issue with universe variables in typeclass search.

jdchristensen (commented):
This LGTM.  I'm puzzled by the universe issues in typeclass search, and so we should definitely figure them out, but I'm ok with merging this and leaving that as a separate task.  There is also more to import from my Smallness.v file, but it depends on the join construction, which is part of another PR.

theories/Colimits/Quotient.v
jdchristensen: And there are places in Smallness.v where the same thing happened.

Alizter:
Could `IsSmall` being cumulative have something to do with it? I'm also thinking that there might be a stray universe somewhere, but I cannot see where it would be.

Alizter:
Yes, setting `IsSmall` to be `NonCumulative` allows your example to succeed. That's interesting. I guess we have a dodgy variance with the universe variables which we need to fix.

Alizter:
I've fixed the variance by putting a `=` before `i` in the definition of `IsSmall@{=i j}`. The reference manual has the relevant documentation: https://coq.inria.fr/doc/v8.20/refman/addendum/universe-polymorphism.html#cumulative-noncumulative

Alizter:
Actually, that broke some things, so I've reverted. Unfortunately, I am out of time today, but it is something to do with the variance.

jdchristensen:
The default variance looks correct to me:
```
IsSmall@{i j} : Type@{j} -> Type@{max(i+1,j)}
(* +i *j |=  *)
```
If `X` satisfies `IsSmall@{i j}` and `i'` and `j'` are two other universes with `i <= i'` and such that `X` is also in universe `j'`, then we expect to have `IsSmall@{i' j'} X` as well.

However, I do see how this flexibility causes Coq to be unsure of what to do.  When it is looking for an instance to prove `IsSmall@{i' j'} X`, it is free to search for an instance of `IsSmall@{i j} X` with `i <= i'`, which introduces extra universe variables.  (This is why I had a comment in Smallness.v saying
```coq
(** Note: making [IsSmall] Cumulative makes the following two not necessary, but also means that Coq can't guess universe variables as well in other spots in the file. *)
```
If we get rid of this flexibility, then we need to reintroduce the `lift_issmall` and `lower_issmall` definitions and use them again.  I'm not sure what is best: having typeclass inference not introduce variables (but needing to explicitly lift and lower in one or two spots) or the reverse.

jdchristensen:
I'd lean towards removing the cumulativity.  Since this is all about tracking universe levels, it's probably best to be explicit.  And having typeclass inference work well will simplify many things.

jdchristensen:
I'm making good progress and will push something soon.

jdchristensen (approved):

jdchristensen:
It turns out that I only ever needed cumulativity in the universe j, while it is the cumulativity in the universe i that caused Coq problems.  So by marking i as invariant (=i), we end up with the best of both worlds: typeclass inference works well, and we don't need to do any manual lifting.

This LGTM.

jdchristensen:
I'd like to get this out to my students, so I'm going to merge.  Feel free to make another PR if you spot any things that can still be improved.

PR #1869, opened 2024-02-24
Improvements to make_equiv

jdchristensen:
The first commit changes `cbn` to `hnf` in the `make_equiv` tactic, so that the lazymatch can correctly detect path types.  If the path type was hidden in record or class, like `IsUnitPreserving`, then `make_equiv` would inadvertently destruct it with `elim`.  With the change, one proof in Group.v becomes much simpler.  `make_equiv` failed there before this change.

The second commit is independent and just gives some improvements coming from looking at places where `make_equiv` was used.  The `cbn` inserted in one spot speeds up the `make_equiv` there.

### DIFF
--- a/theories/Homotopy/HSpace/Coherent.v
+++ b/theories/Homotopy/HSpace/Coherent.v
@@ -21,13 +21,8 @@ Definition issig_iscohhspace (A : pType)
     & { hspace_left_identity : LeftIdentity hspace_op pt
     & { hspace_right_identity : RightIdentity hspace_op pt
     & hspace_left_identity pt = hspace_right_identity pt } } }
-      <~> IsCohHSpace A.
-Proof.
-  transitivity { H : IsHSpace A & IsCoherent A }.
-  2: issig.
-  unfold IsCoherent.
-  make_equiv.
-Defined.
+      <~> IsCohHSpace A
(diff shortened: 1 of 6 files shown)


jdchristensen:
I tested that the change to `hnf` has no measurable affect on the build time.

theories/Homotopy/HSpace/Coherent.v
```diff
@@ -21,13 +21,8 @@ Definition issig_iscohhspace (A : pType)
...
     & hspace_left_identity pt = hspace_right_identity pt } } }
-      <~> IsCohHSpace A.
-Proof.
-  transitivity { H : IsHSpace A & IsCoherent A }.
-  2: issig.
-  unfold IsCoherent.
-  make_equiv.
-Defined.
+      <~> IsCohHSpace A
+  := ltac:(make_equiv).
```
Alizter, line 25: Doesn't `issig` also work here?

Alizter (approved):
Fantastic!

theories/Homotopy/HSpace/Coherent.v
jdchristensen: No, because `issig` only unfolds the top-level record.

PR #1871, opened 2024-02-24
Add instances isconnected_em' and is_minus_one_connected_pointed

jdchristensen:


theories/Pointed/pTrunc.v
```diff
@@ -166,11 +166,16 @@ Definition istrunc_pmap `{Univalence} {m n : trunc_index} (X Y : pType)
   : IsTrunc n (X ->* Y)
   := istrunc_pforall X (pfam_const Y).
 
-(** We can give a different proof of the [n = -1] case (with the conclusion upgraded to contractibility).  This proof works for any reflective subuniverse and avoids univalence.  Is it possible to generalize this to dependent functions while still avoiding univalence and/or keeping [O] a general RSU or modality?  Can [istrunc_pmap] be proven without univalence?  What about [istrunc_pforall]?  If the [n = -2] or [n = -1] cases can be provied without univalence, the rest can be done inductively without univalence. *)
+(** We can give a different proof of the [n = -1] case (with the conclusion upgraded to contractibility).  This proof works for any reflective subuniverse and avoids univalence.  Is it possible to generalize this to dependent functions while still avoiding univalence and/or keeping [O] a general RSU or modality?  Can [istrunc_pmap] be proven without univalence?  What about [istrunc_pforall]?  If the [n = -2] or [n = -1] cases can be proved without univalence, the rest can be done inductively without univalence. *)
```
Alizter, line 169: I think that was supposed to be `provided`

Alizter (approved):

theories/Pointed/pTrunc.v
Alizter: Actually `proved` isn't correct either as it should be `proven`.

PR #1873, opened 2024-02-24
HIT/quotient: fix universe variables, and uses in Classes

jdchristensen:
Just like for Colimits.GraphQuotient, enforce that the quotient in HIT.quotient lands in a universe at least as big as the universe the relation takes values in.


jdchristensen:
(I think we eventually want to delete HIT.quotient and use Colimits.Quotient instead, adding anything from HIT.quotient that doesn't have a parallel in Colimits.Quotient.  But since this fix was trivial, I made it.)

Alizter (approved):

PR #1876, opened 2024-02-27
Groups for a pointed wildcat, and other things about pointedness

jdchristensen:
Commit 1:  Show that Group is a pointed wild category, and then immediately deduce that AbGroup is.  Update some comments and related things.

Commit 2: Make `grp_homo_loops` into a pointed map, and define a pointed version of `fmap loops`.  The underlying map of `grp_homo_loops` is also `fmap loops`, but it lands in group homomorphisms, so I couldn't immediately see a way to unify these two.


theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -139,22 +139,9 @@ Defined.
...
-    exact (grp_homo_unit g)^. }
-  intro A.
-  snrefine (Build_GroupHomomorphism (fun _ => mon_unit); _).
-  1: exact _.
-  { intros x y; symmetry.
-    apply left_identity. }
-  intros g x; cbn.
-  apply path_unit.
+  - exact abgroup_trivial.
+  - intro A; apply ispointedcat_group.
```
Alizter, line 143: Maybe we need an "induced" version of IsPointedCat. That way it would be more inline with the other categorical properties.

theories/Pointed/Loops.v
```diff
@@ -109,6 +109,14 @@ Proof.
   rapply fmap_zero_morphism.
 Defined.
 
+Definition pfmap_loops `{Funext} (A B : pType) : (A ->** B) ->* (loops A ->** loops B).
```
Alizter, line 112: We have a notion of "pointed functor" and we could probably define `pfmap` between two pointed categories to give us a pointed function.

theories/Algebra/AbGroups/AbelianGroup.v
jdchristensen: That would probably work, but we'd need to include a proof that `abgroup_group (abgroup_trivial) $<~> grp_trivial`, since we can't make a lemma that assumes a definitional equality.  But in the above, we use that this is definitionally true, which makes it compute better.  If we used the general lemma, we'd get some `grp_homo_id`s mixed in.  So maybe this way is fine?

theories/Pointed/Loops.v
jdchristensen: Good idea.  I've just done this in Pointed/Core.v

theories/Algebra/AbGroups/AbelianGroup.v
jdchristensen: In the latest commit, it is condensed to a two-line proof.

Alizter (approved):

PR #1878, opened 2024-03-02
Indexed products and coproducts

gio256:
Generalizes binary products and coproducts to indexed products and coproducts, building on work by @Alizter.

The primary changes are to `WildCat/Products.v` and `WildCat/Coproducts.v`, i.e. I haven't touched any of the places where binary products and coproducts are currently used, but that may be ripe for another PR.


theories/WildCat/Opposite.v
```diff
@@ -181,6 +181,16 @@ Proof.
...
 
+Definition equiv_op {A : Type} `{HasEquivs A}
+  {a b : A} (e : a $<~> b)
+  : @CatEquiv A^op _ _ _ _ _ a b.
+Proof.
+  snrapply Build_CatEquiv.
+  - exact (cate_fun e^-1$).
+  - apply isequivs_op.
+    apply cate_isequiv.
+Defined.
```
jdchristensen, line 192: This proof can be replaced with `:= e^-1$.` since `@CatEquiv A^op _ _ _ _ _ a b` is definitionally equal to `b <~> a`.  So then there's the question of whether this definition is needed at all.

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
+  : x $-> cat_prod I (fun _ => x)
+  := cat_prod_corec I (fun _ => Id x).
+
+(** *** Uniqueness of products *)
 
-(** *** Categories with binary products *)
+Definition cate_prod_0gpd {I J : Type} (ie : I <~> J)
+  (G : I -> ZeroGpd) (H : J -> ZeroGpd)
+  (f : forall (i : I), G i $<~> H (ie i))
+  : prod_0gpd I G $<~> prod_0gpd J H.
```
jdchristensen, line 156: This should probably be in ZeroGroupoid.v, I think.

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
+Proof.
```
jdchristensen, line 157: It's often easiest to create an equivalence of 0-groupoids using `isequiv_0gpd_issurjinj`.  It typically involves almost all of the same steps, with one omitted.  Not that important, but you could consider trying it.

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
+            (ap (cate_fun (f i)^-1$) (transport2 _ (eisadj ie i) _))).
+    destruct (eissect ie i).
+    exact (cate_issect (f _) _).
+Defined.
+
+(** [I]-indexed products are unique no matter how they are constructed. *)
+Definition cate_cat_prod {I J : Type} (ie : I <~> J) {A : Type} `{HasEquivs A}
+  (x : I -> A) `{!Product I x} (y : J -> A) `{!Product J y}
+  (e : forall (i : I), x i $<~> y (ie i))
+  : cat_prod I x $<~> cat_prod J y.
```
jdchristensen, line 185: This seems to be mixing the uniqueness of products with the equivalence you get from changing the indexing type up to equivalence.  It might be better to break it into two pieces.  And each of those two pieces naturally break up into smaller pieces.  E.g. uniqueness of products follows from functoriality of products:  given `x y : I -> A` and `f : forall i, x i -> y i` you get an induced map on the products.  This is functorial, so in particular it sends equivalences to equivalences, and the identity family `f` to an equivalence.  For the second part, there is functoriality in precomposition with a function `J -> I`.  Again, not a big deal, but these other facts are probably going to be needed at some point.

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
+  : Is0Functor (fun x : Fun01 I A => cat_prod I x).
+Proof.
+  nrapply Build_Is0Functor.
+  intros x y f.
+  exact (cat_prod_corec I (fun i => f i $o cat_pr i)).
+Defined.
+
+Global Instance is1functor_cat_prod (I : Type) `{IsGraph I}
+  (A : Type) `{HasProducts I A}
+  : Is1Functor (fun x : Fun01 I A => cat_prod I x).
```
jdchristensen, line 223: Ah, I see that you have the functoriality here.  And it's much easier to state it once you assume that all I-products exist, whereas the uniqueness is proven without assume that.  The functoriality is still true just assuming that the products that arise exist, but it's not so convenient to state it that way.  So maybe my earlier remark should just suggest breaking that result into two pieces, one for fixed `I` and one about equivalences `I <~> J`?

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
-      refine (cat_prod_beta_pr1 _ _ $@ _).
+      nrefine (cat_binprod_beta_pr2 _ _ $@R _ $@ _).
+      nrefine (cat_binprod_beta_pr1 _ _ $@ _).
       exact (cat_idr _)^$.
   Defined.
 
-  Lemma cate_prod_swap (x y : A)
-    : cat_prod x y $<~> cat_prod y x.
+  Lemma cate_binprod_swap (x y : A)
+    : cat_binprod x y $<~> cat_binprod y x.
```
jdchristensen, line 419: Does this also follow from `cate_cat_prod` using the non-trivial equivalence `Bool <~> Bool`?

theories/WildCat/Products.v
```diff
@@ -156,160 +143,398 @@ End Lemmata.
...
+    1,2: apply cat_binprod_swap.
+    all: apply cat_binprod_swap_cat_binprod_swap.
   Defined.
 
 End Symmetry.
 
-(** *** Product functor *)
+(** *** Binary product functor *)
 
 (** Binary products are functorial in each argument. *)
```
jdchristensen, line 430: Shouldn't the results below follow from the functoriality of the `I`-indexed product?

theories/WildCat/Coproducts.v
```diff
@@ -1,35 +1,189 @@
...
+(** *** Uniqueness of coproducts *)
+
+(** [I]-indexed coproducts are unique no matter how they are constructed. *)
+Definition cate_cat_coprod {I J : Type} (ie : I <~> J) {A : Type} `{HasEquivs A}
+  (x : I -> A) `{!Coproduct I x} (y : J -> A) `{!Coproduct J y}
+  (e : forall (i : I), x i $<~> y (ie i))
+  : cat_coprod I x $<~> cat_coprod J y.
+Proof.
+  apply equiv_op.
+  exact (cate_cat_prod (A:=A^op) ie x y (fun i => (e i)^-1$)).
```
jdchristensen, line 98: You can remove `equiv_op` and just add `^-1$` around the whole term.  Even better, if you swap `x` and `y` in the statement, it becomes the exact dual statement:
```coq
Definition cate_cat_coprod {I J : Type} (ie : I <~> J) {A : Type} `{HasEquivs A}
  (x : I -> A) `{!Coproduct I x} (y : J -> A) `{!Coproduct J y}
  (e : forall (i : I), y (ie i) $<~> x i)
  : cat_coprod J y $<~> cat_coprod I x
  := cate_cat_prod (A:=A^op) ie x y e.
```

theories/WildCat/Coproducts.v
```diff
@@ -1,35 +1,189 @@
...
+
+(** *** Existence of coproducts *)
+
+Class HasCoproducts (I A : Type) `{Is1Cat A}
+  := has_coproducts :: forall x : I -> A, Coproduct I x.
+
+Class HasAllCoproducts (A : Type) `{Is1Cat A}
+  := has_all_coproducts :: forall I : Type, HasCoproducts I A.
+
+(** *** Coproduct functor *)
```
jdchristensen, line 109: It would be nice if the results after this were also proved by just quoting the dual statements.

jdchristensen (commented):
It's really nice to see this generalization, and the way that some of the proofs become a bit shorter than in the specific case of binary products!

theories/WildCat/Products.v
gio256: This was my previous attempt to specialize the I-indexed product functor, though I'm sure there's a better route.
https://github.com/gio256/Coq-HoTT/blob/6e9662c3496a76c9846850774f86101383ec889b/theories/WildCat/Products.v#L491-L602

Alizter: I suggested that we leave the functoriality of the general product for later so that we can get this merged in the meantime. It is not so straightforward to state but I think I have an idea of what it should be.

theories/WildCat/Products.v
Alizter: I tried this initially but it ended up being the same amount of work. I also had a go at associativity and it seemed to take a lot of work. I suggested to @gio256 to try it again after this PR, but I wasn't able to work out a slick argument.

theories/WildCat/Products.v
jdchristensen: > I suggested that we leave the functoriality of the general product for later

Not sure what you mean, since the functoriality of the I-indexed product is in this PR.

Alizter: Oh yes, in that case it should be straightforward to specialise for bin products. @gio256 could you have another go? It should just be a case of putting I as Bool. You will get the "bundled" version of the functoirality result, i.e. prod is a function from `A * A -> A`. The left and right functoriality should be derivable from this. I think Bifunctor.v might have something along these lines.

theories/WildCat/Coproducts.v
Alizter: @gio256 Have a look at how the proofs worked before, they were recently simplified.

theories/WildCat/Products.v
```diff
@@ -427,89 +418,79 @@ End Symmetry.
 
 (** *** Binary product functor *)
 
-(** Binary products are functorial in each argument. *)
+(** We make this a notation so that the [BinaryProduct x y] instance can still be inferred. *)
+Notation cat_binprod_uncurried := (fun x => cat_binprod (fst x) (snd x)).
```
Alizter, line 422: This gets desugared quite early on so Coq doesn't really have any idea that this is a notation and will just see the equivalent term. I think it would be more correct to say "We make a convenient notation for the uncurried form of cat_binprod". I would also make this a `Local` notation so it doesn't leak elsewhere.

theories/WildCat/Products.v
Alizter: FTR, this form of functoriality is incorrect as it assumes extra structure on the indexing type `I`. The error was mine, since I suggested to @gio256 that `Fun01` should be used instead of the forall wildcat. However since noticing this, the implementation has simplified allowing the binary functoriality results to be easily derived.

theories/WildCat/Products.v
gio256: Is there something unique about this notation that it gets desugared in an earlier phase? Also happy to get rid of it if you think the lambda is more readable.

Alizter: Notations in Coq come in two kinds: Actual notations and "abbreviations". When you write the notation `Notation "a" := term.` it differs from the abbreviation `Notation a := term". Both of these are desugared quite early on, i.e. way before Coq does any type checking. They essentially only help for printing and parsing. This is why, it doesn't make sense to say that it helps with inference as you said in the comment.

Abbreviations are fine to have, we use them all the time in the library. But it is good practice to keep them local so that they don't magically appear when importing this file. It also makes searching for results using `Search` clearer as you can see the results are about a lambda rather than something else.

theories/WildCat/Coproducts.v
```diff
@@ -108,35 +105,26 @@ Class HasAllCoproducts (A : Type) `{Is1Cat A}
...
-    intros i.
-    nrefine (cat_coprod_beta _ _ _ $@ _).
-    nrefine (_ $@ cat_assoc_opp _ _ _).
-    symmetry.
-    nrefine (_ $@L cat_coprod_beta _ _ _ $@ _).
-    nrefine (cat_assoc_opp _ _ _ $@ _).
-    nrefine (cat_coprod_beta _ _ _ $@R _ $@ _).
-    apply cat_assoc.
+  apply is1functor_op'.
+  exact (is1functor_cat_prod I A^op).
```
jdchristensen, line 127: Nice!  It looks like it was a bit tricky to get the statements stated in the right way to make this work.

theories/WildCat/Products.v
gio256: Interesting, thanks. 

What I intended to say in that comment is that making `cat_binprod_uncurried` a `Definition` isn't useful because Coq can no longer infer the `BinaryProduct x y` instance we need, so it needs to be a `Notation` instead. I will update to make it clearer.

Alizter: The Coq reference manual goes into a lot more detail than I can on these concepts however. If you are still confused about abbreviations, here are some differences with ntoations:
- Abbreviations can be declared without typechecking the term, this can be useful if you have say a delayed typeclass resolution.
- Notations are way more powerful even though they subsume abbreviations somewhat. You can describe various syntaxes using Coq's notations, essentially with arbitrary parsing and printing rules. For instance, I believe compcert has the syntax of C as Coq terms in their verified C compiler.
- Under the hood there is a difference with how these things are handled, but they are registered as different datastructures.
- Abbreviations don't have operator precedence, associativity etc. They are just literal expression substitutions.

theories/WildCat/Products.v
```diff
@@ -427,89 +418,79 @@ End Symmetry.
...
+Global Instance is0functor_cat_binprod {A : Type} `{HasBinaryProducts A}
+  : Is0Functor cat_binprod_uncurried.
+Proof.
+  nrefine (is0functor_homotopic cat_binprod_uncurried _ _).
+  - nrapply (is0functor_compose pick2 (fun x => cat_prod Bool x)).
+    + exact is0functor_pick2.
+    + rapply is0functor_cat_prod.
+      exact hasproductsbool_hasbinaryproducts.
+  - reflexivity.
+Defined.
```
jdchristensen, line 458: I can't figure out what is going on in this proof.  You seem to be using that `cat_binprod_uncurried` is homotopic to itself?  So it should work without using `is0functor_homotopic`, but for some reason it does not.  I think this should just require the right annotations, and not `is0functor_homotopic`.

theories/WildCat/Products.v
```diff
@@ -427,89 +418,79 @@ End Symmetry.
...
+
+Global Instance is1functor_cat_binprod {A : Type} `{HasBinaryProducts A}
+  : Is1Functor cat_binprod_uncurried.
+Proof.
+  nrefine (is1functor_homotopic cat_binprod_uncurried _ _).
+  - exact (is1functor_compose pick2 (fun x => cat_prod Bool x)).
+Defined.
```
jdchristensen, line 465: Same comment here (plus I don't know why Coq doesn't require you to provide the homotopy).

theories/WildCat/Products.v
Alizter: @gio256 sorry my second message came at the same time as your reply. Yes, you can more easily write long terms which are literally the term you want typeclass inference on. A definition complicates the situation as it needs to be chosen to be unfolded during typeclass search which is quite an expensive thing to do.

theories/WildCat/Bifunctor.v
```diff
@@ -85,3 +85,67 @@ Proof.
   unfold uncurry; cbn.
   exact ((fmap (flip F b2) f) $o (fmap (F a1) g)).
 Defined.
+
+(** *** (Uncurried) Bifunctors are functorial in each argument. *)
+
+Global Instance is0functor_bifunctor01 {A B C : Type}
+  `{Is01Cat A} `{IsGraph B} `{IsGraph C}
+  (F : A * B -> C) `{!Is0Functor F} (a : A)
+  : Is0Functor (fun b => F (a, b)).
```
jdchristensen, line 94: The material here is using `bifunctor` in a slightly different way.  Earlier, it means `F : A -> B -> C` with certain properties.

1) I think we should distinguish between the two types, but am not sure of the best name.

2) Maybe these results should be combined to prove that a bifunctor of the type `F : A * B -> C` gives rise to a bifunctor of the type `F : A -> B -> C`, since that's missing.  Then the results about functoriality in each variable follow.  (Or these separate results could be kept, with followup results giving the IsBifunctor structure?)

3) I'm not sure we actually need these results.  I'll make another comment about this in the place where these are used.

theories/WildCat/Products.v
jdchristensen, line 422: Why not work with `cat_binprod` directly as a bifunctor `A -> A -> A`.  It should be the composite of a bifunctor `A -> A -> (Bool -> A)` with the product functor `(Bool -> A) -> A`.  Then `isbifunctor_compose` would be what you need, and you wouldn't need the results about functors `A * B -> C` (although they are still useful results to have).

theories/WildCat/Products.v
gio256: Agreed. I haven't had a chance to polish this proof up yet, but it also seems odd to me that `is0functor_homotopic` is needed here and especially in the next proof.

Would you suggest getting rid of `is0functor_homotopic` altogether if I can avoid using it here, or do you think it would be useful to have on its own?

theories/WildCat/Products.v
jdchristensen: But I see now that `IsBifunctor` is really `Is0Bifunctor`, and we are missing `Is1Bifunctor`.  I still think it might be cleaner to implement this rather than work with product categories, but it would be more work than I realized when I wrote the above.  If you want to keep it like this, then the material on functors `F : A * B -> C` would probably fit better in Prod.v than in Bifunctor.v.

gio256: I think the reason was the lack of `Is1Functor` results for curried bifunctors. I will see about changing `Bifunctor.v` to use a single representation of bifunctors and then change this accordingly.

theories/WildCat/Bifunctor.v
```diff
@@ -85,3 +85,67 @@ Proof.
...
+  - intros b.
+    exact (fmap_id F (a, b)).
+  - intros x y z f g.
+    (** This term is just [refine (fmap2 F ((cat_idl _)^$, Id _) $@ _)]. Unfortunately, we need to help inference out quite a bit. *)
+    refine (@gpd_comp (Hom (F (a, x)) (F (a, z))) _ _ _ _ _ _
+            (@fmap2 (A * B) C _ _ _ _ _ _ _ _ F _ _ (a, x) (a, z)
+              (Id a, _) (Id a $o Id a, _) ((cat_idl _)^$, Id _))
+            _).
+    exact (@fmap_comp (A * B) C _ _ _ _ _ _ _ _ F _ _ (a, x) (a, y) (a, z)
+            (Id a, f) (Id a, g)).
```
jdchristensen, line 119:
```suggestion
    refine (_ $@ _).
    2: rapply (fmap_comp (A:=A*B)).
    rapply (fmap2 (A:=A*B) F).
    exact ((cat_idl _)^$, Id _).
```
Sometimes Coq can figure things out if you do it in steps.  A similar thing should work for the other long one (and maybe other places too?).

theories/WildCat/Products.v
jdchristensen: I think it's probably still useful to have `is0functor_homotopic` and `is1functor_homotopic`.

Alizter: @gio256 I don't think it is a particularly useful notion as translating back and forth from a homotopic function for a functoriality result isn't going to be nice. For a hprop like `IsEquiv` that would be fine.

theories/WildCat/Bifunctor.v
gio256: Oh wow. That's much cleaner, thank you.

jdchristensen: I've found many situations where Coq correctly handles `refine (_ $@ _).  1: exact foo. ` but can't handle it when you fill in one of the two terms, like `refine (foo $@ _).`  Maybe we need to use an `Arguments grp_comp & ...` command to make it figure out some implicit arguments before parsing the rest?  Similarly, `$o` gives me lots of trouble with group homomorphisms and group isomorphisms.

theories/WildCat/Products.v
gio256: Subjectively, I wasn't able to improve on this proof using `isequiv_0gpd_issurjinj`. If it's better to use `isequiv_0gpd_issurjinj` anyway to be consistent with similar proofs, I can switch back to it.

theories/WildCat/Products.v
gio256: So far I've made the minimum necessary changes to `Bifunctor.v` to get the approach suggested above working. I'm inclined to save other changes for a separate PR, but I think at the very least it's worth doing some renaming in `Bifunctor.v`.

theories/WildCat/Bifunctor.v
```diff
@@ -8,19 +8,23 @@ Class IsBifunctor {A B C : Type} `{IsGraph A, IsGraph B, Is1Cat C}
...
 
 #[export] Existing Instance bifunctor_isfunctor_10.
 #[export] Existing Instance bifunctor_isfunctor_01.
 Arguments bifunctor_isbifunctor {A B C} {_ _ _ _ _ _}
   F {_} {a0 a1} f {b0 b1} g.
 
+Class Is1Bifunctor {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
+  (F : A -> B -> C) `{!IsBifunctor F}
+  := {
+    bifunctor_is1functor_10 :: forall a : A, Is1Functor (F a);
```
gio256, line 24: I followed the convention above here, but IMO it's more consistent to remove the final underscores and switch the `10` / `01` suffixes. In my mind `bifunctor_is0functor01` should mean that you can `fmap` in the second argument.

Alizter: Rather than using indexes maybe we could write `_fst` and `_snd`? The index is not immediately understandably IMO.

theories/WildCat/Bifunctor.v
```diff
@@ -8,19 +8,23 @@ Class IsBifunctor {A B C : Type} `{IsGraph A, IsGraph B, Is1Cat C}
...
-    bifunctor_isfunctor_01 :
-    forall b, Is0Functor (flip F b);
-    bifunctor_isbifunctor :
-    forall a0 a1 (f : a0 $-> a1) b0 b1 (g : b0 $-> b1),
-      fmap (F _) g $o fmap (flip F _) f $==
-        fmap (flip F _) f $o fmap (F _) g
+    bifunctor_isfunctor_01 : forall b, Is0Functor (flip F b);
+    bifunctor_isbifunctor : forall a0 a1 (f : a0 $-> a1) b0 b1 (g : b0 $-> b1),
+      fmap (F _) g $o fmap (flip F _) f $== fmap (flip F _) f $o fmap (F _) g
   }.
```
Alizter, line 14: I think this should be named `Is0Bifunctor`

Alizter: But maybe leave this for another PR. If you don't feel like doing it, we can open an issue. At the time this was added I was not able to review too well unfortunately.

theories/WildCat/Products.v
jdchristensen: No, that's fine, if it's no easier, no need to use it.  Thanks for checking.

theories/WildCat/Bifunctor.v
jdchristensen: I think either `01`/`10` without the underscore or `_fst`/`_snd` are both fine.

theories/WildCat/Bifunctor.v
```diff
@@ -8,19 +8,23 @@ Class IsBifunctor {A B C : Type} `{IsGraph A, IsGraph B, Is1Cat C}
...
 #[export] Existing Instance bifunctor_isfunctor_10.
 #[export] Existing Instance bifunctor_isfunctor_01.
 Arguments bifunctor_isbifunctor {A B C} {_ _ _ _ _ _}
   F {_} {a0 a1} f {b0 b1} g.
 
+Class Is1Bifunctor {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
+  (F : A -> B -> C) `{!IsBifunctor F}
+  := {
+    bifunctor_is1functor_10 :: forall a : A, Is1Functor (F a);
+    bifunctor_is1functor_01 :: forall b : B, Is1Functor (flip F b);
```
jdchristensen, line 25: I wonder if there also needs to be some coherences?  Not sure.  If we leave it like this, we could add a comment saying that we might find in the future that some coherences are needed.

theories/WildCat/Products.v
```diff
@@ -190,7 +198,7 @@ Proof.
 Defined.
 
 Global Instance is1functor_cat_prod (I : Type) (A : Type) `{HasProducts I A}
-  : @Is1Functor (I -> A) A _ _ _ (is1cat_forall I (fun _ => A)) _ _ _ _
+  : @Is1Functor (I -> A) A (isgraph_forall I (fun _ => A)) _ _ _ _ _ _ _
```
jdchristensen, line 201: For `B : I -> A`, maybe we should define `cat_fam B := forall i, B a` in Forall.v, and then use that notation in the results in that file.  Then here we could say `(A:=cat_fam (fun _ : I => A))` and Coq should be able to find the instance.  And maybe we should have a name for the special case of a constant family, like this.

theories/WildCat/Bifunctor.v
gio256: I think an `Is1Functor (uncurry F)` version of `Is0Functor_uncurry_bifunctor` would be the thing to prove to figure out if we need some coherences, correct?

theories/WildCat/Products.v
jdchristensen: The current commit gets rid of them.  I don't feel strongly about whether to keep them in this PR or not.

theories/WildCat/Bifunctor.v
jdchristensen: Ah, right, good point!  I think the 1-bifunctors `A -> B -> C` should match the 1-functors `A * B -> C`.

theories/WildCat/Products.v
Alizter: @gio256 you can fix this stray instance called `isgraph_arrow` by completely getting rid of it. It is used in one place in `Universe.v` and that is because that uses `equiv_fun` rather than `cate_fun` which should fix it. Once you've done that, the correct instance will be picked up here:
```coq
  : Is0Functor (fun x : I -> A => cat_prod I x).
  : Is1Functor (fun x : I -> A => cat_prod I x).
```
making sure to annotate the source type.

theories/WildCat/Products.v
```diff
@@ -156,160 +142,406 @@ End Lemmata.
...
+    + apply cat_binprod_beta_pr1.
+    + apply cat_binprod_beta_pr2.
+  - intros z f g p.
+    apply cat_binprod_eta_pr.
+    + exact (p true).
+    + exact (p false).
+Defined.
+
+(** From binary products, all Bool-shaped products can be constructed. This should not be an instance. *)
+Definition hasproductsbool_hasbinaryproducts {A : Type} `{HasBinaryProducts A}
```
Alizter, line 337: Making this an instance will make the proofs about binprod being a bifunctor easier. IIUC you don't want typeclass resolution to become cyclic because of this. Coq has a general mechanism for hint databases (of which typeclasses are an example), that allows you to cut the search tree. https://coq.inria.fr/doc/V8.18.0/refman/proofs/automatic-tactics/auto.html#coq:cmd.Hint-Cut

Basically you want to make sure both compositions of the circular instances are disallowed and put that hint cut into the `typeclass_instances` database. Once you've done that, you'll see you won't need to expand implicit arguments in the proofs I mentioned.

theories/WildCat/Products.v
Alizter: I think eventually (not now) we would want to have some dedicated wildcat diagram source categories. So rather than using `Bool` we would have `cat_2 := Bool` and have all the instances defined on that. In the future we might ask for more categorical structure on `cat_2` but I think that would be out of place with `Bool` as there are multiple categories, groupoids etc. whose underlying type is `Bool`.

theories/WildCat/Products.v
Alizter: I can't think of any situation where they would be useful, but then again there may be such a situation. In that situation, we could easily reintroduce it.

theories/WildCat/Products.v
gio256: It turns out `isgraph_arrow` is used in a number of places, including in `Homotopy/Join/` and `Algebra/`. I'm happy to go through and make these instances explicit, but I'm unsure if that's what we want.

If it's just about the two functor instances here, I can make an `Arguments ... : rename` for `Is0Functor` and `Is1Functor` to clean them up. It sounds like it might be preferable to force uses of `isgraph_arrow` to be explicit, though.

Another option is to put `isgraph_arrow`,  `is01cat_arrow`, and `is0gpd_arrow` in their own file. Then they can be imported by the files that use them but don't come along with `WildCat/Universe.v`.

Alizter: I don't think we should keep it general as it is about Universe.v specifically. I'll see if I can remove isgraph_arrow myself in another PR.

Alizter: I've opened https://github.com/HoTT/Coq-HoTT/pull/1884, you can cherry pick that here so you can continue working. When you rebase and that PR is merged then the commit should dissapear.

theories/WildCat/Products.v
```diff
@@ -189,17 +189,16 @@ Class HasAllProducts (A : Type) `{Is1Cat A}
 (** *** Product functor *)
 
 Global Instance is0functor_cat_prod (I : Type) (A : Type) `{HasProducts I A}
-  : @Is0Functor (I -> A) A (isgraph_forall I (fun _ => A)) _
-    (fun x => cat_prod I x).
+  : Is0Functor (fun x : I -> A => cat_prod I x).
 Proof.
+  Set Printing All.
```
jdchristensen, line 194:
```suggestion
```

theories/WildCat/Products.v
jdchristensen: I thought that `pose`ing it might let it be found automatically, but that didn't work.  Anyways, the new approach seems fine.  I'm resolving this conversation, and resolved a few others as well.

theories/WildCat/Products.v
gio256: I believe this is resolved, but not certain. I left `cate_cat_prod` as-is (removing the comment) and used it to prove `cat_prod_unique`. Would you still prefer these be broken up differently?

jdchristensen: My idea was that there are two independent results here, one involving an equivalence `f : I <~> J` between the domains and the other involving equivalences `e i : x i <~> y i` between the values.  In the first case, we would have only one diagram, say `x : J -> A`, and the other would be given by composition with `f`.  In the second case, we would have only one domain `I`.  I feel like the proofs of these two results should be independent, and that someone could combine them in series if they wanted to.  The statement of `cate_cat_prod` is the composite statement, which doesn't feel to me like it would arise very often.  This isn't a big deal, but is just what feels natural to me.  If you think it would be unnecessary work, feel free to skip this.

jdchristensen: Thinking about it more, I like how you've done it. I think it's more efficient than splitting it up. So I'll mark this as resolved.

jdchristensen (approved):

theories/WildCat/Products.v
Alizter: I actually tried splitting this up but it seemed inefficient as you said. It could perhaps benefit from the new displayed equivalence stuff however but that is something to think about.

Alizter:
After finishing the review comments in #1884 and merging, I will take another look at this PR.

Alizter:
@gio256 I've merged #1884, please rebase.

gio256:
@Alizter should I squash this, or is the protocol to squash all commits before a PR and keep the review commits?

Alizter:
@gio256 I think it is fine not to squash this one. I'm currently having another look and once I am done I will merge as is.

theories/WildCat/Coproducts.v
```diff
@@ -1,35 +1,177 @@
-Require Import Basics EquivGpd Types.Prod Types.Sum.
-Require Import WildCat.Core WildCat.ZeroGroupoid WildCat.Equiv WildCat.Yoneda WildCat.Universe WildCat.NatTrans WildCat.Opposite WildCat.Products.
+Require Import Basics.
```
Alizter, line 1:
```suggestion
Require Import Basics.Overture Basics.Tactics.
```

theories/WildCat/ZeroGroupoid.v
```diff
@@ -139,18 +141,68 @@ Proof.
...
+    intros i.
+    exact (prod_0gpd_pr i $o f).
+  - intro f.
+    reflexivity.
+  - intro f.
+    reflexivity.
+  - reflexivity.
+Defined.
+
+Definition cate_prod_0gpd {I J : Type} (ie : I <~> J)
```
Alizter, line 182:
```suggestion
(** Indexed products of groupoids with equivalent indices and fiberwise equivalent factors are equivalent. *)
```

theories/WildCat/ZeroGroupoid.v
```diff
@@ -139,18 +141,68 @@ Proof.
...
+  1: exact (fun f => f i).
   snrapply Build_Is0Functor; cbn beta.
-  intros x y p; simpl; split.
-  - apply (fmap f p).
-  - apply (fmap g p).
+  intros f g p.
+  exact (p i).
+Defined.
+
+(** The universal property of the product of 0-groupoids holds almost definitionally. *)
```
Alizter, line 160:
```suggestion
(** The universal property of the product of 0-groupoids holds almost definitionally. *)
```

theories/WildCat/ZeroGroupoid.v
```diff
@@ -139,18 +141,68 @@ Proof.
...
-Definition prod_0gpd (G H : ZeroGpd) : ZeroGpd.
+Definition prod_0gpd (I : Type) (G : I -> ZeroGpd) : ZeroGpd.
 Proof.
-  rapply (Build_ZeroGpd (G * H)).
+  rapply (Build_ZeroGpd (forall i, G i)).
 Defined.
 
-Definition prod_0gpd_corec {G H K : ZeroGpd} (f : G $-> H) (g : G $-> K)
-  : G $-> prod_0gpd H K.
+Definition prod_0gpd_pr {I : Type} {G : I -> ZeroGpd}
```
Alizter, line 149:
```suggestion
(** The [i]-th projection from the [I]-indexed product of 0-groupoids. *)
```

theories/WildCat/ZeroGroupoid.v
```diff
@@ -139,18 +141,68 @@ Proof.
     apply e0.
 Defined.
 
-Definition prod_0gpd (G H : ZeroGpd) : ZeroGpd.
+Definition prod_0gpd (I : Type) (G : I -> ZeroGpd) : ZeroGpd.
```
Alizter, line 144:
```suggestion
(** [I]-indexed products for an [I]-indexed family of 0-groupoids. *)
```

Alizter:
Something interesting that I noticed was that we don't need `Funext` to show all coproducts exist in Type, but we do for products. The eta-rule for products is obviously equivalent to `path_forall`, so I don't think we can do any better.

theories/WildCat/ZeroGroupoid.v
Alizter: Sorry this should not be a replacement suggestion, but just a comment above.

theories/WildCat/ZeroGroupoid.v
Alizter: Same for here.

theories/WildCat/ZeroGroupoid.v
Alizter: Same for here.

jdchristensen:
> Something interesting that I noticed was that we don't need `Funext` to show all coproducts exist in Type, but we do for products. The eta-rule for products is obviously equivalent to `path_forall`, so I don't think we can do any better.

That's interesting.  This is a spot where the 0-groupoid structure isn't quite working in our favour.  We do have an equivalence of types between `Z -> (forall i, X i)` and `forall i, (Z -> X i)` by swapping the order of variables, and the composites are even definitionally the identity maps.  But when you add the 0-groupoid structure, we get homotopies in the `Z` variable on the left, but in *both* variables on the right, so things don't match up.

theories/WildCat/ZeroGroupoid.v
gio256: Thanks for the comments, those are always helpful.

theories/WildCat/Coproducts.v
```diff
@@ -285,3 +285,16 @@ Proof.
     + exact (p x).
     + exact (q y).
 Defined.
+
+Global Instance hasallcoproducts_type : HasAllCoproducts Type.
```
Alizter, line 289: We can probably remove the one above and just create an instance for it after this one. You can do `:= {}.` for Coq to autofill the fields. Also add a comment saying `(** [Type] has all coproducts *)`

Alizter:
@gio256 Thanks a lot! This turned out nicely.

gio256:
Thanks for the careful feedback, @Alizter @jdchristensen!

Issue #1868, opened 2024-02-21
HITs with rewrite rules


Alizter:
In Coq 8.20, it is expected that https://github.com/coq/coq/pull/18038 will be merged giving us rewrite rules in Coq. From experimentation, it would seem that we should be able to define HITs using these rewrite rules. As an example:

```coq
Axiom Circle : Set.
Symbol base : Circle.
Symbol loop : paths base base.
Symbol Circle_rec : forall (P : Type), P -> (paths base base) -> Circle -> P.

Rewrite Rule loop_rew :=
| Circle_rec _ ?b ?loop base ==> ?b
| match loop with idpath => @idpath (Circle_rec ?B ?b ?loop base) end ==> ?loop
.
```

We can probably write down the dependent eliminators in this way too. (It might have to come after the non-dependent one to be well-typed).

I think we should definitely experiment with these in the library as they have the potential to save us a lot of path algebra. One thing I don't know about is the semantics for HITs with definitional path computation rules. Obviously, there are things like CTT where this holds, but that doesn't give us the right general model IIUC.

The other question is performance, hopefully reducing HITs won't be too costly. But in the end, it means we can get rid of the private inductive types hack.

Alizter:
FTR I asked about semantics of strict path computation rules on Zulip and it doesn't look good. So there is that to keep in mind.

https://hott.zulipchat.com/#narrow/stream/228519-general/topic/Semantics.20of.20HITs.20with.20judgemental.20path.20eliminators

jdchristensen:
It would be nice to resolve the semantic issue, but until then, I guess we should avoid this.

> Symbol Circle_rec : forall (P : Type), P -> (paths base base) -> Circle -> P.

Just a minor point that this doesn't have the right type.  `paths base base` should instead be a loop in `P` at the chosen point.

yannl35133:
I didn't notice it the first time, but since `Circle_rec ?B ?b ?loop base` can reduce, your second rewrite rule would in practice never be triggered and you would be stuck with a term `match loop return paths ?b ?b with idpath => idpath end` which would not reduce.
I guess the better solution using rewrite rules would be to do what Théo Winterhalter suggested, a symbol for `ap` and a reduction for `ap ?f idpath`.

Alizter:
[...]

PR #1880, opened 2024-03-03
redefine DPath as transport and cleanup proofs

Alizter:
We redefine DPath as transport. This simplifies many things in the library. We go through and remove many superfluous lemmas and proof steps. I've also simplified various proofs as I've seen fit especially in Smash.v.

### DIFF
--- a/theories/Cubical/DPath.v
+++ b/theories/Cubical/DPath.v
@@ -7,52 +7,31 @@ Delimit Scope dpath_scope with dpath.
 Local Open Scope dpath_scope.
 
 Definition DPath {A} (P : A -> Type) {a0 a1} (p : a0 = a1)
-  (b0 : P a0) (b1 : P a1) : Type.
-Proof.
-  destruct p.
-  exact (b0 = b1).
-Defined.
+  (b0 : P a0) (b1 : P a1) : Type
+  := transport P p b0 = b1.
 
-(* This allows DPaths to collapse to paths under cbn *)
+(** This allows DPaths to collapse to paths under cbn *)
 Arguments DPath _ / _ _ _ : simpl nomatch.
-
-(* Here is an alternative definition of DPath, for now the original suffices *)
-(* Definition DPath {A} (P : A -> Type) {a0 a1} (p : a0 = a1)
-  (b0 : P a0) (b1 : P a1) := transport P p b0 = b1.  *)
-
-(* We can prove they are equivalent anyway *)
-Definition equiv_dp_path_transport {A : Type} {P : A -> Type}
-           {a0 a1 : A} {p : a0 = a1} {b0 : P a0} {b1 : P a1}
-  : transport P p b0 = b1 <~> DPath P p b0 b1.
-Proof.
-  by destruct p.
-Defined.
-
-(** We abbreviate many names that are equivalences *)
-Notation dp_path_transport := equiv_dp_path_transport.
 
 Global Instance istrunc_dp {A : Type} {P : A -> Type} {n : trunc_index}
  {a0 a1} {p : a0 = a1} {b0 : P a0} {b1 : P a1} `{IsTrunc n.+1 (P a0)}
-  `{IsTrunc n.+1 (P a1)} : IsTrunc n (DPath P p b0 b1).
-Proof.
-  exact (istrunc_equiv_istrunc _ dp_path_transport).
-Defined.
+  `{IsTrunc n.+1 (P a1)} : IsTrunc n (DPath P p b0 b1) := _.
 
 Definition dp_ishprop {A : Type} (P : A -> Type) {a0 a1} {p : a0 = a1}
   {b0 : P a0} {b1 : P a1} `{IsHProp (P a0)} `{IsHProp (P a1)}
   : DPath P p b0 b1.
 Proof.
-  apply dp_path_transport, path_ishprop.
+  apply path_ishprop.
 Defined.
 
-(* We have reflexivity for DPaths, this helps coq guess later *)
+(** We have reflexivity for DPaths, this helps coq guess later *)
 Definition dp_id {A} {P : A -> Type} {a : A} {x : P a} : DPath P 1 x x := 1%path.
 
-(* Althought 1%dpath is definitionally 1%path, coq cannot guess this so it helps
+(** Althought 1%dpath is definitionally 1%path, coq cannot guess this so it helps
    to have 1 be a dpath before hand. *)
 Notation "1" := dp_id : dpath_scope.
 
-(* DPath induction *)
+(** DPath induction *)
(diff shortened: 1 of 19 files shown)


theories/Cubical/DPath.v
```diff
@@ -7,52 +7,31 @@ Delimit Scope dpath_scope with dpath.
...
+  apply path_ishprop.
 Defined.
 
-(* We have reflexivity for DPaths, this helps coq guess later *)
+(** We have reflexivity for DPaths, this helps coq guess later *)
 Definition dp_id {A} {P : A -> Type} {a : A} {x : P a} : DPath P 1 x x := 1%path.
 
-(* Althought 1%dpath is definitionally 1%path, coq cannot guess this so it helps
+(** Althought 1%dpath is definitionally 1%path, coq cannot guess this so it helps
    to have 1 be a dpath before hand. *)
```
jdchristensen, line 31:
```suggestion
(** Although [1%dpath] is definitionally [1%path], coq cannot guess this so it helps to have [1] be a dpath beforehand. *)
```
This fixes some typos in the comment, but to be honest I don't know what it means, especially the "beforehand" part.  (I know this isn't part of this PR, but it was edited, so I noticed it.)

theories/Cubical/DPath.v
```diff
@@ -353,9 +316,7 @@ Definition equiv_path_sigma_dp {A P} {x x' : A} {y : P x} {y' : P x'}
 Proof.
   refine (equiv_path_sigma _ _ _ oE _).
   apply equiv_functor_sigma_id.
-  intro p.
-  symmetry.
-  apply dp_path_transport.
+  reflexivity.
```
jdchristensen, line 319: The whole proof can be replaced with `:= equiv_path_sigma P (x;y) (x';y').`

jdchristensen (approved):
This looks great!  This will also make it easier to use features of dependent paths even when one is working with transports.

theories/Cubical/DPath.v
Alizter: What I meant here is that Coq cannot infer lemmas about DPath's on simplified versions of `transport P p x = y` on it's own. The notation is meant to help this situation since `1 : x = x` is an identity dpath, but Coq is not smart enough to see that. When using this specific notation, Coq can then tell it should be a DPath of the form etc. helping with unification.

jdchristensen: Ok, then I think this can be clarified, e.g. `(** Although [1%dpath] is definitionally [1%path], when [1%path] is used where a dependent path is expected, Coq sometimes has trouble interpreting this.  So we make a custom notation for [1] in [dpath_scope]. *)

jdchristensen:
I also noticed that the changes here change the type of `dp_concat_p1` from
```coq
DPath (fun t : a0 = a1 => DPath P t b0 b1) (concat_p1 p) (dp @Dp 1) dp
```
to
```coq
DPath (fun t : a0 = a1 => DPath P t b0 (transport P 1 b1)) (concat_p1 p) (dp @Dp 1) dp
```
It's probably worth explicitly writing the type family here (and for nearby results) to the simpler form.

Alizter:
@jdchristensen I wasn't able to see the difference in my editor but I will mention that it is the same definitionally. I've gone ahead and added the type annotations now.

theories/Cubical/DPath.v
Alizter: I've changed the comment now to the one you've suggested.

jdchristensen:
LGTM!

mikeshulman:
Sorry to not notice this before this PR was merged, but I don't think this change is semantically compatible with how we would have to deal with `apd` in #1868.  In fact, the only semantic way I know to make sense of a computing `apd` in simplicially-based models would instead require changing `DPath` to be an inductive family with a single "dependent reflexivity" constructor.

Alizter:
@mikeshulman This PR doesn't actually change how we define DPath just that we use `transport` in the definition instead. Before this PR it wasn't actually defined as an inductive type, but by path elimination. This is essentially the same as the transport definition, just that before the target type family was a little larger.

I agree that having an "atomic" `apd` might be necessary if we want definitional computation rules for HITs but I don't see that as an immediate concern. If we do go ahead with defining HITs that way, and we are all reasonably satisfied with the semantics, then we would have to make changes to DPath anyway regardless of whether or not this PR happened.

jdchristensen:
@mikeshulman Just to make sure I understand:  even the previous definition of `DPath` would not be semantically compatible with how we would have to deal with `apd` in #1868.  But before this PR, it would probably be quite easy to change the definition of `DPath` to your proposed inductive family, since other places in the library already used conversion functions, so only those conversion functions and basic fact about dependent paths would need to be updated.  After this PR, all those conversion functions are gone, so it would be more work to change the definition of `DPath`.

I guess the question is whether we are going to trying using rewrite rules to make maps from HITs compute on path constructors.  On the one hand, having more things compute will certainly save us work, especially when you need to prove things about the proofs that currently use beta rules (and this has been troublesome for me, e.g. in work on joins).  On the other hand, adding rewrite rules and changing the definitions of `ap` and `apd` would take us further from Book HoTT and further from the hard work that has been done on initiality and semantics which shows that our proofs imply results in models.

jdchristensen:
@Alizter Our messages crossed, but I think my first paragraph might explain why it might be best to revert this PR (maybe keeping improvements to comments, etc) if we do plan to try #1868.  It'll be easier to revert now than later when other changes will interfere.

jdchristensen:
@Alizter @mikeshulman But maybe I'm wrong about reverting this PR.  E.g. if we try #1868, then maybe lots of the things that changed in this PR (e.g. various beta rules for paths) won't be needed at all.  (And I definitely like this PR.)

mikeshulman:
Yes, @jdchristensen's first paragraph was my point.  There are three possibilities.  This PR changes from the current possibility (A) to a different possibility (B), whereas #1868 would require a third possibility (C), and I suspect that the change from (A) directly to (C) would be easier than from (B) to (C), since (A) is already more opaque.

I'm not sure what my opinion is about whether we should do #1868.  If it were up to me, I think I would probably keep things as they are, so that as Dan says this repository could be said roughly speaking to be a formalization of Book HoTT.  But personally, I'm not working on this repository much any more -- I'm more invested in HOTT and the new proof assistant for it that I'm developing (where I hope the analogous rules ought to hold definitionally in a less ad hoc way, similar to how they do in cubical type theory) -- and the decisions about this repository should be made by the people who are more invested in it.

Alizter:
Thanks for your input @mikeshulman, your thoughts are definitely appreciated. I think a change like #1868 won't happen tomorrow. So I'm not particularly worried if we are in (A) or (B) as there seem to be larger concerns with that proposal. So it should be relatively easy work to adapt to an atomic form of DPath if that's what we choose to do.

jdchristensen:
I'm ok with keeping this PR in place and delaying any decision on #1868 until the future.

PR #1881, opened 2024-03-06
ci: add whitespace check

Alizter:
This adds a whitespace check to the CI that fails if there is trailing white space or a file does not end with a newline.


jdchristensen:
I thought it was pretty funny that in an earlier version, whitespace.yml didn't end in a newline...  :smile:

.github/workflows/whitespace.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  whitespace-check:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Check for trailing whitespaces
```
jdchristensen, line 17: "whitespaces" should be changed to "whitespace" everywhere.  Several blank characters are called "whitespace".

.github/workflows/whitespace.yml
```diff
@@ -0,0 +1,37 @@
...
+          echo "Trailing whitespaces found. Please fix with:\
+          bash ./etc/fix_trailing_whitespaces.sh"
+          exit 1
+        fi
+
+    - name: Check that every file ends correctly
+      run: |
+        while IFS= read -r file; do
+          # Check if the file is a regular file and not in the .git directory
+          if [ -f "$file" ] && [ "$(dirname "$file")" != ".git" ]; then
```
jdchristensen, line 30: The dirname check won't work for files in a subdirectory of the .git directory.  But why is this needed at all?  (And if it is needed, then it should be done in etc/fix_end_newlines.sh as well.)

.github/workflows/whitespace.yml
Alizter: Sorry I was typing a little too fast here.

.github/workflows/whitespace.yml
Alizter: I've changed it now.

Alizter:
@jdchristensen OK sorry for the noise. I think I'm done with this sprint. The first commit adds the CI checks. The second fixes the end newlines and the third fixes the trailing white spaces. Just to be sure I will push each commit separately again.

Alizter:
Here is the first commit, it should fail the check as there are trailing whitespaces in the library. And it does.

Alizter:
Here is the second commit, it should fail because only the trailing white spaces have been fixed but there are still problems with the end newlines.

Alizter:
Once we merge this, I'll prepare a PR that ignores these commits from git blame so that there aren't spurious blames everywhere and our history is clean. I can't do that now as the hash might change.

jdchristensen:
This would also prevent the scripts from being hidden within commits that should otherwise be ignored.

Although I approved, I'm also now having second thoughts about the whitespace commit.  It will make it harder to merge WIP in various trees, causing conflicts, and I don't think it's worth it.  An alternative would be to only complain about trailing whitespace in *new* material.  I'm not sure how to do that, but I'd bet it can be found by googling.

jdchristensen:
Something like this can also be done as a git pre-commit hook, so the person doing the work can fix the issue immediately rather than after creating a PR.

Alizter:
I can split those commits up so that the history of the scripts is preserved like you said.

I'm not sure how annoying the whitespace fix will be overall. I could tweak the CI job so that it is allowed to fail, giving a yellow status. Then at a later date we can commit the whitespace changes. The end newline changes shouldn't be disruptive to developments AFAICT.

Alizter:
I tried to setup a pre-commit hook initially, but I was having trouble getting git to find it so that it could be commited and gave up.

Alizter:
Generally, git is pretty good at rebasing on top of whitespace changes. @jdchristensen Would you be able to check how bad the breakage is when you rebase some of your trees on top of the whitespace changes here? If there are no problems then I think it would be fine to continue. If not, then we can have a look at checking only new commits.

jdchristensen:
> Generally, git is pretty good at rebasing on top of whitespace changes.

I did a test and don't see any special treatment of whitespace changes.  If my branch changes a line that also had whitespace changes in master, then it causes a conflict.  I don't think trailing whitespace is much of an issue, and so I'd like to avoid unnecessary conflicts.  I do think it's reasonable to try to prevent trailing whitespace for future commits.

Alizter:
@jdchristensen Thanks for checking, in that case I will drop the commit about removing the existing whitespace changes and modify the action so that it only checks whitespace in the commit as mentioned in that SO post.

Alizter:
I need to adapt the action from the SO post some more as I see there is a no-longer supported step in the action that was identified as a security risk. Seems sensible since modifying env vars on the fly like this probably leaks secrets.

SkySkimmer:
Coq CI has some scripts for this (the lint job) if you want some inspiration

Alizter:
@jdchristensen This seems to work. I added a dummy commit to demonstrate how it is meant to fail.

Alizter:
@jdchristensen This is perhaps stricter then we would like as it will check every commit individually for whitespaces meaning that you can't just fix the whitespace in a follow up commit. I think this is probably a good practice to have, but just so you are aware.

Alizter:
Oh and I forgot to drop the commit removing the whitespace, here is that done now.

jdchristensen:
@Alizter, thanks for figuring out how to do this.  Is it easy to do something like `git diff --check master...HEAD` to do the check on the entire diff?  (Maybe `master...HEAD` would need to be replaced with `foo HEAD`, where `foo` is the starting commit of the PR?  Not sure the context this is run in.)

Alizter:
@jdchristensen We have a shallow clone of the commits in the PR when doing the whitespace checks. I'll see if I can make it a single check rather than iterating over each commit.

Alizter:
I got something sort-of working, but I can't get github actions to print the commit, file and location of the whitespace. I am out of energy for today, so I will take another look later.

Alizter:
The commit hook suggestion here works fine by the way. I just noticed I still had that commit hook lying around, its just it wasn't marked as executable and that's why it didn't work.

jdchristensen:
> The commit hook suggestion here works fine by the way. I just noticed I still had that commit hook lying around, its just it wasn't marked as executable and that's why it didn't work.

Can you summarize the current situation?  I'm not sure what you mean by "works fine", and I'm not sure if you are talking about something that runs locally when someone creates a commit or about something that runs in the CI.

Alizter:
Yes, we should be able to write a script that git runs in the pre commit hook that checks if there is white space and removes it. That way changes to files that are committed automatically get their whitespace pruned.

Previously I had a script, but I didn't mark it as executable which is why I erroneously thought the precommit hooks were broken.

PR #1884, opened 2024-03-07
get rid of isgraph_arrow

Alizter:
This instance is far too general and should be restricted to Universe.v where it is actually used. A few places in the library were picking this up. This can be fixed by changeing a -> to a $-> so that the correct instance is inferred.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -124,7 +124,7 @@ Defined.
 
 (* A group homomorphism consists of a map between groups and a proof that the map preserves the group operation. *)
 Record GroupHomomorphism (G H : Group) := Build_GroupHomomorphism' {
-  grp_homo_map : G -> H;
+  grp_homo_map : group_type G -> group_type H;
   grp_homo_ishomo :> IsMonoidPreserving grp_homo_map;
 }.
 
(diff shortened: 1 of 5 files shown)


theories/Algebra/Groups/Group.v
```diff
@@ -124,7 +124,7 @@ Defined.
 
 (* A group homomorphism consists of a map between groups and a proof that the map preserves the group operation. *)
 Record GroupHomomorphism (G H : Group) := Build_GroupHomomorphism' {
-  grp_homo_map : G -> H;
+  grp_homo_map : group_type G -> group_type H;
```
Alizter, line 127: I originally tried making this `$->` but it messes with the coercions. I quickly tweaked it to be the underlying type instead, but it can be undone. I think it is better this way.

theories/Algebra/Rings/CRing.v
```diff
@@ -238,7 +238,7 @@ Global Instance is01cat_cring : Is01Cat CRing
   := Build_Is01Cat _ _ rng_homo_id (@rng_homo_compose).
 
 Global Instance is2graph_cring : Is2Graph CRing
-  := fun A B => isgraph_induced (@rng_homo_map A B).
+  := fun A B => isgraph_induced (@rng_homo_map A B : _ -> (cring_type A $-> _)).
```
Alizter, line 241: I don't know why this worked but in Group.v I needed to introduce it everywhere for the induced parts.

theories/WildCat/Universe.v
```diff
@@ -81,7 +77,7 @@ Proof.
   snrapply Build_HasMorExt.
   intros A B f g; cbn in *.
   snrapply isequiv_homotopic.
-  - exact (GpdHom_path o (ap (x:=f) (y:=g) equiv_fun)).
+  - exact (GpdHom_path o (ap (x:=f) (y:=g) cate_fun)).
   - nrapply isequiv_compose.
     1: apply isequiv_ap_equiv_fun.
     exact (isequiv_Htpy_path (uncore A) (uncore B) f g).
```
jdchristensen, line 83:
```suggestion
  - rapply isequiv_compose.
    rapply isequiv_Htpy_path.
```

jdchristensen: BTW, this whole result could be generalized and put into WildCat/Equiv.v, with the assumption that `ap cate_fun` is an equivalence.  Then it could be used both here and for `hasmorext_core_ptype`.

theories/Algebra/Groups/Group.v
```diff
@@ -480,17 +480,20 @@ Global Instance isgraph_group : IsGraph Group
 Global Instance is01cat_group : Is01Cat Group :=
   Build_Is01Cat Group _ (@grp_homo_id) (@grp_homo_compose).
 
+(** Helper notation so that the wildcat instances can easily be inferred. *)
+Local Notation grp_homo_map' := (fun A B => @grp_homo_map A B : _ -> (group_type A $-> _)).
```
jdchristensen, line 484:
```suggestion
Local Notation grp_homo_map' A B := (@grp_homo_map A B : _ -> (group_type A $-> _)).
```
This produces slightly smaller terms.  (The difference is hidden unless you Unset Printing Notations, but it's still there behind the scenes.)

jdchristensen (approved):
I think it's reasonable to get rid of this overly general instance.  Just a few suggestions.

theories/WildCat/Universe.v
Alizter: Doesn't look like this is even used anywhere, but I'll put a more general version next to the `core` stuff for completeness.

Alizter: I'm gonna leave this for another PR as I think I can make a few more improvements.

jdchristensen: Sure, makes sense.

Alizter:
I'll merge this when it is green.

PR #1886, opened 2024-03-07
Update bifunctors

### LINKED ISSUE #1883: Bifunctors
A number of updates to `WildCat/Bifunctor.v` have been discussed in #1878. This is a tracking issue for changes that don't make it into that PR.

###  TODOs include:
- Renaming (in particular, `IsBifunctor` -> `Is0Bifunctor`)
- A consistent plan for working with curried and uncurried bifunctors and moving between the two. Currently, the PR above introduces some lemmas for uncurried bifunctors that don't mesh very well with the rest of the work on curried bifunctors. These should be renamed and either moved to `WildCat/Prod.v` or made to fit better with the other lemmas in `WildCat/Bifunctor.v`.
- Should the bifunctor coherence condition (`bifunctor_isbifunctor`) be moved into `Is1Bifunctor`, leaving `Is0Bifunctor` with just `Is0Functor` instances in each argument? This seems to be more consistent with the unbundled structure of functors. Also, in order to prove that an uncurried bifunctor is a curried bifunctor, I needed `Is1Functor` to prove `Is0Bifunctor`. My hope would be that moving the coherence condition into `Is1Bifunctor` would limit the extra assumptions needed to move between the curried and uncurried forms.

gio256:
Depends on #1878. Addresses main points in #1883, except for changing the definition.

The big change is moving the bifunctor coherence condition to `Is1Bifunctor`. Unfortunately, this means that for any previous proofs of `Is0Bifunctor` we either need 1-functoriality to prove `Is1Bifunctor`, or we throw away the coherence proof. As far as I can tell this has been without incident so far.

### DIFF
--- a/contrib/SetoidRewrite.v
+++ b/contrib/SetoidRewrite.v
@@ -122,9 +122,7 @@ Defined.
 #[export] Instance Is1Functor_uncurry_bifunctor {A B C : Type}
   `{Is1Cat A, Is1Cat B, Is1Cat C}
   (F : A -> B -> C)
-  `{!IsBifunctor F}
-  `{forall a, Is1Functor (F a)}
-  `{forall b, Is1Functor (flip F b)}
+  `{!Is0Bifunctor F, !Is1Bifunctor F}
   : Is1Functor (uncurry F).
 Proof.
   nrapply Build_Is1Functor.
--- a/contrib/SetoidRewrite.v
+++ b/contrib/SetoidRewrite.v
@@ -99,7 +99,7 @@ Proof.
   intros f1 f2.
   apply (is0functor_postcomp a b c g ).
 Defined.
-                  
+
 #[export] Instance IsProper_catcomp {A : Type} `{Is1Cat A}
   {a b c : A}
   : CMorphisms.Proper (GpdHom ==> GpdHom ==> GpdHom)
@@ -111,41 +111,12 @@ Proof.
   exact eq_g.
 Defined.
 
-#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A} 
+#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A}
   {R : Relation B} (F : A -> B)
   `{CMorphisms.Proper _ (GpdHom ==> R) F}
   : CMorphisms.Proper (Hom ==> R) F.
 Proof.
   intros a b eq_ab; apply H2; exact eq_ab.
-Defined.
-
-#[export] Instance Is1Functor_uncurry_bifunctor {A B C : Type}
-  `{Is1Cat A, Is1Cat B, Is1Cat C}
-  (F : A -> B -> C)
-  `{!Is0Bifunctor F, !Is1Bifunctor F}
-  : Is1Functor (uncurry F).
-Proof.
-  nrapply Build_Is1Functor.
-  - intros [a1 a2] [b1 b2] [f1 f2] [g1 g2] [eq_fg1 eq_fg2];
-      cbn in f1, f2, g1, g2, eq_fg1, eq_fg2. cbn.
-    rewrite eq_fg1, eq_fg2.
-    reflexivity.
-  - intros [a b]; cbn.
-    (* rewrite fmap_id generates an extra goal. Not sure how to get typeclass resolution to figure this out automatically. *)
-    rewrite (fmap_id _).
-    rewrite (fmap_id _).
-    rewrite cat_idl.
-    reflexivity.
-  - intros [a1 b1] [a2 b2] [a3 b3] [f1 f2] [g1 g2];
-      cbn in f1, f2, g1, g2.
-    cbn.
-    rewrite (fmap_comp _).
-    rewrite (fmap_comp _).
-    rewrite cat_assoc.
-    rewrite <- (cat_assoc _ (fmap (F a1) g2)).
-    rewrite <- (bifunctor_isbifunctor F f1 g2).
-    rewrite ! cat_assoc.
-    reflexivity.
 Defined.
 
 #[export] Instance gpd_hom_is_proper1 {A : Type} `{Is0Gpd A}
@@ -192,7 +163,7 @@ Defined.
 
 Proposition nat_equiv_faithful {A B : Type}
   {F G : A -> B} `{Is1Functor _ _ F}
-  `{!Is0Functor G, !Is1Functor G} 
+  `{!Is0Functor G, !Is1Functor G}
   `{!HasEquivs B} (tau : NatEquiv F G)
   : Faithful F -> Faithful G.
 Proof.
--- a/contrib/SetoidRewrite.v
+++ b/contrib/SetoidRewrite.v
@@ -71,7 +71,7 @@ Defined.
 
 Open Scope signatureT_scope.
 
-#[export] Instance symmetry_flip {A B: Type} {f : A -> B}
+#[export] Instance symmetry_flip {A B : Type} {f : A -> B}
   {R : Relation A} {R' : Relation B} `{Symmetric _ R}
   (H0 : CMorphisms.Proper (R ++> R') f)
   : CMorphisms.Proper (R --> R') f.
@@ -80,7 +80,7 @@ Proof.
   apply H0. unfold CRelationClasses.flip. symmetry. exact Rab.
 Defined.
 
-#[export] Instance symmetric_flip_snd {A B C: Type} {R : Relation A}
+#[export] Instance symmetric_flip_snd {A B C : Type} {R : Relation A}
   {R' : Relation B} {R'' : Relation C} `{Symmetric _ R'}
   (f : A -> B -> C) (H1 : CMorphisms.Proper (R ++> R' ++> R'') f)
   : CMorphisms.Proper (R ++> R' --> R'') f.
@@ -88,7 +88,7 @@ Proof.
   intros a b Rab x y R'yx. apply H1; [ assumption | symmetry; assumption ].
 Defined.
 
-#[export] Instance IsProper_fmap {A B: Type} `{Is1Cat A}
+#[export] Instance IsProper_fmap {A B : Type} `{Is1Cat A}
   `{Is1Cat A} (F : A -> B) `{Is1Functor _ _ F} (a b : A)
   : CMorphisms.Proper (GpdHom ==> GpdHom) (@fmap _ _ _ _ F _ a b) := fun _ _ eq => fmap2 F eq.
 
@@ -111,7 +111,7 @@ Proof.
   exact eq_g.
 Defined.
 
-#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A}
+#[export] Instance gpd_hom_to_hom_proper {A B : Type} `{Is0Gpd A}
   {R : Relation B} (F : A -> B)
   `{CMorphisms.Proper _ (GpdHom ==> R) F}
   : CMorphisms.Proper (Hom ==> R) F.
--- a/theories/Algebra/AbSES/BaerSum.v
+++ b/theories/Algebra/AbSES/BaerSum.v
@@ -51,11 +51,18 @@ Proof.
   apply abses_pushout_pullback_reorder'.
 Defined.
 
-Global Instance isbifunctor_abses' `{Univalence}
-  : IsBifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
+Global Instance is0bifunctor_abses' `{Univalence}
+  : Is0Bifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
 Proof.
-  eapply Build_IsBifunctor.
-  intros ? ? g ? ? f E; cbn.
+  rapply Build_Is0Bifunctor.
+Defined.
+
+Global Instance is1bifunctor_abses' `{Univalence}
+  : Is1Bifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
+Proof.
+  snrapply Build_Is1Bifunctor.
+  1,2: exact _.
+  - intros ? ? g ? ? f E; cbn.
   apply abses_pushout_pullback_reorder.
 Defined.
 
@@ -222,15 +229,21 @@ Proof.
   - intro; apply baer_sum_unit_r.
 Defined.
 
-Global Instance isbifunctor_abses `{Univalence}
-  : IsBifunctor (AbSES : AbGroup^op -> AbGroup -> pType).
+Global Instance is0bifunctor_abses `{Univalence}
+  : Is0Bifunctor (AbSES : AbGroup^op -> AbGroup -> pType).
 Proof.
-  econstructor.
-  intros ? ? f ? ? g.
-  rapply hspace_phomotopy_from_homotopy.
-  1: apply ishspace_abses.
-  intro E; cbn.
-  apply abses_pushout_pullback_reorder.
+  rapply Build_Is0Bifunctor.
+Defined.
+
+Global Instance is1bifunctor_abses `{Univalence}
+  : Is1Bifunctor (AbSES : AbGroup^op -> AbGroup -> Type).
+Proof.
+  snrapply Build_Is1Bifunctor.
+  (*TODO: why does typeclass search find is1functor_abses' lemmas here? *)
+  - intro; exact is1functor_abses'01.
+  - exact _.
+  - intros ? ? f ? ? g E; cbn.
+    apply abses_pushout_pullback_reorder.
 Defined.
 
 (** ** Pushouts and pullbacks respect the Baer sum *)
--- a/theories/Algebra/AbSES/BaerSum.v
+++ b/theories/Algebra/AbSES/BaerSum.v
@@ -62,8 +62,8 @@ Global Instance is1bifunctor_abses' `{Univalence}
 Proof.
   snrapply Build_Is1Bifunctor.
   1,2: exact _.
-  - intros ? ? g ? ? f E; cbn.
-  apply abses_pushout_pullback_reorder.
+  intros ? ? g ? ? f E; cbn.
+  exact (abses_pushout_pullback_reorder E f g).
 Defined.
 
 (** Given a short exact sequence [A -> E -> B] and maps [f : A -> A'], [g : B' -> B], we can change the order of pushing out along [f] and pulling back along [g]. *)
@@ -236,14 +236,15 @@ Proof.
 Defined.
 
 Global Instance is1bifunctor_abses `{Univalence}
-  : Is1Bifunctor (AbSES : AbGroup^op -> AbGroup -> Type).
+  : Is1Bifunctor (AbSES : AbGroup^op -> AbGroup -> pType).
 Proof.
   snrapply Build_Is1Bifunctor.
-  (*TODO: why does typeclass search find is1functor_abses' lemmas here? *)
-  - intro; exact is1functor_abses'01.
-  - exact _.
-  - intros ? ? f ? ? g E; cbn.
-    apply abses_pushout_pullback_reorder.
+  1,2: exact _.
+  intros ? ? f ? ? g.
+  rapply hspace_phomotopy_from_homotopy.
+  1: apply ishspace_abses.
+  intro E; cbn.
+  apply abses_pushout_pullback_reorder.
 Defined.
 
 (** ** Pushouts and pullbacks respect the Baer sum *)
--- a/theories/Algebra/AbSES/Ext.v
+++ b/theories/Algebra/AbSES/Ext.v
@@ -33,12 +33,11 @@ Defined.
 Definition Ext' (B A : AbGroup@{u}) := Tr 0 (AbSES' B A).
 
 Global Instance is0bifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
-  : Is0Bifunctor@{v v w u u v} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
+  : Is0Bifunctor (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
   := is0bifunctor_compose _ _ (bf:=is0bifunctor_abses').
 
-(**TODO: universes*)
 Global Instance is1bifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
-  : Is1Bifunctor@{v v w u u v v u u} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
+  : Is1Bifunctor (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
   := is1bifunctor_compose _ _ (bf:=is1bifunctor_abses').
 
 (** [Ext B A] is an abelian group for any [A B : AbGroup]. The proof of commutativity is a bit faster if we separate out the proof that [Ext B A] is a group. *)
--- a/theories/Algebra/AbSES/Ext.v
+++ b/theories/Algebra/AbSES/Ext.v
@@ -2,7 +2,7 @@ Require Import Basics Types Truncations.Core.
 Require Import Pointed WildCat.
 Require Import Truncations.SeparatedTrunc.
 Require Import AbelianGroup AbHom AbProjective.
-Require Import AbSES.Pullback AbSES.BaerSum AbSES.Core.
+Require Import AbSES.Pullback AbSES.Pushout AbSES.BaerSum AbSES.Core.
 
 Local Open Scope mc_scope.
 Local Open Scope mc_add_scope.
@@ -27,6 +27,11 @@ Definition Ext' (B A : AbGroup@{u}) := Tr 0 (AbSES' B A).
 Global Instance is0bifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
   : Is0Bifunctor@{v v w u u v} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
   := is0bifunctor_compose _ _ (bf:=is0bifunctor_abses').
+
+(**TODO: universes*)
+Global Instance is1bifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
+  : Is1Bifunctor@{v v w u u v v u u} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
+  := is1bifunctor_compose _ _ (bf:=is1bifunctor_abses').
 
 (** [Ext B A] is an abelian group for any [A B : AbGroup]. The proof of commutativity is a bit faster if we separate out the proof that [Ext B A] is a group. *)
 Definition grp_ext `{Univalence} (B A : AbGroup@{u}) : Group.
@@ -60,7 +65,7 @@ Proof.
   apply baer_sum_commutative.
 Defined.
 
-Global Instance is01functor_ext `{Univalence} (B : AbGroup^op)
+Global Instance is0functor_ext01 `{Univalence} (B : AbGroup^op)
   : Is0Functor (ab_ext B).
 Proof.
   srapply Build_Is0Functor; intros ? ? f.
@@ -72,7 +77,7 @@ Proof.
   apply baer_sum_pushout.
 Defined.
 
-Global Instance is10functor_ext `{Univalence} (A : AbGroup)
+Global Instance is0functor_ext10 `{Univalence} (A : AbGroup)
   : Is0Functor (fun B : AbGroup^op => ab_ext B A).
 Proof.
   srapply Build_Is0Functor; intros ? ? f; cbn.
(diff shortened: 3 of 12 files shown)


theories/Algebra/AbSES/BaerSum.v
```diff
@@ -51,11 +51,18 @@ Proof.
...
-  intros ? ? g ? ? f E; cbn.
+  rapply Build_Is0Bifunctor.
+Defined.
+
+Global Instance is1bifunctor_abses' `{Univalence}
+  : Is1Bifunctor (AbSES' : AbGroup^op -> AbGroup -> Type).
+Proof.
+  snrapply Build_Is1Bifunctor.
+  1,2: exact _.
+  - intros ? ? g ? ? f E; cbn.
```
jdchristensen, line 65: Remove the bullet here.

Alizter:
I will take a look after #1878.

theories/Algebra/AbSES/BaerSum.v
```diff
@@ -222,15 +229,21 @@ Proof.
...
-  1: apply ishspace_abses.
-  intro E; cbn.
-  apply abses_pushout_pullback_reorder.
+  rapply Build_Is0Bifunctor.
+Defined.
+
+Global Instance is1bifunctor_abses `{Univalence}
+  : Is1Bifunctor (AbSES : AbGroup^op -> AbGroup -> Type).
+Proof.
+  snrapply Build_Is1Bifunctor.
```
jdchristensen, line 241: What's confusing Coq here is that `AbSES` lands in `pType`, not `Type`.  If you change it to `pType` then you can use `srapply (Build_Is1Bifunctor AbGroup^op).` and the first two goals are found by typeclass search.  I'm not sure why we need to give Coq a hint about the first category.

That all said, it's a tiny bit slow, so I recommend:
```coq
  snrapply Build_Is1Bifunctor.
  1, 2: exact _.
  intros ? ? f ? ? g.
  rapply hspace_phomotopy_from_homotopy.
  1: apply ishspace_abses.
  intro E; cbn.
  apply abses_pushout_pullback_reorder.
```

jdchristensen (commented):
This looks good!  My only other comment is that there are a few places where the proof of the coherence condition has been removed.  Would it be hard to prove `Is1Bifunctor` in those cases so we still have that information?  Or at least to state the coherence as a separate lemma, which could in the future be merged into a proof of `Is1Bifunctor`?

theories/WildCat/Bifunctor.v
```diff
@@ -4,83 +4,48 @@ Require Import WildCat.Core WildCat.Opposite WildCat.Universe WildCat.Prod.
...
-    bifunctor_is1functor_10 :: forall a : A, Is1Functor (F a);
-    bifunctor_is1functor_01 :: forall b : B, Is1Functor (flip F b);
+    bifunctor_is1functor01 :: forall a : A, Is1Functor (F a);
+    bifunctor_is1functor10 :: forall b : B, Is1Functor (flip F b);
+    bifunctor_isbifunctor : forall a0 a1 (f : a0 $-> a1) b0 b1 (g : b0 $-> b1),
+      fmap (F _) g $o fmap (flip F _) f $== fmap (flip F _) f $o fmap (F _) g
   }.
 
-Definition bifunctor_hom {C : Type} `{IsGraph C}
-  : C^op -> C -> Type := @Hom C _.
```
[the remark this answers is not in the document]
gio256: My thought was that this is all in `Yoneda.v`. I think it can be replaced by a one-liner using `is1bifunctor_bifunctor_uncurried` either in `Bifunctor.v` or in `Yoneda.v`.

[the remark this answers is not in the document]
gio256: The one thing that I think would be lost in that case is the proof of the bifunctor coherence condition in the case of strong categories (but not assuming `HasMorExt`).

gio256:
> There are a few places where the proof of the coherence condition has been removed. 

So far I have only taken the low hanging fruit here, and I haven't looked closely enough to know if there are any instances where we won't be able to prove `Is1Bifunctor` (maybe with an extra assumption). I like the idea of including the coherence as a separate lemma so it's not just lost.

Would it be worth adding an extra class `IsCoherentBifunctor` that includes the coherence separately from `Is1Bifunctor`?

I don't know that there are any lemmas that could be proven from `Is0Bifunctor, !IsCoherentBifunctor` other than the agreement of the two routes to `Is0Functor (uncurry F)`, but maybe there are still interesting cases where that helps.

theories/WildCat/Bifunctor.v
```diff
@@ -85,3 +86,109 @@ Proof.
...
+    refine (_ $@L fmap_id (F _) _ $@ cat_idl _).
+  - intros x y z f g.
+    refine (fmap_comp (flip F _) _ _ $@R _ $@ _).
+    refine (_ $@L fmap_comp (F _) _ _  $@ _).
+    refine (cat_assoc_opp _ _ _ $@ _ $@ cat_assoc _ _ _).
+    refine (cat_assoc _ _ _ $@R _ $@ _ $@ (cat_assoc_opp _ _ _ $@R _)).
+    exact (_ $@L (bifunctor_isbifunctor F _ _)^$ $@R _).
+Defined.
+
+(** *** Uncurried bifunctors are functorial in each argument. *)
```
jdchristensen, line 110: I had one more thought, which could be considered for a future PR.  All of these results about restricting a functor `A * B -> C` to one variable could also be deduced by composing with inclusion functors.  E.g. for `b : B`, the inclusion `A -> A * B` sending `a` to `(a, b)` is a 1-functor.  This might be slightly simpler.

theories/WildCat/Bifunctor.v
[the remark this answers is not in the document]
jdchristensen: Oh, I see.  In Yoneda.v, the 1-functor of 2-vars uses MorExt, while in the removed code here, Strong is used.  We also have that MorExt implies that there is a Strong structure.  So the proof using Strong is in a sense slightly more general.  Not sure the best way to organize this.  Maybe it's ok to lose the proof using Strong, but leave a comment saying that it is possible?  And the one-liner using `is1bifunctor_bifunctor_uncurried` should be added.  I wonder if all of the material about `Hom` should be moved from Yoneda.v to Bifunctor.v?

jdchristensen:
> Would it be worth adding an extra class `IsCoherentBifunctor` that includes the coherence separately from `Is1Bifunctor`?

I would first try to see if we can just prove `Is1Bifunctor` in all cases, rather than create a third Class.

theories/WildCat/Bifunctor.v
gio256: This worked out well, thanks.

theories/Algebra/AbSES/BaerSum.v
gio256: Ah I see this was just an error on my part, thanks.

theories/WildCat/Bifunctor.v
jdchristensen: Great.  I think those facts about the inclusions should be stated as separate lemmas, since they will probably be useful on their own.  Then the results where you compose can probably be `:=` style one-liners.

Also, I think all of these facts that only mention product categories and not bifunctors should be in Prod.v, since that's where someone would naturally look for them.  Not sure about the overall results that say that the curried versions are 0- and 1-bifunctors.  They could go in either file (with the dependency between the two switched if needed).

gio256: I still need to think about what to do with the two-variable functor lemmas in Prod.v, too. Maybe they can be subsumed by Bifunctors.v.

theories/Algebra/AbSES/Ext.v
```diff
@@ -24,9 +24,14 @@ Defined.
...
 Definition Ext' (B A : AbGroup@{u}) := Tr 0 (AbSES' B A).
 
-Global Instance isbifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
-  : IsBifunctor@{v v w u u v v} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
-  := isbifunctor_compose _ _ (P:=isbifunctor_abses').
+Global Instance is0bifunctor_ext'@{u v w | u < v, v < w} `{Univalence}
+  : Is0Bifunctor@{v v w u u v} (Ext' : AbGroup@{u}^op -> AbGroup@{u} -> Type@{v})
+  := is0bifunctor_compose _ _ (bf:=is0bifunctor_abses').
+
+(**TODO: universes*)
```
gio256, line 31: Having checked these now, I believe the universe constraints for both `is0bifunctor_ext'` and `is1bifunctor_ext'` are consistent with the original `isbifunctor_ext'`. That being said, I'm not really sure what I'm doing, so any input here is much appreciated.

theories/Algebra/AbSES/Ext.v
```diff
@@ -84,12 +89,62 @@ Proof.
...
+    apply abses_pushout_homotopic.
+    exact p.
+  - intros A x.
+    strip_truncations; cbn.
+    apply (ap tr).
+    apply abses_pushout_id.
+  - intros A C D f g x.
+    strip_truncations; cbn.
+    apply (ap tr).
+    apply abses_pushout_compose.
```
jdchristensen, line 108: Here's a neat trick.  `ab_ext` is the composite `Tr 0 o (AbSES B)`, except that it has been given an abelian group structure.  The graph and 2-graph structures on abelian groups are induced from the forgetful map to `Type`, which means that from the point of view of 1-functoriality, we may as well be using that composite.  So the entire proof can be replaced by the following, which just uses the corresponding facts for the composite: 
```coq
  pose (c := is1functor_compose (AbSES B) (fun X => Tr 0 X)).
  snrapply Build_Is1Functor.
  - intros A C f g.
    exact (fmap2 _ (is1functor_F:=c)).
  - exact (fmap_id _ (is1functor_F:=c)).
  - intros A C D.
    exact (fmap_comp _ (is1functor_F:=c)).
```
I didn't check how many other places could use this trick, but I think the three "is1*" results in this commit can.

(It might be possible to do this with `fmap2 Ext'`, etc, except that we'd have to manually apply the coercion `pointed_htpy`.)

jdchristensen: Thinking more, that composite is already defined in the file and is called `Ext` (without the prime), so we should just do the one-liners that show that `Ext` is a 0-functor and a 1-functor, and then use `fmap2 Ext`, etc, here.

theories/Algebra/AbSES/Ext.v
jdchristensen: I think the universes in `is0bifunctor_ext'` and `is1bifunctor_ext'` look correct.  Also, if you remove the universe annotations after `Is0Bifunctor` and `Is1Bifunctor`, coq chooses exactly those annotations, so I think it's best to omit them.

gio256: Is it generally safe to remove universe annotations if coq chooses the same, or are there cases when you would provide universe annotations to ensure that they don't change in the future?

jdchristensen: It depends.  There are some times when the universe variable annotations provide documentation and/or assert that things don't change in the future.  But for these particular annotations of `Is?Bifunctor`, I don't see any advantage to the annotations.

gio256: I've removed these now. I also think we can remove the annotations on the definitions themselves. For `is0bifunctor_ext'` without universe annotations, coq chooses all the same ones. For `is1bifunctor_ext'`, coq switches the order of the universe arguments, but otherwise it's unchanged (and this is a new definition, so the order of universe args shouldn't matter).

theories/WildCat/Bifunctor.v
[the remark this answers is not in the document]
gio256: > I wonder if all of the material about Hom should be moved from Yoneda.v to Bifunctor.v?

My only hesitation in doing this is that Yoneda.v currently reads like a very nice tour of hom functors in wild categories.

theories/WildCat/Bifunctor.v
```diff
@@ -67,26 +103,24 @@ Proof.
...
+  `{IsGraph A, IsGraph B, Is1Cat C} (F : A -> B -> C) `{!Is0Bifunctor F}
   : Is0Functor (uncurry F).
 Proof.
-  srapply Build_Is0Functor.
-  intros [a1 b1] [a2 b2] [f g]; cbn in f, g.
-  unfold uncurry; cbn.
-  exact ((fmap (flip F b2) f) $o (fmap (F a1) g)).
+  nrapply Build_Is0Functor.
+  intros [a b] [c d] [f g].
+  exact (fmap (flip F d) f $o fmap (F a) g).
```
jdchristensen: About TODO question:  the next comment says that the coherence condition implies that the two ways one could define `is0functor_uncurry_bifunctor` agree.  Now that that coherence condition is part of `Is1Bifunctor`, that's what one needs in order to know that the two agree.  I think this must have been the original motivation for putting that condition in `Is0Bifunctor`.  But maybe I'm missing something.  Can you explain what you meant by the TODO?

jdchristensen: There's another subtle question one could ask:  if I go back and forth between `Is0Bifunctor F` and `Is0Functor (uncurry F)`, do I get back where I started.  And I think the answer is no, unless `F` is a 1-bifunctor.  Same with the other composite.  Not sure what to make of this.

gio256: > Any 0-bifunctor [A -> B -> C] can be made into a functor from the product category [A * B -> C] in two ways. The bifunctor coherence condition ([bifunctor_isbifunctor]) states precisely that these two routes agree.

The TODO was to make sure it is still clear that `is0functor_uncurry_bifunctor` is not canonical in any way. The reason for changing the comment is that I wanted to make clear that agreement of the two routes to defining the uncurried functor *is* the condition `bifunctor_isbifunctor`. But obviously this condition is now in `Is1Bifunctor`.

gio256: Possibly the simplest way to explain this is just to define `is0functor_uncurry_bifunctor'` with the other route, then prove the round-trip statement you made in the case `Is1Bifunctor F`.

jdchristensen: You could add the other route and that proof, but it's pretty obvious that `bifunctor_isbifunctor` is exactly the statement that the two ways you could do it agree, so it's also ok to leave it I think.

theories/Algebra/AbSES/Ext.v
jdchristensen: Yes, I think it's also ok to remove the annotations on the definitions themselves.  I don't think there is anything subtle going on here, and the rest of the file avoids them except in a couple of places where it ensures that some abelian groups are in the same universe.

theories/WildCat/Monoidal.v
```diff
@@ -8,7 +8,7 @@ Section Monoidal.
   Context `{Is1Cat C}.
   Context `{HasEquivs C}.
   Context (tensor : C -> C -> C).
-  Context `{!IsBifunctor tensor}.
+  Context `{!Is0Bifunctor tensor}.
```
Alizter, line 11: The definition of a monoidal category needs this to be a 1-bifunctor. However this is not the only problem in this file so its fine to leave it like this. I plan to do a cleanup of this file eventually.

jdchristensen (approved):

contrib/SetoidRewrite.v
```diff
@@ -111,43 +111,14 @@ Proof.
...
-#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A} 
+#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A}
   {R : Relation B} (F : A -> B)
   `{CMorphisms.Proper _ (GpdHom ==> R) F}
   : CMorphisms.Proper (Hom ==> R) F.
 Proof.
   intros a b eq_ab; apply H2; exact eq_ab.
 Defined.
 
-#[export] Instance Is1Functor_uncurry_bifunctor {A B C : Type}
```
jdchristensen, line 122: Oh, I didn't remember this being here!

theories/WildCat/Bifunctor.v
```diff
@@ -56,18 +56,84 @@ Definition Build_Is1Bifunctor' {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
...
+  : F a0 b0 $-> F a1 b1
+  := fmap (flip F _) f $o fmap (F _) g.
+
+Definition fmap22 {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
+  (F : A -> B -> C) `{!Is0Bifunctor F, !Is1Bifunctor F}
+  {a0 a1 : A} {f : a0 $-> a1} {f' : a0 $-> a1}
+  {b0 b1 : B} {g : b0 $-> b1} {g' : b0 $-> b1}
+  (p : f $== f') (q : g $== g')
+  : fmap11 F f g $== fmap11 F f' g'
+  := fmap2 (F _) q $@R _ $@ (_ $@L fmap2 (flip F _) p).
```
jdchristensen, line 90:
```suggestion
  := fmap2 (flip F _) p $@@ fmap2 (F _) q.
```
Maybe `$@@` can be used in other places too?

theories/WildCat/DisplayedEquiv.v
```diff
@@ -172,23 +172,43 @@ Global Instance reflexive_dcate {A} {D : A -> Type} `{DHasEquivs A D} {a : A}
...
+  {a b c : A} {g : b $-> c} `{!CatIsEquiv g} {f : a $-> b} `{!CatIsEquiv f}
+  {a' : D a} {b' : D b} {c' : D c}
+  (g' : DHom g b' c') `{ge' : !DCatIsEquiv g'}
+  (f' : DHom f a' b') `{fe' : !DCatIsEquiv f'}
+  : DCatIsEquiv (fe:=compose_catie' g f) (g' $o' f').
 Proof.
   snrapply dcatie_adjointify.
-  - exact (dcate_fun f'^-1$' $o' g'^-1$').
+  - refine (_ $o' _).
+    1: nrapply (Build_DCatEquiv f')^-1$'; exact fe'.
```
jdchristensen, line 184: Having to introduce various "build*equiv*" constructions is a little annoying.  One thing that would help would be to use `pose` or `set` to put these in the context.  This also has the advantage of a smaller proof term, since they will be shared.  This applies to other parts of the recent changes as well.

theories/WildCat/Equiv.v
```diff
@@ -325,7 +335,9 @@ Definition cate_inv_compose {A} `{HasEquivs A} {a b c : A} (e : a $<~> b) (f : b
   : cate_fun (f $oE e)^-1$ $== cate_fun (e^-1$ $oE f^-1$).
 Proof.
   refine (_ $@ (compose_cate_fun _ _)^$).
-  apply cate_inv_adjointify.
+  nrefine (cate_inv_adjointify _ _ _ _ $@ _).
+  nrefine (cate_inv2 (cate_buildequiv_fun _) $@R _ $@ _).
+  exact (_ $@L cate_inv2 (cate_buildequiv_fun _)).
```
jdchristensen, line 340: Here too.

theories/WildCat/Bifunctor.v
```diff
@@ -1,86 +1,153 @@
...
+  : (F a0 b) $-> (F a1 b)
+  := fmap (flip F b) f.
+
+(** [fmap] in the second argument *)
+Definition fmap01 {A B C : Type} `{IsGraph A, IsGraph B, IsGraph C}
+  (F : A -> B -> C) `{!Is0Bifunctor F} (a : A) {b0 b1 : B} (g : b0 $-> b1)
+  : F a b0 $-> F a b1
+  := fmap (F a) g.
+
+(** There are two ways to [fmap] in both arguments of a bifunctor. The bifunctor coherence condition ([bifunctor_isbifunctor]) states precisely that these two routes agree. *)
```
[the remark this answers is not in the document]
jdchristensen: I think the both parts of the comment make sense here, as they explain under what conditions `fmap11` and `fmap11'` are equal

theories/WildCat/DisplayedEquiv.v
gio256: I'm having trouble figuring out what exactly to `pose` in this case. Maybe I'm misinterpreting your suggestion, but I haven't been able to make things cleaner using `pose` here so far.

I also have a tangential question: when we talk about wanting smaller proof terms, is the goal more readable terms, terms that are easier for coq to handle, or both?

jdchristensen: For example, you can add `pose (feq' := Build_DCatEquiv f' (fe':=fe')).` at the start of the proof, since that term is used twice in the proof.  Same for other repeated terms.

In this case, it makes the tactic script easier to read, and also should produce a proof term that is easier for Coq to handle.

theories/WildCat/DisplayedEquiv.v
```diff
@@ -387,8 +407,10 @@ Definition dcate_inv_compose {A} {D : A -> Type} `{DHasEquivs A D}
     (dcate_fun (f' $oE' e')^-1$') (dcate_fun (e'^-1$' $oE' f'^-1$')).
 Proof.
   refine (_ $@' (dcompose_cate_fun e'^-1$' f'^-1$')^$').
-  - snrapply dcate_inv_adjointify.
-  - exact isd0gpd_hom.
+  2: exact isd0gpd_hom.
+  refine (dcate_inv_adjointify _ _ _ _ $@' _).
+  refine (dcate_inv2 (dcate_buildequiv_fun _) $@R' _ $@' _).
+  exact (_ $@L' dcate_inv2 (dcate_buildequiv_fun _)).
```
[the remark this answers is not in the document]
jdchristensen: I don't see why factoring `compose_catie'` out of `compose_catie` should make subsequent things harder to prove.  I suspect that if you prove `compose_catie'` nicely, it should reduce to the same proof of `compose_catie` that you had before.  But maybe the issue is that if you start with an equivalence, and then build a new equivalence with the same underlying map and same proof that it is an equivalence, you don't get back the equivalence you had?  Unfortunately, I'm very busy for the next few days, so I won't have time to look into this, but I encourage you to see if there is a way to make this work out more simply.  One obvious alternative is to revert to the old proof of `compose_catie`, and then define `compose_catie'` using it, instead of the other way around.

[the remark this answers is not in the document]
gio256: I will have a look. I suppose I wasn't sure which way around was preferable, but I think it's all resolvable with something like a `cate_homotopic` lemma.

theories/WildCat/DisplayedEquiv.v
gio256: Oh I see. I think my confusion was really around the second question in this case, so thank you for clarifying.

jdchristensen: If you look at Coq's proof term when the `pose` command is used, it starts with `let feq' := Build_DCatEquiv f' in ...`, which shows that this has been factored out.  It turns out that `feq'` appears 40 more times in the proof term (with implicit arguments shown), so this is a non-trivial savings.  (BTW, this proof term is around 4000 lines long!)

gio256: Interesting. It's very unintuitive to me that this would *improve* coq's performance, so I greatly appreciate the suggestions.

theories/WildCat/DisplayedEquiv.v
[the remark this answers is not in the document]
gio256: I have used `cate_homotopic` to redefine `compose_catie'` in terms of `compose_catie` as you suggested. It would also be straightforward to remove one or both of these definitions if it's just simpler to avoid touching equivalences in this PR.

theories/WildCat/DisplayedEquiv.v
gio256: This comment and the one below also raise two thorny issues with displayed categories I'd like to improve on:
1. Unification and typeclass search are clearly hindered (e.g. we need `exact isd0gpd.` everywhere). I would imagine there are more tricks like the above that can be used to help coq along here.
2. Every change to WildCat/Core.v or WildCat/Equiv.v now needs to be reflected in WildCat/Displayed.v or WildCat/DisplayedEquiv.v, respectively.

theories/WildCat/Equiv.v
```diff
@@ -151,33 +151,43 @@ Global Instance symmetric_cate {A} `{HasEquivs A}
   : Symmetric (@CatEquiv A _ _ _ _ _)
   := fun a b f => cate_inv f.
 
-(** Equivalences can be composed. *)
-Definition compose_catie' {A} `{HasEquivs A} {a b c : A}
-  (g : b $-> c) `{!CatIsEquiv g} (f : a $-> b) `{!CatIsEquiv f}
-  : CatIsEquiv (g $o f).
+(** Anything homotopic to an equivalence is an equivalence. This should not be an instance. *)
+Definition cate_homotopic {A} `{HasEquivs A} {a b : A}
```
jdchristensen, line 155: I think `cate` should be `catie`, since the type is `CatIsEquiv` not `CatEquiv`.  Same for the displayed version.  I didn't check the names of other definitions.

theories/WildCat/DisplayedEquiv.v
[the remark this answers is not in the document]
jdchristensen: Thanks, that looks reasonable.  I thought about this a bit more, and my first intuition was that your previous approach would be better, since normally we first prove `IsEquiv` and then use that to prove `Equiv`.  But with wild categories, it's trickier, since just from the `CatIsEquiv` type, you don't have access to any data, such as the inverse.  You first have to build the element of `CatEquiv`, and even then, that gives you a retraction and section for a potentially different map (which is homotopic to the one you started with, but not definitionally equal).  So this makes working with `CatIsEquiv` assumptions awkward.  We could (1) throw up our hands and avoid making `CatIsEquiv` assumptions, instead having `CatEquiv` assumptions.  (2) Add some lemmas that given a map `f` and a proof of `CatIsEquiv f`, return the inverse `g` and the proofs that `f` itself is a section and a retraction of `g`.  (This is an idiom that you are using in a few places.)  (3) Think hard about the definition of `HasEquivs` and see if there is a way to make this work better.  There's a lot of redundancy in `HasEquivs`, and it's there to try to accommodate different settings, so it might not be possible to do better.

contrib/SetoidRewrite.v
```diff
@@ -111,45 +111,14 @@ Proof.
   exact eq_g.
 Defined.
 
-#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A} 
+#[export] Instance gpd_hom_to_hom_proper {A B: Type} `{Is0Gpd A}
```
Alizter, line 114:
```suggestion
#[export] Instance gpd_hom_to_hom_proper {A B : Type} `{Is0Gpd A}
```

theories/WildCat/Yoneda.v
```diff
@@ -42,6 +43,15 @@ Proof.
     refine (cat_assoc_opp _ _ _).
 Defined.
 
+Global Instance is0bifunctor_hom {A} `{Is01Cat A}
+  : @Is0Bifunctor A^op A Type _ _ _ (@Hom A _)
+  := is0bifunctor_functor_uncurried _.
+
+(** While it is possible to prove the bifunctor coherence condition from [Is1Cat_Strong], 1-functoriality requires morphism extensionality.*)
+Global Instance is1bifunctor_hom {A} `{Is1Cat A} `{HasMorExt A}
```
Alizter, line 51: Could you also prove the 0gpd versions which appear further down? Those won't require `MorphismExtensionality` IIUC.

Alizter (approved):

Alizter:
This looks good to me. I've left a comment about the 0gpd version of the yoneda functor that would also be good to have.

theories/WildCat/Yoneda.v
gio256: What do you think about proving `is?functor_opyon` using the bifunctor instances instead of the explicit proof that is there now? It seems like a repeated proof, but maybe an instructive one.

Alizter: I'm not certain. I think for now lets keep it repeated and merge it at another point if we think it would be slicker.

Alizter: Maybe @jdchristensen will have an opinion when he is available.

jdchristensen: I don't know exactly what is being discussed, but I'm all in favour of reusing proofs instead of repeating them (if it works smoothly).

gio256: I tried proving `is?functor_opyon` using the bifunctor instances, but it leads to a proliferation of extra identity maps in inconvenient places. I will play with it some more, but there might be a reason these weren't proven from the `is?functor_hom` instances previously, in which case maybe best just to leave a comment explaining why it is proven explicitly.

theories/WildCat/DisplayedEquiv.v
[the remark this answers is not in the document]
gio256: Is (2) something worth a small refactor to make `CatIsEquiv` assumptions a better default? I guess maybe best to spend some time on (3) first.

Do you know if anyone has tried to factor `CatEquiv` out of `HasEquivs`? I'm thinking of whether it's possible to provide the simplicity of `a $<~> b` without directly including it in `HasEquivs`.

theories/WildCat/Yoneda.v
```diff
@@ -229,6 +229,41 @@ Defined.
...
+Global Instance is0functor_hom_0gpd {A : Type} `{Is1Cat A}
+  : Is0Functor (A:=A^op*A) (B:=ZeroGpd) (uncurry (opyon_0gpd (A:=A))).
+Proof.
+  nrapply Build_Is0Functor.
+  intros [a1 a2] [b1 b2] [f1 f2]; cbn in *.
+  snrapply Build_Morphism_0Gpd.
+  - exact (cat_postcomp _ f2 o cat_precomp _ f1).
+  - rapply (is0functor_compose _ _).
+    + apply is0functor_precomp.
+    + apply is0functor_postcomp.
```
jdchristensen, line 241: I'm surprised that `rapply (Build_Morphism_0Gpd _ _ (cat_postcomp _ f2 o cat_precomp _ f1)).` doesn't work.  It also doesn't work to replace the last two lines with `+ exact _.`  Why doesn't Coq find the `is0functor_pre/postcomp` instances?

gio256: I was able to improve this a bit with an `unfold op in *`.

jdchristensen: Nice!

theories/WildCat/DisplayedEquiv.v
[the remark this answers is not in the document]
jdchristensen: (3) is definitely something that would take some thought and experimenting, not as part of this PR.  I don't know what people have tried.  Some "obvious" things, like just getting rid of `CatEquiv` as data, and instead defining it to be `{ e : a $-> b & CatIsEquiv e }` won't work because we want to be able to use wild categories is situations where there is already an existing notion of equivalence.  I'm not sure if that's what you meant by factoring `CatEquiv out of `HasEquivs`.  If you have another idea, please share it.

In the meantime, something like (2) could be part of this PR, since I think it would just mean taking a few repeated idioms and factor them out into lemmas.  But it could also be left as something to think about later, and this PR could stay like it is.

theories/WildCat/Yoneda.v
gio256: I have added a comment for this now, though I debated leaving no comment instead. It's a bit misleading, because there are really a number of reasons that we prove `is0functor_opyon` explicitly:
1. The reason stated in the comment, that using the bifunctor instance leads to extra identity maps in certain places.
2. It's an important functor, and it's instructive to prove explicitly and useful to control the resulting terms precisely.
3. It makes it easier to prove `is0functor_yon` as `is0functor_opyon (A:=A^op)`, which is consistent with how the rest of the contravariant lemmas are proven.
4. That's the way it is now, and changing it breaks things

theories/WildCat/DisplayedEquiv.v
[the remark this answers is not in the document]
gio256: I am happy with leaving this to think on and potentially include in a separate PR for now (assuming others feel the same).

gio256:
@Alizter I'm happy with the current state here. A few dangling questions above, but assuming everyone agrees that those can be considered resolved I think it's ready to merge.

jdchristensen:
Thanks for all of the hard work on this!

PR #1887, opened 2024-03-08
refactor: hasmorext_core

Alizter:
It would be nice if we can work out a way to generalise paths of `CatEquiv`s with morphism extensionality, but I wasn't able to work one out.

### DIFF
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -959,10 +959,9 @@ Defined.
 Global Instance hasmorext_core_ptype `{Funext} : HasMorExt (core pType).
 Proof.
   rapply hasmorext_core.
-  intros a b f g.
-  snrapply isequiv_homotopic.
-  1: apply equiv_path_pequiv'. 
-  1: exact _.
+  intros A B f g.
+  snrapply isequiv_homotopic'.
+  1: exact (equiv_path_pequiv' f g)^-1%equiv.
   by intros [].
 Defined.
 
--- a/theories/Pointed/Core.v
+++ b/theories/Pointed/Core.v
@@ -958,14 +958,12 @@ Defined.
 
 Global Instance hasmorext_core_ptype `{Funext} : HasMorExt (core pType).
 Proof.
-  snrapply Build_HasMorExt.
+  rapply hasmorext_core.
   intros a b f g.
-  unfold GpdHom_path.
-  cbn in f, g.
-  (* [GpdHom_path] and the inverse of [equiv_path_pequiv] are not definitionally equal, but they compute to definitionally equal things on [idpath]. *)
-  apply (isequiv_homotopic (equiv_path_pequiv f g)^-1%equiv).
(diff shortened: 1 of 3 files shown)


theories/Pointed/Core.v
```diff
@@ -958,14 +958,12 @@ Defined.
...
   intros a b f g.
-  unfold GpdHom_path.
-  cbn in f, g.
-  (* [GpdHom_path] and the inverse of [equiv_path_pequiv] are not definitionally equal, but they compute to definitionally equal things on [idpath]. *)
-  apply (isequiv_homotopic (equiv_path_pequiv f g)^-1%equiv).
-  intro p; induction p; cbn.
-  reflexivity.
+  snrapply isequiv_homotopic.
+  1: apply equiv_path_pequiv'. 
+  1: exact _.
```
jdchristensen, line 965: This hides that it's actually the inverse of `equiv_pathpequiv'` that applies here.  Also, it is shorter with `isequiv_homotopic'`:
```suggestion
  apply (isequiv_homotopic' (equiv_path_pequiv' f g)^-1%equiv).
```
It's too bad that these aren't actually the same map.  (That could probably be arranged, but isn't part of this PR.)

Another minor thing is that I'd prefer `A` and `B` to `a` and `b` in the unchanged line.

theories/WildCat/Universe.v
```diff
@@ -72,17 +72,7 @@ Proof.
...
-  snrapply Build_HasMorExt.
-  intros A B f g; cbn in *.
-  snrapply isequiv_homotopic.
-  - exact (GpdHom_path o (ap (x:=f) (y:=g) cate_fun)).
-  - nrapply isequiv_compose.
-    1: apply isequiv_ap_equiv_fun.
-    exact (isequiv_Htpy_path (uncore A) (uncore B) f g).
-  - intro p; by induction p.
-Defined.
+Global Instance hasmorext_core_type `{Funext} : HasMorExt (core Type) := _.
```
jdchristensen, line 75: This one worked out very nicely!

jdchristensen (approved):

jdchristensen:
> It would be nice if we can work out a way to generalise paths of `CatEquiv`s with morphism extensionality, but I wasn't able to work one out.

Can you explain what you mean?

Alizter:
@jdchristensen Yes sorry, I wrote that a bit rushed. I'm looking for a way to prove that the equivalence hypothesis for `hasmorext_core`.

Alizter:
I suppose we could add yet another field into `CatEquiv` saying that `ap cate_fun` is an equivalence.

Alizter:
I'm merging this for now as I don't think I will try anything drastic with the definition of CatEquiv to achieve such little unity.

jdchristensen:
And I don't think we'll always expect `ap cate_fun` to be an equivalence without assuming `Funext` globally, so it's best to have that proven on a case-by-case basis.

PR #1895, opened 2024-03-10
pentagon identity for lists

Alizter:
We show that the pentagon identity holds for lists. I haven't shown the triangle identity, but presumably it is easier. This is in preparation for #1896.

### DIFF
--- a/theories/Spaces/List.v
+++ b/theories/Spaces/List.v
@@ -1,4 +1,5 @@
-Require Import Basics.Overture Basics.Tactics.
+Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids.
+Require Import Classes.implementations.list.
 
 Local Open Scope list_scope.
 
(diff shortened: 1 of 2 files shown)


theories/Spaces/List.v
```diff
@@ -43,3 +44,28 @@ Section Fold_Right_Recursor.
...
+    = ap (fun l => w ++ l) (app_assoc _ x y z)
+    @ app_assoc _ w (x ++ y) z
+    @ ap (fun l => l ++ z) (app_assoc _ w x y).
+Proof.
+  induction w as [|? w IHw] in x, y, z |- *.
+  - simpl.
+    lhs nrapply concat_1p.
+    rhs nrapply concat_p1.
+    rhs nrapply concat_p1.
+    symmetry; apply ap_idmap.
```
jdchristensen, line 59: Minor point:  the `rhs` tactic uses path inversion each time, so you'll get a slightly smaller term if you move the `symmetry` before the two `rhs` tactics.

theories/Spaces/List.v
```diff
@@ -43,3 +44,28 @@ Section Fold_Right_Recursor.
...
+  - simpl.
+    lhs_V nrapply ap_pp.
+    lhs nrapply (ap (ap (cons a)) (IHw x y z)).
+    symmetry.
+    rhs nrapply ap_pp.
+    f_ap.
+    { rhs nrapply ap_pp.
+      f_ap.
+      apply ap_compose. }
+    lhs_V nrapply ap_compose.
```
jdchristensen, line 69: Similar here: completely avoiding the `symmetry` might help, as `rhs` and `lhs_V` use path inversion (but `rhs_V` and `lhs` do not).

jdchristensen (approved):

Alizter:
@jdchristensen Thanks, I've shortened the proof now.

PR #1897, opened 2024-03-12
add missing coherence for TwoOneCat

Alizter:
We were missing the bifunctor coherence for composition. This lets us prove the exchange law in a (2,1)-category.

### DIFF
--- a/theories/WildCat/TwoOneCat.v
+++ b/theories/WildCat/TwoOneCat.v
@@ -1,4 +1,4 @@
-Require Import Basics.Overture.
+Require Import Basics.Overture Basics.Tactics.
 Require Import WildCat.Core.
 Require Import WildCat.NatTrans.
 
@@ -10,6 +10,9 @@ Class Is21Cat (A : Type) `{Is1Cat A, !Is3Graph A} :=
   is1gpd_hom : forall (a b : A), Is1Gpd (a $-> b) ;
   is1functor_postcomp : forall (a b c : A) (g : b $-> c), Is1Functor (cat_postcomp a g) ;
   is1functor_precomp : forall (a b c : A) (f : a $-> b), Is1Functor (cat_precomp c f) ;
+  bifunctor_coh_comp : forall {a b c : A} {f f' : a $-> b}  {g' g'' : b $-> c}
+    (p : f $== f') (s : g' $== g''),
+    (g' $@L p) $@ (s $@R f') $== (s $@R f) $@ (g'' $@L p);
 
   (** Naturality of the associator in each variable separately *)
   is1natural_cat_assoc_l : forall (a b c d : A) (f : a $-> b) (g : b $-> c),
@@ -50,3 +53,23 @@ Global Existing Instance is1natural_cat_assoc_m.
 Global Existing Instance is1natural_cat_assoc_r.
 Global Existing Instance is1natural_cat_idl.
 Global Existing Instance is1natural_cat_idr.
+
+(** *** Exchange law *)
+
+Definition cat_exchange {A : Type} `{Is21Cat A} {a b c : A}
+  {f f' f'' : a $-> b} {g g' g'' : b $-> c}
+  (p : f $== f') (q : f' $== f'') (r : g $== g') (s : g' $== g'')
+  : (p $@ q) $@@ (r $@ s) $== (p $@@ r) $@ (q $@@ s).
+Proof.
+  refine ((_ $@@ _) $@ _).
+  1: rapply fmap_comp.
+  1: rapply fmap_comp.
+  refine (cat_assoc _ _ _ $@ _).
+  refine (_ $@ (cat_assoc _ _ _)^$).
+  nrefine (_ $@L _).
+  refine (_ $@ cat_assoc _ _ _).
+  refine ((cat_assoc _ _ _)^$ $@ _).
+  nrefine (_ $@R _).
+  symmetry.
+  apply bifunctor_coh_comp.
+Defined.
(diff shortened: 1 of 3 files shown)


theories/WildCat/TwoOneCat.v
```diff
@@ -10,6 +10,9 @@ Class Is21Cat (A : Type) `{Is1Cat A, !Is3Graph A} :=
   is1gpd_hom : forall (a b : A), Is1Gpd (a $-> b) ;
   is1functor_postcomp : forall (a b c : A) (g : b $-> c), Is1Functor (cat_postcomp a g) ;
   is1functor_precomp : forall (a b c : A) (f : a $-> b), Is1Functor (cat_precomp c f) ;
+  bifunctor_coh_comp : forall {a b c : A} {f f' : a $-> b}  {g' g'' : b $-> c}
```
Alizter, line 13: Not sure about this name. We could refactor this after #1886 and ask for composition to be a bifunctor instead.

Alizter: I'm also not sure about the direction. I've opted for "$@L" and "$@R" first but the other direction would be fine too.

theories/WildCat/TwoOneCat.v
```diff
@@ -50,3 +53,23 @@ Global Existing Instance is1natural_cat_assoc_m.
 Global Existing Instance is1natural_cat_assoc_r.
 Global Existing Instance is1natural_cat_idl.
 Global Existing Instance is1natural_cat_idr.
+
+(** *** Exchange law *)
+
+Definition cat_exchange {A : Type} `{Is21Cat A} {a b c : A}
+  {f f' f'' : a $-> b} {g g' g'' : b $-> c}
+  (p : f $== f') (q : f' $== f'') (r : g $== g') (s : g' $== g'')
```
jdchristensen, line 61: Probably best to name these paths p, p', q, q' respectively, to parallel the naming of the arrows.

theories/WildCat/TwoOneCat.v
```diff
@@ -50,3 +53,23 @@ Global Existing Instance is1natural_cat_assoc_m.
...
+  : (p $@ q) $@@ (r $@ s) $== (p $@@ r) $@ (q $@@ s).
+Proof.
+  refine ((_ $@@ _) $@ _).
+  1: rapply fmap_comp.
+  1: rapply fmap_comp.
```
jdchristensen, line 66: It seems easier for me to follow if some of the notations are unfolded and changed:
```suggestion
  unfold "$@@", "$@".
  refine ((_ $@@ _) $@ _).
  1, 2: rapply fmap_comp.
  change (_ $== ?R) with ((g'' $@L q) $o (g'' $@L p) $o ((s $@R f) $o (r $@R f)) $== R).
```
If we specialized `fmap_comp` to specific lemmas about `$@L` and `$@R`, that would make it even cleaner, avoiding the need for the `change`.  I would then inline them, so the goal would cleanly switch in one step.

theories/WildCat/TwoOneCat.v
jdchristensen: The name seems ok.  I might switch the direction, since the one place it is used, it is in the other direction.  It's probably most important to be consistent with Bifunctor.v, if that determines it.

theories/WildCat/TwoOneCat.v
```diff
@@ -10,6 +10,9 @@ Class Is21Cat (A : Type) `{Is1Cat A, !Is3Graph A} :=
   is1gpd_hom : forall (a b : A), Is1Gpd (a $-> b) ;
   is1functor_postcomp : forall (a b c : A) (g : b $-> c), Is1Functor (cat_postcomp a g) ;
   is1functor_precomp : forall (a b c : A) (f : a $-> b), Is1Functor (cat_precomp c f) ;
+  bifunctor_coh_comp : forall {a b c : A} {f f' : a $-> b}  {g' g'' : b $-> c}
+    (p : f $== f') (s : g' $== g''),
```
jdchristensen, line 14: Why `g'` and `g''` instead of `g` and `g'`?

theories/WildCat/TwoOneCat.v
```diff
@@ -10,6 +10,9 @@ Class Is21Cat (A : Type) `{Is1Cat A, !Is3Graph A} :=
...
+    (g' $@L p) $@ (s $@R f') $== (s $@R f) $@ (g'' $@L p);
```
jdchristensen, line 15: I think we should include comments saying that what is currently the RHS is `cat_comp`, which has notation `$@@`, since that notation is used below.  So this says that the two ways of defining `cat_comp` agree.  And maybe remind the reader that `$@L` and `$@R` are the 0-functoriality of post- and pre-composition.

jdchristensen (commented):
I'll look over the rest in a bit.

theories/WildCat/TwoOneCat.v
Alizter: I'll add those specific lemmas in as that will be useful.

theories/WildCat/TwoOneCat.v
Alizter: Typed too fast, will change.

theories/WildCat/TwoOneCat.v
Alizter: I will annotate the proof with some instructive comments.

Alizter:
@jdchristensen I've addressed your comments.

jdchristensen:
LGTM (modulo the build failures).

Alizter:
Oops, should be fixed now.

PR #1900, opened 2024-03-19
Zero-groupoid universal property for coequalizers

gio256:
Addresses step (3) from #1259. A few notes:

- The graph structure that I put on things "worked", but is there a more correct alternative?
- The naming of the file is a bit awkward. Assuming that we want to add naturality of this universal property w.r.t. `functor_coeq` (step (4) from 1259), my thought was that the combination of the two would make Colimits/Coeq.v untenably long, so I decided to separate this.
- I'm still making sense of the comment [here](https://github.com/gio256/Coq-HoTT/blob/82aece3a1983231dea398550e09031ce3974ec57/theories/Homotopy/Suspension.v#L284), but it seems to imply that we might need a small disclaimer on this, at least until step (4) is added.

### DIFF
--- a/theories/Colimits/CoeqUnivProp.v
+++ b/theories/Colimits/CoeqUnivProp.v
@@ -12,57 +12,26 @@ Require Import WildCat.Forall.
 Require Import WildCat.Paths.
 Require Import WildCat.ZeroGroupoid.
 
-(** Using wild 0-groupoids, the universal property can be proven without funext.  A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
+(** Using wild 0-groupoids, the universal property can be proven without funext. A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
 Section UnivProp.
   Context {B A : Type} (f g : B -> A) (P : Coeq f g -> Type).
 
-  (** The domain of the equivalence is the fully-applied type of [Coeq_ind]: sections of [P] over [Coeq f g] *)
-  Definition Coeq_ind_type := forall z : Coeq f g, P z.
-
-  Local Instance isgraph_Coeq_ind_type : IsGraph Coeq_ind_type.
-  Proof.
-    apply isgraph_forall; intros; apply isgraph_paths.
-  Defined.
-
-  Local Instance is01cat_Coeq_ind_type: Is01Cat Coeq_ind_type.
-  Proof.
-    apply is01cat_forall; intros; apply is01cat_paths.
-  Defined.
-
-  Local Instance is0gpd_Coeq_ind_type: Is0Gpd Coeq_ind_type.
-  Proof.
-    apply is0gpd_forall; intros; apply is0gpd_paths.
-  Defined.
-
-  Definition Coeq_ind_map := forall a : A, P (coeq a).
-
-  Local Instance isgraph_Coeq_ind_map : IsGraph Coeq_ind_map.
-  Proof.
-    apply isgraph_forall. intros; apply isgraph_paths.
-  Defined.
-
-  Local Instance is01cat_Coeq_ind_map : Is01Cat Coeq_ind_map.
-  Proof.
-    apply is01cat_forall; intros; apply is01cat_paths.
-  Defined.
-
-  Local Instance is0gpd_Coeq_ind_map : Is0Gpd Coeq_ind_map.
-  Proof.
-    apply is0gpd_forall; intros; apply is0gpd_paths.
-  Defined.
+  Local Existing Instances isgraph_forall is01cat_forall is0gpd_forall | 1.
+  Local Existing Instances isgraph_total is01cat_total is0gpd_total | 1.
+  Local Existing Instances isgraph_paths is01cat_paths is0gpd_paths | 2.
 
   (** The codomain of the equivalence is a sigma-groupoid of this family. *)
-  Definition Coeq_ind_data' (h : Coeq_ind_map)
+  Definition Coeq_ind_data (h : forall a : A, P (coeq a))
     := forall b : B, DPath P (cglue b) (h (f b)) (h (g b)).
 
-  (** We consider [Coeq_ind_data'] to be a displayed 0-groupoid over [Coeq_ind_map], where objects over [h : forall a : A, P (coeq a)] are homotopies [h o f == h o g] and morphisms over [p : h == k] are witnesses that p commutes with the homotopies over [h] and [k]. *)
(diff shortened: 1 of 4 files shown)


jdchristensen (approved):
That looks very clean!

jdchristensen (approved):
That looks very clean!

jdchristensen:
Just out of curiosity, how helpful was the displayed category framework for this?

A general comment is that it seems like there are two approaches to this kind of thing that have the same power:  (1) An equivalence of 0-groupoids giving the *dependent* induction principle.  (2) A *natural* equivalence of 0-groupoids giving the *non-dependent* recursion principle.  When I worked on the triple join, (2) was quite manageable, but (1) seemed much harder.  I'm wondering if maybe we should try (2) more often?  OTOH, (1) is closer to the extendable-along material which has proven very useful.

jdchristensen:
Just out of curiosity, how helpful was the displayed category framework for this?

A general comment is that it seems like there are two approaches to this kind of thing that have the same power:  (1) An equivalence of 0-groupoids giving the *dependent* induction principle.  (2) A *natural* equivalence of 0-groupoids giving the *non-dependent* recursion principle.  When I worked on the triple join, (2) was quite manageable, but (1) seemed much harder.  I'm wondering if maybe we should try (2) more often?  OTOH, (1) is closer to the extendable-along material which has proven very useful.

gio256:
I don't know that I really needed displayed categories as much as I needed a more general form of `isgraph_sigma`. 

It will be interesting to see if displayed categories are of any use in proving that this is all natural with respect to `functor_coeq`.

theories/Colimits/CoeqUnivProp.v
```diff
@@ -0,0 +1,155 @@
...
+Require Import WildCat.ZeroGroupoid.
+
+(** Using wild 0-groupoids, the universal property can be proven without funext.  A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
+Section UnivProp.
+  Context {B A : Type} (f g : B -> A) (P : Coeq f g -> Type).
+
+  (** The domain of the equivalence is the fully-applied type of [Coeq_ind]: sections of [P] over [Coeq f g] *)
+  Definition Coeq_ind_type := forall z : Coeq f g, P z.
+
+  Local Instance isgraph_Coeq_ind_type : IsGraph Coeq_ind_type.
```
Alizter, line 22: You can avoid having to declare these instances by declaring `Local Existing Instances isgraph_paths is01cat_paths...` etc. That way the structure can be inferred easily making this file much shorter. After you have done this, we can even consider just keeping it in Coeq.v

Alizter: This also means you can avoid having the definitions for the type as it should be inferreable from the type in this context.

gio256: This worked well, thanks.

Alizter:
Now this seems short enough to be put in Coeq.v. WDYT @jdchristensen?

theories/Colimits/CoeqUnivProp.v
```diff
@@ -12,57 +12,26 @@ Require Import WildCat.Forall.
...
-    apply is01cat_forall; intros; apply is01cat_paths.
-  Defined.
-
-  Local Instance is0gpd_Coeq_ind_map : Is0Gpd Coeq_ind_map.
-  Proof.
-    apply is0gpd_forall; intros; apply is0gpd_paths.
-  Defined.
+  Local Existing Instances isgraph_forall is01cat_forall is0gpd_forall | 1.
+  Local Existing Instances isgraph_total is01cat_total is0gpd_total | 1.
+  Local Existing Instances isgraph_paths is01cat_paths is0gpd_paths | 2.
```
jdchristensen, line 21: I like how this cleans things up.  But now it's unclear which instances are being used where.  Maybe add a few comments here and/or below saying what 0gpd structures are being used?

theories/Colimits/CoeqUnivProp.v
```diff
@@ -12,57 +12,26 @@ Require Import WildCat.Forall.
...
 
   (** The codomain of the equivalence is a sigma-groupoid of this family. *)
-  Definition Coeq_ind_data' (h : Coeq_ind_map)
+  Definition Coeq_ind_data (h : forall a : A, P (coeq a))
     := forall b : B, DPath P (cglue b) (h (f b)) (h (g b)).
 
-  (** We consider [Coeq_ind_data'] to be a displayed 0-groupoid over [Coeq_ind_map], where objects over [h : forall a : A, P (coeq a)] are homotopies [h o f == h o g] and morphisms over [p : h == k] are witnesses that p commutes with the homotopies over [h] and [k]. *)
-  Local Instance isdgraph_Coeq_ind_data' : IsDGraph Coeq_ind_data'.
+  (** We consider [Coeq_ind_data] to be a displayed 0-groupoid, where objects over [h : forall a : A, P (coeq a)] are homotopies [h o f == h o g] and morphisms over [p : h == k] are witnesses that p commutes with the homotopies over [h] and [k]. *)
```
jdchristensen, line 27: I think `are homotopies [h o f == h o g]` is misleading.  Maybe better to say `are dependent paths as defined above`.

jdchristensen (approved):
LGTM, modulo the minor comments.

I would lean towards leaving it separate because it's the only file in the Colimits folder that has any WildCat dependencies.  Extensions.v and Modalities/ReflectiveSubuniverse.v depend on Coeq, and lots depend on the latter, so build parallelization would be reduced.  Of course, if this material gets used in one of those files, that's fine, and then this objection goes away.

I think a pointer to this new file should be left as a comment in Coeq.v.

theories/Colimits/CoeqUnivProp.v
```diff
@@ -0,0 +1,109 @@
...
+Require Import WildCat.EquivGpd.
+Require Import WildCat.Forall.
+Require Import WildCat.Paths.
+Require Import WildCat.ZeroGroupoid.
+
+(** Using wild 0-groupoids, the universal property can be proven without funext. A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
+Section UnivProp.
+  Context {B A : Type} (f g : B -> A) (P : Coeq f g -> Type).
+
+  (** This allows Coq to infer 0-groupoid structures of the form [@isgraph_forall _ P (fun a => isgraph_paths (P a))] whenever the (co)domain is of the form [forall z, P z]. *)
```
gio256, line 19: This is slightly imprecise. The structure mentioned is actually placed on the domain of the equivalence and the first projection of the codomain.

jdchristensen: I find the `whenever the (co)domain` part confusing.  How about:
```suggestion
  (** This allows Coq to infer 0-groupoid structures of the form [@isgraph_forall C P (fun c => isgraph_paths (P c))] on any type of the form [forall c, P c]. *)
```

gio256: Perfect, thanks!

PR #1901, opened 2024-03-21
[Chapter 1 Exercise] Solved

MintChocoManju:
Renamed some functions from Exercise 1.8 to avoid collisions

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -129,17 +129,17 @@ End Book_1_3_sig.
 (** Exercise 1.4 *)
 
 Section Book_1_4.
-  Fixpoint Book_1_4_iter (C : Type) (c0 : C) (cs : C -> C) (n : nat) : C :=
-    match n with
-    | O => c0
-    | S m => cs (Book_1_4_iter C c0 cs m)
-    end.
+  Fixpoint Book_1_4_iter (C : Type) (c0 : C) (cs : C -> C) (n : nat) : C
+    := match n with
+      | O => c0
+      | S m => cs (Book_1_4_iter C c0 cs m)
+      end.
 
-  Definition Book_1_4_rec' (C : Type) (c0 : C) (cs : nat -> C -> C) : nat -> nat * C :=
-    Book_1_4_iter (nat * C) (O, c0) (fun x => (S (fst x), cs (fst x) (snd x))).
+  Definition Book_1_4_rec' (C : Type) (c0 : C) (cs : nat -> C -> C) : nat -> nat * C
+    := Book_1_4_iter (nat * C) (O, c0) (fun x => (S (fst x), cs (fst x) (snd x))).
   
-  Definition Book_1_4_rec (C : Type) (c0 : C) (cs : nat -> C -> C) (n : nat) : C :=
-    snd (Book_1_4_rec' C c0 cs n).
+  Definition Book_1_4_rec (C : Type) (c0 : C) (cs : nat -> C -> C) (n : nat) : C
+    := snd (Book_1_4_rec' C c0 cs n).
 
   Lemma Book_1_4_aux : forall C c0 cs n, fst (Book_1_4_rec' C c0 cs n) = n.
   Proof.
@@ -153,8 +153,8 @@ Section Book_1_4.
     intros C c0 cs n. induction n as [| m IH].
     - simpl. reflexivity.
     - unfold Book_1_4_rec. simpl. rewrite Book_1_4_aux.
-      unfold Book_1_4_rec, Book_1_4_rec' in IH. rewrite IH.
-      reflexivity.
+      unfold Book_1_4_rec, Book_1_4_rec' in IH.
+      exact (ap (cs m) IH).
   Qed.
 End Book_1_4.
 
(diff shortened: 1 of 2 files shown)


contrib/HoTTBookExercises.v
```diff
@@ -157,43 +185,207 @@ End Book_1_5.
...
 
+Section Book_1_7.
+  Definition Book_1_7_id {A : Type} {x y : A} (p : x = y)
+    : (x; 1) = (y; p) :> { a : A & x = a }
+    := match p with 1 => 1 end.
+
+  Definition Book_1_7_ind' {A : Type} (a : A) (C : forall x, (a = x) -> Type)
+    (c : C a 1) (x : A) (p : a = x)
+    : C x p
+    := transport (fun r => C (pr1 r) (pr2 r)) (Book_1_7_id p) c.
```
jdchristensen, line 241: The exercise wants you to only use unbased path induction, but I believe that `transport` is defined via `match` which uses based path induction.  Instead, you could use `paths_ind'`, which is the unbased path induction, to define `Book_1_7_transport` that you use here.

contrib/HoTTBookExercises.v
```diff
@@ -157,43 +185,207 @@ End Book_1_5.
...
 
 (* ================================================== ex:nat-semiring *)
 (** Exercise 1.8 *)
 
-Fixpoint rec_nat' (C : Type) c0 cs (n : nat) : C :=
-  match n with
-    O => c0
-  | S m => cs m (rec_nat' C c0 cs m)
-  end.
+Section Book_1_8.
```
jdchristensen, line 253: It would be worth checked whether all or most of these results are in Spaces/Nat/*.v.  If any are missing, it would be better to add them there.  Then this exercise could just point to the right spots in the library.  (In general, if some results in an exercise are of general interest, it's worth putting them in the main library.)

jdchristensen (commented):
I haven't finished reviewing this, but might not have time for a couple of days.  I pushed some stylistic changes that you could take into account in other parts of this PR.  E.g. putting `:` and `:=` at the start of a line instead of the end; avoiding rewrite if it is trivial to do so; adding line breaks in proofs; using `:= 1` idiom for definitional equalities.  (You could change some Theorems to Definitions to use this more.)

contrib/HoTTBookExercises.v
```diff
@@ -458,7 +458,8 @@ End Book_1_13.
 (* ================================================== ex:without-K *)
 (** Exercise 1.14 *)
 
-(* There is no adequate type family Pi_{x, y} C such that C(x, x, p) is p = refl x definitionally *)
+(* There is no adequate type family C : Pi_{x, y, p} U  such that C(x, x, p)
+  is p = refl x definitionally *)
```
jdchristensen, line 471: Actually, our style is to not put line breaks in comments.

jdchristensen:
@MintChocoManju Thanks for these contributions!

jdchristensen:
Thanks for making updates.  BTW, the library was written over many years by many people, so the style is not consistent.  But the style I've suggested is what we've been using lately.  And, just to be clear, `:` and `:=` can appear within a line.  It's just that *if* there is a line break beside one of these, it should occur before, so the punctuation appears at the start of a line making it visually obvious that the line contains the goal or the proof.

contrib/HoTTBookExercises.v
MintChocoManju: I couldn't find the proofs of these properties in Spaces/Nat/*.v. However, I did find these results in Classes/implementations/peano_naturals.v. Is it appropriate to reference nat_semiring from that file directly in this exercise?

jdchristensen: Yes, I think that's best, rather than including similar proofs in the exercises.  Your proofs are more readable, though.  I've just done a bit of work cleaning up a few things in peano_naturals.v.  Can I push it to this PR?  After that, you could adjust things if you want, but the Classes sub-tree is generally written in a very different style from the rest of the HoTT library.

jdchristensen:
Let us know when the second-half is ready for review.  I'm not sure if I get notified if you just hit the "Ready for review" button.

contrib/HoTTBookExercises.v
MintChocoManju: You can push to this PR directly, or open another PR for refactoring peano_naturals.v, whichever you see fit. In both cases I assume the second-half of 1.8 would just be a reference to nat_semiring.

jdchristensen: I just pushed a commit doing minor cleanup to peano_naturals.v.  I'll let you update the exercises.

contrib/HoTTBookExercises.v
```diff
@@ -275,17 +470,18 @@ End Book_1_13.
...
 
-
+(* There is no adequate type family C : Pi_{x, y, p} U  such that C(x, x, p) is p = refl x definitionally *)
 
 (* ================================================== ex:subtFromPathInd *)
 (** Exercise 1.15 *)
 
-(* concat_A1p? *)
+Definition Book_1_15_paths_rec {A : Type} {C : A -> Type} {x y : A} (p : x = y) : C x -> C y 
+  := match p with 1 => idmap end.
```
jdchristensen, line 479: One could also use `:= transport C p`, which would show the reader what this is called in the library.  But we could also keep it like this, and add a comment saying something like:  `(** This is exactly the definition of [transport] from the Basics.Overture. *)`

contrib/HoTTBookExercises.v
```diff
@@ -275,17 +470,18 @@ End Book_1_13.
 (* ================================================== ex:without-K *)
 (** Exercise 1.14 *)
 
-
+(* There is no adequate type family C : Pi_{x, y, p} U  such that C(x, x, p) is p = refl x definitionally *)
```
jdchristensen, line 473: We process the files with tools that hide comments that open with `(* `.  By default, we should use `(** ` to start comments.  (Again, this isn't consistent within the library so you would have no way of knowing this, but we should follow it for changes and additions.)

jdchristensen (commented):
This LGTM, modulo the semiring comment.

contrib/HoTTBookExercises.v
mikeshulman: Apparently this isn't mentioned in `STYLE.md`.  Maybe it should be?

jdchristensen: Actually, I forgot that back in October I changed things so that `(*` comments *do* appear in the documentation.  I think this is better since they are used in many places.  Now the only difference is the formatting:  `(**` are highlighted in a box, while `(*` don't stand out very much.  So I think what STYLE.md says is correct, and it's just a convention that we use `(**` style for most top-level comments.

jdchristensen: For the record, STYLE.md says:

> Section titles [...]. Other comments are generally written using the style (** This is a comment. *) or (* This is a comment. *), with the latter generally used for inline comments during a proof.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
 
 (** [a + 0 =N= a] *)
-Local Instance add_0_r : RightIdentity@{N N} (plus : Plus nat) 0.
+Local Instance add_0_r : RightIdentity@{N N} (plus : Plus nat) (zero : Zero nat).
 Proof.
-  intros a; induction a as [|a IH].
+  intros a; induction a.
   - reflexivity.
-  - apply (ap S), IH.
+  - apply (ap S), IHa.
```
jdchristensen, line 54: Since the induction hypothesis is used by name, you should retain the `as` clause, because Coq might choose a different name in the future.  (The same applies with other tactics that introduce identifiers.)

jdchristensen:
@MintChocoManju Can you summarize the changes in the last commit?  In the future, if you want to make changes to a file and also reorder things, it's best to do this as separate commits.  The last commit is hard to review because things were moved.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
-hnf. apply (nat_rect@{N} (fun a => forall b, _));[|intros a IHa];
-intros b;induction b as [|b IHb].
-- reflexivity.
-- change (S b = S (b + 0)). apply ap,IHb.
-- apply (ap S),IHa.
-- change (S (a + S b) = S (b + S a)).
-  rewrite (IHa (S b)), <- (IHb ). apply (ap S),(ap S),symmetry,IHa.
+  intros a b c; induction a.
+  - reflexivity.
+  - do 3 (rewrite add_S_l).
```
jdchristensen, line 69: These `rewrite`s aren't the style we like, especially since in this case they are definitional equalities.  While it might not be so important here, in general we try to produce proof terms that are as simple as possible, to help with further reasoning.  And proofs of properties of addition do sometimes come up deep in homotopy-theoretic arguments.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
-simpl_nat.
-- intros _ _;reflexivity.
-- intros. rewrite IHa.
-  rewrite <-(associativity b), (associativity c), (commutativity c),
-  <-(associativity (a*b)), (associativity b).
-  reflexivity.
+  intros a b; induction a.
+  - rewrite add_0_r.
+    reflexivity.
+  - rewrite add_S_l, add_S_r.
```
jdchristensen, line 78: Here too.  Can you check for other places where definitional equalities are used in a rewrite?  (We do have a `rewrite_refl` tactic that uses the form of a lemma to change the goal definitionally.)

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
+Local Instance mul_1_l : LeftIdentity@{N N} (mult : Mult nat) (one : One nat)
+  := add_0_r.
+
+Local Instance mul_0_r : RightAbsorb@{N N} (mult : Mult nat) (zero : Zero nat).
 Proof.
-induction a;simpl_nat;trivial.
-reflexivity.
+  intros a; induction a.
+  - reflexivity.
+  - rewrite mul_S_l.
```
jdchristensen, line 95: Here's another spot.

MintChocoManju:
> @MintChocoManju Can you summarize the changes in the last commit? In the future, if you want to make changes to a file and also reorder things, it's best to do this as separate commits. The last commit is hard to review because things were moved.

@jdchristensen I moved the proofs of the semiring properties to peano_naturals.v. Due to the order in which I proved the properties, I had to reorder the pre-existing proofs.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
 
 (** [a * 1 =N= a]. *)
-Local Instance mul_1_r : RightIdentity (mult : Mult nat) 1.
+Local Instance mul_1_r : RightIdentity@{N N} (mult : Mult nat) (one : One nat).
 Proof.
-  intro a.
-  lhs nrapply mul_S_r.
-  lhs nrapply (ap _ (mul_0_r a)).
+  intros a.
+  rewrite mul_S_r, mul_0_r.
```
jdchristensen, line 114: Please use the two `lhs` lines that were there before, since they avoid rewrites.  They are the style we use when it is easy to avoid rewrite.  (`rewrite` produces terms that are harder to reason about.)

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
-(** [1 * a =N= a]. *)
-Local Instance mul_1_l : LeftIdentity (mult : Mult nat) 1 := add_0_r.
+(** [(a + b) * c =N= a * c + b * c]. *)
+Local Instance add_mul_distr_r
+  : RightDistribute@{N} (mult : Mult nat) (plus : Plus nat).
+Proof.
+  intros a b c; induction a.
+  - reflexivity.
+  - rewrite add_S_l.
+    do 2 (rewrite mul_S_l).
```
jdchristensen, line 125: Definitional.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
 
 Local Instance mul_comm : Commutative@{N N} (mult : Mult nat).
 Proof.
-hnf. apply (nat_rect@{N} (fun a => forall b, _));[|intros a IHa];simpl_nat.
-- intros;apply symmetry,mul_0_r.
-- intros b;rewrite IHa. rewrite mul_S_r,<-IHa. reflexivity.
+  intros a b; induction a.
+  - rewrite mul_0_r.
+    reflexivity.
+  - rewrite mul_S_l, mul_S_r.
```
jdchristensen, line 135: Definitional.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -41,106 +41,118 @@ Ltac simpl_nat :=
...
-  lhs nrapply (ap _ (mul_0_r a)).
+  intros a.
+  rewrite mul_S_r, mul_0_r.
   apply add_0_r.
 Qed.
 
-(** [1 * a =N= a]. *)
-Local Instance mul_1_l : LeftIdentity (mult : Mult nat) 1 := add_0_r.
+(** [(a + b) * c =N= a * c + b * c]. *)
+Local Instance add_mul_distr_r
```
jdchristensen, line 119: You should prove `add_mul_distr_l` first, since it is shown in `plus_mult_distr_r` that left distributivity implies right distributivity, and it's an instance.  (Classes/theory/rings.v)

jdchristensen (commented):
Thanks for cleaning this up!  You improved many proofs and fixed the type-class inference problem I had.  I've left some comments.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -165,6 +143,19 @@ Proof.
...
+Local Instance mul_assoc : Associative@{N} (mult : Mult nat).
+Proof.
+ intros a b c; induction a as [| a IHa].
+  - reflexivity.
+  - change (b * c + a * (b * c) = (b + a * b) * c).
+    rhs rapply mul_comm.
+    rhs rapply add_mul_distr_l.
+    rhs rapply (ap (fun x => x + _) (mul_comm _ _)).
+    rapply (ap (plus _)).
+    rhs rapply mul_comm.
```
jdchristensen, line 155: I see now that right distributivity is useful here, and you can't use the result from rings.v, since it requires that you already have a semi-ring.  That result should be weakened, but rather than open up that can of worms, I think it's best to just prove right distributivity here.  I'll push a commit that does this.  Sorry to misdirect you.

jdchristensen:
I think this is good to go.  Thanks again for doing this and improving the library.  I'll merge later today unless there are any other comments.

Alizter (approved):

PR #1902, opened 2024-03-29
(2,1)-cat structure for paths

Alizter:
We show that types are themselves (2,1)-categories in the obvious way. This can be useful when building higher morphisms for various structures. It also serves as a guiding principle for what is missing in PathGroupoids. I've added some TODO's so we can eventually add such lemmas in.

### DIFF
--- a/theories/WildCat/Paths.v
+++ b/theories/WildCat/Paths.v
@@ -92,7 +92,7 @@ Proof.
     by destruct h, s, r, q.
   - intros a b p q h; cbn.
     apply moveL_Mp.
-    (* TODO: the naturliaty condition in path groupoids is not in the most natural form. *)
+    (* TODO: the naturality condition in PathGroupoids.v is not in the most natural form. *)
     lhs nrapply concat_p_pp.
     exact (whiskerR_p1 h).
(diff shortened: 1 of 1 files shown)


theories/WildCat/Paths.v
```diff
@@ -22,3 +22,86 @@ Proof.
...
+    by destruct h, s, r, q.
+  - intros a b c d q r s t h.
+    (* TODO: missing from path groupoids *)
+    by destruct h, s, r, q.
+  - intros a b c d q r s t h.
+    (* TODO: missing from path groupoids *)
+    by destruct h, s, r, q.
+  - intros a b p q h; cbn.
+    apply moveL_Mp.
+    (* TODO: the naturliaty condition in path groupoids is not in the most natural form. *)
```
jdchristensen, line 95:
```suggestion
    (* TODO: the naturality condition in PathGroupoids.v is not in the most natural form. *)
```
If these comments are left in place, I recommend "PathGroupoids.v" instead of "path groupoids".

jdchristensen:
Looks good.  I think it's fine to merge as is (with updated comments), or to add those results to PathGroupoids.v.

PR #1905, opened 2024-04-07
remove bg_scope

Alizter:
This scope was introduced for the `B` notation, however it was later converted to an abbreviation IIRC making leaving it unused.


jdchristensen:
If nothing else is being guarded by this scope, then it makes sense to remove it.  You'll also need to remove it from EMSpace.v.

Alizter:
Yes, I missed the use in EMSpace.v. It was redundant anyway so I removed it.

jdchristensen (approved):

PR #1907, opened 2024-04-08
I-indexed coproducts in pType

### LINKED ISSUE #1890: Show that pType has all coproducts
We can show that pType has all coproducts in Wedge.v. There is a bit of a choice with the binary vs indexed wedge however. I think it would be best to keep these separate and not prove HasBinaryCoproducts by inferring the general indexed wedge. This is how we do it for products and coproducts in Type for instance.

Alizter:
We show that pType has all coproducts.
- closes #1890

### DIFF
--- a/theories/Homotopy/Wedge.v
+++ b/theories/Homotopy/Wedge.v
@@ -194,14 +194,14 @@ Proof.
     exact pt.
 Defined.
 
+Definition fwedge_in' (I : Type) (X : I -> pType)
+  : forall i, X i $-> FamilyWedge I X
+  := fun i => Build_pMap _ _ (fun x => pushl (i; x)) (pglue pt).
+
 (** We have an inclusion map [pushl : sig X -> FamilyWedge X].  When [I] is pointed, so is [sig X], and then this inclusion map is pointed. *)
 Definition fwedge_in (I : pType) (X : I -> pType)
-  : psigma (pointed_fam X) $-> FamilyWedge I X.
-Proof.
-  snrapply Build_pMap.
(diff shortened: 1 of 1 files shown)


theories/Homotopy/Wedge.v
```diff
@@ -194,14 +194,14 @@ Proof.
     exact pt.
 Defined.
 
+Definition fwedge_in' (I : Type) (X : I -> pType)
+  : forall i, X i $-> FamilyWedge I X
+  := fun i => Build_pMap _ _ (fun x => pushl (i; x)) (pglue pt).
```
jdchristensen, line 199:
```suggestion
  := fun i => Build_pMap _ _ (fun x => pushl (i; x)) (pglue i).
```
`pt` was confusing there.  It was the base point of the pointed type `[I, i]`, which is not obvious.

theories/Homotopy/Wedge.v
```diff
@@ -230,6 +230,39 @@ Proof.
...
+    symmetry.
+    exact (Pushout_rec_beta_pglue Z _ (unit_name pt) (fun i => point_eq (f i)) _).
+  - intros Z f g h.
+    snrapply Build_pHomotopy.
+    + snrapply Pushout_ind.
+      * intros [i x].
+        nrapply h.
+      * intros [].
+        exact (point_eq _ @ (point_eq _)^).
+      * intros i.
```
jdchristensen, line 255:
```suggestion
      * intros i; cbn.
```

jdchristensen (approved):

PR #1908, opened 2024-04-08
Add missing coherences from WildCat/Paths.v to PathGroupoids.v

### LINKED ISSUE #1903: Add missing lemmas in PathGroupoids
There are some missing lemmas in PathGroupoids.v that are needed for the (2,1)-cat structure found in WildCat/Paths.v. These should be added in. There is also the question of adapting the existing triangle and pentagon lemmas to better fit the way we write them here.

Alizter:
Fixes #1903.

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -1264,7 +1264,7 @@ Proof.
   destruct p, r, q. reflexivity.
 Defined.
 
-(** Naturality of [concat_p_pp] in left-mode argument. *)
+(** Naturality of [concat_p_pp] in left-most argument. *)
 Definition concat_p_pp_nat_l {A} {w x y z : A}
   {p q : w = x} (r : p = q) (s : x = y) (t : y = z)
   : whiskerR r (s @ t) @ concat_p_pp q s t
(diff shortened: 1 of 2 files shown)


theories/Basics/PathGroupoids.v
```diff
@@ -1264,6 +1264,33 @@ Proof.
   destruct p, r, q. reflexivity.
 Defined.
 
+(** Naturality of [concat_p_pp] in left-mode argument. *)
```
jdchristensen, line 1267:
```suggestion
(** Naturality of [concat_p_pp] in left-most argument. *)
```

theories/Basics/PathGroupoids.v
```diff
@@ -1264,6 +1264,33 @@ Proof.
...
+Definition concat_p_pp_nat_l {A} {w x y z : A}
+  {p q : w = x} (r : p = q) (s : x = y) (t : y = z)
+  : whiskerR r (s @ t) @ concat_p_pp q s t
+    = concat_p_pp p s t @ whiskerR (whiskerR r s) t.
```
jdchristensen, line 1271: For all three of these, the argument names could be better.  Maybe always use p, q, r for the 1-paths, with the duplicated one getting a prime.  And use h for the 2-path.  So in this case we'd have h : p = p', q, r.  In the next we'd have p, h : q = q', r.  And in the third, p, q, h : r = r'.

jdchristensen (approved):

Alizter:
I made a typo with the renaming, fixing now. Renaming is not easy. :innocent:

PR #1909, opened 2024-04-08
chore: update nix flake

Alizter:


jdchristensen (approved):
I don't know what this does, but I trust you know what you are doing.

Alizter:
If you're interested, this lets me setup a developer/working environment quickly using Nix. I just type `nix develop` in the project and coq and dune are all setup correctly. We also use it in the CI and it is the first job to finish.

PR #1910, opened 2024-04-08
bump to dune 3.13

Alizter:
We have some better support for sharing flags in Dune 3.13 allowing us to reduce the size of the coq.theory stanzas.

### DIFF
--- a/coq-hott.opam
+++ b/coq-hott.opam
@@ -16,7 +16,7 @@ license: "BSD-2-Clause"
 homepage: "http://homotopytypetheory.org/"
 bug-reports: "https://github.com/HoTT/HoTT/issues"
 depends: [
-  "dune" {>= "3.8"}
+  "dune" {>= "3.13"}
   "coq" {>= "8.18.0"}
   "odoc" {with-doc}
 ]
--- a/contrib/dune
+++ b/contrib/dune
@@ -1,6 +1,4 @@
 (coq.theory
  (name HoTT.Contrib)
  (package coq-hott)
- (flags -noinit -indices-matter -color on)
- (coqdoc_flags :standard --interpolate --utf8 --no-externals --parse-comments)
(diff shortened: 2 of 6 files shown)


jdchristensen (approved):

Alizter:
I'm going to hold off on this as there is a bug in 3.14 that makes compiling Coq slow. This is fixed in 3.15 but that was only recently released. I'll wait a little longer for this as it isn't critical so that 3.15 is more widespread.

coq-hott.opam
```diff
@@ -16,7 +16,7 @@ license: "BSD-2-Clause"
 homepage: "http://homotopytypetheory.org/"
 bug-reports: "https://github.com/HoTT/HoTT/issues"
 depends: [
-  "dune" {>= "3.8"}
+  "dune" {>= "3.13"}
```
JasonGross, line 19: Why not
```suggestion
  "dune" {>= "3.13" & != "3.14" & build}
```
?

Alizter: I'm not certain when this issue was introduced, I haven't had the time to look into it yet. Presumably something like this would work. However this is best done from the dune-project file since we generate this opam ifle.

It's not that the build doesn't work, but rather rebuilds are slower because the bug causes coqdep calls not to be cached properly making them expensive and numerous. Forcing people who install coq-hott to avoid this dune version would be pretty annoying, so I would probably not want to do it.

JasonGross: Also, I guess we don't want `build` because of https://github.com/ocaml/opam/issues/3850

Alizter:
The bug in Dune has been fixed and the fix is available in 3.15.2. See https://github.com/HoTT/Coq-HoTT/issues/1924 for a more detailed explanation of the issue.

jdchristensen:
LGTM

PR #1912, opened 2024-04-10
noncommutative rings

Alizter:
We generalise much of the theory of commutative rings to the noncommutative case and derive the appropriate commutative results from that.

The theory of ideals was expanded to include left, right and two-sided ideals together with new documentation.

The chinese remainder theorem generalised easily to the noncommutative case.


theories/Algebra/Rings/CRing.v
```diff
@@ -3,432 +3,198 @@ Require Import Spaces.Nat.Core.
...
+(** ** Ideals in commutative rings *)
+  
+Section IdealCRing.
+  Context {R : CRing}.
+  
+  Import Ideal.Notation.
+  Local Open Scope ideal_scope.
+
+  (** Ideal products are commutative in commutative rings. *)
+  Lemma ideal_product_comm (I J : Ideal R) : I ⋅ J ↔ J ⋅ I.
```
jdchristensen, line 74: I thought we avoided unicode in the main library.  And it's not clear to me why ↔ would denote equality of ideals.

Also, maybe the lemma inside the proof should be a top-level statement?

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+Proof.
+  intros [f e].
+  snrapply Build_RingIsomorphism.
+  { snrapply Build_RingHomomorphism.
+    1: exact f^-1.
+    exact _. }
+  exact _.
+Defined.
+
+(** Ring isomorphisms are a reflexive relation *)
```
jdchristensen, line 144:
```suggestion
(** Ring isomorphism is a reflexive relation *)
```

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+Global Instance reflexive_ringisomorphism : Reflexive RingIsomorphism
+  := fun x => Build_RingIsomorphism _ _ (rng_homo_id x) _.
+
+(** Ring isomorphisms are a symmetric relation *)
```
jdchristensen, line 148:
```suggestion
(** Ring isomorphism is a symmetric relation *)
```

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+Global Instance symmetry_ringisomorphism : Symmetric RingIsomorphism
+  := fun x y => rng_iso_inverse.
+
+(** Ring isomorphisms are a transitive relation *)
```
jdchristensen, line 152:
```suggestion
(** Ring isomorphism is a transitive relation *)
```

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+  := Build_RingHomomorphism map
+      (Build_IsSemiRingPreserving _ (grp_homo_ishomo _ _ map) H).
+
+(** We can construct a ring isomorphism from a group isomorphism that preserves multiplication *)
+Definition Build_RingIsomorphism'' (A B : Ring) (e : GroupIsomorphism A B)
+  {H : IsMonoidPreserving (Aop:=ring_mult) (Bop:=ring_mult) (Aunit:=one) (Bunit:=one) e}
+  : RingIsomorphism A B
+  := @Build_RingIsomorphism' A B e (Build_IsSemiRingPreserving e _ H).
+
+(** Here is an alternative way to build a commutative ring using the underlying abelian group. *)
```
jdchristensen, line 177:
```suggestion
(** Here is an alternative way to build a ring using the underlying abelian group. *)
```

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+Definition Build_Ring' (R : AbGroup)
+  `(Mult R, One R, LeftDistribute R mult (@group_sgop R), RightDistribute R mult (@group_sgop R))
+  (iscomm : @IsMonoid R mult one)
```
jdchristensen, line 180:
```suggestion
  (ismonoid : @IsMonoid R mult one)
```

theories/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,411 @@
...
+
+(** ** Ring movement lemmas *)
+
+Section RingMovement.
+
+  (** We adopt a similar naming convention to the [moveR_equiv] style lemmas that can be found in Types.Paths. *)
+
+  Context {R : Ring} {x y z : R}.
+
+  Definition rng_moveL_Mr : - y + x = z <~> x = y + z := @grp_moveL_Mg R x y z.
```
jdchristensen, line 192: Since a ring has an underlying (abelian) group, do we really need to give all of the group lemmas new names?

theories/Algebra/Rings/Ideal.v
```diff
@@ -122,251 +284,608 @@ Definition ideal_sum_ind {R : CRing} (I J : Ideal R)
...
-(** Note that this is not an ideal, but we can fix this later. *)
-Inductive ideal_product_naive_type {R : CRing} (I J : Ideal R) : R -> Type :=
-| ipn_in : forall x y, I x -> J y -> ideal_product_naive_type I J (x * y)
-.
+(** First we form the "naive" product of ideals { a * b | a ∈ I /\ b ∈ J }. Note that this is not an ideal, but we can fix this. *)
+Inductive ideal_product_naive_type {R : Ring} (I J : Subgroup R) : R -> Type :=
+| ipn_in : forall x y, I x -> J y -> ideal_product_naive_type I J (x * y).
 
-(** Now we can close this under addition to get the product ideal. *)
+(** We instead consider the ideal generated by this naive product. *)
```
jdchristensen, line 291:
```suggestion
(** We instead consider the subgroup generated by this naive product. *)
```
(The subgroup generated by a subset and the ideal generated by a subset are not necessarily the same.  It just happens that in this case they are.)

theories/Algebra/Rings/Ideal.v
```diff
@@ -122,251 +284,608 @@ Definition ideal_sum_ind {R : CRing} (I J : Ideal R)
...
+End QuotientIdeal.
+
+(** *** Annihilator *)
+
+Section AnnihilatorIdeal.
+  Context `{Funext}.
+
+  (** The left annihilator of a subgroup is the set of elements that annihilate the subgroup with left multiplication. *)
+  Definition subgroup_ideal_left_annihilator {R : Ring} (I : Subgroup R)
+    : Subgroup R.
```
jdchristensen, line 720:
```suggestion
  (** The left annihilator of a subset is the set of elements that annihilate the subgroup with left multiplication. *)
  Definition subgroup_ideal_left_annihilator {R : Ring} (S : R -> Type)
    : Subgroup R.
```
It works for any subset, in particular, a single element.

theories/Algebra/Rings/Ideal.v
```diff
@@ -122,251 +284,608 @@ Definition ideal_sum_ind {R : CRing} (I J : Ideal R)
...
+  Proof.
+    snrapply Build_Subgroup'.
+    - exact (fun r => forall x, I x -> r * x = ring_zero).
+    - exact _.
```
jdchristensen, line 724: You can avoid Funext by defining:
```suggestion
    - exact (fun r => merely (forall x, I x -> r * x = ring_zero)).
    - exact _.
```
Then, below, you have to use `rapply Trunc_ind`, and do the usual trick to avoid Funext there.

theories/Algebra/Rings/Ideal.v
```diff
@@ -122,251 +284,608 @@ Definition ideal_sum_ind {R : CRing} (I J : Ideal R)
...
+      rewrite (p r s).
+      rewrite rng_mult_negate_l.
+      rewrite (q r s).
+      rewrite <- rng_mult_negate.
+      rewrite rng_mult_zero_r.
+      apply left_identity.
+  Defined.
+  
+  (** The left annihilator is a left ideal of the ring. *)
+  Global Instance isleftideal_ideal_left_annihilator {R : Ring} (I : Subgroup R)
```
jdchristensen, line 738:
```suggestion
  Global Instance isleftideal_ideal_left_annihilator {R : Ring} (S : R -> Type)
```

theories/Algebra/Rings/Ideal.v
```diff
@@ -122,251 +284,608 @@ Definition ideal_sum_ind {R : CRing} (I J : Ideal R)
...
+    : IsLeftIdeal (subgroup_ideal_left_annihilator I).
+  Proof.
+    intros r x p s i.
+    rewrite <- rng_mult_assoc, (p s i).
+    apply rng_mult_zero_r.
+  Defined.
+  
+  (** When the subgroup is a left ideal then the annihilator is also a right ideal. *)
```
jdchristensen, line 746: This could also be done for a subset closed under left-multiplication, but maybe that's not worth it.

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -143,25 +143,43 @@ Section upper_classes.
...
+  Proof.
+    intros H.
+    econstructor; try exact _.
+    intros a b c.
+    lhs rapply commutativity.
+    lhs rapply distribute_l.
+    f_ap; apply commutativity.
+  Defined.
+  
   (* For now, we follow CoRN/ring_theory's example in having Ring and SemiRing
```
jdchristensen, line 170: Delete this comment?

jdchristensen (approved):
It's great to have this in the library!

In addition to the minor comments below, my only global comment is that it would be good to introduce the opposite ring R^op, which has the multiplication reversed.  I think this could be used to remove a lot of the duplication in the theory of ideals, with the results for right ideals just being instances of the results for left ideals, using the opposite multiplication.

theories/Algebra/Rings/CRing.v
Alizter: Way back when I wrote this I think I chose `<->` as a "biimplication" betwen predicates which is what equiality of ideals is equivalent to. I believe I decided to use unicode here for readability purposes. I didn't like the clunkiness without the notation, it made it hard to parser the results.

I'm happy to change it, but I honestly believe it is better this way. The notation is local anyway, meaning that you won't see it unless you specifically ask for `Import Ideal.Notation.`. If you know some ring theory and need some lemmas about ideals, this file will be easier to read compared to if we hadn't used unicode.

theories/Algebra/Rings/CRing.v
```diff
@@ -3,432 +3,198 @@ Require Import Spaces.Nat.Core.
...
-
-Declare Scope ring_scope.
+(** * Commutative Rings *)
 
 Local Open Scope ring_scope.
-(** We want to print equivalences as [≅]. *)
 Local Open Scope wc_iso_scope.
 
-(** A commutative ring consists of the following data *)
+(** A commutative ring consists of the following data: *)
```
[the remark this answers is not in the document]
Alizter: I actually tried this first, but I couldn't find an easy way to share that definition with mathclasses. All of mathclasses is based on commutative rings and semirings and the algebraic hierarchy chosen there is a lot of work to pull apart. I didn't want to spend time fixing mathclasses code, so I opted for this definition.

I might try to change it again in the future, but in practice we have `Build_CRing'` which is what we want.

theories/Algebra/Rings/CRing.v
```diff
@@ -3,432 +3,198 @@ Require Import Spaces.Nat.Core.
...
-  cring_mult : Mult cring_type;
-  cring_zero : Zero cring_type;
-  cring_one  : One  cring_type;
-  cring_negate : Negate cring_type;
-  cring_isring : IsRing cring_type;
+  (** An underlying abelian group. *)
+  cring_abgroup :> AbGroup;
+  (** A multiplication operation. *)
+  cring_mult :: Mult cring_abgroup;
+  (** A multiplicative identity called [one]. *)
```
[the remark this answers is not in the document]
Alizter: The field of the typeclasse `One` is called `one` which I why I mentioned it. I'll remove any reference to the name anyway as it's not very useful.

theories/Algebra/Rings/CRing.v
```diff
@@ -3,432 +3,198 @@ Require Import Spaces.Nat.Core.
...
+  cring_one :: One cring_abgroup;
+  (** Such that all they all satisfy the axioms of a ring. *)
```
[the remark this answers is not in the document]
Alizter: It's interesting to see what you think not come out as what you type sometimes. Fixed.

theories/Algebra/Rings/CRing.v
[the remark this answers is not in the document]
jdchristensen: Ok, makes sense.

theories/Algebra/Rings/CRing.v
Alizter: Ah but I got confused which file I was looking at. Since this isn't the main file about ideals, it is perhaps more appropriate to not use unicode. On the other hand, the reasons above still hold true. Also these results have essentially been plucked from Ideal.v.

jdchristensen: I'm ok with it.  But since this is the first place it is used in this file, maybe put a reminder in a comment?  Also, factoring out the inner lemma would make it more clear what this means.

theories/Algebra/Rings/Ideal.v
jdchristensen: BTW, we should make a lemma that encodes this trick, since we use it so often.  I guess it would produce `Trunc n X -> Trunc n Y -> Z` given that `Z` is n-truncated.  In fact, I think it works for any RSU `O`.  It could be named `O_ind11` or `O_biind`.

Alizter: Nice trick!

Alizter: I think we've typically called such things `O_rec2`. I'll make an issue about this.

Alizter:
> In addition to the minor comments below, my only global comment is that it would be good to introduce the opposite ring R^op, which has the multiplication reversed. I think this could be used to remove a lot of the duplication in the theory of ideals, with the results for right ideals just being instances of the results for left ideals, using the opposite multiplication.

Yes, this is something I noticed when writing the theory about modules which will appear in a later PR. There I only considered left modules and added a note about R^op.

theories/Algebra/Rings/Ideal.v
Alizter: Actually scratch the issue. I've just noticed the same trick can be done for quotient ideals completely removing funext from the file. This idiom occurs again, so I'll see if I can abstract it out.

theories/Algebra/Rings/Ring.v
Alizter: It can be useful since it keep the + operation and doesn't unfold it into a sg_op.

theories/Algebra/Rings/Ideal.v
Alizter: Hmm actually do we need this? I can just use `strip_truncations`.

theories/Algebra/Rings/Ring.v
Alizter: The group lemmas can be used for abelian groups, but it always unfold + to sg_op.

theories/Algebra/Rings/Ideal.v
Alizter: Of course! Interesting that they conincide here.

theories/Algebra/Rings/Ideal.v
Alizter: That seems like a lot of hassle. I haven't used annihilators at all really. I only have a formalization of the Zariski site in HoTT that used it. https://github.com/Alizter/AlgGeom

Alizter: I'll perhaps reconsider when we look at modules. There it might be worth introducing a multiplication preserving subset class.

Alizter: Hmm looks like that repo doesn't have anything about sites in it, that's funny because I seem to recall doing it. However it does have a construction of the localization of a ring that might be of independent interest.

theories/Algebra/Rings/Ideal.v
```diff
@@ -587,236 +587,257 @@ Defined.
...
 (** *** Annihilator *)
 
-Section AnnihilatorIdeal.
-  Context `{Funext}.
+(** The left annihilator of a subset is the set of elements that annihilate the subgroup with left multiplication. *)
+Definition subgroup_ideal_left_annihilator {R : Ring} (S : R -> Type)
+  : Subgroup R.
+Proof.
+  snrapply Build_Subgroup'.
+  (** We use [merely] here to avoid [funext] it isn't necessary as the target is already a proposition. *)
```
jdchristensen, line 731: *If we assume [Funext]*, then it isn't necessary.

theories/Algebra/Rings/Ring.v
jdchristensen: Hmm, that's annoying, but I guess it's a good reason to duplicate things.

theories/Algebra/Rings/Ideal.v
jdchristensen: I guess `strip_truncations` works fine.  I guess what we're really using a lot here is that modalities are bifunctors, which would avoid the `strip_truncations` and the `tr`.  But it's fine how it is.

jdchristensen:
LGTM, after that one comment is fixed.

PR #1915, opened 2024-04-12
Symmetric Monoidal Categories

Alizter:
@jdchristensen here is a definition of symmetric monoidal category together with a partially finished twist construction. By this I mean using a `swap` and `twist` map we get associativity, triangle and hexagon with appropriate assumptions. <s>Only the pentagon is left.</s> Pentagon is done.

Note that this cannot be applied to `JoinAssoc` as is as there is a difference between `fmap11` and `functor_join`. This might be a good reason to redefined bifunctor like I suggested in #1883. It's not too difficult to wrangle it to fit however.

- [x] finish cartesian symmetric monoidal structure
- [ ] optimize proofs in Monoidal.v
- [x] split bifunctor work into another PR
- [ ] refactor twist class? (didn't turn out to be easy)
- [ ] Use in joins?
- [x] modify long form notation of `$==`.

### DIFF
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -1,8 +1,10 @@
-Require Import Basics.Overture.
-Require Import Types.Forall.
-Require Import WildCat.Core WildCat.Bifunctor WildCat.Prod WildCat.Equiv WildCat.NatTrans.
+Require Import Basics.Overture Basics.Tactics Types.Forall.
+Require Import WildCat.Core WildCat.Bifunctor WildCat.Prod WildCat.Equiv.
+Require Import WildCat.NatTrans WildCat.Square.
 
-(** * Monoidal 1-categories *)
+(** * Monoidal Categories *)
+
+(** In this file we define monoidal categories and symmetric monoidal categories. *)
 
 (** ** Typeclasses for common diagrams *)
 
@@ -12,10 +14,10 @@ Require Import WildCat.Core WildCat.Bifunctor WildCat.Prod WildCat.Equiv WildCat
 
 (** A natural equivalence witnessing the associativity of a bifunctor. *)
 Class Associator {A : Type} `{HasEquivs A}
-  (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F} := {   
+  (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F} := {
   (** An isomorphism [associator] witnessing associativity of [F]. *)
   associator a b c : F a (F b c) $<~> F (F a b) c;
-  
+
   (** The [associator] is a natural isomorphism. *)
   is1natural_associator_uncurried
     :: Is1Natural
@@ -56,16 +58,52 @@ Arguments triangle_identity {A _ _ _ _ _} F {_ _ _} unit {_}.
 
 Class PentagonIdentity {A : Type} `{HasEquivs A}
   (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F, !Associator F}
-  (unit : A) `{!LeftUnitor F unit, !RightUnitor F unit}
-  (** The pentagon identity for an associator and unitors. *)
+  (** The pentagon identity for an associator. *)
   := pentagon_identity a b c d
     : associator (F a b) c d $o associator a b (F c d)
       $== fmap10 F (associator a b c) d $o associator a (F b c) d
         $o fmap01 F a (associator b c d).
 Coercion pentagon_identity : PentagonIdentity >-> Funclass.
-Arguments pentagon_identity {A _ _ _ _ _} F {_ _ _} unit {_}.
+Arguments pentagon_identity {A _ _ _ _ _} F {_ _ _}.
 
-(** ** Definition *)
+(** *** Braiding *)
+
+Class Braiding {A : Type} `{Is1Cat A}
+  (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F} := {
+  (** A morphism [braid] witnessing the symmetry of [F]. *)
+  braid a b : F a b $-> F b a;
+  (** The [braid] is a natural transformation. *)
+  is1natural_braiding_uncurried
+    : Is1Natural
+      (uncurry F)
+      (uncurry (flip F))
+      (fun '(a, b) => braid a b);
+}.
+Coercion braid : Braiding >-> Funclass.
+Arguments braid {A _ _ _ _ F _ _ _} a b.
+
+Class SymmetricBraiding {A : Type} `{Is1Cat A}
+  (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F} := {
+  braiding_symmetricbraiding :: Braiding F;
+  braid_braid : forall a b, braid a b $o braid b a $== Id (F b a);
+}.
+(** We could have used [::>] in [braiding_symmetricbraiding] instead however due to bug https://github.com/coq/coq/issues/18971 the coercion isn't registered, so we have to register it manually instead. *)
+Coercion braiding_symmetricbraiding : SymmetricBraiding >-> Braiding.
+Arguments braid_braid {A _ _ _ _ F _ _ _} a b.
+
+(** *** Hexagon identity *)
+
+Class HexagonIdentity {A : Type} `{HasEquivs A}
+  (F : A -> A -> A)
+  `{!Is0Bifunctor F, !Is1Bifunctor F, !Associator F, !Braiding F}
+  (** The hexagon identity for an associator and a braiding. *)
+  := hexagon_identity a b c
+    : fmap10 F (braid b a) c $o associator b a c $o fmap01 F b (braid c a)
+      $== associator a b c $o braid (F b c) a $o associator b c a.
+Coercion hexagon_identity : HexagonIdentity >-> Funclass.
+Arguments hexagon_identity {A _ _ _ _ _} F {_ _}.
+
+(** ** Monoidal Categories *)
 
 (** A monoidal 1-category is a 1-category with equivalences together with the following: *)
 Class IsMonoidal (A : Type) `{HasEquivs A}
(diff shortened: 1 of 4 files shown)


theories/WildCat/Monoidal.v
```diff
@@ -89,3 +128,202 @@ Class IsMonoidal (A : Type) `{HasEquivs A}
...
+    (twist_twist : forall a b c, twist a b c $o twist b a c $== Id _)
+    (twist_nat : forall a a' b b' c c' (f : a $-> a') (g : b $-> b') (h : c $-> c'),
+      twist a' b' c' $o fmap11 cat_tensor f (fmap11 cat_tensor g h)
+      $== fmap11 cat_tensor g (fmap11 cat_tensor f h) $o twist a b c)
+      
+    {left_unitor : LeftUnitor cat_tensor cat_tensor_unit}
+    {right_unitor : RightUnitor cat_tensor cat_tensor_unit}
+  
+    (swap_unitor : forall a, cate_fun (left_unitor a)
+      $== right_unitor a $o swap cat_tensor_unit a)
```
Alizter, line 175: We could even drop the requirement of a left unitor and take this as the definition like we do for join.

Alizter:
I've wrangled with the pentagon proof, cancelling things where possible and making sure no naturality squares remain. The goal I've ended up with looks like:

```coq
    ((τ (a ⊗ b) d c ∘ σ (d ⊗ c) (a ⊗ b)) ∘ τ a (d ⊗ c) b) ∘ a ⊗R σ b (d ⊗ c)
$== ((d ⊗R σ c (a ⊗ b) ∘ d ⊗R τ a c b) ∘ τ a d (c ⊗ b)) ∘ (a ⊗R (d ⊗R σ b c) ∘ a ⊗R τ b d c)
```

I'm not sure if it can be simplified further. The double functor application `(a ⊗R (d ⊗R σ b c)` looks like it might commute by naturality of twist and go somewhere, but I haven't been able to make progress on that. This might be the pentagon for twist that we are looking for, but it doesn't look great. I still feel like there is more that could be done here.

Alizter:
> I'll be taking a look at this PR sometime soon. Also, this paper [arxiv.org/abs/2404.08163](https://arxiv.org/abs/2404.08163) just appeared, and I wonder if anything from it can be used in this context.

Yes I saw that. I'm aware of some of the work from before. They have a nice way of visualising their proof calculus in coq-lsp as string diagrams which is what that work is based on.

ping @bhaktishh, how much work would it take to adapt the visualizer for Coq-HoTT? The tldr is that we have a library that doesn't use the coq standard library. I'm wondering if we can write a small plugin that can directly interface with the vscode extension and render diagrams in our categories.

bhaktishh:
Thanks for the ping! Looping in @lczielinski @wjbs @adrianleh @caldwellb here for discussions about the library itself.

With respect to the visualizer, what prerequisites on opam are you worried about? The visualizer itself should work with `coq-lsp` as its only dependency. The library alone has no dependencies, it is just the examples that are implemented using other, larger libraries -- are these what you are referring to? 

With respect to adapting for Coq-HoTT, it depends -- are you able to instantiate the typeclasses in the `vicar` library? If yes, then you're automatically able to make use of the visualization. If not, we can come up with some translation from your syntax to vicar's, as in theory they should be compatible.

Alizter:
@bhaktishh Thanks for the explanation. I had a quick look at https://github.com/inQWIRE/ViCaR/tree/main/ViCaR/Classes and it seems the definition for the classes that you've chosen are the setoid based ones. In that case, I don't see any issue adapting our "wild cat" definitions to instantiate those. I'll create an issue so that we can track progress with setting this up. I might not have the time to do it now, but its definitely cool to see that it might work.

theories/WildCat/Monoidal.v
```diff
@@ -65,7 +67,44 @@ Class PentagonIdentity {A : Type} `{HasEquivs A}
 Coercion pentagon_identity : PentagonIdentity >-> Funclass.
 Arguments pentagon_identity {A _ _ _ _ _} F {_ _ _} unit {_}.
 
-(** ** Definition *)
+(** *** Symmetors or Braiding *)
+
+Definition is0bifunctor_flip {A B C : Type}
+  (F : A -> B -> C) `{Is0Bifunctor A B C F} : Is0Bifunctor (flip F).
```
jdchristensen, line 73: I guess this is just here temporarily and will move elsewhere.

theories/WildCat/Monoidal.v
```diff
@@ -65,7 +67,44 @@ Class PentagonIdentity {A : Type} `{HasEquivs A}
...
+  - exact _.
+  - exact _.
+Defined.
+
+Hint Immediate is0bifunctor_flip : typeclass_instances.
+
+Class Symmetor {A : Type} `{HasEquivs A}
+  (F : A -> A -> A) `{!Is0Bifunctor F, !Is1Bifunctor F} := {
+  (** An isomorphism [symmetor] witnessing the symmetry of [F]. *)
+  symmetor a b : F a b $<~> F b a;
```
jdchristensen, line 85: The symmetry isomorphism needs to also satisfy that `symmetor a b $o symmetor b a $== Id _`.  Without this condition, you get a braided monoidal category.

BTW, this condition (of course) implies that the symmetor is an equivalence.  So we could drop the condition that it be an equivalence and just assume this instead.  But if we also want to define braided monoidal categories, it would be more natural to define them with symmetor an equivalence, and then define a symmetric monoidal category to be a braided monoidal category with the extra condition above.

One slight concern with the second way is that this means there are two proofs that the symmetor is an equivalence, the one from the braided structure, and the one that follows from the extra condition.  In a WildCat, we don't assume that being an equivalence is a proposition, so this feels a bit weird.  I'm not sure if it would cause any problems.

BTW, I could only find one place where in passing this is called the "symmetor", but I'm ok with that name.

jdchristensen (commented):
I left a couple of comments, and will keep reviewing this.

theories/WildCat/Monoidal.v
mikeshulman: I don't really like "symmetor".  I think a much more common name is "braiding" (even in the symmetric case).

theories/WildCat/Monoidal.v
Alizter: Yes, it will be moved. I need to work out how to make this an instance without causing blow up too.

theories/WildCat/Monoidal.v
Alizter: >The symmetry isomorphism needs to also satisfy that symmetor a b $o symmetor b a $== Id _. Without this condition, you get a braided monoidal category.

I originally asked for a map with the self inverse, however it seemed a bit strange that associators and unitors were natural equivalences and this was not. There are also "lax" variants of monoidal categories where those morphisms are directed, but that probably doesn't concern us. I think I was under the mistaken impression that the braiding in a braided monoidal category was a directed morphism rather than an iso which is why it seemed sensible. I see that I was mistaken. :)

>One slight concern with the second way is that this means there are two proofs that the symmetor is an equivalence, the one from the braided structure, and the one that follows from the extra condition. In a WildCat, we don't assume that being an equivalence is a proposition, so this feels a bit weird. I'm not sure if it would cause any problems.

I think as long as it is under a `cate_fun` it shouldn't cause any problems. I don't think we ever compare equivalences directly, not even in `core A`.

> I don't really like "symmetor". I think a much more common name is "braiding" (even in the symmetric case).

I can't remember where I saw symmetor but it seemed more homogeneous with the other names at the time compared to `swap` or `braiding`. I'll rename it to braiding eventually.

jdchristensen:
@Alizter Can I push a couple of cleanups to this?  If you've made changes locally, you can push first.

Alizter:
@jdchristensen Go ahead!

jdchristensen:
@Alizter, I'm not sure what's causing the build errors.

Alizter:
Completely bizarre. Your changes work fine locally, I have no idea what is going on. I'll try merging the master branch and see if being out of date was the problem.

Alizter:
Yeah, it was due to the difference in the base. I get the same error when building locally now.

Alizter:
In `Monoidal.v` a `Basics.Tactics` was removed in the requires. Adding it back fixes everything.

Alizter:
I've renamed `Symmetor` to `Braiding` and fixed the axioms. Having a morphism rather than equivlaences actually simplified some proofs in a few places.

I also started the hexagon consisting of twists. No idea if it's possible, but I've created some lemmas for moving twists and friends around so that I can attack it with the other hexagon. I'll continue tomorrow.

Alizter:
I was think that the Greek letter notations would be a temporary addition, however it is perhaps useful to include them in a module in this file. The proofs are quite unreadable without some abbreviation.

Alizter:
@jdchristensen I've done some reorganisation. Moved things around. Added many more comments explaining the journey. And also cleaned up some of the proofs. I haven't revised the pentagon yet as I'm still trying to workout what kinds of form we want the 9-gon in.

Alizter:
When this PR is a bit more ready, I will split off the material on bifunctors in a separate PR so that we can better review it.

Alizter:
@jdchristensen I've been adding a lot of lemmas and notations to make the reasoning in the pentagon easier. I've come up with a shorter more readable proof that the pentagon (of associators) "reduces" to a 9-gon of twists. It's stated after the first pentagon proof. I think we should go ahead and make this the main proof and ask for the 9-gon as a piece of data.

jdchristensen:
I don't like that form of the 9-gon, which expresses a homotopy between maps `a ⊗ (b ⊗ (d ⊗ c)) $-> ((a ⊗ b) ⊗ c) ⊗ d`.  In the type of maps, it's hard to remember that you have to reverse the way the objects are associated and swap `c` and `d`, and the two maps themselves aren't obvious.

As we discussed elsewhere, there's a fairly natural 9-gon that you can write down, expressing a homotopy between maps `a ⊗ (b ⊗ (c ⊗ d)) $-> c ⊗ (d ⊗ (a ⊗ b))`.  Here, the parentheses are the same, and you only have to remember that it involves the permutation (1,3)(2,4).  One of the maps expresses that permutation as a product of transpositions of adjacent elements, (2,3)(3,4)(1,2)(2,3), keeping the parenthesization the same.  The other map is obtained by associating the domain and codomain as `(a ⊗ b) ⊗ (c ⊗ d)` and `(c ⊗ d) ⊗ (a ⊗ b)` and using the braiding at the top-level;  when you unfold the associator in terms of the braiding and the twist, two of the braidings cancel, and you are left with five maps.  This is something I'd be able to generate, just remembering that the goal is the permutation (1,3)(2,4).

I'm pretty sure this more memorable 9-gon is the one that came up in one version of your first proof (possibly with permuted letters).

(I'm also not sure what your comment "2 extra sides nested in the left making it an 11-gon" in the proof means.)

Alizter:
(I'm also not sure what your comment "2 extra sides nested in the left making it an 11-gon" in the proof means.)

Goal looks like (with new notation):
```
σ d ((a ⊗ b) ⊗ c)
∘ τ (a ⊗ b) d c
∘ σ (d ⊗ c) (a ⊗ b)
∘ τ a (d ⊗ c) b
∘ a ⊗R σ b (d ⊗ c)
 ~[a ⊗ (b ⊗ (d ⊗ c)) $-> ((a ⊗ b) ⊗ c) ⊗ d]~
 σ d ((a ⊗ b) ⊗ c)
∘ d ⊗R σ c (a ⊗ b)
∘ d ⊗R τ a c b
∘ d ⊗R (a ⊗R σ b c)
∘ τ a d (b ⊗ c)
∘ a ⊗R τ b d c
```
I'm just saying this is the 9-gon, but after you cancel `σ d ((a ⊗ b) ⊗ c)`.

I'll try and understand how we can get the other 9-gon.

Alizter:
I've managed to get:
```
d ⊗R σ (a ⊗ b) c
∘ τ (a ⊗ b) d c
∘ σ (d ⊗ c) (a ⊗ b)
∘ τ a (d ⊗ c) b
∘ a ⊗R σ b (d ⊗ c)
 ~[a ⊗ (b ⊗ (d ⊗ c)) $-> d ⊗ (c ⊗ (a ⊗ b))]~
 d ⊗R τ a c b
∘ d ⊗R (a ⊗R σ b c)
∘ τ a d (b ⊗ c)
∘ a ⊗R τ b d c
```
I believe this is acceptable?

jdchristensen:
Yes, that looks good!  When stated as an axiom, I'd swap the names of `c` and `d`, so they appear in order, but you can of course apply the axiom here with the variables reversed.

theories/WildCat/Monoidal.v
```diff
@@ -206,22 +206,34 @@ Section TwistConstruction.
...
   (** *** Setup *)
 
   (** Before starting the proofs, we need to setup some useful definitions and helpful lemmas for working with diagrams. *)
 
   (** We give notations and abbreviations to the morphisms that will appear in diagrams. This helps us read the goal and understand what is happening, otherwise it is too verbose. *)
-  Local Infix "⊗" := cat_tensor (at level 40).
-  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40).
-  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40).
-  Local Notation σ := swap.
-  Local Notation τ := twist.
```
Alizter: I've removed these as they are already quite short.

theories/WildCat/Monoidal.v
```diff
@@ -206,22 +206,34 @@ Section TwistConstruction.
...
-  Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g").
+  Declare Scope monoidal_scope.
+  Local Infix "⊗" := cat_tensor (at level 40) : monoidal_scope.
+  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40) : monoidal_scope.
+  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40) : monoidal_scope.
+  Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g") : monoidal_scope.
   Local Notation "f ~[ A ]~ g" := (GpdHom (A := A) f g)
     (at level 62, format "'[v' '[v' f ']' '/'  ~[ A ]~ '/'  g ']'")
     : long_path_scope.
```
Alizter, line 235: I think we can promote this to wildcats in general as an alternative way to view homotopies. I think it would be useful to peek at a base type of a morphism with something like this.

theories/WildCat/Monoidal.v
```diff
@@ -206,22 +206,34 @@ Section TwistConstruction.
...
-  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40).
-  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40).
-  Local Notation σ := swap.
-  Local Notation τ := twist.
-  Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g").
+  Declare Scope monoidal_scope.
+  Local Infix "⊗" := cat_tensor (at level 40) : monoidal_scope.
+  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40) : monoidal_scope.
+  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40) : monoidal_scope.
+  Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g") : monoidal_scope.
```
Alizter, line 232: I'm going to suggest modifying the `$o` with this break hint. This means that if a notation needs to break then:
```
f $o g
```
becomes
```
f
$o g
```
which is generally more readable. It also means if we stick this in a vertical formatting box like I've done for `f ~[ A ]~ g` Coq will break all the breaks appearing in f and g.

Alizter: This means I can get rid of \circ too.

theories/WildCat/Monoidal.v
```diff
@@ -206,22 +206,34 @@ Section TwistConstruction.
...
-  Local Infix "⊗" := cat_tensor (at level 40).
-  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40).
-  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40).
-  Local Notation σ := swap.
-  Local Notation τ := twist.
-  Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g").
+  Declare Scope monoidal_scope.
+  Local Infix "⊗" := cat_tensor (at level 40) : monoidal_scope.
+  Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40) : monoidal_scope.
+  Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40) : monoidal_scope.
```
Alizter, line 231: I'm not sure if we want to keep these. I think we probably want modify the scope as it is really only a local one. Call it `twist_monoidal_scope`. It's unclear if the general notation for monoidal categories is useful.

theories/WildCat/Monoidal.v
Alizter: Maybe we could use `f $== g :>` instead however to be more inline with the notation we have for equality.

jdchristensen: I like the `:> A` idea.  The other one was mysterious to me until I looked at the definition.

theories/WildCat/Monoidal.v
jdchristensen: Should our notation `=` for path types have a break hint before the equals sign?  I often find it hard to find the equals sign somewhere buried in the goal.  And maybe similarly for `$==`, `==`, etc?

theories/WildCat/Monoidal.v
jdchristensen: I think there's a good chance these three will be useful to others.

theories/WildCat/Monoidal.v
Alizter: Yes, I think there is some general cleanup we can do there.

Alizter:
I've just pushed a commit partially showing that cartesian products and terminal objects give a symmetric monoidal category. I've finished the triangle and pentagon, didn't have time to look at hexagon yet. Notably it shows that we should streamline some of the assumptions as I needed to repeat arguments in a few places.

jdchristensen:
> I've just pushed a commit partially showing that cartesian products and terminal objects give a symmetric monoidal category.

Why didn't you use the twist/braiding approach?!  It should definitely be easier, and the associator for products is already defined in terms of the twist and braiding, so everything is set up to use the material already in this PR...

Alizter:
@jdchristensen this is using the twist approach. Have I missed something?

jdchristensen:
> @jdchristensen this is using the twist approach. Have I missed something?

Sorry, I just read it too quickly!  I saw how long the proof of the pentagon was and missed the first line.  I'm surprised at how tedious that is, since you "just" need to check that the components of those maps agree.  I wonder if there are some lemmas that can be factored out.

Alizter:
@jdchristensen I will spend some more time refactoring the proof today. I think we can make the twist construction easier to use as currently there are quite a lot of hypotheses that need to be repeated.

Alizter:
@jdchristensen I bundled up the `swap` assumptions so that they can be more easily shared. It does however mean the dependencies may be a bit coarser. Should I revert that and try something looser or do you think its ok?

Also, we could do something similar for twist.

theories/WildCat/Monoidal.v
```diff
@@ -157,10 +160,8 @@ Class IsSymmetricMonoidal (A : Type) `{HasEquivs A}
   := {
   (** A monoidal structure with [cat_tensor] and [cat_tensor_unit]. *)
   issymmetricmonoidal_ismonoidal :: IsMonoidal A cat_tensor cat_tensor_unit;
-  (** A natural transformation [braid] witnessing the symmetry of the tensor product. *)
-  cat_symm_tensor_braiding :: Braiding cat_tensor;
-  (** The [braid] is it's own inverse. *)
-  cat_symm_tensor_issymmetricbraid :: IsSymmetricBraid cat_tensor;
+  (** A natural transformation [braid] witnessing the symmetry of the tensor product that is it's own inverse. *)
```
jdchristensen, line 163:
```
(** A natural transformation [braid] witnessing the symmetry of the tensor product such that [braid] is its own inverse. *)
```
(Sometimes github doesn't like me suggest a change.)

theories/WildCat/Monoidal.v
```diff
@@ -184,11 +185,7 @@ Section TwistConstruction.
     `{!Is0Bifunctor cat_tensor, !Is1Bifunctor cat_tensor}
 
     (** Next we postulate the existence of a [swap] map. This takes a tensor pair and swaps the factors. We also postulate that [swap] is natural in both factors and self-inverse. *)
-    (swap : forall a b, cat_tensor a b $-> cat_tensor b a)
-    (swap_swap : forall a b, swap a b $o swap b a $== Id _)
-    (swap_nat : forall a b c d (f : a $-> c) (g : b $-> d),
-      swap c d $o fmap11 cat_tensor f g
-      $== fmap11 cat_tensor g f $o swap a b)
+    (swap : SymmetricBraiding cat_tensor)
```
jdchristensen, line 188: Should we call it `braid` throughout?

theories/WildCat/Monoidal.v
```diff
@@ -237,7 +234,7 @@ Section TwistConstruction.
 
   (** [swap] is its own inverse and therefore an equivalence. *)
   Local Instance catie_swape a b : CatIsEquiv (swap a b)
-    := catie_adjointify (swap a b) (swap b a) (swap_swap a b) (swap_swap b a).
+    := catie_adjointify (swap a b) (swap b a) (braid_braid a b) (braid_braid b a).
```
jdchristensen, line 237: Here is a place where it looks odd to mix the `swap` and `braid` terminology.

jdchristensen (commented):
I think it seems smoother with the braiding bundled up like this.  I haven't thought through how it would look with the `twist` bundled.

theories/WildCat/Monoidal.v
Alizter: I quickly fixed all the errors. I'll do a naming cleanup too.

Alizter:
@jdchristensen I will rebase and squash. Is that alright with you?

Alizter:
Or actually nevermind, I'll just merge the master branch for now and leave the squashing and rebasing for later.

Alizter:
I attempted to make a "twistor" class however stating the naturality condition is not easy since we are missing a bifunctor instance that populates `Is0Functor (fun '(x, y) => F y x)` for a bifunctor `F`. I'm not particularly keen on adding this, so I think the twist stuff is better left as is.

Alizter:
I've marked this as ready-for-review as I have finished everything I wanted to prove. After the review is finished, I think we should squash the commit history before merging.

theories/WildCat/Monoidal.v
Alizter: I won't do this in this PR, but I've created #1927.

theories/WildCat/Monoidal.v
Alizter: I've created https://github.com/HoTT/Coq-HoTT/issues/1928 for the general idea. I will update the notation here.

Alizter: Fixed.

theories/WildCat/Monoidal.v
```diff
@@ -87,5 +125,679 @@ Class IsMonoidal (A : Type) `{HasEquivs A}
...
+
+  Local Definition moveR_twistR a b c d f (g : _ $-> d)
+    : f $== g $o twist b a c -> f $o twist a b c $== g.
+  Proof.
+    intros p; symmetry; apply moveL_twistR; symmetry; exact p.
+  Defined.
+
+  Local Definition moveL_fmap01_twistL a b c d e f (g : e $-> _)
+    : fmap01 cat_tensor a (twist b c d) $o f $== g
+      -> f $== fmap01 cat_tensor a (twist c b d) $o g.
```
jdchristensen, line 496: Isn't this just an example of a general result about moving an equivalence to the other side?

jdchristensen:
This LGTM.  My only hesitation is that there are a lot of specialized lemmas.  It would be great if many of them could be replaced with a smaller number of general purpose lemmas.  I indicated one as an example, but my comment applies more generally.  Still, it's fine with me if you want to merge it as is.  And I think it makes sense to squash it.  (BTW, I'll be offline for a few days on a short vacation.)

theories/WildCat/Monoidal.v
Alizter: Unfortunately not. The issue is that `twist ...` is not definitionally the same as `cate_fun (twiste ...)`. This means in order to apply a lemma about moving equivalences you have to package and unpackage the equivalence. You might then ask, what if we had "catie" versions of the movement lemmas, and I tried this. The issue there is that we have no equation extracting the inverse from `catie_adjointify` so I couldn't relate `twist^-1$` as an inverted CatIsEquiv to `twist`. This is why I opted to just have the specialised lemmas.

Alizter:
@jdchristensen I've squashed everything. I think we should merge and I will go back to working on additive categories.

theories/WildCat/Monoidal.v
jdchristensen: I think we should be able to figure out a way to make this kind of lemma work (maybe with some tweaks to `HasEquivs`), but that can happen as a separate PR.  Feel free to merge.  Thanks for the great work!

PR #1916, opened 2024-04-14
left R-module basics

Alizter:
In this PR we:
- Define left R-modules over a ring R.
- Define submodules and their module structure.
- Define left module homomorphisms.
- Define the category of left modules and homomorphsims.
- Define kernels and images of module homomorphisms.
- Define quotients of modules by submodules.
- Show the first isomorphism theorem for modules.
- Define the direct product of modules and show that the category of modules has products. (This could be generalised to a treatment of biproducts in categories).

### DIFF
--- a/theories/Algebra/Rings/Module.v
+++ b/theories/Algebra/Rings/Module.v
@@ -14,9 +14,9 @@ Declare Scope module_scope.
 
 (** ** Definition *)
 
-(** A left ring action of a ring [R] on an abelian group [M] consists of the following data: *) 
-Class LeftRingAction (R : Ring) (M : AbGroup) := {
-  (** A function [lact] that takes an element [r : R] and an element [m : M] and returns an element [lact r m : M]. *)
+(** An abelian group [M] is a left [R]-module when equipped with the following data: *) 
+Class IsLeftModule (R : Ring) (M : AbGroup) := {
+  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [lact r m : M]. *)
   lact : R -> M -> M;
   (** Actions distribute on the left over addition in the abelian group. That is [lact r (m + n) = lact r m + lact r n]. *)
   lact_left_dist :: LeftHeteroDistribute lact (+) (+);
@@ -30,10 +30,10 @@ Class LeftRingAction (R : Ring) (M : AbGroup) := {
 
 (** TODO: notation for action. *)
 
-(** A left R-module is an abelian group equipped with a left ring action of R. *)
+(** A left R-module is an abelian group equipped with a left R-module structure. *)
 Record LeftModule (R : Ring) := {
   lm_carrier :> AbGroup;
-  lm_lact :: LeftRingAction R lm_carrier;
+  lm_lact :: IsLeftModule R lm_carrier;
 }.
 
 Section ModuleAxioms.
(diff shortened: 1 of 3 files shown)


jdchristensen:
@Alizter Can I push a commit with a few cleanups, rather than pointing them out in the diff?

theories/Algebra/Rings/Module.v
```diff
@@ -0,0 +1,449 @@
...
+Declare Scope module_scope.
+
+(** * Left modules over a ring. *)
+
+(** TODO: Right modules? We can treat right modules as left modules over the opposite ring. *)
+
+(** ** Definition *)
+
+(** A left ring action of a ring [R] on an abelian group [M] consists of the following data: *) 
+Class LeftRingAction (R : Ring) (M : AbGroup) := {
```
jdchristensen, line 18: We would usually call this a "left R-module structure".

Alizter:
@jdchristensen Yes! Go ahead.

theories/Algebra/Rings/Module.v
```diff
@@ -0,0 +1,449 @@
...
+Class IsLeftSubmodule {R : Ring} {M : LeftModule R} (N : M -> Type) := {
+  ils_issubgroup :: IsSubgroup N;
+  is_left_submodule : forall r m, N m -> N (lact r m) ;
+}.
+
+(** A left submodule is a subgroup of the abelian group closed under the action of R. *)
+Record LeftSubmodule {R : Ring} (M : LeftModule R) := {
+  lsm_carrier :> Subgroup M;
+  lsm_submodule :: IsLeftSubmodule lsm_carrier;
+}.
```
jdchristensen, line 100: This is a bit odd because it assumes twice that the subset is a subgroup, once in `Subgroup` and once in `IsLeftSubmodule`.  Probably better to reorganize in some way, maybe with a coercion to `Subgroup`?

theories/Algebra/Rings/Module.v
```diff
@@ -0,0 +1,449 @@
...
+
+(** Any left submodule of a left R-module is an abelian group. *)
+Definition abgroup_leftsubmodule {R : Ring} {M : LeftModule R} (N : LeftSubmodule M) : AbGroup
+  := Build_AbGroup (lsm_carrier M N) _.
+Coercion abgroup_leftsubmodule : LeftSubmodule >-> AbGroup.
```
jdchristensen, line 105: I wonder if we should have a `SubAbGroup` definition, with a coercion to `AbGroup` (and to `Subgroup`).  Then this coercion wouldn't be needed.

theories/Algebra/Rings/Module.v
```diff
@@ -0,0 +1,449 @@
...
+  - intros M N f; apply eissect.
+  - intros M N f; apply eisretr.
+  - intros M N f g fg gf.
+    exact (isequiv_adjointify f g fg gf).
+Defined.
+
+(** ** Kernel of module homomorphism *)
+
+Definition lm_kernel {R : Ring} {M N : LeftModule R} (f : M $-> N)
+  : LeftSubmodule M.
```
jdchristensen, line 306: Maybe we should have instances saying that the grp_kernel and grp_image of an R-module homomorphism satisfying IsLeftSubmodule?

jdchristensen (approved):
I've made a couple of comments here, and I'm about to push some different minor cleanups.

theories/Algebra/Rings/Module.v
Alizter: Yes, I think I wanted to keep the coercion to subgroup so I ended up duplicating and forgetting. I'll demote the carrier to a predicate and introduce a coercion.

theories/Algebra/Rings/Module.v
Alizter: Some of the books on ring theory that I have call these actions, but I noticed this seems to be an older terminology. I'll change this to `IsLeftModule`.

Alizter: Would you also like me to rename `lact`? I can do `lm_act` or `lm_scalar`.

Alizter: We can also introduce a notation for scalar multiplication. `.*`, `**`, `:*` or I can hack around a bit and generalise `Mult` to be "hetero".

theories/Algebra/Rings/Module.v
jdchristensen: Right, if that works then it's cleaner.  And it would still let us get rid of this Coercion, I think.

theories/Algebra/Rings/Module.v
jdchristensen: I agree that it can be called an action.  But a "left module structure" is a bit more common, and I thought it better to parallel the naming of the predicate and the structure.  I don't have a strong preference about the naming of `lact`.  It could stay the same or change.  And I don't know how important it is at this point to have a notation for it.  If it was easy to reuse `*`, that would look most natural, but might also be ambiguous in some situations.  I don't like the other alternatives but can't think of better ones myself.

Alizter: I've renamed `IsLeftModule` and left `lact` unchanged.

theories/Algebra/Rings/Module.v
Alizter: I've managed to define such a coercion and removed this one.

theories/Algebra/Rings/Module.v
Alizter: I've removed this now.

jdchristensen (approved):

Alizter:
@jdchristensen What would you like to see next about modules?

jdchristensen:
> @jdchristensen What would you like to see next about modules?

Axioms for an abelian category, and showing that R-modules and AbGroup are examples.

Things about PIDs and finitely presented modules, as in Section 2.6 of my paper with Jarl Flaten:  https://arxiv.org/abs/2305.09639

PR #1917, opened 2024-04-14
truncation level of lists

Alizter:
We add <s>some more list functions and</s> a proof that `n.+2` truncated types give `n.+2` truncated lists.

### DIFF
--- a/theories/Spaces/List.v
+++ b/theories/Spaces/List.v
@@ -1,11 +1,12 @@
-Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids.
-Require Import Classes.implementations.list.
+Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids Basics.Trunc.
+Require Import Classes.implementations.list Types.Unit Types.Empty Types.Prod.
+Require Import Modalities.ReflectiveSubuniverse Truncations.Core.
 
 Local Open Scope list_scope.
 
-(** ** Lemmas about lists *)
-
 (** Note that [list] is currently defined in Basics.Datatypes. *)
+
+(** ** Lemmas about lists *)
 
 Section Fold_Left_Recursor.
   Variables (A : Type) (B : Type).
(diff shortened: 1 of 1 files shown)


theories/Spaces/List.v
```diff
@@ -1,8 +1,52 @@
...
+    | O => nil
+    | S n => n :: seq_rev n
+    end.
+
+(** Ascending sequence of natural numbers [< n]. *)
+Definition seq (n : nat) : list nat := reverse (seq_rev n).
+
 (** ** Lemmas about lists *)
 
 (** Note that [list] is currently defined in Basics.Datatypes. *)
```
jdchristensen, line 52: This comment about Basics.Datatypes should remain near the top of the file.

theories/Spaces/List.v
```diff
@@ -69,3 +113,77 @@ Proof.
...
 Defined.
+
+(** ** Path spaces of lists *)
+
+(** This proof was adapted from a proof givaen in agda/cubical by Evan Cavallo. *)
+
+Section PathList.
+  Context {A : Type}.
+  
+  (** This type is equivalent to the path space of lists. We don't actually show that it is equivalent to the type of paths but rather that it is a retract of it. This is sufficient to determine the h-level of the type of lists. *)
```
jdchristensen, line 124: It's the other way around.  The path space is a retract of the type `ListEq`.

theories/Spaces/List.v
```diff
@@ -69,3 +113,77 @@ Proof.
...
+  Proof.
+    induction l1 in l2 |- *.
+    - destruct l2.
+      1,2: exact _.
+    - destruct l2.
+      1: exact _.
+      rapply istrunc_prod.
+  Defined.
+
+  (** Since [ListEq] is a retract of the path space of lists, the latter is n.+1-truncated hence [list A] is n.+2-truncated. *)
```
jdchristensen, line 181: Again, reversed.

theories/Spaces/List.v
```diff
@@ -69,3 +113,77 @@ Proof.
...
+      | cons x xs, cons x' xs' => (x = x') * ListEq xs xs'
+      | _, _ => Empty
+    end.
+
+  Global Instance reflexive_listeq : Reflexive ListEq. 
+  Proof.
+    intros l.
+    induction l as [| a l IHl].
+    - exact tt.
+    - by split.
```
jdchristensen, line 137: Maybe it's worth being explicit, since it's easy?
```suggestion
    - cbn.
      exact (idpath, IHl).
```

jdchristensen (approved):
Also, while it's unrelated to this PR, the `list_pentagon` result would fit better in Classes/implementations/list.v, right after `app_assoc`, as then this file wouldn't need to Require list.v.

theories/Spaces/List.v
Alizter: Ah you're right, I always confuse the direction.

Alizter:
@jdchristensen I might do another cleanup after this where I move all the stuff in `implmentations` to `Spaces` so that we have all the list functions and lemmas in one place. What do you think?

jdchristensen:
> @jdchristensen I might do another cleanup after this where I move all the stuff in `implmentations` to `Spaces` so that we have all the list functions and lemmas in one place. What do you think?

Sounds fine.  I guess that would mean that this file will still depend on some of the Classes library, then.  Hmm, I just noticed that the existing list.v only needs canonical_names.v, not abstract_algebra.v, so that's a change that can also be made.

Alizter:
I'm gonna remove the extra functions I added and leave that for the cleanup. I'll just keep the proof of the truncation level available here.

theories/Spaces/List.v
```diff
@@ -69,3 +70,77 @@ Proof.
     lhs_V nrapply ap_compose.
     nrapply (ap_compose (fun l => l ++ z)).
 Defined.
+
+(** ** Path spaces of lists *)
+
+(** This proof was adapted from a proof givaen in agda/cubical by Evan Cavallo. *)
```
Alizter, line 76:
```suggestion
(** This proof was adapted from a proof given in agda/cubical by Evan Cavallo. *)
```

jdchristensen:
LGTM.

PR #1918, opened 2024-04-15
Reduce Required files

jdchristensen:
The first commit changes eight files to Require canonical_names instead of abstract_algebra.  This has only a minor effect, since both of those depend on ReflectiveSubuniverse.v.  (We could consider trying to make canonical_names not depend on it, though.)  Only abstract_algebra depends on Spaces.Nat.Core, so that's a minor gain.

The second commit just runs my usual script that tries removing imports.  I was surprised that it found a lot of things, since I ran it in January and pushed the changes then, but I guess some cleanups since then have improved the dependency structure.  For example, Types.Forall was eliminated in many places, probably because `istrunc_forall` got moved to Basics/Trunc.v.

There's no measurable change to the build timing with these changes.

### DIFF
--- a/contrib/HoTTBook.v
+++ b/contrib/HoTTBook.v
@@ -58,7 +58,7 @@
 *)
 
 From HoTT Require Import Basics Truncations.
-From HoTT Require Idempotents Spaces.Spheres Spaces.No Spaces.Nat.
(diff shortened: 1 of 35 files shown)


Alizter (approved):
Nice!

PR #1919, opened 2024-04-15
coproduct-product inclusion map

Alizter:
This generalises the wedge inclusion to other categories. I don't see anyway of removing the decidable paths requirement. In general I would assume this is equivalent to AC.


jdchristensen (approved):
Should we redefine `fwedge_incl` using this?  I guess we're missing that `pType` has products?

Alizter:
I've shown that pType has all products and redefined the wedge inclusion. However I couldn't get the typeclasses to be found. AFAICT the instances are declared but Coq just cannot find them. My best guess is that unfolding is blocking Coq from finding things, since `HasAllProducts` is a definitional typeclass.

Alizter:
I worked out what was causing the issue with the instances: `Set Universe Minimization ToSet`. Luckily I can locally disable it around the sensitive instance and everything works fine again.

jdchristensen:
Instead of locally disabling minimization to Set, I added a universe variable.  And I made a couple of other minor changes.  If you agree with my changes, feel free to merge.

PR #1920, opened 2024-04-15
cleanup lists

Alizter:
This PR is a major cleanup of our list implementation and lemmas. Historically, we have patched together various kinds of lemmas from mathclasses, HoTT, etc. in somewhat random places. Making it hard to understand how the library works and where things are.

This PR fixes this issue with lists by doing the following:
- Create a new file called Spaces.List.Core for all the core definitions about lists. This includes the definition of a list which has now been moved out of Basics.Datatypes.
- Create a new file called Spaces.List.Theory containing all the list lemmas we've used so far in the library. It is organised to parallel the order found in definitions.
- Create a new file called Spaces.List.Paths which contains the path type characterization of lists. I've improved it so that we actually characterize the path type and not just show its truncation level. It brings in some more dependencies than we would like so it gets its own file.
- Remove all the `ListNotations` modules. The main reason for its existance is the clashing notation of `[]` which for some reason disrupts ltac `intros []`. Rather than forcing users of lists to import the ntoations every time, we simply get rid of the `[]` notation by commenting it out. It saves a total of 1 character compared to `nil` so not a big loss.
- Improve the naming of variables and structure of proofs for various list lemmas. Also taking the time to make them all transparent.


Alizter:
Another thing we could consider doing is moving `Spaces/List/*` into `Types/List/`.

jdchristensen:
(I edited the first comment to add ".List" after "Spaces" in two spots.)

theories/Spaces/List/Theory.v
```diff
@@ -0,0 +1,143 @@
+Require Import Basics.Overture Spaces.List.Core.
+Require Import canonical_names.
```
jdchristensen, line 2: It's pretty annoying to depend on Truncations.Core just so we have `Associative` and `SgOp` in the name space.  The file canonical_names.v has basic definitions mixed with some results about truncations.  I suggest we split it in two, with one part probably only needing a few files from Basics.  In fact, that part could be called `Basics/Names.v`, for example.  For now, we could omit the `SgOp` definition below and could hard-code the definition of associativity.

theories/Classes/implementations/ne_list.v
```diff
@@ -125,10 +125,10 @@ Lemma tails_are_shorter {T} (y x: ne_list T):
   leq (length (to_list x)) (length (to_list y)).
 Proof with auto.
  induction y; cbn.
- - intros [[] | C].
+ - intros [[ ] | C].
```
jdchristensen, line 128:
```suggestion
 - intros [[] | C].
```

theories/Classes/implementations/ne_list.v
```diff
@@ -125,10 +125,10 @@ Lemma tails_are_shorter {T} (y x: ne_list T):
...
    + constructor.
    + elim C.
- - intros [[] | C]...
+ - intros [[ ] | C]...
```
jdchristensen, line 131:
```suggestion
 - intros [[] | C]...
```
Also, I'm not familiar with what "..." means, and it's impossible to google it, so I'd prefer that this be rephrased.

jdchristensen (approved):

jdchristensen:
About the overall structure:  first, I don't think we can move it all to Types/List, as Types is built before Modalities.

I think it's probably fine to keep it as Spaces/List/*.  But maybe we only need two files, not three?  It we get rid of how the second file Theory.v indirectly depends on Truncations.Core, then most of its contents could be put into List.Core.  The only question is whether to put the "for_all_*" things into Core or into Paths.v (which could be renamed).  The "for_all_*" material just uses a few things from Types, so it's not too bad.  Or we could leave it in its own file List.Forall,, which would probably be rarely used.  Or we could just delete it, as it might never be used anywhere.

theories/Classes/implementations/ne_list.v
Alizter: I really don't like that feature. It's documented here: https://coq.inria.fr/doc/V8.19.0/refman/proofs/automatic-tactics/auto.html#coq:cmd.Proof-with

It basically let's you specify some tactics to be tried when you write `...`.

theories/Spaces/List/Theory.v
Alizter: We already define a few classes like this in Basics. I'm not overly fond of the name `Basics/Names.v` but maybe something like `Basics/Classes.v`. In any case, I'll see if I can drop this dependency entirely. I didn't check what breaks when we remove it.

theories/Classes/implementations/ne_list.v
Alizter: If you prefer these to be written like this, then maybe we can reintroduce the `nil` notation?

jdchristensen:
You may have deleted a comment about this, but I don't think a notation for `nil` is needed.  The more ways we have to write the same thing, the more things somebody has to learn to understand our code, and in this case the benefit is too small I think.

theories/Spaces/List/Theory.v
Alizter: I've managed to remove the dependency completely. I agree that we need to make `canonical_names` more light weight and move it out of classes. There is also the related manner of moving all the `abstract_algebra` declarations to their individual files in Algebra, but we are still missing a lot of theory about various structures so that's not possible yet.

jdchristensen:
I'm ok with keeping List.Theory separate.  But maybe it should `Require Export Spaces.List.Core`, so users of this file only need to `Require Import Spaces.List.Theory`?

jdchristensen:
Other than that, this LGTM.

PR #1929, opened 2024-04-25
biproducts, semiadditive and additive categories

Alizter:
Here is a definition of biproducts, semiadditive categories and additive categories. We show that semiadditive categories have a commutative monoid structure on their hom and additive categories have an abelian group structure.

- [ ] clean up associativity proof

### DIFF
--- a/theories/Algebra/Homological/Biproducts.v
+++ b/theories/Algebra/Homological/Biproducts.v
@@ -1,0 +1,1017 @@
+Require Import Basics.Overture Basics.Decidable Basics.Tactics Basics.Trunc.
+Require Import WildCat.Core WildCat.Products WildCat.Coproducts WildCat.Equiv.
+Require Import WildCat.PointedCat WildCat.Bifunctor WildCat.NatTrans WildCat.Prod.
+Require Import WildCat.Monoidal.
+Require Import Types.Bool Types.Forall.
+Require Import Algebra.AbGroups.AbelianGroup.
+Require Import WildCat.Square.
+
+(** * Categories with biproducts *)
+
+(** ** Biproducts *)
+
+Class Biproduct (I : Type) `{DecidablePaths I} {A : Type}
+  `{HasEquivs A, !IsPointedCat A} {x : I -> A}
+  := Build_Biproduct' {
+  biproduct_product :: Product I x;
+  biproduct_coproduct :: Coproduct I x;
+  catie_cat_coprod_prod_incl :: CatIsEquiv (cat_coprod_prod_incl x);
+}.
+
+Arguments Biproduct I {_ A _ _ _ _ _ _} x.
+
+Section Biproducts.
+  Context {I : Type} {A : Type} (x : I -> A) `{Biproduct I A x}.
+  
+  Definition cate_coprod_prod : cat_coprod I x $<~> cat_prod I x
+    := Build_CatEquiv (cat_coprod_prod_incl x).
+
+  Definition cat_biprod : A
+    := cat_prod I x.
+
+  Definition cat_biprod_pr : forall (i : I), cat_biprod $-> x i
+    := cat_pr.
+
+  Definition cat_biprod_corec {z : A}
+    : (forall i, z $-> x i) -> z $-> cat_biprod
+    := cat_prod_corec I.
+
+  Definition cat_biprod_corec_beta {z : A} (f : forall i, z $-> x i)
+    : forall i, cat_biprod_pr i $o cat_biprod_corec f $== f i
+    := cat_prod_beta I f.
+
+  Definition cat_biprod_corec_eta {z : A} (f : z $-> cat_biprod)
+    : cat_biprod_corec (fun i => cat_biprod_pr i $o f) $== f
+    := cat_prod_eta I f.
+  
+  Definition cat_biprod_corec_eta' {z : A} {f f' : forall i, z $-> x i}
+    : (forall i, f i $== f' i) -> cat_biprod_corec f $== cat_biprod_corec f'
+    := cat_prod_corec_eta I.
+  
+  Definition cat_biprod_pr_eta {z : A} {f f' : z $-> cat_biprod}
+    : (forall i, cat_biprod_pr i $o f $== cat_biprod_pr i $o f') -> f $== f'
+    := cat_prod_pr_eta I.
+
+  Definition cat_biprod_in : forall (i : I), x i $-> cat_biprod
+    := fun i => cate_coprod_prod $o cat_in i.
+
+  Definition cat_biprod_rec {z : A}
+    : (forall i, x i $-> z) -> cat_biprod $-> z
+    := fun f => cat_coprod_rec I f $o cate_coprod_prod^-1$. 
+
+  Definition cat_biprod_rec_beta {z : A} (f : forall i, x i $-> z)
+    : forall i, cat_biprod_rec f $o cat_biprod_in i $== f i.
+  Proof.
+    intros i.
+    unfold cat_biprod_rec, cat_biprod_in.
+    refine (_ $@ cat_coprod_beta I f i).
+    nrefine (cat_assoc _ _ _ $@ (_ $@L _)).
+    apply compose_V_hh.
+  Defined.
+
+  Definition cat_biprod_rec_eta {z : A} (f : cat_biprod $-> z)
+    : cat_biprod_rec (fun i => f $o cat_biprod_in i) $== f.
+  Proof.
+    unfold cat_biprod_rec.
+    apply cate_moveR_eV.
+    apply cat_coprod_in_eta.
+    intros i.
+    refine (cat_coprod_beta I _ i $@ _^$).
+    apply cat_assoc.
+  Defined.
+
+  Definition cat_biprod_rec_eta' {z : A} {f f' : forall i, x i $-> z}
+    : (forall i, f i $== f' i) -> cat_biprod_rec f $== cat_biprod_rec f'.
+  Proof.
+    intros p.
+    unfold cat_biprod_rec.
+    exact (cat_coprod_rec_eta I p $@R _).
+  Defined.
+
+  Definition cat_biprod_in_eta {z : A} {f f' : cat_biprod $-> z}
+    : (forall i, f $o cat_biprod_in i $== f' $o cat_biprod_in i) -> f $== f'.
+  Proof.
+    intros p.
+    apply (cate_epic_equiv cate_coprod_prod).
+    apply cat_coprod_in_eta.
+    intros i.
+    exact (cat_assoc _ _ _ $@ p i $@ (cat_assoc _ _ _)^$).
+  Defined.
+
+End Biproducts.
+
+Arguments cat_biprod I {A} x {_ _ _ _ _ _ _ _}.
+Arguments cat_biprod_pr {I A x _ _ _ _ _ _ _ _} i.
+Arguments cat_biprod_in {I A x _ _ _ _ _ _ _ _} i.
+Arguments cate_coprod_prod {I A} x {_ _ _ _ _ _ _ _}.
+
+(** An inclusion followed by a projection of the same index is the identity. *)
+Definition cat_biprod_pr_in (I : Type) {A : Type} (x : I -> A)
+  `{Biproduct I A x} (i : I)
+  : cat_biprod_pr i $o cat_biprod_in i $== Id _.
+Proof.
+  unfold cat_biprod_pr, cat_biprod_in.
+  refine ((_ $@L _) $@ _).
+  { refine ((cate_buildequiv_fun _ $@R _) $@ _).
+    apply cat_coprod_beta. }
+  refine (cat_prod_beta _ _ _ $@ _).
+  destruct (dec_paths i i); simpl.
+  - (** We cannot eliminate [p : i = i] with path induction, but we do have decidable equality, hence by Hedberg's theorem we must be in a hset and we can replace this with the identity. *)
+    assert (r : (idpath = p)) by apply path_ishprop.
+    by destruct r.
+  - contradiction n.
+    reflexivity.
+Defined.
+
+(** An inclusion followed by a projection of a different index is zero. *)
+Definition cat_biprod_pr_in_ne (I : Type) {A : Type} (x : I -> A)
+  `{Biproduct I A x} {i j : I} (p : i <> j)
+  : cat_biprod_pr j $o cat_biprod_in i $== zero_morphism.
+Proof.
+  unfold cat_biprod_pr, cat_biprod_in.
+  refine ((_ $@L _) $@ _).
+  { refine ((cate_buildequiv_fun _ $@R _) $@ _).
+    apply cat_coprod_beta. }
+  refine (cat_prod_beta _ _ _ $@ _).
+  destruct (dec_paths i j); simpl.
+  - contradiction p.
+  - reflexivity.
+Defined.
+
+Definition cat_biprod_diag I {A} (x : A) `{Biproduct I A (fun _ => x)}
+  : x $-> cat_biprod I (fun _ => x)
+  := cat_prod_diag x.
+
+Definition cat_biprod_codiag I {A} (x : A) `{Biproduct I A (fun _ => x)}
+  : cat_biprod I (fun _ => x) $-> x
+  := cat_coprod_fold x $o (cate_coprod_prod (fun _ => x))^-1$.
+
+(** Compatability of [cat_biprod_rec] and [cat_biprod_corec]. *)
+Definition cat_biprod_corec_rec I `{DecidablePaths I} {A : Type}
+  `{HasEquivs A, !IsPointedCat A} {x y : I -> A}
+  `{!Biproduct I x, !Biproduct I y}
+  (f : forall i, x i $-> y i)
+  : cat_biprod_corec y (fun i => f i $o cat_biprod_pr i)
+    $== cat_biprod_rec x (fun i => cat_biprod_in i $o f i).
+Proof.
+  apply cat_biprod_pr_eta.
+  intros i.
+  refine (cat_biprod_corec_beta _ _ i $@ _).
+  apply cat_biprod_in_eta.
+  intros j.
+  refine (_ $@ (_ $@L (cat_biprod_rec_beta _ _ _)^$) $@ (cat_assoc _ _ _)^$).
+  refine (cat_assoc _ _ _ $@ _ $@ cat_assoc _ _ _).
+  destruct (dec_paths j i) as [p | np].
+  - destruct p.
+    refine ((_ $@L _) $@ cat_idr _ $@ (cat_idl _)^$ $@ (_^$ $@R _)).
+    1,2: apply cat_biprod_pr_in.
+  - refine ((_ $@L _) $@ cat_zero_r _ $@ (cat_zero_l _)^$ $@ (_^$ $@R _)).
+    1,2: apply cat_biprod_pr_in_ne, np.
+Defined.
+
+(** *** Existence of biproducts *)
+ 
+Class HasBiproducts (I A : Type)
+  `{DecidablePaths I, HasEquivs A, !IsPointedCat A}
+  := has_biproducts :: forall (x : I -> A), Biproduct I x.
+
+(** *** Binary biproducts *)
+
+Class BinaryBiproduct {A : Type} `{HasEquivs A, !IsPointedCat A} (x y : A)
+  := binary_biproduct :: Biproduct Bool (fun b => if b then x else y).
+  
+Class HasBinaryBiproducts (A : Type) `{HasEquivs A, !IsPointedCat A}
+  := has_binary_biproducts :: forall (x y : A), BinaryBiproduct x y.
+
+Global Instance hasbinarybiproducts_hasbiproductsbool {A : Type}
+  `{HasEquivs A, !IsPointedCat A, !HasBiproducts Bool A}
+  : HasBinaryBiproducts A
+  := fun x y => has_biproducts (fun b : Bool => if b then x else y).
+
+Global Instance hasbinaryproducts_hasbinarybiproducts {A : Type}
+  `{H : HasBinaryBiproducts A}
+  : HasBinaryProducts A.
+Proof.
+  intros x y.
+  apply biproduct_product.
+Defined.
+
+Global Instance hasbinarycoproducts_hasbinarybiproducts {A : Type}
+  `{H : HasBinaryBiproducts A}
+  : HasBinaryCoproducts A.
+Proof.
+  intros x y.
+  nrapply biproduct_coproduct.
+  apply H.
+Defined.
+
+Section BinaryBiproducts.
+
+  Context {A : Type} `{HasEquivs A, !IsPointedCat A} {x y : A}
+    `{!BinaryBiproduct x y}.
+
+  Definition cat_binbiprod : A
+    := cat_biprod Bool (fun b => if b then x else y).
+  
+  Definition cat_binbiprod_pr1 : cat_binbiprod $-> x := cat_biprod_pr true.
+  Definition cat_binbiprod_pr2 : cat_binbiprod $-> y := cat_biprod_pr false.
+  
+  Definition cat_binbiprod_corec {z : A} (f : z $-> x) (g : z $-> y)
+    : z $-> cat_binbiprod.
+  Proof.
+    apply cat_biprod_corec.
+    intros [|].
+    - exact f.
+    - exact g.
+  Defined.
+
+  Definition cat_binbiprod_corec_beta_pr1 {z : A} (f : z $-> x) (g : z $-> y)
+    : cat_binbiprod_pr1 $o cat_binbiprod_corec f g $== f
+    := cat_biprod_corec_beta _ _ true.
+
+  Definition cat_binbiprod_corec_beta_pr2 {z : A} (f : z $-> x) (g : z $-> y)
+    : cat_binbiprod_pr2 $o cat_binbiprod_corec f g $== g
+    := cat_biprod_corec_beta _ _ false.
+
+  Definition cat_binbiprod_corec_eta {z : A} (f : z $-> cat_binbiprod)
+    : cat_binbiprod_corec (cat_binbiprod_pr1 $o f) (cat_binbiprod_pr2 $o f)
+      $== f.
+  Proof.
+    apply cat_biprod_pr_eta.
+    intros [|].
+    - exact (cat_binbiprod_corec_beta_pr1 _ _).
+    - exact (cat_binbiprod_corec_beta_pr2 _ _).
+  Defined.
+
+  Definition cat_binbiprod_corec_eta' {z : A} {f f' : z $-> x} {g g' : z $-> y}
+    : f $== f' -> g $== g'
+      -> cat_binbiprod_corec f g $== cat_binbiprod_corec f' g'.
+  Proof.
+    intros p q.
+    apply cat_biprod_corec_eta'.
+    intros [|].
+    - exact p.
+    - exact q.
+  Defined.
+
+  Definition cat_binbiprod_pr_eta {z : A} {f f' : z $-> cat_binbiprod}
+     : cat_binbiprod_pr1 $o f $== cat_binbiprod_pr1 $o f'
+      -> cat_binbiprod_pr2 $o f $== cat_binbiprod_pr2 $o f'
+      -> f $== f'.
+  Proof.
+    intros p q.
+    apply cat_biprod_pr_eta.
+    intros [|].
+    - exact p.
+    - exact q.
+  Defined.
+
+  Definition cat_binbiprod_inl : x $-> cat_binbiprod := cat_biprod_in true.
+  Definition cat_binbiprod_inr : y $-> cat_binbiprod := cat_biprod_in false.
+
+  Definition cat_binbiprod_rec {z : A} (f : x $-> z) (g : y $-> z)
+    : cat_binbiprod $-> z.
+  Proof.
+    apply cat_biprod_rec.
+    intros [|].
+    - exact f.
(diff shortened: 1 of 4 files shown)


Alizter:
Unification is failing in 8.18. Not quite sure why that is yet.

theories/Algebra/Homological/Abelian.v
```diff
@@ -0,0 +1,144 @@
...
+  := monic_ker_zero (A:=A^op) (a:=b) (b:=a) f.
+
+(** ** Preabelian categories *)
+
+Class IsPreAbelian {A : Type} `{IsAdditive A} := {
+  preabelian_has_kernels :: forall {a b : A} (f : a $-> b), Kernel f;
+  preabelian_has_cokernels :: forall {a b : A} (f : a $-> b), Cokernel f;
+}.
+
+Definition ispreabelian_canonical_map {A : Type} `{IsPreAbelian A}
```
Alizter, line 122: TODO: change name

jdchristensen (commented):
I've only read partway, and will read more later.

jdchristensen:
I pushed a slight cleanup to using decidability.  I'm guessing the lemmas/tactics I added can be used elsewhere.  But if you think they don't help, feel free to revert.

theories/Algebra/Homological/Biproducts.v
```diff
@@ -0,0 +1,809 @@
...
+Require Import Types.Forall Types.Bool Types.Paths Types.Empty Types.Equiv Types.Sigma.
+Require Import WildCat.Core WildCat.Products WildCat.Coproducts WildCat.Equiv.
+Require Import WildCat.PointedCat WildCat.Bifunctor WildCat.Square.
+Require Import WildCat.Opposite WildCat.Monoidal.
+
+(** * Categories with biproducts *)
+
+(** ** Biproducts *)
+
+Class Biproduct (I : Type) `{DecidablePaths I} {A : Type}
```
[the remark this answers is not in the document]
Alizter: In fact categories with infinite biproducts exhibit some degenerate behavior IIRC.

Alizter:
@jdchristensen They look helpful, thanks!

theories/Algebra/Homological/Biproducts.v
```diff
@@ -0,0 +1,809 @@
...
+  catie_cat_coprod_prod_incl :: CatIsEquiv (cat_coprod_prod_incl x);
+}.
+
+Arguments Biproduct I {_ A _ _ _ _ _ _} x.
+
+Section Biproducts.
+  Context {I : Type} {A : Type} (x : I -> A) `{Biproduct I A x}.
+  
+  Definition cate_coprod_prod : cat_coprod I x $<~> cat_prod I x
+    := Build_CatEquiv (cat_coprod_prod_incl x).
```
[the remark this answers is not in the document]
Alizter: I've renamed it.

theories/WildCat/Coproducts.v
```diff
@@ -303,3 +303,22 @@ Proof.
     exact (Id _).
   - apply zero_morphism. 
 Defined.
+
+(** *** Coproducts in the opposite category *)
+
+Definition coproduct_op {I A : Type} (x : I -> A)
+  `{Is1Cat A} {H' : Product I x}
+  : Coproduct I (A:=A^op) x.
```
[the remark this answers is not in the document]
Alizter: That would be very useful. Hopefully we can make progress on that.

theories/Algebra/Homological/Abelian.v
Alizter: I've renamed this.

theories/WildCat/Coproducts.v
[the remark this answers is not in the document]
Alizter: That simplified very nicely. Thank you very much!

theories/WildCat/Coproducts.v
```diff
@@ -308,17 +308,7 @@ Defined.
...
-  - exact cat_pr.
-  - exact (fun z => cat_prod_corec I).
-  - intros z f i.
-    apply cat_prod_beta.
-  - intros z f g p.
-    apply cat_prod_pr_eta.
-    exact p.
-Defined.
+  : Coproduct I (A:=A^op) x
+  := H'.
```
jdchristensen, line 312: It's interesting that this is very fast, but the test I tried to add showing that the `Is1Cat` structure on `A^op^op` is definitionally equal to the original `Is1Cat` structure is slow.  Any idea why, or how to make that test faster?

jdchristensen:
Did you replace `apply` with `nrapply` to speed things up?  I've noticed this occasionally before, and I don't know why it happens.

Alizter:
@jdchristensen I think `apply` is wasting time trying all sorts of first-order heuristics which are not fast when there are typeclass searches to be done.

theories/WildCat/Coproducts.v
Alizter: I think its just a bad combination of typeclass search and unification. I switched `reflexivity` to `nrefine idpath` and that cut it down to 3s from 6. I think whats left is just reduction.

The issue is that this completely unbundled approach has very poor sharing of terms. If you `Set Printing All` you will see how huge the terms become. Coq is spending most of it's time unfolding and trying to unify stuff I believe.

Alizter:
@jdchristensen FTR, I will push some changes here shortly. I am working on smart constructors for biproducts and showing that AbGroup has them.

Alizter:
I've managed to prove that `AbGroup` is additive. It's good to see that the underlying function of both operations on homs are the same as expected.

theories/WildCat/Coproducts.v
jdchristensen: What I don't understand is that in the proof of `coproduct_op`, Coq should need to know that these `Is1Cat` structures agree, and it does so almost instantly.  Why is it fast here but takes 6s in the test?  (If you add in the `Defined` that adds an additional 6s, so it really takes 12s in the test.)

(Also, `nrefine idpath` and `exact idpath` are both slightly slower for me than `reflexitivity`, and all three take around 6s.)

theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -311,3 +315,30 @@ Proof.
...
+  - exact (fun _ => ab_biprod_rec_inl_beta).
+  - exact (fun _ => ab_biprod_rec_inr_beta).
+  - exact (fun _ => ab_biprod_corec_eta').
+  - cbn; reflexivity.
+  - cbn; reflexivity.
+  - cbn; reflexivity.
+  - cbn; reflexivity.
+Defined.
+
+Global Instance issemiadditive_ab `{Funext} : IsSemiAdditive AbGroup := {}.
```
Alizter, line 344: FTR `funext is required here as `MorExt` is part of the data of a semiadditve category. This is required for homs to be groups.

theories/Algebra/Homological/End.v
```diff
@@ -0,0 +1,29 @@
+Require Import Basics.Overture.
+Require Import WildCat.Core.
+Require Import Algebra.AbGroups.AbelianGroup Algebra.Rings.Ring.
+Require Import canonical_names.
+
+Require Import Algebra.Homological.Additive.
+
+(** * Endomorphism Rings *)
```
Alizter, line 8: I haven't checked this yet, but IINM every `AbHom x y` in an additive category can be given a `(End y, End x)`-bimodule structure. Not sure if this is a useful result. (Also we don't have bimodules just yet).

theories/Algebra/Homological/Additive.v
```diff
@@ -461,14 +461,11 @@ Defined.
...
 Proof.
-  snrapply Build_IsSemiAdditive.
-  - exact _.
-  - exact _.
-  - exact _.
-  - intros a b.
-    change A in a, b.
-    change (IsHSet (b $-> a)).
-    exact _.
+  snrapply Build_IsSemiAdditive; try exact _.
```
jdchristensen, line 441: Can we use `1-3: exact _`?  I keep finding places that use `try` that end up being slow after the library evolves, because Coq tries some fruitless typeclass search.

theories/WildCat/Coproducts.v
Alizter: Let's create an issue about this since I don't want to debug it at the moment. What I would do is create an opam switch with function pointer `fp` enabled in the OCaml compiler. Afterwards I would use something like https://github.com/janestreet/memtrace to inspect the processor trace of the reduction. There we should be able to see what is taking the most time. (fp just names the function symbols in the machine code with nice names so we can find them in Coq's source).

theories/Algebra/Homological/Additive.v
Alizter: Yep, I'll update that.

theories/Algebra/Homological/Biproducts.v
```diff
@@ -103,6 +104,62 @@ Arguments cat_biprod_pr {I A x _ _ _ _ _ _ _ _} i.
 Arguments cat_biprod_in {I A x _ _ _ _ _ _ _ _} i.
 Arguments cate_coprod_prod {I A} x {_ _ _ _ _ _ _ _}.
 
+(** A smart constructor for biproducts. Note that compared to a typical definition of a biproduct, we have not asked for the sum of inclusions composed with projections to be the identity. There are a few reasons for this: In our general setting [I] may be infinite, so it doesn't make sense to ask for an infinite sum; Even if we somehow axiomtised something about the sum, we later show in [Additive.v] that the sum is infact an emergent property; Finally, it doesn't seem necessary to produce an isomorphism since we take that to be the identity in this lemma. We should also note that we require [A] to be a pointed category. In theory this requirement could be dropped. To do this we would weaken the equation showing [cat_pr j $o cat_in i] for different [i] and [j], to postulate that [cat_pr j $o cat_in i] is a split idempotent. This doesn't seem particularly useful however as the finite biproduct case implies the existance of zero objects.  *)
```
jdchristensen, line 107: Should the part of this comment about A being pointed be moved to where Biproduct is defined?

I don't understand some of the earlier comments:  "Even if we somehow axiomtised something about the sum, we later show in [Additive.v] that the sum is infact an emergent property; Finally, it doesn't seem necessary to produce an isomorphism since we take that to be the identity in this lemma."  Maybe drop everything from "There are a few reasons..."?

Typos: "existance", "infact", "axiomtised".

theories/Algebra/Homological/Biproducts.v
```diff
@@ -103,6 +104,62 @@ Arguments cat_biprod_pr {I A x _ _ _ _ _ _ _ _} i.
...
+    * destruct p.
+      exact (cat_pr_in i).
+    * exact (cat_pr_in_ne i j np). }
+  snrapply Build_Biproduct'.
+  - exact product.
+  - exact coproduct.
+  - snrapply catie_adjointify.
+    + exact (Id _).
+    + exact (cat_idr _ $@ r).
+    + exact (cat_idl _ $@ r).
```
jdchristensen, line 160: Here you can use `- exact (catie_homotopic _ r^$).`

theories/Algebra/Homological/Biproducts.v
```diff
@@ -103,6 +104,62 @@ Arguments cat_biprod_pr {I A x _ _ _ _ _ _ _ _} i.
...
+  { nrapply rec_eta.
+    intros i.
+    refine (rec_beta _ _ _ $@ _ $@ (cat_idl _)^$).
+    nrapply corec_eta.
+    intros j.
+    refine (corec_beta _ _ _ $@ _^$).
+    destruct (dec_paths i j) as [p|np].
+    * destruct p.
+      exact (cat_pr_in i).
+    * exact (cat_pr_in_ne i j np). }
```
jdchristensen, line 153: I think this argument about how `cat_coprod_prod` computes when composed with the inclusions and projections has appeared more than once, so it should be factored out as a lemma.  Then, with the change that only uses `r` once (using `catie_homotopic`), maybe there is no longer a need to assert the various pieces?

theories/Algebra/AbGroups/AbHom.v
```diff
@@ -117,3 +119,30 @@ Proof.
   rapply (conn_map_elim (Tr (-1)) f).
   exact (equiv_path_grouphomomorphism^-1 p).
 Defined.
+
+(** ** Additivity of AbGroup *)
+
+(** Here is a sanity check that the abelian group structure that is induced from the semiadditive structure is homotopic to the handcrafted operation on hom. *)
+Definition ab_homo_add_is_semiadditive_add `{Funext}
+  {A B : AbGroup} (f g : A $-> B)
+  : ab_homo_add f g = sgop_hom A B f g.
```
jdchristensen, line 128: In fact, the underlying functions are definitionally equal.  I wonder if things can be tweaked so that the proofs that the maps are homomorphisms are definitionally equal too.  (Not that important, I'm just curious.)

theories/WildCat/Coproducts.v
JasonGross: Try `#[local] Strategy -100 [<list of constants that need to be unfolded for unification>]` or `#[local] Strategy 100 [<list of constants that should not be unfolded>]`.  You could debug by manually doing `unfold Product, Coproduct in *. refine H'.` and seeing how much you have to unfold manually before things become fast

theories/Algebra/Homological/Additive.v
```diff
@@ -0,0 +1,466 @@
...
+        1: nrapply cat_binbiprod_corec_beta_pr1.
+        refine (cat_assoc _ _ _ $@ _).
+        nrapply cat_zero_l.
+      - refine (_ $@ cat_binbiprod_pr2_inr^$).
+        refine ((cat_assoc _ _ _)^$ $@ _).
+        refine ((_ $@R _) $@ _).
+        1: nrapply cat_binbiprod_corec_beta_pr2.
+        refine (cat_assoc _ _ _ $@ _).
+        refine (cat_idl _ $@ _).
+        nrapply cat_binbiprod_corec_beta_pr2. }
```
jdchristensen, line 66: I think here and other places, you should look for general lemmas that will help.  Some examples that would help here:
- If `(f, g) : X -> A ⊕ B` is defined by corecursion, then `(fmap11 h k) $o (f, g) $== (h o f, k o g)` (defined by corecursion).
- The map `(0, Id) : B -> A ⊕ B` is homotopic to the second inclusion.

And duals of these.  And similar statements about fmap01 and fmap10.

theories/Algebra/Homological/Biproducts.v
Alizter: Yeah, I will rewrite this I didn't do a good job.

Alizter: If you have a look at a definition of biproduct from MacLane you see that he asks for `i p + j q = 1`. This comment was really to say that this isn't needed for 3 reasons:
1. It doesn't appear in the proof to get an iso between coproduct and product.
2. It doesn't make sense for infinite biproducts (though we only care about finite ones)
3. The sum is an emergent property from the existence of a biproduct so it is the wrong way round.

The other point I was trying to make is that we can assume no zero object and say that the projection and inclusion form a bunch of split idempotents. This is enough to get a zero object.

Maybe I'll just remove this comment all together.

theories/Algebra/AbGroups/AbHom.v
Alizter: Not sure.

theories/Algebra/Homological/Additive.v
```diff
@@ -0,0 +1,465 @@
...
+Section CMonHom.
+
+  Context {A : Type} `{IsSemiAdditive A} (a b : A).
+  
+  (** We can show that the hom-types of a semiadditive category are commutative monoids. *)
+
+  (** The zero element is given by the zero morphism. This exists from our pointedness assumption. *)
+  Local Instance zero_hom : Zero (a $-> b) := zero_morphism.
+
+  (** TODO: explain a bit more: diagonal duplicates and codiagonal sums. *)
```
Alizter, line 37: Need to come up with a good explanation of this. Also it might be worth redefining as `codiag o corec` to be more inline with the Abelian group definition. I'm not sure however as it might make reasoning about it more tricky.

Alizter:
I've fixed the spelling errors in the new comments I've added. I keep forgetting to turn my spellchecker on.

theories/Algebra/AbGroups/Biproduct.v
jdchristensen: I guess the alternative to requiring Funext is to talk about group objects in 0-groupoids, i.e. groups where the laws hold up to the specified 1-cells.  This would be similar to how using the 0-groupoid version of the Yoneda lemma avoids Funext.  But I doubt that it is worth it here.

theories/Algebra/AbGroups/AbHom.v
```diff
@@ -117,3 +119,30 @@ Proof.
...
+Proof.
+  apply equiv_path_grouphomomorphism.
+  reflexivity.
+Defined.
```
jdchristensen, line 132: Maybe it's better to state this in the stronger form (which then has a simpler proof):
```suggestion
(** The group operation that is induced from the semiadditive structure agrees with the handcrafted operation on hom at the level of the underlying functions. *)
Definition ab_homo_add_is_semiadditive_add' `{Funext}
  {A B : AbGroup} (f g : A $-> B)
  : sgop_hom A B f g = ab_homo_add f g :> (A -> B)
  := 1.
```

theories/Algebra/AbGroups/Biproduct.v
Alizter: Yes, essentially setoid based groups. The issue with those is that presumably applying basic equational reasoning will get heavier than it needs to be.

theories/Algebra/AbGroups/AbHom.v
Alizter: I guess Funext can be dropped too.

theories/Algebra/AbGroups/Biproduct.v
mikeshulman: Intriguing.  It does seem like using setoid-groups would be more in line with the overall philosophy of WildCat of avoiding funext/morext, especially since the definition of a semiadditive category is otherwise phrased in a way that doesn't need funext/morext.  But it also definitely makes sense that we want to apply ordinary group theory to the homs, and the HoTT philosophy is that an ordinary group is a set, not a setoid.  I guess this is a place where we see a limitation of the WildCat philosophy.

theories/Algebra/AbGroups/AbHom.v
jdchristensen: > I guess Funext can be dropped too.

Unfortunately not, since it's needed even to define `sgop_hom`, since that requires the semi-additive structure.

Alizter: The only reason `sgop_hom` requires funext is because it is bundled in the section variable. Technically I could remove the "semiadditive" (really truncated + funext) requirement and ask for pointed + biproducts. Not sure this would gain us anything however.

theories/Algebra/Homological/Additive.v
Alizter: The second lemma is better phrased as `(0, f) $== inr $o f` and it worked out well. It simplified the proof of left_identity quite a bit.

Alizter: I'll see what I can do about the associativity now.

theories/Algebra/Homological/Additive.v
```diff
@@ -0,0 +1,447 @@
...
+  : IsAdditive A^op.
+Proof.
+  snrapply Build_IsAdditive.
+  - exact _.
+  - intros a b.
+    change A in a, b.
+    change (Negate (b $-> a)).
+    exact _.
+  - sorry.
+  - sorry.
```
Alizter, line 444: These need to be finished. They are trickier than I thought they would be.

theories/Algebra/AbGroups/AbHom.v
jdchristensen: I had suggested a revised comment that better describes the new statement.  And I'd prefer if Funext was the first argument.

Alizter: Fixed.

theories/WildCat/Coproducts.v
Alizter: I've created #1936 so we can continue the discussion there.

theories/Algebra/Homological/Biproducts.v
```diff
@@ -0,0 +1,968 @@
...
+  snrapply Build_Biproduct'.
+  (** Products in the opposite category are coproducts in the original category. *)
+  - exact _.
+  (** Coproducts in the opposite category are products in the original category. *)
+  - nrapply coproduct_op. 
+    exact _.
+  - snrapply catie_homotopic.
+    + simpl; exact (cat_coprod_prod (A:=A) x).
+    + simpl; exact _.
+    + (** Showing that these two maps are homotopic is a bit tricky. *)
```
jdchristensen, line 915: I have a feeling that what you are doing here might overlap with `cat_biprod_corec_rec`.  If `cat_biprod_corec_rec` is generalized to be about comparing two maps from a coproduct to a product, then maybe it exactly covers what you need here?  Or maybe a slight variation gives a common generalization of both goals?

jdchristensen:
In Biproducts.v, symmetry and associativity of biproducts is proved using that these hold for products.  Then there is some work in showing that these maps respect the coproduct structure.  I feel like this should be more formal.  In a category with biproducts, every product is _canonically_ a biproduct (as is every coproduct).  This is because the biproduct and the product both satisfy the universal property of the product.  So the maps showing that products are associative and symmetric, defined using the product structure, should "automatically" respect the coproduct structure.  I'm not sure how to make this more precise in a way that would make the proofs simpler, but I suspect that there should be a way.

theories/Algebra/Homological/Additive.v
```diff
@@ -0,0 +1,447 @@
...
+  Local Instance right_identity_ab_hom : RightIdentity sgop_hom zero_hom.
+  Proof.
+    intros f.
+    refine (commutativity _ _ @ _).
+    nrapply left_identity_hom.
+  Defined.
+
+  Local Instance associative_ab_hom : Associative sgop_hom.
+  Proof.
+    (** There is a lot to unfold here. But drawing out the diagram we can split this into 5 separate diagrams that all commute. Much of the work here will be applying associativity and functor laws to get this diagram in the correct shape. *)
```
jdchristensen, line 83: I suspect that these results about the monoid structure on the homs can be proved in a simpler way.  First prove that every object of a semi-additive category is a commutative monoid object.  This should be cleaner, since it involves just one object and doesn't involve morext or the h-set assumption.  Then you prove separately that if Y is a commutative monoid object in a wild category whose hom objects are sets, then X $-> Y is a commutative monoid.

(The definition of additive category could also be adjusted to be the statement that each object is a group object.  I think this is equivalent, but again avoids working with hom types.)

jdchristensen:
I've now make a quick pass over everything.  You've done a heroic amount!  I'm hopeful that some parts can be simplified by thinking of conceptual ways to organize what is going on, but even if not, this will be great to have.

theories/Algebra/Homological/Additive.v
Alizter: That's an interesting idea. I guess I need to flesh out some theory of (commutative) monoid objects in a monoidal category first however. Thinking about this, it also seems closer the the intuition of additive homs too. The biproduct functor sandwiched between the diagonal and codiagonal isn't the easiest thing to think about.

jdchristensen: You only need (commutative) monoid objects with respect to the binary product operation, so it's pretty simple to set up.

Alizter: As a related thought, the monoidal generality would let us do H-space structures on homs too right?

theories/Algebra/AbGroups/Biproduct.v
jdchristensen: Another idea would be the following:  In any wild category A, define the homotopy category ho(A) to have the same objects, but with morphisms being the set-quotient of a $-> b by the 2-cells.  Then if A is semi-additive (without assuming morext or that the hom sets are sets), we get that the hom sets in ho(A) are commutative monoids.  Under the additional assumptions that the hom sets are sets and morext holds, ho(A) and A will have equivalent hom sets.  It's a cute idea, but is probably not worth the trouble.

theories/Algebra/Homological/Additive.v
Alizter: > Not sure what you mean by the "monoidal generality". Do you mean monoids wrt a monoidal product, or just the monoid idea in general? But yes, without h-set assumptions, we'll end up with something like a commutative, associative H-space structure on the homs.

Yes, I meant monoids in a monoidal category. I can't remember if we still call these internal monoids. Presumably they are more general as the internal monoid is just the cartesian version of that.

> BTW, the way biproducts are set-up now, products play a more central role than coproducts, since the biproduct is taken to be the product. Because of this, it might be easier to show that every object is a cocommutative comonoid. That still implies that the homs are commutative monoids.

That's a good point. It also might be useful for definitions of Hopf algebras and Co-H-spaces in the future.

>Alternatively, one could adjust how biproducts are set-up to make it more symmetrical. Right now, the cate_coprod_prod map appears when dealing with the coproduct structure, but I think it should be possible to suppress that, and redefine the Coproduct I x instance to use the same object as the Product I x instance. That might simplify a lot of the path algebra that involves the cate_coprod_prod map. There's no reason for that map to appear in any argument, since the product object also satisfies the universal property of being a coproduct.

Yes, that would be a nice simplification. I'll hold off on doing this for the moment so I can finish the other arguments.

jdchristensen: I think the monoid (or comonoid) approach is going to be best, but if you were to continue with the current approach, then `sgop_hom` could be defined in a simpler (but less symmetric) way:  for `f, g : A $-> B`, we can compose the corec `(f,g) : A $-> B ⊕ B` with the codiagonal.  (Or there is a dual approach.)  This might reduce path algebra.

Another thought is that the ith inclusion map for the coproduct structure on the I-indexed biproduct could be defined by corecursion to be the identity into the ith factor and zero_morphism for the others.  This is homotopic to the current definition, so you could still prove the universal property, but this definition would give maps that compute how you want them to.  Similarly for the binary biproduct.

Alizter:
@jdchristensen In the latest commit, I've revised some of the proofs in Products.v. I've given two new proofs of the pentagon: one with the twist construction and one without the twist construction. The one without ended up being much shorter compared to both other proofs. I'll clean this up later so that we only have a single proof. Something to think about with regards to the twist construction.

Alizter:
@jdchristensen Would you like me to split off the changes to `WildCat/Products.v` so that we can better review in a separate PR?

jdchristensen:
Yes, it might be helpful to split it up.

Alizter:
One of my merges accidentally introduced some WIP changes so the build fails. I will fix the build accordingly soon enough.

Alizter:
I've fixed the build errors and tried to make progress on the cocommutative comonoids giving commutative monoid structures on the homs. Unfortunately, this doesn't follow formally from what I can tell so I had to reprove the argument like I did for commutative monoids giving commutative monoids. Once this is done, I have trouble using it since the is1bifunctor instance for cat_binbiprod and cat_bincoprod are different.

One way to fix this is to redefine biproduct so that the coproduct structure is something that can be proved rather than part of the data. I don't know if this will really work, but I am getting quite stuck with this approach, however I'll continue to experiment when I have more time.

jdchristensen:
> I've [...] tried to make progress on the cocommutative comonoids giving commutative monoid structures on the homs. Unfortunately, this doesn't follow formally

It should be formal.  If `X` is a cocommutative comonoid in a wild category `C`, then you should be able to prove easily (or, if you are lucky, it might be definitionally true) that `X` is a commutative monoid in `C^op`.  From that it follows that `Hom` in `C^op` from any `Y` to `X` is a commutative monoid.  But that Hom type is definitionally equal to `Hom` in `C` from `X` to `Y`, so we have a commutative monoid structure here as well.

I haven't looked at the code you pushed, so I don't know how this compares to what have done, but the point is that the argument factors into two parts, and the second part is definitional, and doesn't require that you check the commutative monoid axioms again.

Alizter:
As you said, it should be possible to prove that the double opposite monoid structure is the original structure, however this does not appear to be the case definitionally hence why I struggled earlier. One possible way to fix this is to fix natural transformations (and equivlaences) to become definitionally involutive as highlighted in #1961. This would allow the associator and unitors to be definitionally involutive. (For the full monoidal structure it seems we would need two pentagons and two traingles) however I don't see any use for that yet)
.

jdchristensen:
Even if it is not definitional that a comonoid in `C` is a monoid in `C^op`, I still suspect that going via this statement might be a better way to prove that hom from a comonoid is a monoid.  First, this statement only involves one object, not two, and shouldn't need any extensionality, so it should have less bureaucracy.  Second, it's a useful statement to know independently of this application . Third, once you have this statement, the monoid structure on the hom will be immediate.

Alizter:
@jdchristensen Do you think it would be a good idea to split off the part on biproducts into a separate PR so that it can be reviewed and merged separately. It mostly mirrors what we have for Products and Coproducts and can be modified later when we wish to make progress on semiadditive categories.

jdchristensen:
@Alizter At this point, I don't remember the status of everything in this PR.  Skimming the above, I see lots of discussions that haven't been marked as resolved, so it seems like there might still be a lot to do for the whole PR.  But smaller pieces that are ready to merge would be easier to review.

Alizter:
@jdchristensen Yes, there are a lot of points and I've lost track of most of them at this point.

My plan for making progress is to isolate the part about biproducts and afterwards we can try to tackle the definition of additivity together with the abgroup homs. I've tried a few times to formalize the argument about cocommutative comonoids but I am always running into difficulties which means I am probably doing something wrong. As this PR is too big for its own good, it would be good to investigate these in its own separate PR.

PR #1937, opened 2024-04-30
single file benchmarks

Alizter:
We add some dune rules to benchmark the compilation of single .v files. It works by writing `(theories/file/to/bench.v)` in a file called `file_to_bench` and running `dune build @bench`. This uses a program called `hyperfine` which must be available for this to work. After the bench is run it outputs a bench report.

Typical output:

```
$ dune build @bench
Starting bench. This may take a while.
Bench finished, report at bench_report:

Benchmark 1: /nix/store/v97xkrdzqhwrz8305w05vfix32y5qifa-coq-8.19.1/bin/coqc -R ./theories HoTT -noinit -indices-matter benched_file.v
  Time (mean ± σ):      6.421 s ±  0.024 s    [User: 6.202 s, System: 0.195 s]
  Range (min … max):    6.364 s …  6.461 s    10 runs
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -110,3 +110,6 @@ _CoqProject
 
 # ignore nix profiles
 nix/profiles/
+
+# Ignore the file to bench
+file_to_bench
--- a/dune
+++ b/dune
@@ -62,3 +62,48 @@
    (glob_files_rec contrib/*.v)))
  (action
   (run etc/emacs/run-etags.sh %{vfile})))
+
+; Bench
+
+; The following rule allows you to bench the running time of a file using
+; "hyperfine". For this to work you must make a file called "file_to_bench"
+; which contains the path to the file you want to bench (with brackets around
+; it). For instance, if we wanted to bench the file
+; "theories/WildCat/Products.v", we would make a file called "file_to_bench" with
+; the following content:
+;
+; (theories/WildCat/Products.v)
+;
+; Afterwards you run "dune build @bench" and it will output the report.
+
+(rule
+ (alias bench)
+ (deps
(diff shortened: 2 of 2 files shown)


jdchristensen (approved):
So there isn't a way to pass the name of the file on the command line?  With `make`, we can tell it to do timing by setting some variables, and then use the usual mechanism to just build certain files:  `make TIMED=1 TIMING=1 path/to/file.vo`

BTW, our `make` set-up has lots of timing options via the scripts in the coq-scripts repo.  E.g. before and after comparisons of the whole library.  And the line-by-line timing like at the link above.  Can those things be done via dune?  Currently I rebuild using make when I want to do timing.

Alizter:
Dune isn't aware of the `.timing` files that get produced even if we pass `-time` to the general Coq flags. For it to be aware, I would have to modify Dune upstream so that we can have an "extra files" field for Coq compilation outputs.

The timing makefile scripts are cool but they only really work for make. The same could be adapted to Dune but I don't see the point since it would be a lot of work. In those cases I would just use make. We could make some dune rules that run the makefile in an isolated environment so that it doesn't litter the build directory however.

Alizter:
Oh, and I couldn't work out a way to pass the name via the command line, at least via the dune cli. If you like I can make a bash script in etc/ that takes in the name and updates the file dune is checking before running it.

Alizter:
Also here is hyperfine: https://github.com/sharkdp/hyperfine

jdchristensen:
It's fine to merge this as is.  And I don't mind using make for the older timing scripts.

PR #1938, opened 2024-04-30
optimize Products.v

Alizter:
This switches `apply` to `nrapply` where applicable and a few miscellaneous inlines.

See #1937 for the script that times.

### Before

```                                                                                   
  Time (mean ± σ):      6.421 s ±  0.024 s    [User: 6.202 s, System: 0.195 s]
  Range (min … max):    6.364 s …  6.461 s    10 runs
```

### After

```                                                                                                                                                                                                                                      
  Time (mean ± σ):      2.588 s ±  0.028 s    [User: 2.402 s, System: 0.176 s]
  Range (min … max):    2.559 s …  2.651 s    10 runs
```

### DIFF
--- a/theories/WildCat/Products.v
+++ b/theories/WildCat/Products.v
@@ -40,15 +40,14 @@ Definition Build_Product (I : Type) {A : Type} `{Is1Cat A} {x : I -> A}
 Proof.
   snrapply (Build_Product' I A _ _ _ _ _ cat_prod cat_pr).
   intros z.
-  apply isequiv_0gpd_issurjinj.
-  snrapply Build_IsSurjInj.
-  - simpl.
-    intros f.
+  nrapply isequiv_0gpd_issurjinj.
+  nrapply Build_IsSurjInj.
+  - intros f.
     exists (cat_prod_corec z f).
     intros i.
-    apply cat_prod_beta_pr.
+    nrapply cat_prod_beta_pr.
   - intros f g p.
-    by apply cat_prod_eta_pr.
+    by nrapply cat_prod_eta_pr.
 Defined.
 
 Section Lemmata.
@@ -104,9 +103,9 @@ Section Lemmata.
       refine (_ $@L _).
       exact p.
     - intros a r i.
-      apply cat_idl.
+      nrapply cat_idl; exact _.
     - intros a b c f g r i.
-      apply cat_assoc.
+      nrapply cat_assoc; exact _.
   Defined.
 
   Definition natequiv_cat_prod_corec_inv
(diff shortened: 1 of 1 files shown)


theories/WildCat/Products.v
```diff
@@ -104,17 +103,17 @@ Section Lemmata.
       refine (_ $@L _).
       exact p.
     - intros a r i.
-      apply cat_idl.
+      nrapply cat_idl; exact _.
     - intros a b c f g r i.
-      apply cat_assoc.
+      nrapply cat_assoc; exact _.
```
jdchristensen, line 108: Are these any different from just using `rapply`?

theories/WildCat/Products.v
```diff
@@ -773,148 +770,148 @@ Section Associativity.
...
           2: refine (_ $@R _).
-          2: apply cat_binprod_beta_pr2.
+          2: nrapply cat_binprod_beta_pr2.
           refine (_ $@ cat_assoc _ _ _).
           refine (_ $@ ((_^$ $@ _^$) $@R _)).
-          3: apply cat_assoc.
+          3: nrapply cat_assoc; exact _.
           2: refine (_ $@L _).
-          2: apply cat_pr2_fmap01_binprod.
+          2: nrapply cat_pr2_fmap01_binprod; exact _.
```
jdchristensen, line 905: Same here (and maybe other spots that I missed).

jdchristensen (approved):
Nice.  I had noticed that this file was the slowest in the library.  Even after this change, it's still a bit slow.  This file https://jdc.math.uwo.ca/tmp/Products.html shows the slow lines, in case you want to improve things further.  But that can also be done in the future.

theories/WildCat/Products.v
Alizter: Seems to be faster as rapply uses typeclasses during unification and `nrapply` doesn't and trivially solves them later.

Alizter: When I say seems to be faster, it's the difference between 0.002 and 0.001.

Alizter:
@jdchristensen I definitely think there are some better optimizations that could be done. I also inspected the timing htmls and saw that the `Defined` lines were taking quite a while. I suspect universes may be playing a role with the performance here.

I just wanted to get these easy optimizations out of the way as I add things to this file in #1929.

theories/WildCat/Products.v
jdchristensen: If it's that small, maybe go with the simpler way?  But no big deal either way.  And it's fine to merge this and deal with the slow Defined and section End lines later.

PR #1939, opened 2024-04-30
Use Opposite.v in Equiv.v to remove duplication

### LINKED ISSUE #1936: Work out why test/WildCat/Opposite.v is slow
We need to work out why Coq is so slow in the following example:
```coq
From HoTT Require Import Basics WildCat.Core WildCat.Opposite WildCat.Equiv.

(** * Opposites are definitionally involutive. *)

Definition test1 A : A = (A^op)^op :> Type := 1.
Definition test2 A `{x : IsGraph A} : x = @isgraph_op A^op (@isgraph_op A x) := 1.
Definition test3 A `{x : Is01Cat A} : x = @is01cat_op A^op _ (@is01cat_op A _ x) := 1.
Definition test4 A `{x : Is2Graph A} : x = @is2graph_op A^op _ (@is2graph_op A _ x) := 1.
Definition test5 A `{x : Is1Cat A} : x = @is1cat_op A^op _ _ _ (@is1cat_op A _ _ _ x) := 1.

(** * [core] only partially commutes with taking the opposite category. *)
Definition core1 A `{HasEquivs A} : (core A)^op = core A^op :> Type := 1.
Definition core2 A `{HasEquivs A} : isgraph_op (A:=core A) = isgraph_core (A:=A^op) := 1.
Definition core3 A `{HasEquivs A} : is01cat_op (A:=core A) = is01cat_core (A:=A^op) := 1.
Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=A^op) := 1.

(** This also passes, but we comment it out as it is slow.  When uncommented, to save time, we end with [Admitted.] instead of [Defined.] *)
(*
Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op).
  Time reflexivity. (* ~6s *)
Admitted.
*)
```
Something fishy is going on in core5, but I haven't been able to workout what.

Some context: https://github.com/HoTT/Coq-HoTT/pull/1929#discussion_r1583085953

jdchristensen:
Fixes #1936 

Currently, Opposite.v requires several other files, such as Equiv.v, and includes results such as `hasequivs_op`.  The first commit reverses such dependencies, so, for example, `hasequivs_op` is now in Equiv.v, and Equiv.v requires Opposite.v.  The same happens for FunctorCat.v and NatTrans.v, so Opposite.v now only depends on Core.v.

This means that opposite categories can be used to remove duplicate code in Equiv.v.  There are many dual results, and the second commit just gives some examples of how we can use opposites to remove proof duplication.

A great side effect of this is that it solves the slowdown in #1936 with test/WildCat/Opposite.v, with the test now taking 0.25s.  However, using opposites to prove `compose_cate_assoc_opp` was important for this, and for some reason it takes 3s for this proof.  (If you split it into a tactic and a Defined line, each takes around 1.5s.)  That's much better than the 12s for the test (including the Defined), but is still terrible.  But it at least gives us what might be a simpler problem:  figure out how to get Coq to accept this quickly.  Can anyone help with this?

(Strangely, if you revert this to the old proof, but leave the other dual proofs I gave, the test takes 35s!)

I also give other examples of places that opposites can be used in Equiv.v, but there are more that could be done if we proceed with this.  And there are probably more places in other files as well.  So I think this has benefits separate from the test speed.

### DIFF
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -15,8 +15,6 @@ Definition core3 A `{HasEquivs A} : is01cat_op (A:=core A) = is01cat_core (A:=A^
 Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=A^op) := 1.
 
 (** This also passes, but we comment it out as it is slow.  When uncommented, to save time, we end with [Admitted.] instead of [Defined.] *)
-(*
-Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op).
-  Time reflexivity. (* ~6s *)
-Admitted.
-*)
+
+Time Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op) := 1.
+(* Was 12s, now 0.25s. *)
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -14,8 +14,6 @@ Definition core2 A `{HasEquivs A} : isgraph_op (A:=core A) = isgraph_core (A:=A^
 Definition core3 A `{HasEquivs A} : is01cat_op (A:=core A) = is01cat_core (A:=A^op) := 1.
 Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=A^op) := 1.
 
-(** This also passes, but we comment it out as it is slow.  When uncommented, to save time, we end with [Admitted.] instead of [Defined.] *)
-
+(** The Opaque line reduces the time from 0.3s to 0.07s. *)
 Opaque compose_catie_isretr.
-Time Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op) := 1.
-(* 0.07s (or 0.3s without the Opaque line). *)
+Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op) := 1.
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -16,5 +16,6 @@ Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=
 
 (** This also passes, but we comment it out as it is slow.  When uncommented, to save time, we end with [Admitted.] instead of [Defined.] *)
 
+Opaque compose_catie_isretr.
 Time Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op) := 1.
-(* Was 12s, now 0.25s. *)
+(* 0.07s (or 0.3s without the Opaque line). *)
--- a/theories/WildCat/Equiv.v
+++ b/theories/WildCat/Equiv.v
@@ -2,6 +2,7 @@
 
 Require Import Basics.Utf8 Basics.Overture Basics.Tactics Basics.Equivalences.
 Require Import WildCat.Core.
+Require Import WildCat.Opposite.
 
 (** We declare a scope for printing [CatEquiv] as [≅] *)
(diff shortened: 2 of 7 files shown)


Alizter (commented):
I can say that the slowness doesn't come from typeclass search. Really bizarre.

Alizter:
I suggest we absorb Opposite into Core at this point. It gets used everywhere and I think it makes sense not to explicitly require it any longer.

jdchristensen:
> I suggest we absorb Opposite into Core at this point.

I was considering that because I expected that we could use duality in Core.v to reduce duplication, but a quick look didn't find anything except one-liners, so there is little gain.  And Core.v is over 600 lines long, so maybe it's better to keep Opposite separate?  There are 19 .v files in WildCat that don't explicitly require Opposite (yet).

jdchristensen:
I found a fix.  See the last commit.  I made the adjointification proofs syntactically dual, and then made the shared proof object opaque for the slow lemma.

Now the `exact` line takes about 0.03s and the `Defined` line takes about 0.01s.  The `core5` test got a touch slower, but using the same Opaque trick sped that up as well, to 0.07s.

The exact line can be made instant by using `exact_no_check`.  Maybe we should use this for slow lines?  The `Defined` line still checks things...  Any objections?

Alizter:
> The exact line can be made instant by using exact_no_check. Maybe we should use this for slow lines? The Defined line still checks things... Any objections?

That seems sensible.

Will you undraft now?

jdchristensen:
I used duality in a few more places.  I think similar tricks would work in DisplayEquiv if the opposite of a displayed category was defined, but it's not that important.

I checked the overall timing of the library, and it's within the noise.  The only files that might be a bit slower are DisplayedEquiv.v and Idempotents.v.  I took a quick look and didn't see anything obviously related to the changes here.  I'll rebase and check the timing using hyperfine.

jdchristensen:
Using hyperfine, I see that Idempotents.v is very slightly _faster_ with this PR, by 0.01s.  DisplayedEquiv is 0.08s slower, which is around 8%.  This is due to a slowdown in `dcompose_cate_idr`, which uses `compose_cate_idr`, which is now defined by duality.  This can be fixed by not defining `compose_cate_idr` by duality, but 0.08s is not much, and I think this approach is worth it.  In addition, if displayed categories get duality, then we'll be able to prove `dcompose_cate_idr` by duality, which should mesh well.  So I think this is ready for final review.

test/WildCat/Opposite.v
```diff
@@ -15,8 +15,7 @@ Definition core3 A `{HasEquivs A} : is01cat_op (A:=core A) = is01cat_core (A:=A^
 Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=A^op) := 1.
 
 (** This also passes, but we comment it out as it is slow.  When uncommented, to save time, we end with [Admitted.] instead of [Defined.] *)
```
Alizter, line 17: This comment needs updating.

theories/WildCat/Equiv.v
```diff
@@ -317,20 +331,16 @@ Defined.
...
   (p : f $== e^-1$ $o g)
-  : e $o f $== g.
-Proof.
-  apply (cate_monic_equiv e^-1$).
-  exact (compose_V_hh _ _ $@ p).
-Defined.
+  : e $o f $== g
+  := cate_moveR_eM (A:=A^op) (a:=c) (b:=b) (c:=a) e f g p.
+
+(* TODO: lots more results are duals. *)
```
Alizter, line 343: Shall we create an issue about this?

jdchristensen: I just forgot to remove that comment.  I didn't see any other obvious things to do, but we can keep our eyes open.

jdchristensen:
@Alizter Thanks for catching those things.  I fixed them all.

Alizter (approved):

Issue #1921, opened 2024-04-17
Opposite ring


Alizter:
We should define the opposite ring and remove the duplication in the theory of ideals.

Alizter:
It occurs to me that if we want `R^op^op` to be definitionally `R` then we need to duplicate the associativity data.

Alizter:
After some quick experimentation I noticed that a ring `R` has quite a few bits of duplicate data that causes equality not to hold:
```
sg_set0 = sg_set
sg_ass0 = sg_ass
monoid_left_id0 = monoid_left_id
monoid_right_id0 = monoid_right_id
negate_l0 = negate_l
negate_r0 = negate_r
abgroup_commutative0 = abgroup_commutative
sg_set1 = sg_set
```
This is obviously not ideal and needs to be cleaned up.

jdchristensen:
Right, the issue is that `IsRing` contains `ring_abgroup :: @IsAbGroup plus_is_sg_op zero_is_mon_unit _;`.

About R^op^op being definitionally R, it's possible that it won't come up as often as opposite categories do, but it's worth thinking about.  One can also reverse groups, semigroups, monoids, etc, and it's probably not worth adding extra associativity clauses for all of these.

mikeshulman:
If one were designing an algebraic hierarchy from the ground up, one might consider abstracting out a notion of "involutive equality", proving general properties about it, and then just phrasing all the algebraic definitions in terms of it.

Alizter:
If I were designing things from the ground up, I would even consider making everything a setoid and having equality as a special case. There is would make sense to have "involutive equality". It would have a lot of unifying potential with non-set based structures.

However I am not particular keen on doing this. At the end of the day, the real motivation behind any algebra that I do is to do some actual calculations in homotopy theory. :)

jdchristensen:
> Hmm it's a lot more work to change up the definitions in mathclasses

Why do you need to change things here?  Can't you just drop the abelian group structure from the Ring Record you defined, and then define a coercion to AbGroup?  Carrying around two abelian group structures seems like a recipe for problems.

Alizter:
> > Hmm it's a lot more work to change up the definitions in mathclasses
> 
> Why do you need to change things there? Can't you just drop the abelian group structure from the Ring Record you defined, and then define a coercion to AbGroup? Carrying around two abelian group structures seems like a recipe for problems.
> 
> (Edit: changed "here" to "there" in first sentence.)

[...]

PR #1940, opened 2024-04-30
opposite rings

Alizter:
This is a start on #1921. I've opened it now so I can get some comments before starting ideals.

### TODO

- [x] deduplicate theory of ideals
- [ ] right modules (perhaps in a later PR?)


test/Algebra/Rings/Ring.v
```diff
@@ -0,0 +1,9 @@
+From HoTT Require Import Basics.Overture Algebra.Rings.Ring Algebra.Rings.CRing.
+From HoTT.Algebra Require Import Groups.Group AbGroups.AbelianGroup.
+From HoTT Require Import abstract_algebra.
+
+Local Open Scope path_scope.
+
+(** Test that opposite rings are definitionally involutive. *)
+Definition test1 (R : Ring) : R = (rng_op (rng_op R)) :> Ring := 1.
+Definition test2 (R : CRing) : R = (rng_op (rng_op R)) :> CRing := 1.
```
Alizter, line 9: Didn't think that this would work but it does. I had to double check but it appears as though it's working somehow?

theories/Classes/theory/groups.v
```diff
@@ -104,7 +104,7 @@ End groupmor_props.
 Section from_another_sg.
 
   Context
-    `{IsSemiGroup A} `{IsHSet B}
+    `{IsSemiGroup A} `{IsHSet A} `{IsHSet B}
```
Alizter, line 107:
```suggestion
    `{IsSemiGroup A} `{IsHSet A}
```

theories/Classes/theory/groups.v
```diff
@@ -104,7 +104,7 @@ End groupmor_props.
 Section from_another_sg.
 
   Context
-    `{IsSemiGroup A} `{IsHSet B}
+    `{IsSemiGroup A} `{IsHSet A}
```
Alizter, line 107:
```suggestion
    `{IsSemiGroup A} `{IsHSet B}
```

test/Algebra/Rings/Ring.v
jdchristensen: Seems reasonable to me.  Redundant associativity was the only thing needed to get double op working for categories.

Alizter: Right but I've only defined `rng_op` for rings yet it works for commutative rings. I am a little surprised.

jdchristensen: Oh, now I see what you mean.  I don't think it is working.  For test1, if you Set Printing All, it shows the goal `@paths Ring R (rng_op (rng_op R))`.  But for test2, it shows `@paths CRing R R`.  So the ops are gone.  That's without any `cbn` or `simpl` commands.  Weird.

test/Algebra/Rings/Ring.v
```diff
@@ -7,3 +7,15 @@ Local Open Scope path_scope.
...
+Definition augment (R : CRing) : CRing.
+Proof.
+  snrapply Build_CRing.
+  - exact R.
+  - exact (fake_comm R).
+Defined.
+
+Fail Definition test3 (R : CRing) : R = augment R :> CRing := 1.
+(* !? *)
+Definition test4 (R : CRing) : R = (rng_op (rng_op (augment R))) :> CRing := 1.
```
Alizter, line 21: Now this is really weird and maybe even inconsistent?

Alizter: What's weirder still is that test4 doesn't have the type I've given it. It just shows `R = R`.

jdchristensen: `augment` gets thrown away when you pass the result to `rng_op`, since `rng_op` ignores the commutativity proof.  So it's not surprising that this is no different from how it was before.  If you put the `augment` on the outside of both `rng_op`s, it will probably fail.  But I'm still surprised that Coq unfolds the `rng_op`s and then uses the original commutativity proof.  In any case, you just need to add a `cring_op` and use that.

Alizter: I'm currently running the bug minimizer so we can get a better look at this throwing-away that Coq is doing.

Alizter:
Mystery solved: Coq makes the `:>` fields in records "reversible" meaning that there is a term inserted during elaboration. We don't see this since we didn't include the following in our prelude:
```
#[universes(polymorphic=yes)] Definition ReverseCoercionSource (T : Type) := T.
#[universes(polymorphic=yes)] Definition ReverseCoercionTarget (T : Type) := T.
#[warning="-uniform-inheritance", reversible=no, universes(polymorphic=yes)]
Coercion reverse_coercion {T' T} (x' : T') (x : ReverseCoercionSource T)
  : ReverseCoercionTarget T' := x'.
Register reverse_coercion as core.coercion.reverse_coercion.
```

In this case, our `CRing` is being coerced (by a reversible coercion) to `Ring. Then the involution happens definitionally giving us the same `Ring`. Coq then says this was coerced by a reversible coercion so let me insert the inverse and brings it back to a `CRing`.

I believe this is unintentional for this coercion to be marked as reversible and has been fixed upstream. https://github.com/coq/coq/pull/18705

We should add the code above so that `Set Printing All` better displays reversed coercions.

jdchristensen:
Ok, that's interesting.  In any case, I still think we want to define `crng_op` and should change test2 to use it.

Alizter:
I simplified the theory of ideals using the opposite ring. However in doing so I redefined the quotient/colon ideal as an intersection of the "left" and "right" versions. This has made the entire thing much more complicated so I don't know what to do with it.

I have scoured all the literature I know on noncommutative rings and I couldn't find anything about colon ideals in this generality. It only seems to appear in commutative algebra/algebraic geometry, in fact this was my original motivation for adding them as they describe complements in the Zariski topology.

I'm tempted to just move all the colon ideal stuff to `CRing` and only have the single two-sided definition.

theories/Algebra/Rings/CRing.v
```diff
@@ -199,3 +222,31 @@ Global Instance isgraph_CRing : IsGraph CRing := isgraph_induced cring_ring.
 Global Instance is01cat_CRing : Is01Cat CRing := is01cat_induced cring_ring.
 Global Instance is2graph_CRing : Is2Graph CRing := is2graph_induced cring_ring.
 Global Instance is1cat_CRing : Is1Cat CRing := is1cat_induced cring_ring.
+Global Instance hasequiv_CRing : HasEquivs CRing := hasequivs_induced cring_ring.
+
+(** ** Opposite commutative ring *)
+
+Definition cring_op (R : CRing) : CRing.
```
jdchristensen, line 229: I know I suggested this, but now I'm wondering whether it's needed, since the opposite of a commutative ring is isomorphic to the original ring.

jdchristensen:
> I simplified the theory of ideals using the opposite ring.

That turned out quite well!

> I'm tempted to just move all the colon ideal stuff to `CRing` and only have the single two-sided definition.

I don't know much about this, but specializing to commutative rings seems fine to me, if that's all you can find written about this.

theories/Algebra/Rings/Ideal.v
```diff
@@ -314,6 +316,7 @@ Proof.
     by apply sgt_op.
 Defined.
 
+(** TODO: not sure how to dualize this. *)
```
jdchristensen, line 319: If you changed `ideal_product_naive_type` to the predicate `fun z => exists x y, I x * J y * (z = x * y)` it might work.  (Two of those asterisks are "and".)  But maybe this would make other things less convenient?

jdchristensen: By creating functor_subgroup_generated, I was able to remove a lot of the duplication.

theories/Algebra/Rings/Ring.v
```diff
@@ -395,7 +407,54 @@ Proof.
...
+  - exact (fun x y => mult y x).
+  - exact one.
+  - exact is_abgroup.
+  - exact monoid_ishset.
+  - exact (fun x y z => mult_assoc_opp z y x).
+  - exact ri.
+  - exact li.
+  - exact (fun x y z => rd y z x).
+  - exact (fun x y z => ld z x y).
+  - exact (fun x y z => mult_assoc z y x).
```
jdchristensen, line 433: If we have `mult_assoc_opp` take its arguments in the reverse order, then lines 428 and 433 will become simpler, and so `rng_op` will be slightly smaller.  Probably worth it.  For Is1Cat too.

jdchristensen (approved):

theories/Algebra/Rings/Ring.v
jdchristensen: Same for the distributivity laws (which also comes up in my last commit, where I needed a `fun x y => ...`).

Alizter:
I think I won't touch it for the moment.

theories/Algebra/Rings/Ring.v
Alizter: I've changed `mult_assoc_opp` like you've suggested but I don't think its worth touching the distributivity laws in general. Your specific case isn't actually using the mathclasses distributivity but the ring law version.

However changing the order of those arguments will be confusing during use. The opp field isn't supposed to actually be used by anybody anyway.

jdchristensen:
LGTM

PR #1941, opened 2024-05-02
register dummy term for inspecting reverse_coercions

Alizter:
As found in #1940.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -770,3 +770,17 @@ Ltac easy :=
   (use_hyps; do_ccl) || fail "Cannot solve this goal".
 
 Tactic Notation "now" tactic(t) := t; easy.
+
+(** *** Reversible Coercions *)
+
+(** Coercions in Coq since 8.16 have the ability to be reversible. These are coercions that are not regular functions but rely on some meta-procedure like typeclass resolution to fill in missing pieces. Examples include marking fields of a record with [:>] which allows Coq to elaborate the projected term to the original term.
+
+This behaviour can have some surprising effects in some places, where you might not expect a term to be elaborated. When inspecting proofs with [Set Printing All] you will not be able to see the reverisble coercion. In order to help with inspecting such situations, Coq exposes a register for a dummy term called [reverse_coercion] which gets inserted during an application of a reversible coercion. This way you can see the application clearly in a proof term.
+
+We register this here. This is standard from the Coq stdlib prelude.*)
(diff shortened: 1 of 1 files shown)


theories/Basics/Overture.v
```diff
@@ -770,3 +770,17 @@ Ltac easy :=
   (use_hyps; do_ccl) || fail "Cannot solve this goal".
 
 Tactic Notation "now" tactic(t) := t; easy.
+
+(** *** Reverisble Coercions *)
```
jdchristensen, line 774: typo

theories/Basics/Overture.v
```diff
@@ -770,3 +770,17 @@ Ltac easy :=
...
+
+(** Coercions in Coq since 8.16 have the ability to be reversible. These are coercions that are not regular functions but rely on some meta-procedure like typeclass resolution to fill in missing pieces. Examples include marking fields of a record with [:>] which allows Coq to elaborate the projected term to the original term.
+
+This behaviour can have some surprising effects in some places, where you might not expect a term to be elaborated. When inspecting proofs with [Set Printing All] you will not be able to see the reverisble coercion. In order to help with inspecting such situations, Coq exposes a register for a dummy term called [reverse_coercion] which gets inserted during an application of a reversible coercion. This way you can see the application clearly in a proof term.
+
+We register this here. This is standard from the Coq stdlib prelude.*)
+#[universes(polymorphic=yes)] Definition ReverseCoercionSource (T : Type) := T.
+#[universes(polymorphic=yes)] Definition ReverseCoercionTarget (T : Type) := T.
+#[warning="-uniform-inheritance", reversible=no, universes(polymorphic=yes)]
```
jdchristensen, line 783: Can we drop the polymorphic annotations, since we have that set globally?

(BTW, I think we should move some of these technical settings out of Overture.v, into an earlier file, e.g. Setup.v, so that Overture.v can be more easily read by someone learning the library.  But that can be a different PR.)

jdchristensen (approved):

theories/Basics/Overture.v
Alizter: Yeah, I was thinking about that too. I would suggest splitting technical aspects and basic HoTT stuff into two files. This would reduce a lot of the cruft we currently have. I've been aware for some time that when people look at the library they want to see something basic, they open this file and they get very intimidated by the business. This is definitely something I want to fix at some point.

Alizter:
I've bitten the bullet and cleaned up basics:
- #1942.

PR #1942, opened 2024-05-02
cleanup Basics/

Alizter:
In this PR we:
* Move global settings from Overture.v to a new file Settings.v
* Absorb Datatypes.v into Overture.v
* Absorb Logic.v into Overture.v
* Give prod_rect and sum_rect the proper prod_ind and sum_ind names
* Add explanations for some unexplained settings.
* Moved numeral definitions used only for parsing `1234` to a subdirectory `Basics/Numerals`
* Removed `True` and `False` types. We have `Unit` and `Empty` for that.
* Subsume #1941. Note that I would rather keep the explicit polymorphism attributes so that if things move around we don't accidentally cause some hard-to-debug issues.


theories/Basics/Overture.v
```diff
@@ -2,47 +2,91 @@
...
 
-(** This command makes it so that you don't have to declare universes explicitly when mentioning them in the type.  (Without this command, if you want to say [Definition foo := Type@{i}.], you must instead say [Definition foo@{i} := Type@{i}.]. *)
-Global Unset Strict Universe Declaration.
+Arguments inl {A B} _ , [A] B _.
+Arguments inr {A B} _ , A [B] _.
 
-(** This command makes it so that when we say something like [IsHSet nat] we get [IsHSet@{i} nat] instead of [IsHSet@{Set} nat]. *)
-Global Unset Universe Minimization ToSet.
+(* A notation for coproduct that's less overloaded than [+] *)
+Notation "x |_| y" := (sum x y) (only parsing) : type_scope.
```
Alizter, line 50: Should we get rid of this?

theories/Basics/Overture.v
```diff
@@ -2,47 +2,91 @@
...
+Notation "x * y" := (prod x y) : type_scope.
+Notation "( x , y , .. , z )" := (pair .. (pair x y) .. z) : core_scope.
+Notation "A /\ B" := (prod A B) (only parsing) : type_scope.
+Notation and := prod (only parsing).
+Notation conj := pair (only parsing).
+
+#[export] Hint Resolve pair inl inr : core.
+
+Definition prod_curry (A B C : Type) (f : A -> B -> C)
+  (p : prod A B) : C := f (fst p) (snd p).
```
Alizter, line 77: This seems useless.

theories/Basics/Overture.v
jdchristensen: I didn't even know about it, so I don't mind getting rid of it.  Having too many different notations for the same thing makes the learning curve harder.

theories/Basics/Overture.v
jdchristensen: I guess if it's never been used, it's not worth having.  And I see that `uncurry` is only used in one other spot.

theories/Basics/Overture.v
```diff
@@ -2,47 +2,91 @@
...
+
+Notation "A <-> B" := (iff A B) : type_scope.
+
+(** ** Type classes *)
+
+(** This command prevents Coq from trying to guess the values of existential variables while doing typeclass resolution.  If you don't know what that means, ignore it. *)
+Local Set Typeclasses Strict Resolution.
 
 (** Apply using the same opacity information as typeclass proof search. *)
 Ltac class_apply c := autoapply c with typeclass_instances.
```
jdchristensen, line 92: Would some of the type classes settings also fit into Settings.v?

theories/Basics/Settings.v
```diff
@@ -0,0 +1,81 @@
+(** * General Settings *)
+
+(** This file contains all the tweaks and settings we make to Coq as to not clutter Overture.v *)
```
jdchristensen, line 3:
```suggestion
(** This file contains all the tweaks and settings we make to Coq. *)
```

theories/Categories/Comma/ProjectionFunctors.v
```diff
@@ -46,8 +46,8 @@ Section comma.
              (ST : object ((A -> C)^op * (B -> C)))
   : Cat / !((A * B; PAB) : Cat).
   Proof.
-    exists (Datatypes.fst ST / Datatypes.snd ST; P_comma _ _) (center _).
-    exact (comma_category_projection (Datatypes.fst ST) (Datatypes.snd ST)).
+    exists (Basics.Overture.fst ST / Basics.Overture.snd ST; P_comma _ _) (center _).
+    exact (comma_category_projection (Basics.Overture.fst ST) (Basics.Overture.snd ST)).
```
jdchristensen, line 50: Better to Require Import Basics.Overture at top?  Same below and in other files.  E.g. several files make custom notations for fst and snd, which seems odd.

theories/Basics/Overture.v
Alizter: They appear to be local so it didn't make sense to move them.

theories/Categories/Comma/ProjectionFunctors.v
Alizter: Here it isn't needed but elsewhere there is a name clash with the projection functor in the Categories library.

theories/Basics/Overture.v
mikeshulman: I guess this is one place where avoiding unicode kind of bites us.

theories/Categories/Comma/ProjectionFunctors.v
Alizter: I think it's best to keep them qualified since the categories library uses `fst` and `snd` alot.

theories/Basics/Overture.v
Alizter: That's true, but I still think its a minor issue compared to the obfuscation that full unicode everywhere can bring. There are ways to unambiguously setup + so that it can be used in tandem using the bind arguments feature, however I don't want to spend time setting that up in mathclasses.

Alizter: With more modern Coq editors, it wouldn't be too difficult to setup latex notations for everything and render them using mathjax in the goals window. But typically when I get stuck in an argument and I need to freehand prove something, I just raid the printer tray.

theories/Basics/Overture.v
jdchristensen: Oh right, the first one is explicitly labelled Local.  But the second one is used in Classes.  Maybe we should move all tactics out of Overture.v and into Basics/Tactics.v?

Alizter: Apparently strict resolution made no difference here. I'm inclined to remove it.

Alizter: Actually it did, but in other files?! I must be misunderstanding how the locality here is working.

Alizter: I found the issue. It affects the instances that are declared: https://coq.inria.fr/doc/V8.18.0/refman/addendum/type-classes.html#coq:flag.Typeclasses-Strict-Resolution

So this setting belongs here.

jdchristensen:
What do you think about moving the `easy` and `now` tactics to Tactics.v as well?

Alizter:
We could also move `reflexivity` etc. but those are used in the file so it would be a little trickier. Also I think they are more instructive for somebody new to ltac than the other tactics.

jdchristensen:
> We could also move `reflexivity` etc. but those are used in the file so it would be a little trickier. Also I think they are more instructive for somebody new to ltac than the other tactics.

Yeah, I thought about those.  The other reason to leave them in Overture.v is that they involve things defined there.  I think for tactics like that, it makes sense to define them right where the relevant material is defined.

jdchristensen (approved):

PR #1943, opened 2024-05-02
shorter proofs of monoidal structure in Products.v

Alizter:
By not going through the twist construction for the pentagon and hexagon in Products.v we can make the proof shorter. This is because we can efficiently compute the projections of an associator.

This was needed in #1929 so that we could more easily work with the associator.

### DIFF
--- a/theories/WildCat/Products.v
+++ b/theories/WildCat/Products.v
@@ -331,6 +331,51 @@ Proof.
     + exact (p false).
 Defined.
 
+Definition cat_binprod_eta_pr_x_xx {A : Type} `{HasBinaryProducts A} {w x y z : A}
+  (f g : w $-> cat_binprod x (cat_binprod y z))
+  : cat_pr1 $o f $== cat_pr1 $o g
+  -> cat_pr1 $o cat_pr2 $o f $== cat_pr1 $o cat_pr2 $o g
+  -> cat_pr2 $o cat_pr2 $o f $== cat_pr2 $o cat_pr2 $o g
+  -> f $== g.
+Proof.
+  intros p q r.
+  snrapply cat_binprod_eta_pr.
+  - exact p.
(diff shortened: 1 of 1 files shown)


jdchristensen:
It's hard to tell whether the smaller proofs of the hexagon and pentagon are mainly due to not using the twist or due to the various helper lemmas you added.  It's definitely smart to add and use those lemmas!  And since that path algebra is so tricky, I don't think it's worth investigating whether the twist approach could be equally short.  So, this LGTM.

(BTW, it just occurred to me that the lhs and rhs tactics could probably be adapted to work for any transitive relation, so they could be used for `$==`, `==`, `<~>`, etc.  That would probably help with these proofs!)

jdchristensen (approved):

Alizter:
> It's hard to tell whether the smaller proofs of the hexagon and pentagon are mainly due to not using the twist or due to the various helper lemmas you added. It's definitely smart to add and use those lemmas! And since that path algebra is so tricky, I don't think it's worth investigating whether the twist approach could be equally short. So, this LGTM.

For the pentagon, its a matter of size. Reducing 5 terms is inherently shorter than reducing 9 terms. I know this because I originally shortened the pentagon proof to use some similar lemmas and then avoided the twist construction.

For the hexagon, I didn't try the twist construction directly. But presumably because the proof fits on my monitor it is good enough.

Again, this is only because products have this nice property that we can prove these lemmas more easily. So this idea doesn't apply to other monoidal products like the join. Although, for the join it's not entirely certain yet if the 9-gon axiom is manageable yet. 

> (BTW, it just occurred to me that the lhs and rhs tactics could probably be adapted to work for any transitive relation, so they could be used for `$==`, `==`, `<~>`, etc. That would probably help with these proofs!)

Yes, I also had this thought a few times to. I would definitely welcome such a PR.

PR #1944, opened 2024-05-02
Monoidal: convenient naturality lemmas

Alizter:
The curried naturality we ask for is compact, however it makes it difficult to use in practice. These lemmas show that it is a fair bit of work to actual get the more streamlined versions out.

Split from #1929

### DIFF
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -147,6 +147,60 @@ Class IsSymmetricMonoidal (A : Type) `{HasEquivs A}
   cat_symm_tensor_hexagon :: HexagonIdentity cat_tensor;
 }.
 
+(** *** Theory about [Associator] *)
+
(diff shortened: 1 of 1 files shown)


jdchristensen:
Do these results follow from some more general results about naturality of curried and uncurried bifunctors?

If not, is it any easier to deduce the last three from the first one, rather than from the definition?

Alizter:
> Do these results follow from some more general results about naturality of curried and uncurried bifunctors?

Probably, but I haven't had the time to work it out. We should definitely look into it in the future.

> If not, is it any easier to deduce the last three from the first one, rather than from the definition?

I'll check.

Alizter:
Nice, it worked. The proofs are much shorter and cleaner. This will make generalising this argument to general natural transformations with 3 variables easier in the future.

jdchristensen (approved):

PR #1947, opened 2024-05-02
Bool: add convenient variant of negb_ne

Alizter:
This is more useful in proof when you want to apply it to a hypothesis and then destruct.

### DIFF
--- a/theories/Types/Bool.v
+++ b/theories/Types/Bool.v
@@ -92,6 +92,16 @@ Proof.
   - intros oops; case (oops idpath).
 Defined.
 
+(** This version of [negb_ne] is more convenient to [destruct] against. *)
+Definition negb_ne' {b1 b2 : Bool}
+  : (b1 <> b2) -> (negb b1 = b2).
+Proof.
+  intros oops.
+  symmetry.
+  apply negb_ne.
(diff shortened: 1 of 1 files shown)


theories/Types/Bool.v
```diff
@@ -92,6 +92,16 @@ Proof.
...
 Defined.
 
+(** This version of [negb_ne] is more convenient to [destruct] against. *)
+Definition negb_ne' {b1 b2 : Bool}
+  : (b1 <> b2) -> (negb b1 = b2).
+Proof.
+  intros oops.
+  symmetry.
+  apply negb_ne.
+  intros p; symmetry in p; contradiction.
```
jdchristensen, line 102: Maybe this is a bit cleaner?
```suggestion
  exact (symmetric_neq oops).
```
(And then some of the lines could be merged.)

jdchristensen (approved):

theories/Types/Bool.v
Alizter: Ah I thought we had that. Maybe I should add a symmetric instance.

Alizter: Hmm, making that a typeclass instance doesn't seem to work which is probably why it wasn't one before.

PR #1951, opened 2024-05-04
[coq] Overlay for coq/coq#18385

ejgallego:


Alizter:
When I add the plugins field I get:
```
Error: path outside the workspace:       
../../../../../../nix/store/yb1h6vdd03mr0mlzznsx9yfd183f0n2x-coq-8.19.1/lib/ocaml/4.14.2/site-lib/coq-core/META
from default/theories
-> required by _build/default/theories/Basics.glob
-> required by alias theories/all
-> required by alias default in dune:44
```
So it seems there is some bad interaction with the plugin loading and noinit.

ejgallego:
That's an interesting bug due in Coq due to the workaround for https://github.com/ocaml/dune/issues/5833 that was implemented in coqdep (cc: https://github.com/coq/coq/pull/18165 )

So indeed, coqdep path normalization works fine for example in Coq's CI, however in build setups where Coq lives outside the build path, the normalization returns a path that is a bit non-sensical as it is relativized up to root.

We need to think a bit more how to handle this, best would be to solve Dune's bug.

theories/Basics/Settings.v
```diff
@@ -7,9 +7,9 @@
 (** ** Plugins *)
 
 (** Load the Ltac plugin. This is the tactic language we use for proofs. *)
-Declare ML Module "ltac_plugin".
+Declare ML Module "ltac_plugin:coq-core.plugins.ltac".
```
SkySkimmer, line 10: seems to work better with the semi-legacy loading syntax

SkySkimmer: (after https://github.com/coq/coq/pull/18385 the legacy part is ignored and prints a deprecation warning)

Alizter: Nice fix! I'm happy with this now.

Alizter (approved):

Issue #1933, opened 2024-04-27
Experiment with redefining bifunctors as uncurried functors


Alizter:
In #1883 I mentioned that we should consider redefining `Is0Bifunctor F` as `Is0Functor (ucurry F)`. This has a few advantages:

1. It allows us to use our "functorial-at-the-same-time" definitions.
2. We don't have extra terms with the fmap01 and fmap10 split.
3. Bifunctor coherence becomes a lemma and a way of constructing bifunctors.
4. It generalised much more easily to the ternary and n-ary case.

I think we would be able to make some arguments slicker by doing this, but I haven't spent anytime really checking. It may however present other problems that I haven't thought about. I'll record this issue so that we don't forget about it.

PR #1952, opened 2024-05-05
redefine Bifunctor

Alizter:
In this PR, we redefine `Is0Bifunctor` and `Is1Bifunctor` as discussed in #1933.

This has simplified things in lots of places however I got stuck with a universe issue in AbExt. @jdchristensen would you be able to take a look at that?

The reason I am doing this PR now is that I need some definitionally involutive opposite behaviour for bifunctors in #1929 and this isn't possible with the bifunctor coherence defintion.

### DIFF
--- a/theories/Algebra/AbSES/SixTerm.v
+++ b/theories/Algebra/AbSES/SixTerm.v
@@ -98,7 +98,8 @@ Proof.
       destruct g as [g k].
     exists g.
     apply path_sigma_hprop; cbn.
-    exact k^.
+    elim k^. 
+    by apply equiv_path_grouphomomorphism.
 Defined.
 
 (** *** Exactness of [ab_hom A G -> Ext1 B G -> Ext1 E G]. *)
--- a/theories/Algebra/AbSES/SixTerm.v
+++ b/theories/Algebra/AbSES/SixTerm.v
@@ -98,7 +98,7 @@ Proof.
       destruct g as [g k].
     exists g.
     apply path_sigma_hprop; cbn.
-    exact k^. 
+    exact k^.
 Defined.
 
 (** *** Exactness of [ab_hom A G -> Ext1 B G -> Ext1 E G]. *)
--- a/theories/Algebra/AbSES/SixTerm.v
+++ b/theories/Algebra/AbSES/SixTerm.v
@@ -98,8 +98,7 @@ Proof.
       destruct g as [g k].
     exists g.
     apply path_sigma_hprop; cbn.
-    elim k^. 
-    by apply equiv_path_grouphomomorphism.
+    exact k^. 
 Defined.
 
 (** *** Exactness of [ab_hom A G -> Ext1 B G -> Ext1 E G]. *)
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -562,7 +562,6 @@ Section FunctorJoin.
 
   Global Instance is0bifunctor_join : Is0Bifunctor Join.
   Proof.
-    rapply Build_Is0Functor.
     apply Build_Is0Functor.
     intros A B [f g].
     exact (functor_join f g).
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -572,7 +572,7 @@ Section FunctorJoin.
   Global Instance is1bifunctor_join : Is1Bifunctor Join.
   Proof.
     snrapply Build_Is1Bifunctor'.
-    snrapply Build_Is1Functor.
+    nrapply Build_Is1Functor.
     - intros A B f g [p q].
       exact (functor2_join p q).
     - intros A; exact functor_join_idmap.
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -562,6 +562,8 @@ Section FunctorJoin.
 
   Global Instance is0bifunctor_join : Is0Bifunctor Join.
   Proof.
+    snrapply Build_Is0Bifunctor'.
+    1,2: exact _.
     apply Build_Is0Functor.
     intros A B [f g].
     exact (functor_join f g).
@@ -569,6 +571,7 @@ Section FunctorJoin.
 
   Global Instance is1bifunctor_join : Is1Bifunctor Join.
   Proof.
+    snrapply Build_Is1Bifunctor'.
     snrapply Build_Is1Functor.
     - intros A B f g [p q].
       exact (functor2_join p q).
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -562,7 +562,7 @@ Section FunctorJoin.
 
   Global Instance is0bifunctor_join : Is0Bifunctor Join.
   Proof.
-    rapply Build_Is0Bifunctor'.
+    rapply Build_Is0Functor.
     apply Build_Is0Functor.
     intros A B [f g].
     exact (functor_join f g).
@@ -570,8 +570,7 @@ Section FunctorJoin.
 
   Global Instance is1bifunctor_join : Is1Bifunctor Join.
   Proof.
-    snrapply Build_Is1Bifunctor'.
-    nrapply Build_Is1Functor.
+    snrapply Build_Is1Functor.
     - intros A B f g [p q].
       exact (functor2_join p q).
     - intros A; exact functor_join_idmap.
--- a/theories/WildCat/Bifunctor.v
+++ b/theories/WildCat/Bifunctor.v
@@ -7,10 +7,49 @@ Require Import WildCat.Core WildCat.Prod WildCat.Equiv WildCat.NatTrans WildCat.
 (** ** Definition *)
 
 Class Is0Bifunctor {A B C : Type}
-  `{IsGraph A, IsGraph B, IsGraph C} (F : A -> B -> C)
-  := is0functor_bifunctor_uncurried :: Is0Functor (uncurry F).
+  `{IsGraph A, IsGraph B, IsGraph C} (F : A -> B -> C) := {
+  is0functor_bifunctor_uncurried :: Is0Functor (uncurry F);
+  is0functor01_bifunctor :: forall a, Is0Functor (F a);
+  is0functor10_bifunctor :: forall b, Is0Functor (flip F b);
+}.
 
-Global Instance is0functor01_bifunctor {A B C : Type}
+Definition Build_Is0Bifunctor' {A B C : Type}
+  `{Is01Cat A, Is01Cat B, IsGraph C} (F : A -> B -> C)
+  `{!Is0Functor (uncurry F)}
+  : Is0Bifunctor F.
+Proof.
+  snrapply Build_Is0Bifunctor.
+  - exact _.
+  - intros a.
+    snrapply Build_Is0Functor.
+    intros b b' g.
+    change (uncurry F (a, b) $-> uncurry F (a, b')).
+    refine (fmap (uncurry F) (_, _)).
+    + exact (Id a).
+    + exact g.
+  - intros b.
+    snrapply Build_Is0Functor.
+    intros a a' f.
+    change (uncurry F (a, b) $-> uncurry F (a', b)).
+    refine (fmap (uncurry F) (_, _)).
+    + exact f.
+    + exact (Id b).
+Defined.
+
+Definition Build_Is0Bifunctor'' {A B C : Type}
+  `{IsGraph A, IsGraph B, Is01Cat C} (F : A -> B -> C)
+  `{!forall a, Is0Functor (F a), !forall b, Is0Functor (flip F b)}
+  : Is0Bifunctor F.
+Proof.
+  snrapply Build_Is0Bifunctor.
+  2,3: exact _.
+  snrapply Build_Is0Functor.
+  intros [a b] [a' b'] [f g].
+  change (F a b $-> F a' b').
+  exact (fmap (flip F b') f $o fmap (F a) g).
+Defined.
+
+(* Global Instance is0functor01_bifunctor {A B C : Type}
   `{Is01Cat A, IsGraph B, IsGraph C} (F : A -> B -> C) `{!Is0Bifunctor F}
   : forall a, Is0Functor (F a).
 Proof.
@@ -21,9 +60,9 @@ Proof.
   refine (fmap (uncurry F) (_, _)).
   - exact (Id a).
   - exact g.
-Defined.
+Defined. *)
 
-Global Instance is0functor10_bifunctor {A B C : Type}
+(* Global Instance is0functor10_bifunctor {A B C : Type}
   `{IsGraph A, Is01Cat B, IsGraph C} (F : A -> B -> C) `{!Is0Bifunctor F}
   : forall b, Is0Functor (flip F b).
 Proof.
@@ -34,9 +73,9 @@ Proof.
   refine (fmap (uncurry F) (_, _)).
   - exact f.
   - exact (Id b).
-Defined.
+Defined. *)
 
-Definition Build_Is0Bifunctor {A B C : Type}
+(* Definition Build_Is0Bifunctor {A B C : Type}
   `{IsGraph A, IsGraph B, Is01Cat C} (F : A -> B -> C)
   `{!forall a, Is0Functor (F a), !forall b, Is0Functor (flip F b)}
   : Is0Bifunctor F.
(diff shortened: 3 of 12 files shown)


theories/WildCat/Bifunctor.v
```diff
@@ -1,117 +1,191 @@
...
+    nrapply bifunctor_coh.
+Defined.
 
 (** ** Bifunctor lemmas *)
 
 (** *** 1-functorial action *)
 
 (** [fmap] in the first argument. *)
-Definition fmap10 {A B C : Type} `{IsGraph A, IsGraph B, IsGraph C}
+Definition fmap10 {A B C : Type} `{IsGraph A, Is01Cat B, IsGraph C}
```
jdchristensen, line 130: Why do you need a strong assumption on `B` here?  (And same for `A` in the next result.)

theories/WildCat/Bifunctor.v
```diff
@@ -1,117 +1,191 @@
...
+  : fmap01 F a1 g $o fmap10 F f b0 $== fmap10 F f b1 $o fmap01 F a0 g
+  := (fmap11_is_fmap01_fmap10 _ _ _)^$ $@ fmap11_is_fmap10_fmap01 _ _ _.
+
 (** [fmap11] with right map the identity gives [fmap10]. *)
 Definition fmap10_is_fmap11 {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
   (F : A -> B -> C) `{!Is0Bifunctor F, !Is1Bifunctor F}
   {a0 a1 : A} (f : a0 $-> a1) (b : B)
   : fmap11 F f (Id b) $== fmap10 F f b
-  := (fmap_id _ _ $@R _) $@ cat_idl _.
+  := Id _.
```
jdchristensen, line 181: I'm guessing this is definitionally true.  If so, maybe it's better to record it as a definitional equality, so people can use `change` or `rewrite_refl`?  (Same with next one.)

theories/WildCat/Bifunctor.v
```diff
@@ -230,69 +306,56 @@ Definition emap11 {A B C : Type} `{HasEquivs A, HasEquivs B, HasEquivs C}
...
-Hint Immediate is1bifunctor_flip : typeclass_instances.
 
 (** ** Composition of bifunctors *)
 
 (** There are 4 different ways to compose a functor with a bifunctor. *)
 
 (** Restricting a functor along a bifunctor yields a bifunctor. *)
 Global Instance is0bifunctor_postcompose {A B C D : Type}
-  `{IsGraph A, IsGraph B, IsGraph C, IsGraph D}
+  `{Is01Cat A, Is01Cat B, Is01Cat C, Is01Cat D}
```
jdchristensen, line 353: Also extra hypotheses here.  (And other places.  I won't mention them again.)

theories/Homotopy/Join/Core.v
```diff
@@ -562,16 +562,15 @@ Section FunctorJoin.
 
   Global Instance is0bifunctor_join : Is0Bifunctor Join.
   Proof.
-    rapply Build_Is0Bifunctor'.
+    rapply Build_Is0Functor.
```
jdchristensen, line 565: This line can be removed.
```suggestion
```

theories/WildCat/Bifunctor.v
```diff
@@ -230,69 +306,56 @@ Definition emap11 {A B C : Type} `{HasEquivs A, HasEquivs B, HasEquivs C}
...
+
+(** Here is an alternative constructor that uses [fmap01] and [fmap10] instead. *)
+Definition Build_Is1Bifunctor' {A B C : Type}
+  `{Is1Cat A, Is1Cat B, Is1Cat C} (F : A -> B -> C)
+  `{!forall a, Is0Functor (F a), !forall b, Is0Functor (flip F b)}
+  `{!forall a, Is1Functor (F a), !forall b, Is1Functor (flip F b)}
+  (Is0Bifunctor_F := Build_Is0Bifunctor F)
+  (bifunctor_coh : forall a0 a1 (f : a0 $-> a1) b0 b1 (g : b0 $-> b1),
+    fmap01 F a1 g $o fmap10 F f b0 $== fmap10 F f b1 $o fmap01 F a0 g)
+  : Is1Bifunctor F.
```
jdchristensen, line 337: Isn't the type of this definitionally the same as the type of `Build_Is1Bifunctor`?  If so, is this needed?   ...  Oh, I see what is going on here, and it is very confusing.  In the context of the declaration of `bifunctor_coh`, there are two ways that `fmap01` could be interpreted.  One is using `Is0Bifunctor_F` and the instances in this file to deduce `Is0Functor (f a1)`, for example.  And the other is to use the hypothesis that gives `Is0Functor (f a1)`.  I think this would be more clear if the `Is0Functor` and `Is1Functor` assumptions were named, and used by name in place of `fmap01`, etc.

But do we really need this?  Maybe the place(s) that use it can be converted to defining a functor on the product category right from the start?

theories/WildCat/Bifunctor.v
Alizter: In order to have `fmap10` the way I defined it as `fmap11 F (f, Id _)` I need to have a 1-functor for identity and composition laws. This means that we cannot just assume `IsGraph`. Alternatively, I could redefine `fmap10` to be `fmap (flip F a) f` but I haven't tried this yet.

theories/WildCat/Bifunctor.v
Alizter: Yes, these could probably be removed. I've simply updated the proofs and not gone over the needs yet.

Alizter:
@jdchristensen I have some applications in mind for this. It has helped simplify the opposite of a monoidal category. I will work a bit more on that so that we can be sure this is worth it.

The reason I am interested in that is so that I can write down a cocommutative comonoid in a monoidal category which is just a commutative monoid in the opposite monoidal category.

theories/Homotopy/Join/Core.v
Alizter: Thanks, fixed.

theories/WildCat/Bifunctor.v
Alizter: Yes, I'm not sure why I changed these now. I'll revert those changes and see if we can weaken any others.

theories/WildCat/Bifunctor.v
Alizter: Actually ignore what I said about redefining. That's exactly the current definition.

theories/WildCat/Bifunctor.v
Alizter: Removing these simplified a few more things.

theories/WildCat/Bifunctor.v
Alizter: I am confused. I can't remember why I introduced this, presumably because the terms appearing in Ext.v were different to the ones I expected. I need to spend some more time thinking about this.

theories/WildCat/Bifunctor.v
jdchristensen: Nice, that's further evidence that this approach is better.

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -98,7 +98,8 @@ Proof.
       destruct g as [g k].
     exists g.
     apply path_sigma_hprop; cbn.
-    exact k^.
+    elim k^. 
+    by apply equiv_path_grouphomomorphism.
```
Alizter, line 102: This is unfortunate, but the proofs of this map being a group homomorphism changed.

jdchristensen: I'll discuss this at the top-level.

jdchristensen:
The approach to bifunctors in this PR causes some awkwardness that I'd like to discuss, and which showed up in SixTerm.v.  `ab_hom A B` is the abelian group of group homomorphisms from `A` to `B`.  It's a 0-functor in each variable in a natural way using pre- and post-composition.  This makes it into a 0-bifunctor.  However, when you then apply `fmap10` to this bifunctor (using the set-up in this PR) along a map `f : A -> A'`, you get the map sending `g : A' -> B` to `Id $o g $o f`, with an extra identity map involved.  (We could instead use `fmap (flip ab_hom B) f`, but I couldn't get Coq to accept this in the context it comes up for some reason, even with various implicit arguments provided.)  The same thing will happen for hom functors in general.  And even for more general bifunctors, if we construct them by starting with a 0-functor in each variable, it would seem better if `fmap10` and `fmap01` returned the same definitions that we provided.

Couldn't we define the bifunctor class so that it remembers the functoriality in each variable and returns it when `fmap10` and `fmap01` are applied, and then have wrappers that provide constructors for how it is in this PR as well as accessor functors that provide the data used in this PR?  I think this roughly describes how things are currently in master branch.  It should be possible to use this set-up to make things as clean as they are in this PR, but retain the good behaviour of `fmap10` and `fmap01`.  In other words, I'm suggesting that if we go back to how things are in master, we should be able to find places where using the constructors/accessors I describe allow things to simplify like they do in this PR.

@Alizter said that he wanted this approach for other reasons related to opposite categories.  Maybe those definitional equalities can still be achieved if things are done carefully?  Or, if not, we have to think about whether it is worth having more complicated `fmap10` and `fmap01`.

Alizter:
Yes, this is an unfortunate side-effect and I am not happy with it. I will attempt to mix the two definitions and see if we can get the best of both worlds. This will mean I have to reintroduce the `fmap01_is_fmap11` homotopies back as they would become data in the definition.

jdchristensen:
> I will attempt to mix the two definitions and see if we can get the best of both worlds.

This isn't what I meant above, but I was also thinking about it as an option.  The class would have the fmap01, fmap10 and fmap11 data, plus a single(?) law saying that fmap11 can be expressed in terms of the other two.  (The laws about expressing fmap10 and fmap01 in terms of fmap11 should follow, I think.)  And in addition to the default constructor, there would be one taking just the fmap10 and fmap01 data and deriving fmap11.  And there would be one taking just fmap11 and deriving fmap10 and fmap01 (using Is01Cat structures).  Then depending on which constructor you used, some of the data you get back when using the instance would agree with what you put it.

That might work.  But it might also work to just have the fmap01 and fmap10 data in the class, like in current master, and use wrappers for fmap11 and the fmap11 version of the constructor.

Alizter:
I'm currently trying this out:
```
Class Is0Bifunctor {A B C : Type}
  `{IsGraph A, IsGraph B, IsGraph C} (F : A -> B -> C) := {
  is0functor_bifunctor_uncurried :: Is0Functor (uncurry F);
  is0functor01_bifunctor :: forall a, Is0Functor (F a);
  is0functor10_bifunctor :: forall b, Is0Functor (flip F b);
}.

...

Class Is1Bifunctor {A B C : Type}
  `{Is1Cat A, Is1Cat B, Is1Cat C} (F : A -> B -> C) `{!Is0Bifunctor F} := {

  is1functor_bifunctor_uncurried :: Is1Functor (uncurry F);
  is1functor01_bifunctor :: forall a, Is1Functor (F a);
  is1functor10_bifunctor :: forall b, Is1Functor (flip F b);

  fmap11_is_fmap01_fmap10 {a0 a1} (f : a0 $-> a1) {b0 b1} (g : b0 $-> b1)
    : fmap11 F f g $== fmap01 F a1 g $o fmap10 F f b0;
  fmap11_is_fmap10_fmap01 {a0 a1} (f : a0 $-> a1) {b0 b1} (g : b0 $-> b1)
    : fmap11 F f g $== fmap10 F f b1 $o fmap01 F a0 g;
}.
```

jdchristensen:
> I'm currently trying this out:

That looks reasonable!  It makes sense to have those two laws, not just one like I said.  And I think many bifunctors naturally have efficient descriptions of all three functorialities, so they might just use the default constructors.

mikeshulman:
I haven't looked at this particular case at all, but just wanted to point out that sometimes when there's more than one possible definition of something, it's useful to have both of them in a library, along with the equivalence between them.  Then the more convenient definition can be used whenever a choice is possible, and we control when to go across the equivalence.

Alizter:
> I haven't looked at this particular case at all, but just wanted to point out that sometimes when there's more than one possible definition of something, it's useful to have both of them in a library, along with the equivalence between them. Then the more convenient definition can be used whenever a choice is possible, and we control when to go across the equivalence.

In this particular case we have two equivalent definitions that each have convenient behaviour in different places. We are looking into generalising both at the same time to have both sets of behaviours whilst still specialising to both definitions.

Alizter:
I've pushed the best-of-both-worlds definition and it builds. I haven't gone over the proofs in AbSES yet to make sure they are simpler.

Alizter:
It seems to have worked. Overall many things got renamed, but there is a clear advantage in the join formalization. IINM this should allow us to use the twist construction directly there.

theories/WildCat/Yoneda.v
```diff
@@ -44,13 +44,20 @@ Proof.
...
 
 Global Instance is0bifunctor_hom {A} `{Is01Cat A}
-  : Is0Bifunctor (A:=A^op) (B:=A) (C:=Type) (@Hom A _)
-  := is0functor_hom.
+  : Is0Bifunctor (A:=A^op) (B:=A) (C:=Type) (@Hom A _).
+Proof.
+  nrapply Build_Is0Bifunctor'.
+  1-2: exact _.
+  exact is0functor_hom.
+Defined.
```
jdchristensen: I wonder if other places that use fmap01 and fmap10 for this bifunctor would get simpler if we used `Build_Is0Bifunctor` and provided the more direct functoriality in each variable as well?

jdchristensen (commented):
I'm not quite done my review, but will continue later.  I'm also about to push some simplifications.

theories/WildCat/Bifunctor.v
```diff
@@ -74,44 +53,90 @@ Definition fmap01 {A B C : Type} `{IsGraph A, IsGraph B, IsGraph C}
...
+  - exact (is1functor_functor_uncurried01 (uncurry F)).
+  - exact (is1functor_functor_uncurried10 (uncurry F)).
+  - intros a0 a1 f b0 b1 g.
+    refine (fmap2 (uncurry F) _^$ $@ fmap_comp (uncurry F)
+      (a := (a0, b0)) (b := (a1, b0)) (c := (a1, b1)) (f, Id b0) (Id a1, g)).
+    exact (cat_idl _, cat_idr _).
+  - intros a0 a1 f b0 b1 g.
+    refine (fmap2 (uncurry F) _^$ $@ fmap_comp (uncurry F)
+      (a := (a0, b0)) (b := (a0, b1)) (c := (a1, b1)) (Id a0, g) (f, Id b1)).
+    exact (cat_idr _, cat_idl _).
```
jdchristensen, line 101: I wonder if the last two bullets should be abstracted out and put into Prod.v?

theories/WildCat/Bifunctor.v
```diff
@@ -74,44 +53,90 @@ Definition fmap01 {A B C : Type} `{IsGraph A, IsGraph B, IsGraph C}
...
+  - snrapply Build_Is1Functor.
+    + intros [a b] [a' b'] [f g] [f' g'] [p p']; unfold fst, snd in * |- .
+      exact (fmap2 (F a) p' $@@ fmap2 (flip F b') p).
+    + intros [a b].
+      exact ((fmap_id (F a) b $@@ fmap_id (flip F b) _) $@ cat_idr _).
+    + intros [a b] [a' b'] [a'' b''] [f g] [f' g']; unfold fst, snd in * |- .
+      refine ((fmap_comp (F a) g g' $@@ fmap_comp (flip F b'') f f') $@ _).
+      nrefine (cat_assoc_opp _ _ _ $@ (_ $@R _) $@ cat_assoc _ _ _).
+      refine (cat_assoc _ _ _ $@ (_ $@L _^$) $@ cat_assoc_opp _ _ _).
+      nrapply bifunctor_coh.
```
jdchristensen, line 123: Maybe this bullet is also a lemma that can be put into Prod.v?

theories/WildCat/Bifunctor.v
```diff
@@ -136,27 +107,138 @@ Definition fmap11 {A B C : Type} `{IsGraph A, IsGraph B, IsGraph C}
...
+      exact ((fmap_id (F a) b $@@ fmap_id (flip F b) _) $@ cat_idr _).
+    + intros [a b] [a' b'] [a'' b''] [f g] [f' g']; unfold fst, snd in * |- . 
+      refine ((fmap_comp (F a) g g' $@@ fmap_comp (flip F b'') f f') $@ _).
+      nrefine (cat_assoc_opp _ _ _ $@ (_ $@R _) $@ cat_assoc _ _ _).
+      refine (cat_assoc _ _ _ $@ (_ $@L _^$) $@ cat_assoc_opp _ _ _).
+      nrapply bifunctor_coh.
+  - exact _.
+  - exact _.
+  - intros a0 a1 f b0 b1 g.
+    exact (bifunctor_coh a0 a1 f b0 b1 g)^$.
```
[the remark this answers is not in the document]
Alizter: I tried this, but IIRC I had to switch direction in other places as well.

theories/WildCat/Bifunctor.v
Alizter: I'm unsure what you mean. What would the lemma look like?

theories/WildCat/Bifunctor.v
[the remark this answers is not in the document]
jdchristensen: > Alternatively, we could split the assumption into two.

Not sure what you mean by this.  But another alternative is to change the definition of `Is0Bifunctor''` to be the other choice.  But then probably lots of other things would change, and there's little benefit.

theories/WildCat/Bifunctor.v
jdchristensen: `test` below works for the first bullet:
```coq
Definition fmap_pair {A B C : Type}
  `{IsGraph A, IsGraph B, IsGraph C}
  (F : A * B -> C) `{!Is0Functor F}
  {a0 a1 : A} (f : a0 $-> a1) {b0 b1 : B} (g : b0 $-> b1)
  : F (a0, b0) $-> F (a1, b1)
  := fmap (a:=(a0,b0)) (b:=(a1,b1)) F (f,g).

Definition test {A B C : Type}
  `{Is1Cat A, Is1Cat B, Is1Cat C}
  (F : A * B -> C) `{!Is0Functor F, !Is1Functor F}
  : forall (a0 a1 : A) (f : a0 $-> a1) (b0 b1 : B) (g : b0 $-> b1),
    fmap_pair F f g $== fmap (fun b => F (a1,b)) g $o fmap (fun a => F(a,b0)) f.
Admitted.
```
It fits in Prod.v, since it is about the 0-functor instances defined there.

theories/WildCat/Bifunctor.v
```diff
@@ -124,8 +149,11 @@ Definition fmap02 {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
 Definition fmap12 {A B C : Type} `{Is1Cat A, Is1Cat B, Is1Cat C}
   (F : A -> B -> C) `{!Is0Bifunctor F, !Is1Bifunctor F}
   {a0 a1 : A} (f : a0 $-> a1) {b0 b1 : B} {g g' : b0 $-> b1} (q : g $== g')
-  : fmap11 F f g $== fmap11 F f g'
-  := fmap02 F _ q $@R _.
+  : fmap11 F f g $== fmap11 F f g'.
+Proof.
+  refine (fmap2 (uncurry F) _).
+  exact (Id _, q).
```
jdchristensen, line 155: It's annoying that in many places where we treat `uncurry F` as a functor, we either need to split a term across two tactics like this, or specify `(a:=(a0,b0)) (b:=(a1,b1))`.  Is there a way we can get Coq to recognize that `(f,g)` is a morphism in `A * B` in all of these places?  Maybe a wrapper `prod_mor f g` that produces `(f,g)`?

theories/WildCat/Products.v
```diff
@@ -558,12 +538,8 @@ Definition cat_pr1_fmap10_binprod {A : Type} `{HasBinaryProducts A}
...
 Definition cat_pr1_fmap11_binprod {A : Type} `{HasBinaryProducts A}
   {w x y z : A} (f : w $-> y) (g : x $-> z)
-  : cat_pr1 $o fmap11 (fun x y => cat_binprod x y) f g $== f $o cat_pr1.
-Proof.
-  refine ((cat_assoc _ _ _)^$ $@ (_ $@R _) $@ cat_assoc _ _ _ $@ _).
-  1: nrapply cat_binprod_beta_pr1.
-  exact (cat_idl _ $@ cat_binprod_beta_pr1 _ _).
-Defined.
+  : cat_pr1 $o fmap11 (fun x y => cat_binprod x y) f g $== f $o cat_pr1
+  := cat_binprod_beta_pr1 _ _.
```
jdchristensen, line 542: I wonder if this lemma and others nearby are still needed, or if one can just use `cat_binprod_beta_pr1`?

jdchristensen (approved):
I've now finished reviewing, and think it looks good.  There are still a few comments that could maybe be addressed, but it's ok if they don't work out.

theories/WildCat/Bifunctor.v
Alizter: I wondered the same thing. I think I encountered a similar issue with cubes and squares in the torus proof. I'm not sure if there is a way to hint unification in these cases.

theories/WildCat/Bifunctor.v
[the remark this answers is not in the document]
Alizter: I guess I wasn't thinking straight earlier. I was thinking of the `fun11_is_fun01_fun10` coherences but they don't make sense in this context.

theories/WildCat/Bifunctor.v
Alizter: This doesn't fall into that category since it is specifically about the instances built in Build_Is0Bifunctor''. If you unfold the proof here you will see that it isn't the usual uncurry instance but rather one where it is built from the fmap in each parameter.

theories/WildCat/Bifunctor.v
Alizter: I've fixed this with the new fmap_pair stuff in Prod.

theories/WildCat/Yoneda.v
Alizter: One thing I would like to use this in is the twist construction for the smash product. I was thinking before that we would run into the awkwardness that join did and that it would be less manageable since the smash is tricky to work with. Now however I think everything is in place to prove the associativity of smash. We just need to construct the twist map.

theories/WildCat/Bifunctor.v
Alizter: Thanks, I've created that and some related lemmas and used them in bifunctor.

theories/WildCat/Products.v
Alizter: I would opt to keep this as it also serves as documentation on how `fmap11` is suppsoed to behave. Its not always easy to remember how you've defined `fmap11` in a proof so having an explicit lemma like this can be helpful. It doesn't do any harm since it is equivalent to the thing you should use.

theories/WildCat/Bifunctor.v
Alizter: I may have misunderstood what you meant however.

Alizter:
I will merge once the CI is green.

theories/WildCat/Bifunctor.v
jdchristensen: I'll show what I meant in a fresh PR.

PR #1953, opened 2024-05-07
monoids and comonoids in a monoidal category

Alizter:
Here is a definition of monoid and comonoid object in a monoidal category.

They live in a new file called `Algebra/Categorical/MonoidObject.v`. This should be a place for "categorical algebra" that is theory about algebraic objects defined using category theory.

In order to define a comonoid as a monoid in the opposite monoidal category we need to be able to take take the opposite monoidal category. This requires restructuring how we define things in `Monoidal.v` using natural transformations and equivalences where we can. The work in #1952 helped some properties hold more easily. The result is that the monoidal structure is inherited in the opposite category in a fairly straightforward way, modulo moving some inverted morphisms around.

As an "application" we show that `x $-> y` is a (commutative) monoid when `y` is a (commutative) monoid object in a cartesian category. (Equivalently, `x` can be a comonoid). This greatly simplifies the analogous proof in #1929 leaving us only to show that objects of additive categories are comonoids or monoids (whichever turns out to be eaiser).

I've taken the time to also simplify a few proofs in Products.v where new lemmas I've introduced can break up the work.

### DIFF
--- a/theories/Algebra/Categorical/MonoidObject.v
+++ b/theories/Algebra/Categorical/MonoidObject.v
@@ -1,0 +1,197 @@
+Require Import Basics.Overture Basics.Tactics.
+Require Import WildCat.Core WildCat.Equiv WildCat.Monoidal WildCat.Bifunctor
+  WildCat.NatTrans WildCat.Opposite WildCat.Products.
+Require Import abstract_algebra.
+
+(** * Monoids and Comonoids *)
+
+(** Here we define a monoid internal to a monoidal category. Various algebraic theories such as groups and rings may also be internalized, however these specifically require a cartesian monoidal structure. The theory of monoids however has no such requirement and can therefore be developed in much greater generality. This can be used to define a range of objects such as R-algebras, H-spaces, Hopf algebras and more. *)
+
+(** * Monoid objects *)
+
+Section MonoidObject.
+  Context {A : Type} {tensor : A -> A -> A} {unit : A}
+    `{HasEquivs A, !Is0Bifunctor tensor, !Is1Bifunctor tensor}
+    `{!Associator tensor, !LeftUnitor tensor unit, !RightUnitor tensor unit}.
+
+  (** An object [x] of [A] is a monoid object if it comes with the following data: *)
+  Class IsMonoidObject (x : A) := {
+    (** A multiplication map from the tensor product of [x] with itself to [x]. *)
+    mo_mult : tensor x x $-> x;
+    (** A unit of the multplication. *)
+    mo_unit : unit $-> x;
+    (** The multiplication map is associative. *)
+    mo_assoc : mo_mult $o fmap10 tensor mo_mult x $o associator x x x
+        $== mo_mult $o fmap01 tensor x mo_mult;
+    (** The multiplication map is left unital. *)
+    mo_left_unit : mo_mult $o fmap10 tensor mo_unit x $== left_unitor x;
+    (** The multiplication map is right unital. *)
+    mo_right_unit : mo_mult $o fmap01 tensor x mo_unit $== right_unitor x;
+  }.
+
+  Context `{!Braiding tensor}.
+
+  (** An object [x] of [A] is a commutative monoid object if: *)
+  Class IsCommutativeMonoidObject (x : A) := {
+    (** It is a monoid object. *)
+    cmo_mo :: IsMonoidObject x;
+    (** The multiplication map is commutative. *)
+    cmo_comm : mo_mult $o braid x x $== mo_mult;
+  }.
+
+End MonoidObject.
+
+Arguments IsMonoidObject {A} tensor unit {_ _ _ _ _ _ _ _ _ _} x.
+Arguments IsCommutativeMonoidObject {A} tensor unit {_ _ _ _ _ _ _ _ _ _ _} x.
+
+Section ComonoidObject.
+  Context {A : Type} (tensor : A -> A -> A) (unit : A)
+    `{HasEquivs A, !Is0Bifunctor tensor, !Is1Bifunctor tensor}
+    `{!Associator tensor, !LeftUnitor tensor unit, !RightUnitor tensor unit}.
+
+  (** A comonoid object is a monoid object in the opposite category. *)
+  Class IsComonoidObject (x : A)
+    := ismonoid_comonoid_op :: IsMonoidObject (A:=A^op) tensor unit x.
+
+  (** We can build comonoid objects from the following data: *)
+  Definition Build_IsComonoidObject (x : A)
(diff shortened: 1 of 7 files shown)


theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -0,0 +1,197 @@
...
+   := mo_mult (A:=A^op) (tensor:=tensor) (unit:=unit) (x:=x).
+
+  (** Counit *)
+  Definition co_counit {x : A} `{!IsComonoidObject x} : x $-> unit
+    := mo_unit (A:=A^op) (tensor:=tensor) (unit:=unit) (x:=x).
+
+  Context `{!Braiding tensor}.
+
+  (** A cocommutative comonoid objects is a commutative monoid object in the opposite category. *)
+  Class IsCocommuativeComonoidObject (x : A)
```
Alizter, line 99:
```suggestion
  Class IsCocommutativeComonoidObject (x : A)
```

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -0,0 +1,197 @@
...
+    := iscommuatativemonoid_cocomutativemonoid_op
+      :: IsCommutativeMonoidObject (A:=A^op) tensor unit x.
+
+  (** We can build cocommutative comonoid objects from the following data: *)
+  Definition Build_IsCocommuatativeComonoidObject (x : A)
```
Alizter, line 104:
```suggestion
  Definition Build_IsCocommutatativeComonoidObject (x : A)
```

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -0,0 +1,197 @@
...
+    := iscommuatativemonoid_cocomutativemonoid_op
+      :: IsCommutativeMonoidObject (A:=A^op) tensor unit x.
+
+  (** We can build cocommutative comonoid objects from the following data: *)
+  Definition Build_IsCocommutatativeComonoidObject (x : A)
+    (** A comonoid. *)
+    `{!IsComonoidObject x}
+    (** Together with a proof of cocommutativity. *)
+    (cco_cocomm : braid x x $o co_comult $== co_comult)
+    : IsCocommuativeComonoidObject x.
```
Alizter, line 109:
```suggestion
    : IsCocommutativeComonoidObject x.
```

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -0,0 +1,197 @@
...
+
+  Context `{!Braiding tensor}.
+
+  (** A cocommutative comonoid objects is a commutative monoid object in the opposite category. *)
+  Class IsCocommutativeComonoidObject (x : A)
+    := iscommuatativemonoid_cocomutativemonoid_op
+      :: IsCommutativeMonoidObject (A:=A^op) tensor unit x.
+
+  (** We can build cocommutative comonoid objects from the following data: *)
+  Definition Build_IsCocommutatativeComonoidObject (x : A)
```
Alizter, line 104:
```suggestion
  Definition Build_IsCocommutativeComonoidObject (x : A)
```
Spelling is hard.

theories/WildCat/Bifunctor.v
```diff
@@ -389,3 +396,71 @@ Proof.
...
   exact (hconcat (nat_l _ _ _ f) (nat_r _ _ _ f')).
 Defined.
+
+(** Flipping a natural transformation between bifunctors. *)
+Definition nattrans_flip {A B C : Type}
+  `{Is1Cat A, Is1Cat B, Is1Cat C}
+  {F : A -> B -> C} `{!Is0Bifunctor F, !Is1Bifunctor F}
+  {G : B -> A -> C} `{!Is0Bifunctor G, !Is1Bifunctor G}
+  : NatTrans (uncurry F) (uncurry (flip G))
+    -> NatTrans (uncurry (flip F)) (uncurry G).
```
jdchristensen, line 406: How is `(uncurry (flip G))` being interpreted as a functor here? I think it uses `is1bifunctor_flip`, right?

It seems to me like the natural statement would have no `flip`s in the hypotheses and two `flip`s in the final conclusion.  And then, I was hoping that since `flip` is involutive on functions, the above would follow, as would the primed version below.  But I think that `is1bifunctor_flip` is probably not definitionally involutive, so maybe we end up needing to state four versions if we want to cover all possibilities?

theories/WildCat/Bifunctor.v
```diff
@@ -14,6 +15,11 @@ Class Is0Bifunctor {A B C : Type}
   is0functor10_bifunctor :: forall b, Is0Functor (flip F b);
 }.
 
+Arguments Is0Bifunctor {A B C _ _ _} F.
+Arguments is0functor_bifunctor_uncurried {A B C _ _ _} F {_}.
+Arguments is0functor01_bifunctor {A B C _ _ _} F {_} a : rename.
+Arguments is0functor10_bifunctor {A B C _ _ _} F {_} b : rename.
+
```
jdchristensen, line 22: Line 41 below has a stack overflow.  For some reason, with other changes, that line needs to be `nrapply Build_Is0Bifunctor; exact _`.  But it would be good to figure out what is causing the stack to overflow.

jdchristensen (commented):
So far I just looked at the WildCat part.  I read the new file when I have more time.

jdchristensen (approved):
Nice!

theories/WildCat/Bifunctor.v
Alizter: I tried to figure out what was causing the stack overflow but to no avail. FTR limiting the typeclass depth to 1 locally allows the search to succeed. It seems we are wasting a lot of time searching for instances that we don't care about. In order to fix this, I tried setting `Hint Mode` for `Is0Functor` and `Is1Functor` but I couldn't get anything to work. The documentation on hint mode is not great and I don't really know what I am doing.

For now I will just do your suggestion.

theories/WildCat/Bifunctor.v
Alizter: > How is `(uncurry (flip G))` being interpreted as a functor here? I think it uses `is1bifunctor_flip`, right?

Yes, that is correct.

> It seems to me like the natural statement would have no `flip`s in the hypotheses and two `flip`s in the final conclusion. And then, I was hoping that since `flip` is involutive on functions, the above would follow, as would the primed version below. But I think that `is1bifunctor_flip` is probably not definitionally involutive, so maybe we end up needing to state four versions if we want to cover all possibilities?

I changed it to having no `flip`s in the hypotheses and it appears to work. We don't even actually use the primed version so I've removed that. If we ever need it we can add it back.

PR #1954, opened 2024-05-07
WildCat: generalize a couple of proofs from Bifunctor.v to Prod.v

jdchristensen:
A couple of the embedded proofs in Bifunctor.v can be generalized slightly.  Not sure if this is really an improvement.  This relates to the discussion in #1952.


Alizter (approved):
Nice.

PR #1956, opened 2024-05-13
abstract out some idioms for dealing with decidable types

Alizter:
This commit is from #1929. I ended up using this elsewhere outside of #1929 so I deemed it was of interest on its own.

### DIFF
--- a/theories/Basics/Decidable.v
+++ b/theories/Basics/Decidable.v
@@ -42,7 +42,7 @@ Proof.
   - contradiction n.
 Defined.
 
-(** Replace a term [p] of the form [Decidable A] with [inl x] if we know [A] is true. *)
+(** Replace a term [p] of the form [Decidable A] with [inl x] if we have a term [a : A] showing that [A] is true. *)
 Ltac decidable_true p a :=
   generalize p;
   rapply (decidable_true a);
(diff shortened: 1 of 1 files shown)


theories/Basics/Decidable.v
```diff
@@ -31,6 +31,40 @@ Ltac decide :=
...
+  (P : forall (p : Decidable A), Type)
+  (p : forall x, P (inl x))
+  : forall p, P p.
+Proof.
+  intros [x|n].
+  - apply p.
+  - contradiction n.
+Defined.
+
+(** Replace a term [p] of the form [Decidable A] with [inl x] if we know [A] is true. *)
```
jdchristensen, line 45:
```suggestion
(** Replace a term [p] of the form [Decidable A] with [inl x] if we have a term [a : A] showing that [A] is true. *)
```

theories/Basics/Decidable.v
```diff
@@ -31,6 +31,40 @@ Ltac decide :=
...
+  intros [x|n'].
+  - contradiction n.
+  - apply p.
+Defined.
+
+(** Replace a term [p] of the form [Decidable A] with [inr na] if we know [A] is false. *)
+Ltac decidable_false p a :=
+  generalize p;
+  rapply (decidable_false a);
+  try intro.
```
jdchristensen, line 66:
```suggestion
(** Replace a term [p] of the form [Decidable A] with [inr na] if we have a term [n : not A] showing that [A] is false. *)
Ltac decidable_false p n :=
  generalize p;
  rapply (decidable_false n);
  try intro.
```

jdchristensen (approved):

PR #1958, opened 2024-05-14
Use correct scheme kind for CarriersTermAlgebra_ind

SkySkimmer:
Elimination is documented as nonrecursive although incorrectly behaves like Induction until coq/coq#19017


Alizter:
Thanks! LGTM

PR #1959, opened 2024-05-14
finite sums of ring elements

Alizter:
In this PR we define finite sums of ring elements and prove some basic properties about them. We also introduce the Kronecker delta function and prove some standard results about its interaction with sums.

Depends on #1956.


theories/Algebra/Rings/Ring.v
```diff
@@ -468,3 +468,232 @@ Proof.
   refine ((rng_mult_assoc _ _ _)^ @ _).
   exact (ap (x *.) IHn).
 Defined.
+
+(** ** Finite Sums *)
```
jdchristensen, line 472: Almost all of this material works in any abelian group, so should be done in that generality.  And the stuff about Kronecker delta functions should move elsewhere (and most could be generalized from nat to any decidable type).  (I know this is still a draft, but I thought I'd still add my two cents.)

Alizter: How does the Kronecker delta work for an abelian group? I don't see how you can define it without 1 in addition to 0.

Alizter: Also something else I would like to experiment with when I find the time is to do generic sums over finite sets. This comes up when defining the determinant of a matrix as a sum over the symmetric group for instance.

Alizter: Ah I see I misunderstood your comment earlier. I guess it makes sense to move Kronecker delta to its own file in rings.

Alizter:
I've rebased and split off the Kronecker delta stuff, generalising it to an arbitrary decidable type where applicable. I'll see about moving things about sums to abelian groups.

Alizter:
@jdchristensen I've generalised sums from rings to abelian groups. However when the sum takes place in a ring, I've kept the lemma names as `rng_sum`. We could specialise `ab_sum` to rings with `rng_sum` however it doesn't seem to be a big issue reusing it for now.

jdchristensen:
This looks better.  It's actually possible to get rid of the ring assumption for most of the material on the Kronecker delta.  It could be defined to land in `nat`, and then we could use the fact that in any abelian group you can multiply by a natural number (which we write as `grp_pow`).  Also, there should be a coercion from `nat` to any ring `R`, and applying the coercion followed by multiplying in `R` is the same as `grp_pow` in the underlying abelian group.  So then we'd be able to use lemmas like `rng_sum_kronecker_delta_l` in any abelian group.  But maybe this would just make practical usage of this trickier?

jdchristensen:
Feel free to either merge as is, or to try changing the Kronecker delta to land in `nat`, whichever you prefer.

Alizter:
I'll leave generalising the Kronecker delta to a future PR. The only application I have in mind for this currently is for the identity matrix in #1965. The sum identities here allow us to show the identity matrix is a left and right identity.

PR #1961, opened 2024-05-15
more opposite tests

Alizter:
We add more tests showing functors and now bifunctors are definitionally involutive. We also show that opposite natural transformations do not have this behaviour since the proof of naturality is inverted. I think it might be time to consider adding redundant information to `Is1Natural` so that we also keep the inverted naturality around like we did for `cat_assoc_opp`.

### DIFF
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -23,7 +23,7 @@ Definition test6 A B (F : A -> B) `{x : Is0Functor A B F}
   : @is0functor_op _ _ F _ _ (@is0functor_op _ _ F _ _ x) = x
   := 1.
 Definition test7 A B (F : A -> B) `{x : Is1Functor A B F}
-  : @is1functor_op _ _ F _ _ _ _ _ _ _ _ _(@is1functor_op _ _ F _ _ _ _ _ _ _ _ _ x) = x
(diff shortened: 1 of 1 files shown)


test/WildCat/Opposite.v
```diff
@@ -17,3 +17,26 @@ Definition core4 A `{HasEquivs A} : is2graph_op (A:=core A) = is2graph_core (A:=
 (** The Opaque line reduces the time from 0.3s to 0.07s. *)
 Opaque compose_catie_isretr.
 Definition core5 A `{HasEquivs A} : is1cat_op (A:=core A) = is1cat_core (A:=A^op) := 1.
+
+(** Opposite functors are definitionally involutive. *)
+Definition test6 A B (F : A -> B) `{x : Is0Functor A B F}
+  : @is0functor_op _ _ F _ _ (@is0functor_op _ _ F _ _ x) = x
+  := 1.
+Definition test7 A B (F : A -> B) `{x : Is1Functor A B F}
+  : @is1functor_op _ _ F _ _ _ _ _ _ _ _ _(@is1functor_op _ _ F _ _ _ _ _ _ _ _ _ x) = x
```
Alizter, line 26:
```suggestion
  : @is1functor_op _ _ F _ _ _ _ _ _ _ _ _ (@is1functor_op _ _ F _ _ _ _ _ _ _ _ _ x) = x
```

jdchristensen (approved):

PR #1966, opened 2024-05-18
Prove that the universal property of pullback is an equivalence

MintChocoManju:
I couldn't find this proof anywhere in the library, although this result maybe of interest somewhere else.


jdchristensen (approved):
Thanks for this!  I just pushed some minor style cleanups.  I'll let Ali take a look and merge if he approves.

Alizter (approved):

PR #1968, opened 2024-05-27
cleanup NatTrans.v

Alizter:
We cleanup NatTrans.v by doing the following:

- Transformations, natuarlity lemmas, and natural transformations each appear separately in their own sections making it easier to see what lemmas are available for each.
- There are now more comments explaining what is happening.
- Is1Natural is no longer a definitional typeclass in preperation for adding the transposed naturality condition to the data of a natural transformation.
- Whitespace style has been homogenized.

### DIFF
--- a/theories/WildCat/NatTrans.v
+++ b/theories/WildCat/NatTrans.v
@@ -31,14 +31,14 @@ Definition trans_comp {A B : Type} `{Is01Cat B}
   : F $=> K
   := fun a => gamma a $o alpha a.
 
-(** Transformations can be prewhiskered by a function. This means we precompose either side of the transformation with a function. *)
+(** Transformations can be prewhiskered by a function. This means we precompose both sides of the transformation with a function. *)
 Definition trans_prewhisker {A B : Type} {C : B -> Type} {F G : forall x, C x}
   `{Is01Cat B} `{!forall x, IsGraph (C x)}
   `{!forall x, Is01Cat (C x)} (gamma : F $=> G) (K : A -> B) 
   : F o K $=> G o K
   := gamma o K.
 
-(** Transformations can be postwhiskered by a function. This means we postcompose either side of the transformation with a function. *)
+(** Transformations can be postwhiskered by a function. This means we postcompose both sides of the transformation with a function. *)
 Definition trans_postwhisker {A B C : Type} {F G : A -> B}
   (K : B -> C) `{Is01Cat B, Is01Cat C, !Is0Functor K} (gamma : F $=> G)
   : K o F $=> K o G
@@ -221,7 +221,7 @@ Definition issig_NatEquiv {A B : Type} `{IsGraph A} `{HasEquivs B}
   (F G : A -> B) `{!Is0Functor F, !Is0Functor G}
   : _ <~> NatEquiv F G := ltac:(issig).
 
-(** From a given natural transformation, we can get the underlying natural transformation. *)
+(** From a given natural equivalence, we can get the underlying natural transformation. *)
 Lemma nattrans_natequiv {A B : Type} `{IsGraph A} `{HasEquivs B}
   {F G : A -> B} `{!Is0Functor F, !Is0Functor G}
   : NatEquiv F G -> NatTrans F G.
--- a/theories/WildCat/NatTrans.v
+++ b/theories/WildCat/NatTrans.v
@@ -4,8 +4,13 @@ Require Import WildCat.Equiv.
 Require Import WildCat.Square.
 Require Import WildCat.Opposite.
 
-(** ** Natural transformations *)
+(** * Wild Natural Transformations *)
 
+(** ** Transformations *)
+
+(** *** Definition *)
+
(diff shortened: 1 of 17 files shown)


theories/WildCat/NatTrans.v
```diff
@@ -15,198 +20,192 @@ Identity Coercion fun_trans : Transformation >-> Funclass.
...
-  rapply is1natural_id.
-Defined.
-
+(** Transformations can be composed pointwise. *)
 Definition trans_comp {A B : Type} `{Is01Cat B}
            {F G K : A -> B} (gamma : G $=> K) (alpha : F $=> G)
   : F $=> K
   := fun a => gamma a $o alpha a.
 
+(** Transformations can be prewhiskered by a function. This means we precompose either side of the transformation with a function. *)
```
jdchristensen, line 34:
```suggestion
(** Transformations can be prewhiskered by a function. This means we precompose both sides of the transformation with a function. *)
```

theories/WildCat/NatTrans.v
```diff
@@ -15,198 +20,192 @@ Identity Coercion fun_trans : Transformation >-> Funclass.
...
 Definition trans_prewhisker {A B : Type} {C : B -> Type} {F G : forall x, C x}
   `{Is01Cat B} `{!forall x, IsGraph (C x)}
   `{!forall x, Is01Cat (C x)} (gamma : F $=> G) (K : A -> B) 
   : F o K $=> G o K
   := gamma o K.
 
+(** Transformations can be postwhiskered by a function. This means we postcompose either side of the transformation with a function. *)
```
jdchristensen, line 41:
```suggestion
(** Transformations can be postwhiskered by a function. This means we postcompose both sides of the transformation with a function. *)
```

theories/WildCat/NatTrans.v
```diff
@@ -15,198 +20,192 @@ Identity Coercion fun_trans : Transformation >-> Funclass.
...
-  - exact (transformation_op F G alpha).
-  - exact _.
-Defined.
+    -> NatTrans (A:=A^op) (B:=B^op) (G : A^op -> B^op) (F : A^op -> B^op)
+  := fun alpha => Build_NatTrans (trans_op F G alpha) _.
 
-(** Natural equivalences *)
+(** ** Natural equivalences *)
 
+(** Natural equivalences are families of equivlaences that are natural. *)
```
jdchristensen, line 204:
```suggestion
(** Natural equivalences are families of equivalences that are natural. *)
```

theories/WildCat/NatTrans.v
```diff
@@ -220,9 +219,7 @@ Definition issig_NatEquiv {A B : Type} `{IsGraph A} `{HasEquivs B}
   (F G : A -> B) `{!Is0Functor F, !Is0Functor G}
   : _ <~> NatEquiv F G := ltac:(issig).
 
-Global Existing Instance is1natural_natequiv.
-Coercion cat_equiv_natequiv : NatEquiv >-> Funclass.
-
+(** From a given natural transformation, we can get the underlying natural transformation. *)
```
jdchristensen, line 222:
```suggestion
(** From a given natural transformation, we can get the underlying natural transformation. *)
```
```suggestion
(** From a given natural equivalence, we can get the underlying natural transformation. *)
```

theories/WildCat/NatTrans.v
```diff
@@ -15,198 +20,192 @@ Identity Coercion fun_trans : Transformation >-> Funclass.
...
+  (F : A -> B) (G : A -> B) (alpha : F $=> G)
+  : Transformation (A:=A^op) (B:=fun _ => B^op) G (F : A^op -> B^op)
+  := alpha.
+
+(** ** Naturality *)
+
+(** A transformation is 1-natural if there exists a 2-cell witnessing the naturality square. The codomain of the transformation must be a wild 1-category. *)
+Class Is1Natural {A B : Type} `{IsGraph A, Is1Cat B}
+  (F : A -> B) `{!Is0Functor F} (G : A -> B) `{!Is0Functor G}
+  (alpha : F $=> G) := Build_Is1Natural' {
```
jdchristensen, line 58: What is there a prime on `Build_Is1Natural'`?  Can we just omit the name and then by default get the unprimed version?

jdchristensen (approved):

theories/WildCat/NatTrans.v
Alizter: This is primed for the time being as I will introduce a field with the duplicated field soon. There I want to see more clearly where I use the future `Build_Is1Natural` constructor which will prove the transposed naturality proof instead of being given it.

Alizter:
OK this is ready now. I previously I didn't build the entire library by accident which meant that I missed adding `snrapply Build_Is1Natural'` in various places. I've updated that now. As mentioned earlier, I will introduce a version of `Is1Natural` where there is a field for the transposed naturality proof. There I want to be clear that I am using `Build_Is1Natural` or `Build_Is1Natural'` so I will make all the versions primed first.

Alizter:
I've checked and all the primed versions will go away in the next PR adding the extra field.

PR #1969, opened 2024-05-27
make opposite Is1Natural definitionally involutive

Alizter:
Carry out the suggestion in #1961. This will allow for more of the data of a monoidal category or (2,1)-category to be definitionally involutive.

### DIFF
--- a/test/WildCat/Opposite.v
+++ b/test/WildCat/Opposite.v
@@ -35,8 +35,14 @@ Definition test9 A B C (F : A -> B -> C) `{x : Is1Bifunctor A B C F}
     (@is1bifunctor_op _ _ _ F _ _ _ _ _ _ _ _ _ _ _ _ _ x) = x
   := 1.
 
-(** Opposite natural transformations are *not* definitionally involutive. *)
-Fail Definition test10 A `{Is01Cat A} B `{Is1Cat B} (F G : A -> B)
+(** Opposite natural transformations are definitionally involutive. *)
+Definition test10 A `{Is01Cat A} B `{Is1Cat B} (F G : A -> B)
   `{!Is0Functor F, !Is0Functor G} (n : NatTrans F G)
   : nattrans_op (nattrans_op n) = n
+  := 1.
+
+(** Opposite natural equivalences are definitionally involutive. *)
+Definition test11 A `{Is01Cat A} B `{HasEquivs B} (F G : A -> B)
+  `{!Is0Functor F, !Is0Functor G} (n : NatEquiv F G)
+  : natequiv_op (natequiv_op n) = n
   := 1.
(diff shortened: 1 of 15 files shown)


theories/WildCat/NatTrans.v
```diff
@@ -145,8 +151,10 @@ Global Instance is1natural_op A B `{Is01Cat A} `{Is1Cat B}
 Proof.
   unfold op.
   snrapply Build_Is1Natural'.
-  intros a b f.
-  srapply isnat_tr.
+  - intros a b f.
+    by snrapply isnat_tr.
+  - intros a b f.
+    by snrapply isnat.
```
jdchristensen, line 157: Not sure if it makes much different here, but this is a bit more explicit, and avoids doing an `intro` on `f`, which in some cases reduces the size of the term, but doesn't seem to matter here.
```suggestion
  - intros a b.
    exact (isnat_tr alpha).
  - intros a b.
    exact (isnat alpha).
```

jdchristensen (approved):

Alizter:
Turns out we don't get definitionally involutive associators and unitors since we are taking inverses there. I have updated the tests to show the current state of the opposite involutions. This isn't strictly needed for #1929 anyway.

test/WildCat/Opposite.v
```diff
@@ -1,42 +1,72 @@
...
-  WildCat.NatTrans WildCat.Bifunctor.
+  WildCat.NatTrans WildCat.Bifunctor WildCat.Monoidal.
 
 (** Opposites are definitionally involutive. *)
-Definition test1 A : A = (A^op)^op :> Type := 1.
-Definition test2 A `{x : IsGraph A} : x = @isgraph_op A^op (@isgraph_op A x) := 1.
-Definition test3 A `{x : Is01Cat A} : x = @is01cat_op A^op _ (@is01cat_op A _ x) := 1.
-Definition test4 A `{x : Is2Graph A} : x = @is2graph_op A^op _ (@is2graph_op A _ x) := 1.
-Definition test5 A `{x : Is1Cat A} : x = @is1cat_op A^op _ _ _ (@is1cat_op A _ _ _ x) := 1.
+Succeed Definition t A : A = (A^op)^op :> Type := 1.
```
Alizter, line 5: `Succeed` reverses the effect after a successful command so we can give all the tests the same name. This prevents us from having to track the numbering in the tests.

jdchristensen: > `Succeed` reverses the effect after a successful command so we can give all the tests the same name. This prevents us from having to track the numbering in the tests.

Oh, nice!

jdchristensen: The `t` looks like it might be an argument, so maybe `test` is a better placeholder name?

Alizter: Sure, I'll change it to be longer.

jdchristensen:
LGTM.

PR #1970, opened 2024-05-28
further coherence conditions for monoidal categories

Alizter:
Here are the extra axioms that MacLane gave for mmonoidal categories that were later shown to be redundant by Kelly.

At least one of these actually appears in the monoidal coherence proof so it is not something that follows from coherence for monoidal categories.

cc @patrick-nicodemus who stated two of these in the first (wild) monoidal categories formalization but didn't finish them.

### DIFF
--- a/theories/WildCat/Equiv.v
+++ b/theories/WildCat/Equiv.v
@@ -484,6 +484,12 @@ Proof.
   exact (cate_buildequiv_fun _)^$.
 Defined.
 
+Definition emap_inv' {A B : Type} `{HasEquivs A} `{HasEquivs B}
+  (F : A -> B) `{!Is0Functor F, !Is1Functor F}
+  {a b : A} (e : a $<~> b)
+  : cate_fun (emap F e)^-1$ $== fmap F e^-1$
+  := emap_inv F e $@ cate_buildequiv_fun _.
+
 (** When we have equivalences, we can define what it means for a category to be univalent. *)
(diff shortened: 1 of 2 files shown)


patrick-nicodemus:
This is really nice.
Do you know if there has been any progress on the issue where importing the standard library zeroes out the typeclass database and replaces it with the empty typeclass database defined in the standard library?
Because it would be nice to have setoid rewriting for these kinds of theorems.

Alizter:
@patrick-nicodemus I have not been tracking the progress on generalized rewriting on the Coq side. I also haven't had a chance to experiment with it in the library as of yet.

I think the issue of generalized rewriting and noinit actually stems from the issue of using Program without the standard library. This is something that has been known to be unsatisfactory for some time, but I only recently created an issue for it: https://github.com/coq/coq/issues/19033

If we have the ability to register library constants in Program and by extension SetoidRewrite then we could register those directly ourselves without having to worry about importing the Coq standard library.

I don't however believe any of the Coq devs are looking into this at the moment.

patrick-nicodemus:
That's interesting. I was not aware that Setoid rewriting depends (nominally?) on Program.

Alizter:
Ltac2 depends on Ltac but could in theory be separated. There has been work in this direction I believe. The overlap is due to syntax parsing of ltac/ltac2 grammar.

For Elpi, it should be more straightforward to support without the Coq stdlib. It may already be the case today, though I haven't checked.

PR #1973, opened 2024-06-01
injectivity of nat addition

Alizter:
This is a simple lemma that comes in useful later. We could probably give a shorter proof with commutativity of nat addition, but I would prefer to keep this earlier on.

### DIFF
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -268,20 +268,6 @@ Proof.
   auto.
 Defined.
 
-Definition isinj_nat_add_l k : forall x y, k + x = k + y -> x = y.
-Proof.
-  simple_induction' k; simpl; auto.
-Defined.
-
-Definition isinj_nat_add_r k x y (H : x + k = y + k) : x = y.
-Proof.
-  induction k as [|k IHk] in x, y, k, H |- *.
-  - by rewrite <- !add_n_O in H.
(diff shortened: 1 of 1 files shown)


theories/Spaces/Nat/Core.v
```diff
@@ -268,6 +268,20 @@ Proof.
   auto.
 Defined.
 
+Definition isinj_nat_add k : forall x y, k + x = k + y -> x = y.
+Proof.
+  simple_induction' k; simpl; auto.
+Defined.
+
+Definition isinj_nat_add' x y k (H : x + k = y + k) : x = y.
```
jdchristensen, line 276:
```suggestion
Definition isinj_nat_add_l k : forall x y, k + x = k + y -> x = y.
Proof.
  simple_induction' k; simpl; auto.
Defined.

Definition isinj_nat_add_r k x y (H : x + k = y + k) : x = y.
```
I changed the names, and also changed the argument order for the second one. Now `isinj_nat_add_r k` is the statement that addition on the right by `k` is injective.

jdchristensen (approved):
I don't see why you wouldn't want to put these after commutativity of addition, to make the second one follow from the first.

Alizter:
I can't remember my reasoning now so it probably wasn't very good. I think I may have confused myself about where the commutativity lemma was and perhaps thought it was in Arithmetic.

PR #1976, opened 2024-06-01
interaction of scalar and matrix multiplication

Alizter:
We show that scalar multiplication distributes over the first factor of matrix multiplication. We also clarify that something similar should hold for right multiplication and the right factor.

We also clarify an earlier comment about an R-algebra over a noncommutative ring R. It's not a priori clear what the definition should be in this case. R-matrices form an R-bimodule with a product that interacts with the left and right actions nicely. This might be a good candidate for a definition, but it is not clear if it is standard.

When the ring R is commutative then it is clear that this becomes the usual definition of an associative algebra. Equivalently, a map `R -> Z(A)` where A is the algebra and Z is the center.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -225,7 +225,7 @@ Proof.
   - exact (right_identity_matrix_mult R n n).
 Defined.
 
-(** Matrix multiplication on the right preserves scalar multiplication. *)
+(** Matrix multiplication on the right preserves scalar multiplication in the sense that [matrix_lact r (matrix_mult M N) = matrix_mult (matrix_lact r M) N] for [r] a ring element and [M] and [N] matrices of compatible sizes. *)
 Definition matrix_mult_lact_l {R : Ring} {m n p : nat}
   : HeteroAssociative (@matrix_lact R m p) (@matrix_mult R m n p)
       (@matrix_mult R m n p) (@matrix_lact R m n).
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -208,7 +208,7 @@ Proof.
   nrapply rng_sum_kronecker_delta_r'.
 Defined.
 
-(** TODO: define this as an R-algebra *)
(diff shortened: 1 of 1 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -225,6 +225,24 @@ Proof.
   - exact (right_identity_matrix_mult R n n).
 Defined.
 
+(** Matrix multiplication on the right preserves scalar multiplication. *)
+Definition matrix_mult_lact_l {R : Ring} {m n p}
```
jdchristensen, line 229:
```suggestion
Definition matrix_mult_lact_l {R : Ring} {m n p : nat}
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -225,6 +225,24 @@ Proof.
   - exact (right_identity_matrix_mult R n n).
 Defined.
 
+(** Matrix multiplication on the right preserves scalar multiplication. *)
```
jdchristensen, line 228:
```suggestion
(** Matrix multiplication on the right preserves scalar multiplication in the sense that [matrix_lact r (matrix_mult M N) = matrix_mult (matrix_lact r M) N] for [r] a ring element and [M] and [N] matrices of compatible sizes. *)
```
(The heteroassociativity statement is hard to decipher.)

jdchristensen (approved):
This makes me think that we should probably have defined a matrix as a vector of *column* vectors.  Then, a matrix product `M * N` is the list of the `M * v` for `v` a column vector in `N`.  So many things about matrix multiplication can first be proved for the action of matrices on vectors, and then generalized to matrix products.

The dual statement for rows is also true, but it's more common to think of matrices acting on the left on column vectors, and the library is currently focusing on left actions.  So we'd be in better harmony with this bias if matrices were lists of column vectors.  But it's probably not worth changing everything (again) at this point.

theories/Algebra/Rings/Matrix.v
Alizter: Yes, maybe its best to state it directly. The typeclass probably won't be needed.

Alizter:
I think it should be possible to define matrices as a vector of columns. It might be worth trying if we move to defining them as functions as mentioned in #1979.

PR #1971, opened 2024-05-30
subrings

Alizter:
We introduce the notion of a subring, some (very) basic theory about it and redefine the image of a ring homomorphism as a subring.

This will be useful later for giving subrings of matrix rings.

PR #1977, opened 2024-06-01
theory of diagonal matrices

Alizter:
We write down the property of being a diagonal matrix, show that this is a hprop (how nice), and show that standard opertions on matrices preserve diagonal matrices. This lets us define the subring of diagonal matrices of the ring of matrices.

Depends on:
- #1971
- #1978

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -480,8 +480,13 @@ Definition matrix_diag_vector {R : Ring} {n : nat} (M : Matrix R n n)
 (** Diagonal matrices form a subring of the ring of square matrices. *)
 Definition matrix_diag_ring (R : Ring) (n : nat) : Subring (matrix_ring R n).
 Proof.
-  snrapply (Build_Subring' (fun M : matrix_ring R n => IsDiagonal M));
-    cbn beta; exact _.
+  snrapply (Build_Subring' (fun M : matrix_ring R n => IsDiagonal M) _); hnf.
+  - intros; exact _.
+  - intros x y dx dy.
+    nrapply isdiagonal_matrix_plus; trivial.
+    by nrapply isdiagonal_matrix_negate.
+  - nrapply isdiagonal_matrix_mult.
+  - nrapply isdiagonal_identity_matrix.
 Defined.
 
 (** ** Trace *)
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -310,16 +310,178 @@ Defined.
 
 (** ** Diagonal matrices *)
 
-(** A diagonal matrix is a matrix with zeros everywhere except on the diagonal. *)
-Definition diagonal_matrix {R : Ring@{i}} {n : nat} (v : list R) (p : length v = n)
+(** A diagonal matrix is a matrix with zeros everywhere except on the diagonal. It's entries are given by a vector. *)
+Definition matrix_diag {R : Ring@{i}} {n : nat} (v : Vector R n)
   : Matrix R n n.
 Proof.
   snrapply Build_Matrix.
   intros i j H1 H2.
-  destruct (dec (i = j)).
+  exact (kronecker_delta i j * Vector.entry v i).
+Defined.
+
+(** Addition of matrices is the same as addition of the corresponding vectors. *)
+Definition matrix_diag_plus {R : Ring@{i}} {n : nat} (v w : Vector R n)
+  : matrix_plus (matrix_diag v) (matrix_diag w) = matrix_diag (vector_plus v w).
+Proof.
+  snrapply path_matrix.
+  intros i j Hi Hj.
+  rewrite 2 entry_Build_Matrix, 4 entry_Build_Vector.
+  cbv beta zeta; destruct (dec (i = j)) as [p | np].
   - destruct p.
-    exact (nth' v i H1).
-  - exact 0.
+    rewrite kronecker_delta_refl.
+    rewrite <- rng_dist_l.
+    by rewrite entry_Build_Vector.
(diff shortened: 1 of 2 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
 
 (** ** Diagonal matrices *)
 
-(** A diagonal matrix is a matrix with zeros everywhere except on the diagonal. *)
-Definition diagonal_matrix {R : Ring@{i}} {n : nat} (v : list R) (p : length v = n)
+(** A diagonal matrix is a matrix with zeros everywhere except on the diagonal. It's entries are given by a vector. *)
```
jdchristensen, line 313:
```suggestion
(** A diagonal matrix is a matrix with zeros everywhere except on the diagonal. Its entries are given by a vector. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
...
+Definition matrix_diag {R : Ring@{i}} {n : nat} (v : Vector R n)
   : Matrix R n n.
 Proof.
   snrapply Build_Matrix.
   intros i j H1 H2.
-  destruct (dec (i = j)).
+  exact (kronecker_delta i j * Vector.entry v i).
+Defined.
+
+(** Addition of matrices is the same as addition of the corresponding vectors. *)
```
jdchristensen, line 322:
```suggestion
(** Addition of diagonal matrices is the same as addition of the corresponding vectors. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
...
+  rewrite 2 entry_Build_Matrix, 4 entry_Build_Vector.
+  cbv beta zeta; destruct (dec (i = j)) as [p | np].
   - destruct p.
-    exact (nth' v i H1).
-  - exact 0.
+    rewrite kronecker_delta_refl.
+    rewrite <- rng_dist_l.
+    by rewrite entry_Build_Vector.
+  - rewrite (kronecker_delta_neq np).
+    by rewrite 3 rng_mult_zero_l, rng_plus_zero_l.
```
jdchristensen, line 335: This is more direct, with no case split:
```coq
  symmetry.
  snrapply path_matrix.
  intros i j Hi Hj.
  rewrite 2 entry_Build_Matrix, 5 entry_Build_Vector.
  nrapply rng_dist_l.
```
If the equality in the statement is reversed, then the `symmetry` can be avoided too.

theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
...
+    = matrix_diag (vector_map2 (.*.) v w).
+Proof.
+  snrapply path_matrix.
+  intros i j Hi Hj.
+  rewrite 2 entry_Build_Matrix.
+  lhs snrapply path_ab_sum.
+  { intros k Hk.
+    rewrite 2 entry_Build_Matrix.
+    rewrite rng_mult_assoc.
+    rewrite <- (rng_mult_assoc (kronecker_delta _ _ )).
```
jdchristensen, line 350:
```suggestion
    rewrite <- (rng_mult_assoc (kronecker_delta _ _)).
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
...
+Definition matrix_transpose_diag {R : Ring} {n : nat} (v : Vector R n)
+  : matrix_transpose (matrix_diag v) = matrix_diag v.
+Proof.
+  snrapply path_matrix.
+  intros i j Hi Hj.
+  rewrite 3 entry_Build_Matrix.
+  rewrite kronecker_delta_symm.
+  unfold kronecker_delta.
+  destruct (dec (i = j)) as [p|np].
+  1: destruct p; f_ap; f_ap; apply path_ishprop.
```
jdchristensen, line 368: I think this line should be a lemma saying that if `i = j`, then `entry v i = entry v j`, and it should use a corresponding lemma about `nth'`.

theories/Algebra/Rings/Matrix.v
```diff
@@ -310,16 +310,178 @@ Defined.
...
+(** Given a square matrix, we can extract the diagonal as a vector. *)
+Definition matrix_diag_vector {R : Ring} {n : nat} (M : Matrix R n n)
+  : Vector R n
+  := Build_Vector R n (fun i _ => entry M i i).
+
+(** Diagonal matrices form a subring of the ring of square matrices. *)
+Definition matrix_diag_ring (R : Ring) (n : nat) : Subring (matrix_ring R n).
+Proof.
+  snrapply (Build_Subring' (fun M : matrix_ring R n => IsDiagonal M));
+    cbn beta; exact _.
```
jdchristensen, line 484: This two lines are slow.  Can you speed them up?

theories/Algebra/Rings/Matrix.v
```diff
@@ -365,7 +360,7 @@ Proof.
   rewrite kronecker_delta_symm.
   unfold kronecker_delta.
   destruct (dec (i = j)) as [p|np].
-  1: destruct p; f_ap; f_ap; apply path_ishprop.
+  1: destruct p; f_ap; apply path_entry_vector.
```
jdchristensen, line 363: I was thinking that the `p : i = j` would be an argument to `path_entry_vector`, so the `destruct` would also move there.

Alizter: I've done this but unfortunately, we require `j = i` rather than `i = j` so I had to use a symmetry.

jdchristensen (approved):

PR #1978, opened 2024-06-01
more lemmas about matrix transpose

Alizter:
We prove that:
- matrix transposes preserve negation
- matrix transposes preserve zero
- matrix transposes preserve the identity matrix

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -298,7 +298,7 @@ Proof.
   by rewrite !entry_Build_Matrix.
 Defined.
 
-(** The transpose of the idenetity matrix is the identity matrix. *)
+(** The transpose of the identity matrix is the identity matrix. *)
 Definition matrix_transpose_identity {R : Ring} {n}
   : matrix_transpose (identity_matrix R n) = identity_matrix R n.
 Proof.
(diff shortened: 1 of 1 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -276,6 +289,25 @@ Proof.
...
+(** The transpose of the zero matrix is the zero matrix. *)
+Definition matrix_transpose_zero {R : Ring} {m n}
+  : matrix_transpose (matrix_zero R m n) = matrix_zero R n m.
+Proof.
+  apply path_matrix.
+  intros i j Hi Hj.
+  by rewrite !entry_Build_Matrix.
+Defined.
+
+(** The transpose of the idenetity matrix is the identity matrix. *)
```
jdchristensen, line 301:
```suggestion
(** The transpose of the identity matrix is the identity matrix. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -276,6 +289,25 @@ Proof.
...
+Definition matrix_transpose_identity {R : Ring} {n}
+  : matrix_transpose (identity_matrix R n) = identity_matrix R n.
```
jdchristensen, line 303: This is true more generally for diagonal matrices.  Maybe the identity matrix should be defined as a diagonal matrix?  Or if that makes other things too complicated, then there should be a lemma that the identity matrix is equal to the diagonal matrix for the constant vector of ones, and then this would follow from the corresponding factor for diagonal matrices.

Alizter: I did this initially, but it was annoying to have `1 * kronecker_delta` around. Later in the PR about diagonal matrices we state the property of being a diagonal matrix and prove the general version too.

jdchristensen (approved):

PR #1979, opened 2024-06-01
more theory about matrix traces

Alizter:
We prove that:
- The trace of a matrix sum is the sum of traces.
- The trace preserves scalar multiplication (it is a linear map)
- In a commutative ring, a cyclic permutation of a product of matrices has invariant trace.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -319,7 +319,8 @@ Proof.
 Defined.
 
 (** The trace of a matrix multiplication is the same as the trace of the reverse multiplication. This holds only in a commutative ring. *)
-Definition matrix_trace_mult {R : CRing} {n} (M N : Matrix R n n)
+Definition matrix_trace_mult {R : CRing} {m n : nat}
+  (M : Matrix R m n) (N : Matrix R n m)
   : matrix_trace (matrix_mult M N) = matrix_trace (matrix_mult N M).
 Proof.
   lhs nrapply path_ab_sum.
@@ -336,7 +337,8 @@ Proof.
 Defined.
 
 (** More generally, the trace of a matrix is invariant under cyclic permutations. This fails for other permutations. *)
-Definition matrix_trace_mult_triple {R : CRing} {n} (M N P : Matrix R n n)
+Definition matrix_trace_mult_triple {R : CRing} {m n p}
+  (M : Matrix R m n) (N : Matrix R n p) (P : Matrix R p m)
   : matrix_trace (matrix_mult (matrix_mult M N) P)
     = matrix_trace (matrix_mult (matrix_mult P M) N).
 Proof.
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -336,30 +336,6 @@ Proof.
   reflexivity.
 Defined.
 
-(** More generally, the trace of a matrix is invariant under cyclic permutations. This fails for other permutations. *)
-Definition matrix_trace_mult_triple {R : CRing} {m n p}
-  (M : Matrix R m n) (N : Matrix R n p) (P : Matrix R p m)
-  : matrix_trace (matrix_mult (matrix_mult M N) P)
-    = matrix_trace (matrix_mult (matrix_mult P M) N).
-Proof.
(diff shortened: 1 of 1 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -296,23 +296,68 @@ Defined.
 Definition matrix_trace {R : Ring} {n} (M : Matrix R n n) : R
   := ab_sum n (fun i Hi => entry M i i).
 
-(** The trace of a matrix multiplication is the same as the trace of the reverse multiplication. *)
+(** The trace of a matrix preserves addition. *)
+Definition matrix_trace_plus {R : Ring} {n} (M N : Matrix R n n)
+  : matrix_trace (matrix_plus M N) = (matrix_trace M) + (matrix_trace N).
```
jdchristensen, line 301: Why doesn't `M + N` work on the LHS?  We should figure that out and use `+` here and in other places too.

theories/Algebra/Rings/Matrix.v
```diff
@@ -296,23 +296,68 @@ Defined.
...
+Proof.
+  unfold matrix_trace.
+  lhs nrapply path_ab_sum.
+  { intros i Hi.
+    by rewrite entry_Build_Matrix. }
```
jdchristensen, line 306: It's too bad that none of our operations on vectors and matrices compute, requiring us to use things like `entry_Build_Matrix`.  This is because we represent them with lists but work with them using functions on a prefix of nat.  Would it work to *define* them to be functions on a prefix of nat?  (Or should we use functions on `Fin n`?)

(This shouldn't block this PR, but is something to think about for the future.)

theories/Algebra/Rings/Matrix.v
```diff
@@ -296,23 +296,68 @@ Defined.
...
+  1,2: cbn; rewrite entry_Build_Matrix.
+  1,2: rewrite rng_sum_dist_r.
+  1,2: nrapply path_ab_sum.
+  1,2: intros k Hk.
+  2: rewrite <- rng_mult_assoc, rng_mult_comm.
+  1,2: reflexivity.
+  rewrite ab_sum_sum.
+  lhs snrapply path_ab_sum.
+  1: intros i Hi; by rewrite ab_sum_sum.
+  by rewrite ab_sum_sum.
```
jdchristensen, line 360:
```suggestion
  lhs nrapply matrix_trace_mult.
  apply ap, associative_matrix_mult.
```

jdchristensen (approved):

theories/Algebra/Rings/Matrix.v
Alizter: I didn't make those an instance, so unless they are `x : abgroup_matrix` they won't be interpreted that way. I can do a cleanup of that in another PR, it should be possible to make it work.

theories/Algebra/Rings/Matrix.v
Alizter: I think we should be able to define a matrix as a finite map. IIRC it is possible to prove funext for functions from a finite domain so we could define matrices directly as functions. This would also generalise nicely to functions with finite support allowing us to write down examples of "infinite matrices" which provide nice counter examples in ring theory.

This will of course have to come later as I have to think about it some more.

theories/Algebra/Rings/Matrix.v
Alizter: Oh yes of course. Maybe it's not even worth stating the result then?

theories/Algebra/Rings/Matrix.v
Alizter: I actually experimented with this idea previously, but working with `Fin` was a bit more awkward and working directly with `{k : nat & k < n}` seemed much better. This is essentially what our matrix entry functions are mapping from.

theories/Algebra/Rings/Matrix.v
jdchristensen: Maybe it's not worth it, unless it's needed somewhere.  But one similar thing that does explicitly come up a lot is that the trace of N^-1 M N is equal to the trace of M.

Alizter: It's not needed anywhere for now. The lemma about the conjugate could probably be added if and when we define matrix inverses.

PR #1981, opened 2024-06-02
upper/lower triangular matrix rings

Alizter:
We define upper and lower triangular matrices, show that they satisfy various closure properties and define the subring of upper and lower triangular matrices.

Depends on:

- #1980 
- #1977

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -339,7 +339,7 @@ Proof.
 Defined.
 
 (** The transpose of the identity matrix is the identity matrix. *)
-Definition matrix_transpose_identity {R : Ring} {n}
+Definition matrix_transpose_identity@{i} {R : Ring@{i}} {n}
   : matrix_transpose (identity_matrix R n) = identity_matrix R n.
 Proof.
   apply path_matrix.
@@ -391,7 +391,7 @@ Proof.
 Defined.
 
 (** The transpose of a diagonal matrix is the same diagonal matrix. *)
-Definition matrix_transpose_diag {R : Ring} {n : nat} (v : Vector R n)
+Definition matrix_transpose_diag {R : Ring@{i}} {n : nat} (v : Vector R n)
   : matrix_transpose (matrix_diag v) = matrix_diag v.
 Proof.
   snrapply path_matrix.
@@ -405,7 +405,7 @@ Proof.
 Defined.
 
 (** The diagonal matrix construction is injective. *)
-Global Instance isinj_matrix_diag {R : Ring} {n : nat}
+Global Instance isinj_matrix_diag {R : Ring@{i}} {n : nat}
   : IsInjective (@matrix_diag R n).
 Proof.
   intros v1 v2 p.
@@ -418,7 +418,7 @@ Proof.
 Defined.
 
 (** A matrix is diagonal if it is equal to a diagonal matrix. *)
-Class IsDiagonal {R : Ring@{i}} {n : nat} (M : Matrix R n n) : Type@{i} := {
+Class IsDiagonal@{i} {R : Ring@{i}} {n : nat} (M : Matrix R n n) : Type@{i} := {
   isdiagonal_diag_vector : Vector R n;
   isdiagonal_diag : M = matrix_diag isdiagonal_diag_vector;
 }.
@@ -426,12 +426,12 @@ Class IsDiagonal {R : Ring@{i}} {n : nat} (M : Matrix R n n) : Type@{i} := {
 Arguments isdiagonal_diag_vector {R n} M {_}.
 Arguments isdiagonal_diag {R n} M {_}.
 
-Definition issig_IsDiagonal {R : Ring} {n : nat} {M : Matrix R n n}
+Definition issig_IsDiagonal {R : Ring@{i}} {n : nat} {M : Matrix R n n}
   : _ <~> IsDiagonal M
   := ltac:(issig).
 
 (** A matrix is diagonal in a unique way. *)
-Global Instance ishprop_isdiagonal {R : Ring} {n : nat} (M : Matrix R n n)
+Global Instance ishprop_isdiagonal {R : Ring@{i}} {n : nat} (M : Matrix R n n)
   : IsHProp (IsDiagonal M).
 Proof.
   snrapply hprop_allpath.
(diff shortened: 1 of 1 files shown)


jdchristensen:
I see that you marked this a draft again, but I had noticed a few simplifications to the proofs, so I have pushed them.  There is still a build problem having to do with the other PRs this depends on.

theories/Algebra/Rings/Matrix.v
```diff
@@ -454,3 +454,205 @@ Proof.
...
+Proof.
+  unfold IsLowerTriangular.
+  rewrite matrix_transpose_negate.
+  exact _.
+Defined.
+
+(** The product of two upper triangular matrices is upper triangular. *)
+Global Instance upper_triangular_mult {R : Ring} {n : nat} (M N : Matrix R n n)
+  {H1 : IsUpperTriangular M} {H2 : IsUpperTriangular N}
+  : IsUpperTriangular (matrix_mult M N).
```
jdchristensen, line 542: In this and the next, I reduced from three cases to two cases.

theories/Algebra/Rings/Matrix.v
```diff
@@ -454,3 +454,205 @@ Proof.
...
+  1: by rewrite rng_mult_zero_l.
+  assumption.
+Defined.
+
+(** The product of two lower triangular matrices is lower triangular. *)
+Global Instance lower_triangular_mult {R : Ring} {n : nat} (M N : Matrix R n n)
+  {H1 : IsLowerTriangular M} {H2 : IsLowerTriangular N}
+  : IsLowerTriangular (matrix_mult M N).
+Proof.
+  (** Annoyingly, we cannot derive this from [upper_trianglar_mult] and [matrix_transpose_mult] because the latter requires a commutative ring. Therefore we have to repeat the proof with a few modifications. *)
```
jdchristensen, line 565: `matrix_transpose_mult` should also work over a non-commutative ring, using the opposite ring, so you should be able to deduce this result from the previous result.

jdchristensen:
I repushed my changes since they got lost in the force push you did.

theories/Algebra/Rings/Matrix.v
jdchristensen: I've address this in my most recent push to this PR.

theories/Algebra/Rings/Matrix.v
```diff
@@ -616,3 +637,191 @@ Proof.
...
+  rewrite matrix_transpose_diag.
+  apply upper_triangular_diag.
+Defined.
+
+(** Upper triangular matrices are a subring of the ring of matrices. *)
+Definition upper_triangular_matrix_ring (R : Ring) (n : nat)
+  : Subring (matrix_ring R n).
+Proof.
+  nrapply (Build_Subring' (fun M : matrix_ring R n => IsUpperTriangular M));
+    exact _.
```
jdchristensen, line 817: This lines are a bit slow.

theories/Algebra/Rings/Matrix.v
```diff
@@ -616,3 +637,191 @@ Proof.
...
+Defined.
+
+(** Lower triangular matrices are a subring of the ring of matrices. *)
+Definition lower_triangular_matrix_ring (R : Ring) (n : nat)
+  : Subring (matrix_ring R n).
+Proof.
+  nrapply (Build_Subring' (fun M : matrix_ring R n => IsLowerTriangular M));
+    exact _.
```
jdchristensen, line 825: This lines too.

jdchristensen (approved):
LGTM once the two sets of slow lines are improved.

Alizter:
@jdchristensen This commit speeds up those lines a bit, I could probably speed it up further if I split it up like I did for the diagonal. WDYT?

theories/Algebra/Rings/Matrix.v
Alizter: Nice, thanks!

jdchristensen:
> This commit speeds up those lines a bit, I could probably speed it up further if I split it up like I did for the diagonal. WDYT?

I think they are still too slow, so should be split up.  ...  And I checked `matrix_diag_ring` and it's actually still a bit slow, but you don't notice it because the time is spread over multiple tactics.  I wonder if part of the slowness is because some of the lemmas about diagonal matrices use almost 600 universe variables?  And the results about triangular matrices involve ~2500 universe variables!

In general, all of this should be really fast, so it would be good to check for slow lines and fix them, beginning with a fix to the universe variables.

Alizter:
>~2500 universe variables!

:exclamation:

I will see about fixing this.

jdchristensen:
A quick look shows that the problems occur already in the List/Theory.v file, e.g. with `nth'_seq'`.  I wonder if we should be using minimization to set in List/Core.v, etc.  E.g. does `seq` need to land in `list@{i} nat`, or would `list@{Set} nat` be good enough?  Do we want to have `Local Set Polymorphic Inductive Cumulativity.` in most of these files?

Maybe the thing to do is to make a separate PR with cleanups to the universe variables, and then rebase this one on that one to see if the speed issue goes away.

Alizter:
@jdchristensen I've been tweaking it from where you said since yesterday but the work grows a lot. I need to finish and split up the work. You're correct that seq and seq' should land in Set.

jdchristensen:
@Alizter I'm curious how things look here after rebasing, now that some universe variables are fixed in other places.  Can you rebase?  Or I can try hitting the "Update branch" button.

Alizter:
I also opened this: https://github.com/HoTT/Coq-HoTT/pull/1988

jdchristensen:
Already the fixes to lists have made a huge improvement here, reducing the number of universe variables from thousands to dozens (e.g. 33 for `lower_triangular_matrix_ring`).  But the proof of `lower_triangular_matrix_ring` is still a bit slow, unfortunately.

Alizter:
After the universe modifications to vector and matrix, we can revisit the speed in this proof.

Alizter:
I want to do another cleanup of universe levels but this time in Matrix. The fact that matrix rings are taking 30 universes seems to be the crux of the issue.

Alizter:
@jdchristensen I've annotated the file so that the subrings for diagonal and upper/lower triangular matrices have the correct number of universes. This can probably be improved so that annotations can be removed but I don't see how to avoid `Ring` being reduced to `Set` without unsetting minimization to set.

jdchristensen:
Yes, it's odd that sometimes `Ring` is interpreted as `Ring@{i}` and sometimes as `Ring@{Set}`.  Your new annotations look fine.  With the various improvements to universes, the proof of `matrix_diag_ring` became fast.  I still thought the triangular_matrix_ring results were a touch slow, so I expanded them in my last commit, and now they are fast.

So this LGTM!

PR #1984, opened 2024-06-05
fix universes in list library

Alizter:
We explicitly annotate all the lemmas in the list library in order to make sure there is no universe blow up. The universe blow up stems from poor implementation of various tactics with respect to universe polymorphism.

I will get onto cleaning up universes in the linear algebra library too, but that is also affected by a more critical universe blow up which will have to be considered separately.

### DIFF
--- a/test/Spaces/List.v
+++ b/test/Spaces/List.v
@@ -38,11 +38,11 @@ Succeed Check fold_left_app@{_ _}.
 Succeed Check fold_right_app@{_ _}.
 Succeed Check length_list_map@{_ _}.
 Succeed Check inlist_map@{_ _}.
-Succeed Check inlist_map'@{_ _}.
+Succeed Check inlist_map'@{_ _ _}.
 Succeed Check list_map_id@{_}.
 Succeed Check list_map_compose@{_ _ _}.
 Succeed Check length_list_map2@{_ _ _}.
-Succeed Check inlist_map2@{_ _ _}.
+Succeed Check inlist_map2@{_ _ _ _}.
 Succeed Check list_map2_repeat_l@{_ _ _}.
 Succeed Check list_map2_repeat_r@{_ _ _}.
 Succeed Check length_reverse_acc@{_}.
@@ -101,11 +101,11 @@ Succeed Check length_repeat@{_}.
 Succeed Check inlist_repeat@{_}.
 Succeed Check for_all_inlist@{_ _}.
 Succeed Check inlist_for_all@{_ _}.
-Succeed Check for_all_list_map@{_ _ _ _ _}.
+Succeed Check for_all_list_map@{_ _ _ _}.
 Succeed Check for_all_list_map'@{_ _ _}.
-Succeed Check for_all_list_map2@{_ _ _ _ _ _ _}.
-Succeed Check for_all_list_map2'@{_ _ _ _ _ _ _}.
-Succeed Check fold_left_preserves@{_ _ _ _ _}.
+Succeed Check for_all_list_map2@{_ _ _ _ _ _}.
+Succeed Check for_all_list_map2'@{_ _ _ _ _ _}.
+Succeed Check fold_left_preserves@{_ _ _ _}.
 Succeed Check istrunc_for_all@{_ _}.
 Succeed Check istrunc_for_all'@{_ _}.
 Succeed Check for_all_repeat@{_ _}.
--- a/test/Spaces/List.v
+++ b/test/Spaces/List.v
@@ -1,0 +1,114 @@
+From HoTT Require Import Basics.
+From HoTT.Spaces.List Require Import Core Theory.
+
+(** Here we check the universe levels of the lemmas from List.Core *)
+Succeed Check list@{_}.
+Succeed Check nil@{_}.
+Succeed Check cons@{_}.
+Succeed Check list_rect@{_ _}.
+Succeed Check list_ind@{_ _}.
+Succeed Check list_rec@{_ _}.
+Succeed Check length@{_}.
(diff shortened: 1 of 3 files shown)


theories/Spaces/List/Theory.v
```diff
@@ -211,62 +226,10 @@ Proof.
     exact (y'; z'; (q, (inr r, inr s))).
 Defined.
 
-(** If an operation given to [list_map2] is heteroassociative, then [list_map2] is also heteroassociative with the caveat that the lists must be the same length. *)
-Definition heteroassociative_list_map2 {A B C AB BC ABC}
```
Alizter, line 215: I've also removed these as they are useless and I didn't feel like adding universe variable to them.

Alizter:
It seems that this is not ready. I need to work out a way to keep it compatible with 8.18. It seems that inference improved since 8.18 meaning that I need to work in 8.18 and be more explicit to be fully compatible.

jdchristensen:
Ok, tag me when this is ready to review.  BTW, I think that in many case, Coq does the right thing with universe variables by default, and I prefer to not annotate things in those cases, as it makes the statements harder to read.

Alizter:
@jdchristensen I've tried to be as conservative as possible with Coq 8.18 however it ends up with me needing to annotate the term beyond recognition. I'm wondering if it might be a good time to bump to 8.19 as the universe inference in 8.18 is just too poor to be usable.

An example of a problem that I am encountering. In 8.18, all the lemmas about paths with `nth` need to live in the path type of the option type. However Coq introduced a new universe variable because of the option type, which ends up with me having to write `paths@{i}`.

SkySkimmer:
What changed about universes in 8.19? There's kernel sort polymorphism but that shouldn't matter for hott.

jdchristensen:
> I'm wondering if it might be a good time to bump to 8.19

Since 8.19 was released more than four months ago, I'm fine with this.

Alizter:
@jdchristensen I've relaxed some of the annotations in Theory.v and kept only the ones where the annotations matter. I think this is ready for another round of review.

Let me know if you still think there are too many universe annotations in some places. I've had a go, but perhaps there are areas I can improve.

test/Spaces/List.v
```diff
@@ -0,0 +1,114 @@
+From HoTT Require Import Basics.
+From HoTT.Spaces.List Require Import Core Theory.
+
+(** Here we check the universe levels of the lemmas from List.Core *)
```
jdchristensen, line 4:
```suggestion
(** Here we check the number of universe variables for the definitions from List.Core *)
```

test/Spaces/List.v
```diff
@@ -0,0 +1,114 @@
...
+Succeed Check last@{_}.
+Succeed Check nth@{_}.
+Succeed Check remove_last@{_}.
+Succeed Check seq_rev@{}.
+Succeed Check seq@{}.
+Succeed Check repeat@{_}.
+Succeed Check InList@{_}.
+Succeed Check for_all@{_ _}.
+
+(** Here we check the universe levels of the lemmas from List.Theory *)
```
jdchristensen, line 30:
```suggestion
(** Here we check the number of universe variables for the definitions from List.Theory *)
```

theories/Spaces/List/Core.v
```diff
@@ -37,7 +39,7 @@ Notation "[ x , y , .. , z ]" := (x :: (y :: .. (z :: nil) ..)) : list_scope.
 (** Notice that the definition of a list looks very similar to the definition of [nat]. It is as if each [S] constructor from [nat] has an element of [A] attached to it. We can discard this extra element and get a list invariant that we call [length]. *)
 
 (** The length (number of elements) of a list. *)
-Fixpoint length {A} (l : list A) :=
+Fixpoint length@{i|} {A : Type@{i}} (l : list@{i} A) :=
```
jdchristensen, line 42: I checked the first two places that new annotations remain, `length` and `app`, and neither seems necessary to me.  How did you decide which to remove and which to keep?

jdchristensen: Oh, I see that you said that you only removed some in Theory.v.  Why not in Core.v too?

Alizter: I ran out of time yesterday, I'll try and have a go today for Core.v.

jdchristensen:
I just pushed changes to List/Theory.v.  There were several cases where there was a universe constraint `i <= j` that was not needed, but could be replaced with an upper bound universe `u` with `i <= u` and `j <= u`.  While the upper bound approach has more universe variables, it's still the correct thing to do, as the other approach introduces an unnecessary constraint.  (And I think that in the future Coq might support using `max(i,j)` in more places, which would allow us to get rid of `u`.)

I also removed lots of annotations that had no effect, as Coq 8.19 does the right thing by default.

And I simplified two proofs, near the beginning.

When trying to fix issues with too many universe variables, I always try to trace things down to the earliest result with extra variables.  After fixing that, I check whether later results still have problems.  If so, I again search for the earliest result that has issues, and fix it, always working bottom up.  I often find that I just need to annotate a small number of early results and that the later results get fixed automatically.

jdchristensen:
I forgot to fix the tests, and will do so in a minute.

jdchristensen:
I'm also in the process of using a `fix` tactic to simplify things further, so don't push any changes.

jdchristensen:
I added `simple_list_induction` which is just like `simple_induction`, and it works well.  Those upper bound universes are always free, so they tend to multiply, so using this trick to get rid of them is useful.

Alizter:
Thanks @jdchristensen LGTM!

PR #1986, opened 2024-06-10
drop extra universe in istrunc_sigma

Alizter:
This was quite difficult to fix, but I worked it out in the end.

The previous proof was using the generated polymorphic induction principle which is actually less general than what we could be. This meant that we required an extra strictly larger universe for the induction hypothesis to work.

Rewriting the proof using a fixpoint allows us to use the most general version of the "induction principle", i.e. the rules of the underlying fixpoint. This allows us to remain in the same universe, simplifying the universe requirements elsewhere too.

The main motivation for this patch was to reduce the universes in `istrunc_vector` and now we don't need a superfluous larger universe.

cc @mikeshulman who participated in the corresponding Zulip discussion: https://hott.zulipchat.com/#narrow/stream/228519-general/topic/.E2.9C.94.20Truncation.20level.20of.20sigma.20type

### DIFF
--- a/theories/Types/Sigma.v
+++ b/theories/Types/Sigma.v
@@ -613,7 +613,7 @@ Global Instance istrunc_sigma `{P : A -> Type}
 : IsTrunc n (sig P) | 100.
 Proof.
   generalize dependent A.
-  induction n; simpl; intros A P ac Pc.
+  simple_induction' n; simpl; intros A P ac Pc.
   { apply (Build_Contr _ (center A; center (P (center A)))).
     intros [a ?].
     refine (path_sigma' P (contr a) (path_contr _ _)). }
--- a/theories/PropResizing/Nat.v
+++ b/theories/PropResizing/Nat.v
@@ -375,9 +375,6 @@ Section AssumeStuff.
       apply path_N; reflexivity.
   Qed.
 
-  (** Sometimes we just need a bigger fish. *)
-  Universe large.
-
   (** A first application *)
   Definition N_neq_succ@{} (n : N) : n <> succ n.
   Proof.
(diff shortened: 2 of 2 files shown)


theories/Types/Sigma.v
```diff
@@ -608,20 +608,23 @@ Definition equiv_sig_coind
...
-    refine (path_sigma' P (contr a) (path_contr _ _)). }
-  apply istrunc_S.
+    refine (path_sigma'@{i j k} P (contr@{i} a) (path_contr@{j} _ _)). }
+  apply istrunc_S@{k}.
   intros u v.
-  refine (istrunc_isequiv_istrunc _ (path_sigma_uncurried P u v)).
+  refine (istrunc_isequiv_istrunc@{k k} _ (path_sigma_uncurried@{i j k} P u v)).
 Defined.
 
+Existing Instance istrunc_sigma | 100.
```
Alizter, line 626:
```suggestion
Global Existing Instance istrunc_sigma | 100.
```

jdchristensen:
Instead of your approach, you can just change `induction` to `simple_induction'`, and it goes through with the reduced universe variables.  No annotations required.

Alizter:
@jdchristensen I don't understand your suggestion. If I change `induction` to `simple_induction'` like you said, I end up having unbound universes. Could you elaborate a bit more?

jdchristensen:
If you completely revert your changes to Types/Sigma.v and just change `induction` to `simple_induction'`, doesn't it work?

Alizter:
> If you completely revert your changes to Types/Sigma.v and just change `induction` to `simple_induction'`, doesn't it work?

I tried that, it didn't work for me.

Alizter:
Oh yes you are correct. I had an annotation left over. Whoops. I will force push the fix.

Alizter:
Thanks @jdchristensen that is a much cleaner "fix"!

jdchristensen (approved):

PR #1987, opened 2024-06-10
fix universes in vector

Alizter:
With a little bit of help, we can get Coq to minimize the number of universes it uses for Vector.v.

### DIFF
--- a/theories/Algebra/Rings/Vector.v
+++ b/theories/Algebra/Rings/Vector.v
@@ -16,8 +16,8 @@ Local Set Polymorphic Inductive Cumulativity.
 
 (** ** Definition *)
 
-Definition Vector@{i} (A : Type@{i}) (n : nat) : Type@{i}
- := sig@{i Set} (fun l : list@{i} A => length l = n).
+Definition Vector@{i|} (A : Type@{i}) (n : nat) : Type@{i}
+  := { l : list A & length l = n }.
 
 (** *** Constructors *)
 
--- a/theories/Algebra/Rings/Vector.v
+++ b/theories/Algebra/Rings/Vector.v
@@ -7,7 +7,8 @@ Require Import abstract_algebra.
 
 Local Open Scope mc_scope.
 
-Set Universe Minimization ToSet.
+Local Set Universe Minimization ToSet.
+Local Set Polymorphic Inductive Cumulativity.
 
 (** * Vectors *)
 
@@ -16,7 +17,7 @@ Set Universe Minimization ToSet.
 (** ** Definition *)
 
 Definition Vector@{i} (A : Type@{i}) (n : nat) : Type@{i}
- := { l : list@{i} A & length l = n }.
+ := sig@{i Set} (fun l : list@{i} A => length l = n).
 
 (** *** Constructors *)
 
@@ -31,7 +32,7 @@ Defined.
 
 (** *** Projections *)
 
-Definition entry {A : Type} {n} (v : Vector A n) i {Hi : (i < n)%nat} : A
+Definition entry {A : Type} {n : nat} (v : Vector A n) i {Hi : (i < n)%nat} : A
   := nth' (pr1 v) i ((pr2 v)^ # Hi).
 
 (** *** Basic properties *) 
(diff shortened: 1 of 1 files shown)


jdchristensen:
@Alizter Is this a duplicate of #1988?

Alizter:
@jdchristensen Seems GitHub created two PRs due to my flaky internet connection. I've closed the other one.

theories/Algebra/Rings/Vector.v
```diff
@@ -16,7 +17,7 @@ Set Universe Minimization ToSet.
 (** ** Definition *)
 
 Definition Vector@{i} (A : Type@{i}) (n : nat) : Type@{i}
- := { l : list@{i} A & length l = n }.
+ := sig@{i Set} (fun l : list@{i} A => length l = n).
```
jdchristensen, line 20: This seems to do the same thing without using the `sig` constructor.  Revert?

jdchristensen (approved):
LGTM, modulo the one comment.

theories/Algebra/Rings/Vector.v
Alizter: I get a floating algebraic universe when I do that. If you do it, you'll see that other lemmas start assuming an extra algebraic universe which you can see with Show Universes.

Alizter: So this change makes the universe this sigma type lives in equal to `i` rather than `max(i, u)` for some floating `u`. Here `u` should be `Set` no idea why Coq doesn't see that.

jdchristensen: I don't get a floating universe with this:
```coq
Definition Vector@{i} (A : Type@{i}) (n : nat) : Type@{i}
  := { l : list A & length l = n }.

(* Vector@{i}` : Type -> nat -> Type *)
```

jdchristensen: To clarify:
```coq
Definition Vector@{i} (A : Type@{i}) (n : nat) : Type@{i}
 := { l : list A & length l = n }.
Unset Printing Notations.
Set Printing Universes.
Print Vector.
```
produces
```
Vector@{i} =
fun (A : Type@{i}) (n : nat) =>
sig@{i Set} (fun l : list@{i} A => paths@{Set} (length@{i} l) n)
     : forall (_ : Type@{i}) (_ : nat), Type@{i}
(* i |= Set <= i *)
```
with the same `sig@{i Set}` that the PR does manually.

jdchristensen: I'm not sure what you mean.  With my version of `Vector`, I get
```
Build_Vector@{u} =
fun (A : Type@{u}) (n : nat) (f : forall (i : nat) (_ : Core.lt i n), A) =>
{|
  proj1 :=
    list_map@{Set u}
      (fun pat : sig@{Set Set} (fun i : nat => Core.lt i n) =>
       let x := pat in let i := proj1@{Set Set} x in let Hi := proj2@{Set Set} x in f i Hi)
      (seq' n);
  proj2 :=
    concat@{Set}
      (length_list_map@{Set u}
         (fun pat : sig@{Set Set} (fun i : nat => Core.lt i n) =>
          let x := pat in
          let i := proj1@{Set Set} x in let Hi := proj2@{Set Set} x in f i Hi) 
         (seq' n)) (length_seq' n)
|}
     : forall (A : Type@{u}) (n : nat) (_ : forall (i : nat) (_ : Core.lt i n), A),
       Vector@{u} A n
(* u |=  *)
```
I'm using Coq 8.19.1.

jdchristensen: Please show your test code and the output from Coq.

jdchristensen: I looked into this further, and found that with the version of `Vector` that I posted, `path_vector` had an issue.  Depending on exactly how you write `Vector`, you might end up with it landing in `max(Set,i)`, or you might end up with it landing in `i` but with a constraint `Set <= i`.  Either way gives problems later.  But I just pushed a version that gets rid of both of those issues, while still using the `{ ... }` notation for the sigma type.

Why does Coq produce constraints like `Set <= i` that are always true?  And why doesn't it simplify `max(Set,i)`to `i`?

Alizter: I can't reproduce. I guess I must have been tired yesterday and made a mistake somewhere while testing.

PR #1989, opened 2024-06-10
nix: add coq master to nix flake

Alizter:
This lets you setup coq master with nix. To use it you do `nix develop --impure`. (`--impure` is required because `master` is a rolling version and cannot have a hash so it is not a "pure" evaluation according to nix.)


Alizter:
Merging this as this Nix job has passed so this is fine as far as the CI is concerned.

PR #1991, opened 2024-06-11
1-Cat instance for MatrixCat

CelestialCry:
Defines a type MatrixCat R that consists of the natural numbers. We turn MatrixCat into a 1-category by considering the homsets (m,n) = M_{m╳n}(R) of m╳n-dimensional matrices.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -4,6 +4,7 @@ Require Import Spaces.List.Core Spaces.List.Theory Spaces.List.Paths.
 Require Import Algebra.Rings.Ring Algebra.Rings.Module Algebra.Rings.CRing
   Algebra.Rings.KroneckerDelta Algebra.Rings.Vector.
 Require Import abstract_algebra.
+Require Import WildCat.
 
 Set Universe Minimization ToSet.
 
@@ -639,3 +640,95 @@ Proof.
   intros k l Hk Hl.
   by rewrite 4 entry_Build_Matrix.
 Defined.
+
+Section MatrixCatIs1Cat.
+
+  (** The wildcategory MatrixCat R of R-valued matrices. 
+  This category has natural numbers as objects, and m x n - dimensional matrices between m and n. *)
+  Definition MatrixCat (R : Ring) := nat.
+
+  Global Instance isgraph_matrixcat {R} : IsGraph (MatrixCat R).
+  Proof.
+    snrapply Build_IsGraph.
+    intros m n.
+    exact (Matrix R m n).
+  Defined.
+
+  Global Instance is01cat_matrixcat {R} : Is01Cat (MatrixCat R).
+  Proof.
+    snrapply Build_Is01Cat.
+    - intros a. 
+      exact (identity_matrix R a).
+    - intros a b c f g.
+      exact (matrix_mult g f).
+  Defined.
+
+  Global Instance is2graph_matrixcat {R} : Is2Graph (MatrixCat R).
+  Proof.
+    unfold Is2Graph.
+    intros M N. 
+    snrapply Build_IsGraph.
+    intros A B.
+    exact (A = B).
+  Defined.
+
+  Global Instance isgraph_morphismmatrix {R} (m n : MatrixCat R) : IsGraph (m $-> n).
+  Proof.
+    snrapply Build_IsGraph.
+    intros A B.
+    exact (A = B).
+  Defined.
+
+  Global Instance is01cat_morphismmatrix {R} (m n : MatrixCat R) : Is01Cat (m $-> n).
+  Proof.
+    snrapply Build_Is01Cat.
+    - intros a.
+      exact (idpath a).
+    - intros a b c A B.
+      exact (B @ A).
+  Defined.
+
+  Global Instance is0gpd_morphismmatrix {R} (m n : MatrixCat R) : Is0Gpd (m $-> n).
+  Proof.
+    snrapply Build_Is0Gpd.
+    intros a b A.
+    destruct A.
+    exact (idpath _).
+  Defined.
+
+  Global Instance is0functor_postcomp_morphismmatrix {R} {m n p : MatrixCat R} (h : n $-> p)
+    : Is0Functor (@cat_postcomp (MatrixCat R) _ _ m n p h).
+  Proof.
+    snrapply Build_Is0Functor.
+    intros a b f.
+    destruct f.
+    exact (idpath _).
+  Defined.
+
+  Global Instance is0functor_precomp_morphismmatrix {R} {m n p : MatrixCat R} (h : m $-> n)
+    : Is0Functor (@cat_precomp (MatrixCat R) _ _ m n p h).
+  Proof.
+    snrapply Build_Is0Functor.
+    intros a b f.
+    destruct f.
+    exact (idpath _).
+  Defined.
+  (** I don't know if we can do something smart with these 3 functions, as they are constructed identically. *)
+
+  (** MatrixCat R forms a strong 1-category. *)
+  Global Instance is1catstrong_matrixcat {R} : Is1Cat_Strong (MatrixCat R).
+  Proof.
+    snrapply Build_Is1Cat_Strong.
+    - apply is01cat_morphismmatrix.
+    - apply is0gpd_morphismmatrix.
+    - intros m n p h. exact (is0functor_postcomp_morphismmatrix h).
+    - intros m n p h. exact (is0functor_precomp_morphismmatrix h).
+    - apply (associative_matrix_mult R).
+    - intros m n p q A B C. apply inverse. apply (associative_matrix_mult R).
+    - apply right_identity_matrix_mult.
+    - apply left_identity_matrix_mult.
+  Defined.
+
+(** TODO: Define HasEquivs for MatrixCat.  *)
+
+End MatrixCatIs1Cat.


theories/Algebra/Rings/Matrix.v
```diff
@@ -4,6 +4,7 @@ Require Import Spaces.List.Core Spaces.List.Theory Spaces.List.Paths.
 Require Import Algebra.Rings.Ring Algebra.Rings.Module Algebra.Rings.CRing
   Algebra.Rings.KroneckerDelta Algebra.Rings.Vector.
 Require Import abstract_algebra.
+Require Import WildCat.
```
Alizter, line 8:
```suggestion
Require Import WildCat.Core.
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
   intros k l Hk Hl.
   by rewrite 4 entry_Build_Matrix.
 Defined.
+
+Section MatrixCatIs1Cat.
+
```
Alizter, line 645:
```suggestion
Section MatrixCat.
```

This is a shorter name.

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+  (** The wildcategory MatrixCat R of R-valued matrices. 
+  This category has natural numbers as objects, and m x n - dimensional matrices between m and n. *)
+  Definition MatrixCat (R : Ring) := nat.
+
+  Global Instance isgraph_matrixcat {R} : IsGraph (MatrixCat R).
+  Proof.
+    snrapply Build_IsGraph.
+    intros m n.
+    exact (Matrix R m n).
+  Defined.
```
Alizter, line 655:
```suggestion
  Global Instance isgraph_matrixcat {R : Ring} : IsGraph (MatrixCat R)
    := {| Hom := Matrix R |}.
```

You can write this in a more compact way like this. I would also suggest that we write `{R : Ring}` as it might not always be inferrable in contrast to say natural numbers. There are a few results in this file that require a commutative ring, but we still call that `R` for example.

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+  Global Instance is2graph_matrixcat {R} : Is2Graph (MatrixCat R).
+  Proof.
+    unfold Is2Graph.
+    intros M N. 
+    snrapply Build_IsGraph.
+    intros A B.
+    exact (A = B).
+  Defined.
+
+  Global Instance isgraph_morphismmatrix {R} (m n : MatrixCat R) : IsGraph (m $-> n).
```
Alizter, line 675: This is not needed as this is what Is2Graph gives.

Alizter (commented):
I've got some cleanups for this PR. I can continue to review through GitHub or I can push my changes here. Which one would you prefer?

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
   intros k l Hk Hl.
   by rewrite 4 entry_Build_Matrix.
 Defined.
+
+Section MatrixCatIs1Cat.
+
+  (** The wildcategory MatrixCat R of R-valued matrices. 
+  This category has natural numbers as objects, and m x n - dimensional matrices between m and n. *)
```
jdchristensen, line 647: We put square brackets around things that parse as Coq syntax, as the documentation generator uses this to know how to format them.  We also don't put newlines in comments generally.
```suggestion
  (** The wild category [MatrixCat R] of [R]-valued matrices. This category has natural numbers as objects and m x n matrices as the arrows between [m] and [n]. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+    snrapply Build_IsGraph.
+    intros m n.
+    exact (Matrix R m n).
+  Defined.
+
+  Global Instance is01cat_matrixcat {R} : Is01Cat (MatrixCat R).
+  Proof.
+    snrapply Build_Is01Cat.
+    - intros a. 
+      exact (identity_matrix R a).
```
jdchristensen, line 661: Probably these two lines can be simplfied:
```suggestion
    - exact (identity_matrix R).
```
It's the difference between the terms `f` and `(fun x => f x)`.  We try to store the smaller term.  This trick doesn't work for the `matrix_mult` term below, because the argument order changes.

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+    - intros a b c f g.
+      exact (matrix_mult g f).
+  Defined.
+
+  Global Instance is2graph_matrixcat {R} : Is2Graph (MatrixCat R).
```
jdchristensen, line 666: Do we have a helper function for when we want to just use paths for the higher structure?  I couldn't find one, but maybe we should add one.  If we do, then a lot of lemmas below should also follow automatically from paths forming a groupoid.

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+  Proof.
+    unfold Is2Graph.
+    intros M N.
```
jdchristensen, line 669:
```suggestion
    intros m n. 
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+      exact (matrix_mult g f).
+  Defined.
+
+  Global Instance is2graph_matrixcat {R} : Is2Graph (MatrixCat R).
+  Proof.
+    unfold Is2Graph.
+    intros M N. 
+    snrapply Build_IsGraph.
+    intros A B.
+    exact (A = B).
```
jdchristensen, line 672:
```suggestion
    exact paths.
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+  Defined.
+
+  Global Instance is01cat_morphismmatrix {R} (m n : MatrixCat R) : Is01Cat (m $-> n).
+  Proof.
+    snrapply Build_Is01Cat.
+    - intros a.
+      exact (idpath a).
```
jdchristensen, line 686:
```suggestion
    - exact (@idpath _).
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -639,3 +640,95 @@ Proof.
...
+    - intros a b c A B.
+      exact (B @ A).
+  Defined.
+
+  Global Instance is0gpd_morphismmatrix {R} (m n : MatrixCat R) : Is0Gpd (m $-> n).
+  Proof.
+    snrapply Build_Is0Gpd.
+    intros a b A.
+    destruct A.
+    exact (idpath _).
```
jdchristensen, line 696: Matrices should use letters like `M` and `N`, and paths should use letters like `p`.  And we can reuse the existing proof of inverses of paths.
```suggestion
    intros M N.
    exact inverse.
```

jdchristensen: There are three more results below, but I'll let @ThomatoTomato apply some similar simplifications to them.

theories/Algebra/Rings/Matrix.v
Alizter: I was thinking about something similar.

theories/Algebra/Rings/Matrix.v
Alizter: Better to just use WildCat.paths instances here actually.

theories/Algebra/Rings/Matrix.v
Alizter: I've created https://github.com/HoTT/Coq-HoTT/pull/1994 so you can do just that. You shouldn't need any of the instances after Is2Graph and you can write `exact _` for the instances `Build_Is1Cat` asks for up to the associativity which you will prove with the lemma anyway.

theories/Algebra/Rings/Matrix.v
ThomatoTomato: Does this notation has a name? Can I read up on it in the documentation, or is this a feature of this library?

theories/Algebra/Rings/Matrix.v
ThomatoTomato: I'm a little confused. Doesnt Is2Graph depend on IsGraph? Will the little accent ` make the other dependencies into a goal if they aren't instantiated?

jdchristensen: @ThomatoTomato I'm confused about what you are confused about.  :smile:  Is2Graph depends on IsGraph, but you already proved isgraph_matrixcat.  Then you proved is2graph_matrixcat.  And now isgraph_morphismmatrix is proving the same thing as is2graph_matrixcat.  Both state that the type of 1-morphisms is a graph.  So the second one can be deleted.

theories/Algebra/Rings/Matrix.v
jdchristensen: @ThomatoTomato It's documented here:  https://coq.inria.fr/doc/V8.19.2/refman/language/core/records.html#constructing-records
(Classes behave just like Records.)

theories/Algebra/Rings/Matrix.v
ThomatoTomato: Ah, I remember now that I got very confused about this when I was writing up the code. Everything makes sense now, ty.

theories/Algebra/Rings/Matrix.v
```diff
@@ -641,85 +644,67 @@ Proof.
...
   Proof.
     snrapply Build_Is1Cat_Strong.
-    - apply is01cat_morphismmatrix.
-    - apply is0gpd_morphismmatrix.
-    - intros m n p h. exact (is0functor_postcomp_morphismmatrix h).
-    - intros m n p h. exact (is0functor_precomp_morphismmatrix h).
+    - apply is01cat_matrixcat_hom.
+    - apply is0gpd_matrixcat_hom.
+    - snrapply @is0functor_postcomp_matrixcat_hom.
+    - snrapply @is0functor_precomp_matrixcat_hom.
```
jdchristensen, line 701: Since we've defined instances, we can use them.  But trying this shows that there is something wrong with one of the instances.  I'm not sure what is going on.
```suggestion
    1,2,4: exact _.
    (* I'm not sure why typeclass search doesn't find the next one.  Something must be slightly broken. *)
    - snrapply @is0functor_postcomp_matrixcat_hom.
```

theories/Algebra/Rings/Matrix.v
ThomatoTomato: I've been looking at WildCat.Paths for some time now, lol 😅. I will take a look at the new things now ig.

theories/Algebra/Rings/Matrix.v
ThomatoTomato: I'm still seeing the issue that it can't find #3 :/

theories/Algebra/Rings/Matrix.v
jdchristensen: @Alizter I think most of the Local Instances involving the _paths wild category structures should be Global Instances, except isgraph{,2,3}_paths, which need to be chosen by the user.  I have tested this and it gets rid of lots of boiler plate.  If you think this makes sense, I'll push, probably tomorrow.

jdchristensen: Actually, I'm ready to push now.  @ThomatoTomato is it ok if I push to this PR, or do you have any local changes that would get in the way?

ThomatoTomato: Local branch is up to date! Go ahead whenever

theories/Algebra/Rings/Matrix.v
```diff
@@ -641,85 +644,67 @@ Proof.
...
     := {| Hom := Matrix R |}.
 
-  Global Instance is01cat_matrixcat {R} : Is01Cat (MatrixCat R).
+  Global Instance is01cat_matrixcat {R : Ring} : Is01Cat (MatrixCat R).
   Proof.
     snrapply Build_Is01Cat.
-    - exact (identity_matrix R).
-    - intros a b c f g.
-      exact (matrix_mult g f).
+    - exact (@identity_matrix R).
```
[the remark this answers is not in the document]
jdchristensen:
```suggestion
    - exact (identity_matrix R).
```

theories/Algebra/Rings/Matrix.v
jdchristensen: Actually, I need more time, so I'll make a separate PR.  I think this PR is ready to merge, unless @Alizter has any other comments.

jdchristensen:
Merging.  Thanks @ThomatoTomato for your contribution!  I'll next create a PR about WildCat.Paths that allows us to simplify things here a bit too.

PR #1992, opened 2024-06-11
Z-module structure on abelian groups

ThomatoTomato:
We want to make lm_carrier into an equivalence of categories.

### Pipeline:
- [x] Turn lm_carrier : LeftModule R -> AbGroup into a functor
- [ ] Define a canonical Z-module structure
- [ ] Prove that lm_carrier is an equivalence
- [ ] Define a coercion from AbGroup into LeftModule cring_Z and write tests

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -1,5 +1,5 @@
 Require Import Basics Types.
-Require Import Spaces.Nat.Core.
+Require Import Spaces.Nat.Core Spaces.Int.
 Require Export Classes.interfaces.canonical_names (Zero, zero, Plus).
 Require Export Classes.interfaces.abstract_algebra (IsAbGroup(..), abgroup_group, abgroup_commutative).
 Require Export Algebra.Groups.Group.
@@ -228,26 +228,39 @@ Proof.
     1-2: exact negate_involutive.
 Defined.
 
-(** Multiplication by [n : nat] defines an endomorphism of any abelian group [A]. *)
-Definition ab_mul_nat {A : AbGroup} (n : nat) : GroupHomomorphism A A.
+(** Multiplication by [n : Int] defines an endomorphism of any abelian group [A]. *)
+Definition ab_mul {A : AbGroup} (n : Int) : GroupHomomorphism A A.
 Proof.
   snrapply Build_GroupHomomorphism.
   1: exact (fun a => grp_pow a n).
   intros a b.
   induction n; cbn.
-  1: exact (grp_unit_l _)^.
-  refine (_ @ associativity _ _ _).
-  refine (_ @ ap _ (associativity _ _ _)^).
-  rewrite (commutativity (grp_pow a n) b).
-  refine (_ @ ap _ (associativity _ _ _)).
-  refine (_ @ (associativity _ _ _)^).
-  apply grp_cancelL.
-  assumption.
+  - exact (grp_unit_l _)^.
+  - destruct n.
+    + reflexivity.
+    + rhs_V srapply (associativity a).
+      rhs srapply (ap _ (associativity _ b _)).
+      rewrite (commutativity (nat_iter _ _ _) b).
+      rhs_V srapply (ap _ (associativity b _ _)).
+      rhs srapply (associativity a).
+      apply grp_cancelL.
+      exact IHn.
+  - destruct n.
+    + rewrite (commutativity (-a)).
+      exact (grp_inv_op a b).
+    + rhs_V srapply (associativity (-a)).
+      rhs srapply (ap _ (associativity _ (-b) _)).
+      rewrite (commutativity (nat_iter _ _ _) (-b)).
+      rhs_V srapply (ap _ (associativity (-b) _ _)).
+      rhs srapply (associativity (-a)).
+      rewrite (commutativity (-a) (-b)), <- (grp_inv_op a b).
+      apply grp_cancelL.
+      exact IHn.
 Defined.
 
-Definition ab_mul_nat_homo {A B : AbGroup}
-  (f : GroupHomomorphism A B) (n : nat)
-  : f o ab_mul_nat n == ab_mul_nat n o f
+Definition ab_mul_homo {A B : AbGroup}
+  (f : GroupHomomorphism A B) (n : Int)
+  : f o ab_mul n == ab_mul n o f
   := grp_pow_homo f n.
 
 (** The image of an inclusion is a normal subgroup. *)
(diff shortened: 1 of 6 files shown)


theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+  Defined.
+
+  Global Instance lm_carrieris1fun {R} : Is1Functor (lm_carrier R).
+  Proof.
+    snrapply Build_Is1Functor.
+    - intros a b f g e. assumption.
+    - reflexivity.
+    - reflexivity.
+  Defined.
+  (* I think the above should be moved to Module.v, as it is not specifically a property of the integers. *)
```
Alizter, line 277: Yes, I think these should be moved there. While you are doing that I would suggest renaming them `is0functor_lm_carrier`. The basic naming scheme is `whatyouwant_whatyouhave`.

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+
+  (** Every abelian group admits a canonical left Z-module structure. *)
+  Definition can_Z : AbGroup -> (LeftModule cring_Z).
+  Proof.
+    intros A. snrapply Build_LeftModule.
+    - assumption.
+    - snrapply (Build_IsLeftModule _).
+      + intros n a. exact (ab_mul n a).
+      + unfold LeftHeteroDistribute. intros n. exact preserves_sg_op.
```
Alizter, line 286: This is `grp_homo_op`. We define `ab_mul` as a map from Z to group endomorphisms. This means that the left dist property for modules is just the group homomorphism preserving the operation.

Alizter:
Actually it seems we are missing a lot here which is a shame. I think the best way forward would be to try to generalise what I've called `cring_catamorphism` which is integer multiplication with 1 to an arbitrary ring element. That way it would generalise the ab_mul function and the lemmas we prove about cring_catamorphism should generalise past multiplication with 1. That should encompass most of what we want to prove here.

The issue with this however is that I would like to switch from BinInt to Int soon, which should simplify the proofs here. So maybe we should hold off on this for the time being until we clean that up.

Thank you for taking a look at this however.

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
     exact IHp.
 Defined.
+
+Section Lm_carrierIsEquiv.
+
+  (** lm_carrier is a 1-functor (LeftModule R) ->  AbGroup. *)
+  Global Instance lm_carrieris0fun {R} : Is0Functor (lm_carrier R).
+  Proof.
+    snrapply Build_Is0Functor.
+    intros a b f.
```
jdchristensen, line 265: It's important to choose suggestive variable names, since the reader can't tell what the types are.
```suggestion
    intros M N f.
```

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+    destruct f.
+    exact lm_homo_map.
```
jdchristensen, line 267: `lm_homo_map` is an accessor function that can be applied to `f` without destructing it.
```suggestion
    exact (lm_homo_map _ _ f).
```

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+  Defined.
+
+  Global Instance lm_carrieris1fun {R} : Is1Functor (lm_carrier R).
+  Proof.
+    snrapply Build_Is1Functor.
+    - intros a b f g e. assumption.
```
jdchristensen, line 273: Since it's so easy to be explicit about which assumption is being used, it's better to spell it out:
```suggestion
    - intros a b f g e. exact e.
```
(Also, the variable names could be better.)

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
     apply ap.
     exact IHp.
 Defined.
+
+Section Lm_carrierIsEquiv.
+
+  (** lm_carrier is a 1-functor (LeftModule R) ->  AbGroup. *)
```
jdchristensen, line 261: Comment style as I mentioned in the other PR.  I'll let you check all comments.

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+    - reflexivity.
+    - reflexivity.
+  Defined.
+  (* I think the above should be moved to Module.v, as it is not specifically a property of the integers. *)
+
+  (** Every abelian group admits a canonical left Z-module structure. *)
+  Definition can_Z : AbGroup -> (LeftModule cring_Z).
+  Proof.
+    intros A. snrapply Build_LeftModule.
+    - assumption.
```
jdchristensen, line 283:
```suggestion
    - exact A.
```

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+    - snrapply (Build_IsLeftModule _).
```
jdchristensen, line 284:
```suggestion
    - snrapply Build_IsLeftModule.
```

theories/Algebra/Rings/Z.v
```diff
@@ -254,3 +255,38 @@ Proof.
...
+      + intros n a. exact (ab_mul n a).
+      + unfold LeftHeteroDistribute. intros n. exact preserves_sg_op.
+      + unfold RightHeteroDistribute. intros m n a. destruct m, n; simpl.
+        -- 
+        (* This might be the wrong way to do this. On this path I need to prove that grp_pow respects addition of natural numbers. *)
+  Admitted.
+
+End Lm_carrierIsEquiv.
```
jdchristensen, line 292: Make sure to end files with a newline.

jdchristensen:
> I think the best way forward would be to try to generalise what I've called `cring_catamorphism` which is integer multiplication with 1 to an arbitrary ring element.

In fact, as far as I can see, `cring_catamorphism` doesn't use anything about `R` being a ring except that it has a chosen element `1`.  Once you generalize it to an arbitrary ring element, then `R` should only be assumed to be an abelian group, and the material on `cring_catamorphism` should be renamed and moved to the `AbGroups` folder.  (In fact, it probably doesn't even use the the group is abelian, so could be moved to `Groups`.)

jdchristensen:
We should probably call this `grp_pow_int` and put it and its properties in Groups.v, right after `grp_pow`.  (Unless, for dependency reasons, we don't want Groups.v to depend on the definition of the integers?)  And I agree that we should use the newer definition of the integers for this.  Since the newer definition is in the library, this could be done now, right?

Alizter:
@jdchristensen Yes, we could do this immediately. I can work on that today if you like.

jdchristensen:
> I can work on that today if you like.

Sounds good to me, unless you think it might be a good thing for @ThomatoTomato to tackle (with an outline of what is needed).  Probably we want to implement and use something like what is in Spaces/BinInt/Equiv.v, using that in a group, multiplication by a fixed element is an equivalence.  (OTOH, if we define `grp_pow_int` directly from the definition of Int, it might compute better?  Not sure.)

jdchristensen:
Ah, I see that you already started on this, asking @ThomatoTomato for help.

Alizter:
I created #1995 which is partially finished.

PR #1993, opened 2024-06-12
Rings: remove some universe variables

jdchristensen:
This is just a start at removing some universe variables in Rings/*. 

@Alizter , please feel free to contribute to this PR, as I may not have time to investigate further.  Or we can merge this and look at other things later.

### DIFF
--- a/theories/Algebra/Rings/KroneckerDelta.v
+++ b/theories/Algebra/Rings/KroneckerDelta.v
@@ -10,15 +10,16 @@ Local Set Polymorphic Inductive Cumulativity.
 Section AssumeDecidable.
   (** Throughout this section, we assume that we have a type [A] with decidable equality. This will be our indexing type and can be thought of as [nat] for reading purposes. *)
 
-  Context {A : Type} `{DecidablePaths A}.
+  Universes u v.
+  Context {A : Type@{u}} `{DecidablePaths@{u} A} {R : Ring@{v}}.
 
   (** The Kronecker delta function is a function of elements of [A] that is 1 when the two numbers are equal and 0 otherwise. It is useful for working with finite sums of ring elements. *)
-  Definition kronecker_delta {R : Ring} (i j : A) : R
+  Definition kronecker_delta@{} (i j : A) : R
     := if dec (i = j) then 1 else 0.
 
   (** Kronecker delta with the same index is 1. *)
-  Definition kronecker_delta_refl {R : Ring} (i : A)
-    : kronecker_delta (R:=R) i i = 1.
+  Definition kronecker_delta_refl@{} (i : A)
+    : kronecker_delta i i = 1.
   Proof.
     unfold kronecker_delta.
     generalize (dec (i = i)).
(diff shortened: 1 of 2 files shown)


theories/Algebra/Rings/KroneckerDelta.v
```diff
@@ -2,6 +2,9 @@ Require Import Basics.Overture Basics.Decidable Spaces.Nat.
 Require Import Algebra.Rings.Ring.
 Require Import Classes.interfaces.abstract_algebra.
 
+Local Set Universe Minimization ToSet.
```
jdchristensen, line 5: This gets rid of lots of universe variables.

theories/Algebra/Rings/KroneckerDelta.v
```diff
@@ -111,7 +115,7 @@ Proof.
     + rewrite (kronecker_delta_neq p).
       rewrite rng_mult_zero_l.
       rewrite grp_unit_l.
-      f_ap; apply path_ishprop.
+      apply ap, path_ishprop.
```
jdchristensen, line 118: The changes to this proof only get rid of one universe variable.

theories/Algebra/Rings/Ring.v
```diff
@@ -78,7 +78,7 @@ Section RingLaws.
 
   (** Many of these ring laws have already been proven. But we give them names here so that they are easy to find and use. *)
 
-  Context {A B : Ring} (f : RingHomomorphism A B) (x y z : A).
+  Context {A : Ring} (x y z : A).
```
jdchristensen, line 81: Having an unused type `B` in context means that its universe variable gets added to every definition in the Section.  So all of those lemmas about ring operations had free universe variables, which just accumulate when the lemmas are used.  The simple change to the Sections here saves many universe variables in other places.

Alizter: It's a shame that the universe variables get used everywhere.

theories/Algebra/Rings/KroneckerDelta.v
```diff
@@ -10,32 +10,33 @@ Local Set Polymorphic Inductive Cumulativity.
 Section AssumeDecidable.
   (** Throughout this section, we assume that we have a type [A] with decidable equality. This will be our indexing type and can be thought of as [nat] for reading purposes. *)
 
-  Context {A : Type} `{DecidablePaths A}.
+  Universes u v.
+  Context {A : Type@{u}} `{DecidablePaths@{u} A} {R : Ring@{v}}.
```
jdchristensen, line 14: Just to clarify, it looks to me like the changes in this commit don't affect the universe variables at all, right?  But it does clean things up to put `R` in the Context line.

Alizter: Yes there are no universe changes but it does clear up the `R` that is shared. It also prevents any new universes from appearing without looking too messy. If you revert the changes you did for the ring laws it will complain about the extra universes now. We could perhaps transfer these to a test but this seems fine.

jdchristensen:
I'm happy to merge this for now, and do additional work in another PR, unless @Alizter plans to add more to this PR.

PR #1994, opened 2024-06-12
More flexible WildCat.Paths instances

Alizter:
This allows wildcat structures to be augmented with paths in the higher cells.

cc @ThomatoTomato

### DIFF
--- a/theories/WildCat/Paths.v
+++ b/theories/WildCat/Paths.v
@@ -19,18 +19,18 @@ Definition is2graph_paths (A : Type) `{IsGraph A} : Is2Graph A
 Definition is3graph_paths (A : Type) `{Is2Graph A} : Is3Graph A
   := fun _ _ => is2graph_paths _.
 
-(** We assume these as indexes for the rest of the file with a low priority. *)
+(** We assume these as instances for the rest of the file with a low priority. *)
 Local Existing Instances isgraph_paths is2graph_paths is3graph_paths | 10.
 
 (** Any type has composition and identity morphisms given by path concatenation and reflexivity. *)
 Local Instance is01cat_paths (A : Type) : Is01Cat A.
 Proof.
   snrapply Build_Is01Cat.
-  - intros a; exact idpath.
+  - exact (@idpath _).
   - intros a b c q p; exact (p @ q).
 Defined.
 
-(** Any type has a 0-groupoid strucutre with inverse morphisms given by path inversion. *)
+(** Any type has a 0-groupoid structure with inverse morphisms given by path inversion. *)
 Local Instance is0gpd_paths (A : Type) : Is0Gpd A.
 Proof.
   snrapply Build_Is0Gpd.
--- a/theories/WildCat/Paths.v
+++ b/theories/WildCat/Paths.v
@@ -39,22 +39,22 @@ Defined.
 
 (** Postcomposition is a 0-functor when the 2-cells are paths. *)
 Local Instance is0functor_cat_postcomp_paths (A : Type) `{Is01Cat A}
-  : forall (a b c : A) (g : b $-> c), Is0Functor (cat_postcomp a g).
+  (a b c : A) (g : b $-> c)
+  : Is0Functor (cat_postcomp a g).
 Proof.
-  intros x y z p.
   snrapply Build_Is0Functor.
-  intros q r h.
-  by destruct h.
+  intros f h p.
+  by destruct p.
 Defined.
 
 (** Precomposition is a 0-functor when the 2-cells are paths. *)
 Local Instance is0functor_cat_precomp_paths (A : Type) `{Is01Cat A}
-  : forall (a b c : A) (f : a $-> b), Is0Functor (cat_precomp c f).
+  (a b c : A) (f : a $-> b)
+  : Is0Functor (cat_precomp c f).
 Proof.
-  intros x y z p.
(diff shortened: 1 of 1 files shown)


theories/WildCat/Paths.v
```diff
@@ -4,41 +4,67 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
+
+(** Any graph is a 2-graph with 2-cells given by the identity type. *)
+Definition is2graph_paths (A : Type) `{IsGraph A} : Is2Graph A
+  := fun _ _ => isgraph_paths _.
+
+(** Any 2-graph is a 3-graph with 3-cells given by the identity type. *)
+Definition is3graph_paths (A : Type) `{Is2Graph A} : Is3Graph A
+  := fun _ _ => is2graph_paths _.
+
+(** We assume these as indexes for the rest of the file with a low priority. *)
```
jdchristensen, line 22:
```suggestion
(** We assume these as instances for the rest of the file with a low priority. *)
```

theories/WildCat/Paths.v
```diff
@@ -4,41 +4,67 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
+Local Existing Instances isgraph_paths is2graph_paths is3graph_paths | 10.
+
+(** Any type has composition and identity morphisms given by path concatenation and reflexivity. *)
 Local Instance is01cat_paths (A : Type) : Is01Cat A.
 Proof.
-  unshelve econstructor.
-  - intros a; reflexivity.
+  snrapply Build_Is01Cat.
+  - intros a; exact idpath.
```
jdchristensen, line 29:
```suggestion
  - exact (@idpath _).
```

theories/WildCat/Paths.v
```diff
@@ -4,41 +4,67 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
   - intros a b c q p; exact (p @ q).
 Defined.
 
+(** Any type has a 0-groupoid strucutre with inverse morphisms given by path inversion. *)
```
jdchristensen, line 33:
```suggestion
(** Any type has a 0-groupoid structure with inverse morphisms given by path inversion. *)
```

theories/WildCat/Paths.v
```diff
@@ -4,41 +4,67 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
 Local Instance is0gpd_paths (A : Type) : Is0Gpd A.
 Proof.
-  constructor.
-  intros x y p; exact (p^).
+  snrapply Build_Is0Gpd.
+  intros x y p; exact p^.
```
jdchristensen, line 37:
```suggestion
  exact (@inverse A).
```

theories/WildCat/Paths.v
```diff
@@ -4,41 +4,67 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
+Defined.
+
+(** Postcomposition is a 0-functor when the 2-cells are paths. *)
+Local Instance is0functor_cat_postcomp_paths (A : Type) `{Is01Cat A}
+  : forall (a b c : A) (g : b $-> c), Is0Functor (cat_postcomp a g).
+Proof.
+  intros x y z p.
+  snrapply Build_Is0Functor.
+  intros q r h.
+  by destruct h.
```
jdchristensen, line 47: I'd prefer to use the same variable names in the statement as in the proof, to make the types clear.  Maybe a b c p q r h?  Also, why not use whiskerL/R instead of destruct, to have a smaller term?  (All of this applies to the next result too.)

jdchristensen (approved):
LGTM once the comments are addressed.

theories/WildCat/Paths.v
Alizter: This is exactly because the 1-cells are not necessarily paths here.

Alizter: I'll change the names however.

theories/WildCat/Paths.v
```diff
@@ -4,41 +4,60 @@ Require Import WildCat.Core WildCat.TwoOneCat WildCat.NatTrans.
...
+Local Instance is0functor_cat_postcomp_paths (A : Type) `{Is01Cat A}
+  (a b c : A) (g : b $-> c)
+  : Is0Functor (cat_postcomp a g).
 Proof.
-  unshelve econstructor.
-  - intros a; reflexivity.
-  - intros a b c q p; exact (p @ q).
+  snrapply Build_Is0Functor.
+  intros f h p.
+  by destruct p.
```
jdchristensen, line 40:
```suggestion
  exact (@ap _ _ (cat_postcomp a g)).
```
(`exact (@ap _ _ _).`also works, but the above is more clear.)  So this could also be a one-liner.  (Also applies to next result.)

jdchristensen: I hope it's ok that I just pushed this, along with removing a bunch of beta expansions.  I'm just try to keep terms syntactically small in some of this basic material.  Feel free to revert if you don't like it.

Alizter: That all looks fine to me.

jdchristensen:
Ok, this LGTM!   @ThomatoTomato this should help with the matrix category.

PR #1995, opened 2024-06-12
integer group powers

Alizter:
We generalise group powers to integers. The nat based powers become a special case.

@ThomatoTomato I've left two parts incomplete if you would like to finish them. Make sure to remove any axioms or `Admitted`s. You may push directly to this branch. If pushing doesn't work, then you can just create a new PR.

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -237,18 +237,27 @@ Proof.
   induction n; cbn.
   - exact (grp_unit_l _)^.
   - destruct n.
-    + cbn; by rewrite !grp_unit_r.
+    + reflexivity.
     + simpl in IHn |- *.
-      rewrite IHn.
-      rewrite !simple_associativity.
-      do 2 f_ap.
-      rewrite (commutativity a).
-      rewrite <- 2 simple_associativity.
-      rewrite (commutativity b).
-      by rewrite !simple_associativity.
-  - (** TODO *)
-    Axiom transparent_admit : Empty.
-    snrapply (Empty_rec transparent_admit).
+      rewrite <- (simple_associativity a (nat_iter _ _ _)).
+      rewrite (simple_associativity (nat_iter n (sg_op a) a)).
+      rewrite (commutativity (nat_iter _ _ _) b).
+      rewrite <- (simple_associativity b (nat_iter _ _ _) (nat_iter _ _ _)).
+      rewrite (simple_associativity a).
+      f_ap.
+  - destruct n.
+    + simpl. 
+      rewrite (commutativity (- a)).
+      exact (grp_inv_op a b).
+    + simpl in IHn |- *.
+      rewrite <- (simple_associativity (- a)).
+      rewrite (simple_associativity (nat_iter _ _ _)).
+      rewrite (commutativity (nat_iter _ _ _) (-b)).
+      rewrite <- (simple_associativity (- b)).
+      rewrite (simple_associativity (-a) (-b)).
+      rewrite (commutativity (-a) (-b)).
+      rewrite <- (grp_inv_op a b).
+      f_ap.
 Defined.
 
 Definition ab_mul_homo {A B : AbGroup}
@@ -301,14 +310,12 @@ Definition ab_sum_const {A : AbGroup} (n : nat) (r : A)
   : ab_sum n f = grp_pow r n.
 Proof.
   induction n as [|n IHn] in f, p |- *.
-  1: reflexivity.
-  simpl; f_ap.
-  destruct n.
-  1: reflexivity.
-  rewrite IHn.
   - reflexivity.
-  - intros k Hk.
-    apply p.
+  - rewrite (grp_pow_nat_add_1 n r).
+    simpl. f_ap.
+    rewrite IHn.
+    + reflexivity.
+    + intros. apply p.
 Defined.
 
 (** If the function is zero in the range of a finite sum then the sum is zero. *)
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -238,24 +238,21 @@ Proof.
   - exact (grp_unit_l _)^.
   - destruct n.
     + reflexivity.
-    + simpl in IHn |- *.
-      refine (_ @ associativity _ _ _).
-      refine (_ @ (ap _ (associativity _ _ _)^)).
+    + rhs_V srapply (associativity a).
+      rhs srapply (ap _ (associativity _ b _)).
       rewrite (commutativity (nat_iter _ _ _) b).
-      refine (_ @ (ap _ (associativity _ _ _))).
-      refine (_ @ (associativity _ _ _)^).
+      rhs_V srapply (ap _ (associativity b _ _)).
+      rhs srapply (associativity a).
       apply grp_cancelL.
       exact IHn.
   - destruct n.
-    + simpl. 
-      rewrite (commutativity (- a)).
+    + rewrite (commutativity (-a)).
       exact (grp_inv_op a b).
-    + simpl in IHn |- *.
-      refine (_ @ associativity _ _ _).
-      refine (_ @ ap _ (associativity _ _ _)^).
+    + rhs_V srapply (associativity (-a)).
+      rhs srapply (ap _ (associativity _ (-b) _)).
       rewrite (commutativity (nat_iter _ _ _) (-b)).
-      refine (_ @ ap _ (associativity _ _ _)).
-      refine (_ @ (associativity _ _ _)^).
+      rhs_V srapply (ap _ (associativity (-b) _ _)).
+      rhs srapply (associativity (-a)).
       rewrite (commutativity (-a) (-b)), <- (grp_inv_op a b).
       apply grp_cancelL.
       exact IHn.
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -261,15 +261,6 @@ Proof.
       exact IHn.
 Defined.
 
-  (* 1: exact (grp_unit_l _)^.
-  refine (_ @ associativity _ _ _).
-  refine (_ @ ap _ (associativity _ _ _)^).
-  rewrite (commutativity (grp_pow a n) b).
-  refine (_ @ ap _ (associativity _ _ _)).
-  refine (_ @ (associativity _ _ _)^).
-  apply grp_cancelL.
-  assumption. *)
-
 Definition ab_mul_homo {A B : AbGroup}
   (f : GroupHomomorphism A B) (n : Int)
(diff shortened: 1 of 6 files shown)


theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -228,26 +228,32 @@ Proof.
...
+      rewrite IHn.
+      rewrite !simple_associativity.
+      do 2 f_ap.
+      rewrite (commutativity a).
+      rewrite <- 2 simple_associativity.
+      rewrite (commutativity b).
+      by rewrite !simple_associativity.
+  - (** TODO *)
+    Axiom transparent_admit : Empty.
+    snrapply (Empty_rec transparent_admit).
```
Alizter, line 251: Here

theories/Algebra/Groups/Group.v
```diff
@@ -474,28 +477,61 @@ End GroupMovement.
...
+      by snrapply ap.
 Defined.
 
+(** Note that powers don't preserve the group operation as it is not commutative. This does hold in an abelian group so such a result will appear later. *)
+
+(** [grp_pow] satisfies a law of exponents. *)
+Definition grp_pow_int_add {G : Group} (m n : Int) (g : G)
+  : grp_pow g (n + m)%int = grp_pow g n * grp_pow g m.
+Proof.
+  (** TODO: *)
```
Alizter, line 532: And here.

Let me know if you need any hints.

jdchristensen:
Another approach to `grp_pow` is to develop something like Spaces/BinInt/Equiv.v for the newer integers, which let you define a function from the integers into a pointed type with a self-equivalence.  Some of the properties are true in that generality.  Not sure whether it would complicate things here to use that more general construction.

Alizter:
This implementation doesn't compute particularly well to begin with. For instance `grp_pow g n.+1` where `n` is a `nat` cannot be unfolded but `n.+2` can. The reason is that we treat 0 differently in `Int` and we have to do a nat_ind starting from 1 giving us 3 cases.

I'm open to trying out the equivalence approach if @ThomatoTomato would like to have a go. I haven't thought much about it however.

ThomatoTomato:
Okay! Thank you for making me some exercises 🥰 I will see what I will be able to do.

theories/Algebra/Groups/Group.v
```diff
@@ -474,26 +477,102 @@ End GroupMovement.
...
+      apply grp_homo_unit.
+    + lhs nrapply grp_homo_op.
+      snrapply ap.
+      assumption.
+  - destruct n.
+    + lhs snrapply grp_homo_op.
+      cbn. rewrite grp_homo_unit. 
+      (* I don't remember how to not use rewrite here 😩 *)
+      snrapply (ap (.* _)).
+      by lhs nrapply grp_homo_inv.
```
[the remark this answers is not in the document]
jdchristensen: When it's easy to avoid rewrites, we try to.  But for set-level work like this, if using `rewrite` makes it a lot simpler, that's fine.

theories/Algebra/Groups/Group.v
```diff
@@ -518,19 +520,60 @@ Proof.
...
+    + cbn in IHn |- *.
+      rewrite grp_inv_unit in IHn |- *. 
+      by lhs nrapply grp_unit_l.
 Defined.
 
 (** Note that powers don't preserve the group operation as it is not commutative. This does hold in an abelian group so such a result will appear later. *)
 
+(** Helper functions for [grp_pow_int_add] add *)
+Definition grp_pow_int_add_1 {G : Group} (n : Int) (g : G)
+  : grp_pow g (n.+1)%int = g * grp_pow g n.
```
jdchristensen, line 532: This is a perfect example of a result that will hold in general for the `int_iter` that I am proposing.  Also, `grp_pow_homo` is a naturality statement for `int_iter`: if you have pointed types `A` and `B` with self-equivalences `g` and `h`, and a pointed map `f : A -> B` that commutes with the self-equivalences, then `f (int_iter n g pt) = int_iter n h pt`.

theories/Algebra/Groups/Group.v
```diff
@@ -518,19 +520,60 @@ Proof.
...
 (** [grp_pow] satisfies a law of exponents. *)
 Definition grp_pow_int_add {G : Group} (m n : Int) (g : G)
   : grp_pow g (n + m)%int = grp_pow g n * grp_pow g m.
 Proof.
-  (** TODO: *)
-Abort.
+  induction n; cbn.
+  - exact (grp_unit_l _)^.
+  - rewrite int_add_succ_l.
+    do 2 rewrite grp_pow_int_add_1.
```
jdchristensen, line 564: `rewrite` has build-in support for repetition, so you can use `rewrite 2 ...` or `rewrite ! ...` for as many as possible.  But for some reason it sometimes does more than one even if you use `1`.

theories/Algebra/Groups/Group.v
```diff
@@ -518,19 +520,60 @@ Proof.
...
+    rhs srapply (simple_associativity _ _ _)^.
+    snrapply (ap (_ *.)). exact IHn.
+  - rewrite <- int_neg_succ.
+    rewrite <- (int_neg_neg m) in IHn |- *.
+    rewrite <- int_neg_add in IHn |- *.
+    do 2 rewrite grp_pow_int_sign_commute in IHn |- *.
+    rewrite int_add_succ_l.
+    do 2 rewrite grp_pow_int_add_1.
+    rhs srapply (simple_associativity _ _ _)^.
```
jdchristensen, line 573: Here you should use `rhs_V srapply simple_associativity.`  The `rhs` tactic already has to invert the path you provide, so your line does double-inversion.  And using `rhs_V` lets you avoid counting the arguments.

theories/Algebra/Groups/Group.v
```diff
@@ -518,19 +520,60 @@ Proof.
...
-Abort.
+  induction n; cbn.
+  - exact (grp_unit_l _)^.
+  - rewrite int_add_succ_l.
+    do 2 rewrite grp_pow_int_add_1.
+    rhs srapply (simple_associativity _ _ _)^.
+    snrapply (ap (_ *.)). exact IHn.
+  - rewrite <- int_neg_succ.
+    rewrite <- (int_neg_neg m) in IHn |- *.
+    rewrite <- int_neg_add in IHn |- *.
```
jdchristensen, line 569: These do more rewriting than is necessary, but I couldn't figure out how to limit them.

ThomatoTomato:
(I need to learn how to make suggestions)

I've completed the two parts, but I've also changed a lot things along the way. I cannot tell if it is for the better or the worse.

I think Dan's idea about having a nat_iter for types with an auto-equivalence sounds nice, but can't we further generalize this to group actions? Do we have any theory of group actions atm?

theories/Algebra/Groups/Group.v
```diff
@@ -520,6 +520,16 @@ Defined.
 
 (** Note that powers don't preserve the group operation as it is not commutative. This does hold in an abelian group so such a result will appear later. *)
 
+Definition grp_pow_nat_add_1 {G : Group} (n : nat) (g : G) 
+  :  grp_pow g (n.+1)%nat = g * grp_pow g n.
```
jdchristensen, line 524: Isn't this a special case of the next result?  Why is it needed?

ThomatoTomato: In ab_sum we're using a natural number and not an int, and I wasn't able to coerce the natural into an integer. It shouldn't be needed, but I gave up 😔

jdchristensen: Natural numbers are automatically coerced into integers.  But that's not the issue.  It's the difference between `(n+.1)%nat` and `(n%int.+1)`.  Including the coercion `int_of_nat`, it's the difference between `int_of_nat (n.+1)` and `(int_of_nat n).+1`.  The first `.+1` is the natural number successor, while the second `.+1` is the successor on the integers.  I think we should add a lemma to Int.v saying that `int_of_nat` commutes with successors.  Then it can be used to rewrite `(n+.1)%nat` to `(n%int.+1)` so that `grp_pow_int_add_1` applies.  (When we create `int_iter`, we should similarly show that it computes as expected both on nat successors and int successors.)

ThomatoTomato:
I think I've adressed everything that has been commented on now. The change I made in Int.v probably has a shorter form, but I went with the easy solution 🥰 I managed to get rid of many of the rewrites, but I made a helper function that is very similar to grp_pow_int_add_1, called grp_pow_int_sub_1. I could prove it using the former, but then I had to use a lot of rewrites again. I think this is the cleanest way to do this, but idk.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -239,27 +239,37 @@ Proof.
   - destruct n.
     + reflexivity.
     + simpl in IHn |- *.
-      rewrite <- (simple_associativity a (nat_iter _ _ _)).
-      rewrite (simple_associativity (nat_iter n (sg_op a) a)).
+      refine (_ @ associativity _ _ _).
+      refine (_ @ (ap _ (associativity _ _ _)^)).
```
jdchristensen, line 243: These and others below look like perfect candidates for `rhs` and `rhs_V`, which let you avoid some of the underscores, etc.

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -239,27 +239,37 @@ Proof.
...
 
+  (* 1: exact (grp_unit_l _)^.
+  refine (_ @ associativity _ _ _).
+  refine (_ @ ap _ (associativity _ _ _)^).
+  rewrite (commutativity (grp_pow a n) b).
+  refine (_ @ ap _ (associativity _ _ _)).
+  refine (_ @ (associativity _ _ _)^).
+  apply grp_cancelL.
+  assumption. *)
+
```
jdchristensen, line 272:
```suggestion
```

theories/Spaces/Int.v
```diff
@@ -78,6 +78,13 @@ Definition int_pred (n : Int) : Int :=
 
 Notation "n .-1" := (int_pred n) : int_scope.
 
+(** [int_of_nat] is commutes with taking succesors *)
+
+Definition int_of_nat_succ_commute (n : nat) : S n = int_succ(n) :> Int.
```
jdchristensen, line 83: First, a minor suggestion about use of parentheses:
```suggestion
Definition int_of_nat_succ_commute (n : nat) : S n = int_succ n :> Int.
```
But I also think it would be better to explicitly write the coercion function on both sides of the statement.

jdchristensen:
> I think I've adressed everything that has been commented on now.

In addition to the new comments, there are still a couple of older ones that haven't been addressed.  I marked the completed things as resolved.  (A couple that aren't resolved are about int_iter, which we'll postpone to a future PR.)

jdchristensen:
This LGTM.  @ThomatoTomato is there anything else you want to do, or should I go ahead and merge?

ThomatoTomato:
I'm ready to merge, so go ahead.

PR #1996, opened 2024-06-12
recommend coq-lsp over vscoq

Alizter:
Users are free to use whichever editor they like, but I would recommend `coq-lsp` over vscoq due to the smoother experience for Coq-HoTT.


jdchristensen:
LGTM.  I don't use VSCode, but is it worth mentioning VSCoq as another option, with a warning that there might be issues?

Alizter:
I've also added a link to Coq's list of editors where vscoq appears. I would rather keep the list here short and only consisting of recommended editors.

jdchristensen (approved):

PR #1997, opened 2024-06-14
Make some WildCat.Path instances Global, and use them

jdchristensen:
We don't want `isgraph_paths`, `is2graph_paths`, or `is3graph_paths` to be global instances, as then Coq could apply them in situations where we prefer a different wild category structure.  But the further instances that require having these ones already applied can safely be global instances, so that's what this PR does. 

This allows us to simplify things in the three files that currently use WildCat.Paths:  Suspension.v, CoeqUnivProp.v and Matrix.v.

I have checked that the build time is unchanged within the noise.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -4,8 +4,7 @@ Require Import Spaces.List.Core Spaces.List.Theory Spaces.List.Paths.
 Require Import Algebra.Rings.Ring Algebra.Rings.Module Algebra.Rings.CRing
   Algebra.Rings.KroneckerDelta Algebra.Rings.Vector.
 Require Import abstract_algebra.
-Require Import WildCat.Core.
-Require Import WildCat.Paths.
+Require Import WildCat.Core WildCat.Paths.
 
 Set Universe Minimization ToSet.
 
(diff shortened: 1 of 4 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -658,34 +657,15 @@ Section MatrixCat.
...
 
   (** MatrixCat R forms a strong 1-category. *)
   Global Instance is1catstrong_matrixcat {R : Ring} : Is1Cat_Strong (MatrixCat R).
   Proof.
     snrapply Build_Is1Cat_Strong.
-    1,2,4: exact _.
-    (* I'm not sure why typeclass search doesn't find the next one.  Something must be slightly broken. *)
-    - snrapply @is0functor_postcomp_matrixcat_hom.
+    (* Most of the structure comes from typeclasses in WildCat.Paths. *)
+    1-4: exact _.
```
jdchristensen, line 668: Note that the problem involving typeclass inference for goal 3 has disappeared.  Further testing shows that it *did* work using is0functor_postcomp_matrixcat_hom, but took around 5s.  Using the instance from WildCat.Paths, like in this PR, it is fast.  I don't know why there is a difference.

Alizter (approved):
This seems fine to me although Ibhaven't had a chance to inspect the PR interactively.

PR #1998, opened 2024-06-27
fix nix flake

Alizter:
I missed an extra argument here before which meant that the development shell didn't build. I've fixed this and updated the flake.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1717774105,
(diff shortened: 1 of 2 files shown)


jdchristensen:
Just to confirm:  the flake.lock file is intentionally tracked in git?

Alizter:
@jdchristensen Yes. The flake.nix file tells nix how to build something and the flake.lock file tells nix which version of its dependencies (mostly nixpkgs (a huge monorepo of packages)) to use. This makes nix environments reproducible.

jdchristensen (approved):

PR #1999, opened 2024-06-27
fix universes in Matrix.v

Alizter:
In this PR, we reduce the number of universe variables appearing in Matrix.v. I'm not particularly sure why Coq cannot minimize these cases itself, but adding a few annotations helps it come to the correct number of variables.

The number of universes in matrix_ring went from 4 to 1. This will help with the slow down in #1981.

### DIFF
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -309,7 +309,8 @@ Definition ab_sum_const {A : AbGroup} (n : nat) (r : A)
 Proof.
   induction n as [|n IHn] in f, p |- *.
   - reflexivity.
-  - rewrite int_of_nat_succ_commute, grp_pow_int_add_1.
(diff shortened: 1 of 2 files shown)


jdchristensen:
The issues you found were caused by an underlying issue, namely a stray universe variable in `ab_sum_const`.  Getting rid of that made Coq do the right thing in Matrix.v, so all but one of your fixes is not required.  The stray universe variable in `ab_sum_const` was introduced in a `rewrite`, and I was able to avoid it by not using `rewrite`.

jdchristensen:
Feel free to merge once you have reviewed my commit.  (Github won't let me request you as a reviewer, probably because you created the PR.)

jdchristensen:
BTW, the fix to `ab_sum_const` also removed stray universe variables from `ab_sum_zero`, `ab_sum_sum`, `rng_sum_kronecker_delta_l`, `rng_sum_kronecker_delta_r'`, and probably several other results.  That's why it's always good to chase the problem to the source of the free universe variable.

PR #2000, opened 2024-06-28
switch to Int in cring_Z

Alizter:
This was a bit more work than I anticipated but I seem to have something workable. It could use some more cleanup as I think some of these proofs could be slicker.

In this PR we port the remaining uses of BinInt to Int and introduce `int_iter` with standard exponentiation results on equivalences and loops.

Feel free to push any commits.

@ThomatoTomato this should help you with #1992.

### DIFF
--- a/theories/Algebra/Rings/Z.v
+++ b/theories/Algebra/Rings/Z.v
@@ -26,12 +26,12 @@ Local Open Scope mc_scope.
 
 (** Every ring has a unique ring map from the integers. *)
 (** TODO: rename? *)
-Definition rng_int_mult (R : Ring) (z : cring_Z) : R
-  := int_iter (fun x => 1 + x) z 0.
+Definition rng_int_mult (R : Ring) (r : R) (z : cring_Z) : R
+  := int_iter (fun x => r + x) z 0.
 
 (** Preservation of + *)
 Global Instance issemigrouppreserving_plus_rng_int_mult (R : Ring)
-  : IsSemiGroupPreserving (Aop:=(+)) (Bop:=(+)) (rng_int_mult R).
+  : IsSemiGroupPreserving (Aop:=(+)) (Bop:=(+)) (rng_int_mult R 1).
 Proof.
   intros x y.
   induction x as [|x|x].
@@ -55,19 +55,23 @@ Proof.
     f_ap.
 Defined.
 
+Global Instance isunitpreserving_plus_rng_int_mult (R : Ring)
+  : IsUnitPreserving (Aunit:=zero) (Bunit:= canonical_names.zero) (rng_int_mult R 1)
+  := idpath.
+
+Global Instance ismonoidpreserving_plus_rng_int_mult (R : Ring)
+  : IsMonoidPreserving (Aop:=(+)) (Bop:=(+)) (rng_int_mult R 1)
+  := {}.
+
 Lemma rng_int_mult_neg {R} x
-  : rng_int_mult R (- x) = - rng_int_mult R x.
+  : rng_int_mult R 1 (- x) = - rng_int_mult R 1 x.
 Proof.
-  snrapply (groups.preserves_negate _).
-  1-6: typeclasses eauto.
-  snrapply Build_IsMonoidPreserving.
-  1: exact _.
-  split.
+  snrapply (groups.preserves_negate _); exact _.
 Defined.
 
 (** Preservation of * (multiplication) *)
 Global Instance issemigrouppreserving_mult_rng_int_mult (R : Ring)
-  : IsSemiGroupPreserving (Aop:=(.*.)) (Bop:=(.*.)) (rng_int_mult R).
+  : IsSemiGroupPreserving (Aop:=(.*.)) (Bop:=(.*.)) (rng_int_mult R 1).
 Proof.
   intros x y.
   induction x as [|x|x].
@@ -98,7 +102,7 @@ Defined.
 Definition rng_homo_int (R : Ring) : (cring_Z : Ring) $-> R.
 Proof.
   snrapply Build_RingHomomorphism.
-  1: exact (rng_int_mult R).
+  1: exact (rng_int_mult R 1).
   repeat split.
   1,2: exact _.
   hnf.
--- a/theories/Algebra/Rings/Z.v
+++ b/theories/Algebra/Rings/Z.v
@@ -37,14 +37,14 @@ Proof.
   - simpl.
     by rhs nrapply rng_mult_zero_l.
   - rewrite int_mul_succ_l.
-    rewrite grp_pow_int_add.
-    rewrite grp_pow_int_add_1.
+    rewrite grp_pow_add.
+    rewrite grp_pow_succ.
     rhs nrapply rng_dist_r.
     rewrite rng_mult_one_l.
     f_ap.
   - rewrite int_mul_pred_l.
-    rewrite grp_pow_int_add.
-    rewrite grp_pow_int_sub_1.
+    rewrite grp_pow_add.
+    rewrite grp_pow_pred.
     rhs nrapply rng_dist_r.
     rewrite IHx.
     f_ap.
@@ -72,12 +72,12 @@ Proof.
   unfold rng_homo_int, rng_int_mult; cbn.
   induction x as [|x|x].
   - by rhs nrapply (grp_homo_unit g).
-  - rewrite grp_pow_int_add_1.
+  - rewrite grp_pow_succ.
     change (x.+1%int) with (1 + x)%int.
     rewrite (rng_homo_plus g 1 x).
     rewrite rng_homo_one.
     f_ap.
-  - rewrite grp_pow_int_sub_1.
+  - rewrite grp_pow_pred.
     rewrite IHx.
     clear IHx.
     rewrite <- (rng_homo_one g).
--- a/theories/Algebra/Rings/Z.v
+++ b/theories/Algebra/Rings/Z.v
@@ -24,36 +24,31 @@ Defined.
 
 Local Open Scope mc_scope.
 
-Definition rng_int_mult (R : Ring) := abgroup_int_mult R. 
+(** Given a ring element [r], we get a map [Int -> R] sending an integer to that multiple of [r]. *)
+Definition rng_int_mult (R : Ring) := grp_pow_homo : R -> Int -> R.
 
 (** [rng_int_mult R 1] preserves multiplication.  This requires that the specified ring element is the unit. *)
 Global Instance issemigrouppreserving_mult_rng_int_mult (R : Ring)
   : IsSemiGroupPreserving (A:=cring_Z) (Aop:=(.*.)) (Bop:=(.*.)) (rng_int_mult R 1).
 Proof.
   intros x y.
+  cbn.
   induction x as [|x|x].
   - simpl.
     by rhs nrapply rng_mult_zero_l.
-  - cbn.
-    rewrite int_mul_succ_l.
-    unfold rng_int_mult in *.
-    rewrite issemigrouppreserving_plus_abgroup_int_mult.
-    unfold abgroup_int_mult in *.
+  - rewrite int_mul_succ_l.
+    rewrite grp_pow_int_add.
     rewrite grp_pow_int_add_1.
     rhs nrapply rng_dist_r.
     rewrite rng_mult_one_l.
-    rewrite IHx.
-    reflexivity.
-  - cbn.
-    rewrite int_mul_pred_l.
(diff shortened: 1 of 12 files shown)


ThomatoTomato:
I added a small lemma that I think can be useful in the future. This lemma also has an accompanying lemma that I do not think belong to this file, and it might already exist.

theories/Spaces/Int.v
```diff
@@ -617,6 +617,39 @@ Proof.
     f_ap.
 Defined.
 
+(** This is a general lemma about commutativity of monoid units. Don't know what file this should belong to. *)
+Lemma inverse_commute_commute {A} (f : A -> A) `{!IsEquiv f} (g : A -> A) (p : g o f == f o g)
+  : g o f^-1 == f^-1 o g.
```
Alizter, line 622: I think this is a simple application of `moveR_equiv_eM`, associativity followed by `moveL_equiv_Me`. Since functions are definitionally associative you probably don't need the associativity step.

Alizter:
@ThomatoTomato I simplified the proof of the lemma you gave.

theories/Spaces/Int.v
ThomatoTomato: Uhm, I might be stupid, but I cannot find any instances for moveR_equiv_eM, nor the other guy. Should I make these maps, put them in Basics/Equivalences.v, and rewrite the lemma using these maps?

jdchristensen:
I think `int_iter` has worked well to simplify a lot of things.  Thanks for tackling this, @Alizter !

The proofs about iterated powers of loops and the proof that the natural map from the integers to a ring respects addition all should be special cases about suitable functions of two variables.  I started to write some thing down, but ran out of time.  Maybe someone can figure out the right abstraction and use it in all of those places?  Here's what I had:
```coq
(** ** Iterating left-invertible binary operations. *)

Definition int_iter_op_l {A : Type} (f : A -> A -> A)
  (n : Int) (a b : A) `{IsEquiv _ _ (f a)}
  : A
  := int_iter (f a) n b.

Definition int_iter_op_l_succ_l {A : Type} (f : A -> A -> A)
  (n : Int) (a b : A) `{IsEquiv _ _ (f a)}
  : int_iter_op_l f n.+1 a b = f a (int_iter_op_l f n a b)
  := int_iter_succ_l (f a) n b.

Definition int_iter_op_l_succ_r {A : Type} (f : A -> A -> A)
  (n : Int) (a b : A) `{IsEquiv _ _ (f a)}
  : int_iter_op_l f n.+1 a b = int_iter_op_l f n a (f a b)
  := int_iter_succ_r (f a) n b.

(** When [f] is associative, this can be written a different way. *)
Definition int_iter_op_l_succ_r' {A : Type} (f : A -> A -> A)
  (n : Int) (a b : A) `{IsEquiv _ _ (f a)}
  {ass : forall x , f (f a x) b = f a (f x b)}
  : int_iter_op_l f n.+1 a b = f (int_iter_op_l f n a a) b.
Proof.
  lhs nrapply int_iter_op_l_succ_r.
  symmetry; exact (int_iter_commute_function_application (f a) (fun x => f x b) ass n a).
Defined.

(** When [f] is associative and [b] is a unit, we have another way. *)
Definition int_iter_op_l_succ_r'' {A : Type} (f : A -> A -> A)
  (n : Int) (a b : A) `{IsEquiv _ _ (f a)}
  {ass : forall x , f (f a x) b = f a (f x b)}
  {unit_l : forall x, f b x = x}
  {unit_r : forall x, f x b = x}
  : int_iter_op_l f n.+1 a b = f (int_iter_op_l f n a b) a.
Proof.
  (* Probably have to do another induction here.  I couldn't see an abstract way to prove it. *)
Abort.
```

theories/Algebra/Rings/Z.v
```diff
@@ -10,155 +10,53 @@ Require Import WildCat.Core.
...
-  1: exact (fun y => sg_op y (- x)).
-  all: intro y.
-  all: refine ((associativity _ _ _)^ @ _ @ right_identity y).
-  all: refine (ap (y *.) _).
-  1: apply left_inverse.
-  apply right_inverse.
-Defined.
+(** Every ring has a unique ring map from the integers. *)
+(** TODO: rename? *)
+Definition rng_int_mult (R : Ring) (z : cring_Z) : R
```
jdchristensen, line 29: I don't think `ring_int_mult` is a good name, as it's not so much about multiplication as about iteration.  (If we were multiplying a general ring element by an integer, then this name might be ok.)

theories/Algebra/Rings/Z.v
```diff
@@ -167,59 +65,44 @@ Proof.
   split.
```
jdchristensen, line 65: It looks to me like the `Build_IsMonoidPreserving` part should probably be a separate lemma.

theories/Spaces/Int.v
```diff
@@ -526,3 +531,201 @@ Proof.
...
+Definition int_iter_neg {A} (f : A -> A) `{IsEquiv _ _ f} (n : Int) (a : A)
+  : int_iter f (- n) a = int_iter f^-1 n a.
+Proof.
+  by destruct n.
+Defined.
+
+Definition int_iter_succ_l {A} (f : A -> A) `{IsEquiv _ _ f} (n : Int) (a : A)
+  : int_iter f (int_succ n) a = f (int_iter f n a).
+Proof.
+  induction n as [|n|n].
```
jdchristensen, line 557: One pattern I noticed is that `int_iter` is defined directly from the definition of `Int`, and this doesn't mesh well with the custom induction principle we defined.  That's why `destruct n` works quite well when the goal doesn't need an induction hypothesis.  I wonder if we should also have the default induction principle available, and if it would help with some of these goals?  Or if we should rephrase the custom induction principle so we don't need things like `int_pred_succ` so often?

jdchristensen:
I just pushed another commit and simplifies things by noticing that they are common instances of a general result.

I wonder how close our results on `int_iter` are to the Scoccola-Altenkirch induction principle?  Once we have that, we can probably avoid doing induction on integers in many cases.

theories/Algebra/Rings/Z.v
Alizter: I think that is what I had in mind when I renamed it (from "catamorphism").

Alizter:
@jdchristensen Isn't `int_ter` exactly the Scoccola-Altenkirch recursor?

theories/Spaces/Int.v
Alizter: I'm not sure what you mean by default induction principle here. The only eliminator we have for `Int` is one by cases and it is non-recursive. This is essentially because we just have two copies of `nat` stuck together around a base point as a sum type.

Even when we destruct by cases we still have to case split on `n` in `(negS n).+1` being `0` or `.+1` given how we define successor. So it doesn't appear to gain us anything as we run into the same problems. I am however unsure if a Scoccola-Altenkirch induction prinicple would work, however it seems pretty complicated to write down. I guess in the hset case it would be simplified.

jdchristensen:
Now the name `rng_int_mult` makes sense.  I generalized the results about it to not assume that the ring element is `1`, except for preservation of multiplication.  I also updated the comments.  I removed the comment about opacity, since I didn't see anything opaque there.

theories/Algebra/Rings/Z.v
```diff
@@ -1,225 +1,114 @@
...
-(** We call this a "catamorphism" which is the name of the map from an initial object. It seems to be a more common terminology in computer science. *)
-Definition cring_catamorphism_fun (R : CRing) (z : cring_Z) : R
-  := match z with
-     | neg z => pos_peano_rec R (-1) (fun n nr => -1 + nr) z
-     | 0%binint => 0
-     | pos z => pos_peano_rec R 1 (fun n nr => 1 + nr) z
-     end.
+(** Every ring element can be multiplied by an integer. *)
+Definition rng_int_mult (R : Ring) (r : R) (z : cring_Z) : R
+  := int_iter (fun x => r + x) z 0.
```
jdchristensen, line 29: This definition and the first three instances below all work when `R` is replaced by any abelian group, so I think they should be renamed and generalized, and moved to the AbGroups folder.

jdchristensen:
@Alizter My thoughts on these induction principles are a bit vague, but I get the sense that very similar manipulations are often being done when using our current induction principle.  One idea I had would be to change the hypotheses to
```coq
  (HP : forall n : nat, P n -> P (n.+1%nat))
  (HN : forall n : nat, P (- n) -> P (-(n.+1%nat)))
```
instead of
```coq
  (HP : forall n : nat, P n -> P (int_succ n))
  (HN : forall n : nat, P (- n) -> P (int_pred (-n)))
```
(The proof is unchanged.)  That might make some arguments go through more smoothly, but at least one becomes trickier, so it's not clear if this is a good idea.  Maybe we should make both available?

Yes, I think `int_iter` is the Scoccola-Altenkirch recursor, but I was really thinking about the induction principle.  Again, this is just a vague thought that might guide what we do.

theories/Spaces/Int.v
jdchristensen: (I should have put my top-level comment here.  The idea is that maybe by using natural number successor instead of int_succ, we might not have to destruct n in as many places.)

theories/Algebra/Rings/Z.v
jdchristensen: We'd like to show that the category of abelian groups is equivalent to the category of Z-modules, and these four things will be the key ingredients for that, so we'll need to have them for abelian groups, not just for rings.

Alizter: I've generalised and moved the part that works for abelian groups to AbGroups/Z.v.

jdchristensen:
In the last commit, I've done some cleanups to grp_pow, ab_mul, abgroup_int_mult and rng_int_mult.  Note that the last three are all special cases of grp_pow.

There are still some tasks to do.

- The grp_pow material in Group.v should be redone using int_iter.  This is essentially what Ali did earlier in this PR, when defining rng_int_mult and proving properties about it.  Those proofs were discarded in a recent commit, but should be brought back and used to replace the proofs of grp_pow in Group.v.

- abgroup_int_mult is just grp_pow, so I think we can just drop it.

- ab_mul is grp_pow, but regarded as a homomorphism A -> A for fixed n (needs A abelian), so this should be kept.  And rng_int_mult again should be kept.

- I think issemigrouppreserving_mult_rng_int_mult can be proven by combining some pieces that will be useful on their own.  See below for a sketch.  This is optional, but the lemmas might be good tasks for @ThomatoTomato .

Group.v:

```coq
(** [grp_pow g n] commutes with [g]. *)
Definition grp_pow_commutes {G : Group} (n : Int) (g : G)
  : (grp_pow g n) * g = g * (grp_pow g n).

(** If [g] and [h] commute, then [grp_pow (g * h) n] = (grp_pow g n) * (grp_pow h n)]. *)
(* Note that part of the proof that [ab_mul] is a homomorphism will be covered by this. *)
Definition grp_pow_mul {G : Group} (n : Int) (g h : G)
  : grp_pow (g * h) n] = (grp_pow g n) * (grp_pow h n).

(** [grp_pow] satisfies a multiplicative law of exponents. *)
Definition grp_pow_int_mul {G : Group} (m n : Int) (g : G)
  : grp_pow g (m * n)%int = grp_pow (grp_pow g m) n.
(* This will follow from the previous two. *)
```

Rings/Z.v:

```coq
Definition rng_int_mult_foo {R : Ring} (r : R) (n : Int)
  : rng_int_mult r n = (rng_int_mult 1 n) * r.

(* I think issemigrouppreserving_mult_rng_int_mult will follow from the previous item and grp_pow_int_mul. *)
```

jdchristensen:
> * The grp_pow material in Group.v should be redone using int_iter. 

My second last commit does this.  The last commit just does some cleanups to the naming and the order.  So the second-last commit should be viewed on its own to better see the simplifications from using int_iter.

> * abgroup_int_mult is just grp_pow, so I think we can just drop it.

I'll do this soon unless someone objects.

> * I think issemigrouppreserving_mult_rng_int_mult can be proven by combining some pieces that will be useful on their own.  See below for a sketch.  This is optional, but the lemmas might be good tasks for @ThomatoTomato .

If @Alizter or @ThomatoTomato plan to do this, write here first so nobody duplicates work.

Alizter:
I'll leave it to @ThomatoTomato.

jdchristensen:
I think we should merge this as is, and leave my suggestions about the proof of `issemigrouppreserving_mult_rng_int_mult` for a future PR.

Alizter:
@jdchristensen Very well. Could you create an issue so we don't forget?

PR #2002, opened 2024-06-29
skew-symmetric matrices

Alizter:
In this PR we define what is means for a matrix to be skew-symmetric and prove some basic closure properties.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -955,6 +955,69 @@ Defined.
 
 (** Note that symmetric matrices do not form a subring (or subalgebra) but they do form a submodule of the module of matrices. *)
 
+(** ** Skew-symmetric Matrices *)
+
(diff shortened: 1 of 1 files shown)


theories/Algebra/Rings/Matrix.v
```diff
@@ -955,6 +955,69 @@ Defined.
...
+Global Instance isskewsymmetric_matrix_negate {R : Ring@{i}} {n : nat}
+  (M : Matrix R n n) `{!IsSkewSymmetric M}
+  : IsSkewSymmetric (matrix_negate M).
+Proof.
+  unfold IsSkewSymmetric.
+  rewrite matrix_transpose_negate.
+  f_ap.
+Defined.
+
+(** The scalar multiplication of a skew-symmetric matrix is skew-symmetric. *)
```
jdchristensen, line 986:
```suggestion
(** A scalar multiple of a skew-symmetric matrix is skew-symmetric. *)
```

jdchristensen (approved):

PR #2003, opened 2024-06-30
centrosymmetric matrices

Alizter:
In this PR we define centrosymmetric matrices which are matrices which are symmetric about their center. Unlike symmetric and skew-symmetric matrices, these form a subring of the ring of matrices.


theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
 (** Skew-symmetric matrices degenerate to symmetric matrices in rings with characteristic 2. In odd characteristic the module of matrices can be decomposed into the direct sum of symmetric and skew-symmetric matrices. *)
 
+(** ** Exchange matrix *)
+
+(** The exchange matrix is the matrix with ones on the anti-diagonal and zeros elsewhere. It will play an important role in defining classes of matrices with certain symmetric properties. *)
+Definition exchange_matrix (R : Ring) n : Matrix R n n
+  := Build_Matrix R n n (fun i j Hi Hj
+      => kronecker_delta (R:=R) (i + j) (pred n))%nat.
+
+(** Multiplying a matrix by the exchange matrix on the left is equivalent to exchanging the rows. Similarly multiplying on the right exchanges the columns. *)
```
jdchristensen, line 1028:
```suggestion
(** Multiplying a matrix by the exchange matrix on the left reverses the order of the rows. Similarly multiplying on the right reverses the order of the columns. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
+Defined.
+
+(** The exchange matrix is symmetric. *)
+Global Instance issymmetric_exchange {R : Ring} {n : nat}
+  : IsSymmetric (exchange_matrix R n)
+  := exchange_matrix_transpose.
+
+(** The exchange matrix has order 2. This proof is only long because of arithmetic. *)
+Definition exchange_matrix_square {R : Ring} {n : nat}
+  : matrix_mult (exchange_matrix R n) (exchange_matrix R n) = identity_matrix R n.
```
jdchristensen, line 1047: This also follows from the claim above that `(exchange_matrix R n) * M` reverses the order of the rows of `M`.  I wonder if this claim is a bit easier to prove?

theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
+  : IsCentrosymmetric (matrix_lact r M).
+Proof.
+  unfold IsCentrosymmetric.
+  (** TODO: Need associative algebra structure. *)
+Abort.
+
+(** A centrosymmetric matrix is also centrosymmetric when considered over the opposite ring. *)
+Global Instance iscentrosymmetric_rng_op {R : Ring@{i}} {n : nat} (M : Matrix R n n)
+  `{!IsCentrosymmetric M}
+  : IsCentrosymmetric (R := rng_op R) M.
```
jdchristensen, line 1132: This is trivial if you know that being centrosymmetric is equivalent to the (i,j) entry being equal to the (n-1-i, n-1-j) entry, since this condition doesn't involve the ring structure at all.  And if the claim about how the exchange matrix multiple other matrices is proven, this characterization would be easy.  Just a thought.

theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
+  - apply path_ab_sum.
+    intros k Hk.
+    rewrite 2 entry_Build_Matrix.
+    rapply kronecker_delta_comm.
+Defined.
+
+(** The transpose of a centrosymmetric matrix is centrosymmetric. *)
+Global Instance iscentrosymmetric_matrix_transpose {R : Ring@{i}} {n : nat}
+  (M : Matrix R n n) {H : IsCentrosymmetric M}
+  : IsCentrosymmetric (matrix_transpose M).
```
jdchristensen, line 1155: This might also be easier with the equivalent characterization?

theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
+  : IsCentrosymmetric (matrix_mult M N).
+Proof.
+  lhs nrapply (rng_mult_assoc (A:=matrix_ring R n)).
+  rhs_V nrapply (rng_mult_assoc (A:=matrix_ring R n)).
+  hnf in H1, H2; cbn.
+  rewrite H1.
+  lhs_V nrapply (rng_mult_assoc (A:=matrix_ring R n)); f_ap.
+Defined.
+
+(** Centrosymmetric rings form a subring of the matrix ring. *)
```
jdchristensen, line 1179:
```suggestion
(** Centrosymmetric matrices form a subring of the matrix ring. *)
```

theories/Algebra/Rings/Matrix.v
```diff
@@ -1018,6 +1018,172 @@ Defined.
...
+Definition centrosymmetric_matrix_ring (R : Ring@{i}) (n : nat)
+  : Subring (matrix_ring R n).
+Proof.
+  nrapply (Build_Subring' (fun M : matrix_ring R n => IsCentrosymmetric M));
+    cbn beta; exact _.
```
jdchristensen, line 1184: Can you check if this is fast?

jdchristensen (approved):

theories/Algebra/Rings/Matrix.v
Alizter: I've sped it up the same as we did before.

theories/Algebra/Rings/Matrix.v
Alizter: This may work out, however I found it quite awkward to write the condition as we need Coq to infer that `n - 1 - i < n` and I couldn't find a way to do this easily.

theories/Algebra/Rings/Matrix.v
Alizter: I've factored out that part of the proof, but it is still a little tricky. I might be missing something but I'll have to take another look when I have more time.

Alizter:
@jdchristensen I've tried to make progress on this but haven't been able to. I tried to characterize centrosymmetric matrices using the equivalent definition however getting Coq to find `pred n - i <= n` automatically hasn't been easy. I've tried to use `let`s to do this, but it complicated the proofs later on. I think I would prefer to keep it this way for now. Have you got any other suggestions for improvement?

theories/Algebra/Rings/Matrix.v
jdchristensen: It doesn't work to define a (possibly Local) instance stating this inequality?

jdchristensen: Or maybe it helps to write it as `n - (i.+1) < n`?

jdchristensen: I made a couple of minor changes.  If `i <= pred n` can be inferred from `i < n` with some kind of hint, then with the new style we should be able to simply remove the `auto with nat` lines.  (I also added a comment explaining the `set t`, which confused me at first.)

Alizter:
@jdchristensen Thanks a lot for the improvements. I would like to spend some more time thinking about this and see if I can improve things further. I will have another go at defining centrosymmetric matrices using the alternative characterisation.

Alizter:
@jdchristensen I got further than before but I run into issues with showing the identity matrix is centrosymmetric. Overall it seems the index approach might be more general, but it greatly complicates even basic things like this. Would you like to have a look at the WIP commit to see the difficulty I am running into.

jdchristensen:
I'm busy for a few days, but I'll try to take a look at this when I'm free.

Alizter:
Alright, no rush. This is a relatively low priority PR anyway and should be the last of my matrix backlog. I'm going to be focusing on abelian groups and modules some more.

jdchristensen:
> Overall it seems the index approach might be more general, but it greatly complicates even basic things like this.

I tried a few more things in the file, and many got easier (including some that were much easier).  (See my last commit.)  So I definitely think we want to have the index version around.

> I got further than before but I run into issues with showing the identity matrix is centrosymmetric. 

I don't think that particular goal looks that tricky.  But I'm also ok with using the exchange matrix approach when it is more convenient, as in cases like this.  So I suggest one of two approaches:

- Change the definition is IsCentroSymmetric back to the exchange matrix definition.  Keep the lemma that says that the index version implies the exchange matrix version, and use that lemma to prove that many things are centrosymmetric using indices.  (And we can drop the `merely` if we go this route.)

- Keep the definition of IsCentroSymmetric using indices, and add a lemma saying that the exchange matrix definition implies the index definition.  Then, we use this new lemma when it makes things easier, like for the identity matrix.

Alizter:
I'm converting this to a draft for now as we can come back to it when we have the time.

PR #2004, opened 2024-06-30
comonoid objects in opposite category are monoids

Alizter:
Here we prove that comonoid objects in an opposite category are monoid objects in the original category.

This is for #1929.

### DIFF
--- a/theories/Algebra/Categorical/MonoidObject.v
+++ b/theories/Algebra/Categorical/MonoidObject.v
@@ -95,6 +95,37 @@ Section ComonoidObject.
   Definition co_counit {x : A} `{!IsComonoidObject x} : x $-> unit
     := mo_unit (A:=A^op) (tensor:=tensor) (unit:=unit) (x:=x).
 
+  (** Coassociativity *)
+  Definition co_coassoc {x : A} `{!IsComonoidObject x}
+    : associator x x x $o fmap01 tensor x co_comult $o co_comult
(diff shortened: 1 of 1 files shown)


Alizter:
cc @ThomatoTomato as you had a look at this before.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -115,8 +146,58 @@ Section ComonoidObject.
...
+    apply cate_moveL_eM.
+    refine (cat_assoc _ _ _ $@ _).
+    exact (co_left_counit (A:=A^op) tensor unit (x:=x)).
+  - simpl; nrefine (_ $@ cat_idl _).
+    apply cate_moveL_eM.
+    refine (cat_assoc _ _ _ $@ _).
+    exact (co_right_counit (A:=A^op) tensor unit (x:=x)).
+Defined.
+
+(** A cocoassociative cocomonoid object in [A^op] is a commutative monoid object in [A]. *)
```
jdchristensen, line 187:
```suggestion
(** A cocommutative cocomonoid object in [A^op] is a commutative monoid object in [A]. *)
```

jdchristensen (approved):

theories/Algebra/Categorical/MonoidObject.v
Alizter: Oops looks like I was coconfused when I wrote that.

PR #2005, opened 2024-07-01
invertible ring elements

Alizter:
In this PR we:
- Define notions of left and right invertibility of ring elements and define invertible elements as those which are both left and right invertible.
- We prove basic uniqueness properties of inverses. 
- We prove basic closure properties of invertibility.
- We define the group of units of a ring.

### DIFF
--- a/theories/Algebra/Rings/Ring.v
+++ b/theories/Algebra/Rings/Ring.v
@@ -30,9 +30,9 @@ Record Ring := Build_Ring' {
 }.
 
 
-Arguments ring_mult {_}.
-Arguments ring_one {_}.
-Arguments ring_isring {_}.
+Arguments ring_mult {R} : rename.
+Arguments ring_one {R} : rename.
+Arguments ring_isring {R} : rename.
 
 Definition issig_Ring : _ <~> Ring := ltac:(issig).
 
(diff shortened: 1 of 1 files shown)


theories/Algebra/Rings/Ring.v
```diff
@@ -572,3 +572,261 @@ Proof.
...
+    exists (left_inverse_elem x); split.
+    + apply rng_inv_l.
+    + apply rng_inv_r.
+  - intros i.
+    unfold Build_IsInvertible; cbn.
+    apply (ap (Build_IsInvertible' R x _)).
+    apply (equiv_ap_inv (issig_IsLeftInvertible _)); cbn.
+    change (left_inverse_elem (R:=rng_op R) x) with (right_inverse_elem x).
+    rapply path_sigma_hprop; cbn.
+    apply path_left_inverse_elem_right_inverse_elem.
```
jdchristensen, line 694: The changes I made here avoid destructing `i` and also avoid using `issig_IsInvertible`, which is not a super-useful lemma.

BTW, this whole proof can be replaced with:
```coq
  equiv_via { i : IsInvertible R x & left_inverse_elem x = right_inverse_elem x }.
  1: make_equiv_contr_basedpaths.
  apply equiv_sigma_contr; intro i.
  rapply contr_inhabited_hprop.
  apply path_left_inverse_elem_right_inverse_elem.
```
which is a lot more conceptual.  But this proof doesn't seem to compute as well, which means that  the proof of the hprop property doesn't go through as written.

theories/Algebra/Rings/Ring.v
```diff
@@ -572,3 +572,261 @@ Proof.
...
+
+(** An element [x] of a ring [R] is invertible if it is both left and right invertible. *)
+Class IsInvertible (R : Ring) (x : R) := Build_IsInvertible' {
+  isleftinvertible_isinvertible :: IsLeftInvertible R x;
+  isrightinvertible_isinvertible :: IsRightInvertible R x;
+}.
+
+Definition issig_IsInvertible {R : Ring} (x : R)
+  : _ <~> IsInvertible R x
+  := ltac:(issig).
```
jdchristensen, line 610: Maybe we should omit this?  It would be more natural to either unfold the inner records, or to represent it as a product, not a sigma type.  But since we no longer need it, we could just drop it.

theories/Algebra/Rings/Ring.v
```diff
@@ -572,3 +572,261 @@ Proof.
...
+Global Instance isinvertible_one {R} : IsInvertible R 1.
+Proof.
+  snrapply Build_IsInvertible.
+  - exact one.
+  - apply rng_mult_one_l.
+  - apply rng_mult_one_l.
+Defined.
+
+(** [-1] is always invertible. *)
+Global Instance isinvertible_neg_one {R} : IsInvertible R (-1) := _.
```
jdchristensen, line 804: Since this is found by typeclass search, maybe it can be omitted?

theories/Algebra/Rings/Ring.v
jdchristensen: I managed to figure out how to make this go through.  But if you prefer the more concrete proof, I can undo this last change.

Alizter: If you can push those changes that would be good. You can also get rid of the old issig lemma.

Alizter: Oh actually I see that you've pushed it.

Alizter:
@jdchristensen I've done the changes you suggested and will merge when this is green.

PR #2014, opened 2024-07-05
more on coproducts

Alizter:
In this PR, we add some more results about coproducts. This includes the following:
- Bifunctoriality results
- Products and coproducts in opposite categories
- The cocartesian monoidal structure
- And various miscellaneous results

### DIFF
--- a/theories/WildCat/Coproducts.v
+++ b/theories/WildCat/Coproducts.v
@@ -216,107 +216,107 @@ End Lemmata.
 (** Hint: Use [Set Printing Implicit] to see the implicit arguments in the following proofs. *)
 
 Global Instance is0functor_cat_bincoprod_l {A : Type}
-  `{H0 : HasBinaryCoproducts A} y
+  `{hbc : HasBinaryCoproducts A} y
   : Is0Functor (A:=A) (fun x => cat_bincoprod x y).
 Proof.
   rapply is0functor_op'.
-  exact (is0functor_cat_binprod_l (A:=A^op) (H0:=H0) y).
+  exact (is0functor_cat_binprod_l (A:=A^op) (H0:=hbc) y).
 Defined.
 
 Global Instance is1functor_cat_bincoprod_l {A : Type}
-  `{H0 : HasBinaryCoproducts A} y
+  `{hbc : HasBinaryCoproducts A} y
   : Is1Functor (fun x => cat_bincoprod x y).
 Proof.
   rapply is1functor_op'.
-  exact (is1functor_cat_binprod_l (A:=A^op) (H0:=H0) y).
+  exact (is1functor_cat_binprod_l (A:=A^op) (H0:=hbc) y).
 Defined.
 
 Global Instance is0functor_cat_bincoprod_r {A : Type}
-  `{H0 : HasBinaryCoproducts A} x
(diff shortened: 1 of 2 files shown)


theories/WildCat/Coproducts.v
```diff
@@ -211,59 +211,190 @@ Section Lemmata.
...
+(** *** Products and coproducts in the opposite category *)
+
+Definition hasbinarycoproducts_op_hasbinaryproducts {A : Type}
+  `{hbp : HasBinaryProducts A}
+  : HasBinaryCoproducts A^op
+  := hbp.
+Hint Immediate hasbinarycoproducts_op_hasbinaryproducts : typeclass_instances.
+
+Definition hasbinarycoproducts_hasbinaryproducts_op {A : Type}
+  `{HasEquivs A, hbp : !HasBinaryProducts A^op}
```
jdchristensen, line 275: Why is `HasEquivs A` needed here and 12 lines down?

jdchristensen (approved):
I inlined the one bit that was posed, and unified the naming of some arguments.

theories/WildCat/Coproducts.v
Alizter: Good point. I tracked down a few unneeded `HasEquivs` instances that actually stem from `Products.v`.

theories/WildCat/Coproducts.v
```diff
@@ -211,59 +211,190 @@ Section Lemmata.
...
+    $== cat_bincoprod_rec f (cat_bincoprod_rec g h).
+Proof.
+  nrapply cate_moveR_eV.
+  symmetry.
+  exact (cat_binprod_associator_corec
+           (HasBinaryProducts0:=hasbinaryproducts_op_hasbinarycoproducts (hbc:=hbc))
+           f g h).
+Defined.
+
+Definition cat_bincoprod_swap_rec {A : Type} `{HasEquivs A}
```
jdchristensen, line 374: I also noticed that this `HasEquivs` can be changed to `Is1Cat`.

theories/WildCat/Coproducts.v
```diff
@@ -211,59 +211,190 @@ Section Lemmata.
...
+      hasbinaryproducts_op_hasbinarycoproducts _ _ _ _ _.
+
+(** *** Cocartesian Monoidal Structure *)
+
+Global Instance ismonoidal_cat_bincoprod {A : Type} `{HasEquivs A}
+  `{!HasBinaryCoproducts A} (zero : A) `{!IsInitial zero}
+  : IsMonoidal A (fun x y => cat_bincoprod x y) zero.
+Proof.
+  nrapply ismonoidal_op'.
+  nrapply (ismonoidal_binprod (A:=A^op) zero).
```
jdchristensen, line 387: The coproduct version has `cat` in the name, but the product version does not.  We should make them match.  This also applies to `issymmetricmonoidal_cat_bincoprod` and `associator_cat_bincoprod`.  Since everything else has `cat`, I guess the right change is to add `cat` to the product versions.  But we could also consider whether `cat` is needed in all of these names...

theories/WildCat/Coproducts.v
Alizter: #2017

PR #2016, opened 2024-07-05
biproducts

Alizter:
In this PR we define what we mean by a category having biproducts. We prove standard lemmas, monoidal structure, funtoriality results and duality.

cc @ThomatoTomato you might want to take a look at this since it is directly part of the abelian category formalization in #1929.

### DIFF
--- a/theories/WildCat/Coproducts.v
+++ b/theories/WildCat/Coproducts.v
@@ -426,7 +426,18 @@ Instance hasbinarycoproducts_type : HasBinaryCoproducts Type
 
 (** ** Canonical coproduct-product map *)
 
-(** There is a canonical map from a coproduct to a product when the indexing set has decidable equality and the category is pointed. *)
+(** There is a canonical map from a coproduct to a product when the indexing set has decidable equality and the category is pointed.  We factor out the components of this map into a separate definition to make goals involving [cat_coprod_prod] easier to read. *)
+Definition cat_coprod_prod_component {I : Type} `{DecidablePaths I} {A : Type}
+  `{IsPointedCat A}
+  (x : I -> A) (i j : I)
+  : x i $-> x j.
+Proof.
+  destruct (dec_paths i j) as [p|].
+  - destruct p.
+    exact (Id _).
+  - exact zero_morphism.
+Defined.
+
 Definition cat_coprod_prod {I : Type} `{DecidablePaths I} {A : Type}
   `{Is1Cat A, !IsPointedCat A}
   (x : I -> A) `{!Coproduct I x, !Product I x}
@@ -435,11 +446,8 @@ Proof.
   apply cat_coprod_rec.
   intros i.
   apply cat_prod_corec.
-  intros a.
-  destruct (dec_paths i a) as [p|].
-  - destruct p.
-    exact (Id _).
-  - exact zero_morphism.
+  intros j.
+  exact (cat_coprod_prod_component x i j).
 Defined.
 
 Definition cat_bincoprod_binprod {A : Type} `{Is1Cat A, !IsPointedCat A}
--- a/theories/WildCat/Coproducts.v
+++ b/theories/WildCat/Coproducts.v
@@ -79,6 +79,11 @@ Section Lemmata.
   Definition cat_coprod_in_eta {z : A} {f g : cat_coprod I x $-> z}
     : (forall i, f $o cat_in i $== g $o cat_in i) -> f $== g
     := cat_prod_pr_eta I (A:=A^op) (x:=x).
+
+  (** A categorical equivalence out of a coproduct induces a coproduct structure on the codomain. *)
+  Definition cat_coprod_coprod_equiv `{!HasEquivs A} (y : A) (f : cat_coprod I x $<~> y)
+    : Coproduct I x
+    := cat_prod_equiv_prod _ (A:=A^op) (x:=x) _ f.
 End Lemmata.
 
 (** *** Codiagonal / fold map *)
@@ -346,7 +351,7 @@ Definition cat_bincoprod_fmap01_rec {A : Type}
 
 Definition cat_bincoprod_fmap10_rec {A : Type}
   `{Is1Cat A, !HasBinaryCoproducts A} {w x y z : A}
-  (f : y $-> x) (g : x $-> w) (h : z $-> w) 
+  (f : y $-> x) (g : x $-> w) (h : z $-> w)
   : cat_bincoprod_rec g h $o fmap10 (fun x y => cat_bincoprod x y) f z
     $== cat_bincoprod_rec (g $o f) h
   := @cat_binprod_fmap10_corec A^op _ _ _ _
@@ -425,7 +430,7 @@ Instance hasbinarycoproducts_type : HasBinaryCoproducts Type
 Definition cat_coprod_prod {I : Type} `{DecidablePaths I} {A : Type}
   `{Is1Cat A, !IsPointedCat A}
   (x : I -> A) `{!Coproduct I x, !Product I x}
-  : cat_coprod I x $-> cat_prod I x. 
+  : cat_coprod I x $-> cat_prod I x.
 Proof.
   apply cat_coprod_rec.
   intros i.
@@ -450,5 +455,5 @@ Definition coproduct_op {I A : Type} (x : I -> A)
   `{Is1Cat A} {H' : Product I x}
   : Coproduct I (A:=A^op) x
   := H'.
-  
+
 Hint Immediate coproduct_op : typeclass_instances.
--- a/theories/WildCat/Coproducts.v
+++ b/theories/WildCat/Coproducts.v
@@ -364,6 +364,48 @@ Definition hasbinaryproducts_hasbinarycoproducts_op {A : Type}
   := hbc.
 Hint Immediate hasbinaryproducts_hasbinarycoproducts_op : typeclass_instances.
 
+(** *** Lemmas about [cat_bincoprod_rec] *)
+
+Definition cat_bincoprod_fmap01_rec {A : Type}
+  `{Is1Cat A, !HasBinaryCoproducts A} {w x y z : A}
+  (f : z $-> w) (g : y $-> x) (h : x $-> w)
+  : cat_bincoprod_rec f h
+      $o fmap01 cat_bincoprod z g
+    $== cat_bincoprod_rec f (h $o g)
+  := @cat_binprod_fmap01_corec A^op _ _ _ _
+      hasbinaryproducts_op_hasbinarycoproducts _ _ _ _ f g h.
+
+Definition cat_bincoprod_fmap10_rec {A : Type}
+  `{Is1Cat A, !HasBinaryCoproducts A} {w x y z : A}
+  (f : y $-> x) (g : x $-> w) (h : z $-> w)
+  : cat_bincoprod_rec g h
+      $o fmap10 (fun x y => cat_bincoprod x y) f z
+    $== cat_bincoprod_rec (g $o f) h
+  := @cat_binprod_fmap10_corec A^op _ _ _ _
+      hasbinaryproducts_op_hasbinarycoproducts _ _ _ _ f g h.
+
+Definition cat_bincoprod_fmap11_rec {A : Type}
+  `{Is1Cat A, !HasBinaryCoproducts A} {v w x y z : A}
+  (f : y $-> w) (g : z $-> x) (h : w $-> v) (i : x $-> v)
+  : cat_bincoprod_rec h i
+      $o fmap11 cat_bincoprod f g
+    $== cat_bincoprod_rec (h $o f) (i $o g)
+  := @cat_binprod_fmap11_corec A^op _ _ _ _
+      hasbinaryproducts_op_hasbinarycoproducts _ _ _ _ _ f g h i.
+
+(** *** Codiagonal *)
+
+Definition cat_bincoprod_codiag {A : Type} `{Is1Cat A} (x : A)
+  (cat_bincoprod : A) {isbincoprod : IsBinaryCoproduct x x cat_bincoprod}
+  : cat_bincoprod $-> x
+  := @cat_binprod_diag _ _ _ _ _ _ _ isbincoprod.
+
+Definition cat_bincoprod_fmap11_codiag {A : Type}
+  `{HasBinaryCoproducts A} {x y : A} (f : x $-> y)
+  : f $o cat_bincoprod_codiag x _
+    $== cat_bincoprod_codiag y _ $o fmap11 (fun x y => cat_bincoprod x y) f f
+  := cat_binprod_fmap11_diag (A:= A^op) _.
+
 (** *** Symmetry of coproducts *)
 
 Definition cat_bincoprod_swap {A : Type} `{Is1Cat A}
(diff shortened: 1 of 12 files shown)


jdchristensen:
Some of the comments in #1929 that aren't marked as resolved are about the file Biproducts.v.  Have they been addressed here?  (I haven't had time to take a look at this PR yet, but will do so soon.)

Alizter:
@jdchristensen I don't remember. I will take a closer look ASAP.

Alizter:
I'll reproduce the outstanding comments here so we can address them in one place:

1. Regarding line 143: 
> I think this argument about how `cat_coprod_prod` computes when composed with the inclusions and projections has appeared more than once, so it should be factored out as a lemma. Then, with the change that only uses `r` once (using `catie_homotopic`), maybe there is no longer a need to assert the various pieces?

I'll write down and use these lemmas. They can probably appear in Coproducts.v.

I need to think a bit more about the rest of the comments.

2. Regarding line 907:
> I have a feeling that what you are doing here might overlap with `cat_biprod_corec_rec`. If `cat_biprod_corec_rec` is generalized to be about comparing two maps from a coproduct to a product, then maybe it exactly covers what you need here? Or maybe a slight variation gives a common generalization of both goals?

3. 
> In Biproducts.v, symmetry and associativity of biproducts is proved using that these hold for products. Then there is some work in showing that these maps respect the coproduct structure. I feel like this should be more formal. In a category with biproducts, every product is _canonically_ a biproduct (as is every coproduct). This is because the biproduct and the product both satisfy the universal property of the product. So the maps showing that products are associative and symmetric, defined using the product structure, should "automatically" respect the coproduct structure. I'm not sure how to make this more precise in a way that would make the proofs simpler, but I suspect that there should be a way.

Alizter:
@jdchristensen I haven't been able to make progress with the comments that we had, perhaps I no longer understand what needs to be done. Would you be able to take another look at this and give a fresh review, that should allow us to make some progress on this.

I would like to get this merged eventually, so that we can make room for the other abelian cat stuff.

jdchristensen:
@Alizter It will take me some time to remember the context here.  I might have time tomorrow, but if not, it will likely be next week.

jdchristensen:
I'm about to (finally) review this again.  Is it ok if I rebase this (or at least merge master into it) before I do so?

Alizter:
That's fine with me @jdchristensen

jdchristensen:
I squashed the previous commits into a single commit, adjusted the result so it builds with current master, and pushed that as a single commit.  Then I pushed small commits which make some minor fixes here and there.  A couple of small changes also improved many of the slow `Defined.` lines.

I spoke to @ThomatoTomato today, and we discussed ways that Biproducts.v could be adjusted to make many things smoother.  The main idea is to make the smart constructor (or something like it) the main definition of a biproduct.  In this version, instead of having separate coproduct and product objects, there is just a single object.  There will be instances showing that this object is both a coproduct and a product, so many of the lemmas about biproducts can be skipped, since we can just apply the results about (co)products.  It should also be trivial to show that this data gives a biproduct in the opposite category, and moreover this proof will likely be definitionally involutive.  What is currently the definition of Biproduct would instead be a different constructor.

A second thing that will likely be helpful is to add `IsProduct` and `IsCoproduct` which say when an object with projection/inclusion maps satisfies the universal property.  And there will be a lemma saying that if `X` along with some projections is a product and `Y -> X` is an equivalence, then `Y` with the composite maps is a product.  The dual of this would make the proof that the current definition of Biproduct implies the new definition trivial, and avoid some duplication.

(As an aside, I think the new definition can drop the `HasEquivs` assumption on `A`.)

@ThomatoTomato is willing to try to tackle this reorganization.  Is that ok with you @Alizter?

Alizter:
@jdchristensen  That sounds good to me.

jdchristensen:
@ThomatoTomato I think it's possible to get some of the benefits mentioned above without doing a major restructuring, so I think this makes a good first goal.  The idea is to (1) use the instance `biproduct_product` and drop all of the `cat_biprod_*` wrapper results about products, and (2) add an instance that shows that the product is also a coproduct, and drop all of the `cat_biprod_*` results about coproducts.  I've already done (1) as a test, and it worked;  I'll push it in a minute.  For (2), I think the right way to proceed is to add a result to Products.v that says that if `X` along with some projections is a product and `Y -> X` is an equivalence, then `Y` with the composite maps is a product.  Then add the dual result to Coproducts.v, and use it to create the instance using `cate_cat_prod_corec_inv`.

At this point, you wouldn't yet change the definition of `Biproduct` or add `IsProduct` and `IsCoproduct`, so it's a self-contained small change.

After that, we can decide on the next steps.

ThomatoTomato:
(2) has now been completed. In the first commit we add a result that let us transfer (co)product structures along equivalences in a category. Typeclass search tricks itself into looking for a different goal than what we want, so the second commit tweaks the priorities of typeclass search. In the final commit we remove wrappers for the coproduct. It was still not successfull at infering every parameter, so at some places we need to tell it what the diagram is.

Binary biproducts still keep making wrappers. So a next step would be to keep simplifying. However, it looks like adding `IsProduct`and `IsCoproduct` could help make this process easier, as typeclass search still gets itself into trouble at this point.

We've also added a couple of new tricks to the main library. Would it be wise to rebase this on master?

jdchristensen:
Thanks.  I'll take a look soon.  About rebasing, I guess `lhs'` and `rhs'` might be useful, but please wait until I review before rebasing.

Alizter:
I'll see if I can find some time this week to take a closer look.

jdchristensen:
@ThomatoTomato How close are you to finishing this?

ThomatoTomato:
Hello! I have finally been able to make some more progress on these files. It was difficult to make all the files to compile by simply changing the definitions to IsProduct and IsCoproduct. As these changes affected Biproducts.v, I also added an IsBiproduct in the same way. Now these three commits are split up in the following order:
1. Add IsProduct, IsCoproduct and IsBiproduct
2. Add IsBinaryProduct, IsBinaryCoproduct and IsBinaryBiproduct
3. Move some definitions out of Biproducts.v

You can see that I gave up on making the first pass fully compile, and some code is either admitted or commented out. This problem is fixed once the IsBinary... versions are added.

I did my best to try and keep track of everything that was done here, so that there are as few back and forth changes as possible, but I cannot promise that these commits are by any means perfect.

ThomatoTomato:
Now, I did notice that the old definition for Product and Coproduct did not use HasEquivs, even though we want to say that a map is an equivalence. Is this something that should be changed for the new IsProduct and IsCoproduct definitions?

Alizter:
> Now, I did notice that the old definition for Product and Coproduct did not use HasEquivs, even though we want to say that a map is an equivalence. Is this something that should be changed for the new IsProduct and IsCoproduct definitions?

Would you mind explaining the distinction a bit further? It has been a long time and I am not able to follow.

ThomatoTomato:
In the definition for IsProduct, we are specifying that a morphism is a categorical equivalence, so wouldn't it be natural to pass in HasEquiv as an argument to this class? I believe that doing so would give us more control over the class of equivalences. Right now it should be using the standard instance for HasEquiv that exists for any 1-category. I don't see anything wrong with that, but it seems to go against the practices for this library.

jdchristensen:
> In the definition for IsProduct, we are specifying that a morphism is a categorical equivalence, so wouldn't it be natural to pass in HasEquiv as an argument to this class?

No, the equivalence there is an equivalence in the wild category of 0-groupoids, not in the category `A`, so `A` does not need to have any notion of equivalence.

jdchristensen:
Can you rebase this on master?  I can't build it with Rocq 9.1, and I'm not sure if it works with 9.0 either.

ThomatoTomato:
Are you sure you want to rebuild this on master? It will add a couple of things to Products.v and mess up the history a little. This version is compatible with Coq 8.20, and you can build it with that.

jdchristensen:
It's got to happen eventually, so why not now?  Are the conflicts serious?

ThomatoTomato:
I have rebased the branch on master, and it builds with Rocq 9.1 on my computer.

Alizter:
@copilot Why do we have alectryon errors?

jdchristensen:
> @copilot Why do we have alectryon errors?

It doesn't look like anything related to this PR, unless it was an error in the rebase?  Or maybe just a random glitch?

```
Exiting early due to an error; use --traceback to diagnose:
  `sertop` binary not found (bin=None); please run `opam install coq-serapi`
Error: Process completed with exit code 1.
```

(BTW, I see "Copilot isn't available for cross-repository pull requests", but I'm not sure what that means.)

jdchristensen:
I restarted the alectryon job, so we'll see if the problem goes away.

etc/alectryon
jdchristensen, line 1: The alectryon build still fails.  It looks like the alectryon submodule has changes compared to master, so that probably needs to be fixed.  I don't know exactly what command will fix this, but a search/LLM can probably help.

Alizter:
@ThomatoTomato I've pushed reverting the accidental submodule change preserving everything else.

Alizter:
@jdchristensen Are we happy with this branch yet? Is this waiting on me to review?

jdchristensen:
@ThomatoTomato Can you remind us of the status?  Did I forget to review your most recent changes?

ThomatoTomato:
@jdchristensen we are indeed waiting for your review. Nothing has happened here in a month, so everything still needs to be reviewed. You and Ali fixed the problem that arose with Alectryon.

jdchristensen:
@Alizter @ThomatoTomato I'm not done yet, and will continue later.  One question for now:  I think it would make more sense to put Biproducts.v in the WildCat folder, alongside Coproducts and Products, since it is parallel to them.  Is that ok?

Alizter:
@jdchristensen Seems fine to me.

jdchristensen:
@Alizter Ok, I'll move it after I've finished working through some other fixes.

I have another question:  Products.v has instances (exported) for:
```
Associator cat_binprod
IsMonoidal A (fun x y => cat_binprod x y) unit
IsSymmetricMonoidal A (fun x y => cat_binprod x y) unit
```
Coproducts.v currently has instances for the duals of the last two, but the first is just a definition.  Biproducts.v has
```
IsSymmetricMonoidal A (fun x y => cat_binbiprod x y) zero_object
```
as a definition, and doesn't have the other two at all, either as definitions or instances.

So three different set-ups.  I'm wondering what is best.  It seems to me that IsSymmetricMonoidal has IsMonoidal as an instance (via `::`), and IsMonoidal has Associator as an instance, so maybe the right thing is to only have IsSymmetricMonoidal as an instance, rather than having redundant ways to fill the same class?  So in both Products.v and Coproducts.v, I would make the first two either plain definitions or Local Instances.

For Biproducts, if we make IsSymmetricMonoidal an instance, is there any chance of a conflict between the instances, since biproducts are also products (and are also coproducts)?

Similarly, and maybe even worse, if we are in a category with both coproducts and products (but not necessarily the same), there are two symmetric monoidal structures.  Will having both as instances cause a problem?

I'm guessing none of these will be an issue, but I wondered if this line of thought was why in Biproducts no instances were declared.

Alizter:
I think it's less likely that instances, due to how they are set up, will conflict due to the changing head terms that key the instance. For instance, we use cat_binbiprod and cat_binprod separately. Even though these are the same under the hood, but type class search shouldn't be exposing this.

The symmetric monoid structure for biproducts is unique up to monoidal products, so for that product and unit it shouldn't matter, but might be annoying if two are defined, I.e. one represented with coproducts rather than products. In general, you are correct, we shouldn't make such things instances unless they are vaguely unique or canonical.

I think if we were working in a context where the actual monoidal structure was sensative however we should stead use a bundled structure anyway.

jdchristensen:
So what do you think we should actually do here?  Make none instances?  Or make just IsSymmetricMonoidal an instance in all three files?  Even in a category with biproducts, there will often be another symmetric monoidal structure.  E.g. for abelian groups, tensor product is symmetric monoidal.  But, as you said, these classes are keyed by the functor, so I guess that's not a problem?  So maybe I'll make (just) IsSymmetricMonoidal an instance in all three files?

Alizter:
Yes, that seems the most sensible. I would remove the redundant instances you've found with cat_binprod and rely on the projections from IsSynmetricMonoidal.

jdchristensen:
It turns out that there are a few subtleties.  For some reason, Rocq sometimes fails to find Associator instances when IsMonoidal can be found.  I wonder if this is because Associator doesn't depend on the unit, but IsMonoidal does?  In some very naive tests, Rocq is able to guess the unit based on the IsMonoidal Instance it finds, but in Products.v, the following fails when placed right after the Associativity Section:

```
Section Test.
  Context {A : Type} `{HasEquivs A} `{!HasBinaryProducts A}.
  Context (unit : A) `{!IsTerminal unit}.

  (** This is inferred from issymmetricmonoidal_cat_binprod. *)
  Instance test1 : IsMonoidal A cat_binprod unit := _.

  Typeclasses eauto := debug 2.
  Definition test2 : Associator cat_binprod.
  Proof.
    Fail exact _.
    (* The above doesn't even try cat_tensor_associator. *)
    (* And if we try it directly, it fails: *)
    Fail simple apply @cat_tensor_associator.
    (* But this works. *)
    exact cat_tensor_associator.
  Defined.
End Test.
```

So I made the Associator into an instance, and took a few guesses and made some other things into instances.  When this material is used in practice, we'll find out what other instances are needed and maybe figure out how to get the typeclass search to find these more consistently via IsMonoidal.

PS: I've still got a few more things I want to clean up.  I'll let you know when it's ready for review.

jdchristensen:
Ok, I think I'm finally done.  I made so many commits that it will be hard to review them, except to just skim the commit messages to see if anything stands out.  Then I would just look at the overall diff compared to master to see if things look good.  @Alizter, @ThomatoTomato:  it would be good if you both could review.

One recent commit deals with `isgraph_op`, adding a `Hint Extern` for it.  This could use a quick check.  The new comments explain the motivation, and there are tests.

Long ago I added `decidable_hprop_true` and `decidable_hprop_false` to Basics/Decidable.v, and used them in this PR.  But with the restructuring, they were no longer needed.  I left them since they might be useful, but let me know if you think they should be deleted.

I compared the timing to master.  The new Biproducts.v takes 0.67s.  Products.v and MonoidObject.vo take about 0.08s longer each.  Coproducts.v takes about 0.03s longer.  The rest of the library takes about 0.8s longer in total, but no file increased by more than 0.04s.  The slowest lines in Biproducts.v and Products.v are ending the Associativity sections, taking about 10% of the whole file's time in both cases.  We could get rid of the sections at the cost of repeating the args many times.  It's not a lot of time, so probably not worth it.

When this is ready to merge, maybe we should squash the commits down to one with all three of us (and Claude) as co-authors?

Alizter:
@jdchristensen I read through the final diff and it LGTM. Feel free to squash and merge.

Thanks a lot for working on this, it's exciting to be able to make good progress.

jdchristensen:
I'll also wait to see if @ThomatoTomato has any feedback before squashing and merging.

theories/WildCat/Coproducts.v
```diff
@@ -7,248 +7,317 @@ Require Import WildCat.Core WildCat.Equiv WildCat.Forall WildCat.NatTrans
...
+
+  (** The induced inclusion is given by the equivalence. *)
+  Definition cat_in_comp (i : I)
+    : cat_in i $== f $o cat_in i
+    := cat_pr_comp cat_coprod y _ i.
+
+  (** The induced recursion is given by the equivalence. *)
+  Definition cat_coprod_rec_comp {z : A} (D : forall i, x i $-> z)
+    : cat_coprod_rec y D $o f $== cat_coprod_rec cat_coprod D
+    := cat_prod_corec_comp (A:=A^op) (H0:=hasequivs_op) cat_coprod y f D.
```
ThomatoTomato: On this line, and around 12 other lines in this file, we use the named field `H0` to specify the equivalences in `A`. Is it good enough practice to use the name that Rocq provides for us, or do we want to go back to Products.v and name this field ourselves?

ThomatoTomato:
This looks good to me as well! I only have one comment/question regarding the use of names provided by Rocq. After this is resolved, I will be happy to merge the way you see best fit.

theories/WildCat/Coproducts.v
jdchristensen: Thanks, good catch.  Fixed and pushed as a separate commit in case anyone wants to look at it.  Next I'll squash and force push.

Alizter: Rocq has support for name mangling which can locally help detect these kinds of uses.

Naming the field is always preferable so that future rocq changes don't break proofs.

jdchristensen:
I pushed five commits, and the library builds after each one.  I'll merge when the CI completes.

PR #2018, opened 2024-07-05
fix naming of monoidal instances in (co)products.v

Alizter:


jdchristensen (commented):
Also `associator_binprod`.

Alizter:
@jdchristensen Have I been a bit overzealous with the renaming?

theories/WildCat/Products.v
```diff
@@ -894,54 +894,54 @@ Section Associativity.
     intros a b c d.
     nrapply cat_binprod_eta_pr_xx_x.
     - nrefine (cat_assoc_opp _ _ _ $@ (_ $@R _) $@ _).
-      1: nrapply cat_pr1_pr1_associator_binprod.
+      1: nrapply cat_pr1_pr1_associator_cat_binprod.
```
jdchristensen, line 897: Since we allow things like `associator_cat_binprod` without a leading `cat`, maybe `pr1_pr1_associator_cat_binprod` is more appropriate here (and with other similar names that have `cat` duplicated)?

Alizter: Yes, this is probably better. I've reverted the names.

jdchristensen: My suggestion wasn't to revert, but to use `pr1_pr1_associator_cat_binprod`.  But now that I think about it, `cat_pr1` appears in the statement of the result too, so maybe just keeping `cat` at the front like you've now done is a good choice.  (Really it should have three `cat`s in it!)

So, this LGTM.

PR #2020, opened 2024-07-08
free abelian groups

Alizter:
In this PR we:
- Characterise the universal property of the free abelian group on a type S.
- Show that an abelianization of a free group produces a free abelian group.
- As a special case, construct the free abelian group on a type S as the abelianization of the free group on the set.

This construction will be important in the theory of modules as we will be taking quotients of free abelian groups to produce constructions such as tensor products of abelian groups.

I had a bit of trouble managing funext. I tried to see if it was possible to stay funext free but I didn't see a way.

Classically, the free abelian group is defined as a fucntion from the generating set into Z together with a finite support. This definition is very appealing for us but would also require a lot of funext in order to equate terms. It is also not clear how to constructively reason about finite supports as their definition contains an awkward negation that means basic lemmas require decidablility or LEM in general.

### DIFF
--- a/theories/Algebra/AbGroups/Abelianization.v
+++ b/theories/Algebra/AbGroups/Abelianization.v
@@ -23,6 +23,22 @@ Class IsAbelianization {G : Group} (G_ab : AbGroup)
       IsSurjInj (group_precomp A eta).
 Global Existing Instance issurjinj_isabel.
 
+Definition isequiv_group_precomp_isabelianization `{Funext}
+  {G : Group} {G_ab : AbGroup} (eta : GroupHomomorphism G G_ab)
+  `{!IsAbelianization G_ab eta} (A : AbGroup)
+  : IsEquiv (group_precomp A eta).
+Proof.
+  snrapply isequiv_adjointify.
+  - intros g.
+    rapply (surjinj_inv (group_precomp A eta) g).
+  - intros f.
+    snrapply equiv_path_grouphomomorphism.
+    exact (eisretr0gpd_inv (group_precomp A eta) f).
+  - intros f.
+    snrapply equiv_path_grouphomomorphism.
+    exact (eissect0gpd_inv (group_precomp A eta) f).
+Defined.
+
 (** Here we define abelianization as a HIT. Specifically as a set-coequalizer of the following two maps: (a, b, c) |-> a (b c) and (a, b, c) |-> a (c b).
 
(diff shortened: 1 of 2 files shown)


Alizter:
I've noticed that in `FreeGroup.v` there are comments saying we should remove funext from the universal property. We should probably do this before this PR so that we can introduce a fuent free version of the universal property of the free abelian group here.

theories/Algebra/AbGroups/Abelianization.v
```diff
@@ -23,6 +23,22 @@ Class IsAbelianization {G : Group} (G_ab : AbGroup)
...
+  snrapply isequiv_adjointify.
+  - intros g.
+    rapply (surjinj_inv (group_precomp A eta) g).
+  - intros f.
+    snrapply equiv_path_grouphomomorphism.
+    exact (eisretr0gpd_inv (group_precomp A eta) f).
+  - intros f.
+    snrapply equiv_path_grouphomomorphism.
+    exact (eissect0gpd_inv (group_precomp A eta) f).
+Defined.
```
jdchristensen, line 40: The proof of this looks like it can be abstracted out as a general lemma.  Maybe it would say something like:  If A is a wild category with MorExt and `f : Hom a b -> Hom c d` is a map of zero groupoids that satisfies `IsSurjInj`, then it is an equivalence on the underlying types?

But really it should be a bit more general.  We could say that a zero groupoid is "extensional" if the 1-cells are equivalent to the paths in the underlying type.  Then the lemma would say that if G and H are extensional zero groupoids and `f : G -> H` is a map of zero groupoids that satisfies `IsSurjInj`, then it is an equivalence on the underlying types.  And we'd have a typeclass instance that says that if A is a wild category that satisfies MorExt, then its Hom zero groupoids are extensional.

(I'm not sure if "extensional" is the right word.  It's vague.)

jdchristensen:
This PR is fairly short and sweet (and I just pushed a commit making it a bit shorter), so I'm ok with merging it and leaving the cleanup of Funext for a future PR.  My comment above could also be left as a TODO item in the file, or could be handled.  I'm fine either way.

theories/Algebra/AbGroups/Abelianization.v
Alizter: Alternatively, we wouldn't need this at all if we removed funext from the free group universal property. I haven't had a look at that in detail yet.

Issue #2015, opened 2024-07-05
grp_pow and related things


jdchristensen:
Based on a comment I made in in https://github.com/HoTT/Coq-HoTT/issues/2000#issuecomment-2198748547_

I think `issemigrouppreserving_mult_rng_int_mult` can be proven by combining some pieces that will be useful on their own.  See below for a sketch.

Group.v:

```coq
(** [grp_pow g n] commutes with [g].  Actually, for the inductive step below, we might need to know that if [g] commutes with [h], then [g] commutes with powers of [h]? *)
Definition grp_pow_commutes {G : Group} (n : Int) (g : G)
  : (grp_pow g n) * g = g * (grp_pow g n).

(** If [g] and [h] commute, then [grp_pow (g * h) n] = (grp_pow g n) * (grp_pow h n)]. *)
(* Note that part of the proof that [ab_mul] is a homomorphism will be covered by this. *)
Definition grp_pow_mul {G : Group} (n : Int) (g h : G)
  (c : g * h = h * g)
  : grp_pow (g * h) n] = (grp_pow g n) * (grp_pow h n).

(** [grp_pow] satisfies a multiplicative law of exponents. *)
Definition grp_pow_int_mul {G : Group} (m n : Int) (g : G)
  : grp_pow g (m * n)%int = grp_pow (grp_pow g m) n.
(* This will follow from the previous two. *)
```

Rings/Z.v:

```coq
Definition rng_int_mult_foo {R : Ring} (r : R) (n : Int)
  : rng_int_mult r n = (rng_int_mult 1 n) * r.

(* I think issemigrouppreserving_mult_rng_int_mult will follow from the previous item and grp_pow_int_mul. *)
```

cc: @ThomatoTomato

ndcroos:
For Group.v I have this at the moment:

```coq
(** [grp_pow g n] commutes with [g].*)
Definition grp_pow_commutes'' {G : Group} (n : Int) (g : G)
  : (grp_pow g n) * g = g * (grp_pow g n).
Proof.
  symmetry.
  by apply grp_pow_commutes.
Defined.

(** If [g] and [h] commute, then [grp_pow (g * h) n] = (grp_pow g n) * (grp_pow h n)]. *)
Definition grp_pow_mul {G : Group} (n : Int) (g h : G)
  (c : g * h = h * g)
  : grp_pow (g * h) n = (grp_pow g n) * (grp_pow h n).
Proof.
  induction n.
  - simpl.
[...]

PR #2021, opened 2024-07-09
tensor products of abelian groups

Alizter:
This PR constructs the tensor product of abelian groups. We show that tensor products give a symmetric monoidal structure to `AbGroup` using the twist construction for symmetric monoidal categories.

This PR depends on:
- [x] #2020 
- [x] #2015 

So it would be good to finish those before. In the meantime this PR can be read. There are some issues I would like to address:
- [ ] `ab_tensor_prod` and its constructor `tensor` are slow to work with. If you have a goal involving them and you do a `cbn` or `simpl`, Coq will hang. We should think carefully about helping Coq to not unfold certain parts to make this reasonable to work with.
- [x] The file `TensorProduct.v` needs comments.
- [x] Some of the `grp_pow` lemmas should follow from the lemmas outlined in #2015.
- [x] There is some duplication between the different induction principles for `ab_tensor_prod`. We need to think how to share more here.

cc @ThomatoTomato as this is something you are presumably interested in.


theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -0,0 +1,840 @@
...
+  - refine (_ - _ - _ = x).
+    1-3: apply freeabgroup_in.
+    + exact (a1 + a2, b).
+    + exact (a1, b).
+    + exact (a2, b).
+  - refine (_ - _ - _ = x).
+    1-3: apply freeabgroup_in.
+    + exact (a, b1 + b2).
+    + exact (a, b1).
+    + exact (a, b2).
```
jdchristensen, line 39: I guess we should have a version of `subgroup_generated` that lets you provide a family `gen : I -> A` of elements, and constructs the predicate `fun x => exists i, gen i = x` for you.  Also, I think `(a1 + a2, b) - ((a1,b) + (a2,b))` would be the more natural parenthesization.  And it might make tensor_dist_* a bit easier. And ab_tensor_prod_rec.  In fact, we should probably generalize this to an abelian group with generators and relations, with relations given by two functions `I -> A` representing the left and right hand sides of the relations, and a recursion principle in this generality.  It should be cleaner.

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -0,0 +1,840 @@
...
+  := fun a b => grp_quotient_map (freeabgroup_in (a, b)).
+
+(** ** Properties of tensors *)
+
+(** The characterizing property of elementary tensors are that they are bilinear in their arguments. *)
+
+(** A [tensor] of a sum distributes over the sum on the left. *) 
+Definition tensor_dist_l {A B : AbGroup} (a : A) (b b' : B)
+  : tensor a (b + b') = tensor a b + tensor a b'.
+Proof.
```
jdchristensen, line 57: The goal looks pretty terrible in these proofs.  Can we improve that somehow?

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -0,0 +1,840 @@
...
+
+The hprop induction principle therefore requires that the codomain is a hprop in addition to being closed under the various group operations of the domain. Most importantly we hypothsise that we can prove a property on an elementary tensor [tensor a b]. *)
+Definition ab_tensor_prod_ind_hprop {A B : AbGroup}
+  (P : ab_tensor_prod A B -> Type)
+  {H : forall x, IsHProp (P x)}
+  (H1 : P 0)
+  (H2 : forall a b, P (tensor a b))
+  (H3 : forall x, P x -> P (-x))
+  (H4 : forall x y, P x -> P y -> P (x + y))
+  : forall x, P x.
```
jdchristensen, line 229: I don't think you should need H1 or H3.  You can get the negative of a tensor with tensor (-a) b, and you can get 0 by adding tensor a b to tensor (-a) b.  But if that makes the proof too complicated (especially omitting H3), it's ok to keep them.

jdchristensen (commented):
This is looking good so far!

jdchristensen:
I like that the twist construction worked well here.  I suspect that you can also prove symmetry and the twist using a Yoneda argument, using the universal property and properties of biadditive maps.  But the way you've done it is pretty straightforward.

theories/Algebra/AbGroups/TensorProduct.v
jdchristensen: Actually, you can get zero as `tensor 0 0`, so I've gone ahead and done this.  Getting rid of H3 might take a bit more work.

Alizter:
When we later characterize the universal property as a right adjoint to the internal hom, the Yoneda lemma will play a crucial role in proving things like distributivity over the direct sum.

theories/Algebra/AbGroups/Abelianization.v
```diff
@@ -275,6 +275,19 @@ Section AbelGroup.
...
+    split.
+    1: exact _.
+    reflexivity.
+  Defined.
+  
+  (** [ab] preserves negation *)
+  Definition ab_negate x : ab (-x) = - ab x.
+  Proof.
+    exact (preserves_negate (f:=ab) x).
+  Defined.
```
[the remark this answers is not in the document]
Alizter: Yes, `ab` here actually clashes with `ab_comm` stating commutativity in an abelian group so it would be good to revise the names here and define it directly as a group homomorphism. I'll see about doing that in another PR.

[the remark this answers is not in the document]
Alizter: We can't define `ab` to be a group homomorphism as the group structure is based on `ab`, however we do infact define the group homomorphism I was looking for before, later in the file. I've gone ahead and renamed `ab` to `abel_in` and I'll derive the negation result where I need it using the grp_homo lemma.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: Yes, I think it would be better if we made a custom induction principle for QuotientGroup.

theories/Algebra/AbGroups/Abelianization.v
```diff
@@ -270,29 +278,23 @@ Section AbelGroup.
...
-  (** By definition, the map ab is also a group homomorphism. *)
-  Global Instance issemigrouppreserving_ab : IsSemiGroupPreserving ab.
+  (** By definition, the map [abel_in] is also a group homomorphism. *)
+  Global Instance issemigrouppreserving_abel_in : IsSemiGroupPreserving abel_in.
   Proof.
     by unfold IsSemiGroupPreserving.
   Defined.
    
-  Global Instance ismonoidpreserving_ab : IsMonoidPreserving ab.
+  Global Instance ismonoidpreserving_abel_in : IsMonoidPreserving abel_in.
```
jdchristensen, line 287: Maybe we don't need this result either?

Alizter: Yes, this doesn't seem useful anymore.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: I've been struggling to prove H3 isn't needed. Have you got an idea how to make it go through?

theories/Algebra/AbGroups/TensorProduct.v
Alizter: I've made a custom induction principle that assume `grp_quotient_map` in the hypotheses which makes it nicer.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: Looks like we already have this for subgroups of free groups. I'll see if I can use it.

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -223,27 +223,21 @@ Definition ab_tensor_prod_ind_hprop {A B : AbGroup}
   (P : ab_tensor_prod A B -> Type)
   {H : forall x, IsHProp (P x)}
   (H2 : forall a b, P (tensor a b))
-  (H3 : forall x, P x -> P (-x))
-  (H4 : forall x y, P x -> P y -> P (x + y))
+  (H4 : forall x y, P x -> P y -> P (x - y))
```
jdchristensen, line 226: If we're going to keep negation, I'm not convinced that it's better to combine addition and negation into one hypothesis.  It seems to make a few things easier but a few other things harder.  But if you prefer it, that's fine.  (I'll comment elsewhere about why I think addition alone suffices.)

Alizter: I've combined them for now, but I also agree that it ought to be possible to get by with only addition. The reason I joined them is to mimic the subgroup criterion.

theories/Algebra/AbGroups/TensorProduct.v
jdchristensen: The rough idea is that we know that the simple tensors `tensor a b` generate the tensor product, in the sense that every element of the tensor product is a sum of these and their inverses.  But since the inverse of `tensor a b` is also a simple tensor, the simple tensors generate the group as a monoid.  So to prove something about all elements, it should be enough to prove it for the simple tensors and show that the property is closed under the operation.

Put another way, if you drop the zero and negation parts of the induction principle, it really is the statement that the tensor product is generated as a monoid by the simple tensors.

Maybe we could add a general lemma that says that the subgroup generated by a subset is equal to the monoid generated by the subset when the subset is closed under negation and contains zero?

Alizter: I've worked out how to prove it now.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: I've managed to work out how to prove it only for addition so I've gone ahead and done that. It's now much clearer to see how a dependent biadditive map should look like.

jdchristensen: Nice!  I looked at what you did and realized that your subgoal was a slightly harder version of the overall goal, so I could delete the earlier part of the proof and just use your method instead.  Pushed.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: What I said earlier helps in some other places but not here. I'm not particularly sure what the best thing to do here is. I've simplified the proofs here after using your suggestion of `x - (y + z)` from earlier, but it still looks kind of ugly.

Alizter:
The pentagon takes 20 universes or so which isn't ideal. I've been trying to chase down the source of the issue but it seems to be quite systemic. I think basically any time we use `$->` instead of `GroupHomomorphism` we end up introducing a strictly larger universe for the wildcat `Group`/`AbGroup` which can be a bit annoying.

theories/Algebra/AbGroups/TensorProduct.v
jdchristensen: I just pushed a change that makes these goals clean.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: Could you expand on this idea a bit further? What would `gen` be in this case?

jdchristensen:
> The pentagon takes 20 universes or so which isn't ideal. I've been trying to chase down the source of the issue but it seems to be quite systemic throughout the groups library. I think basically any time we use `$->` instead of `GroupHomomorphism` we end up introducing a strictly larger universe for the wildcat `Group`/`AbGroup` which can be a bit annoying.

I'm hoping that algebraic universes get merged into Coq soon, which I think will get rid of this problem.  We'll be able to use `u+1` as the universe that is currently free.  So it's probably find to not worry about this too much here.  (And 20 isn't too bad.)

https://github.com/coq/coq/pull/18903

theories/Algebra/AbGroups/TensorProduct.v
jdchristensen: In my first sentence, we could take `I` to be `A * A * B + A * B * B` and `gen` would send `(a1,a2,b)` to `(a1+a2,b) - ((a1,b) + (a2,b))` and act similarly on the `A * B * B` summand.

In my last sentence, `I` would stay the same, and we'd have `lhs` sending `(a1,a2,b)` to `(a1+a1,b)` and `rhs` sending `(a1,a2,b)` to `(a1,b) + (a2,b)` (and both would act similarly on the other summand).  Then the induction principle would involve an equation, rather than a difference, which might eliminate some of the manipulations needed.

I don't think either thing I'm suggesting saves very much, though, so maybe it's not worth changing.

Alizter:
@jdchristensen I proved one of the lemmas in #2015 here (in a commit near the top). Is this the proof you had in mind for that issue? If so, I will check it off my checklist.

Alizter:
> I like that the twist construction worked well here.

This might be a superficial success however. If we write down the action of `associator` on `tensor a (tensor b c)` then it might be possible to just rewrite using that computation rule.

Alizter:
> > Unfortunately, it seems that the 9-gon isn't yet useful.
> 
> Just to make sure I understand: the proof of the pentagon in the current version of this PR is using the 9-gon involving the swap and the twist. The code you posted in the comment above shows that you can also prove the pentagon directly. Both are about a similar level of difficulty (not that difficult at all), so the 9-gon seems to work equally well as the more obvious proof. But we hope that with some other monoidal structure, we might find the 9-gon to be an easier way to prove it.

Yes, that is what I was getting at. My confidence in the 9-gon being simpler than the pentagon is not very high at the moment however.

jdchristensen:
> I proved one of the lemmas in #2015 here (in a commit near the top). Is this the proof you had in mind for that issue? If so, I will check it off my checklist.

I factored out the parts about commuting elements into new lemmas.  Other than that, this part looks good.  (There are still things in #2015 that can be done, but they don't have to be part of this PR.)

While I was at it, I cleaned up some other things.  To fix notation, recall that for `n : Int` and `g` in a group `G`, `grp_pow g : Int -> G` is a homomorphism, defined as `grp_pow_homo`.  On the other hand, when `G` is abelian, `ab_mul n : G -> G` is a homomorphism.  Both of these have `grp_pow` as the underlying map, so which you use depends upon which form of the homomorphism you are considering.  In light of this:

- `grp_pow_plus` was redundant, since we already have `ab_mul`.  (Also, the proof of the homomorphism property in `ab_mul` is shorter, probably because of the helper lemma.)
- I generalized `abgroup_Z_grp_pow_1` to `abgroup_Z_ab_mul` which shows that `ab_mul` is the same as `int_mul` when `G` is the integers.
- `tensor_grp_pow_l/r` are just special cases of the naturality of `ab_mul`, so they are both one-liners.

Alizter:
@jdchristensen Thanks! This looks good. The only remaining issue for this PR is working out why `cbn` is slow in the presence of `tensor`. If we can't find the issue however it is not a big deal, but we need to be careful when working with tensors so as to not do `cbn` as Coq getting stuck is a bit annoying.

jdchristensen:
> The only remaining issue for this PR is working out why `cbn` is slow in the presence of `tensor`.

Have you tried making various things opaque?

> Oh and also there is the issue of making the use of `subgroup_generated` slicker, but I don't know if it is needed for this PR or something we could consider in the future.

It's not necessary for this PR.

Alizter:
@jdchristensen Yes, but I've had no luck so far. I'll try some more.

Alizter:
I think I've found the culprit. The main issue was unfold `grp_homo_ishomo`. We can make this opaque but a better thing to do is to redefine `ab_tensor_prod_rec` using `Build_GroupHomomorphism`. This encapsulates the proof a bit making `simpl` and `cbn` unfold it less. Maybe there is a better way to get this, but for now this seems better.

jdchristensen:
> The main issue was unfold `grp_homo_ishomo`. We can make this opaque but a better thing to do is to redefine `ab_tensor_prod_rec` using `Build_GroupHomomorphism`. This encapsulates the proof a bit making `simpl` and `cbn` unfold it less. Maybe there is a better way to get this, but for now this seems better.

I don't like or understand the last commit.  In the proof, you construct a term that I'll call `f` whose type is a group homomorphism, and then you build a group homomorphism as `Build_GroupHomomorphism f _`, where `_` is inferred since `f` is a group homomorphism already.  This already seems weird.  But in fact, `f` is defined using `grp_quotient_rec`, and the head term of that is already `Build_GroupHomomorphism`, so what we've now got is `Build_GroupHomomorphism (Build_GroupHomomorphism foo bar) _` which seems pretty silly.  Can you give a concrete example of what improves before and after this change?  Is there another way we can achieve it?

jdchristensen:
I don't know if it's related, but I've wondered if we should remove `IsUnitPreserving` from the record that defines a group homomorphism, since it can be deduced from the map preserving the operation.  (And in fact that's almost always how we construct examples.)  There are some cases where Coq tries to unfold that proof and it slows things down.

Alizter:
I will revert that commit as I noticed it doesn't actually affect the things I claimed it did. I guess I jumped to a conclusion a little fast there.

Alizter:
Removing IsUnitPreserving seems like a sensible thing to do. I guess my original intention was for groups to be monoids with extra structure, but I don't think the category of monoids is interesting enough to restrict ourselves this way. I'll do that in another PR.

jdchristensen:
> Removing IsUnitPreserving seems like a sensible thing to do. I guess my original intention was for groups to be monoids with extra structure, but I don't think the category of monoids is interesting enough to restrict ourselves this way. I'll do that in another PR.

The reason I haven't tried it is that the Classes library is designed with groups being special kinds of monoids, so we might end up having trouble with the results we use from there.  So it's not completely clear if this change is worth it.

Another approach I considered is to factor out the proof that preserving the operation implies preserving the unit, and make that lemma opaque (which can be toggled if we need it to compute in some cases).  E.g.
```coq
(** A map between groups that preserves the operation, automatically preserves the identity. *)
Definition isunitpreserving_issemigrouppreserving {G H : Group}
  (f : G -> H) {h : IsSemiGroupPreserving f}
  : IsUnitPreserving f.
Proof.
  unfold IsUnitPreserving.
  apply (group_cancelL (f mon_unit)).
  refine (_ @ (grp_unit_r _)^).
  refine (_ @ ap _ (monoid_left_id _ mon_unit)).
  symmetry.
  apply h.
Defined.
Opaque isunitpreserving_issemigrouppreserving.

(** When building a group homomorphism we only need that it preserves the group operation, since we can prove that the identity is preserved. *)
Definition Build_GroupHomomorphism {G H : Group}
  (f : G -> H) {h : IsSemiGroupPreserving f}
  : GroupHomomorphism G H.
Proof.
  srapply (Build_GroupHomomorphism' _ _ f).
  split.
  1: exact h.
  apply isunitpreserving_issemigrouppreserving.
  exact h.
Defined.
```
I've thought about this because there was a particular slowdown a few months ago that I'd like to fix, but I haven't got around to working out the best approach.

Alizter:
@jdchristensen It seems to work fine: https://github.com/HoTT/Coq-HoTT/pull/2029

Alizter:
When #2029 is merged we will need to fix this PR as we made some arguments to `grp_homo_map` implicit.

Alizter:
Things seem to have improved with regards to `simpl` and `cbn`. The only time I see a hang is if you do `cbn`/`simpl` in the beta rule for `ab_tensor_prod_twist`. The beta rules that cam before all reduce nicely which wasn't the case before.

jdchristensen:
> Things seem to have improved with regards to `simpl` and `cbn`. The only time I see a hang is if you do `cbn`/`simpl` in the beta rule for `ab_tensor_prod_twist`. The beta rules that came before all reduce nicely which wasn't the case before.

Does that mean that some of the `change` lines can be changed to `cbn` or `simpl`?

Doing `cbn` at the start of the proof of `ab_tensor_prod_twist` produces a goal around 2000 lines long on my machine, and takes 17s.  Not terrible, but maybe there is still something that can be done to improve this?

jdchristensen:
Ah, I think you mean `ab_tensor_prod_twist_beta_tensor_tensor`.  There, `cbn` didn't complete after trying for 10 minutes.  In fact, after doing `unfold ab_tensor_prod_twist.`, even `unfold ab_tensor_prod_rec.` was very slow, taking around 5 minutes and producing  a goal around 4000 lines long.  I wonder if we could make part of `ab_tensor_prod_rec` opaque, e.g. by factoring out part of the proof?   ....    Yes, I've done this, and it makes that second `unfold` fast.  Next is `grp_quotient_rec`.  Again, maybe the part that is checking that the equivalence relation is respected should be made opaque?  (Or maybe both of these should be marked as `simpl never` or something like that?)

While I was looking at this, I removed a couple of rewrites from `grp_quotient_rec`, as I saw that the term produced was a bit large.

Alizter:
@jdchristensen Did you mean to push any changes?

jdchristensen:
Ah, my push failed because I hadn't yet pulled, and I didn't see the error message.  Now I've rebased and pushed.

Alizter:
I've made some changes to how the twist map is defined by breaking it into pieces and removing rewrites. `cbn` still has trouble unfolding the beta rule, but this time `simpl` goes through without any issue.

jdchristensen:
> `cbn` still has trouble unfolding the beta rule, but this time `simpl` goes through without any issue.

I couldn't find a place where `simpl` works better after this PR.  Here's what I tried:
```coq
Definition ab_tensor_prod_twist_beta_tensor_tensor {A B C} a b c
  : @ab_tensor_prod_twist A B C (tensor a (tensor b c)) = tensor b (tensor a c).
Proof.
  simpl. (* Very fast, but leaves us with ab_tensor_prod_twist_map. *)
  unfold ab_tensor_prod_twist_map.
  simpl. (* Very slow. *)
```

Alizter:
I've pushed a commit allowing simpl to unfold `ab_tensor_prod_twist_ap`

This works for me, although it is not as simple as I claimed before:
```coq
  do 2 (simpl; unfold uncurry, fst, snd).
```

maybe we should add simpl hints for `uncurry`, `fst` and `snd` but that might be problematic if anything relies on it not unfolding at the moment.

jdchristensen:
It might be a good idea to have `uncurry` unfold when given enough arguments.  `fst` and `snd` already do unfold with `simpl` when given a constructor, so I think we're good there.  You could try adding a hint for `uncurry`, and seeing if the build succeeds and how the time compares.  I recall having to add some `unfold uncurry` commands elsewhere recently that could then be omitted.

I'm still not sure why `cbn` is so slow here.  Even after your commands get the goal looking nice, `cbn` takes forever.  And `Opaque tensor.` doesn't help.  Oh, here's something interesting.  With just your `simpl` and `unfold` commands, the "small" term
```coq
 tensor b (tensor a c) + mon_unit + mon_unit = tensor b (tensor a c)
```
that results is actually 9000 lines long with `Set Printing All.`  So even at this point there's a huge amount of unfolded context that should be left folded.  We probably need to make more things opaque or mark them with `simpl never`.

Alizter:
I'm working on a patch that will get rid of the `mon_unit` we get from `FreeGroup_rec` reducing. It will save us having to do quite a few rewrites actually.

Alizter:
I've fixed the slow inference by doing a `change` rather than a `simpl`. The issue appears to be that `simpl` is fast, but unnecessarily unfolds the base type of the path type. If we make that type `Opaque` then simpl becomes slow. I've therefore opted for `simpl` to be fast, but when giving terms as we do in the coherences for monoidal categories, it's best to use `change`.

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -260,40 +260,61 @@ Section Reduction.
...
-    1: exact mon_unit.
-    refine (_ * IHxs).
-    destruct x as [x|x].
-    1: exact (s x).
-    exact (- s x).
+    induction x as [|x xs IHxs].
+    - exact mon_unit.
+    - destruct xs.
+      + exact (word_rec G s x).
+      + exact (word_rec G s x * IHxs).
```
jdchristensen, line 278: This change makes it compute better on singleton lists (avoiding a multiplication by the unit), but makes it not compute definitionally on cons, and also makes some other things harder.  At first, this doesn't seem like a good tradeoff, but since `freegroup_in` produces a singleton list, this change makes the recursor compute definitionally, which then makes many other things *much* simpler.  Clever.

It took me a while to understand this, so I think you should explain why the second destruct is done and why this is important.

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -752,17 +717,8 @@ Proof.
...
-        (ab_tensor_prod_twist (tensor b (tensor a c))))).
-  rewrite functor_ab_tensor_prod_beta_tensor.
-  rewrite ab_tensor_swap_beta_tensor.
-  rewrite 2 ab_tensor_prod_twist_beta_tensor_tensor.
-  rewrite 2 functor_ab_tensor_prod_beta_tensor.
-  rewrite 2 ab_tensor_swap_beta_tensor.
-  rewrite ab_tensor_prod_twist_beta_tensor_tensor.
-  reflexivity.
+  change (tensor c (tensor a b) = tensor c (tensor a b)).
+  exact idpath.
```
jdchristensen, line 719: I'm just curious about using `exact idpath` vs `reflexivity`.

jdchristensen:
Nice work!  Many things have simplified.  I pushed a few more simplifications.

> In fact, just doing this has made all the monoidal coherences hold definitionally!

You mean, after applying the propositional induction principle, right?

Alizter:
@jdchristensen Yes, sorry I should have said that. It's not quite a "strict" monoidal category, but its pretty close.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: Reflexivity does some other reductions which I suspected were causing a slow down. I'm not entirely sure however I should check the timings.

theories/Algebra/Groups/FreeGroup.v
Alizter: Yes, it is a bit of a trick, but in practice we care more about computing the recurosr on generators than on lists of elements. I'll see if I can add some comments explaining this.

Alizter:
It no longer makes sense to keep the beta rules around so I have removed them and tried to simplify some of the proofs. I'll work on commenting the "trick" in the free group recursor next.

theories/Algebra/AbGroups/TensorProduct.v
Alizter: I've removed them all now.

theories/Algebra/Groups/FreeGroup.v
Alizter: I've added a comment explaining this behaviour in FreeGroup.v. I also did a general cleanup in that file.

Alizter:
@jdchristensen If you agree I think we have resolved the final issue here with the slowness.

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -96,87 +80,89 @@ Section Reduction.
...
+    change (freegroup_eta ((c ++ [b] ++ [b^] ++ d) ++ a)
+      = freegroup_eta ((c ++ d) ++ a)).
+    lhs_V nrapply ap.
+    1: nrapply app_assoc.
+    lhs_V nrapply (ap (fun x => freegroup_eta (c ++ x))).
+    1: nrapply app_assoc.
+    lhs_V nrapply (ap (fun x => freegroup_eta (c ++ _ ++ x))).
+    1: nrapply app_assoc.
+    rhs_V nrapply ap.
+    2: nrapply app_assoc.
```
jdchristensen, line 106: Just curious why this went from two uses of associativity to four.

jdchristensen:
> If you agree I think we have resolved the final issue here with the slowness.

Well, there are still issues with using `cbn` which could bite us in the future, but that's also the case in other parts of the library, so I'm happy with how this works now.

It's also great how much simplified by making the free group recursor compute to the expected thing definitionally!

What's left to do in this PR?  If you think it's ready to merge, let me make one more pass over it before you do.  It might be a day or two before I get a chance.

theories/Algebra/Groups/FreeGroup.v
```diff
@@ -96,87 +80,89 @@ Section Reduction.
...
 
   (** We can change the sign of all the elements in a word and reverse the order. This will be the inversion in the group *)
   Fixpoint word_change_sign (x : Words) : Words.
   Proof.
     destruct x as [|x xs].
-    1: exact nil.
-    exact (word_change_sign xs @ [change_sign x]).
+    - exact nil.
+    - exact (word_change_sign xs ++ [ change_sign x ]).
   Defined.
```
jdchristensen, line 120: This could be replaced with `reverse (list_map change_sign x).` (but then one or more later proofs would need to be updated).  In general, I like your changes, and think it would be great if more proofs here were replaced with general facts about lists.

theories/Algebra/Groups/FreeGroup.v
Alizter: The list concatenation operation associates the other way so the "tau" constructor has a different bracketing to before.

Alizter: To expand on this comment: Previously we had a left-associative notation `@` for list concatenation, but I changed it to use the usual `++` notation which is right-associative.

List concatenation is better when it is right-associative since you can write `a :: b ++ c` unambiguously with the way `::` works.

We end up having to do more rewriting but I don't think it really matters in the end. We could probably optimise the number of steps further at some point.

theories/Algebra/Groups/FreeGroup.v
Alizter: Yes, that seems like a sensible thing to do.

Alizter: I've redefined `word_change_sign` like you've suggested and updated the later proofs. There were some missing list lemmas so I've added those as necessary.

Alizter:
@jdchristensen I think another pass would be good. Here is a checklist for the remaining issues that I can think of:
- [ ] I think some of the comments in `TensorProduct.v` need updating as they are referencing the old way of reducing terms.
- [ ] Have another look at why `cbn` is slow. I think this is due to the unfolding of the type, but I can't understand why the `simpl never` isn't preventing this from happening.
- [ ] Consider renaming `ab_tensor_prod` to `ab_tensor`? I've got some preliminary stuff on the tensor product of modules, and there I've named simple tensors `mtensor` and I can give the abelian group/bimodule a better name. To me `ab_tensor` seems more economical.

Alizter:
I wrote a little intro to the file to explain the tensor product of abelian groups. When I was first learning about tensor products as an undergraduate I struggled with understanding what they are supposed to be useful for. I think the punchline that should be kept in mind is that they describe biadditive maps. It seems obvious to us now, but it might help any future readers who are maybe not so familiar with the concept.

jdchristensen:
> * Consider renaming `ab_tensor_prod` to `ab_tensor`?

I'm fine with this, but how will we name the simple tensors in the module case?  One pattern would be `ab_tensor_prod`, `ab_tensor`, `mtensor_prod` and `mtensor` for the operations on groups/modules and the operations on elements.  (Right now we just use `tensor` for the second one, but it's hard to see how that will have a corresponding name in the module case.)

theories/Classes/theory/premetric.v
```diff
@@ -299,8 +298,7 @@ Class Continuous@{UA UB}
 Arguments continuous {A _ B _} f {_} _ _.
 
 Definition BinaryDup@{i} {A : Type@{i} } : A -> A /\ A := fun x => (x, x).
-Definition uncurry {A B C} (f : A -> B -> C) : A /\ B -> C
```
Alizter, line 302: Ah I was wondering why this wasn't working! Thanks for finding this.

Alizter:
I don't yet know what the best names will be once we introduce tensor products of modules, so perhaps the best thing to do would be to rename `tensor` to `ab_tensor`.

Alizter:
After thinking about it some more, at some point when we start doing calculations with modules we might prefer to work only with R-modules and treat abelian groups as Z-modules in that case. Then it might make sense for the module tensor product to have the name `mtensor` for the tpye (don't know yet if it will be a bimodule or what) and `tensor` for simple tensors.

Alizter:
Out of curiosity here are the longest lines:
```
./theories/Algebra/AbGroups/TensorProduct:     Chars 199 - 264 [Require~Import~Algebra.Groups....] 0.07 secs (0.04u,0.029s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 3416 - 3424 [Defined.] 0.038 secs (0.024u,0.014s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 23833 - 23864 [(apply~(grp_homo_op~(ab_mul~z))).] 0.036 secs (0.036u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 25793 - 25822 [exact~(tensor_ab_mul~_~_~_)^.] 0.034 secs (0.034u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 14868 - 14899 [(rapply~path_sigma_hprop;~simpl).] 0.029 secs (0.029u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 3736 - 3744 [Defined.] 0.026 secs (0.026u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 11658 - 11725 [rapply~(ab_tensor_prod_ind_hpr...] 0.025 secs (0.025u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 7021 - 7029 [Defined.] 0.022 secs (0.022u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 6539 - 6557 [strip_truncations.] 0.017 secs (0.016u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 16661 - 16669 [Defined.] 0.017 secs (0.017u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 12230 - 12282 [rapply~(ab_tensor_prod_ind_hpr...] 0.017 secs (0.017u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 3396 - 3415 [by~exists~a,b,b'.] 0.016 secs (0.002u,0.014s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 11038 - 11077 [rapply~ab_tensor_prod_ind_hpro...] 0.016 secs (0.016u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 88 - 165 [Require~Import~WildCat.Core~Wi...] 0.015 secs (0.009u,0.005s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 14917 - 14925 [Defined.] 0.014 secs (0.014u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 0 - 46 [Require~Import~Basics.Overture...] 0.012 secs (0.008u,0.004s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 9530 - 9562 [rapply~ab_tensor_prod_ind_hprop.] 0.01 secs (0.01u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 20239 - 20245 [Proof.] 0.01 secs (0.01u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 8990 - 8998 [Defined.] 0.009 secs (0.009u,0.s)
./theories/Algebra/AbGroups/TensorProduct:     Chars 8948 - 8972 [(rewrite~<-~tensor_neg_l).] 0.008 secs (0.008u,0.s)
```

And the entire file is under 1s.
```
Benchmark 1: /nix/store/6hyzsj2kifsymlv8zm39lnqrybfbigi9-coq-8.19.2/bin/coqc -R ./theories HoTT -noinit -indices-matter benched_file.v
  Time (mean ± σ):     939.0 ms ±  18.5 ms    [User: 817.2 ms, System: 118.1 ms]
  Range (min … max):   900.0 ms … 965.7 ms    10 runs
```

Alizter:
We can mark `ab_tensor_prod` as opaque and remove a few `unfold`s and the entire file compiles.  We can even remove most of the change statements and replace it with `simpl`.

However this has a detrimental effect on the timing of the file, whereby the compilation time goes from 1s to around 1.2s. There is a lot of noise, however the increase appears to be consistent.

Also, `cbn` still hangs.

jdchristensen:
However the naming is done, I think we need some pattern between the abelian group names and the module names.  And `ab_tensor_prod/mtensor_prod` isn't really bad for the name of type.

jdchristensen:
> We can mark `ab_tensor_prod` as opaque and remove a few `unfold`s and the entire file compiles. We can even remove most of the change statements and replace it with `simpl`.
> 
> However this (only marking as opaque not removing change) has a detrimental effect on the timing of the file, whereby the compilation time goes from 1s to around 1.2s. There is a lot of noise, however the increase appears to be consistent.
> 
> Also, `cbn` still hangs.

Since it doesn't fix `cbn`, then I think we should leave it as it is.

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -0,0 +1,691 @@
...
+Definition ab_tensor_swap_natural {A B A' B'} (f : A $-> A') (g : B $-> B')
+  : ab_tensor_swap $o functor_ab_tensor_prod f g
+    $== functor_ab_tensor_prod g f $o ab_tensor_swap.
+Proof.
+  snrapply ab_tensor_prod_ind_homotopy.
+  simpl. (* This speeds up the [reflexivity] and the [Defined]. *)
+  reflexivity.
+Defined.
+
+(** The data of swap together gives us a symmetric braiding on the category of abelian groups. We will later show it is a full symmetric monoidal category. *)
```
jdchristensen, line 474: "The data of swap together gives us ..." doesn't sound right.  What did you intend?

jdchristensen (approved):
I've reread the entire diff, and think it looks good!

theories/Algebra/AbGroups/TensorProduct.v
Alizter: Not sure, maybe I tried to say "the data of the swap map together with the naturality" but that seems kind of redundant to say so I reworded it.

PR #2023, opened 2024-07-11
abgroup coequalizers

Alizter:
In this PR we define the coequalizer of two group homomorphisms in the category of abelian groups. This is defined as the cokernel of the difference of the homomorphisms.

### DIFF
--- a/theories/Algebra/AbGroups/AbHom.v
+++ b/theories/Algebra/AbGroups/AbHom.v
@@ -2,6 +2,9 @@ Require Import Basics Types.
 Require Import WildCat HSet Truncations.Core Modalities.ReflectiveSubuniverse.
 Require Import Groups.QuotientGroup AbelianGroup Biproduct.
 
+Open Scope mc_scope.
+Open Scope mc_add_scope.
+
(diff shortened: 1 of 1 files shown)


jdchristensen (approved):
I just cleaned up a couple of things.

theories/Algebra/AbGroups/AbHom.v
```diff
@@ -2,6 +2,9 @@ Require Import Basics Types.
 Require Import WildCat HSet Truncations.Core Modalities.ReflectiveSubuniverse.
 Require Import Groups.QuotientGroup AbelianGroup Biproduct.
 
+Open Scope mc_scope.
+Open Scope mc_add_scope.
```
Alizter, line 6: Should this be local?

Alizter:
@jdchristensen `Open Scope` is by default `Global` which is why it interfered with a later file in the dependency chain. I went ahead and made it `Local`. It would be nice if we could set the default behaviour to `Local` somewhere.

jdchristensen (approved):

PR #2024, opened 2024-07-11
notations for left and right module actions

Alizter:
In this PR we introduce notations `r *L m` for `lact r m` and `m *R r` for `ract m r`. If we don't like them, then we can drop this PR.

### DIFF
--- a/theories/Algebra/Rings/Module.v
+++ b/theories/Algebra/Rings/Module.v
@@ -17,7 +17,7 @@ Local Open Scope module_scope.
 
 (** An abelian group [M] is a left [R]-module when equipped with the following data: *) 
 Class IsLeftModule (R : Ring) (M : AbGroup) := {
-  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [r *L m : M]. *)
+  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [lact r m : M], which we also denote [r *L m]. *)
   lact : R -> M -> M;
   (** Actions distribute on the left over addition in the abelian group. That is [r *L (m + n) = r *L m + r *L n]. *)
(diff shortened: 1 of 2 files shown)


theories/Algebra/Rings/Module.v
```diff
@@ -16,18 +17,20 @@ Declare Scope module_scope.
 
 (** An abelian group [M] is a left [R]-module when equipped with the following data: *) 
 Class IsLeftModule (R : Ring) (M : AbGroup) := {
-  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [lact r m : M]. *)
+  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [r *L m : M]. *)
```
jdchristensen, line 20: In this comment, the notation is used, and in the comment about ract, ract is used.  Maybe in both places give both descriptions?
```suggestion
  (** A function [lact] (left-action) that takes an element [r : R] and an element [m : M] and returns an element [lact r m : M], which we also denote [r *L m]. *)
```

theories/Algebra/Rings/Module.v
```diff
@@ -107,11 +110,13 @@ Defined.
 Class IsRightModule (R : Ring) (M : AbGroup)
   := isleftmodule_op_isrightmodule :: IsLeftModule (rng_op R) M.
   
-(** [ract] (right-action) that takes an element [m : M] and an element [r : R] and returns an element [lact r m : M]. *)
+(** [ract] (right-action) that takes an element [m : M] and an element [r : R] and returns an element [ract r m : M]. *)
```
jdchristensen, line 113:
```suggestion
(** [ract] (right-action) that takes an element [m : M] and an element [r : R] and returns an element [ract m r : M] which we also denote [m *R r]. *)
```
(Fixes order of arguments to ract and adds notation.)

jdchristensen (approved):

PR #2026, opened 2024-07-16
fix induction principles for nat

Alizter:
This patch fixes the induction principles for nat. Unfortunately due to coq/coq#10766 the `induction` tactic will choose `nat_rec` (??) when elminating `nat`, but this can be worked around by defining `nat_rec` as `nat_ind`. This doesn't affect anything else and is something we have done before. For simple inductive types like this, we really care about the dependent eliminator anyway.


theories/Basics/Overture.v
```diff
@@ -698,15 +698,15 @@ Ltac path_via mid :=
...
 (**  Natural numbers. *)
 Inductive nat : Type0 :=
 | O : nat
 | S : nat -> nat.
 
+Scheme nat_ind := Induction for nat Sort Type.
+Scheme nat_rect := Induction for nat Sort Type.
+Scheme nat_rec := Induction for nat Sort Type.
+
 Local Unset Elimination Schemes.
```
jdchristensen, line 710: I guess this line can be deleted too, since it's already done earlier in the file?

jdchristensen (approved):
I'm not sure what exactly is being fixed, since I never noticed a problem with induction over a natural number, but I see that a few `with` annotations can now be removed, so I guess that's good.

Alizter:
@jdchristensen We were using the autogenerated `nat_rec` (which is `nat_ind` when autogenerated) however I think the universe variables were a bit messed up. This meant when `induction` was used it was making incorrect universe assumptions. With this fix everything is working as it should.

PR #2027, opened 2024-07-16
move SuccessorStructure.v over to using Int

Alizter:
SuccessorStructure.v was still using `BinInt` we switch it over to `Int`.

### DIFF
--- a/theories/Homotopy/PinSn.v
+++ b/theories/Homotopy/PinSn.v
@@ -18,7 +18,6 @@ Local Open Scope pointed_scope.
 Section Pi1S1.
   Context `{Univalence}.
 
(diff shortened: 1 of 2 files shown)


theories/Homotopy/PinSn.v
```diff
@@ -18,7 +18,6 @@ Local Open Scope pointed_scope.
 Section Pi1S1.
   Context `{Univalence}.
 
-  Local Open Scope binint_scope.
```
Alizter, line 21: This was left over from a previous cleanup. Since the scope is no longer in context this line would fail. We remove it entirely as it is unneeded.

jdchristensen (approved):

PR #2028, opened 2024-07-16
Cleanup of Nat.v part 1

Alizter:
This PR depends on:
- #2025
- #2026

In this PR we begin the cleanup of Nat.v. The work will be split in a few PRs as there is a lot of work to do. The end goal will be to have a single file `theories/Spaces/Nat.v` that contains all of the lemmas about `nat` including any missing lemmas in one place.

This PR focuses on consolidating the basic operations of `nat` in one place, showing that `nat` has all the properties of a semiring, focuses on naming these correctly and removes any unused (and probably useless) operations on `nat`. Further cleanups will follow that address orders which for now are left untouched.

The commits are all self-contained to aid review.


theories/Spaces/Nat/Core.v
```diff
@@ -57,31 +57,33 @@ Fixpoint sub n m : nat :=
 
 Notation "n - m" := (sub n m) : nat_scope.
 
-(** ** Minimum, maximum *)
-
+(** TODO: rename [nat_max]. *)
+(** The [max n m] of two natural numbers [n] and [m]. *)
```
jdchristensen: The comments for the next three items should include the words "maximum", "minimum" and "power" to aid in searching for things.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -28,17 +28,17 @@ Defined.
 Global Instance nat_0: Zero@{N} nat := 0%nat.
 Global Instance nat_1: One@{N} nat := 1%nat.
 
-Global Instance nat_plus: Plus@{N} nat := Nat.Core.add.
+Global Instance nat_plus: Plus@{N} nat := Nat.Core.nat_add.
```
jdchristensen, line 31: Does it work to make this an Existing Instance instead, so we don't have two names for the same thing?

theories/Spaces/Nat/Core.v
```diff
@@ -647,12 +646,12 @@ Proof.
...
 Defined.
 
-Global Instance isinj_nat_add_l@{} k : IsInjective (add k).
+Global Instance isinj_nat_add_l@{} k : IsInjective (nat_add k).
 Proof.
   simple_induction k k Ik; exact _.
 Defined.
 
-Definition isinj_nat_add_r@{} k : IsInjective (fun x => add x k).
+Definition isinj_nat_add_r@{} k : IsInjective (fun x =>nat_add x k).
```
jdchristensen, line 611: Missing a space before `nat_add`.

theories/Classes/implementations/peano_naturals.v
```diff
@@ -574,15 +574,15 @@ Definition nat_full@{} := ltac:(first[exact nat_full'@{Ularge Ularge}|
                                       exact nat_full'@{}]).
 Local Existing Instance nat_full.
 
-Lemma le_nat_max_l n m : n <= Nat.Core.max n m.
+Lemma le_nat_max_l n m : n <= Nat.Core.nat_max n m.
```
jdchristensen, line 577: Should we import Nat.Core to simplify this?

theories/Spaces/Nat/Core.v
```diff
@@ -527,46 +526,46 @@ Defined.
 
 (** Maximum and minimum : definitions and specifications *)
```
jdchristensen, line 536: This comment is wrong, since the definitions are earlier.  (Also, remove space before colon.)

theories/Spaces/Nat/Core.v
```diff
@@ -146,51 +146,6 @@ Fixpoint sqrt_iter k p q r : nat :=
...
-  match k with
-  | O    => p
-  | S k' =>
-    match r with
-    | O => log2_iter k' (S p) (S q) q
-    | S r' => log2_iter k' p (S q) r'
-    end
-  end.
-
-Definition log2 n : nat := log2_iter (nat_pred n) 0 1 0.
```
jdchristensen: Why removed?  Seems non-trivial so maybe worth keeping?

theories/Spaces/Nat/Core.v
```diff
@@ -119,33 +119,6 @@ Fixpoint gcd a b :=
...
-  match k with
-  | O => p
-  | S k' =>
-    match r with
-    | O => sqrt_iter k' p.+1 q.+2 q.+2
-    | S r' => sqrt_iter k' p q r'
-    end
-  end.
-
-Definition sqrt n : nat := sqrt_iter n 0 0 0.
```
jdchristensen, line 152: Again, since non-trivial, maybe worth keeping?

theories/Spaces/Nat/Core.v
```diff
@@ -196,25 +196,22 @@ Defined.
...
-(* Local Definition ap011_add := @ap011 _ _ _ add. *)
-(* Local Definition ap011_nat := @ap011 nat nat. *)
-(* #[export] Hint Resolve ap011_add : core. *)
-(* #[export] Hint Resolve ap011_nat : core. *)
+(** *** Properties of Addition *)
 
-(** TODO: rename [nat_add_zero_r] *)
-(** TODO: reverse direction *)
-Lemma add_n_O : forall (n : nat), n = n + 0.
+Definition nat_add_zero_r@{} n : n + 0 = n.
```
jdchristensen, line 230: It's odd that almost every usage of this wants it in the other direction.  I agree that this direction is the natural way to state it.

theories/Colimits/Sequential.v
```diff
@@ -208,6 +208,11 @@ Proof.
   destruct p; rewrite !concat_1p, concat_p1; reflexivity.
 Defined.
 
+Local Lemma add_n_O : forall (n : nat), n = n + 0.
+Proof.
+  simple_induction' n; simpl; auto.
+Defined.
+
```
jdchristensen, line 215: Why is this needed?

theories/Colimits/Sequential.v
```diff
@@ -213,6 +213,13 @@ Proof.
   simple_induction' n; simpl; auto.
 Defined.
 
+Local Lemma nat_add_n_Sm (n m : nat) : (n + m).+1 = n + m.+1.
```
jdchristensen, line 216: Why is this needed?

theories/Spaces/Nat/Core.v
```diff
@@ -248,19 +248,22 @@ Proof.
     exact IHn.
 Defined.
 
-(** Multiplication *)
+(** *** Properties of Multiplication *)
```
jdchristensen: Lower case "m".

jdchristensen:
Thanks for tackling this, and for breaking it up into easy to review commits.  One other thing I noticed is that some comments have line breaks, so that's another thing that can be fixed at some point.

jdchristensen:
BTW, did we really not have associativity of multiplication or distributivity of multiplication over addition anywhere?  They must be in the Classes library, since there it is shown that the naturals form a semiring.  That raises the question of dealing with overlap between this material and the material in Classes...

theories/Spaces/Nat/Core.v
Alizter: I don't have any plans for using it and I can't see why we would use it. It can always be added back in, its just the Coq stdlib implementation.

theories/Colimits/Sequential.v
Alizter: This file has some weird issues that I haven't been able to work out. AFAICT the proofs are expected to be in a certain form so simply replacing these results with the new version doesn't work. This lemma for instance is inverted, but this causes some nontrivial issues with the proof later on. I therefore opted to just duplicate and leave the cleanup as a TODO.

theories/Colimits/Sequential.v
Alizter: Same reason as I gave above.

theories/Spaces/Nat/Core.v
Alizter: I'm writing titles in title case since it looks better in the documentation.

theories/Classes/implementations/peano_naturals.v
Alizter: We cannot make this an existing instance since nat_add does not have the type of a class. If we change the type there to `Plus` that would work, but I don't plan on using more mathclasses in Nat/Core.v. There is also another difference with the universe variable, for some reason the nat implementation in mathclasses takes a universe whereas we pick Set. I don't know if this matters in practice.

theories/Spaces/Nat/Core.v
Alizter:
```suggestion
Definition isinj_nat_add_r@{} k : IsInjective (fun x => nat_add x k).
```

theories/Classes/implementations/peano_naturals.v
Alizter: I think its fine for the time being. I don't want to touch mathclasses more than I have to.

theories/Spaces/Nat/Core.v
Alizter: This comment will be replaced in a later PR.

theories/Spaces/Nat/Core.v
Alizter: One reason this might exist at all is for run time complexity of algorithms, and I don't have any plans of adding that in. If somebody wants to use it in the future, they should be able to copy the Coq implementation.

theories/Spaces/Nat/Core.v
Alizter: Same reason as log2.

theories/Spaces/Nat/Core.v
Alizter: I've started to notice the other direction being used more however. The reverse usage is better suited when you `elim`/`destruct` the equality directly, but if you are using `rewrite` or `lhs` then this direction is better.

Alizter:
@jdchristensen We did not. They are definitely in mathclasses, but I would prefer not to use the mathclasses stuff. The advantage of this approach is that we have all the important results about `nat` in one place with a homegenous style and documentation.

theories/Spaces/Nat/Core.v
Alizter: I've pushed a commit addressing "power".

theories/Colimits/Sequential.v
jdchristensen: Ok, can you add a TODO comment before these two results?

Alizter: Done.

theories/Spaces/Nat/Core.v
Alizter: I've also done this in the future PRs so I need to remember to revise the headers there too.

Alizter:
@jdchristensen I've addressed all your comments from before. Will you do another round or should I merge and submit part 2?

jdchristensen (approved):

jdchristensen:
I've fixed the issue in Colimits/Sequential and will push it here.  It actually simplifies things there a bit.

Alizter:
@jdchristensen Sorry, I merged at the same time as you posted. Could you create a PR?

jdchristensen:
Sure, will do.

PR #2029, opened 2024-07-16
remove IsUnitPreserving from GroupHomomorphism

Alizter:
This PR removes `IsUnitPreserving` from `GroupHomomorphism` as it can always be inferred. And in practice, that is how we built group homomoprhisms anyway.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -130,8 +130,17 @@ Record GroupHomomorphism (G H : Group) := Build_GroupHomomorphism {
 
 Arguments grp_homo_map {G H}.
 Arguments Build_GroupHomomorphism {G H} _ _.
+Arguments issemigrouppreserving_grp_homo {G H} f _ : rename.
 
-(** Group homomorphisms are unit preserving *)
+(** ** Basic properties of group homomorphisms *)
+
+(** Group homomorphisms preserve group operations. This is an alias for [issemigrouppreserving_grp_homo] with the identity written explicitly. *)
+Definition grp_homo_op
+  : forall {G H : Group} (f : GroupHomomorphism G H) (x y : G), f (x * y) = f x * f y
+  := @issemigrouppreserving_grp_homo.
+#[export] Hint Immediate grp_homo_op : group_db.
+
+(** Group homomorphisms are unit preserving. *)
 Global Instance isunitpreserving_grp_homo {G H : Group}
   (f : GroupHomomorphism G H)
   : IsUnitPreserving f.
@@ -144,6 +153,13 @@ Proof.
   nrapply issemigrouppreserving_grp_homo.
 Defined.
 
+(** Group homomorphisms preserve identities. This is an alias for the previous statement. *)
+Definition grp_homo_unit
+  : forall {G H : Group} (f : GroupHomomorphism G H), f mon_unit = mon_unit
+  := @isunitpreserving_grp_homo.
+#[export] Hint Immediate grp_homo_unit : group_db.
+
+(** Therefore, group homomorphisms are monoid homomorphisms. *)
 Global Instance ismonoidpreserving_grp_homo {G H : Group}
   (f : GroupHomomorphism G H)
   : IsMonoidPreserving f
@@ -173,24 +189,6 @@ Proof.
   apply istrunc_S.
   intros f g; apply (istrunc_equiv_istrunc _ equiv_path_grouphomomorphism).
 Defined.
-
-(** ** Basic properties of group homomorphisms *)
-
-(** Group homomorphisms preserve identities. *)
-Definition grp_homo_unit {G H} (f : GroupHomomorphism G H)
-  : f (mon_unit) = mon_unit.
-Proof.
-  nrapply isunitpreserving_grp_homo.
-Defined.
-#[export] Hint Immediate grp_homo_unit : group_db.
-
-(** Group homomorphisms preserve group operations. *)
-Definition grp_homo_op {G H} (f : GroupHomomorphism G H)
-  : forall x y : G, f (x * y) = f x * f y.
-Proof.
-  nrapply issemigrouppreserving_grp_homo.
-Defined.
-#[export] Hint Immediate grp_homo_op : group_db.
 
 (** Group homomorphisms preserve inverses. *)
 Definition grp_homo_inv {G H} (f : GroupHomomorphism G H)
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -138,8 +138,8 @@ Global Instance isunitpreserving_grp_homo {G H : Group}
 Proof.
   unfold IsUnitPreserving.
   apply (group_cancelL (f mon_unit)).
-  nrefine (_ @ (grp_unit_r _)^).
-  refine (_ @ ap _ (monoid_left_id _ mon_unit)).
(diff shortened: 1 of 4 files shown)


theories/Algebra/Groups/Group.v
```diff
@@ -123,14 +123,35 @@ Defined.
...
+  apply (group_cancelL (f mon_unit)).
+  nrefine (_ @ (grp_unit_r _)^).
+  refine (_ @ ap _ (monoid_left_id _ mon_unit)).
+  symmetry.
+  nrapply issemigrouppreserving_grp_homo.
+Defined.
+
+Global Instance ismonoidpreserving_grp_homo {G H : Group}
+  (f : GroupHomomorphism G H)
+  : IsMonoidPreserving f
```
Alizter, line 149: This instance allows all those classes lemmas we were using before to still be usable.

theories/Algebra/Groups/Group.v
```diff
@@ -159,15 +180,15 @@ Defined.
...
 #[export] Hint Immediate grp_homo_unit : group_db.
 
 (** Group homomorphisms preserve group operations. *)
 Definition grp_homo_op {G H} (f : GroupHomomorphism G H)
   : forall x y : G, f (x * y) = f x * f y.
 Proof.
-  apply monmor_sgmor.
+  nrapply issemigrouppreserving_grp_homo.
 Defined.
 #[export] Hint Immediate grp_homo_op : group_db.
```
jdchristensen, line 193: Do we need `grp_homo_op` anymore?  If we rename the second field of the GroupHomomorphism record to `grp_homo_op`, will that just work?

theories/Algebra/Groups/Group.v
```diff
@@ -123,14 +123,35 @@ Defined.
...
+  issemigrouppreserving_grp_homo :: IsSemiGroupPreserving grp_homo_map;
 }.
 
+Arguments grp_homo_map {G H}.
+Arguments Build_GroupHomomorphism {G H} _ _.
+
+(** Group homomorphisms are unit preserving *)
+Global Instance isunitpreserving_grp_homo {G H : Group}
+  (f : GroupHomomorphism G H)
+  : IsUnitPreserving f.
```
jdchristensen, line 137: I wonder if the slightly more general result that I posed, saying that if you have the unbundled data of a map f and a proof that it is a semigroup morphism implies that it is unit preserving, would be useful?  Could it possibly help avoid the `pose` in the proof that groups are univalent?

jdchristensen (approved):
It's great that this turned out so simple!  I just pushed some tiny cleanups.

theories/Algebra/Groups/Group.v
Alizter: I think I am inclined towards keeping it around as it also serves a documentation purpose. The typeclasses fields while useful can be difficult when reading.

theories/Algebra/Groups/Group.v
Alizter: I had a go, but it requires unbundling the groups too. I think this makes this kind of lemma too verbose. I agree that the univalence proof could probably be improved, but I think for now it is good enough.

Alizter:
@jdchristensen Unless you have anything else to add, I think this is ready to go?

theories/Algebra/Groups/Group.v
jdchristensen: Can you explain more why you think we should duplicate this?  We don't generally write alternative names for all record fields.  Doing so means there are two names that mean the same thing, which can make it harder to learn.  Can you point to exactly what was done with modules and also why that makes sense?

theories/Algebra/Groups/Group.v
jdchristensen: It doesn't require unbundling the groups.  What I mean is what I posted in #2021:
```coq
Definition isunitpreserving_issemigrouppreserving {G H : Group}
  (f : G -> H) {h : IsSemiGroupPreserving f}
  : IsUnitPreserving f.
```

Alizter: @jdchristensen Yes, I used that, but to be useful in the univalence proof the group needs to be unbundled.

theories/Algebra/Groups/Group.v
Alizter: If I do `Search GroupHomomorphism.` as a user and look for lemmas about group homomorphisms preserving operations, its less obvious to see `IsSemiGroupPreserving` than an explicit `f (x * y) = f x * f y`.

My point about modules is that we had say the associativity of the left action together with the ring operation as a `HeteroAssociative` instance. I don't think it is even possible to understand what `HeteroAssociative _ _ _ _` means without thinking really hard, so the more obvious axiom to look for is the explicitly stated `(x * y) *L m = x *L (y *L m)`.

The question then comes up: Why do we even bother with these hard to read typeclasses? The answer I have is that they can be useful as they are typeclasses and allow us to build records easily by inferring fields. This is not possible with just an equation.

Generally, my philosophy regarding duplication is that it is fine if it saves a reader from having to jump to another file. I don't think a user interested in groups should be expected to know what a semigroup is and what exactly is being preserved.

Alizter: Anyway, that is my explanation. If you are still not convinced, I don't mind doing the change you suggested.

jdchristensen: I agree that there's a tension here.  If you avoid using the type classes, then some type class search fails.  And if you only use the type classes, then it can be hard to read, and some unfolding is needed in some spots.  So I guess I'm ok with this in a limited way.  I just pushed a commit that at least puts the duplicated things consecutively in the file, and labels them as aliases.  I also got rid of the beta expansion in their definitions.  (And note that this all applies to grp_homo_unit as well.)

theories/Algebra/Groups/Group.v
jdchristensen: The group is already unbundled in that proof, so it would need to be assembled with Build_Group, like it is now.  So all this would change is avoiding having to bundle idmap into a group homomorphism, which is not a big deal.  (And I think it might make other things more awkward.)  So I'm fine with how it is.

jdchristensen:
This LGTM!

jdchristensen:
Hmm, I don't understand the build failure, for just the coqchk (dev) build.  
```
Fatal Error: User error: Inconsistent assumptions over module
              Coq.Setoids.Setoid .
```

jdchristensen:
BTW, one other thought about this idea of getting rid of IsUnitPreserving in the definition of a group homomorphism:  it will sometimes be the case that a particular group homomorphism has a simple proof that it preserves the identity, while the proof that we get via IsSemiGroupPreserving will be more complicated.  In particular, the coercion to pointed maps will be more involved than it has to be.  But since almost all of the time we were using the proof via IsSemiGroupPreserving, there's no difference in practice.  Moreover, this proof lies in an h-set, so can usually be ignored.

Alizter:
> Hmm, I don't understand the build failure, for just the coqchk (dev) build.
> 
> ```
> Fatal Error: User error: Inconsistent assumptions over module
>               Coq.Setoids.Setoid .
> ```

That seems like the dev version of Coq changed in a major way between jobs. This error usually happens when there is a version mismatch. I will restart the CI.

Alizter:
Checking the Coq changes, only a month ago in master was the version bumped to 8.21 internally so I don't know if that was the culprit. In any case, rerunning seems to have fixed the issue so merging.

PR #2030, opened 2024-07-17
revise definition of normal subgroup

Alizter:
We revise the definition of normal subgroup to fit the classical definition. Previously I opted for the equivalent characterization as a subgroup whose left and right cosets are equivalent thinking this would be easier to formalise. However it is quite easy to show this property starting from the closure under conjugation definition which is in practice easier to satisify.


theories/Algebra/Groups/Subgroup.v
```diff
@@ -349,25 +349,52 @@ Record NormalSubgroup (G : Group) := {
 Coercion normalsubgroup_subgroup : NormalSubgroup >-> Subgroup.
 Global Existing Instance normalsubgroup_isnormal.
 
-(* Inverses are then respected *)
+(** A product [x * y] is in a normal subgroup if and only if [y * x] is. *)
+Definition symmetric_in_normalsubgroup {G : Group}
+  (N : NormalSubgroup G)
+  : forall x y, N (x * y) -> N (y * x).
```
jdchristensen, line 355: Isolating this property of normal subgroups simplifies the next result.  This property can in fact be taken to be the definition, and that might make several things easier.  It's more symmetrical (since both x and y are general elements), and it avoids inverses.  Also, it's trivially true in abelian groups, making that proof easier.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -380,11 +407,11 @@ Proof.
   rewrite negate_sg_op, <- simple_associativity.
   apply symmetric_in_cosetL; cbn.
   rewrite simple_associativity.
-  apply isnormal; cbn.
+  apply equiv_isnormalsubgroup_in_cosetL_in_cosetR; cbn.
```
jdchristensen, line 410: `in_cosetL_cong` and `in_cosetR_cong` can also be proved directly from the definition of normal, in which case `equiv_isnormalsubgroup_in_cosetL_in_cosetR` might not be needed at all.  But it's fine this way now that that result has a short proof.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -349,25 +349,52 @@ Record NormalSubgroup (G : Group) := {
...
+  (N : NormalSubgroup G)
+  : forall x y, N (x * y) <~> N (y * x).
+Proof.
+  intros x y.
+  rapply equiv_iff_hprop.
+  all: apply symmetric_in_normalsubgroup.
+Defined.
+
+(** Left and right cosets are equivalent in normal subgroups. *)
+Definition equiv_isnormalsubgroup_in_cosetL_in_cosetR {G : Group}
```
jdchristensen, line 375: I forgot to include a name change in my commit:
```suggestion
Definition equiv_in_cosetL_in_cosetR_normalsubgroup {G : Group}
```

theories/Algebra/Groups/Subgroup.v
Alizter: That's an interesting point. I've pushed a commit switching to this definition and it seems slightly better. I wonder why algebra texts don't use this definition? (I think I've seen it as an exercise before though).

theories/Algebra/Groups/Subgroup.v
Alizter: It is still used in some other proofs so I'll keep it for now. Most of these proofs/statements are the "same thing" and we are just repackaging it in equivalent ways.

jdchristensen:
This LGTM.  Maybe we should include the usual definition of normal subgroup as a side corollary, just so the reader knows that the usual property holds, and also because it will probably come up?

Alizter:
@jdchristensen I've pushed a commit allowing you to build a proof of `IsNormalSubgroup` by showing conjugation invariance. I've also shown conjugation invariance is implied by our definition and therefore both definitions are equivalent.

Alizter:
I will fix the CI in another PR.

jdchristensen (approved):

jdchristensen:
I think it's ok to merge with the CI failures, since they are unrelated.

PR #2038, opened 2024-07-31
Fix binding of type_scope to Sortclass

Alizter:
We fix the order in which we bind `type_scope` to `Sortclass` allowing targets of function types to automatically be interpreted as types. This allows us to remove `%type` annotations throughout the library.

One obvious benefit from this is that when working with `nat` or mathclasses, Coq will no longer find it ambiguous to have `A -> B + C` as a type.

### DIFF
--- a/theories/Basics/Notations.v
+++ b/theories/Basics/Notations.v
@@ -1,3 +1,7 @@
+(** [type_scope] must be declared and bound early on so that later reserved notations register correctly. *)
+Declare Scope type_scope.
+Bind Scope type_scope with Sortclass.
+
 (** To reserve this notation, we must first bootstrap, and preserve the underlying [forall] notation *)
 Notation "'forall'  x .. y , P" := (forall x , .. (forall y, P) ..) (at level 200, x binder, y binder, right associativity).
 Reserved Notation "'exists' x .. y , p"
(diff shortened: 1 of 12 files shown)


theories/Basics/Notations.v
```diff
@@ -1,3 +1,7 @@
+(** [type_scope] must be declared and bound early on so that later reserved notations register correctly. *)
+Declare Scope type_scope.
+Bind Scope type_scope with Sortclass.
+
 (** To reserve this notation, we must first bootstrap, and preserve the underlying [forall] notation *)
 Notation "'forall'  x .. y , P" := (forall x , .. (forall y, P) ..) (at level 200, x binder, y binder, right associativity).
```
Alizter, line 6: Previously, this notation was declared before the bind which meant when `->` was defined in `Overture.v` it registered the parser before the binding. By switching the order, we can get the correct binding behaviour.

jdchristensen (approved):
Nice catch!  I used my coqcreplace.py script and noticed that some `%type` annotations can also be removed in the Categories part of the library.  I think they are independent of your changes, but I could push them to this PR if you think that's ok.

Alizter:
@jdchristensen That's fine!

PR #2039, opened 2024-08-01
Use lemmas from Nat.Core in Colimits/Sequential instead of custom ones

jdchristensen:
The proofs of the lemmas didn't matter, just the direction.  Reversing the direction got rid of some lines dealing with `ap_V`.


Alizter:
Great! I missed the `p^` in `J` which led to me being very confused trying to fix this proof hence why I left it like that. It would be good to revise this file again in the future as it is an important result for spectra.

PR #2040, opened 2024-08-01
Cleanup of Nat.v part 2

Alizter:
We continue the cleanup of `Nat/` started in #2028. This PR focuses on the comparison operators for `nat`. There is still more cleanup that needs to be done, but this gets a good deal out of the way.

This PR consists of moving results from arithmetic, renaming things properly and proving missing results.

There is more to come after this.

### DIFF
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -906,8 +906,7 @@ Defined.
 
 (** There are various lemmas we can state about negating the comparison operators on [nat]. To aid readability, we opt to keep the order of the variables in each statement consistent. *)
 
-(** TODO: rename not_lt_implies_geq -> geq_not_lt *)
-Definition not_lt_implies_geq {n m} : ~(n < m) -> n >= m.
+Definition geq_not_lt {n m} : ~(n < m) -> n >= m.
 Proof.
   intros not_lt.
   destruct (@leq_dichot m n); [ assumption | contradiction].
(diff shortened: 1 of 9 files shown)


theories/Spaces/Nat/Core.v
```diff
@@ -467,190 +521,429 @@ Proof.
...
+Global Instance transitive_lt : Transitive lt := @lt_trans.
+Global Instance ishprop_lt n m : IsHProp (n < m) := _.
 Global Instance decidable_lt n m : Decidable (lt n m) := _.
 
-Definition ge n m := leq m n.
-Existing Class ge.
-#[export] Hint Unfold ge : core typeclass_instances.
-Infix ">=" := ge : nat_scope.
-Global Instance ge_is_leq n m : leq m n -> ge n m | 100 := idmap.
+(** *** Basic Properties of [>=] *)
```
jdchristensen, line 553: Maybe at the end of this series you should remove all trailing whitespace in the file.

theories/Spaces/Nat/Core.v
```diff
@@ -467,190 +521,429 @@ Proof.
...
+Definition nat_add_sub_cancel_r m n : m + n - n = m.
 Proof.
-  revert m; simple_induction' n; destruct m; cbn; auto.
+  rhs_V nrapply (nat_add_sub_cancel_l m n).
+  nrapply (ap (fun x => x - n)).
+  nrapply nat_add_comm.
 Defined.
 
-Lemma nat_max_0_n n : nat_max 0 n = n.
+(** We can cancel a right subtrahend when adding it on the right to a subtraction if the subtrahend is less than the number being subtracted from. *)
```
jdchristensen, line 639: Is "subtrahend" a word?

jdchristensen (approved):

theories/Spaces/Nat/Core.v
Alizter: I also learnt this for the first time. Apparently yes: https://www.merriam-webster.com/dictionary/subtrahend

I'm ok changing it to something else if you think it is too obscure. I just wouldn't know what to call it.

PR #2041, opened 2024-08-03
EMSpace: every n-connected and (n+1)-truncated type is an EM space

jdchristensen:


theories/Homotopy/EMSpace.v
```diff
@@ -10,7 +10,7 @@ Require Import Homotopy.Hopf.
 Require Import Truncations.Core Truncations.Connectedness.
 Require Import WildCat.
 
-(* Formalisation of Eilenberg-MacLane spaces *)
```
Alizter, line 13: I've asked a question on the categories mailing list on which spelling is correct. I seem to recall a story where his wife who was his typist preferred the latter spelling due to it being easier to type on a typewriter. I'll see in a few days what other people think it should be. But for now this is fine.

Alizter (approved):
LGTM

theories/Homotopy/EMSpace.v
jdchristensen: He wasn't completely consistent himself, but he used "Mac Lane" most often, e.g. in authorship of his book Categories for the Working Mathematician.

PR #2042, opened 2024-08-03
Cleanup of Nat.v part 3

Alizter:
This is part 3 of the cleanup of Nat following:
- #2028 
- #2040

In this PR we:
- Show various properties of subtraction
- Show all movement lemmas for summands around a `<=` or `<`.
- Show monotonicity of powers
- Move and reprove "strong induction" for nat. This will be a useful lemma when we start characterizing divisibility.

We have some TODOs which I will complete when I find the time:
- [x] Fix the capitalisation of the headers

### DIFF
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -1224,8 +1224,7 @@ Proof.
   by rewrite nat_add_sub_cancel_l in H2.
 Defined.
 
-Definition leq_moveL_nM {k m n}
-  : n <= k -> k - n <= m -> k <= m + n.
+Definition leq_moveL_nM {k m} n : n <= k -> k - n <= m -> k <= m + n.
 Proof.
   rewrite nat_add_comm.
   apply leq_moveL_Mn.
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -875,17 +875,15 @@ Defined.
 
 (** *** Dichotomy of [<=] *)
 
-(** TODO: rename *)
-Fixpoint leq_dichot m n : (m <= n) + (m > n).
+Definition leq_dichot m n : (m <= n) + (m > n).
 Proof.
-  simple_induction' m; simple_induction' n.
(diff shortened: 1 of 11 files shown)


theories/Spaces/Nat/Core.v
```diff
@@ -436,24 +445,46 @@ Global Existing Instance leq_succ | 100.
 
 (** The converse to [leq_succ] also holds. *)
 Definition leq_succ' {n m} : n.+1 <= m.+1 -> n <= m := leq_pred.
+Hint Immediate leq_succ' : typeclass_instances.
```
jdchristensen, line 448: Does it work to get rid of `leq_succ'` and make `leq_pred` a hint?

jdchristensen (approved):

theories/Spaces/Nat/Core.v
jdchristensen: If not, maybe it should be renamed to `leq_pred'`.  Using a prime to denote the converse doesn't match our other conventions.

Alizter: Making only `leq_pred` a hint does not work. I will update the name however.

theories/Spaces/Nat/Core.v
```diff
@@ -835,24 +850,24 @@ Defined.
...
 Proof.
-  snrapply equiv_adjointify.
+  srapply equiv_iff_hprop.
+  - srapply hprop_allpath.
+    intros x y.
+    snrapply equiv_path_sum.
+    destruct x as [l|p], y as [q|r].
+    1,4: rapply path_ishprop.
+    + destruct r; contradiction (lt_irrefl _ _).
+    + destruct p; contradiction (lt_irrefl _ _).
```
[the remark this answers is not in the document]
Alizter: What's the best way to state mutual exclusivity?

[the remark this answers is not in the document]
Alizter: Actually, we already had such a lemma so I've gone ahead and used it.

Issue #2022, opened 2024-07-10
Tensor product of cyclic groups


Alizter:
Now that we have tensor products of abelian groups in #2021, a nice exercise (and something useful for calculations) would be to show that

$$\mathbb{Z}/a\mathbb{Z} \otimes_{\mathbb{Z}} \mathbb{Z}/b\mathbb{Z}\cong \mathbb{Z}/(a,b)\mathbb{Z}$$

For this we will need some improvement to the material we have on cyclic groups. One important thing would be to have a multiplication formula for equivalence classes, i.e. $[xy]_a=[x]_a y$ for integers $x$ and $y$.

Once we have that the proof of the result is very straightforward. We have a bilinear (biadditive) map $f : \mathbb{Z}/a\mathbb{Z} \times \mathbb{Z}/b\mathbb{Z}\to \mathbb{Z}/(a,b)\mathbb{Z}$ given by $f([x]_a,[y]_b)=[xy]_{(a,b)}$ and an inverse map $g : \mathbb{Z}/(a,b)\mathbb{Z} \to \mathbb{Z}/a\mathbb{Z} \otimes_{\mathbb{Z}} \mathbb{Z}/b\mathbb{Z}$ given by $g([x]_{(a,b)})=[x]_a \otimes [1]_b$. We only need show one direction is a homomorphism which seems easiest for $f$. Checking the inverses should then be a straightforward exercise.

ndcroos:
Hi, I'd like to work on this issue, please.

Alizter:
@ndcroos Great! I would start with introducing a new version of the cyclic group in `AnGroups/Cyclic.v`. The definition there is defined as a quotient of the free abelian group on one generator, but we have recently improved our own `Int` type so we should use that instead.

I think at some point we want to show that `abgroup_Z` (based on `Int`) is the free abelian group on one generator, but this is probably not the time to do it. Instead, I would suggest introducing an alternative `cyclic'` definition defined as the cokernel of the multiplication map which can be found in `AbGroups/Z.v`.

If you create a PR for that first, then I can help you set up the statement for this lemma.

ndcroos:
@Alizter A definition of `cyclic` as the cokernel of the multiplication map seems to be already present in `Cyclic.v` here:
https://github.com/HoTT/Coq-HoTT/blob/f367ba9002fdd1d6b2ed076ac158f515c99edf45/theories/Algebra/AbGroups/Cyclic.v#L78

[...]

PR #2050, opened 2024-08-06
Added definition for cyclic' to AbGroups/Cyclic.v

ndcroos:
See https://github.com/HoTT/Coq-HoTT/issues/2022

### DIFF
--- a/theories/Algebra/AbGroups/Cyclic.v
+++ b/theories/Algebra/AbGroups/Cyclic.v
@@ -77,8 +77,10 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
 Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
-  := ab_cokernel@{u v} (Z1_mul_nat n).
+  := ab_cokernel@{u} (Z1_mul_nat n).
+
+(** ** Alternative definition of cyclic group *)
 
 (** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
-Definition cyclic'@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
-  := ab_cokernel@{u v} (ab_mul (A:=abgroup_Z) n).
+Definition cyclic'@{u} (n : nat) : AbGroup@{u}
+  := ab_cokernel@{u} (ab_mul (A:=abgroup_Z) n).
--- a/theories/Algebra/AbGroups/Cyclic.v
+++ b/theories/Algebra/AbGroups/Cyclic.v
@@ -78,3 +78,7 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
 Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
   := ab_cokernel@{u v} (Z1_mul_nat n).
+
+(** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
+Definition cyclic'@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
+  := ab_cokernel@{u v} (ab_mul (A:=abgroup_Z) n).
--- a/theories/Algebra/AbGroups/Cyclic.v
+++ b/theories/Algebra/AbGroups/Cyclic.v
@@ -2,8 +2,6 @@ Require Import Basics Types WildCat.Core Truncations.Core
   AbelianGroup AbHom Centralizer AbProjective
   Groups.FreeGroup AbGroups.Z Spaces.Int.
 
-Local Set Universe Minimization ToSet.
-
 (** * Cyclic groups *)
 
 (** ** The free group on one generator *)
@@ -78,11 +76,11 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 (** * Finite cyclic groups *)
 
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
-Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
-  := ab_cokernel@{u} (Z1_mul_nat n).
+Definition cyclic `{Funext} (n : nat) : AbGroup
+  := ab_cokernel (Z1_mul_nat n).
 
 (** ** Alternative definition of cyclic group *)
 
 (** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
-Definition cyclic'@{} (n : nat) : AbGroup
+Definition cyclic' (n : nat) : AbGroup
   := ab_cokernel (ab_mul (A:=abgroup_Z) n).
--- a/theories/Algebra/AbGroups/Cyclic.v
+++ b/theories/Algebra/AbGroups/Cyclic.v
@@ -2,6 +2,8 @@ Require Import Basics Types WildCat.Core Truncations.Core
   AbelianGroup AbHom Centralizer AbProjective
   Groups.FreeGroup AbGroups.Z Spaces.Int.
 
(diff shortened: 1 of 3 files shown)


theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -78,3 +78,7 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
 Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
   := ab_cokernel@{u v} (Z1_mul_nat n).
+
+(** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
+Definition cyclic'@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
+  := ab_cokernel@{u v} (ab_mul (A:=abgroup_Z) n).
```
Alizter, line 84:
```suggestion
  := ab_cokernel@{u v} (ab_mul (A:=abgroup_Z) n).

```
We usually end with a new line.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -78,3 +78,7 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
 Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
   := ab_cokernel@{u v} (Z1_mul_nat n).
+
+(** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
+Definition cyclic'@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
```
Alizter, line 83: This was surprising to see as it's not immediately obvious why this should require a strictly larger universe. Upon further investigation, I see that `grp_image` in `theories/Algebra/Groups/Image.v` is the one that requires this strictly larger universe. This is because the hypothesis for a morhpism is `A $-> B` which is a morphism in the category of groups.

Unfortunately, stating the category of groups requires a larger universe therefore we should modify `grp_image` to take `GroupHomomorphism A B` instead. This drops the strictly larger universe requirement and allows you to write `cyclic@{u}` here.

Also, you shouldn't need `Funext`. That was something that the freegroup construction required from before.

Alizter:
Also I would add a comment dividing this definition from the top half of the file:
```
(** ** Alternative definition of cyclic group *)
```

The format for comments is:
```
(* Comment that doesn't appear in documentation *)
(** Comment that appears in documentation *)
(** * Title *)
(** ** Header 1 *)
(** *** Header 2 *)
```

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -220,7 +220,7 @@ Defined.
 (** The main result of this section. *)
 Theorem ext_cyclic_ab@{u v w | u < v, v < w} `{Univalence}
```
Alizter, line 221: I believe `w` and it's condition can be dropped here.

Alizter: Actually never mind, I am mistaken.

Alizter (approved):
Thanks, this looks good to me. I'll create an issue so that we remember to replace the old definition of `cyclic` at some point.

theories/Algebra/AbSES/SixTerm.v
ndcroos: I get an "universe is unbound" error in Coq, if I drop `w` and its condition.

theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -77,4 +77,10 @@ Definition Z1_to_Z `{Funext} : ab_Z1 $-> abgroup_Z
 
 (** The [n]-th cyclic group is the cokernel of [Z1_mul_nat n]. *)
 Definition cyclic@{u v | u < v} `{Funext} (n : nat) : AbGroup@{u}
-  := ab_cokernel@{u v} (Z1_mul_nat n).
+  := ab_cokernel@{u} (Z1_mul_nat n).
+
+(** ** Alternative definition of cyclic group *)
+
+(** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
+Definition cyclic'@{u} (n : nat) : AbGroup@{u}
```
Alizter, line 85: Actually, upon further thinking, this shouldn't take any universe variables. If you remove all references of `u` and the universe annotation on AbGroup and ab_cokernel you will see Coq complain about an unbound universe.

To fix this, at the top of the file before the `(** * Cyclic groups *)` title put ```
```
This will tell Coq to try to minimize unbound universe variables to `Set` the lowest universe. Afterwards, the definition of `cyclic'@{}` should be accepted and you will see it lands in `AbGroup@{Set}` if you inspect it with `About` and `Set Printing Universes`.

Comment (ndcroos) [reply theories/Algebra/AbGroups/Cyclic.v]: I have an issue with equiv_Z1_home that I can't resolve by removing universe variables

https://github.com/HoTT/Coq-HoTT/blob/a6c4ec7b5517e21b714e8906a33c40d584adfe03/theories/Algebra/AbGroups/Cyclic.v#L32

Comment (Alizter) [reply theories/Algebra/AbGroups/Cyclic.v]: OK it looks like `Local Set Universe Minimization ToSet` messed some things up earlier up in the file. In that case, let's move it to after the "alternative def" title.

**Comment by jdchristensen** [comment]
> I'll create an issue so that we remember to replace the old definition of `cyclic` at some point.

I've wondered whether defining the integers as the free group on one generator might be the best basic definition.  You get a lot of properties for free...  But now that Int is developed so thoroughly, it does make sense to focus on that definition.

File: theories/Algebra/Groups/Image.v [hunk 5]
```diff
@@ -11,7 +11,7 @@ Local Open Scope mc_scope.
 Local Open Scope mc_mult_scope.
 
 (** The image of a group homomorphism between groups is a subgroup *)
-Definition grp_image {A B : Group} (f : A $-> B) : Subgroup B.
+Definition grp_image {A B : Group} (f : GroupHomomorphism A B) : Subgroup B.
```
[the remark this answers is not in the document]
Alizter: This was my suggestion from earlier, and yes, this will go away when Coq gets algebraic universes. I don't know if it will be the next release of Coq however. It still seems unfinished.

FTR @ndcroos the issue here is that the wildcat of groups should land in the unvierse `Set+1` however Coq doesn't let you write that for the moment, meaning we have to introduce a new universe variable strictly larger.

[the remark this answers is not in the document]
jdchristensen: Ok, I missed that comment.  But it seems weird to optimize that one result in Image.v to save a universe variable, but not the other results.  And it's really orthogonal to this PR.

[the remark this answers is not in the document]
Alizter: One difference here is that we are expecting no universe variables, whereas in other places it is one extra. Anyway, we can cleanup other universe issues some other time.

theories/Algebra/AbGroups/Cyclic.v
ndcroos: Coq complains here using this defiition about an unbound universe.

ndcroos:
I reverted the changes in `grp_image` in `Image.v`, and `SixTerm.v`.
The universe annotations are now removed for both the old and new version of `cyclic`.

Alizter (approved):

ndcroos:
Thanks @Alizter and @jdchristensen for your helpful feedback.

Alizter:
@ndcroos Thank you for the contribution. For #2022, I believe the next step (new PR) would be to define
```coq
Definition cyclic'_in {n} : abgroup_Z $-> cyclic' n := grp_quotient_map _. 
```

and then show

```coq
Definition ab_mul_cyclic'_in (n : nat) (x y : abgroup_Z)
  : ab_mul (@cyclic'_in n x) y = cyclic'_in (x * y).
```

jdchristensen (approved):
Thanks for this!

PR #2052, opened 2024-08-06
Nat: basics on mod, div and divisibility

Alizter:
In this PR we:
- Introduce a predicate of divisibility and prove some standard facts.
- Rename `divmod`, `div` and `modulo` to `nat_divmod`, `nat_div` and `nat_mod`.
- Make a technical adjustment to the definition of `nat_mod` so that `n mod 0 = n` by convention. This allows us to drop many conditions from statements such as the quotient-remainder form of division.
- Show basic lemmas about division and modulo and prove that they uniquely satisfy a quotient-remainder form.
- Rename `gcd` to `nat_gcd`.
- Prove that `nat_gcd` is the greatest common divisor and various standard facts about `gcd`.


theories/Spaces/Nat/Core.v
```diff
@@ -1437,3 +1456,202 @@ Proof.
     by apply IH_strong.
 Defined.
 
+(** ** Divisibility *)
+
+(** We define divisibility as a relation between natural numbers. *)
+Class NatDivides (n m : nat) : Type0 := nat_divides : {k : nat & k * n = m}.
+
+Notation "( n | m )" := (NatDivides n m) : nat_scope.
```
Alizter, line 1464: I've chosen this notation for divisibility which matches the Coq stdlib one. Before I tried `n | m` however this messes with the parsing of `Global Instance foo : bar | weight.`.

Alizter:
Actually, I've remarked this as draft for now. I think I can make some more progress on this.

Alizter:
This is now ready again. Before I had failed to prove that div and mod satisfy the quotient-remainder form. Part of the problem was the base case for `m mod 0` which I found was defined somewhat incorrectly. I've fixed this so that no extra side conditions are needed to make up for it.

While I was at it, I also proved many basic properties about `nat_gcd` which should hopefully be enough for @ndcroos working on #2022.

theories/Spaces/Nat/Core.v
```diff
@@ -81,41 +82,40 @@ Fixpoint nat_min n m :=
 
 (** *** Euclidean division *)
 
-(** This division is linear and tail-recursive. In [divmod], [y] is the predecessor of the actual divisor, and [u] is [y] sub the real remainder. *)
+(** This division is linear and tail-recursive. In [nat_divmod], [y] is the predecessor of the actual divisor, and [u] is [y] sub the real remainder. *)
```
jdchristensen, line 85: I found this hard to understand, so I recommend updating the comment.  Also, maybe the name should be changed to [nat_divmod_helper], to indicate that it isn't meant to be used directly?
```suggestion
(** This division takes time linear in `x` and is tail-recursive. In [nat_divmod x y q u], [x + y.+1 * q + (y - u)] is the quantity being divided and [y] is the predecessor of the divisor.  It will be called with [q] zero and [u] equal to [y], so that [x] is the quantity being divided.  The return value is a pair [(q', u')] with [x + y.+1 * q + (y - u) = y.+1 * q' + (y - u')], as least when [u <= y], as shown in [nat_divmod_spec_helper] below. *)
```

theories/Spaces/Nat/Core.v
```diff
@@ -81,41 +82,40 @@ Fixpoint nat_min n m :=
...
     match u with
-    | 0 => divmod x' y (S q) y
-    | S u' => divmod x' y q u'
+    | 0 => nat_divmod x' y (S q) y
+    | S u' => nat_divmod x' y q u'
     end
   end.
 
-Definition div x y : nat :=
+Definition nat_div x y : nat :=
```
jdchristensen, line 97:
```suggestion
(** [nat_div x y] is floor of the quotient.  In other words, it is the largest natural number [d] such that [y * d <= x].  [nat_div x 0] is defined to be [0]. See [nat_div_mod_spec] and related results below. *)
Definition nat_div x y : nat :=
```

theories/Spaces/Nat/Core.v
```diff
@@ -81,41 +82,40 @@ Fixpoint nat_min n m :=
...
   match y with
   | 0 => y
-  | S y' => fst (divmod x y' 0 y')
+  | S y' => fst (nat_divmod x y' 0 y')
   end.
+Infix "/" := nat_div : nat_scope.
 
-Definition modulo x y : nat :=
+Definition nat_mod x y : nat :=
```
jdchristensen, line 104:
```suggestion
(** [nat_mod x y] is the remainder when [x] is divided by [y].  When [y] is zero, it is defined to be [x]. See [nat_div_mod_spec] and related results below. *)
Definition nat_mod x y : nat :=
```

theories/Spaces/Nat/Core.v
```diff
@@ -1437,3 +1464,488 @@ Proof.
...
+  1: exact ((nat_mul_one_r _)^ @ p @ q^ @ nat_mul_one_r _).
+  refine (fst (nat_divmod_unique n.+1 x y 0 0 _ _ _)).
+  lhs nrapply nat_add_zero_r.
+  rhs nrapply nat_add_zero_r.
+  rewrite 2 (nat_mul_comm n.+1).
+  exact (p @ q^).
+Defined.
+
+(** The quotient-remainder form is invariant under [nat_divmod]. *)
+Definition nat_divmod_spec_helper x y q u (H : u <= y)
```
jdchristensen, line 1653:
```suggestion
(** This specifies the behaviour of [nat_divmod_helper] when [u <= y]. *)
Definition nat_divmod_helper_spec x y q u (H : u <= y)
```
The renaming is because of the suggested renaming of `nat_divmod` to `nat_divmod_helper` above.

theories/Spaces/Nat/Core.v
```diff
@@ -1437,3 +1464,488 @@ Proof.
...
+    split; trivial.
+    rewrite <- p, 2 nat_add_succ_l, <- nat_add_succ_r.
+    snrapply ap.
+    rewrite nat_sub_succ_r.
+    apply nat_succ_pred.
+    rapply lt_moveL_nV.
+Defined.
+
+(** Division and modulo can be put in quotient-remainder form. *)
+Definition nat_divmod_spec x y : x = y * (x / y) + x mod y.
```
jdchristensen, line 1675: This this specifies `nat_div` and `nat_mod`, I think an extra underscore in the name would be helpful:
```suggestion
Definition nat_div_mod_spec x y : x = y * (x / y) + x mod y.
```
In fact, I'd suggest changing `divmod` to `div_mod` throughout the file.

theories/Spaces/Nat/Core.v
```diff
@@ -700,9 +700,12 @@ Proof.
 Defined.
 
 (** We can move a subtracted number to the right-hand side of an equation. *)
-Definition nat_moveR_nV {k m} n : k = n + m -> k - m = n
-  := fun p => (nat_moveL_nV _ p^)^.
-  
+Definition nat_moveR_nV {n m} k : n = k + m -> n - k = m.
```
jdchristensen, line 703: Shouldn't it be:
```suggestion
Definition nat_moveR_nV {n m} k : n = m + k -> n - k = m.
```

theories/Spaces/Nat/Core.v
```diff
@@ -1437,3 +1464,488 @@ Proof.
...
+Defined.
+
+Global Instance antisymmetric_divides : AntiSymmetric NatDivides
+  := nat_divides_antisym.
+
+(** ** Properties of division *)
+
+Local Definition nat_divmod_unique_helper b q1 q2 r1 r2
+  : r1 < b -> r2 < b -> q1 < q2 -> r2 < r1 -> b * q1 + r1 = b * q2 + r2
+    -> q1 = q2 /\ r1 = r2.
```
jdchristensen, line 1577: This has `q1 < q2` as a hypothesis and `q1 = q2` as a conclusion.  So this gives a contradiction.  Something is fishy here. Or, if this is really intended, the comment should explain it, and maybe the conclusion should be changed to `False`?

theories/Spaces/Nat/Core.v
```diff
@@ -1437,3 +1464,488 @@ Proof.
...
+    apply leq_iff_not_gt in H5. 
+    contradiction H5.
+    rapply lt_moveR_nV.
+  - rapply nat_mul_l_monotone.
+Defined.
+
+(** Quotients and remainders are uniquely determined. *)
+Definition nat_divmod_unique d q1 q2 r1 r2
+  : r1 < d -> r2 < d -> d * q1 + r1 = d * q2 + r2
+    -> q1 = q2 /\ r1 = r2.
```
jdchristensen, line 1599: Is there an approach to this uniqueness that doesn't involve nine cases?  I'll make a general comment at the end that I think will be relevant.

jdchristensen (commented):
I didn't check everything carefully, but overall it looks fine.  One issue is that I think this file is getting much too large, and it would be better to break it into a few pieces.  E.g. all of this material about divisibility would fit in a separate file.

jdchristensen:
I also have some general comments about all of this work on the natural numbers (not specifically this PR):

1) All of these results must already have proofs in Coq.  Have you checked how others do it?  There might be some clever ways to organize things that simplify the work.

2) Some of these results are true in any semiring, and could be proved in that generality.

3) Many of these results are easier for the integers than for the natural numbers, and so could be proven there instead.  In some cases, the statements about the natural numbers would follow, using the the inclusion `nat -> int` is an injective semiring homomorphism.  One example that I think would be much easier when using the integers is the uniqueness of the divisor and remainder.  That's because in the integers you can subtract the two equations and get that `0 = y * q + r` with `-y < r < y`, which forces `r` to be `0`.  It then follows that `q` is `0`.  So I wonder if it's better to have some of these results proven in Spaces.Int or in a file in Algebra/ that depends on Spaces.Int.

theories/Spaces/Nat/Core.v
Alizter: I think the conclusion should be `Empty` here as you've observed. I believe this case was meant to get a common contradiction in the following proof.

Alizter:
> I also have some general comments about all of this work on the natural numbers (not specifically this PR):
> 
> 1. All of these results must already have proofs in Coq.  Have you checked how others do it?  There might be some clever ways to organize things that simplify the work.

Yes. My main sources of inspiration have been the standard library, mathcomp and lean's mathlib. Generally I read what results I expect to hold true, and attempt to prove things myself. If I cannot make progress on the proof, I usually check the original implementation.

I spent a long time here trying to prove the unprovable, due to the mistaken definition of `nat_mod` that I caught. After fixing it, the proofs from the stdlib more or less translated directly. There is a difference with the style of proofs since we don't use the same set of tactics and rely on typeclasses.

Before I managed to prove the quotient-remainder form for divmod `nat_divmod_spec`, I tried proving things like `n / n = 1` manually by some clever inductions. This wasn't really scalable, and having `nat_divmod_spec` simplified a lot of things.

> 2. Some of these results are true in any semiring, and could be proved in that generality.

Yes, I would assume this is the case. My overall aim here was to get something practically usable rather than picking the most general framework. I will leave it to a future look to try to generalise these results.

> 3. Many of these results are easier for the integers than for the natural numbers, and so could be proven there instead.  In some cases, the statements about the natural numbers would follow, using the the inclusion `nat -> int` is an injective semiring homomorphism.  One example that I think would be much easier when using the integers is the uniqueness of the divisor and remainder.  That's because in the integers you can subtract the two equations and get that `0 = y * q + r` with `-y < r < y`, which forces `r` to be `0`.  It then follows that `q` is `0`.  So I wonder if it's better to have some of these results proven in Spaces.Int or in a file in Algebra/ that depends on Spaces.Int.

The Coq standard library actually proves much of these results from a set of axioms that both `nat` and `Int` would satisfy. The only difference here is that the int results commonly have a `0 <= n` hypothesis littered about as having negative cases causes some issues with uniquness.

Overall, I cannot say if it would be *easier* to work directly with the integers and transplant these results back to `nat` without actually developing the corresponding theory for the integers. My, perhaps mistaken, impression is that these lemmas are easier to prove for `nat` then they are for `Int`.

theories/Spaces/Nat/Core.v
Alizter: Not certain. I tried all sorts of tricks to prove this in a shorter way, but in the end I just brought the hammer down.

theories/Spaces/Nat/Core.v
Alizter: That was what it was before actually, but I swapped the order separately. It's hard to tell here because of the name change being shown too. In my mind, moving something is going from the left to the right, but of course that's kind of nonsense in a non-commutative setting and keeping the same side makes more sense.

theories/Spaces/Nat/Core.v
Alizter: Yes, that makes more sense as a name.

theories/Spaces/Nat/Core.v
Alizter: I didn't write this comment. It was written or copied when parts of the Coq stdlib prelude were copied.

I'll change the name to `nat_divmod_helper` like you suggest.

Alizter:
OK thinking about this a bit more, I think you are right that the integers would be better to develop the theory of divisibility in. I've been sketching out the proof of #2022 in detail and I noticed at some point we will need to use Bezout's identity which doesn't hold true for `nat`. In that way, it is inevitable for us to develop the theory for integers.

I think for now, the nat stuff is good. It plugs some gaps on some missing essentials we've had for a while. I'll have to think about the integers when I get some more time.

theories/Spaces/Nat/Core.v
Alizter: I'm a little hesitant to say "floor of the quotient" since there is only one quotient that makes sense for integers anyway. This makes more sense when we are talking about the rationals or the reals, but that's not what we are talking about here.

How about saying something like: 
```suggestion
(** [nat_div x y] returns the quotient of the division of [x] by [y], defined as the largest natural number [q] such that [y * q <= x]. This is the standard procedure of "long-division", where the result is the number of times [y] can be subtracted from [x] without making [x] negative. If [y] is [0], [nat_div x 0] is defined to be [0]. See [nat_div_mod_spec] for further details and related results. *)
Definition nat_div x y : nat :=
```

theories/Spaces/Nat/Core.v
Alizter: Checking the Coq stdlib proof again, they only destruct on the trichotomy of `q1` and `q2` rather than the finer cases I've done here. The reasoning is a bit hard to understand however since they use `intuition`. I haven't spent any time stepping through the proof however. I will see if I can prove this with less cases.

Alizter: I've simplified the proof now.

theories/Spaces/Nat/Core.v
Alizter: Fixed.

Alizter:
I've split the division parts into its own file. I also noticed that it is still possible to prove Bezout's identity for `nat` by rearranging terms slightly. This means we actually don't need division for integers just yet, though it would be nice to have in the future.

theories/Spaces/Nat/Core.v
jdchristensen: I don't think "floor of the quotient" would confuse anyone, and it's immediately clear what it means, but your version is also fine.

theories/Spaces/Nat/Core.v
jdchristensen: Great, the new proof looks much better!

jdchristensen (approved):
This LGTM.  There may be ways to generalize and/or simplify the structure, but I think it's great to have all of this material!

Issue #2035, opened 2024-07-25
use CongruenceQuotient to define coequalizer


jdchristensen:
We should experiment with using `CongruenceQuotient` to define the coequalizer of two maps between groups.  This should avoid having to deal with terms of the form `-foo * bar`.  Given `f, g : A $-> B`, we probably want to define `R b b'` to be `{ a : A & f a = b * g a = b' }`.

We should also delete the hypothesis in the Section defining `ConguenceQuotient` that asserts that the relation is an hprop, since it is never used.  So we'll also avoid having strip truncations and apply tr.

Note that tensor products should simplify a bit as well.

If we do this, it's not clear whether to define the cokernel as the coequalizer against the zero map, or to keep the cokernel as the quotient by the image subgroup.

This is based on https://github.com/HoTT/Coq-HoTT/issues/2034#issuecomment-2249107312_

jdchristensen:
I think it would also be helpful to allow a subgroup of a group to be defined with a predicate that isn't assume to take values in h-propositions.  This would avoid us needing to `strip_truncations` and apply `tr` in various places, and would mean that we don't need a separate `ab_cokernel_embedding`.

jdchristensen:
> We should also delete the hypothesis in the Section defining `ConguenceQuotient` that asserts that the relation is an hprop, since it is never used

#2057 does this.

> So we'll also avoid having strip truncations and apply tr.

#2057 alone doesn't simplify anything else, since those section variables were ignored.  What would help would be to directly define coequalizers as the quotient by a non-truncated congruence relations, and to define the quotient of a group by a not-necessarily truncated subgroup.

Alizter:
@jdchristensen I've tried formalising this, but I don't follow your idea for the relation `R`. You've written `{ a : A & f a = b * g a = b' }` but I don't see how this relation is supposed to be reflexive.

Alizter:
It seems to me, that we should really be using `f a - g a = b - b'`. This is a congruence, but this has the same issues that we were trying to avoid in the first place.

jdchristensen:
@Alizter I think you're right.  Would it help things if we wrote it as `f a + b' = g a + b`, so it's fairly symmetric and avoids negation?

Alizter:
That appears to have worked, but I have yet to see if it offers any simplifications elsewhere.

jdchristensen:
[...]

PR #2053, opened 2024-08-06
replace cyclic with cyclic'

### LINKED ISSUE #2051: Replace old definition of `cyclic` with new `cyclic'`
In #2050, a new definition of `cyclic` based on `Int` is introduced. Replacing the old one is a bit non-trivial however due to its extensive use.

It's also not clear if it is easy to show `abgroup_Z` is the free group on one generator. I had a quick look, but it seems awkward to prove.

Alizter:
Actually, I misremembered. We don't use `cyclic` anywhere, so we are free to rename it. I was confused instead with the definition of the integers as a free group.

Fix #2051.

At some point we should move this definition of the integers to its own file or replace it. I opt for moving it since it is still of interest.

cc @ndcroos in case you are rebasing


Alizter:
I wonder if while we are at it we might rename `abgroup_Z` to `ab_Z` and `cyclic` to `ab_cyclic`?

jdchristensen:
I think we should use the same model for the infinite cyclic group as for its finite quotients.  So if we want the new definition of `cyclic n` using `abgroup_int` to be the official one, then we also should move `Z1` out of the file Cyclic.v.  Alternatively, the file could contain both definitions, one primed.  But this PR leaves things in a mixed state.

jdchristensen:
Maybe `Z1` should be moved to `Z.v` as an alternate definition of the integers, with a TODO to prove that it is isomorphism to `abgroup_Z`?

Alizter:
I've moved the free group integer construction to `FreeInt.v`. `Cylic.v` looks a bit bare for now, but we will soon add lots of lemmas.

theories/Spaces/FreeInt.v
```diff
@@ -0,0 +1,71 @@
...
+Definition Z1_gen : Z1 := freegroup_in tt. (* The generator *)
+
+(** The recursion principle of [Z1] and its computation rule. *)
+Definition Z1_rec {G : Group} (g : G) : Z1 $-> G
+  := FreeGroup_rec Unit G (unit_name g).
+
+Definition Z1_rec_beta {G : Group} (g : G) : Z1_rec g Z1_gen = g
+  := FreeGroup_rec_beta _ _ _.
+
+(** The free group [Z] on one generator is isomorphic to the subgroup of [Z] generated by the generator.  And such cyclic subgroups are known to be commutative, by [commutative_cyclic_subgroup]. *)
```
jdchristensen, line 19:
```suggestion
(** The free group [Z1] on one generator is isomorphic to the subgroup of [Z1] generated by the generator.  And such cyclic subgroups are known to be commutative, by [commutative_cyclic_subgroup]. *)
```

theories/Spaces/FreeInt.v
```diff
@@ -0,0 +1,71 @@
+Require Import Basics Types WildCat.Core Truncations.Core Spaces.Int
+  AbelianGroup AbHom Centralizer AbProjective Groups.FreeGroup AbGroups.Z.
+
+(** * The free group on one generator *)
+
+(** We can define the integers as the free group on one generator, which we denote [Z1] below. Results from Centralizer.v and Groups.FreeGroup let us show that [Z1] is abelian. *)
+
+(** We define [Z] as the free group with a single generator. *)
```
jdchristensen, line 8:
```suggestion
(** We define [Z1] as the free group with a single generator. *)
```

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -1,6 +1,6 @@
 Require Import Basics Types WildCat HSet Pointed.Core Pointed.pTrunc Pointed.pEquiv
   Modalities.ReflectiveSubuniverse Truncations.Core Truncations.SeparatedTrunc
-  AbGroups Homotopy.ExactSequence Spaces.Int
+  AbGroups Homotopy.ExactSequence Spaces.Int Spaces.FreeInt
```
jdchristensen, line 3: It shouldn't be hard to switch this file to use the new definition of cyclic groups.  Do we know that multiplication by a non-zero integer is an injective map?  And that the integers are projective?  (Both regarding Spaces.Int.)

jdchristensen (approved):

jdchristensen:
> I wonder if while we are at it we might rename `abgroup_Z` to `ab_Z` and `cyclic` to `ab_cyclic`?

We have other examples like `abgroup_vector` and `abgroup_matrix`, so maybe `abgroup_Z` is best?  And maybe `cyclic` should be `abgroup_cyclic`?

theories/Algebra/AbSES/SixTerm.v
Alizter: I think I will leave this to a later cleanup. We are redefining cyclic' here using `ab_cokernel_embedding`. The cleanup in #2035 should clean up this situation, afterwards it would make sense to switch to `cyclic`.

Alizter:
I think I'll hold off on any renaming. I like `cyclic` since it is short and otherwise it gets too verbose. Later we will have things like `cyclic_ind_homotopy` so you can see how sticking `abgroup` will get old quickly.

theories/Algebra/AbSES/SixTerm.v
jdchristensen: We should be able to define `cyclic` in this file using `ab_cokernel_embedding` as well.  Is the fact that multiplication by a non-zero integer is an embedding already in the library?  I don't see it, so that could be another "first issue" goal.

Alizter: We have it for `nat` but not for `Int`.

PR #2054, opened 2024-08-08
more power laws for groups

Alizter:
We show that inverses are preserved by powers (surprisingly difficult?) and that multiples can be pulled apart in the exponent.


Alizter:
Oh I forgot about the plan in #2015. Let me take another look.

Alizter:
I'll leave this for now. It can be proven in a better way as Dan suggested and I will leave it as a long hanging fruit for anybody else who is interested.

PR #2055, opened 2024-08-08
Int: conversion lemmas with nat

Alizter:
This allows for nat lemmas to be used in place of integer ones under the right conditions.

### DIFF
--- a/theories/Spaces/Int.v
+++ b/theories/Spaces/Int.v
@@ -19,7 +19,7 @@ Inductive Int : Type0 :=
 | zero : Int
 | posS : nat -> Int.
 
-(** We can convert a [nat] to an [Int] by mapping [0] to [zero] and [S n] to [posS n]. *)
+(** We can convert a [nat] to an [Int] by mapping [0] to [zero] and [S n] to [posS n]. Various operations on [nat] are preserved by this function. See the section on conversion functions starting with [int_nat_succ]. *)
 Definition int_of_nat (n : nat) :=
   match n with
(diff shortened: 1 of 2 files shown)


theories/Spaces/Int.v
```diff
@@ -732,3 +732,46 @@ Proof.
   refine (ap (fun q => equiv_path A A (loopexp q z) a) _).
   apply eissect.
 Defined.
+
+(** ** Converting between integers and naturals *)
+
+Definition int_nat_succ (n : nat)
```
jdchristensen, line 738: We already have `int_of_nat_succ_commute`, but this name is probably better.  Get rid of the other one?

theories/Spaces/Int.v
```diff
@@ -732,3 +732,46 @@ Proof.
   refine (ap (fun q => equiv_path A A (loopexp q z) a) _).
   apply eissect.
 Defined.
+
+(** ** Converting between integers and naturals *)
```
jdchristensen, line 736: Somewhere near here we should have a comment mentioning that this shows that the coercion `int_of_nat` is a "semiring homomorphism", to help people find these results.  (But the subtraction part goes beyond being a semiring homomorphism.)  Note the explicit mention of `int_of_nat`, since the name doesn't show up here.

Also, near the definition of the coercion, point the reader to here for results about the `int_of_nat` function.

jdchristensen (approved):
LGTM, modulo the comments I made.

PR #2056, opened 2024-08-08
Added proof for ab_mul_cyclic_in

ndcroos:
See https://github.com/HoTT/Coq-HoTT/issues/2022

### DIFF
--- a/theories/Algebra/AbGroups/Cyclic.v
+++ b/theories/Algebra/AbGroups/Cyclic.v
@@ -85,11 +85,11 @@ Definition cyclic `{Funext} (n : nat) : AbGroup
 Definition cyclic' (n : nat) : AbGroup
   := ab_cokernel (ab_mul (A:=abgroup_Z) n).
 
-Definition cyclic_in {n : nat} : abgroup_Z $-> cyclic' n
+Definition cyclic_in (n : nat) : abgroup_Z $-> cyclic' n
   := grp_quotient_map. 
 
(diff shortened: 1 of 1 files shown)


theories/Algebra/AbGroups/Cyclic.v
```diff
@@ -84,3 +84,13 @@ Definition cyclic `{Funext} (n : nat) : AbGroup
 (** The [n]-th cyclic group is the cokernel of [ab_mul n]. *)
 Definition cyclic' (n : nat) : AbGroup
   := ab_cokernel (ab_mul (A:=abgroup_Z) n).
+
+Definition cyclic_in {n : nat} : abgroup_Z $-> cyclic' n
```
Alizter, line 88: I think it would be better for `n` to be explicit.

Alizter: This also allows you to remove the `@` later.

ndcroos: There is a error here in ab_mul_cyclic_in:
The term "ty * x)%int" has type "Int"
while it is expected to have type "nat"

Changing to to %nat gives another (different) error.

jdchristensen: If you make `n` explicit, then it needs to be provided both times you use `cyclic_in`.

jdchristensen (approved):

Alizter (approved):
Looks great!

PR #2057, opened 2024-08-08
QuotientGroup: remove unused context variables and add comments

jdchristensen:
This doesn't affect the definitions themselves:  since the context variables were not used, they were not made into arguments of the various things defined in the section.

### DIFF
--- a/theories/Algebra/Groups/QuotientGroup.v
+++ b/theories/Algebra/Groups/QuotientGroup.v
@@ -19,10 +19,12 @@ Local Open Scope wc_iso_scope.
 
 Section GroupCongruenceQuotient.
 
+  (** A congruence on a group is a relation satisfying [R x x' -> R y y' -> R (x * y) (x' * y')].  Note that requiring [R] to be reflexive and symmetric imposes additional conditions when combined with the congruence condition, but that we don't need to assume that [R] is transitive. *)
   Context {G : Group} {R : Relation G}
-    `{is_mere_relation _ R, !IsCongruence R,
-      !Reflexive R, !Symmetric R, !Transitive R}.
+    `{!IsCongruence R,
+      !Reflexive R, !Symmetric R}.
 
+  (** The type underlying the quotient group is [Quotient R]. *)
   Definition CongruenceQuotient := G / R.
 
   Global Instance congquot_sgop : SgOp CongruenceQuotient.
--- a/theories/Algebra/Groups/QuotientGroup.v
+++ b/theories/Algebra/Groups/QuotientGroup.v
@@ -19,9 +19,8 @@ Local Open Scope wc_iso_scope.
 
 Section GroupCongruenceQuotient.
 
-  (** A congruence on a group is a relation satisfying [R x x' -> R y y' -> R (x * y) (x' * y')].  Note that requiring [R] to be reflexive and symmetric imposes additional conditions when combined with the congruence condition, but that we don't need to assume that [R] is transitive. *)
-  Context {G : Group} {R : Relation G}
-    `{!IsCongruence R, !Reflexive R, !Symmetric R}.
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/QuotientGroup.v
```diff
@@ -19,10 +19,12 @@ Local Open Scope wc_iso_scope.
 
 Section GroupCongruenceQuotient.
 
+  (** A congruence on a group is a relation satisfying [R x x' -> R y y' -> R (x * y) (x' * y')].  Note that requiring [R] to be reflexive and symmetric imposes additional conditions when combined with the congruence condition, but that we don't need to assume that [R] is transitive. *)
   Context {G : Group} {R : Relation G}
-    `{is_mere_relation _ R, !IsCongruence R,
-      !Reflexive R, !Symmetric R, !Transitive R}.
+    `{!IsCongruence R,
+      !Reflexive R, !Symmetric R}.
```
Alizter, line 25:
```suggestion
    `{!IsCongruence R, !Reflexive R, !Symmetric R}.
```

theories/Algebra/Groups/QuotientGroup.v
```diff
@@ -19,10 +19,11 @@ Local Open Scope wc_iso_scope.
 
 Section GroupCongruenceQuotient.
 
+  (** A congruence on a group is a relation satisfying [R x x' -> R y y' -> R (x * y) (x' * y')].  Note that requiring [R] to be reflexive and symmetric imposes additional conditions when combined with the congruence condition, but that we don't need to assume that [R] is transitive. *)
```
Alizter, line 22:
```suggestion
  (** A congruence on a group is a relation satisfying [R x x' -> R y y' -> R (x * y) (x' * y')].  Note that requiring [R] to be reflexive and symmetric imposes additional conditions when combined with the congruence condition, but that we don't need to assume that [R] is transitive since this is ensured by transitivity of paths. *)
```

jdchristensen: I don't think that's the right thing to say.  Reflexivity and symmetry are also ensured by transitivity of paths, and yet they are needed, since they interact with the congruence condition.  I don't fully grok why transitivity isn't needed, but I don't think that comment properly explains it.

Alizter: Yes, I'm a little surprised this works. If you want to state that equality of classes reflects the relation then you need transitivity together with univalence. It's interesting that the reflexive-symmetric closure is enough for the group structure however.

jdchristensen: BTW, the congruence condition plus reflexivity gives:  `R y y' -> R (x * y) (x * y')` and similarly for multiplication on the right by `x`.   And the congruence condition plus symmetry gives `R x y -> R (-x) (-y)`.  Oh, I've just figured out how to avoid using symmetry.  I'll push in a minute.

jdchristensen: Ok, I've pushed a commit removing symmetry as well.

Note that instead of assuming the congruence condition and reflexivity, we could instead assume just the two conditions `R y y' -> R (x * y) (x * y')` and `R x x' -> R (x * y) (x' * y)`.  I'm not sure which is more natural, so probably it's best to leave it as is.

Alizter: According to wikipedia, a reflexive-symmetric relation preserving the operation is called a "tolerance" https://en.wikipedia.org/wiki/Tolerance_relation

And they lack uniqueness which is the comment I made about univalence.

Alizter:
Nice trick for avoiding symmetry! It's unlikely that we can drop reflexivity right?

jdchristensen:
I'm pretty sure I have a counterexample showing that you can't drop reflexivity.

theories/Algebra/Groups/QuotientGroup.v
jdchristensen: For the record, a tolerance on a group is automatically a congruence, so it has the uniqueness mentioned in the wikipedia article.  It's also worth mentioning that, according to wikipedia, a congruence needs to be reflexive, symmetric and transitive, while the Coq-HoTT definition of `IsCongruence` doesn't require any of these.  So the naming we use is not ideal.

PR #2058, opened 2024-08-12
Prime numbers

### LINKED ISSUE #2008: Prime numbers
We should develop some theory about divisibility for `nat` and define a predicate `IsPrime`. This will be useful for various things that require prime numbers such as algebra and localization.

Alizter:
We define what it means to be a prime number and give a decision procedure for deciding if a given nat is prime. We give the first 8 primes as examples of this procedure.

This fixes #2008.

### DIFF
--- a/theories/Basics/Decidable.v
+++ b/theories/Basics/Decidable.v
@@ -27,8 +27,9 @@ Ltac decide_type A :=
   end.
 
 Ltac decide :=
-  match goal with
+  multimatch goal with
   | [|- ?A] => decide_type A
+  | [|- ~ ?A] => decide_type A
   end.
 
 Definition decidable_true {A : Type} `{Decidable A}
--- a/theories/Spaces/Nat/Division.v
+++ b/theories/Spaces/Nat/Division.v
@@ -641,7 +641,7 @@ Defined.
 (** A prime number is a number greater than [1] that is only divisible by [1] and itself. *)
 Class IsPrime (n : nat) : Type0 := {
   gt_one_isprime :: 1 < n;
-  is_prime : forall m, (m | n) -> (m = 1) + (m = n);
+  isprime : forall m, (m | n) -> (m = 1) + (m = n);
 }.
 
 Definition issig_IsPrime n : _ <~> IsPrime n := ltac:(issig).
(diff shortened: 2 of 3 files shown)


Alizter:
It's probably possible to simplify some of the proofs and definitions, so I will leave this PR here so people can give feedback. I'll try to iterated over it in the coming days.

jdchristensen (approved):

theories/Spaces/Nat/Division.v
```diff
@@ -329,19 +330,18 @@ Proof.
   nrapply nat_div_mod_spec.
 Defined.
 
+(** For [m > 0], [n mod m = 0] is equivalent to [m] dividing [n]. *)
+Definition equiv_nat_mod_divides n m : 0 < m -> n mod m = 0 <~> (m | n) .
+Proof.
+  intros H; srapply equiv_iff_hprop_uncurried.
+  apply iff_nat_mod_divides.
+Defined.
```
jdchristensen, line 338: This isn't used, so I'm not sure if we need to include it.  Asking for an equivalence requires the extra hypothesis that `0 < m` in order that the codomain be an hprop.

theories/Spaces/Nat/Division.v
```diff
@@ -314,6 +315,36 @@ Proof.
   apply nat_moveR_nV, nat_mul_comm.
 Defined.
 
+(** [n mod m = 0] is equivalent to [m] dividing [n]. *)
+Definition equiv_nat_mod_divides n m : 0 < m -> n mod m = 0 <~> (m | n) .
+Proof.
+  intros H; srapply equiv_iff_hprop.
+  2: exact (nat_mod_divides _ _ H).
```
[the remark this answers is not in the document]
Alizter: This can be fixed with `Local Set Universe Minimization ToSet`.

theories/Spaces/Nat/Division.v
Alizter: I've removed it.

theories/Basics/Decidable.v
```diff
@@ -27,8 +27,9 @@ Ltac decide_type A :=
   end.
 
 Ltac decide :=
-  match goal with
+  multimatch goal with
```
Alizter, line 30: I've changed this to a `multimatch` which means later branches will also be tried, if the former branches fail. This means that in practice, if `decide` fails to fill `A` then it will try to fill `~A` before giving up.

theories/Spaces/Nat/Division.v
```diff
@@ -604,3 +629,85 @@ Definition nat_bezout n m
...
+Global Instance isprime_3 : IsPrime 3 := ltac:(decide).
+Global Instance isprime_5 : IsPrime 5 := ltac:(decide).
+Global Instance isprime_7 : IsPrime 7 := ltac:(decide).
+Global Instance isprime_11 : IsPrime 11 := ltac:(decide).
+Global Instance isprime_13 : IsPrime 13 := ltac:(decide).
+Global Instance isprime_17 : IsPrime 17 := ltac:(decide).
+Global Instance isprime_19 : IsPrime 19 := ltac:(decide).
+
+(** Similarly, we can see that other natural numbers are not prime. *)
+Definition not_isprime_0 : not (IsPrime 0) := ltac:(decide).
```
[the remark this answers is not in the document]
jdchristensen: Nice, I thought something like that should work.

PR #2060, opened 2024-08-17
Use congruence quotient for ab_coeq

Alizter:
In this PR we use `CongruenceQuotient` for `ab_coeq`. I've marked this as a draft for now in case we can improve it.

This was proposed in #2035.

### DIFF
--- a/theories/Algebra/AbGroups/AbHom.v
+++ b/theories/Algebra/AbGroups/AbHom.v
@@ -46,29 +46,58 @@ Defined.
 (** ** Coequalizers *)
 
 (** Using the cokernel and addition and negation for the homs of abelian groups, we can define the coequalizer of two group homomorphisms as the cokernel of their difference. *)
-Definition ab_coeq {A B : AbGroup} (f g : GroupHomomorphism A B)
-  := ab_cokernel (ab_homo_add g (negate_hom f)).
+Definition ab_coeq {A B : AbGroup} (f g : GroupHomomorphism A B) : AbGroup.
+Proof.
+  snrapply Build_AbGroup'.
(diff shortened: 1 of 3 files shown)


theories/Algebra/AbGroups/AbHom.v
```diff
@@ -46,29 +46,58 @@ Defined.
...
+  - intros x.
+    exists 0.
+    apply (ap (+ _)).
+    exact (grp_homo_unit _ @ (grp_homo_unit _)^).
+Defined.
 
 Definition ab_coeq_in {A B} {f g : A $-> B} : B $-> ab_coeq f g.
 Proof.
-  snrapply grp_quotient_map.
+  snrapply Build_GroupHomomorphism.
```
Alizter, line 75: It's possible we can factor this with `grp_quotient_map` to share the underlying definition.

jdchristensen:
If we proceeded with this, several of the things done for `ab_coeq` could be generalized to be for any congruence quotient.  E.g. we could have `CongruenceQuotientGroup` for the bundled group, `CongruenceQuotientAbgroup` for the abelian case, `congruencequotient_in`, `congruencequotient_rec`, etc.  Then the diff would be smaller in the code involving `ab_coeq`.  But still we'd need to prove that the given relation *is* a congruence quotient, whereas with the cokernel definition that follows from general theory, so this approach would still be longer, and in the end there is no benefit that I can see.  So I guess this wasn't really a helpful idea.  Thanks for trying it, @Alizter !

I do think that there would be some minor simplifications if we allowed quotients of groups by subgroups whose predicate was non-truncated, but it's not clear that this would be worth it.

PR #2061, opened 2024-08-18
Primes: division lemma for primes

Alizter:
In this PR we show a prime divisor is necessarily a divisor of one of the factors. This is otherwise known as Euclid's lemma.


jdchristensen (approved):

jdchristensen:
Nice!

Issue #2059, opened 2024-08-16
Fundamental theorem of arithmetic


Alizter:
We should prove the fundamental theorem of arithmetic for `nat` and `IsPrime`. The basic statement says that any given `nat` has a factorisation and any two factorisations are unique up to a permutation of factors. There is also a generalisation to unique factorisation domains with a possible semiring variant.

Another way to modify the statement for `nat` is to force uniqueness by enforcing an order of the prime factors of a given factorisation. Then the fundamental theorem of arithmetic essentially says that the type of factorisations of a `nat` is contractible.

In our calculations in homotopy theory, it's not apparent how useful this theorem actually is given that we can always manually factor a concrete number. So I wouldn't say this issue is a top priority.

One place to draw inspiration from is the following: https://github.com/coq-contribs/fundamental-arithmetics/blob/master/primes.v

Alizter:
I've managed to prove the existence of factorizations, however the proof does not compute efficiently. This means you cannot do say `Eval cbv in ...` to see what the factorisation should be. This leaves the following concrete things to do in the future for this issue:

- Show that any two factorizations are unique up to permutation of prime factors.
- Show that ordered factorisations by `<=` are contractible.
- Give a more efficient factorisation algorithm so we can compute concrete cases.

klinashka:
Is this issue still something that could be worked on? I would like to try and work on this if yes. Also, could you tell me where I can find your work in the codebase?

jdchristensen:
@klinashka As far as I know, this is still open.  @Alizter might have more to say.

The existing work was merged in #2062 and is in Spaces/Nat/Division.v.

If you want to work in this repo, be sure to read STYLE.md and to push a small amount of work early (say, 100-150 lines or so), so we can try to get the style right from the start rather than trying to fix it later.

Also, I'll warn you that @Alizter has been very busy lately, and I have a couple of weeks of travel coming up and may also be slow to respond.

PR #2062, opened 2024-08-18
prime factorization

Alizter:
In this PR we prove the first part of the fundamental theorem of arithmetic outlined in #2059. We show that any natural number has a factorisation into primes.

In order to do this we define `IsComposite` which checks if a natural number has a proper divisor. Now this allows us to partition `nat` into 4 parts, zero, one (unit), primes and composites.

I noticed that in the classical setting, the line between being not prime and composite is blurred due to classical logic. This means texts on elementary number theory rarely distinguish between these two concepts. In the constructive setting, it can be shown that they are equivalent since we can apply classical reasoning to IsComposite and IsPrime precisely because they are decidable.

This allows us to show that any natural has a prime divisor which can be iterated to provide a prime factorisation of any natural number.

Decidability of IsComposite required me to introduce a new `list_exists` predicate on lists which checks for a witness in a list satisfying a given predicate, analogous to `for_all`. This allows us to show decidability of certain existential statements as long as witnesses can be collected in a list (related to enumerability).

Unfortunately, I needed to tweak a lot of universe levels for example in Sum.v as the default Coq ones introduced too many "phantom" universes which didn't appear to have any bearing on the terms but caused proofs to be left hanging.

Hopefully, this PR can be improved a bit, but I will be slightly less active over the coming days.

As mentioned in #2059, showing that any two factorisations are unique upto permutation will be left to future work. We are missing a lot of material on permutations of lists so this will have to wait.

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -239,8 +239,11 @@ Global Instance symmetric_iff : Symmetric iff | 1
   := @iff_inverse.
 
 (** And reflexivity of them *)
+Definition iff_refl {A} : A <-> A
+  := (idmap , idmap).
+
 Global Instance iff_reflexive : Reflexive iff | 1
-  := fun A => (idmap , idmap).
+  := @iff_refl.
 
 (** Dependent composition of functions. *)
 Definition composeD {A B C} (g : forall b, C b) (f : A -> B) := fun x : A => g (f x).
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -73,13 +73,6 @@ Notation conj := pair (only parsing).
 
 #[export] Hint Resolve pair inl inr : core.
 
-(** If and only if *)
-
-(** [iff A B], written [A <-> B], expresses the logical equivalence of [A] and [B] *)
-Definition iff (A B : Type) := prod (A -> B) (B -> A).
-
-Notation "A <-> B" := (iff A B) : type_scope.
-
 (** ** Type classes *)
 
 (** This command prevents Coq from trying to guess the values of existential variables while doing typeclass resolution.  If you don't know what that means, ignore it. *)
@@ -223,27 +216,6 @@ Notation "g 'o' f" := (compose g%function f%function) : function_scope.
 
 (** This definition helps guide typeclass inference. *)
 Definition Compose {A B C : Type} (g : B -> C) (f : A -> B) : A -> C := compose g f.
-
-(** Composition of logical equivalences *)
-Definition iff_compose {A B C} (f : A <-> B) (g : B <-> C) : A <-> C
-  := (fst g o fst f , snd f o snd g).
-
-Global Instance transitive_iff : Transitive iff | 1
-  := @iff_compose.
-
-(** While we're at it, inverses of logical equivalences *)
-Definition iff_inverse {A B} : (A <-> B) -> (B <-> A)
-  :=  fun f => (snd f , fst f).
-
-Global Instance symmetric_iff : Symmetric iff | 1
-  := @iff_inverse.
-
-(** And reflexivity of them *)
-Definition iff_refl {A} : A <-> A
-  := (idmap , idmap).
-
-Global Instance iff_reflexive : Reflexive iff | 1
-  := @iff_refl.
 
 (** Dependent composition of functions. *)
 Definition composeD {A B C} (g : forall b, C b) (f : A -> B) := fun x : A => g (f x).
@@ -538,9 +510,6 @@ Notation "f ^-1" := (@equiv_inv _ _ f _) : function_scope.
 
 Definition ap10_equiv {A B : Type} {f g : A <~> B} (h : f = g) : f == g
   := ap10 (ap equiv_fun h).
-
-Coercion iff_equiv {A B : Type} (f : A <~> B)
-  : A <-> B := (equiv_fun f, f^-1).
 
 (** ** Contractibility and truncation levels *)
 
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -536,6 +536,9 @@ Notation "f ^-1" := (@equiv_inv _ _ f _) : function_scope.
 Definition ap10_equiv {A B : Type} {f g : A <~> B} (h : f = g) : f == g
   := ap10 (ap equiv_fun h).
 
+Coercion iff_equiv {A B : Type} (f : A <~> B)
+  : A <-> B := (equiv_fun f, f^-1).
+
 (** ** Contractibility and truncation levels *)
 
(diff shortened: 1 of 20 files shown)


theories/Basics/Overture.v
```diff
@@ -225,14 +225,18 @@ Notation "g 'o' f" := (compose g%function f%function) : function_scope.
...
 
 (** Composition of logical equivalences *)
-Global Instance iff_compose : Transitive iff | 1
-  := fun A B C f g => (fst g o fst f , snd f o snd g).
-Arguments iff_compose {A B C} f g : rename.
+Definition iff_compose {A B C} (f : A <-> B) (g : B <-> C) : A <-> C
+  := (fst g o fst f , snd f o snd g).
+
+Global Instance transitive_iff : Transitive iff | 1
+  := @iff_compose.
```
jdchristensen, line 232: I think the reason you are introducing redundant names here is because the typeclass `Transitive` introduces an extra universe variable.  I wonder if we'll be able to get rid of this once we have algebraic universes?

jdchristensen (commented):
I'm partway through reviewing this, but will have to come back and finish later.  I'll push a commit that removes some universe annotations that aren't needed, but there is more that could be done like that.  That commit also reduces universe variables in `prod_empty_l`.

jdchristensen:
I'll look into the build failures later today.  It builds for me locally.

jdchristensen:
The error was in the tests, which I hadn't built locally. `decidable_for_all` had changed to having three universe variables, instead of two.  But with an improvement to `decidable_prod`, this goes away.

theories/Spaces/Nat/Division.v
```diff
@@ -760,3 +793,130 @@ Proof.
...
+    + exact c.1.
+    + apply inlist_seq.
+      exact (snd (fst c.2)).
+    + exact c.2.
+  - exact _.
+Defined.
+
+(** For a number larger than [1], being prime is equivalent to not being composite. *)
+Definition isprime_iff_not_iscomposite@{} n
+  : 1 < n -> IsPrime n <-> ~ IsComposite n.
```
jdchristensen, line 823: I think this should be restated as a sharper iff:  `IsPrime n <-> (1 < n) /\ not (IsComposite n)`.  The point is that one of the two directions doesn't require the `1 < n` assumption, so this is slightly stronger.  This also applies to the next result.

theories/Spaces/Nat/Division.v
```diff
@@ -760,3 +793,130 @@ Proof.
...
+Proof.
```
jdchristensen, line 824: I think the proof of this is the composite of two general purpose results:
1) If `P : nat -> Type` and `forall a, P a -> a < n`, then `exists a, P a <-> list_exists P (seq n)`.
2) If in addition `P` takes values in decidable types, then `exists a, P a` is decidable.
I think it would be good to reorganize it like this, as then the technique can be used for other predicates as well.

theories/Spaces/Nat/Division.v
```diff
@@ -760,3 +793,130 @@ Proof.
...
+  destruct (IHn d _ _) as [p r].
+  exists p.
+  exact _.
+Defined.
+
+(** Any natural number can either be written as a product of primes or is zero. *)
+Definition prime_factorization@{} n
+  : (exists (l : list Prime),
+      n = fold_right (fun (p : Prime) n => nat_mul p n) 1 l)
+    + (n = 0).
```
jdchristensen, line 900: Maybe instead of `foo + (n = 0)` we should write it as `(0 < n) -> foo`?  This would be the usual statement that every positive natural number is a product of primes.

jdchristensen:
My last commit simplifies the proof of the prime factorization, in part by factoring out some things, but in part but using some shorter arguments.   Probably some things could be cleaned up further.

theories/Basics/Overture.v
Alizter: I think this stems from `A : Type` and `R : A -> A -> Type` so these two universes are generally unrelated. When we pass something like `iff` it has to land in a strictly larger universe because we don't have algebraic universes which causes issues here. On the other hand, Equiv lands in `max` (an algebraic universe) so it has no problem I believe (or maybe it does). This case is unproblematic for say `paths` since there is only one universe.

theories/Spaces/Nat/Division.v
Alizter: Did you mean to write this comment for decidability of `IsComposite`?

Alizter: I've simplified the proof of decidability of is composite.

theories/Basics/Overture.v
Alizter: But to answer your question, I would assume algebraic universes solves this issue.

theories/Spaces/Nat/Division.v
Alizter: It was trickier than I expected, since I had to introduce some demorgan laws, but it seems to have worked.

theories/Spaces/Nat/Division.v
jdchristensen: Yes, I meant the decidability of IsComposite.  Thanks for understanding what I meant!

theories/Spaces/Nat/Division.v
```diff
@@ -673,6 +673,22 @@ Class IsPrime (n : nat) : Type0 := {
 
 Definition issig_IsPrime n : _ <~> IsPrime n := ltac:(issig).
 
+Global Instance ishprop_isprime `{Funext} n : IsHProp (IsPrime n).
```
jdchristensen, line 676: If it would be useful to have `IsPrime` be an hprop without Funext, there are logically equivalent definitions that would achieve this.  But I don't know if there's any reason to do so.

Alizter: We could truncate, but juggling around the truncation can be annoying.

jdchristensen: I wasn't thinking of truncating.  We could probably do something like form the list of natural numbers which are divisors of n and require that the length of the list be 2.  But it would still be awkward compared to what we have.

theories/Types/Sum.v
```diff
@@ -1049,3 +1043,33 @@ Proof.
...
+  - intros np.
+    destruct (dec A) as [a|na].
+    + exact (inr (fun b => np (a, b))).
+    + exact (inl na).
+  - intros [na|nb] [a b].
+    + exact (na a).
+    + exact (nb b).
+Defined.
+
+Definition iff_contradiction A : A * ~A <-> Empty.
```
jdchristensen, line 1070: This doesn't really belong in Types/Sum.v.  Maybe we need a Types/Iff.v file for everything involving `iff`?  (Or Basics/Iff.v, if dependencies allow it.)

jdchristensen (commented):
I still need a bit more time to finish reviewing this, but I'll do it soon.

jdchristensen:
Ok, this LGTM modulo gathering most `iff` results to one spot.  I'd say that the results that are about `iff` itself would make sense in their own file, whereas results where `iff` is simply a hypothesis should stay where they are (like `decidable_iff`).

Alizter:
I'm having a bit of trouble with the dependencies of `Iff.v` and `Decidable.v`. Some of the De Morgan laws require decidability as an assumption, however the decidability file requires some facts about `Iff.v`. For now, I will put the laws requiring decidability in Decidable. I don't see a clear way to make this separation currently.

Alizter:
@jdchristensen Is this what you had in mind?

jdchristensen:
Yes, that looks like a fine way to organize it.  Alternatively, the Decidable class could be defined in Overture.v.  In many cases, that's the only thing needed for users of Decidable.v, and it would follow the pattern of putting some of the very basic definitions in Overture.v.  But I also think it's fine like this.

I noticed that the dual De Morgan's law only needs A to be decidable with the existing proof, so I dropped the assumption that B is decidable, and also added the symmetrical version.

PR #2065, opened 2024-08-29
Nat: theory about factorials

Alizter:
We develop some theory about factorials in a new file `Spaces/Nat/Factorial.v`. This will be useful for some upcoming work on binomial coefficients.

Notably we show that `(n! * m! | (n + m)!)` which is actually quite a tricky proof to get right. We have to induct on `n + m` rather than just `n` or `m`.

This is a draft for now as it depends on #2063.

### DIFF
--- a/theories/Spaces/Nat/Factorial.v
+++ b/theories/Spaces/Nat/Factorial.v
@@ -1,5 +1,5 @@
 Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids
-  Basics.Decidable Spaces.Nat.Core Spaces.Nat.Division.
+  Basics.Decidable Spaces.Nat.Core Spaces.Nat.Division Tactics.EvalIn.
 
 Local Set Universe Minimization ToSet.
 Local Unset Elimination Schemes.
@@ -33,10 +33,12 @@ Defined.
 (** ** Divisibility *)
 
 (** Any number less than or equal to [n] divides [factorial n]. *)
-Global Instance nat_divides_facttorial_factor n m
+Global Instance nat_divides_factorial_factor n m
   : 0 < n -> n <= m -> (n | factorial m).
 Proof.
-  intros H1 H2; destruct H1; induction H2; exact _.
+  intros [] H2.
+  1: exact _.
+  induction H2; exact _.
 Defined.
 
 (** [factorial] is a monotone function from [nat] to [nat] with respect to [<=] and divides. *)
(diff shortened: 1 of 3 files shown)


theories/Spaces/Nat/Factorial.v
```diff
@@ -0,0 +1,87 @@
...
+Defined.
+
+(** ** Divisibility *)
+
+(** Any number less than or equal to [n] divides [factorial n]. *)
+Global Instance nat_divides_factorial_factor n m
+  : 0 < n -> n <= m -> (n | factorial m).
+Proof.
+  intros [] H2.
+  1: exact _.
```
jdchristensen, line 40: There's a repeated idiom where a hypothesis of the form `0 < n` is used to split into two cases, one where `n` is replaced with `1` and one where `n` is of the form `k.+1` and we know `0 < k`.  But oten the `n = 1` case is also handled by the other case, since all that is used is that `n` is a successor.  Is there any easy way to instead use a hypothesis `0 < n` to produce a single goal where `n` is a successor?

theories/Spaces/Nat/Factorial.v
```diff
@@ -0,0 +1,87 @@
...
+  revert k n m p; snrapply nat_ind_strong; hnf; intros k IH n m p.
+  destruct k.
+  { apply equiv_nat_add_zero in p.
+    destruct p as [p q].
+    destruct p^, q^.
+    exact _. }
+  rewrite_refl nat_factorial_succ.
+  rewrite <- p.
+  rewrite nat_dist_r.
+  assert (helper : forall n' m' (p' : n' + m' = k.+1), (factorial n' * factorial m' | n' * factorial k)).
```
jdchristensen, line 65: To avoid duplicated code, I proved this helper lemma.  This helper lemma might be of separate interest, in which case we could do a simultaneous induction of a statement like:
```coq
Definition foo n m k (p : n + m = k)
  : (factorial n * factorial m | factorial k) * (factorial n * factorial m | n * factorial (pred k).
```
or something like that, and deduce the two separate facts from this one.

theories/Spaces/Nat/Factorial.v
```diff
@@ -0,0 +1,87 @@
+Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids
+  Basics.Decidable Spaces.Nat.Core Spaces.Nat.Division Tactics.EvalIn.
+
+Local Set Universe Minimization ToSet.
+Local Unset Elimination Schemes.
```
Alizter, line 5:
```suggestion
```

theories/Spaces/Nat/Factorial.v
```diff
@@ -0,0 +1,91 @@
+Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids
+  Basics.Decidable Spaces.Nat.Core Spaces.Nat.Division.
+
+Local Set Universe Minimization ToSet.
+Local Unset Elimination Schemes.
+
+Local Open Scope nat_scope.
+
+(** * Factorials *)
+
```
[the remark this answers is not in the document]
Alizter: I think I will move it here.

theories/Spaces/Nat/Factorial.v
Alizter: I suppose `destruct n; only 1: contradiction (lt_irrefl _ _ (_ : 0 < n))` would do what you want.

theories/Spaces/Nat/Factorial.v
Alizter: That lemma doesn't look too useful on its own, so I'm happy with keeping it as an assert for the time being.

theories/Spaces/Nat/Factorial.v
jdchristensen: What about a lemma of type `(forall n, P n.+1) -> (forall n, 0 < n, P n)`?  I think that would work cleanly and would be useful in many situations.  Alternatively, this slight variant of what you suggested works:
```coq
  intro H; destruct n; only 1: contradiction (lt_irrefl _ H); clear H.
```
But I'd prefer to get rid of boilerplate lines like that, and would also like to avoid a special tactic for a situation where a lemma could work instead.

jdchristensen: This PR LGTM, and the lemma I suggest could be part of another PR.  Or it could be added here.

Alizter: I'll make a separate PR for that, as it will allow us to get rid of that idiom in various places.

PR #2066, opened 2024-08-29
Functoriality of FreeGroup

Alizter:
In this PR we show that `FreeGroup` is a functor. Together with #2064, this allows us to show that `FreeAbGroup` is a functor also.


jdchristensen (approved):
LGTM.  (Makes me wonder if there is a way to state and prove a meta-theorem that says that if some construction has an induction principle, then it is a functor.  We keep repeating this idiom.  Or maybe a tactic that does it.)

Alizter:
I also wondered something similar, but I couldn't come up with anything useful.

Issue #2048, opened 2024-08-06
Binomial Coefficients


Alizter:
There are various situations in algebra where it can be useful to perform a binomial expansion of a term. This leads to a need for the binomial coefficients in order to write such terms down.

We should define the binomial coefficients in `theories/Spaces/Nat/Core.v` and call them `nat_choose`, since it is about the number of ways of choosing $k$ elements from a set of $n$. The factorial formula is taken as a common definition of the binomial coefficient:

$$\binom{n}{k} = \frac{n!}{k!(n-k)!}$$

However I would argue that this is a poor definition for formalisation.

1. It contains division, which is tricky to reason about.
2. It is inefficient, this can be fixed somewhat by replacing $\frac{n!}{(n-k)!}$ by a falling factorial.
3. It is tricky to show binomial identities from this formula.

I therefore propose we use the much simpler "Pascal's triangle" formula:

$$\binom{n+1}{k+1} = \binom{n}{k} + \binom{n}{k+1}$$

with the appropriate base cases. This definition is recursive, meaning it is better suited to reasoning by induction.

As a bonus it might be nice to show some identities involving the binomial coefficients, but the most useful identities appear as sums of binomial coefficients. We don't currently have sums of elements in `nat`. This will probably appear as a generalisation of `ab_sum` for monoids and then semirings. We do however have `ab_sum` and `rng_sum` so such identities could be shown in an abelian group or ring.

The binomial theorem for commutative (semi)rings can be left for a later issue.

PR #2070, opened 2024-08-30
Nat: binomial coefficients

Alizter:
In this PR we introduce binomial coefficients as discussed in #2048. We prove basic identities about binomial coefficients, including the factorial formula.

In order to prove all of this, we also add a few more division lemmas.

### DIFF
--- a/theories/Basics/Notations.v
+++ b/theories/Basics/Notations.v
@@ -50,7 +50,7 @@ Reserved Notation "x < y <= z" (at level 70, y at next level).
 Reserved Notation "x + y" (at level 50, left associativity).
 Reserved Notation "x - y" (at level 50, left associativity).
 Reserved Notation "x * y" (at level 40, left associativity).
-Reserved Notation "x / y" (at level 40, left associativity).
+Reserved Notation "x / y" (at level 40, no associativity).
 Reserved Notation "- x" (at level 35, right associativity).
 Reserved Notation "/ x" (at level 35, right associativity).
 Reserved Notation "x ^ y" (at level 30, right associativity).
--- a/theories/Categories/Comma.v
+++ b/theories/Categories/Comma.v
@@ -1,6 +1,7 @@
 (** * Comma Categories *)
 (** Since there are only notations in [Comma.Notations], we can just export those. *)
 Local Set Warnings Append "-notation-overridden".
+Require Import Basics.Notations.
 Require Export Comma.Notations.
 
 (** ** Definitions *)
--- a/theories/Classes/theory/ua_first_isomorphism.v
+++ b/theories/Classes/theory/ua_first_isomorphism.v
@@ -4,6 +4,7 @@
     identification theorem [id_first_isomorphism] follows. *)
 
 Require Import
+  Basics.Overture
   HSet
   Colimits.Quotient
   Classes.interfaces.canonical_names
--- a/theories/Classes/theory/ua_first_isomorphism.v
+++ b/theories/Classes/theory/ua_first_isomorphism.v
@@ -4,7 +4,7 @@
     identification theorem [id_first_isomorphism] follows. *)
 
 Require Import
-  Basics.Overture
+  Basics.Notations
   HSet
   Colimits.Quotient
   Classes.interfaces.canonical_names
--- a/theories/Classes/theory/ua_second_isomorphism.v
+++ b/theories/Classes/theory/ua_second_isomorphism.v
@@ -1,7 +1,7 @@
 (** The second isomorphism theorem [isomorphic_second_isomorphism]. *)
 
 Require Import
-  Basics.Overture
+  Basics.Notations
   HSet
   Colimits.Quotient
   Classes.interfaces.canonical_names
--- a/theories/Classes/theory/ua_second_isomorphism.v
+++ b/theories/Classes/theory/ua_second_isomorphism.v
@@ -1,6 +1,7 @@
 (** The second isomorphism theorem [isomorphic_second_isomorphism]. *)
 
 Require Import
+  Basics.Overture
   HSet
   Colimits.Quotient
   Classes.interfaces.canonical_names
--- a/theories/Classes/theory/ua_third_isomorphism.v
+++ b/theories/Classes/theory/ua_third_isomorphism.v
@@ -1,7 +1,7 @@
 (** This file proves the third isomorphism theorem,
     [isomorphic_third_isomorphism]. *)
 Require Import
-  Basics.Overture
+  Basics.Notations
   Colimits.Quotient
   Classes.interfaces.canonical_names
   Classes.theory.ua_quotient_algebra
--- a/theories/Classes/theory/ua_third_isomorphism.v
+++ b/theories/Classes/theory/ua_third_isomorphism.v
@@ -1,6 +1,7 @@
 (** This file proves the third isomorphism theorem,
     [isomorphic_third_isomorphism]. *)
 Require Import
+  Basics.Overture
   Colimits.Quotient
   Classes.interfaces.canonical_names
   Classes.theory.ua_quotient_algebra
--- a/theories/Spaces/Nat/Binomial.v
+++ b/theories/Spaces/Nat/Binomial.v
@@ -85,23 +85,22 @@ Defined.
 Definition nat_choose_factorial@{} n m
   : m <= n -> nat_choose n m = factorial n / (factorial m * factorial (n - m)).
 Proof.
-  revert n m; snrapply nat_ind_strong; hnf; intros n IHn m H.
-  destruct H as [|n H].
+  revert n m; apply nat_double_induction_leq; intro n.
+  (* The case when [m = 0]. *)
+  { rewrite nat_mul_one_l.
+    rewrite nat_sub_zero_r.
+    symmetry; rapply nat_div_cancel. }
+  (* The case when [m = n]. *)
   { rewrite nat_sub_cancel.
     rewrite nat_mul_one_r.
     rewrite nat_div_cancel.
     1: nrapply nat_choose_diag.
     exact _. }
-  destruct m.
-  { rewrite nat_choose_zero_r.
-    rewrite nat_sub_zero_r.
-    rewrite nat_mul_one_l.
-    symmetry.
-    apply nat_div_cancel.
-    exact _. }
+  (* The case with [n.+1] and [m.+1] with [m < n] and an induction hypothesis. *)
+  intros m H IHn.
   rewrite nat_choose_succ.
   rewrite 2 IHn.
-  2-5: exact _.
+  2,3: exact _.
   rewrite <- (nat_div_cancel_mul_l _ _ m.+1). 
   2: exact _.
   rewrite nat_mul_assoc.
--- a/theories/Spaces/Nat/Binomial.v
+++ b/theories/Spaces/Nat/Binomial.v
@@ -46,12 +46,6 @@ Proof.
   exact IHn.
 Defined.
 
-(** There are no ways to choose more than [0] elements from [0] elements. *)
-Definition nat_choose_zero_l@{} n : 0 < n -> nat_choose 0 n = 0.
-Proof.
-  apply nat_choose_lt.
-Defined.
-
 (** There is only one way to choose [0] elements from any number of elements. *)
 Definition nat_choose_zero_r@{} n : nat_choose n 0 = 1.
 Proof.
--- a/theories/Spaces/Nat/Binomial.v
+++ b/theories/Spaces/Nat/Binomial.v
@@ -98,7 +98,7 @@ Proof.
     exact _. }
   (* The case with [n.+1] and [m.+1] with [m < n] and an induction hypothesis. *)
   intros m H IHn.
-  rewrite nat_choose_succ.
+  rewrite_refl nat_choose_succ.
   rewrite 2 IHn.
   2,3: exact _.
   rewrite <- (nat_div_cancel_mul_l _ _ m.+1). 
--- a/theories/Spaces/Nat/Binomial.v
+++ b/theories/Spaces/Nat/Binomial.v
@@ -31,9 +31,9 @@ Definition nat_choose_succ@{} n m
 (** The binomial coefficient is zero if [m] is greater than [n]. *)
 Definition nat_choose_lt@{} n m : n < m -> nat_choose n m = 0.
 Proof.
-  revert n m; snrapply nat_ind_strong; hnf; intros n IHn m H.
-  destruct H, n; only 1,3: reflexivity;
-  exact (ap011 nat_add (IHn _ _ _ _) (IHn _ _ _ _)).
+  revert m; induction n; hnf; intros m H; destruct H.
+  1, 2: reflexivity.
+  1, 2: exact (ap011 nat_add (IHn _ _) (IHn _ _)).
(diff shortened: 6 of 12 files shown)


theories/Spaces/Nat/Binomial.v
```diff
@@ -0,0 +1,158 @@
...
+  - rapply lt_lt_leq_trans.
+  - rewrite <- nat_dist_r.
+    rewrite nat_add_succ_l.
+    rewrite nat_add_sub_r_cancel.
+    2: exact _.
+    rewrite <- nat_factorial_succ.
+    reflexivity.
+Defined.
+
+Definition nat_choose_succ_mul@{} n m
```
Alizter, line 135: TODO: I need to write a comment here.

theories/Spaces/Nat/Binomial.v
```diff
@@ -0,0 +1,158 @@
...
+  2,3: exact _.
+  rewrite nat_div_mul_l.
+  2: exact _.
+  rewrite nat_div_div_l; rewrite (nat_mul_comm _ m.+1), nat_mul_assoc.
+  1: reflexivity.
+  change (factorial m.+1 * factorial (n.+1 - m.+1) | factorial n.+1).
+  exact _.
+Defined.
+
+Definition nat_choose_sub@{} n m
```
Alizter, line 149: And here

theories/Spaces/Nat/Division.v
```diff
@@ -423,6 +429,48 @@ Proof.
   nrapply nat_div_cancel_mul_l.
 Defined.
 
+(** We can swap the order of division and multiplication on the left under certain conditions. *)
+Definition nat_div_mul_l n m k : (m | n) -> k * (n / m) = (k * n) / m.
```
Alizter, line 433: This condition is sufficient since `2 * (3 / 2) = 2 =/= 3 = (2 * 3) / 2`. It's probably necessary but I'm not certain.

jdchristensen: The proof shows that the hypothesis `(m | n)` is sufficient.  Your example shows that it is necessary (at least for general `k`) since `m * (n / m)` is not equal to `(m * n) / m` when `m` doesn't divide `n`.  (The latter is `n` while the former is strictly less than `n`.)

theories/Spaces/Nat/Division.v
```diff
@@ -423,6 +429,48 @@ Proof.
...
+
+(** We can swap the order of division and multiplication on the right under certain conditions. *)
+Definition nat_div_mul_r n m k : (m | n) -> (n / m) * k = n * k / m.
+Proof.
+  rewrite 2 (nat_mul_comm _ k).
+  snrapply nat_div_mul_l.
+Defined.
+
+(** Dividing a quotient is the same as dividing by the product of the divisors. *)
+Definition nat_div_div_l n m k : (m * k | n) -> (n / m) / k = n / (m * k).
```
jdchristensen, line 460: I don't think the `(m * k | n)` hypothesis is needed here.

jdchristensen (commented):
I still need to review Binomial.v.

theories/Spaces/Nat/Division.v
```diff
@@ -450,7 +450,7 @@ Proof.
 Defined.
 
 (** We can swap the order of division and multiplication on the right under certain conditions. *)
-Definition nat_div_mul_r n m k : (m | n) -> (n / m) * k = n * k / m.
+Definition nat_div_mul_r n m k : (m | n) -> (n / m) * k = (n * k) / m.
```
Alizter, line 453: I need to work out how to get these to be parenthesised all the time. `/` and `*` are defined at the same level which makes reading terms very confusing.

theories/Spaces/Nat/Binomial.v
```diff
@@ -0,0 +1,158 @@
...
+  1: rapply nat_choose_lt.
+  lhs nrapply nat_add_zero_r.
+  exact IHn.
+Defined.
+
+(** There are no ways to choose more than [0] elements from [0] elements. *)
+Definition nat_choose_zero_l@{} n : 0 < n -> nat_choose 0 n = 0.
+Proof.
+  apply nat_choose_lt.
+Defined.
```
jdchristensen, line 53: Since this is just a special case of `nat_choose_lt`, maybe we should delete it?

theories/Spaces/Nat/Division.v
Alizter: The proof I have used `nat_div_mul_l` which requires that assumption. I had an earlier attempt which required the equivalent `(m | n / k)`. Do you see another way of proving this?

I struggled to come up with a counter example, so presumably you are correct.

Alizter: I also tried to use `nat_div_mod_spec` twice which leads to
```
m * k * (n / (m * k)) + n mod (m * k) = m * k * (n / (m * k)) + 0
```
this is close, but getting `n mod (m * k)` to equal zero is the assumption we are trying to drop.

theories/Spaces/Nat/Binomial.v
Alizter: Done.

theories/Spaces/Nat/Binomial.v
Alizter: I couldn't come up with a comment for this.

theories/Spaces/Nat/Division.v
Alizter: I've fixed the printing of `/` now by marking it as `no associativity` which means parentheses will always be printed. In order to do this, I had to fix some dodgy `Require`s elsewhere in the library which were causing notation conflicts.

theories/Spaces/Nat/Binomial.v
Alizter: @jdchristensen Any ideas how you might describe this?

theories/Spaces/Nat/Division.v
Alizter: I've managed to weaken the hypothesis to `(m | n)`, however I still feel like it should be possible to get rid of this. I haven't been able to do it however.

theories/Spaces/Nat/Division.v
```diff
@@ -392,13 +405,19 @@ Defined.
...
   destruct n; trivial.
-  symmetry.
   snrapply (nat_mod_unique _ _ 1); only 1: exact _.
   lhs nrapply nat_add_zero_r.
   nrapply nat_mul_one_r.
 Defined.
 
-(** ** Further Properties of Division and Modulo *)
+(** A number can be corrected so that it is divisible by subtracting the modulo. *)
+Definition nat_divides_sub_mod n m : (n | m - m mod n).
```
Alizter, line 414:
```suggestion
Global Instance nat_divides_sub_mod n m : (n | m - m mod n).
```

theories/Basics/Notations.v
```diff
@@ -50,7 +50,7 @@ Reserved Notation "x < y <= z" (at level 70, y at next level).
 Reserved Notation "x + y" (at level 50, left associativity).
 Reserved Notation "x - y" (at level 50, left associativity).
 Reserved Notation "x * y" (at level 40, left associativity).
-Reserved Notation "x / y" (at level 40, left associativity).
```
jdchristensen, line 53: Does `no associativity` mean that Coq will declare expressions like `x / y / z` invalid?  If so, that seems reasonable.

theories/Categories/Comma.v
```diff
@@ -1,6 +1,7 @@
 (** * Comma Categories *)
 (** Since there are only notations in [Comma.Notations], we can just export those. *)
 Local Set Warnings Append "-notation-overridden".
+Require Import Basics.Notations.
```
jdchristensen, line 4: Just curious why this line needed to be added to many files.

theories/Classes/theory/ua_first_isomorphism.v
```diff
@@ -4,6 +4,7 @@
     identification theorem [id_first_isomorphism] follows. *)
 
 Require Import
+  Basics.Overture
```
jdchristensen, line 7: Add this line.

theories/Spaces/Nat/Division.v
jdchristensen: It might be a few days before I respond.  I'll be on a long flight tomorrow and might get a chance to see if I can improve this during the flight.

theories/Basics/Notations.v
Alizter: Not quite. `no associativity` is essentially just a printing modification. It is treated the same as `left associativity`, but will always be printed with parentheses. There is nothing stopping Coq from parsing `x / y / z`. That's not a problem, as it should be up to us to write the correct expression, and more importantly the expression will appear correctly in proofs.

theories/Categories/Comma.v
Alizter: I'm not certain, but for some reason if I don't add it, Coq thinks that `level 40` is `right associativity` whereas in `Basics.Notations` it should be `left associativity`. This is fixed by adding this line.

theories/Classes/theory/ua_first_isomorphism.v
Alizter: This should be `Basics.Notations` instead. I will fix it.

Alizter:
```suggestion
  Basics.Notations
```

theories/Classes/theory/ua_second_isomorphism.v
```diff
@@ -1,6 +1,7 @@
 (** The second isomorphism theorem [isomorphic_second_isomorphism]. *)
 
 Require Import
+  Basics.Overture
```
Alizter, line 4:
```suggestion
  Basics.Notations
```

theories/Classes/theory/ua_third_isomorphism.v
```diff
@@ -1,6 +1,7 @@
 (** This file proves the third isomorphism theorem,
     [isomorphic_third_isomorphism]. *)
 Require Import
+  Basics.Overture
```
Alizter, line 4:
```suggestion
  Basics.Notations
```

jdchristensen:
Just a note that this document provides a nice summary of what's known about binomials in Coq:  https://inria.hal.science/hal-04550762v1/document

jdchristensen:
I just pushed four commits:￼

- Nat/Division: remove some divisibility hypotheses, using new lemmas
  This removes divisibility hypotheses in two places.  I also added a result about multiplication being injective, but didn't end up using it.  And I didn't always add the symmetrical versions of results.

- Nat/Binominal: make nat_choose compute on edge cases as well
  I used nested `Fixpoint`s to make `nat_choose` compute on all of the cases of its definition.  An unfortunate side-effect is that `cbn` and `simpl` don't generate good terms, so it's not 100% clear that this is worth it.  Let me know what you think.

- Nat/Binomial: remove hypothesis from nat_choose_succ_mul and add comment
  Removed a hypothesis here.  Also, note that this could be taken to be the definition of nat_choose.  It would compute nicely, is efficient, and probably implies both the factorial definition and the Pascal identity fairly easily.  But a disadvantage is that it isn't clear that it produces natural numbers.

- Nat/*: abstract out the double induction used in nat_choose_factorial
  I didn't understand why strong induction was used here, when the induction hypothesis was only used for the previous n; but it was tricky to convince Coq to do this any other way, so I encoded my method in a separate lemma. Maybe there's an even easier way?

I think this all looks good to merge, but let me know if there is any feedback.

Alizter:
> * Nat/Division: remove some divisibility hypotheses, using new lemmas
>   This removes divisibility hypotheses in two places.  I also added a result about multiplication being injective, but didn't end up using it.  And I didn't always add the symmetrical versions of results.

That's good that we have injectivity of multiplication.

> * Nat/Binominal: make nat_choose compute on edge cases as well
>   I used nested `Fixpoint`s to make `nat_choose` compute on all of the cases of its definition.  An unfortunate side-effect is that `cbn` and `simpl` don't generate good terms, so it's not 100% clear that this is worth it.  Let me know what you think.

I think this is fine. I would actually prefer if binomial coefficients are more rigid than they currently are, perhaps using `simpl never` so that we don't rely on `cbn` behaviour. I think its always helpful to be explicit about how you want to rewrite such terms.

> * Nat/Binomial: remove hypothesis from nat_choose_succ_mul and add comment
>   Removed a hypothesis here.  Also, note that this could be taken to be the definition of nat_choose.  It would compute nicely, is efficient, and probably implies both the factorial definition and the Pascal identity fairly easily.  But a disadvantage is that it isn't clear that it produces natural numbers.

Yes, this is something I mentioned briefly in #2048. Having to prove that we always get an integer and that the division definition is "well-defined" seems like more work than sticking to the simple recursive additive description. If we are interested in actually computing some larger values, it might be worth introducing a variant that computes this way. For now, I don't think we will have many "concrete" examples of binomial coefficients.

In the undergraduate algebra texts that I checked, the factorial definition is always preferred and Pascal's identity as a definition is left as an exercise. However in the history of the subject, Pascal's identity came way before Pascal around 1100 in India and the factorial definition appears to likely have been introduced by Newton, though it is likely the statement of `nat_choose_succ_mul` was known. There is some evidence to suggest that the Ancient Greeks where aware of similar identities for certain values of `n`. From what I can recall.

> * Nat/*: abstract out the double induction used in nat_choose_factorial
>   I didn't understand why strong induction was used here, when the induction hypothesis was only used for the previous n; but it was tricky to convince Coq to do this any other way, so I encoded my method in a separate lemma. Maybe there's an even easier way?

Strong induction seems to be pretty common here for on-paper proofs. I will need to research a bit what "double induction" lemmas are possible, since I believe there a few variants.

> I think this all looks good to merge, but let me know if there is any feedback.

I think have some minor whitespace improvements that I'll push. I'll merge when I deem it ready.

jdchristensen:
> I think this is fine. I would actually prefer if binomial coefficients are more rigid than they currently are, perhaps using `simpl never` so that we don't rely on `cbn` behaviour. I think its always helpful to be explicit about how you want to rewrite such terms.

`simpl never` might work well.

> >   Also, note that this could be taken to be the definition of nat_choose. 

> Having to prove that we always get an integer and that the division definition is "well-defined" seems like more work than sticking to the simple recursive additive description. 

I agree.

> >   I didn't understand why strong induction was used here, when the induction hypothesis was only used for the previous n

> Strong induction seems to be pretty common here for on-paper proofs.

But I think we should avoid it when we aren't actually doing strong induction, as it is confusing to the reader.

In this case, I suspect that we could use some lemmas about proving statements of the form `forall n, n >= m -> P n` (with `m` in the context) to avoid fiddling with induction on `n` and destructing the inequality.

Alizter:
As a side note, I had a go at proving something like:
```coq
Definition nat_div_div_r n m k : k < m -> n / (m / k) = n * k / m.
Proof.
  intros H.
  destruct (nat_zero_or_gt_zero k) as [[] | kp].
  1: by rewrite nat_mul_zero_r, nat_div_zero_l.
  destruct (nat_zero_or_gt_zero m) as [[] | mp].
  1: by rewrite nat_div_zero_l.
```

But I wasn't able to work out how to continue. It's not needed, but it would complement our other division lemmas nicely.

Alizter:
`simpl never` works well now. The binomial terms don't get unfolded, which should allow us to use `rewrite_refl` when needed.

PR #2072, opened 2024-09-01
Basics on monoids

Alizter:
We introduce a definition of monoid and monoid homomorphism. We show that it forms a category with equivalences and products.


theories/Algebra/Monoids/Monoid.v
```diff
@@ -0,0 +1,268 @@
...
+Require Export Classes.interfaces.canonical_names (SgOp, sg_op, One, one,
+    MonUnit, mon_unit, LeftIdentity, left_identity, RightIdentity, right_identity,
+    Negate, negate, Associative, simple_associativity, associativity,
+    LeftInverse, left_inverse, RightInverse, right_inverse, Commutative, commutativity).
+Export canonical_names.BinOpNotations.
+Require Export Classes.interfaces.abstract_algebra (IsSemiGroup(..), sg_set, sg_ass,
+    IsMonoid(..), monoid_left_id, monoid_right_id, monoid_semigroup,
+    IsMonoidPreserving(..), monmor_unitmor, monmor_sgmor,
+    IsSemiGroupPreserving, preserves_sg_op, IsUnitPreserving, preserves_mon_unit).
+Require Import Classes.theory.groups.
```
jdchristensen, line 17: It would be nice to avoid depending on Classes.theory.groups.  Would it be worth separately out the parts about monoids to Classes.theory.monoids?

jdchristensen (approved):

theories/Algebra/Monoids/Monoid.v
Alizter: I've moved it to abstract_algebra. I want to refrain introducing more stuff in Classes.theory and would rather the things there appear in Algebra directly without having to take the classes detour. Since these lemmas on morphisms are of general interest, I've moved them to abstract_algebra instead.

PR #2074, opened 2024-09-03
prove nat_div_div_r

Alizter:
I'm not sure if the side condition can be improved.

### DIFF
--- a/theories/Spaces/Nat/Division.v
+++ b/theories/Spaces/Nat/Division.v
@@ -556,6 +556,15 @@ Proof.
   snrapply nat_div_mul_l.
 Defined.
 
+Definition nat_div_sub_mod n m : n / m = (n - n mod m) / m.
+Proof.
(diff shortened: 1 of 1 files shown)


Alizter:
I've added a variant of `nat_div_div_r` which doesn't have the divisibility assumption, but it does mean that `m` gets modified by subtracting `m mod k`.

jdchristensen (approved):

theories/Spaces/Nat/Division.v
```diff
@@ -579,6 +588,26 @@ Proof.
     symmetry; apply nat_div_mod_spec.
 Defined.
 
+(** Dividing a number by a quotient is the same as dividing the product of the number with the denominator the quotient by the numerator of the quotient. *)
```
jdchristensen, line 591:
```suggestion
(** Dividing a number by a quotient is the same as dividing the product of the number with the denominator of the quotient by the numerator of the quotient. *)
```

PR #2079, opened 2024-09-07
tensor product distributes over direct sum

Alizter:
We show that tensor products distribute over sums.

### DIFF
--- a/theories/Algebra/AbGroups/Biproduct.v
+++ b/theories/Algebra/AbGroups/Biproduct.v
@@ -224,7 +224,7 @@ Defined.
 (** *** Lemmas for working with biproducts *)
 
 (* Maps out of biproducts are determined on the two inclusions. *)
-Lemma equiv_path_biprod_corec `{Funext} {A B X : AbGroup} (phi psi : ab_biprod A B $-> X)
+Lemma equiv_ab_biprod_ind_homotopy `{Funext} {A B X : AbGroup} (phi psi : ab_biprod A B $-> X)
   : ((phi $o ab_biprod_inl == psi $o ab_biprod_inl) * (phi $o ab_biprod_inr == psi $o ab_biprod_inr))
       <~> phi == psi.
 Proof.
--- a/theories/Algebra/AbGroups/Biproduct.v
+++ b/theories/Algebra/AbGroups/Biproduct.v
@@ -23,6 +23,45 @@ Definition ab_biprod_inr {A B : AbGroup} : B $-> ab_biprod A B := grp_prod_inr.
 Definition ab_biprod_pr1 {A B : AbGroup} : ab_biprod A B $-> A := grp_prod_pr1.
 Definition ab_biprod_pr2 {A B : AbGroup} : ab_biprod A B $-> B := grp_prod_pr2.
 
+(** Pairs in biproducts can be decomposed *)
+Definition ab_biprod_decompose {A B : AbGroup} (a : A) (b : B)
+  : (a, b) = ((a, group_unit) : ab_biprod A B) + (group_unit, b).
+Proof.
+  snrapply path_prod; symmetry.
+  - snrapply grp_unit_r.
+  - snrapply grp_unit_l.
+Defined.
+
+Definition ab_biprod_ind_hprop {A B : AbGroup}
+  (P : ab_biprod A B -> Type)
+  (H : forall x, IsHProp (P x))
+  (Hinl : forall a, P (ab_biprod_inl a))
+  (Hinr : forall b, P (ab_biprod_inr b))
+  (Hop : forall x y, P x -> P y -> P (x + y))
+  : forall x, P x.
+Proof.
+  intros [a b].
+  snrapply ((ab_biprod_decompose a b)^ # _).
+  apply Hop.
+  - exact (Hinl a).
+  - exact (Hinr b).
+Defined.
+
+Definition ab_biprod_ind_homotopy {A B C : AbGroup}
+  {f g : ab_biprod A B $-> C}
+  (Hinl : f $o ab_biprod_inl $== g $o ab_biprod_inl)
+  (Hinr : f $o ab_biprod_inr $== g $o ab_biprod_inr)
+  : f $== g.
+Proof.
+  rapply ab_biprod_ind_hprop. 
+  - exact Hinl.
+  - exact Hinr.
+  - intros x y p q.
+    lhs nrapply grp_homo_op.
+    rhs nrapply grp_homo_op.
+    f_ap.
+Defined.
+
 (** Recursion principle *)
 Proposition ab_biprod_rec {A B Y : AbGroup}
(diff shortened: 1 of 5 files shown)


theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -23,6 +23,45 @@ Definition ab_biprod_inr {A B : AbGroup} : B $-> ab_biprod A B := grp_prod_inr.
 Definition ab_biprod_pr1 {A B : AbGroup} : ab_biprod A B $-> A := grp_prod_pr1.
 Definition ab_biprod_pr2 {A B : AbGroup} : ab_biprod A B $-> B := grp_prod_pr2.
 
+(** Pairs in biproducts can be decomposed *)
+Definition ab_biprod_decompose {A B : AbGroup} (a : A) (b : B)
+  : (a, b) = ((a, group_unit) : ab_biprod A B) + (group_unit, b).
```
jdchristensen, line 28: This is true for non-abelian groups as well, so it could be moved to Group.v and stated for grp_prod.

theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -23,6 +23,45 @@ Definition ab_biprod_inr {A B : AbGroup} : B $-> ab_biprod A B := grp_prod_inr.
...
+Proof.
+  snrapply path_prod; symmetry.
+  - snrapply grp_unit_r.
+  - snrapply grp_unit_l.
+Defined.
+
+Definition ab_biprod_ind_hprop {A B : AbGroup}
+  (P : ab_biprod A B -> Type)
+  (H : forall x, IsHProp (P x))
```
jdchristensen, line 37: This assumption isn't used.

theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -23,6 +23,45 @@ Definition ab_biprod_inr {A B : AbGroup} : B $-> ab_biprod A B := grp_prod_inr.
...
+  apply Hop.
+  - exact (Hinl a).
+  - exact (Hinr b).
+Defined.
+
+Definition ab_biprod_ind_homotopy {A B C : AbGroup}
+  {f g : ab_biprod A B $-> C}
+  (Hinl : f $o ab_biprod_inl $== g $o ab_biprod_inl)
+  (Hinr : f $o ab_biprod_inr $== g $o ab_biprod_inr)
+  : f $== g.
```
jdchristensen, line 54: This is already later in the file with the name `ab_biprod_corec_eta'`.  I don't know why that name has "corec" in it, since it seems to be about the recursion principle.  The same comment about naming applies to `equiv_path_biprod_corec`.  Also, both of those results seem to follow from `isequiv_ab_biprod_rec`.  Or maybe the proof of `isequiv_ab_biprod_rec` can be simplified by using these results?  There is definitely some overlap here.

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -807,4 +807,52 @@ Proof.
     reflexivity.
 Defined.
 
+(** ** Tensor products distribute over direct sums *)
+
+Definition ab_tensor_prod_dist_l {A B C : AbGroup}
+  : ab_tensor_prod A (ab_biprod B C)
+    $<~> ab_biprod (ab_tensor_prod A B) (ab_tensor_prod A C).
+Proof.
```
jdchristensen, line 815: Is it possible to make the goals easier to read in this proof?  Maybe the trick of naming the two functions would help?  And maybe some carefully `simpl` or `cbn` commands would help?

theories/Algebra/AbGroups/Biproduct.v
Alizter: I don't know about `isequiv_ab_biprod_rec` but I managed to show the others are special cases. I think this entire file needs some clean up as there are some duplicated results, but I don't really want to do that here.

theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -23,6 +23,35 @@ Definition ab_biprod_inr {A B : AbGroup} : B $-> ab_biprod A B := grp_prod_inr.
 Definition ab_biprod_pr1 {A B : AbGroup} : ab_biprod A B $-> A := grp_prod_pr1.
 Definition ab_biprod_pr2 {A B : AbGroup} : ab_biprod A B $-> B := grp_prod_pr2.
 
+Definition ab_biprod_ind_hprop {A B : AbGroup}
```
jdchristensen, line 26:
```suggestion
Definition ab_biprod_ind {A B : AbGroup}
```
(with a corresponding change when used)

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -807,4 +807,52 @@ Proof.
...
+  srapply (let f := _ in let g := _ in cate_adjointify f g _ _).
+  - snrapply ab_tensor_prod_rec.
+    + intros a bc.
+      exact (tensor a (fst bc), tensor a (snd bc)).
+    + intros a bc bc'.
```
jdchristensen, line 820:
```suggestion
    + intros a bc bc'; cbn beta.
```

theories/Algebra/AbGroups/TensorProduct.v
```diff
@@ -807,4 +807,52 @@ Proof.
...
+      snrapply path_prod'; snrapply tensor_dist_l.
+    + intros a a' bc.
```
jdchristensen, line 822:
```suggestion
    + intros a a' bc; cbn beta.
```

theories/Algebra/AbGroups/Biproduct.v
jdchristensen: Ok, that's fine.

theories/Algebra/AbGroups/Biproduct.v
```diff
@@ -175,33 +223,14 @@ Defined.
...
-Proof.
-  intros h k.
-  intros [a b].
-  refine (ap f (ab_biprod_decompose _ _) @ _ @ ap g (ab_biprod_decompose _ _)^).
-  exact (grp_homo_op_agree f g (h a) (k b)).
-Defined.
-
 (* Maps out of biproducts are determined on the two inclusions. *)
-Lemma equiv_path_biprod_corec `{Funext} {A B X : AbGroup} (phi psi : ab_biprod A B $-> X)
+Lemma equiv_ab_biprod_ind_homotopy `{Funext} {A B X : AbGroup} (phi psi : ab_biprod A B $-> X)
```
jdchristensen, line 227: I think this should be moved to right after `ab_biprod_ind_homotopy`.

jdchristensen (approved):
LGTM after the minor cleanups I just indicated.

Alizter:
FTR we will eventually prove a more general version of this lemma for indexed products/biproducts but this will probably end up requiring funext if we go through the adjunction route. So its good to have this simple case explicitly written out.

PR #2082, opened 2024-09-07
Factorial: except for 0! = 1!, factorial is strictly monotone

jdchristensen:
This is left over from some work on #2070, as it didn't end up being needed there.

### DIFF
--- a/theories/Spaces/Nat/Factorial.v
+++ b/theories/Spaces/Nat/Factorial.v
@@ -37,6 +37,24 @@ Proof.
   induction n; exact _.
 Defined.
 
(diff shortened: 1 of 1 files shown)


Alizter:
Do you think it would be useful to introduce a notation for `factorial`? I think we cannot do `!` do to parsing clashes, however I have seen `` `! `` being used also.

Alizter (approved):

jdchristensen:
> Do you think it would be useful to introduce a notation for `factorial`? I think we cannot do `!` do to parsing clashes, however I have seen `` `! `` being used also.

I guess something like that would be ok.

PR #2083, opened 2024-09-12
cleanup FreeProduct.v

Alizter:
We use the relevant list lemmas in FreeProduct.v which allows us to remove some redundant lemmas.


jdchristensen (approved):
I had deja vu at first, but now I remember that we made the same changes in FreeGroup.v

PR #2085, opened 2024-09-12
Replace ab_biprod_corec_eta with grp_prod_corec_natural

jdchristensen:
Since we generalized ab_biprod_corec to grp_prod_corec, we should also generalize ab_biprod_corec_eta to grp_prod.  In the process, I renamed it from _eta to _natural.  It's a definitional equality that isn't used anywhere, but is probably worth keeping as a form of documentation.

I had to switch grp_prod_corec to use wildcat notation, and I switched a couple of nearby things as well.


Alizter (approved):

Alizter:
Switching `GroupHomomorphism` to `$->` will introduce a strictly larger universe in those expressions, but as we've said before we don't care when algebraic universes land.

PR #2086, opened 2024-09-12
add lemma turning 0 < n -> n.+1

Alizter:
If there are any more places this can be used, feel free to push here.

### DIFF
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -877,6 +877,19 @@ Proof.
   rewrite nat_mul_comm; exact _.
 Defined.
 
+(** ** Eliminating positive assumptions *)
+
+(** Sometimes we want to prove a predicate which assumes that [0 < x]. In that case, it suffices to prove it for [x.+1] instead. *)
+Definition lt_zero_ind (P : forall x, 0 < x -> Type)
+  (H : forall x l, P x.+1 l)
+  : forall x l, P x l.
+Proof.
+  intros x l.
+  destruct x.
+  1: contradiction (lt_irrefl _ l).
+  apply H.
+Defined.
+
 (** Alternative Characterizations of [<=] *)
 
 (** [n <= m] is equivalent to [(n < m) + (n = m)]. This justifies the name "less than or equal to". Note that it is not immediately obvious that the latter type is a hprop. *)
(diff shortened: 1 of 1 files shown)


theories/Spaces/Nat/Core.v
```diff
@@ -877,6 +877,19 @@ Proof.
...
+(** Sometimes we want to prove a predicate which assumes that [0 < x]. In that case, it suffices to prove it for [x.+1] instead. *)
+Definition lt_zero_ind (P : forall x, 0 < x -> Type)
+  (H : forall x l, P x.+1 l)
+  : forall x l, P x l.
+Proof.
+  intros x l.
+  destruct x.
+  1: contradiction (lt_irrefl _ l).
+  apply H.
+Defined.
```
jdchristensen, line 891: Typically the predicate won't depend on the proof that `0 < x` (since that's an hprop), so this form will be cleaner, eliminating a useless ` 0 < x.+1` hypothesis in the context:
```coq
Definition gt_zero_ind (P : nat -> Type)
  (H : forall n, P n.+1)
  : forall n (l : 0 < n), P n.
Proof.
  intros n l.
  destruct n.
  1: contradiction (lt_irrefl _ l).
  apply H.
Defined.
```
I also changed `lt_zero` to `gt_zero`, but we could instead change it to `zero_lt` if you want.

jdchristensen: (It's also possible to allow the predicate to depend on `l` and still eliminate the redundant hypothesis in the context, using the version below, but it's a bit uglier.)
```coq
Definition gt_zero_ind (P : forall n, 0 < n -> Type)
  (H : forall n, P n.+1 (leq_add_r 1 n))
  : forall n l, P n l.
Proof.
  intros n l.
  destruct n.
  1: contradiction (lt_irrefl _ l).
  destruct (path_ishprop (leq_add_r 1 n) l).
  apply H.
Defined.
```

Alizter: Yes, I'm unsure which one is more useful in practice since I (surprisingly) couldn't find many uses for it. The only use we have so far doesn't rely on the proof of `0 < n` so I've simplified it like you suggested above. I also had a version similar to your second version before, but it didn't seem right to choose `leq_add_r` specifically.

jdchristensen (approved):

PR #2089, opened 2024-09-13
localizations of commutative rings

Alizter:
In this PR we introduce localizations of commutative rings. This is based on a formalization of the Zariski topology I started back in 2020, but I think the localization is of independent interest.

FTR there is no good generalization of localization for noncommutative rings so I've stuck to commutative rings.

We begin by defining the notion of a multiplicative subset of a ring and give a few examples. We then define the underlying type of a ring localized with respect to a multiplicative subset as a quotient type of fractions. This generalises the field of fractions construction.

The rest of the work consists of showing that we have a commutative ring structure and that we have the expected recursion principle.

Along the way, we introduce various ring lemmas of general use.

### DIFF
--- a/theories/Algebra/Rings/Localization.v
+++ b/theories/Algebra/Rings/Localization.v
@@ -239,18 +239,16 @@ Section Localization.
 
   Instance mult_localization_type : Mult Localization_type.
   Proof.
-    intros x; srapply Localization_type_rec.
-    - intros f2; revert x.
-      srapply Localization_type_rec.
-      + intros f1.
-        apply loc_frac.
-        exact (frac_mult f1 f2).
-      + intros f1 f1' p.
-        by apply loc_frac_eq, frac_mult_wd_l.
-    - intros f2 f2' p; revert x.
-      srapply Localization_type_ind_hprop.
-      intros f1.
-      by apply loc_frac_eq, frac_mult_wd_r.
+    srapply Quotient_rec2.
+    - rapply fraction_eq_refl.
+    - cbn.
+      intros f1 f2.
+      exact (loc_frac (frac_mult f1 f2)).
+    - cbn beta.
+      intros f1 f1' p f2 f2' q.
+      transitivity (loc_frac (frac_mult f1' f2)).
+      + by apply loc_frac_eq, frac_mult_wd_l.
+      + by apply loc_frac_eq, frac_mult_wd_r.
   Defined.
 
   (** *** Zero element *)
--- a/theories/Algebra/Rings/Localization.v
+++ b/theories/Algebra/Rings/Localization.v
@@ -76,6 +76,7 @@ Section Localization.
     in_mult_subset_denominator : S denominator ;
   }.
 
+  (** We consider two fractions to be equal if we can rearrange the fractions as products and ask for equality upto some scaling factor from the multiplicative subset [S]. *)
   Definition fraction_eq : Relation Fraction.
   Proof.
     intros [n1 d1 ?] [n2 d2 ?].
@@ -95,21 +96,26 @@ Section Localization.
     exact (ap (1 *.) p).
   Defined.
 
+  (** Fraction equality is a reflexive relation. *)
   Definition fraction_eq_refl f1 : fraction_eq f1 f1.
   Proof.
     apply fraction_eq_simple.
     reflexivity.
   Defined.
 
+  (** The underlying type of the localization of a commutative ring is the quotient of the type of fractions by fraction equality. *)
   Definition Localization_type : Type := Quotient fraction_eq.
 
+  (** Given any fraction, we have an obvious inclusion into the localization type. *)
   Definition loc_frac : Fraction -> Localization_type
     := class_of fraction_eq.
 
+  (** In order to give an equality of included fractions, it suffices to show that the fractions are equal under fraction equality. *)
   Definition loc_frac_eq {f1 f2 : Fraction} (p : fraction_eq f1 f2)
     : loc_frac f1 = loc_frac f2
     := qglue p.
 
+  (** We can give an induction principle for the localization type. *)
   Definition Localization_type_ind (Q : Localization_type -> Type)
     `{forall x, IsHSet (Q x)}
     (frac : forall f, Q (loc_frac f))
--- a/theories/Algebra/Rings/Localization.v
+++ b/theories/Algebra/Rings/Localization.v
@@ -614,6 +614,75 @@ Section Localization.
     Defined.
   
   End Rec.
+  
+  (** Elements belonging to the multiplicative subset [S] of [R] become invertible in [rng_localization R S]. *)
+  Global Instance isinvertible_rng_localization (x : R) (Sx : S x)
+    : IsInvertible rng_localization (loc_in x).
+  Proof.
+    snrapply isinvertible_cring.
+    - exact (loc_frac (Build_Fraction 1 x Sx)).
+    - apply loc_frac_eq, fraction_eq_simple.
+      exact (rng_mult_assoc _ _ _)^.
+  Defined.
+  
+  (** As a special case, any denominator of a fraction must necesserily be invertible. *)
+  Global Instance isinvertible_denominator (f : Fraction)
+    : IsInvertible rng_localization (loc_in (denominator f)).
+  Proof.
+    snrapply isinvertible_rng_localization.
+    exact (in_mult_subset_denominator f).
+  Defined.
+  
+  (** We can factor any fraction as the multiplication of the numerator and the inverse of the denominator. *)
+  Definition fraction_decompose (f : Fraction)
+    : loc_frac f
+      = loc_in (numerator f) * inverse_elem (loc_in (denominator f)).
+  Proof.
+    apply loc_frac_eq, fraction_eq_simple.
+    nrapply rng_mult_assoc.
+  Defined.
+
+  Section Ind.
+  
+    Context (P : rng_localization -> Type)
+      (H : forall x, IsHProp (P x))
+      (Hin : forall x, P (loc_in x))
+      (Hinv : forall x (H : IsInvertible rng_localization x),
+        P x -> P (inverse_elem x))
+      (Hmul : forall x y, P x -> P y -> P (x * y)) 
+      .
+
+    Definition rng_localization_ind
+      : forall x, P x.
+    Proof.
+      srapply Localization_type_ind.
+      - intros f.
+        refine (transport P (fraction_decompose f)^ _).
+        apply Hmul.
+        + apply Hin.
+        + apply Hinv, Hin.
+      - intros f1 f2 p.
+        apply path_ishprop.
+    Defined.
+
+  End Ind.
+  
+  Definition rng_localization_ind_homotopy {T : CRing}
+    {f g : rng_localization $-> T}
+    (p : f $o loc_in $== g $o loc_in)
+    : f $== g.
+  Proof.
+    srapply rng_localization_ind.
+    - exact p.
+    - hnf; intros x H q.
+      change (inverse_elem (f x) = inverse_elem (g x)).
+      apply equiv_path_inverse_elem.
+      exact q.
+    - hnf; intros x y q r.
+      lhs nrapply rng_homo_mult.
+      rhs nrapply rng_homo_mult.
+      f_ap.
+  Defined.
 
 End Localization.
 
--- a/theories/Algebra/Rings/Localization.v
+++ b/theories/Algebra/Rings/Localization.v
@@ -286,6 +286,17 @@ Section Localization.
   Defined.
 
   (** *** Ring laws *)
+  
+  (** Commutativity of addition *)
+  Instance commutative_plus_localization_type
+    : Commutative plus_localization_type.
+  Proof.
+    intros x; srapply Localization_type_ind_hprop; intros y; revert x.
+    srapply Localization_type_ind_hprop; intros x.
+    apply loc_frac_eq, fraction_eq_simple; simpl.
+    rewrite (rng_mult_comm (denominator y) (denominator x)).
+    f_ap; apply rng_plus_comm.
+  Defined.
 
   (** Left additive identity *)
   Instance leftidentity_plus_localization_type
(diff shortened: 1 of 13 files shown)


theories/Algebra/Rings/Ring.v
```diff
@@ -616,6 +633,15 @@ Proof.
     by exists (inv : rng_op R).
 Defined.
 
+(** The invertible elements in [R] and [rng_op R] agree, by swapping the proofs of left and right invertibility. *)
+Definition isinvertible_rng_op (R : Ring) (x : R) `{!IsInvertible R x}
+  : IsInvertible (rng_op R) x.
```
jdchristensen, line 638: I added this because I thought it might be useful later.  It turns out that that didn't work out, but I pushed it anyways, since it is good to have.

(I still need to review the main Localization.v file.  I'll try to get to it soon.)

Alizter:
I've added an induciton princple for `rng_localization` and a lemma for proving uniquness of morphsims out of the localization type. I believe this confirms we have the correct universal property.

jdchristensen:
By slightly generalizing `isinvertible_unique` (which incidentally makes it easier to state) and making the proof of `isinvertible_inverse_elem` compute better, we get that `inverse_elem_inverse_elem` is definitionally true and `equiv_path_inverse_elem` has a three line proof.  I'll also handle the TODO I added, but wanted it as a separate commit, to make it clear what changed.

jdchristensen:
I still have more to review, and it might be a couple more days before I get to it.

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,686 @@
...
+    numerator : R ;
+    denominator : R ;
+    in_mult_subset_denominator : S denominator ;
+  }.
+
+  Definition fraction_eq : Relation Fraction.
+  Proof.
+    intros [n1 d1 ?] [n2 d2 ?].
+    refine (exists (x : R), S x /\ _).
+    exact (x * (n1 * d2 - n2 * d1) = 0).
```
jdchristensen, line 83: Why is this expressed using subtraction instead of having the parts on opposite sides of the equality?

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,686 @@
...
+  (** We can factor any fraction as the multiplication of the numerator and the inverse of the denominator. *)
+  Definition fraction_decompose (f : Fraction)
+    : loc_frac f
+      = loc_in (numerator f) * inverse_elem (loc_in (denominator f)).
+  Proof.
+    apply loc_frac_eq, fraction_eq_simple.
+    nrapply rng_mult_assoc.
+  Defined.
+
+  Section Ind.
```
jdchristensen, line 642: Maybe no need for a Section containing just one result?

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,680 @@
...
+      exact q.
+    - hnf; intros x y q r.
+      lhs nrapply rng_homo_mult.
+      rhs nrapply rng_homo_mult.
+      f_ap.
+  Defined.
+
+End Localization.
+
+(** TODO: Show construction is a localization. *)
```
Alizter, line 680: The idea here is to show that a certain precomposition map is an equivalence as we do for other kinds of localizations. The LHS should be the hom from the localized ring to another ring T and the RHS should be S-inverting maps. The categories are a little awkward to describe however so I haven't been able to state this. We also would make this an equivalence of groupoids since otherwise funext would be introduced.

theories/Algebra/Rings/Localization.v
Alizter: I don't recall the reason, but it is likely because it was how it appeared in whatever text I was following at the time. I've made it symmetrical and it appears to simplify some later arguments, but it will be some time before I finish.

jdchristensen: Glad to see that this simplified things a fair bit.  I wondered if `x * (n1 * d2) = x * (n2 * d1)` might make things even a bit simpler, but a quick try shows that it would be a fair bit of work with little gain.

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,636 @@
...
+    - rewrite rng_mult_zero_l.
+      rewrite rng_plus_zero_l.
+      apply rng_mult_one_r.
+    - symmetry.
+      apply rng_mult_one_l.
+  Defined.
+
+  Instance rightidentity_plus_localization_type
+    : RightIdentity plus_localization_type zero_localization_type.
+  Proof.
```
jdchristensen, line 308: If you move commutativity of plus and mult earlier, then various "right" results will follow from the "left" results.

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,636 @@
...
+    intros x y z.
+    rewrite 3 (commutative_mult_localization_type _ z).
+    snrapply leftdistribute_localization_type.
+  Defined.
+
+  Instance isring_localization_type : IsRing Localization_type
+    := ltac:(repeat split; exact _).
+
+  Definition abgroup_localization : AbGroup.
```
jdchristensen, line 438: We don't have a coercion from rings to abelian groups?

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,634 @@
...
+  Definition Localization_type : Type := Quotient fraction_eq.
+
+  Definition loc_frac : Fraction -> Localization_type
+    := class_of fraction_eq.
+
+  Definition loc_frac_eq {f1 f2 : Fraction} (p : fraction_eq f1 f2)
+    : loc_frac f1 = loc_frac f2
+    := qglue p.
+
+  Definition Localization_type_ind (Q : Localization_type -> Type)
```
jdchristensen, line 113: STYLE.md says to not capitalize type names in results about the type (e.g. `path_contr`) but only in the type name itself, but I know our existing library has a mix.  Also, I'm not sure about have `_type` everywhere.  Would it work to just drop that?

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,634 @@
...
+
+  Definition frac_add_wd_r (f1 f2 f2' : Fraction) (p : fraction_eq f2 f2')
+    : fraction_eq (frac_add f1 f2) (frac_add f1 f2').
+  Proof.
+    pose (fraction_eq_refl f1).
+    by apply frac_add_wd.
+  Defined.
+
+  (** The addition operation on the localization is induced from the addition operation for fractions. *)
+  Instance plus_localization_type : Plus Localization_type.
```
jdchristensen, line 194: For example, this ends up being the plus operation on `rng_localization`, so calling it `plus_localization_type` seems awkward.  OTOH, it would be rare that we have to use the name.

theories/Algebra/Rings/Localization.v
Alizter: We do. I think I just did it this way to not have to indent the definition so much. I've merged them now.

theories/Algebra/Rings/Localization.v
Alizter: I think it would be better to make a smart constructor for CRing instead.

Alizter: We already had a smart constructor for `CRing` but it asked for `IsMonoid` which was proven with `repeat split` in the only use. So I opted to explode the data there and make `Build_CRing'` ask for the non-redundant ring data instead. I also introduced `Build_AbGroup'` in a similar way, but stopped short of a smart constructor for `Group` which still has some redundant data.

This allowed us to drop all the right laws here as expected.

theories/Algebra/Rings/Localization.v
Alizter: Yes, I have been inconsistent about this in the past, but it's not clear what the best way to proceed is. The `Localization_type` was named in 2020. I'm weary of calling it just `Localization` due to that having other meanings elsewhere. What about `rng_localization_type`?

theories/Algebra/Rings/Localization.v
Alizter: Ideally we wouldn't even name these as they never have to be accessed on their own, we have ring laws for that. We could actually leave it as a nameless instance but that's guaranteed poor practice.

jdchristensen:
As an aside, it would be great to have some "reflection" tools in Coq-HoTT that let us quickly solve all of these goals involving group and ring laws.  I heard that the Cubical Agda library has recently added such things, and of course there's lots of work of this nature done in Coq.  I wonder if any of it can be easily imported into Coq-HoTT?

theories/Algebra/Rings/Localization.v
jdchristensen: We're allowed to capitalize `Localization_type`, we're just not supposed to capitalize the results about it.  But the type itself is too specialized to deserve capitalization?  I think the name `rng_localization_type` might be best.  There's still the question of whether all of the results about it need to have `_type` in their names...

theories/Algebra/Rings/Localization.v
jdchristensen: Maybe `plus_rng_localization` is still better?

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,585 @@
...
+        snrapply loc_frac_eq.
+        apply fraction_eq_simple.
+        by simpl; rewrite 3 rng_mult_one_r.
+      + reflexivity.
+  Defined.
+  
+  Section Rec.
+
+    Context (T : CRing) (f : R $-> T)
+      (H : forall x, IsInvertible T (f x)).
```
jdchristensen, line 429: Shouldn't this be:
```suggestion
    Context (T : CRing) (f : R $-> T)
      (H : forall x, S x -> IsInvertible T (f x)).
```
I'll push a fix.

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,589 @@
...
+    rewrite rng_mult_negate_l.
+    apply rng_plus_negate_l.
+  Defined.
+
+  Instance associative_plus_localization_type
+    : Associative plus_localization_type.
+  Proof.
+    intros x y; srapply Localization_type_ind_hprop; intros z; revert y.
+    srapply Localization_type_ind_hprop; intros y; revert x.
+    srapply Localization_type_ind_hprop; intros x.
```
jdchristensen, line 337: Should we make `Quotient_ind2_hprop` to handle this case?  (And maybe `Quotient_ind2` to complete the picture?)  (Here I'm following the existing capitalization in that file.)

jdchristensen:
Ok, I've finished my review.  The only things left to resolve are the naming and whether to add `ind2` results to handle the repeated idiom.

theories/Algebra/Rings/Localization.v
```diff
@@ -0,0 +1,589 @@
...
+  Definition Localization_type : Type := Quotient fraction_eq.
+
+  (** Given any fraction, we have an obvious inclusion into the localization type. *)
+  Definition loc_frac : Fraction -> Localization_type
+    := class_of fraction_eq.
+
+  (** In order to give an equality of included fractions, it suffices to show that the fractions are equal under fraction equality. *)
+  Definition loc_frac_eq {f1 f2 : Fraction} (p : fraction_eq f1 f2)
+    : loc_frac f1 = loc_frac f2
+    := qglue p.
```
jdchristensen, line 116: There are a number of aliases in this file that I'm not sure are worth it.  `Localization_type` is defined as a Quotient, so we have various results available to us.  Examples are:
`loc_frac_eq`, `Localization_type_ind/ind_hprop/rec` and `isinvertible_rng_localization_preserved`.

theories/Algebra/Rings/Localization.v
Alizter: That seems like a reasonable thing to do.

theories/Algebra/Rings/Localization.v
Alizter: Yes, I think it would be best to remove those.

theories/Algebra/Rings/Localization.v
Alizter: I'll get started on that since I have a use for that elsewhere.

Alizter: I don't know what `Quotinet_ind2` should look like yet, but I will do `Quotinet_ind2_hprop` since it is simpler and is what we need.

theories/Algebra/Rings/Localization.v
Alizter: I've removed all of these now since we no longer work with `Localization_type`.

theories/Algebra/Rings/Localization.v
Alizter: I added `Quotient_ind2` and used it.

theories/Colimits/Quotient.v
```diff
@@ -91,6 +91,37 @@ Arguments Quotient_rec_beta_qglue : simpl never.
 
 Notation "A / R" := (Quotient (A:=A) R).
 
+(* Quotient induction into a hprop. *)
+Definition Quotient_ind_hprop {A : Type@{i}} (R : Relation@{i j} A)
+  (P : A / R -> Type) `{forall x, IsHProp (P x)}
+  (dclass : forall x, P (class_of R x)) : forall q, P q.
```
jdchristensen, line 97:
```suggestion
(** Quotient induction into a hprop. *)
Definition Quotient_ind_hprop {A : Type@{i}} (R : Relation@{i j} A)
  (P : A / R -> Type) `{forall x, IsHProp (P x)}
  (dclass : forall x, P (class_of R x))
   : forall x, P x.
```
Does the rest of the file have a convention about variable names?  E.g. using a, b, c for elements of A and x, y, z for elements of the quotient?  (If not, some convention like that would be good, especially for these statements where the types are not made explicit.)

theories/Algebra/Rings/QuotientRing.v
```diff
@@ -138,6 +138,25 @@ Definition QuotientRing_ind_hprop {R : Ring} {I : Ideal R} (P : R / I -> Type)
...
+  snrapply Build_RingHomomorphism'.
+  - snrapply (grp_quotient_rec _ _ f).
+    exact H.
+  - split.
+    + srapply QuotientRing_ind2_hprop.
+      nrapply rng_homo_mult.
+    + nrapply rng_homo_one.
+Defined.
+
 (** ** Quotient thoery *)
```
jdchristensen, line 160:
```suggestion
(** ** Quotient theory *)
```

jdchristensen (approved):
Nice, I like how that simplified things!

theories/Colimits/Quotient.v
Alizter: I don't think I followed any convention. I'll do a pass making it consistent throughout with the convention you suggested.

Alizter:
I pushed some further cleanups for uses of `Quotient_ind2/3` and `Quotient_rec2`.

Alizter:
> As an aside, it would be great to have some "reflection" tools in Coq-HoTT that let us quickly solve all of these goals involving group and ring laws. I heard that the Cubical Agda library has recently added such things, and of course there's lots of work of this nature done in Coq. I wonder if any of it can be easily imported into Coq-HoTT?

We have some ring reflection tactics already in the library. Thought it is from mathclasses so I don't know how easy it would be to get working with the main algebra library. Coq also has its own `ring` tactics that can solve polynomial expressions in semirings. I don't know if it relies on the stdlib in anyway, but it would be interesting to see if we can adapt its usage here.

PR #2092, opened 2024-09-16
More STYLE.md updates

jdchristensen:
Typos I noticed today, plus a couple of obsolete things.

I pretty sure that the whole last part of the bug minimizer discussion is not needed anymore, since it was about using hoqc, which we no longer use.  @JasonGross , let me know if I'm right about that.  (The third commit below.)

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1315,7 +1315,7 @@ for some other reason.
 
 We use the [Travis Continuous Integration Platform][travis] to check
 that pull requests do not break anything, and also to automatically
-update various things (such as the documentation, proviola, and
+update various things (such as the documentation and
 dependency graph linked on the [project wiki][wiki]).  Normally you
 shouldn't need to know anything about this; Travis automatically
 checks every pull request made to the central repository.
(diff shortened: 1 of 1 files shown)


Alizter:
Changes LGTM. Would you also be able to hard-wrap some of the paragraphs? I think we wrap at 80 (emacs default).

jdchristensen:
Lines wrapped.  Should we wait to hear what Jason says about the bug minimizer, or just go ahead and merge?

STYLE.md
```diff
@@ -1428,13 +1444,12 @@ You will need to pass the bug-finder several arguments to tell it to
 use the HoTT version of Coq and where to find the rest of the library;
```
JasonGross, line 1444:
```suggestion
pass the right flags and where to find the rest of the library;
```

STYLE.md
```diff
@@ -1428,13 +1444,12 @@ You will need to pass the bug-finder several arguments to tell it to
...
 a common invocation would be something like
 
-    $ /path/to/find-bug.py --coqc ../hoqc --coqtop ../hoqtop -R . HoTT Path/To/Buggy.v bug_minimized.v
+    $ /path/to/find-bug.py --coqc .../coqc --coqtop .../coqtop -R . HoTT Path/To/Buggy.v bug_minimized.v
```
JasonGross, line 1447:
```suggestion
    $ /path/to/find-bug.py --arg -noinit --arg -indices-matter -R . HoTT Path/To/Buggy.v bug_minimized.v
```

Alizter: Alternatively you can use `-f`(?) to give `_CoqProject` instead.

jdchristensen:
I'll go ahead and merge unless I hear any other suggestions.

PR #2095, opened 2024-09-21
silence "From Coq" warning

Alizter:
We can't fix this warning until we bump the minimal version to 8.21 so for now the best we can do is silence it.


Alizter:
I can't be more specific with the warning category disabling since these categories don't exist in <= 8.20. `deprecated` was the finest category available. Alternatively we could have done `-w` for disabling all warnings, but I'm uncertain if that's a good idea.

SkySkimmer:
>I can't be more specific with the warning category disabling since these categories don't exist in <= 8.20.

There should be no warning for disabling an unknown warning so why not be specific?

Alizter:
I think I got confused with `#[warnings="warning"]` which warns but as you said `#[warnings="-warning"]`, does not.

jdchristensen (approved):

PR #2096, opened 2024-09-21
Misc cleanups to Overture, Colimits/Quotient

jdchristensen:
Added a remark to STYLE.md.

Moved path_via tactic from Overture to Basics/Tactics, and made other minor fixes.

Removed reflexivity hypothesis from Quotient_rec2, and instead assume exactly what we need.

I also moved Quotient_rec2 earlier in the file, so it's best to view these one commit at a time to see the actual change there.


Alizter:
I pushed a commit introducing `Quotient_ind2` which is easier now that the hypothesis is split. The proof is the same as the old `Quotient_rec2` proof which I reproved using `Quotient_ind2`.

Alizter (approved):

PR #2097, opened 2024-09-21
bump delete-artifact

Alizter:
This was giving a warning in the CI since v1 is not compatible with upload-artifact@v4. We fix this by bumping the version to v5.


jdchristensen:
If the CI runs fine, then this LGTM.

Alizter:
Before:
image


After:
image

PR #2099, opened 2024-09-22
Basics and Types cleanup

jdchristensen:
This PR cleans up a few files in Basics/ and Types/ that I happened to be looking through.  Many of the changes are simply indentation or other whitespace changes.  Some others move things to a more logical spot.  And a couple actually change things.  I made sure to separate out the non-trivial changes into their own commits, to make it easy to review this.

(There's of course a lot more that could be done, but I'm going over these files with students tomorrow, so I'm just fixing things as I notice them.)

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -1,6 +1,8 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 
-(** * Basic definitions of homotopy type theory, particularly the groupoid structure of identity types *)
+(** * Basic definitions of homotopy type theory *)
+
+(** This file defines some of the most basic types and type formers, such as sums, products, Sigma types and path types.  It defines the action of functions on paths [ap], transport, equivalences, and function extensionality.  It also defines truncatedness, and a number of other fundamental definitions used throughout the library. *)
 
 (** Import the file of reserved notations so we maintain consistent level notations throughout the library. *)
 Require Export Basics.Settings Basics.Notations.
@@ -21,8 +23,8 @@ Notation "A -> B" := (forall (_ : A), B) : type_scope.
 
 (** [option A] is the extension of [A] with an extra element [None] *)
 Inductive option (A : Type) : Type :=
-  | Some : A -> option A
-  | None : option A.
+| Some : A -> option A
+| None : option A.
 
 Scheme option_rect := Induction for option Sort Type.
 
@@ -35,8 +37,8 @@ Register option as core.option.type.
 
 (** [sum A B], written [A + B], is the disjoint sum of [A] and [B] *)
 Inductive sum (A B : Type) : Type :=
-  | inl : A -> sum A B
-  | inr : B -> sum A B.
+| inl : A -> sum A B
+| inr : B -> sum A B.
 
 Scheme sum_rect := Induction for sum Sort Type.
 Scheme sum_ind := Induction for sum Sort Type.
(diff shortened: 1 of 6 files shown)


theories/Basics/Overture.v
```diff
@@ -1,7 +1,8 @@
 (* -*- mode: coq; mode: visual-line -*-  *)
 
-(** * Basic definitions of homotopy type theory, particularly the groupoid structure of identity types. *)
-(** Import the file of reserved notations so we maintain consistent level notations throughout the library *)
+(** * Basic definitions of homotopy type theory, particularly the groupoid structure of identity types *)
```
Alizter, line 3: I feel like the wording here is a little long for a section title. I think keeping the first part as the title would be better. We can have a succinct paragraph giving a description of the contents of this file following it.

theories/Basics/Overture.v
```diff
@@ -582,8 +578,8 @@ Local Open Scope trunc_scope.
 (** We define truncatedness using an inductive type [IsTrunc_internal A n].  We use a notation [IsTrunc n A] simply to swap the orders of arguments, and notations [Contr], [IsHProp] and [IsHSet] which specialize to [n] being [-2], [-1] and [0], respectively.  An alternative is to use a [Fixpoint], and that was done in the past.  The advantages of the inductive approach are:  [IsTrunc_internal] is cumulative; typeclass inherence works smoothly; the library builds faster.  Some disadvantages are that we need to manually apply the constructors when proving that something is truncated, and that the induction principle is awkward to work with. *)
 
 Inductive IsTrunc_internal (A : Type@{u}) : trunc_index -> Type@{u} :=
-| Build_Contr : forall (center : A) (contr : forall y, center = y), IsTrunc_internal A minus_two
-| istrunc_S : forall {n:trunc_index}, (forall x y:A, IsTrunc_internal (x = y) n) -> IsTrunc_internal A (trunc_S n).
+  | Build_Contr : forall (center : A) (contr : forall y, center = y), IsTrunc_internal A minus_two
```
Alizter, line 581: Not certain about the indentation here. I generally prefer flush with the `I` since the `| ` acts as a de facto indent.

theories/Basics/PathGroupoids.v
```diff
@@ -1415,13 +1418,15 @@ Proof.
...
 
-(** [concat], with arguments flipped. Useful mainly in the idiom [apply (concatR (expression))]. Given as a notation not a definition so that the resultant terms are literally instances of [concat], with no unfolding required. *)
-Notation concatR := (fun p q => concat q p).
+(** We declare some more [Hint Resolve] hints, now in the "hint database" [path_hints].  In general various hints (resolve, rewrite, unfold hints) can be grouped into "databases". This is necessary as sometimes different kinds of hints cannot be mixed, for example because they would cause a combinatorial explosion or rewriting cycles.  A specific [Hint Resolve] database [db] can be used with [auto with db].
+
+    The hints in [path_hints] are designed to push concatenation *outwards*, eliminate identities and inverses, and associate to the left as far as possible. *)
 
 #[export]
 Hint Resolve
+  inverse
```
Alizter, line 1426: You forgot to include `idpath` from the previous hints.

theories/Types/Sigma.v
```diff
@@ -1,16 +1,16 @@
 (* -*- mode: coq; mode: visual-line -*- *)
+
 (** * Theorems about Sigma-types (dependent sums) *)
```
Alizter, line 4: Shall we move the section title to after the imports and remove the emacs header?

Alizter (approved):
LGTM modulo a few comments

theories/Basics/Overture.v
jdchristensen: It looks like we're roughly split in the library, but proof general does not indent in this case, so let's go with that.

theories/Basics/PathGroupoids.v
jdchristensen: I purposely left it out, since it wasn't needed.  Is this because it is in the `core` hint database, or is it just a coincidence that nothing using `path_hints` needs it?  If you think I should still include it, I will.

Alizter: No idea. We don't use this so often any longer so it probably doesn't matter too much.

theories/Types/Sigma.v
jdchristensen: I kind of like having the section title as high as possible in the file.

About the emacs header, our STYLE.md document currently suggests adding this to all files, but I'd be inclined to remove it, since many people working on the library don't use emacs, and in emacs, you can make this setting yourself (which I do, so the headers aren't useful for me).  If we decide to remove these headers, let's do it in a separate PR across the whole library, updating STYLE.md at the same time.

Alizter: Yes, I think the emacs header is really an editor option and it was a legacy consideration to add it in many places. If you don't feel like doing that now lets create an issue about that cleanup.

Alizter:
Feel free to merge when you are ready.

PR #2103, opened 2024-09-30
ImpredicativeTruncation: redo using universe annotations

jdchristensen:
I noticed that this file generated lots of universe variables, and could be made a lot more clear now that Coq supports specifying the constraints between universe variables.  Because of this, some of the examples weren't really needed anymore, so I trimmed it to make the main point more clear.  I also updated the style of comments and other formatting issue.

### DIFF
--- a/theories/Metatheory/IntervalImpliesFunext.v
+++ b/theories/Metatheory/IntervalImpliesFunext.v
@@ -1,10 +1,12 @@
-(** * Theorems about the homotopical interval. *)
+(** * Theorems about the homotopical interval *)
 
 Require Import HoTT.Basics.
 Require Import HIT.Interval.
 Require Import Metatheory.Core Metatheory.FunextVarieties.
 
-(** ** From an interval type, we can prove function extensionality. *)
+(** ** An interval type implies function extensionality *)
+
+(** From an interval type with definitional computation rules on the end points, we can prove function extensionality. *)
 
 Definition funext_type_from_interval : Funext_type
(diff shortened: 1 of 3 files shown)


jdchristensen:
I have a bit more to add, which I'll do tomorrow.

jdchristensen:
I reorganized and expanded ImpredicativeTruncations a bit, including more about how things work without using PropResizing, and also a way to make it work in the lowest universe.  The reason I was interested in what happens without PropResizing is that in that case, you get definitional computation rules, which means that you can define an interval type using Funext but no HITs.  In TruncImpliesFunext.v I added an illustration that this works.  So we get that Funext is equivalent to the existence of an interval type (with definitional computation rules on the endpoints).

theories/Metatheory/IntervalImpliesFunext.v
```diff
@@ -1,14 +1,18 @@
-(** * Theorems about the homotopical interval. *)
+(** * Theorems about the homotopical interval *)
 
 Require Import HoTT.Basics.
 Require Import HIT.Interval.
 Require Import Metatheory.Core Metatheory.FunextVarieties.
 
-(** ** From an interval type, we can prove function extensionality. *)
+(** ** An interval type implies function extensionality *)
```
Alizter, line 7: Perhaps this should be the title of the section instead and we should remove the title above.

jdchristensen: Yes, done.  (This was leftover from when I thought I would put the converse into this file as well.)

Alizter (approved):
Nice

PR #2104, opened 2024-09-30
github754: make this test work again

jdchristensen:
Change the Require line from DProp to Universes.DProp to make this work again.  I'm not sure why the CI hasn't complained about this, but my build was failing.


Alizter:
Not sure what is going on, but it should be a reminder that ambiguously qualified imports are asking for trouble. (Unfortunately).

PR #2106, opened 2024-10-02
Naturality of coequalizer universal property w.r.t. functor_coeq

gio256:
Addresses step (4) from #1259.

Still TODO:
- Clean up / generalize / rename `Coeq_ind_functor_coeq_beta_cglue`.
- Comments and documentation. It would be nice to also add some documentation explaining what `CoeqUnivProp.v` is doing more generally.

### DIFF
--- a/theories/Colimits/Coeq.v
+++ b/theories/Colimits/Coeq.v
@@ -64,6 +64,58 @@ Proof.
   rapply path_ishprop.
 Defined.
 
+Definition Coeq_eta_homot {B A f g} {P : @Coeq B A f g -> Type}
+  (h : forall w : Coeq f g, P w)
+  : h == Coeq_ind P (h o coeq) (fun b => apD h (cglue b)).
+Proof.
+  unfold pointwise_paths.
+  nrapply (Coeq_ind _ (fun _ => 1)).
+  intros b.
+  lhs nrapply transport_paths_FlFr_D.
+  lhs nrapply (whiskerL _ (Coeq_ind_beta_cglue _ _ _ _)).
+  lhs nrapply (whiskerR (concat_p1 _)).
+  nrapply concat_Vp.
+Defined.
+
+Definition Coeq_rec_eta_homot {B A f g} {P : Type} (h : @Coeq B A f g -> P)
+  : h == Coeq_rec P (h o coeq) (fun b => ap h (cglue b)).
+Proof.
+  unfold pointwise_paths.
+  nrapply (Coeq_ind _ (fun _ => 1)).
+  intros b.
+  lhs nrapply transport_paths_FlFr.
+  lhs nrapply (whiskerL _ (Coeq_rec_beta_cglue _ _ _ _)).
+  lhs nrapply (whiskerR (concat_p1 _)).
+  nrapply concat_Vp.
+Defined.
+
+Definition Coeq_eta `{Funext}
+  {B A f g} {P : @Coeq B A f g -> Type} (h : forall w : Coeq f g, P w)
+  : h = Coeq_ind P (h o coeq) (fun b => apD h (cglue b))
+  := path_forall _ _ (Coeq_eta_homot h).
+
+Definition Coeq_rec_eta `{Funext}
+  {B A f g} {P : Type} (h : @Coeq B A f g -> P)
+  : h = Coeq_rec P (h o coeq) (fun b => ap h (cglue b))
+  := path_forall _ _ (Coeq_rec_eta_homot h).
+
+Definition Coeq_ind_homotopy {B A f g} (P : @Coeq B A f g -> Type)
+  {m n : forall a, P (coeq a)} (u : m == n)
+  {r : forall b, (cglue b) # (m (f b)) = m (g b)}
+  {s : forall b, (cglue b) # (n (f b)) = n (g b)}
+  (p : forall b, ap (transport P (cglue b)) (u (f b)) @ (s b) = r b @ u (g b))
+  : Coeq_ind P m r == Coeq_ind P n s.
+Proof.
+  unfold pointwise_paths.
+  nrapply Coeq_ind; intros b.
+  lhs nrapply (transport_paths_FlFr_D (f:=Coeq_ind P m r) (g:=Coeq_ind P n s)).
+  lhs nrapply (whiskerL _ (Coeq_ind_beta_cglue P n s b)).
+  lhs nrapply (whiskerR (whiskerR (ap inverse (Coeq_ind_beta_cglue P m r b)) _)).
+  lhs nrapply concat_pp_p; nrapply moveR_Mp.
+  rhs nrapply (whiskerR (inv_V _)).
+  exact (p b).
+Defined.
+
 (** ** Universal property *)
 (** See Colimits/CoeqUnivProp.v for a similar universal property without [Funext]. *)
 
--- a/theories/Colimits/CoeqUnivProp.v
+++ b/theories/Colimits/CoeqUnivProp.v
@@ -4,11 +4,13 @@ Require Import Basics.PathGroupoids.
 Require Import Types.Paths.
 Require Import Colimits.Coeq.
 Require Import Cubical.DPath.
+Require Import Cubical.DPathSquare.
 Require Import WildCat.Core.
 Require Import WildCat.Displayed.
 Require Import WildCat.Equiv.
 Require Import WildCat.EquivGpd.
 Require Import WildCat.Forall.
+Require Import WildCat.NatTrans.
 Require Import WildCat.Paths.
 Require Import WildCat.ZeroGroupoid.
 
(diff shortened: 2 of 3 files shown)


theories/Colimits/Coeq.v
```diff
@@ -64,6 +64,58 @@ Proof.
   rapply path_ishprop.
 Defined.
 
+Definition Coeq_eta_homot {B A f g} {P : @Coeq B A f g -> Type}
+  (h : forall w : Coeq f g, P w)
+  : h == Coeq_ind P (h o coeq) (fun b => apD h (cglue b)).
```
jdchristensen, line 69: About the name `Coeq_eta_homot`:  first, maybe `Coeq_ind_eta_homot` would be better?  Second, there are only a couple of places in the library that use "homot".  In general, we use "homotopic" or "homotopy".  So I suggest `Coeq_ind_eta_homotopic`.

(I think this naming is paralleling what was done for suspensions, but I think `Susp_eta_homot` should be `Susp_ind_eta_homotopic`.)

If this is changed, then the other similar names should be changed as well.

theories/Colimits/CoeqUnivProp.v
```diff
@@ -107,3 +109,95 @@ Section UnivProp.
   Defined.
 
 End UnivProp.
+
+Section UnivPropNat.
```
jdchristensen, line 113: I'm having a bit of trouble figuring out what is going on in this new Section.  Can you add lots of comments?

theories/Colimits/CoeqUnivProp.v
```diff
@@ -107,3 +109,95 @@ Section UnivProp.
...
+    nrapply moveL_Mp.
+    lhs nrapply concat_p_pp.
+    lhs nrefine (transport_paths_FlFr_D (cglue b) (Coeq_ind_homotopy P u v _))^.
+    apply (ds_dp (Coeq_ind P m r o _) (Coeq_ind P n s o _) _ _ _).
+    exact (dp_apD_nat (Coeq_ind_homotopy P u v o _) (cglue b)).
+  Defined.
+
+  Definition functor_Coeq_ind_type
+    : (forall z : Coeq f' g', P z) -> forall z : Coeq f g, P (functor_coeq h k p q z)
+    := fun x => x o functor_coeq h k p q.
```
jdchristensen, line 177: Why does this have `_type` in the name?

gio256: This comes from `Homotopy/Suspension.v`, where the type `forall z:Susp X, P z` is given the name `Susp_ind_type`. Then `Susp_ind` is a map from `Susp_ind_data` to `Susp_ind_type`.

Because the groupoid structure is less explicit in `CoeqUnivProp.v`, the naming here has no obvious parallel. Maybe `Coeq_ind_dom` and `Coeq_ind_cod` in place of `Coeq_ind_data` and `Coeq_ind_type` would be more clear?

theories/Colimits/Coeq.v
gio256: I will replace "homot" here and elsewhere. Also, the eta rule material here could be removed entirely. I thought I would need it, then decided it might be nice to have even though it turned out not to be necessary for proving `Coeq_ind_inv_nat`.

theories/Colimits/CoeqUnivProp.v
jdchristensen: @gio256 I don't understand how your comment relates to the code above.  What's written above isn't a type at all, but is a function from one type to another.  For example, the next result proves `Is0Functor functor_Coeq_ind_type`.  So I don't see how it would make sense to call this `Coeq_ind_cod` either.

gio256: My point was really that `functor_Coeq_ind_type` is named after `functor_Susp_ind_type`, but the connection with `_type` here is sort of lost because we don't explicitly name an equivalent to `Susp_ind_type` for the coequalizer.

I agree the naming in the whole file is a bit unclear. I will revisit naming along with trying to add some helpful comments.

jdchristensen: Oh, now I see.  It's really `functor_(Coeq_ind_type)` saying that the unnamed `Coeq_ind_type` is a functor.  Maybe just naming `Coeq_ind_type` would make this more clear?

jdchristensen:
One high level comment: would it work to prove similar things about any GraphQuotient, and then deduce the results about Coeq from them?  GraphQuotient is underlies many other things defined in the library, so it would be ideal if as much as possible was proved about GraphQuotient, so that similar things would follow for everything built using it.  OTOH, maybe some things are simpler when stated for coequalizers and would be hard to generalize.

theories/Colimits/CoeqUnivProp.v
gio256: Oh, I see how that's very confusing, and I doubt my explanation was all that helpful. I will try to clarify things and hopefully still keep some parallels with `Homotopy/Suspension.v`.

gio256:
@jdchristensen that's an interesting idea, and my initial thought is that it could work quite well. I am still trying to wrap my head around the full funext-free proof of `O_inverts_functor_coeq`, so I'm not certain how much of the work here would generalize smoothly to GraphQuotient. But at the very least, it seems worthwhile to explore some zero-groupoid tools for GraphQuotient and see what they might be useful for.

theories/Colimits/CoeqUnivProp.v
```diff
@@ -107,3 +112,113 @@ Section UnivProp.
...
+  (** We now have two different ways of mappping from [Coeq_ind_type f' g' P] to [sig (Coeq_ind_data f g (P o functor_coeq h k p q))]. Here we construct a transformation between these two maps.
+
+  Coeq_ind_type f' g' P ---------functor_Coeq_ind_type---> Coeq_ind_type f g (P o functor_coeq h k p q)
+        |                                                                   |
+        |                                                                   |
+  Coeq_ind_inv f' g' P                                      Coeq_ind_inv f g (P o functor_coeq h k p q)
+        |                                                                   |
+        |                                                                   |
+        V                                                                   V
+  sig (Coeq_ind_data f' g' P) ---functor_Coeq_ind_data---> sig (Coeq_ind_data f g (P o functor_coeq h k p q))
```
gio256, line 206: I'm not sure how nicely this will play with the generated documentation (or if the generated docs are a priority). Happy to remove, but I thought it would be useful for review at least.

Alizter: Looks like this:
image
Can be fixed by adding
```
<<
  keeps spaces
     in here
>>
```
image

gio256: Thanks! Are the docs hosted somewhere that I can check the fixed version?

Alizter: Unfortunately we don't host them for PRs, but they are hosted once it hits `master`. The wiki page has all the links: https://github.com/HoTT/Coq-HoTT/wiki

You can view the changes yourself locally by doing something like `dune build @theories/doc` and then inspecting:
```
_build/default/theories/HoTT.html/HoTT.Colimits.CoeqUnivProp.html
```
with a web browser. VSCode is able to run a local server for me when I right click a HTML document so I can see it with the CSS loaded in a browser on localhost.

Alizter: Your latest changes look fine to me:
image

gio256: Thanks for checking. On building I got the following warning:
```
File "Colimits/CoeqUnivProp.v", line 206, character 109, warning: missing newline after ">>" block
```
The latest commit should fix this though, and I double-checked that the diagram is still rendered correctly.

Alizter: Yes, the warnings are pretty easy to fix. I think they just try to enforce best practices. There are other tools like alectryon so its generally good to keep a consistent style that most tools expect from doc comments.

jdchristensen:
@gio256 Did you think any more about my "high level comment" from Oct 2?  It was only an idea, and if you think it's not feasible, that's fine.  I'll wait to hear your plan before re-reviewing this.

gio256:
@jdchristensen Assuming that it would be useful to have more of these 0-groupoid universal properties in the library, I think it is worth exploring whether they could be derived from the analogous universal property for GraphQuotient. Unfortunately, I haven't had a chance to investigate this yet.

I am happy to push this PR forward in the meantime or to close it until we are able to formulate a more general approach.

jdchristensen (approved):
I think it's probably best to merge this as it is, and then keep in mind for the future the idea of doing something similar for graph quotients and then using that work to deduce what is done here (if possible; not sure).  Maybe open an issue, or edit an existing issue, to mention this?

@Alizter , it wouldn't hurt if you took a quick look at this before merging.

Incidentally, I think it's also possible to state something a bit more sophisticated than what is the main naturality result here, `Coeq_ind_inv_nat`.  That result is stated by manually invoking the functoriality of `Coeq_ind_type` and `Coeq_ind_data`.  One could actually make these into functors, with domain the wild-category whose objects consist of `A, B, f, g` and whose morphisms consist of `h, k, p, q`, and with codomain the wild category of 0-groupoids.  Then we could say that they are naturally isomorphic, which might in the end boil down to exactly `Coeq_ind_inv_nat`.  But that would add some complexity, and hopefully won't be needed.

Alizter (approved):
LGTM

PR #2111, opened 2024-10-08
chore: remove Cubical.v indexing file

Alizter:
We remove the Cubical.v indexing file since we always wish to have finer dependencies when using the Cubical library. This allows us to refine some imports elsewhere.


jdchristensen:
I'm neutral about this.  It's sometimes convenient to not have to be that specific.  But I'm also in favour of having fewer things in the theories/ folder, and since Cubical.v would only be used in three places, maybe it's best to remove it.  So I guess this LGTM.

PR #2112, opened 2024-10-09
istrunc_self_implies_istrunc

ThomatoTomato:
Added this cute little lemma about truncations that might come in handy. All it says is that a map (A -> istrunc n.+1 A) is sufficient to deduce that A is n+1 truncated.

I wasn't so sure what to name it, so feel free to make it fit into your mold.

### DIFF
--- a/theories/Basics/Trunc.v
+++ b/theories/Basics/Trunc.v
@@ -535,4 +535,13 @@ Proof.
   exact (equiv_iff_hprop_uncurried (iff_contr_hprop A)).
 Defined.
 
+(** If a type [A] implies that it is [n.+1]-truncated, then it is [n.+1]-truncated. **)
+Definition istrunc_self_implies_istrunc {n : trunc_index} {A : Type} (H : A -> IsTrunc n.+1 A)
+  : IsTrunc n.+1 A.
+Proof.
+  apply istrunc_S.
+  intros a b.
+  exact (H a a b).
+Defined.
+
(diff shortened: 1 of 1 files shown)


theories/Basics/Trunc.v
```diff
@@ -535,4 +535,13 @@ Proof.
   exact (equiv_iff_hprop_uncurried (iff_contr_hprop A)).
 Defined.
 
+(** If a type [A] implies that it is [n.+1]-truncated, then it is [n.+1]-truncated. **)
+Definition istrunc_self_implies_istrunc {n : trunc_index} {A : Type} (H : A -> IsTrunc n.+1 A)
```
Alizter, line 539: I would call this `istrunc_inhabited_istrunc` drawing inspiration from `HProp.v`.

Alizter (approved):
LGTM. Did you find a nice use of this lemma yet? I couldn't think of any truncation proofs that are simplified by knowing you have an inhabitant.

Alizter:
I found one place it could potentially be useful in `equiv_istrunc_istrunc_loop` from `Loops.v`. It doesn't make the proof shorter, but I believe its the same argument as in this proof.

ThomatoTomato:
My motivation to prove this was originally as a stepping stone to `equiv_istrunc_istrunc_loop`, but it already exists in the library. Sadly, I currently don't have any other plans for it :(

theories/Basics/Trunc.v
Alizter: @ThomatoTomato looks like you missed this comment. I'll rename it in another PR.

PR #2113, opened 2024-10-10
ind_homotopy lemma for FreeAbGroup

Alizter:


theories/Algebra/AbGroups/Abelianization.v
```diff
@@ -149,7 +149,7 @@ Section Abel.
     srapply Coeq_ind_hprop.
     exact a.
   Defined.
-
+  
   (** And its recursion version. *)
   Definition Abel_rec_hprop (P : Type) `{IsHProp P}
```
jdchristensen, line 154: I know it's not part of this PR, but this result seems useless.  If P is a proposition and you have a map `G -> P`, then P is contractible, since we have the identity element of G.

jdchristensen (approved):

theories/Algebra/AbGroups/Abelianization.v
Alizter: Good point. It wasn't used anyway so I've removed it.

PR #2115, opened 2024-10-10
lax monoidal functors

Alizter:
We introduce the notion of a `lax` monoidal functor. This notion of "morphism of monoidal categories" has the nice property of preserving monoid objects in monoidal categories.

The "lax" part is due to requiring a "reverse morphism" `F a ⊗ F b $-> F (a ⊗ b)`. This is probably contrary to the perhaps more natural reverse direction that you might come up with. The important thing here is that monoids are preserved. In most cases, we actually have what is known as a "strong monoidal functor" where this morphism is an isomorphism. The unit preservation also is "lax" and therefore looks backwards.

This sets us up for some future work on showing that `FreeAbGroup`, and perhaps a future `FreeLeftModule`, are monoidal functors from `Type` to `AbGroup` or `LeftModule R`. This lets us define monoid rings over `cring_Z` and generally monoid rings. Hopefully, this will form the basis of further work on group rings, polynomial rings, etc. 

TODO:
- [ ] Better names for fields of IsMonoidalFunctor?

### DIFF
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -1060,7 +1060,7 @@ Class IsMonoidalFunctor {A B : Type}
         (uncurry (fun a b => tensorB (F a) (F b)))
         (uncurry (fun a b => F (tensorA a b)));
 
-  (** A morhpism [fmap_unit] relating the two unit objects. *)
+  (** A morphism [fmap_unit] relating the two unit objects. *)
   fmap_unit : IB $-> F (IA);
 
   (** Such that the following coherence conditions hold: 
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -1045,3 +1045,86 @@ Section TwistConstruction.
   Abort.
 
 End TwistConstruction.
+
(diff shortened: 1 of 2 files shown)


theories/WildCat/Monoidal.v
```diff
@@ -1045,3 +1045,86 @@ Section TwistConstruction.
...
+  `{IsMonoidal A tensorA IA, IsMonoidal B tensorB IB}
+  (F : A -> B) `{!Is0Functor F, !Is1Functor F} := {
+
+  (** A natural transformation [fmap_tensor] from [F a ⊗ F b] to [F (a ⊗ b)]. *)
+  fmap_tensor
+    : NatTrans
+        (uncurry (fun a b => tensorB (F a) (F b)))
+        (uncurry (fun a b => F (tensorA a b)));
+
+  (** A morhpism [fmap_unit] relating the two unit objects. *)
```
jdchristensen, line 1063:
```suggestion
  (** A morphism [fmap_unit] relating the two unit objects. *)
```

jdchristensen (approved):

PR #2118, opened 2024-10-10
subgroup and ideal preimage, ideal extension

Alizter:
In this lemma we show that the preimage of a subgroup with respect to a group homomoprhism is itself a subgroup. We then show the same holds for rings and ideals (a.k.a contraction). Finally, we define what is known as the extension of an ideal and show a basic property that the extension of the contraction of an ideal is included in the original ideal.

These are useful notions to have around for describing the relationship between ideals in a localization and in the base ring.


jdchristensen (approved):
Looks great!  Sorry for the delay.

PR #2121, opened 2024-10-10
associativity of ideal product

Alizter:
This was left as a TODO so I did a quick proof for it. It's possible it can be simplified.

### DIFF
--- a/theories/Algebra/Rings/Ideal.v
+++ b/theories/Algebra/Rings/Ideal.v
@@ -904,12 +904,37 @@ Section IdealLemmas.
     + by apply ideal_in_plus_negate.
   Defined.
 
-  (** TODO: *)
   (** The product of ideals is an associative operation. *)
-  (* Lemma ideal_product_assoc (I J K : Ideal R) : I ⋅ (J ⋅ K) ↔ (I ⋅ J) ⋅ K.
+  Lemma ideal_product_assoc (I J K : Ideal R) : I ⋅ (J ⋅ K) ↔ (I ⋅ J) ⋅ K.
   Proof.
-    intros r; split; apply Trunc_functor.
-  Abort. *)
+    snrapply ideal_subset_antisymm.
+    - intros x.
+      apply Trunc_rec. 
+      intros p; induction p as [g [y z p q] | | g h p1 IHp1 p2 IHp2].
+      + strip_truncations.
+        induction q as [t [z w r s] | | t k p1 IHp1 p2 IHp2].
+        * rewrite rng_mult_assoc.
+          by apply tr, sgt_in, ipn_in; [ apply tr, sgt_in, ipn_in | ].
+        * rewrite rng_mult_zero_r.
+          apply ideal_in_zero.
+        * rewrite rng_dist_l_negate.
+          by apply ideal_in_plus_negate.
+      + apply ideal_in_zero.
+      + by apply ideal_in_plus_negate.
+    - intros x.
+      apply Trunc_rec.
+      intros p; induction p as [g [y z p q] | | g h p1 IHp1 p2 IHp2].
+      + strip_truncations.
+        induction p as [t [y w r s] | | t k p1 IHp1 p2 IHp2].
+        * rewrite <- rng_mult_assoc.
+          by apply tr, sgt_in, ipn_in; [ | apply tr, sgt_in, ipn_in ].
+        * rewrite rng_mult_zero_l.
+          apply ideal_in_zero.
+        * rewrite rng_dist_r_negate.
+          by apply ideal_in_plus_negate.
+      + apply ideal_in_zero.
+      + by apply ideal_in_plus_negate.
+  Defined.
 
   (** Products of ideals are subsets of their intersection. *)
   Lemma ideal_product_subset_intersection (I J : Ideal R) : I ⋅ J ⊆ I ∩ J.


theories/Algebra/Rings/Ideal.v
```diff
@@ -904,12 +904,37 @@ Section IdealLemmas.
...
 
-  (** TODO: *)
   (** The product of ideals is an associative operation. *)
-  (* Lemma ideal_product_assoc (I J K : Ideal R) : I ⋅ (J ⋅ K) ↔ (I ⋅ J) ⋅ K.
+  Lemma ideal_product_assoc (I J K : Ideal R) : I ⋅ (J ⋅ K) ↔ (I ⋅ J) ⋅ K.
   Proof.
-    intros r; split; apply Trunc_functor.
-  Abort. *)
+    snrapply ideal_subset_antisymm.
+    - intros x.
```
jdchristensen, line 911: For ideals I and J, is `I ⋅ J` definitionally the same as `J ⋅ I` as ideals over the opposite ring?  Coq doesn't seem to think so, but if this were true, then both halves of the proof could be replaced with one result, used for R and then for `rng_op R`.  If it's not definitionally true, then we should have a lemma `ideal_product_op` that relates them, and try to use that here and in the other `_l` and `_r` results that duplicate work.

Alizter: I tried this last week but this was a little trickier than I thought it would be due to the way we define product ideals. I'll have another stab at it again soon.

Alizter:
@jdchristensen I managed to carry out the opposite argument you suggested, but it required a lot of refactoring in Ideal. Namely, exploding the Section IdealLemmas which is probably for the best. While the diff is big, it might be a good time to perform some more style changes like changing Lemma to Definition.

theories/Algebra/Rings/Ideal.v
```diff
@@ -139,6 +145,80 @@ Section IdealElements.
...
+(** Ideal equality is transitive. *)
+Global Instance transitive_ideal_eq {R : Ring} : Transitive (@ideal_eq R).
+Proof.
+  intros I J K p q x; specialize (p x); specialize (q x); by transitivity (J x).
+Defined.
+
+(** *** Subset relation on ideals *)
+
+(** We define the subset relation on ideals in the usual way: *)
+Definition ideal_subset {R : Ring} (I J : Subgroup R) := (forall x, I x -> J x).
```
jdchristensen, line 186: Since this is about Subgroups, not Ideals, it should go in Groups/Subgroup.v.  Same with reflexivity, transitivity and antisymmetry.  But really, this is just about any predicates on any type, so maybe it just belongs in Types/Forall or Basics/Overture or something like that?

theories/Algebra/Rings/Ideal.v
```diff
@@ -139,6 +145,80 @@ Section IdealElements.
...
+Coercion ideal_eq_subset {R : Ring} {I J : Subgroup R} : ideal_eq I J -> ideal_subset I J.
+Proof.
+  intros f x; apply f.
+Defined.
+
+(** *** Left and right ideals are invariant under ideal equality *)
+
+(** Left ideals are invariant under ideal equality. *)
+Definition isleftideal_eq {R : Ring} (I J : Subgroup R) (p : ideal_eq I J)
+  : IsLeftIdeal I -> IsLeftIdeal J.
```
jdchristensen, line 209: This can be strengthened:  If `I` is an ideal, and `J` is any predicate on `R` which is pointwise logically equivalent to `I`, then `J` is a subgroup and an ideal.

jdchristensen (commented):
I started reviewing this, but the change to the indentation made it impossible to see what changed later on in the file, as github got quite confused with the diff it displayed.  It's better if large indentation changes can be moved to a separate commit.  Maybe you can force push to fix this, replacing just the last commit?  (I think the comments I already made should still be visible and understandable after this.)

theories/Algebra/Rings/Ideal.v
```diff
@@ -76,6 +76,12 @@ Definition ideal_op (R : Ring) : Ideal R -> Ideal (rng_op R)
   := fun I => Build_Ideal (rng_op R) I _.
 Coercion ideal_op : Ideal >-> Ideal.
 
+Definition rightideal_op (R : Ring) : LeftIdeal R -> RightIdeal (rng_op R)
+  := fun I => Build_RightIdeal (rng_op R) I _.
+
+Definition leftideal_op (R : Ring) : RightIdeal R -> LeftIdeal (rng_op R)
+  := fun I => Build_LeftIdeal (rng_op R) I _.
```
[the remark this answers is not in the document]
Alizter: Yes. The same can't be done for `ideal_op` which is why I probably did it this way for these two.

theories/Algebra/Rings/Ideal.v
Alizter: Isn't this just the composition of two results then? Any `J` logically equivalent to a subgroup [I] is also a subgroup, furthermore also an ideal when [I] is one.

jdchristensen: Yes.  But we may as well state the stronger version if we generalize these predicate comparisons to any predicates on any types.

theories/Algebra/Rings/Ideal.v
```diff
@@ -921,19 +967,16 @@ Proof.
...
-    + by apply ideal_in_plus_negate.
+  - apply ideal_subset_antisymm; only 1: apply f.
+    intros x i.
+    apply ideal_product_op.
+    nrapply (ideal_product_subset_pres_l (R:=rng_op R)).
+    1: rapply ideal_product_op.
+    apply f.
+    apply (ideal_product_op (R:=rng_op R)).
+    nrapply (ideal_product_subset_pres_l (R:=R)).
+    1: rapply (ideal_product_op (R:=rng_op R)).
```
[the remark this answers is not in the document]
jdchristensen: I think I'll just push this along with my other idea to see what you think.

jdchristensen:
I pushed three commits.  The first two I think are clear benefits.  And `subgroup_generated_subset_subgroup` can be used in other places in Ideal.v and probably in other places in Subgroup.v or maybe elsewhere.  Look for places using `ideal_in_zero` and `ideal_in_plus_negate`.

The third commit is the thing I wrote a comment about.  Feel free to revert if you think it is not worth it.  Or maybe with the new approach to the assertion, it's not worth using opposite rings here?

I do think opposite rings help with ideals in other parts of the file, so that's great.  Can you find more places where this helps?

I have to run and won't have time to look at this for a while, but hopefully you'll be able to use some of these ideas to clean things up further.

Alizter:
@jdchristensen I've added some material on predicates of types. This generalises all the subset material we had in `Ideal` and makes it available for `Subgroup`.

theories/Algebra/Rings/Ideal.v
Alizter: I suppose we could ask for `J` to be hprop valued separately. That should solve this issue.

theories/Algebra/Rings/Ideal.v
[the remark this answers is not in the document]
Alizter: I like it.

jdchristensen (commented):
I've carefully looked at Classes.v and Predicate.v, but not at the rest of the PR.  I'm about to push a lot of changes involving those two files, so I thought I should pause my review here and see if you like the changes.  I broke the changes into small commits, and the library builds after each commit, so you can read them independently.

theories/Basics/Classes.v
```diff
@@ -2,6 +2,63 @@ Require Import Basics.Overture Basics.Tactics.
...
+  transitivity (P (f (f^-1 a))).
+  2: apply r.
+  rapply related_reflexive_path.
+  symmetry; apply (ap P), eisretr.
+Defined.
+
+Definition pointwise_moveR_equiv {A A' B : Type} (R : Relation B) `{Reflexive _ R} `{Transitive _ R}
+  (f : A <~> A') (P : A -> B) (Q : A' -> B)
+  : relation_pointwise (fun _ => R) P (Q o f) -> relation_pointwise (fun _ => R) (P o f^-1) Q
+  := pointwise_moveL_equiv (flip R) f Q P.
```
jdchristensen: Note how using `flip R` saved work here;  but I had to change the statement to make this work.  That's fine, as later users just have to provide `f^-1` instead of `f`.

theories/Basics/Overture.v
```diff
@@ -278,6 +311,13 @@ Proof.
   exact u.
 Defined.
 
+Definition related_reflexive_path {A : Type} (R : Relation A) `{Reflexive A R}
+  {a b : A} (p : a = b)
+  : R a b.
```
jdchristensen, line 316: I was surprised that we didn't have something like this already.  Did I miss something?  Also, not sure of the name.

theories/Basics/Classes.v
```diff
@@ -2,6 +2,63 @@ Require Import Basics.Overture Basics.Tactics.
 
 (** * Classes *)
 
+(** ** Pointwise Lemmas *)
+
+Section Pointwise.
```
jdchristensen, line 7: I like this material.  In a separate PR, we should consider porting the pointwise_paths stuff to use this.  That would mean rearranging things somehow.  Either this gets moved into Overture.v, or everything involving homotopies (and therefore everything involving equivalences) would have to move to a new file, e.g. Basics/Functions.v.  But even that won't quite work, since currently Classes uses equivalences...

theories/Basics/Classes.v
jdchristensen: (In the last commit, I swapped these two results, which saved one use of `symmetry`.)

Alizter:
Feel free to push ahead with this, I won't be able to take a look until the weekend.

jdchristensen:
I just pushed a few more changes.  I still want to think a bit more about this, hopefully tomorrow.

theories/Algebra/Rings/Ideal.v
jdchristensen: I wasn't sure whether to bother changing this.  Maybe it's fine like it is.

theories/Algebra/Rings/Ideal.v
```diff
@@ -781,8 +769,8 @@ Defined.
 (** We declare and import a module for various (Unicode) ideal notations. These exist in their own special case, and can be imported and used in other files when needing to reason about ideals. *)
 
 Module Import Notation.
-  Infix "⊆" := ideal_subset       : ideal_scope.
-  Infix "↔" := ideal_eq           : ideal_scope.
+  Infix "⊆" := pred_subset        : ideal_scope.
+  Infix "↔" := pred_eq            : ideal_scope.
   Infix "+" := ideal_sum          : ideal_scope.
   Infix "⋅" := ideal_product      : ideal_scope.
   Infix "∩" := ideal_intersection : ideal_scope.
```
jdchristensen, line 776: I wonder if ⊆, ↔ and ∩ should be defined for any predicates?  They would certainly make my additions to Predicate.v more readable, but OTOH they can be hard to type if you don't know how.  So maybe it's best if they are limited to more specialized files?

theories/Algebra/Rings/Ideal.v
```diff
@@ -791,431 +779,377 @@ Module Import Notation.
...
+      rewrite rng_mult_negate_r.
+      by apply ideal_in_plus_negate.
+  - induction jk as [? [] | | ? ? ? ? ? ? ].
+    + by apply q.
+    + by apply v.
+    + apply v, ideal_in_zero.
+    + change (I ((g - h) * r)). 
+      rewrite rng_dist_r.
+      rewrite rng_mult_negate_l.
+      by apply ideal_in_plus_negate.
```
jdchristensen, line 1112: I've given some examples of how most uses of ideal_in_zero, ideal_in_plus_negate, etc can be avoided by using results that work with subgroups rather than elements.  Here is another spot where it might be possible.  It might require a `ideal_product_rec` lemma, or something like that.

jdchristensen (commented):
I've finished my review and don't plan to push any more changes.  There are still a number of comments, but none of them are very important, so some could be handled by just adding TODO items to the file.

A couple more I noticed: `ideal_subset_extension_preimage` could be simplified with a `_rec` lemma for `ideal_generated`.  And `subgroup_left_ideal_quotient` is really the statement that an arbitrary intersection of subgroups is a subgroup, which could be proved in Subgroup.v and then used here.  (Each `I (_ * x)` is a subgroup since `(_ * x)` is a group homomorphism, and we intersect these over all `x` satisfying `J x`.)

theories/Basics/Overture.v
jdchristensen: I noticed that `Hom_path` is a special case of this.  Maybe we have other special cases as well?

Alizter: I think the name is fine. Not sure about other special cases, but its a simple application of J.

theories/Algebra/Rings/Ideal.v
Alizter: I'm inclined to say we define it for any predicate. It aids readability immensely and its not worth sacrificing to save somebody from copy-and-pasting. The latter issue is really a documentation issue and maybe we can add a section in the STYLE.md guide about typing unicode. In my editor (vscode + coq-lsp) I can just type `\` and it starts unicode input. Same in CoqIDE (though very rarely used) and I believe PG has good ways of inputting unicode.

theories/Basics/Predicate.v
```diff
@@ -0,0 +1,79 @@
...
+Instance transitive_pred_subset {A : Type} : Transitive (@pred_subset A)
+ := transitive_pointwise (B:=fun x => Type) (fun _ A B => A -> B).
+
+Coercion pred_eq_subset {A : Type} (P Q : A -> Type)
+  : pred_eq P Q -> pred_subset P Q
+  := fun p x => fst (p x).
+
+Definition pred_eq_subset' {A : Type} (P Q : A -> Type)
+  : pred_eq P Q -> pred_subset Q P
+  := fun p x => snd (p x).
```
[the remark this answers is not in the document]
Alizter: Yes, I will rename them.

theories/Basics/Classes.v
Alizter: Yes, lets leave that to a future PR. I think its time we rethink some of the order of the earlier material.

theories/Algebra/Rings/Ideal.v
Alizter: I've pushed a commit with predicate notations. I've only done \subseteq and \leftrightarrow for now, but \cup and \cap would be straightforward additions. We would still have to define the ideal notations for \cup and \cap however since they require an ideal to be returned rather than a predicate.

jdchristensen: Those notations do make things very readable.  I'm fine with adding \cup and \cap too, but don't mind if they are left out.

Alizter: I've gone ahead and used these notations in `Subgroup.v` too.

theories/Algebra/Rings/Ideal.v
Alizter: I've simplified the argument here. Let me know if you find any others that need attention.

Alizter:
@jdchristensen This is ready for another round of review.

jdchristensen:
`subgroup_leftideal_quotient`:  As I mentioned earlier, this just seems to be the fact that an arbitrary intersection of subgroups is a subgroup.  My last commit makes this more clear, since `J` can be any predicate.

`isleftideal_subgroup_leftideal_quotient`:  And this seems to be the fact that an arbitrary intersection of left ideals is a left ideal.

Both of these facts should be proved in general, and then used in these two places.  (Or maybe both of these results can then be dropped, as you can just directly form the left ideal as the intersection.)

Alizter:
@jdchristensen Good observation. I will try to do that today.

Alizter:
This is not so straightforward due to the truncation. I can show that `merely (forall i, H i)` is a subgroup, but in this particular case we have `merely (forall i, J i -> H i)` which means the first isn't a more general version of the second (without funext).

Alizter:
@jdchristensen I didn't manage to find a nice way to generalise it for the reasons I mentioned before. Did you want to have a go too or should we just merge this now?

jdchristensen:
If we have `H : I -> Subgroup G`, then the intersection will be defined as `fun g => merely (forall i, H i g)` (so that we don't need to use Funext to know that it is hprop-valued).  So there are two ways to proceed.  One is to change the definition of the ideal intersection to
```coq
exact (fun r => merely (forall x : sig J, subgroup_preimage (grp_homo_rng_left_mult r) I x.1)).
```
Here `sig J` plays the role of the indexing type `I`, and this is exactly expressed as an intersection of subgroups, so it should work fine.  However, with this change, some later proofs need to change, which might be a bit annoying.

An alternative is to use the same predicate we currently have, but then use `issubgroup_equiv` to convert the predicate to the other form.  This is how it would start:
```coq
Definition subgroup_leftideal_quotient {R : Ring} (I : Subgroup R) (J : R -> Type)
  : Subgroup R.
Proof.
  snapply Build_Subgroup.
  1: exact (fun r => merely (forall x, J x -> subgroup_preimage (grp_homo_rng_left_mult r) I x)).
  snapply issubgroup_equiv.
  - exact (fun r => merely (forall (x : sig J), subgroup_preimage (grp_homo_rng_left_mult r) I x.1)).
  - intro r.
    apply ReflectiveSubuniverse.equiv_O_functor. (* Need to add an import. *)
    symmetry.
    napply equiv_sig_ind.
  - (* Remaining goal: use that intersections of subgroups are subgroups. *)
Admitted.
```
(This just shows that it is possible.  But probably we want to have a general proof that an intersection of ideals is an ideal, and use that, so this result wouldn't appear in this form.)

jdchristensen:
Another option, which ties in to things I've been suggesting about subgroups, is to allow subgroups to be defined by non-hprop-predicates.  We'd probably need two kinds of subgroups.  Maybe the kind I'm talking about could be called "presubgroups".  Then we would have a result that says that if `H : G -> Type` is a presubgroup, then `fun g => (W -> H g)` is a presubgroup.  And another result that says that if `H : I -> Presubgroup G` is a family of presubgroups, then `fun g => forall i, H i g` is a presubgroup.  And finally if `H` is a presubgroup, then `fun g => merely (H g)` is a subgroup.  Those could be instances, and from that Coq could automatically infer that `fun r => merely (forall x, J x -> subgroup_preimage (grp_homo_rng_left_mult r) I x)` is a subgroup.

Alizter:
I like the idea of showing it is equivalent to an uncurried version. I won't be able to finish it tonight, but that's probably the direction we should go in. I had some stuff on showing intersections are subgroups and ideals that I wrote today, so its just going to require some more time.

Alizter:
@jdchristensen I sank too much time into trying to fix the Rocq master build last week. I am not sure when I will have enough time to look at this again. In the meantime, if you feel like pushing ahead with this, I don't mind.

jdchristensen:
I merged with master.  No conflicts and no changes required.  I'll see if I can simplify those last two results.

JasonGross:
> I started reviewing this, but the change to the indentation made it impossible to see what changed later on in the file, as github got quite confused with the diff it displayed. It's better if large indentation changes can be moved to a separate commit.

Very belatedly, https://github.blog/news-insights/product-news/ignore-white-space-in-code-review/ might help with such issues

PR #2122, opened 2024-10-10
cleanup: make into aliases: easy -> done, now -> by

Alizter:
These tactics have been around and pretty much do the same thing. We therefore define them in terms of the more commonly used variant. We could consider removing them completely, but the diff will be a bit larger.

### DIFF
--- a/contrib/SetoidRewrite.v
+++ b/contrib/SetoidRewrite.v
@@ -55,7 +55,7 @@ Defined.
          CRelationClasses.arrow) (GpdHom a).
 Proof.
   intros x y eq_xy eq_ax.
-  now transitivity x.
+  by transitivity x.
 Defined.
 
 (* forall a : A, x $== y -> a $== y -> a $== x *)
--- a/theories/Algebra/Universal/Algebra.v
+++ b/theories/Algebra/Universal/Algebra.v
@@ -128,7 +128,7 @@ Proof.
   destruct p, q.
   unfold path_algebra, path_sigma_hprop, path_sigma_uncurried.
   cbn -[center].
-  now destruct (center (ha = hb)).
+  by destruct (center (ha = hb)).
 Defined.
 
 Arguments path_ap_carriers_path_algebra {_} {_} (A B)%_Algebra_scope (p q)%_path_scope.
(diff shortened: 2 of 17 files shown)


jdchristensen (commented):
It looks like `easy` was fairly different from `done` before.  Unfortunately, it's hard to parse the old definition of `easy`, so it's not clear how it is different.  I guess these come from ssreflect, so maybe it's better to not reuse the name `easy` for something different?  How about changing `now` to use `done`, deleting `easy` completely, and using `done` for the one or two places that currently use `easy`?

I also wonder whether `easy` might be worth keeping for some reason.  Maybe it can do some things that `done` can't?  Or maybe it's faster?  Not sure.

Alizter:
> It looks like `easy` was fairly different from `done` before. Unfortunately, it's hard to parse the old definition of `easy`, so it's not clear how it is different.

They main difference is that `done` does not try `inversion` or `contradiction`. This means it is generally faster. In these cases where inversion is needed, its not typically considered a "finishing step" anyway.

> I guess these come from ssreflect, so maybe it's better to not reuse the name `easy` for something different?

`done` and `by` historically come from ssreflect although they now have implementations in the Coq stdlib.

>How about changing `now` to use `done`, deleting `easy` completely, and using `done` for the one or two places that currently use `easy`?

Sure. Would you prefer `now` to be `done` or `by`? I feel like the latter is closer to how it is used.

> I also wonder whether `easy` might be worth keeping for some reason. Maybe it can do some things that `done` can't? Or maybe it's faster? Not sure.

As I answered above, it can do some inversion in rare cases, but is slower overall.

jdchristensen:
Ok, it sounds like getting rid of `easy` is a good idea, since it's not something we've used.  And maybe it's best to get rid of `now` and just keep `by`, even if it makes the diff bigger?  I'm also ok with keeping `now` as an alias of `by`, but in general I think each alias makes it slightly harder for beginners to understand the library.  (Incidentally, I think some of our tactics files contain various unused things, but I don't know if there's much benefit to going through and deleting them.)

Alizter:
I've removed `now` and `easy` now.

jdchristensen (approved):

Alizter:
FTR, I have some patches lying around cleaning up the unused tactics you discussed. I think its worth refining them at some point so that potential contributors don't get overwhelmed by useless information.

PR #2127, opened 2024-10-30
Adapt to Coq#19611

Tragicus:
https://github.com/coq/coq/pull/19611 changes the behavior of the unification algorithm.

### DIFF
--- a/theories/Algebra/Rings/Matrix.v
+++ b/theories/Algebra/Rings/Matrix.v
@@ -711,7 +711,6 @@ Proof.
   by rewrite rng_plus_zero_l.
 Defined.
 
-
 (** The sum of two lower triangular matrices is lower triangular. *)
 Global Instance lower_triangular_plus {R : Ring@{i}} {n : nat}
   (M N : Matrix R n n) `{!IsLowerTriangular M} `{!IsLowerTriangular N}
@@ -787,7 +786,7 @@ Defined.
 Global Instance upper_triangular_zero {R : Ring@{i}} {n : nat}
   : IsUpperTriangular (matrix_zero R n n).
 Proof.
-  apply tr@{i}.
+  apply tr.
   by hnf; intros; rewrite entry_Build_Matrix.
 Defined.
 
@@ -825,7 +824,7 @@ Global Instance upper_triangular_diag {R : Ring@{i}} {n : nat} (v : Vector R n)
   : IsUpperTriangular (matrix_diag v).
 Proof.
   unfold IsUpperTriangular.
-  apply tr@{i}.
+  apply tr.
   intros i j Hi Hj lt_i_j.
   rewrite entry_Build_Matrix.
   rewrite kronecker_delta_lt.
@@ -846,12 +845,12 @@ Defined.
 Definition upper_triangular_matrix_ring@{i} (R : Ring@{i}) (n : nat)
   : Subring@{i i} (matrix_ring@{i} R n).
 Proof.
-  nrapply (Build_Subring'@{i i} (fun M : matrix_ring R n => IsUpperTriangular M)).
+  nrapply (Build_Subring' (fun M : matrix_ring R n => IsUpperTriangular M)).
   - exact _.
   (* These can all be found by typeclass search, but being explicit makes this faster. *)
-  - intros x y ? ?; exact (upper_triangular_plus@{i} x (-y)).
-  - exact upper_triangular_mult@{i}.
-  - exact upper_triangular_identity@{i}.
+  - intros x y ? ?; exact (upper_triangular_plus x (-y)).
+  - exact upper_triangular_mult.
+  - exact upper_triangular_identity.
 Defined.
 
 (** Lower triangular matrices are a subring of the ring of matrices. *)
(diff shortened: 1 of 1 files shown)


jdchristensen:
These changes aren't pretty.  Can you explain why the changes in coq/coq#19611 make them necessary?  It's also surprising to me that so much needs to change in this one file, but nothing else in the entire library needs to change.

Tragicus:
I am afraid I will have a hard time answering, since I have no idea how coq deals with universes and my changes only touch the unification algorithm. I am also surprised, but since there already are annotations I was not too worried. I am available if anyone can help me understand what happens.

theories/Algebra/Rings/Matrix.v
```diff
@@ -775,7 +787,7 @@ Defined.
 Global Instance upper_triangular_zero {R : Ring@{i}} {n : nat}
   : IsUpperTriangular (matrix_zero R n n).
 Proof.
-  apply tr.
+  apply tr@{i}.
```
jdchristensen, line 790: @Tragicus In my tests, none of the universe annotations added from this line onwards are needed.  So the only changes are the inlining of `strip_truncations` in three places.

theories/Algebra/Rings/Matrix.v
```diff
@@ -707,6 +711,7 @@ Proof.
   by rewrite rng_plus_zero_l.
 Defined.
 
+
```
jdchristensen, line 714:
```suggestion
```

theories/Algebra/Rings/Matrix.v
jdchristensen: I can push my edits to this PR if you agree.

theories/Algebra/Rings/Matrix.v
```diff
@@ -697,7 +697,11 @@ Global Instance upper_triangular_plus {R : Ring@{i}} {n : nat} (M N : Matrix R n
   : IsUpperTriangular (matrix_plus M N).
 Proof.
   unfold IsUpperTriangular.
-  strip_truncations; apply tr.
+  revert_opaque H1.
+  refine (@Trunc_ind _ _ _ (fun _ => istrunc_truncation _ _) _).
+  intro H1.
+  revert_opaque H2; refine (Trunc_ind@{i i} _ _); intro H2.
+  apply tr.
```
jdchristensen, line 704: Ah, I figured out what is special about these uses of strip_truncations.  The goal is non-dependent, so Coq isn't able to guess which truncation to apply to the domain.  If we replace them with `strip_reflections` (adding `Require Import Modalities.ReflectiveSubuniverse.` to the top of the file), then those three lines are the only changes that need making.

Even better would be for `strip_truncations` to first try `Trunc_rec`, the non-dependent eliminator.  I'll see if I can do this, but it won't be right away.  @Alizter , if you have a chance, feel free to take a look.

jdchristensen: Hmm, no, that explanation is not correct.  The same universe issue happens whether you use `Trunc_ind` or `Trunc_rec`, but things work out ok with either `O_ind` or `O_rec`.  Somehow the cumulativity of our truncations is causing Coq to generate a free universe variable.  That kind of thing has happened to us here and there with other lemmas involving cumulative inductives, but I can't figure out why the update to Coq is causing it to happen just in this file when we use `strip_truncations`, but not everywhere else that uses that tactic.  E.g. Rings/Ideal.v uses it a lot, without issues.

If any Coq developers have any insight into this, it would be very helpful.

theories/Algebra/Rings/Matrix.v
Tragicus: Sure, please feel free.

theories/Algebra/Rings/Matrix.v
jdchristensen: Is there a way in Ltac to find out the universe arguments in a term `t`?  E.g. something like `unify t identifier@{?u}` that would instantiate a "universe evar `u`" that we could use later?

JasonGross: For things like this you should use [Ltac2's `Constr.Unsafe.kind`](https://github.com/coq/coq/blob/master/user-contrib/Ltac2/Constr.v)

SkySkimmer: I don't think using Unsafe would make the proof particularly cleaner.

jdchristensen: I would like the `strip_truncations` tactic to operate on a goal of the form `Trunc@{u} X -> Y` using `Trunc_ind@{u _}`.  The way it works now, it seems to use `Trunc_ind@{v _}` (probably with a `<=` constraint on `v`), which is valid since `Trunc` is cumulative, but which generates an unwanted universe variable.  (In some cases, Coq later identifies this universe variable with `u`, but in Matrix.v, it seems to be leaving it free, which is causing problems.)  So if some trick can be done in one place to make `strip_truncations` work this work, it would make proofs cleaner.  Unfortunately, I don't understand @JasonGross 's comment well enough to try this.

Alternatively, can a universe variable be passed into a tactic?  E.g. could we have `strip_truncations@{i j}` or `strip_truncations i j` which tells it which universe variables to use?

jdchristensen:
@Tragicus I'll close this, since I think it is handled by #2130, which has now been merged.  So you can probably adjust the Coq PR to not need an overlay for Coq-HoTT.

PR #2129, opened 2024-11-04
Prove Yoneda-like result for paths, and required lemmas

jdchristensen:
The third commit proves that `paths : X -> (X -> Type)` is an embedding.  For this we needed that `functor_forall_id` sends families of O-modal maps to an O-modal map, which is in commit 1.  And that `equiv_fun` is an embedding, which is in commit 2.

### DIFF
--- a/theories/HFiber.v
+++ b/theories/HFiber.v
@@ -262,6 +262,13 @@ Proof.
   exact (equiv_iff_hprop (@isequiv_ap_isembedding _ _ f) (@isembedding_isequiv_ap _ _ f)).
 Defined.
 
+(** It follows from [isembedding_isequiv_ap] and [isequiv_ap_equiv_fun] that [equiv_fun] is an embedding. *)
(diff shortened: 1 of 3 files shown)


theories/HFiber.v
```diff
@@ -200,6 +200,17 @@ Section UnstableOctahedral.
 
 End UnstableOctahedral.
 
+(** We characterize the fibers of [functor_forall], but only in the special case where the base map is [idmap]. This doesn't dependn on anything else in this file, but can't be put in Types/Forall.v, because it requires results from Types/Sigma.v. *)
```
Alizter, line 203:
```suggestion
(** We characterize the fibers of [functor_forall], but only in the special case where the base map is [idmap]. This doesn't depend on anything else in this file, but can't be put in Types/Forall.v, because it requires results from Types/Sigma.v. *)
```

Alizter (approved):
LGTM modulo a typo

PR #2130, opened 2024-11-07
Matrix: use strip_reflections to avoid extra universe variables

jdchristensen:
This makes the library build when https://github.com/coq/coq/pull/19611 is applied to Coq, and can replace #2127.

(It also builds with 8.19 and other versions.  And I added universe annotations to make sure the problem doesn't crop up again.)


jdchristensen:
I have another clean-up PR coming that will include some comments in Truncations/Core.v about this issue.

Alizter (approved):

PR #2133, opened 2024-11-07
cleanup: unused discriminates tactic

Alizter:
In this PR we cleanup some unused tactics. Each tactic is removed in its own commit so that it can easily be reverted in the future. There is still some ways to go, but this is a start.

### DIFF
--- a/theories/Basics/Tactics.v
+++ b/theories/Basics/Tactics.v
@@ -58,14 +58,6 @@ Ltac destruct_one_pair :=
 
 Ltac destruct_pairs := repeat (destruct_one_pair).
 
-(** Destruct one existential package, keeping the name of the hypothesis for the first component. *)
-
-Ltac destruct_one_ex :=
-  let tacT H := let ph := fresh "X" in (destruct H as [H ph]) in
-    match goal with
-      | [H : (sig ?P) |- _ ] => tacT H
-    end.
-
 (** Revert the last hypothesis. *)
 
 Ltac revert_last :=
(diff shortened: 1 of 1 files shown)


theories/Basics/Tactics.v
```diff
@@ -224,29 +163,14 @@ Ltac destruct_all_rec_calls :=
...
-Ltac autoinjection tac := idtac.
-  (* match goal with *)
-  (*   | [ H : ?f ?a = ?f' ?a' |- _ ] => tac H *)
-  (* end. *)
-
-Ltac inject H := progress (inversion H ; subst*; clear_dups) ; clear H.
+Ltac inject H := progress (inversion H ; clear_dups) ; clear H.
 
-Ltac autoinjections := repeat (clear_dups ; autoinjection ltac:(inject)).
+Ltac autoinjections := repeat (clear_dups ; ltac:(inject)).
```
jdchristensen, line 168: Is the `ltac:(  )` part needed?

jdchristensen (approved):
Seems fine to me.  We could also consider moving any that could possibly be useful in the future to contrib.

theories/Basics/Tactics.v
Alizter: I have no idea so I left it as-is.

jdchristensen: It was probably only there because it was being based as an argument.  The tactic doesn't give an error with the `ltac:( )` part removed.

PR #2134, opened 2024-11-07
cycle construction for symmetric monoidal categories

Alizter:
In this PR, we introduce what I've termed the "cycle construction" for symmetric monoidal categories. Like the "twist construction" it is a slicker way to build symmetric monoidal categories without having to give all the data from the beginning. But this time, instead of asking for a twist map ABC -> BAC we ask for a cycle map ABC -> CAB. We then have variants of the hexagon and pentagon axioms which appear to be slicker than in the twist construction.

### TODO
- [ ] more comments
- [ ] use somewhere?

### DIFF
--- a/theories/WildCat/MonoidalCycleConstruction.v
+++ b/theories/WildCat/MonoidalCycleConstruction.v
@@ -1,0 +1,207 @@
+Require Import Basics.Overture Basics.Tactics Types.Forall WildCat.Monoidal.
+Require Import WildCat.Core WildCat.Bifunctor WildCat.Prod WildCat.Equiv.
+Require Import WildCat.NatTrans WildCat.Square WildCat.Opposite.
+
+(** * Cycle Construction for Symmetric Monoidal Categories *)
+
+(** The following construction is what we call the "cycle construction". Like the "twist construction" found in MonoidalTwistConstruction.v, this procedure allows one to build a symmetric monoidal category from simpler pieces.
+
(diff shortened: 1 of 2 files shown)


theories/WildCat/MonoidalCycleConstruction.v
```diff
@@ -0,0 +1,207 @@
...
+  
+  Instance ismonoidal_cycle
+    : IsMonoidal A cat_tensor cat_tensor_unit
+    := {}.
+  
+  Instance issymmetricmonoidal_cycle
+    : IsSymmetricMonoidal A cat_tensor cat_tensor_unit
+    := {}.
+
+End CycleConstruction.
```
Alizter, line 207:
```suggestion
End CycleConstruction.

```

jdchristensen (approved):
This is great!  Do you think it can be used to prove that Join satisfies the pentagon law?  That would be helpful for something I'm working on.

Alizter:
@jdchristensen I don't recall off the top of my head, but you are more than welcome to try!

jdchristensen:
@Alizter Do you have any partial work on the pentagon law for Join, either using the current twist approach to associativity or the cycle approach added in this PR?  I vaguely recall that you worked on it.

Alizter:
> @Alizter Do you have any partial work on the pentagon law for Join, either using the current twist approach to associativity or the cycle approach added in this PR? I vaguely recall that you worked on it.

I can't find anything about joins, but I have some partial work on instantiating these constructions with the smash product.

PR #2135, opened 2024-11-07
prove twist hexagon in twist construction

Alizter:
We finally prove the remaining hexagon consisting of only twists that was left as a WIP. It uses two instances of the 9-gon meaning it isn't equivalent to it in any way.


Alizter:
ping @jdchristensen

jdchristensen:
Oh, I see I already gave this a LGTM, and that my previous comments were all about possible future improvements, so feel free to merge.

PR #2138, opened 2024-11-14
cleanup in FinNat

Alizter:
We had a lot of universes being generated here, I've cut them down a bit by using `Universe Minimization ToSet`.

### DIFF
--- a/theories/Spaces/Finite/FinNat.v
+++ b/theories/Spaces/Finite/FinNat.v
@@ -16,9 +16,9 @@ Definition zero_finnat@{} (n : nat) : FinNat n.+1
 Definition succ_finnat@{} {n : nat} (u : FinNat n) : FinNat n.+1
   := (u.1.+1; leq_succ u.2).
 
-Definition path_succ_finnat {n : nat} (x : nat) (h : x.+1 < n.+1)	
-  : succ_finnat (x; leq_pred' h) = (x.+1; h).	
-Proof.	
+Definition path_succ_finnat {n : nat} (x : nat) (h : x.+1 < n.+1)
+  : succ_finnat (x; leq_pred' h) = (x.+1; h).
+Proof.
   by apply path_sigma_hprop.
 Defined.
 


jdchristensen (commented):
Are the `Monomorphic` modifiers still needed?

(Also, I don't get why there are various `path_*` lemmas like `path_zero_finnat` which are just aliases for `path_ishprop`.  Can't `path_ishprop` just be used?)

jdchristensen:
I got rid of the unneed `path_*_finnat` definitions, and adapted `path_succ_finnat` to fit better what is needed.  I also simplified a couple of proofs.  I suspect `finnat_ind` could be made to compute better.

I also got rid of one unneeded universe variable in `finnat_ind`, and annotated a few things to enforce that universe variables don't accidentally arise again.

I didn't think about why some things are marked `Monomorphic`.  Can we just get rid of that?

Alizter:
@jdchristensen I think that was misused to get things to not be polymorphic, but that's already the case when we have no Univ variables, therefore it's fine to remove.

I also had some patches cleaning up things here similarly to what you did, but I'll need to check in detail later as I may have changed some proofs differently.

jdchristensen:
Ok, I removed `Monomorphic`.  We still use it in other places, e.g. things like `Funext` are marked as monomorphic even though they don't take any universe variables.  I'm not sure what advantage adding `Monomorphic` gives if something already doesn't use any universe variables, but maybe there's some tiny speedup from this?

jdchristensen:
PS: Sorry if I interfered with other patches you had queued up!

Alizter:
> Ok, I removed `Monomorphic`. We still use it in other places, e.g. things like `Funext` are marked as monomorphic even though they don't take any universe variables. I'm not sure what advantage adding `Monomorphic` gives if something already doesn't use any universe variables, but maybe there's some tiny speedup from this?

I am not sure if there is a speed up, but its probably not worth confusing readers.

> PS: Sorry if I interfered with other patches you had queued up!

No worries, I've pushed some of the changes I was going to do while we were here.

theories/Spaces/Finite/FinNat.v
```diff
@@ -50,9 +32,10 @@ Proof.
   - elim (not_lt_zero_r u.1 u.2).
   - destruct u as [x h].
     destruct x as [| x].
-    + exact (transport (P n.+1) (path_finnat _ _ h) (z _)).
-    + refine (transport (P n.+1) (path_succ_finnat x h) _).
-      apply s. apply IHn.
+    + nrefine (transport (P n.+1) _ (z _)).
+      by apply path_sigma_hprop.
+    + refine (transport (P n.+1) _ (s _ _ (IHn (x; leq_pred' _)))).
```
jdchristensen, line 37: This is less clear than before.  `path_succ_finnat` is slightly non-trivial, and says that any element of the form `(x.+1; _)` is equal to something in the image of `succ_finnat`, which seems useful to record.  So I'd vote for reinstating `path_succ_finnat` and using it here.

jdchristensen:
This looks good to me, modulo the build error and the one comment about `succ_finnat`.

Alizter:
I've fixed the universe issue in 8.19 (for some reason `idpath` wasn't being minimized to `idpath@{Set}` and I've reverted the removal of `path_succ_finnat` like you suggested.

theories/Spaces/Finite/FinNat.v
```diff
@@ -15,6 +15,12 @@ Definition zero_finnat@{} (n : nat) : FinNat n.+1
 
 Definition succ_finnat@{} {n : nat} (u : FinNat n) : FinNat n.+1
   := (u.1.+1; leq_succ u.2).
+  
+Definition path_succ_finnat@{} {n : nat} (u : FinNat n) (h : u.1.+1 < n.+1)
+  : succ_finnat u = (u.1.+1; h).
+Proof.
+  by apply path_sigma_hprop.
+Defined.
```
jdchristensen, line 23: I think you should revert to the version I had:
```coq
Definition path_succ_finnat {n : nat} (x : nat) (h : x.+1 < n.+1)	
  : succ_finnat (x; leq_pred' h) = (x.+1; h).	
Proof.	
  by apply path_finnat.	
Defined.
```
The advantage here is that the caller doesn't need to figure out that `leq_pred'` needs to be used.

jdchristensen: (It should work with `by apply path_sigma_hprop.` as well.)

theories/Spaces/Finite/FinNat.v
```diff
@@ -15,10 +15,10 @@ Definition zero_finnat@{} (n : nat) : FinNat n.+1
...
 Definition succ_finnat@{} {n : nat} (u : FinNat n) : FinNat n.+1
   := (u.1.+1; leq_succ u.2).
-  
-Definition path_succ_finnat@{} {n : nat} (u : FinNat n) (h : u.1.+1 < n.+1)
-  : succ_finnat u = (u.1.+1; h).
-Proof.
+
+Definition path_succ_finnat {n : nat} (x : nat) (h : x.+1 < n.+1)	
+  : succ_finnat (x; leq_pred' h) = (x.+1; h).	
+Proof.
```
jdchristensen, line 21: Trailing whitespace here.  Otherwise, LGTM!

Issue #2144, opened 2024-12-03
Simpler definition of span pushout


Alizter:
When reading Blakers-Massey I noticed that we could use a more efficient implementation of `SPushout`. Currently, we define it as a special case of a `Pushout`, which is in turn a special case of `Coeq`, which in turn is a special case of `GraphQuotient`.

It's not hard to see that you can simplify this all the way through to end up with something like this:
```coq
  Definition SPushout := 
    @GraphQuotient (X + Y) (fun xy xy' => match xy, xy' with
                                        | inl x, inr y => Q x y
                                        | _, _ => Empty
                                        end).
```
Proving the rec/ind is also simple and modulo some changes to Freudenthal.v, it should work.

I created this issue so we can discuss the tradeoffs first. The main advantage would be the more compact term size and the main disadvantage would be pushout lemmas not being applicable to span pushouts. Interested to know what others think.

Alizter:
> That said, while SPushout isn't used much currently, David Warn's shorter type-theoretic paper on path spaces of pushouts uses that set-up, and some students in my class should be "pushing" work on that soon. I had forgotten to point out SPushout to them, but they might want to use it, in which case they'll probably need to know basic facts about it, which would need to be reproved if the definition changes. So maybe we should wait a bit to see how that plays out?

That's a good point, no trouble waiting. 

> BTW, if you trace through the definition of Pushout, you'll see that it is equivalent to a very similar family, with `Q x y` replaced with `hfiber f x * hfiber g y`, and again, the other cases being sent to Empty. So we could in fact define Pushout in terms of SPushout if we wanted to, and prove the fundamental facts about SPushout, inheriting them to Pushout, instead of the other way around.

This is quite interesting. I would be curious if span pushouts simplify any of the pushout proofs. For instance, functionality has a more succinct description. Compare the following:

```coq
Definition functor_spushout_homotopic {X Y Z W : Type}
  {Q : X -> Y -> Type} {Q' : Z -> W -> Type}
  {f g : X -> Z} (h : f == g)
  {i j : Y -> W} (k : i == j)
  (l : forall x y, Q x y -> Q' (f x) (i y))
  (m : forall x y, Q x y -> Q' (g x) (j y))
  (H : forall x y q,
    spglue Q' (l x y q) @ ap (spushr Q') (k y)
      = ap (spushl Q') (h x) @ spglue Q' (m x y q))
  : functor_spushout f i l == functor_spushout g j m.
```

```coq
[...]

PR #2145, opened 2024-12-03
functoriality of span pushouts

Alizter:
Functoriality of span pushouts are simpler than their pushout counterparts in that they have simpler data.

As discussed in #2144 we could also simplify the definition of SPushout, but we will hold off on that for now.

Regarding naming, the beta rule for `rec` has two problems. Firstly the path constructor is called `spglue` not `sglue` and secondly we don't capitlize. I've done it this way to stay consistent with the `_ind` analogues, but I think it might be better to change both of them.


theories/Colimits/SpanPushout.v
```diff
@@ -46,3 +54,75 @@ Section SpanPushout.
...
+  (h' : forall x y, Q' x y -> Q'' (f' x) (g' y))
+  : functor_spushout f' g' h' o functor_spushout f g h
+    == functor_spushout (f' o f) (g' o g) (fun x y => h' _ _ o h x y).
+Proof.
+  snrapply SPushout_ind.
+  1,2: reflexivity.
+  intros x y q.
+  snrapply transport_paths_FlFr'.
+  lhs nrapply concat_p1.
+  rhs nrapply concat_1p.
```
jdchristensen, line 82: The idiom I like for this common situation is:
```suggestion
  nrapply transport_paths_FlFr'; apply equiv_p1_1q.
```
(Same below.)

theories/Colimits/SpanPushout.v
```diff
@@ -46,3 +54,75 @@ Section SpanPushout.
...
+  lhs nrapply (ap_compose _ (functor_spushout f' g' h') (spglue Q q)).
+  lhs nrapply ap.
+  1: nrapply spushout_rec_beta_sglue.
+  symmetry.
+  rhs nrapply spushout_rec_beta_sglue.
```
jdchristensen, line 87: `rhs` uses an inversion in its proof, so you'll get a smaller proof term with:
```suggestion
  lhs nrapply spushout_rec_beta_sglue.
  symmetry.
```
(`lhs` and `rhs_V` don't have the inversion, while `lhs_V` and `rhs` do.)
(Note: none of these suggestions have been tested.)

theories/Colimits/SpanPushout.v
```diff
@@ -46,3 +54,75 @@ Section SpanPushout.
...
+Definition functor_spushout_idmap {X Y : Type} {Q : X -> Y -> Type}
+  : functor_spushout idmap idmap (fun x y (q : Q x y) => q) == idmap.
+Proof.
+  snrapply SPushout_ind.
+  1,2: reflexivity.
+  intros x y q.
+  snrapply transport_paths_FlFr'.
+  lhs nrapply concat_p1.
+  rhs nrapply concat_1p.
+  rhs nrapply ap_idmap.
```
jdchristensen, line 101: Should we add `transport_paths_Flr'` to avoid the need for `ap_idmap`?

theories/Colimits/SpanPushout.v
```diff
@@ -22,6 +23,13 @@ Section SpanPushout.
     intros [[x y] q]; cbn in *.
     apply sglue'; assumption.
   Defined.
+  
+  Definition spushout_rec_beta_sglue (R : Type)
+             (spushl' : X -> R) (spushr' : Y -> R)
+             (sglue' : forall x y (q : Q x y), spushl' x = spushr' y)
+             (x:X) (y:Y) (q:Q x y)
+    : ap (SPushout_rec R spushl' spushr' sglue') (spglue q) = sglue' x y q
+    := Pushout_rec_beta_pglue _ _ _ _ ((x, y); q).
```
jdchristensen, line 32: I think we should change `sglue` to `spglue` everywhere.

Not sure about capitalization.  I thought our pattern was that we sometimes use capitals when defining a type, like `Pushout`, but that results *about* that type would use lower case.  However, there are lots of examples where the capitals are kept, and lots where they are not.  And there are situations where they are kept for some of the results, but removed for others.  This makes it hard to remember the names. I think moving forward we should always get rid of them, except for in the name of the type itself.  (Also note that many of our types do not use capitalization, e.g. `paths`, `sig`, `prod`, `forall`, `sum`, etc.  So we could also consider not using capitalization at all...)

Alizter: The situations in which the capitals are kept are mostly my fault and are due to me trying to replicate the autogenerated induction principles. Sometimes Coq will have a terrible lemma finding method like in `induction` that will use that autogenerated name to find said lemma without going through the standard `Register` mechanism. This isn't a problem for HITs however since we can't use such tactics anyway.

Alizter: I've renamed `sglue` however, but opted to keep lower case.

jdchristensen: I guess you meant "keep upper case"?  In any case, not a big deal.

jdchristensen (approved):

theories/Colimits/SpanPushout.v
Alizter: I was thinking about lemmas about `_rec` and `_ind`, but I've also changed those too now.

PR #2147, opened 2024-12-05
Path spaces of graph quotients, coequalizers, and pushouts

ThomatoTomato:
Now that we have identity systems, #2143, we can use these to characterize path spaces of certain HITs. In this PR we characterize path spaces of graph quotients, coequalizers and pushouts, using the induction principle given by [Kraus and von Raumer](https://arxiv.org/pdf/1901.06022). These results will be helpful towards characterizing the colimits in Wärn's zigzag-construction.

We first introduce transportDD. This is a doubly dependent transport over doubly dependent type families, defined in PathGroupods.v. In Equivalences.v, we prove that both transportD and transportDD are equivalences. In Sigma.v, we show that transportDD is a curried version of transport over a sigma type.

Next up we prove induction and recursion principles for descent data over graphs, parallel pairs, and spans. We then reprove the flattening lemma using this framework. Finally, we deduce the characterization of path spaces.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -93,6 +93,33 @@ Section EquivTransport.
 
 End EquivTransport.
 
+(** Doubly dependent transport is an equivalence *)
+Section EquivTransportD.
+
+    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type) 
+      {x y : A} (p : x = y) {px : P x}.
+
+    Global Instance isequiv_transportD : IsEquiv (transportD P Q p px).
+    Proof.
+      snrapply Build_IsEquiv.
+      { refine (_ o transportD P Q p^ (transport P p px)).
+        exact (transport (Q x) (transport_Vp _ p px)). }
+      all: by destruct p.
+    Defined.
+
+    Definition equiv_transportD : Q x px <~> Q y (transport P p px)
+      := Build_Equiv _ _ (transportD P Q p px) _.
+
+    Context {py : P y} (q : transport P p px = py).
+
+    Global Instance isequiv_transportDD : IsEquiv (transportDD P Q p q)
+     := isequiv_compose.
+
+    Definition equiv_transportDD : Q x px <~> Q y py
+      := Build_Equiv _ _ (transportDD P Q p q) _.
+
+  End EquivTransportD.
+
 (** In all the above cases, we were able to directly construct all the structure of an equivalence.  However, as is evident, sometimes it is quite difficult to prove the adjoint law.
 
    The following adjointification theorem allows us to be lazy about this if we wish.  It says that if we have all the data of an (adjoint) equivalence except the triangle identity, then we can always obtain the triangle identity by modifying the datum [equiv_is_section] (or [equiv_is_retraction]).  The proof is the same as the standard categorical argument that any equivalence can be improved to an adjoint equivalence.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -84,7 +84,7 @@ Section EquivTransport.
 
   Context {A : Type} (P : A -> Type) {x y : A} (p : x = y).
 
-  (** The inverse and the homotopies of transport is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
+  (** The inverse and the homotopies of transport are defined explicitly.  This allows us to reason about the inverse when the input is not [idpath]. *)
   Global Instance isequiv_transport : IsEquiv (transport P p) | 0
     := Build_IsEquiv (P x) (P y) (transport P p) (transport P p^)
     (transport_pV P p) (transport_Vp P p) (transport_pVp P p).
@@ -94,13 +94,13 @@ Section EquivTransport.
 
 End EquivTransport.
 
-(** Doubly dependent transport is an equivalence *)
+(** Doubly dependent transport is an equivalence. *)
 Section EquivTransportD.
 
     Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type)
       {x y : A} (p : x = y) {px : P x}.
 
-    (** The inverse of transportD is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
+    (** The inverse of transportD is defined explicitly.  This allows us to reason about the inverse when the input is not [idpath]. *)
     Global Instance isequiv_transportD : IsEquiv (transportD P Q p px).
     Proof.
       snrapply Build_IsEquiv.
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -84,6 +84,7 @@ Section EquivTransport.
 
   Context {A : Type} (P : A -> Type) {x y : A} (p : x = y).
 
+  (** The inverse and the homotopies of transport is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
   Global Instance isequiv_transport : IsEquiv (transport P p) | 0
     := Build_IsEquiv (P x) (P y) (transport P p) (transport P p^)
     (transport_pV P p) (transport_Vp P p) (transport_pVp P p).
@@ -96,9 +97,10 @@ End EquivTransport.
 (** Doubly dependent transport is an equivalence *)
 Section EquivTransportD.
 
-    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type) 
+    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type)
       {x y : A} (p : x = y) {px : P x}.
 
+    (** The inverse of transportD is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
     Global Instance isequiv_transportD : IsEquiv (transportD P Q p px).
     Proof.
       snrapply Build_IsEquiv.
@@ -110,12 +112,8 @@ Section EquivTransportD.
     Definition equiv_transportD : Q x px <~> Q y (transport P p px)
       := Build_Equiv _ _ (transportD P Q p px) _.
 
-    Context {py : P y} (q : transport P p px = py).
-
-    Global Instance isequiv_transportDD : IsEquiv (transportDD P Q p q)
-     := isequiv_compose.
-
-    Definition equiv_transportDD : Q x px <~> Q y py
+    Definition equiv_transportDD {py : P y} (q : transport P p px = py)
+      : Q x px <~> Q y py
       := Build_Equiv _ _ (transportDD P Q p q) _.
 
   End EquivTransportD.
@@ -658,7 +656,7 @@ Goal forall (A : Type) (B : A -> Type) (C : forall a:A, B a -> Type) (D : forall
     intros b; cbn in b.
     intros x; cbn in x.
     elim x; clear x.
-    intros c; cbn in c. 
+    intros c; cbn in c.
     intros d; cbn in d.
     (** Here begins [build_record] *)
     cbn; unshelve econstructor.
@@ -674,7 +672,7 @@ Goal forall (A : Type) (B : A -> Type) (C : forall a:A, B a -> Type) (D : forall
     intros x; cbn in x.
     elim x; clear x.
     intros x; cbn in x.
-    elim x; clear x. 
+    elim x; clear x.
     intros b; cbn in b.
     intros c; cbn in c.
     intros d; cbn in d.
@@ -691,7 +689,7 @@ Goal forall (A : Type) (B : A -> Type) (C : forall a:A, B a -> Type) (D : forall
     intros x; cbn in x.
     elim x; clear x.
     intros x; cbn in x.
-    elim x; clear x. 
+    elim x; clear x.
     intros b; cbn in b.
     intros c; cbn in c.
     intros d; cbn in d.
@@ -704,7 +702,7 @@ Goal forall (A : Type) (B : A -> Type) (C : forall a:A, B a -> Type) (D : forall
     intros b; cbn in b.
     intros x; cbn in x.
     elim x; clear x.
-    intros c; cbn in c. 
+    intros c; cbn in c.
     intros d; cbn in d.
     cbn; exact idpath.
 Defined.
@@ -728,7 +726,7 @@ Goal forall (A:Type) (R:A->A->Type),
     intros a2; cbn in a2.
     intros r; cbn in r.
     cbn; unshelve econstructor.
-    { cbn; unshelve econstructor. 
+    { cbn; unshelve econstructor.
       { (** [build_record] can't guess at this point that it needs to use [a1] instead of [a2], and in fact it tries [a2] first; but later on, [exact r] fails in that case, causing backtracking to this point and a re-try with [a1].  *)
         cbn; exact a1. }
       { cbn; exact a2. } }
@@ -829,7 +827,7 @@ Section Examples.
       intros x; cbn in x.
       elim x; clear x.
       intros a; cbn in a.
-      intros x; cbn in x. 
+      intros x; cbn in x.
       elim x; clear x.
       intros b; cbn in b.
       intros p; cbn in p.
@@ -855,7 +853,7 @@ Section Examples.
     - intros x; cbn in x.
       elim x; clear x.
       intros a; cbn in a.
-      intros x; cbn in x. 
+      intros x; cbn in x.
       elim x; clear x.
       intros b; cbn in b.
       intros p; cbn in p.
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -756,17 +756,26 @@ Defined.
 
 (** Dependent transport in doubly dependent types and more. *)
 
+(** Singly dependent transport over doubly dependent types. *)
 Definition transportD {A : Type} (B : A -> Type) (C : forall a:A, B a -> Type)
   {x1 x2 : A} (p : x1 = x2) (y : B x1) (z : C x1 y)
   : C x2 (p # y)
   :=
   match p with idpath => z end.
 
+(** Singly dependent transport over doubly dependent types of 2 variables. *)
 Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a -> Type)
   {x1 x2 : A} (p : x1 = x2) (y : B x1) (z : C x1) (w : D x1 y z)
   : D x2 (p # y) (p # z)
   :=
   match p with idpath => w end.
+
+(** Doubly dependent transport over doubly dependent types.  *)
+Definition transportDD {A : Type} (B : A -> Type) (C : forall a : A, B a -> Type)
+  {a1 a2 : A} (pA : a1 = a2)
+  {b1 : B a1} {b2 : B a2} (pB : transport B pA b1 = b2)
+  (c1 : C a1 b1) : C a2 b2
+  := transport (C a2) pB (transportD B C pA b1 c1).
 
 (** *** [ap] for curried two variable functions *)
 
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -774,7 +774,8 @@ Definition transportD2 {A : Type} (B C : A -> Type) (D : forall a:A, B a -> C a 
 Definition transportDD {A : Type} (B : A -> Type) (C : forall a : A, B a -> Type)
   {a1 a2 : A} (pA : a1 = a2)
   {b1 : B a1} {b2 : B a2} (pB : transport B pA b1 = b2)
-  (c1 : C a1 b1) : C a2 b2
+  (c1 : C a1 b1)
+  : C a2 b2
   := transport (C a2) pB (transportD B C pA b1 c1).
 
 (** *** [ap] for curried two variable functions *)
@@ -874,7 +875,7 @@ Defined.
 
 (** Naturality of [transport011]. *)
 Definition ap_transport011{A B} {P Q : A -> B -> Type}
-  {a1 a2 : A} {b1 b2 : B} (p : a1 = a2) (q : b1 = b2)  
+  {a1 a2 : A} {b1 b2 : B} (p : a1 = a2) (q : b1 = b2)
   (f : forall {a b}, P a b -> Q a b) (x : P a1 b1)
   : f (transport011 P p q x) = transport011 Q p q (f x).
 Proof.
@@ -992,6 +993,32 @@ Definition transport2_const {A B : Type} {x1 x2 : A} {p q : x1 = x2}
   (r : p = q) (y : B)
   : transport_const p y = transport2 (fun _ => B) r y @ transport_const q y
   := match r with idpath => (concat_1p _)^ end.
+
+(** We need a better naming scheme for transportD_const, as there are several places where you can have a constant type family. *)
+Definition transportD_const' {A B : Type} (C : A -> B -> Type) {x1 x2 : A}
+  (p : x1 = x2) {y : B} (z : C x1 y)
+  : transportD (fun x => B) C p y z
+    = transport (fun x => C x (transport (fun _ => B) p y)) p
+      (transport (C x1) (transport_const p y)^ z).
+Proof.
+  by destruct p.
+Defined.
+
+Definition transportD_const {A : Type} (B : A -> Type) (C : Type)
+  {x1 x2 : A} (p : x1 = x2) (y : B x1) (z : C)
+  : transportD B (fun _ _ => C) p y z = z.
+Proof.
+  by destruct p.
+Defined.
+
+Definition transportDD_const {A : Type} (B : A -> Type) (C : Type)
+  {a1 a2 : A} (pA : a1 = a2)
+  {b1 : B a1} {b2 : B a2} (pB : transport B pA b1 = b2)
+  (c1 : C)
+  : transportDD B (fun _ _ => C) pA pB c1 = c1.
+Proof.
+  by destruct pB, pA.
+Defined.
 
 (** Transporting in a pulled back fibration. *)
 Lemma transport_compose {A B} {x y : A} (P : B -> Type) (f : A -> B)
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -994,19 +994,18 @@ Definition transport2_const {A B : Type} {x1 x2 : A} {p q : x1 = x2}
   : transport_const p y = transport2 (fun _ => B) r y @ transport_const q y
   := match r with idpath => (concat_1p _)^ end.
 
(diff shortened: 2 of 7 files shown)


theories/Basics/PathGroupoids.v
```diff
@@ -993,6 +994,32 @@ Definition transport2_const {A B : Type} {x1 x2 : A} {p q : x1 = x2}
...
+  : transportD (fun x => B) C p y z
+    = transport (fun x => C x (transport (fun _ => B) p y)) p
+      (transport (C x1) (transport_const p y)^ z).
+Proof.
+  by destruct p.
+Defined.
+
+Definition transportD_const {A : Type} (B : A -> Type) (C : Type)
+  {x1 x2 : A} (p : x1 = x2) (y : B x1) (z : C)
+  : transportD B (fun _ _ => C) p y z = z.
```
jdchristensen, line 1009: Maybe put this simpler, non-primed version before the other?

theories/Colimits/GraphQuotient.v
```diff
@@ -300,6 +301,39 @@ Section Flattening.
...
+  Context `{Univalence} {A : Type} {R : A -> A -> Type} (a0 : A)
+    (Pe : gqDescent R) (p0 : gqd_fam Pe a0)
+    (based_gqdescent_ind : forall (Qe : gqDepDescent Pe) (q0 : gqdd_fam Qe a0 p0),
+      gqDescentSection Qe)
+    (based_gqdescent_ind_beta : forall (Qe : gqDepDescent Pe) (q0 : gqdd_fam Qe a0 p0),
+      gqds_sect (based_gqdescent_ind Qe q0) a0 p0 = q0).
+
+  (** Under these hypotheses, we get an identity system structure on [Dgqd Pe]. *)
+  Local Instance idsys_flatten_gqdescent
+    : @IsIdentitySystem _ (gq a0) (Dgqd Pe) p0.
```
jdchristensen, line 320: The name `Dgqd` is a bit cryptic (and the same is true for the two parallel files).  Maybe a longer name would be better, as well as a reminder in the comment of what it means?

jdchristensen (commented):
The changes to Coeq.v and Pushout.v ended up adding a lot more code than I realized, compared to deducing flattening from the GraphQuotient case.  But maybe it's worth having the whole theory of descent spelled out in all three cases?  I'm not 100% sure.  @Alizter , what do you think?

theories/Basics/Equivalences.v
```diff
@@ -93,6 +93,33 @@ Section EquivTransport.
...
 End EquivTransport.
 
+(** Doubly dependent transport is an equivalence *)
+Section EquivTransportD.
+
+    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type) 
+      {x y : A} (p : x = y) {px : P x}.
+
+    Global Instance isequiv_transportD : IsEquiv (transportD P Q p px).
+    Proof.
```
Alizter, line 103: There is a shorter proof:
```coq
destruct p; exact _.
```
Since `transportD` just becomes `idmap` in that case.

theories/Basics/Equivalences.v
```diff
@@ -93,6 +93,33 @@ Section EquivTransport.
...
+        exact (transport (Q x) (transport_Vp _ p px)). }
+      all: by destruct p.
+    Defined.
+
+    Definition equiv_transportD : Q x px <~> Q y (transport P p px)
+      := Build_Equiv _ _ (transportD P Q p px) _.
+
+    Context {py : P y} (q : transport P p px = py).
+
+    Global Instance isequiv_transportDD : IsEquiv (transportDD P Q p q)
```
Alizter, line 115: This is not needed. You can make it a `Definition` and `:= _` since typeclasses can pick it up anyway.

theories/Basics/PathGroupoids.v
```diff
@@ -993,6 +994,32 @@ Definition transport2_const {A B : Type} {x1 x2 : A} {p q : x1 = x2}
   : transport_const p y = transport2 (fun _ => B) r y @ transport_const q y
   := match r with idpath => (concat_1p _)^ end.
 
+(** We need a better naming scheme for transportD_const, as there are several places where you can have a constant type family. *)
+Definition transportD_const' {A B : Type} (C : A -> B -> Type) {x1 x2 : A}
```
Alizter, line 998: An idea: How about `transportD_const`, `transportD_const_base` and `transportD_const_fiber`?

theories/Types/Sigma.v
```diff
@@ -265,6 +265,16 @@ Proof.
   destruct p. reflexivity.
 Defined.
 
+(** Doubly dependent transport is the same as transport along a [path_sigma]. *)
+Definition transportDD_is_transport {A : Type} (B : A -> Type) (C : forall a : A, B a -> Type)
```
Alizter, line 269: I'm inclined to call this `transport_path_sigma` and flip the direction. This way we are "computing" transports over `path_sigma`.

theories/Types/Sigma.v
```diff
@@ -265,6 +265,16 @@ Proof.
...
+  {a1 a2 : A} (pA : a1 = a2)
+  {b1 : B a1} {b2 : B a2} (pB : transport B pA b1 = b2)
+  (c1 : C a1 b1)
+  : transportDD B C pA pB c1 = transport (fun (ab : sig B) => C ab.1 ab.2) (path_sigma _ (a1; b1) (a2; b2) pA pB) c1.
```
Alizter: Also `sig_rec C` is probably a better function name here. Any map `sig C -> B` can be eta-expanded to a `sig_rec`. I would also rephrase this lemma as being about `path_sigma'` and add a variant with `a : sig B` instead for `path_sigma`.

Alizter (commented):
Here are some initial comments on the material coming before descent/flattening. I still think we can do a better job explaining what it actually is, but I will suggest changes at a later date after I've had some time to think and review.

Alizter:
I'm happy with the theory of descent in all cases. I'm weary that making things special cases of other things to reduce clutter will just make everything harder to use.

I am not a big fan of the naming convention for the descent stuff however, but I need some more time to conjure up a better suggestion.

theories/Basics/Equivalences.v
jdchristensen: That was the original proof, but I suggested that it would be better if the inverse computed.  This is analogous to how we prove `isequiv_transport`, which could also be done by destructing the path.

Alizter: Oh I see, in that case this can be ignored. Maybe a comment saying that we give an explicit inverse so that it can be reasoned about could be added.

theories/Basics/Equivalences.v
Alizter: Actually it's probably better to get rid of it completely.

theories/Types/Sigma.v
Alizter: I see that there already is `transportD_is_transport` so this isn't something you introduced. We can consider renaming this at a later date then.

theories/Types/Sigma.v
Alizter: I see you already changed this, so I am happy with how it is now.

Alizter:
What is the difference between descent data and recursion / induction data? I haven't been able to change it all the way through, but it seems if you replace the equivalence with a path, you can do most of the work here without univalence. I wonder if we are really just looking at a special case of a general fact about recursion data.

jdchristensen:
> What is the difference between descent data and recursion / induction data?

Descent data on, say, a graph A, R is the special case of recursion data for a function A/R -> Type, combined with univalence to change the paths to equivalences.  But because the codomain is Type instead of a particular type B, you can now talk about *sections* of the induced type family A/R -> Type, and talk about the descent data you need to define such a section.  And that data can be expressed cleanly in terms of the equivalences that define the descent data, instead of the paths they correspond to via univalence.

It might be possible to express a lot of this in terms of paths instead of equivalences, but univalence is certainly needed in some places, so it's probably best to keep things expressed in terms of equivalences.

Alizter:
>It might be possible to express a lot of this in terms of paths instead of equivalences, but univalence is certainly needed in some places, so it's probably best to keep things expressed in terms of equivalences.

Do you have an example where it is necessary? I couldn't see where it becomes so.

theories/Types/Sigma.v
jdchristensen: Replacing `(fun (ab : sig B) => C ab.1 ab.2)` with `sig_rec C` seems like it might be a good idea.  `sig_rec` is the dependent version of `uncurry`, and we should maybe use it more often.

jdchristensen:
> > It might be possible to express a lot of this in terms of paths instead of equivalences, but univalence is certainly needed in some places, so it's probably best to keep things expressed in terms of equivalences.
> 
> Do you have an example where it is necessary? I couldn't see where it becomes so.

Well, it's known that flattening and univalence are equivalent, so that's why it has to come up.  I think it will explicitly appear when you have some transport equivalences that you have to regard as paths, but since the current formalization is done fully with equivalences, it's hard to point to a specific place.

Alizter:
@jdchristensen Right, so the stuff about DescentData *could* be a special case of recursion data. The stuff about flattening is actually where the univalence comes in. Therefore I am thinking that it might be better to do everything for a `RecData` instead and introduce univalence for flattening.

Alizter:
Anyway, its not super important right now, mostly I was just wondering.

ThomatoTomato:
The motivation to introduce descent data was to break up the characterization of path spaces into smaller pieces. Dan noticed that this proved part of the flattening, so we decided to incorporate descent into flattening. In this case, both flattening and characterization of paths depend on univalence. Having record types for dependent descent data and sections of descent data made the code a lot more readable, so I don't suggest giving those up. There might be a good way to state the principles we need using recursion data instead of descent data, but I don't know how on top of my head.

theories/Colimits/GraphQuotient.v
```diff
@@ -70,169 +71,269 @@ Proof.
...
   Defined.
 
-  (** We can also prove an induction principle for [sig DGraphQuotient]. We won't show that it satisfies the relevant computation rules as these will not be needed. Instead we will prove the non-dependent eliminator directly so that we can better reason about it. In order to get through the path algebra here, we have opted to use dependent paths. This makes the reasoning slightly easier, but it should not matter too much. *)
-  Definition flatten_ind {Q : sig DGraphQuotient -> Type}
-    (Qgq : forall a (x : F a), Q (flatten_gq x))
-    (Qgqglue : forall a b (s : R a b) (x : F a),
-      flatten_gqglue s x # Qgq _ x = Qgq _ (e _ _ _ x))
-    : forall x, Q x.
+  (** A section of dependent descent data [Qe : gqDepDescent Pe] is a fiberwise section [gqds_sect], together with coherences [gqdd_e]. *)
+  Record gqDescentSection {Pe : gqDescent R} (Qe : gqDepDescent Pe) := {
```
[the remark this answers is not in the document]
ThomatoTomato: > I guess this should be called `gqDepDescentSection`

The name `gqDescentSection` should also makes sense in its own regard. If you uncurry `Qe`, then it is singly-dependent in the total space `sig (Dgqd Pe)`, so *sections* of this space are exactly *sections* of the descent data. If it matters, Agda UniMath uses the name section for such data.

> And that it would make sense to add `gqDescentSection` giving the data needed for a section of `Dgqd Pe` for the singly-dependent case, just for completeness?

I'm not so sure what this is supposed to add further. The point of `gqDescentSection` and `gqDescentSectionConst` is that we can express induction and recursion principles for these records. If the intermediate data type has an interesting induction principle, then I would be fully on-board.

Alizter:
I've checked and its possible to remove `Univalence` completely from DescentData leaving just a `Funext` assumption. So it really is a special case of rec data. It also lets you prove flattening where you ask only for an equality at the type level. Its then possible to use univalence as a special case to build descent data. But all-in-all descent data is just a special case of rec data.

Alizter:
However, I don't want to ask that you change the direction of this PR, so far everything looks good and it will be a useful addition. I just think there is a cleaner concept at play here which we can think about in the future.

jdchristensen:
> I've checked and its possible to remove `Univalence` completely from DescentData leaving just a `Funext` assumption.

Wow, I didn't realize that would be possible.  Of course, univalence will be needed to provide the paths in Type, but I thought it would also be needed to prove any kind of flattening lemma.  The diff you provided has some awkward parts, but I suspect that with some clean-up it might become pretty reasonable.  Of course, every time someone wants to *use* the various results, they would need to invoke univalence, so that might make this a bit more awkward.  And the second field of the `gqDescentSection` Record would be more annoying to deal with.  Still, it's really worth thinking about whether we should do it in this form.

Does the section on Kraus-von-Raumer also go through?

Alizter:
> Does the section on Kraus-von-Raumer also go through?

Yes, just with `funext`.

theories/Colimits/GraphQuotient.v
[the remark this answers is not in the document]
jdchristensen: In the Descent Section, first you give the data for defining a type family P over the quotient A/R.  Then you give the data for defining a type family Q over the total space of P.  Then you give the data for defining a section of Q over the total space of P.  But you missed giving the data for how to define a section of P itself (see below).  I don't know if this is implicitly coming up later in the file (which means it could factor through this), but even if not, it seems like a very natural thing to have.  Once we know how to make bundles over A/R, we should know how to make sections of them.

And even if this is not included, the name of the Record you give should be DepDescentSection, since it describes sections of the bundles you create using DepDescent data.

```coq
  Record gqDescentSection' (Pe : gqDescent R) := {
    gqds_sect' (a : A) : gqd_fam Pe a;
    gqds_e' {a b : A} (r : R a b) 
      : gqd_e Pe r (gqds_sect' a) = gqds_sect' b
  }.

  (** A descent section induces a genuine section of [Dgqd Pe]. *)
  Definition gqdescent_ind' {Pe : gqDescent R}
    (f : gqDescentSection' Pe)
    : forall (x : GraphQuotient R), Dgqd Pe x.
  Proof.
    snrapply (GraphQuotient_ind _ (gqds_sect' _ f)).
    intros a b r.
    refine (transport_Dgqd_gqglue Pe r _ @ gqds_e' _ f r).
  Defined.
```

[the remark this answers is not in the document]
jdchristensen: (I put primes above to distinguish it from the existing names, but if you rename the existing things, then the primes would of course not be needed.)

Alizter:
Again, sorry for hijacking this thread Thomas, but it seems the general pattern here is that we have a (higher) inductive type which corresponds to some monad looking thing. Then the recursion and induction data correspond to algebras and "displayed algebras" respectively. Sections of these displayed algebras are the sections that appear here. Flattening says something about the colimits in this category of algebras.

I haven't been motivated enough to work out the full details, but this is the gist I got from what is going on here. The reason I've noticed this now is because the same story occurs for all sorts of other structures we work on.

ThomatoTomato:
You are raising valuable points Ali. I'm not sure if it is best to spend time to do it this way right now, or if we should make an issue that references this thread, so we remember this for the future.

jdchristensen:
I think anytime these results would be used, it would be via equivalences (and therefore with univalence), so I think it's best to leave things as they are.  Otherwise we'd need a path-based version, and then almost as much code again to have wrappers which convert everything to equivalences.  What I propose is that we add a comment to all three files saying something like:
```coq
(** It's possible to prove the results in the Descent, Flattening and Paths Sections without univalence, by replacing all equivalences with paths, but in practice these results will be used with equivalences as input, making the form below more convenient.  See https://github.com/HoTT/Coq-HoTT/pull/2147#issuecomment-2521570830 for further information. *)
```

As a separate point, before the flattening lemma for coequalizers, we should add something like:
```coq
(** The flattening lemma here also follows from the flattening lemma for GraphQuotients, avoiding the need for the material in Section Descent.  However, that material is likely useful in general, so we have given an independent proof of the flattening lemma.  See versions of the library before December 5, 2024 for the proof using the flattening lemma for GraphQuotients. *)
```
and similarly for pushouts, with GraphQuotients replaced twice by Coeq.

theories/Colimits/GraphQuotient.v
ThomatoTomato: In the old files, this type family was known as `DGraphQuotient`. However, its analogue in Coeq.v was named `flatten_coeq_fam`. Could `flatten_gqd_fam` work as a name instead of `Dgqd`?

theories/Basics/Equivalences.v
```diff
@@ -84,6 +84,7 @@ Section EquivTransport.
 
   Context {A : Type} (P : A -> Type) {x y : A} (p : x = y).
 
+  (** The inverse and the homotopies of transport is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
```
jdchristensen, line 87: Plural subject -> "are".  Typo in "explictly".  Please proofread comments to save review time...

theories/Basics/Equivalences.v
```diff
@@ -96,9 +97,10 @@ End EquivTransport.
 (** Doubly dependent transport is an equivalence *)
 Section EquivTransportD.
 
-    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type) 
+    Context {A : Type} (P : A -> Type) (Q : forall a : A, P a -> Type)
       {x y : A} (p : x = y) {px : P x}.
 
+    (** The inverse of transportD is defined explictly. This allows us to reason about the map when the input is not [idpath]. *)
```
jdchristensen, line 103: Typo.

theories/Colimits/Coeq.v
```diff
@@ -480,23 +482,46 @@ Section Descent.
...
+  Definition transport_induced_fam_cd_cglue
     (Pe : cDescent f g) (b : B) (pf : cd_fam Pe (f b))
-    : transport (Dcd Pe) (cglue b) pf = cd_e Pe b pf.
+    : transport (induced_fam_cd Pe) (cglue b) pf = cd_e Pe b pf.
   Proof.
     nrapply transport_path_universe'.
     nrapply Coeq_rec_beta_cglue.
   Defined.
 
+  (** A section on the descent data is a fiberwise section that respects the equivalence. *)
```
jdchristensen, line 502: "equivalences".

theories/Colimits/Coeq.v
```diff
@@ -508,101 +533,101 @@ Section Descent.
   Global Arguments cdd_fam {Pe} Qe a pa : rename.
   Global Arguments cdd_e {Pe} Qe b pf : rename.
 
-  (** A dependent type family over [Dcd Pe] induces dependent descent data over [Pe]. *)
-  Definition cdepdescent_fam {Pe : cDescent f g}
-    (Q : forall x : Coeq f g, (Dcd Pe) x -> Type)
+  (** A dependent type family over [induced_fam_cd Pe] induces dependent descent data over [Pe]. *)
+  Definition induced_fam_cdd {Pe : cDescent f g}
```
jdchristensen, line 537: I liked the old name `cdepdescent_fam` better, since it starts with the type of the output.  And `cdd` is a bit cryptic.  In fact, the name `induced_fam_cdd` is very misleading when you compare it to `induced_fam_cd`; they are doing the opposite thing, aren't they?

jdchristensen: Actually, aren't we missing the function that starts with `cDepDescent` data and produces a bundle over `induced_fam_cd Pe`?  We don't need it, but either we should include it, or we should have a comment saying that `cDepDescent` data will induce a bundle `induced_fam_cd Pe`, but we don't need that so we won't use it.

jdchristensen (commented):
I didn't read GraphQuotient.v or Pushout.v, but presumably similar comments apply to them.

theories/Colimits/Coeq.v
ThomatoTomato: I agree with what you are saying, since they are in fact doing the opposite thing. Would it be better to name `cdepdescent_fam` as `cdepdescent_coeq_fam`? Then the opposite operation can be reversing the output and input, such as `coeq_fam_cdepdescent`.

theories/Colimits/GraphQuotient.v
```diff
@@ -70,169 +71,294 @@ Proof.
...
+    {a b : A} (r : R a b)
+    : apD (gqdepdescent_ind f) (gqglue r) = transport_gqds_gqglue f a b r
+    := GraphQuotient_ind_beta_gqglue _ (gqdds_sect f) (transport_gqds_gqglue f) a b r.
+
+  (** The data for a section into a constant type family. *)
+  Record gqDepDescentConstSection (Pe : gqDescent R) (Q : Type) := {
+    gqdcs_sect (a : A) : gqd_fam Pe a -> Q;
+    gqdcs_e {a b : A} (r : R a b) (pa : gqd_fam Pe a)
+      : gqdcs_sect a pa = gqdcs_sect b (gqd_e Pe r pa)
+  }.
```
ThomatoTomato, line 208: In this record, the fields are named `gqdcs_sect` and `gqdcs_e`. These names belong to a pattern made across the different records in Descent. `gqdcs_sect` is an abbreviation for "graph quotient dependent descent constant section section". (I realize now that there's missing a "d", as it should be `gqddcs_sect`.) However, this is very vague and cryptic. Is there a better way to make up names for these entries, or should I explain the nomenclature behind these names?

theories/Colimits/Coeq.v
jdchristensen: I thought that the `c` in `cdepdescent` already stood for `coeq`?  So adding `coeq` here wouldn't fit the pattern used in the rest of the file.  I think I'm happy with `cdepdescent_fam`.  Then `induced_fam_cd` should be `fam_cdescent`, and the definition we are missing would be `fam_cdepdescent` if we include it.

theories/Colimits/GraphQuotient.v
jdchristensen: I think this is ok with the extra `d` added.

theories/Colimits/GraphQuotient.v
```diff
@@ -70,169 +71,321 @@ Proof.
...
+  Global Arguments gqdds_sect {Pe Qe} f a pa : rename.
+  Global Arguments gqdds_e {Pe Qe} f {a b} r pa : rename.
+
+  (** Transporting [gqdds_sect f a] over [Q] along [gqglue r] is [gqdds_sect f b]. *)
+  Definition transport_fam_gqglue_gqdds_sect {Pe : gqDescent R}
+    {Q : forall x : GraphQuotient R, (fam_gqdescent Pe) x -> Type}
+    (f : gqDepDescentSection (gqdepdescent_fam Q))
+    (a b : A) (r : R a b)
+    : transport (fun x : GraphQuotient R => forall px : fam_gqdescent Pe x, Q x px)
+        (gqglue r) (gqdds_sect f a) = gqdds_sect f b.
```
ThomatoTomato, line 206: I realized that the names of the different transport lemmas were inconsistent. I changed `transport_gqdds_gqglue` to `transport_fam_gqglue_gqdds_sect`. Every such transport lemma are now of the form `transport_"type family"_"path constructor"_"optional function"`. Does this look good?

theories/Colimits/Coeq.v
ThomatoTomato: Since we only did `fam_cdescent`, I also added `cdescent_fam`.

theories/Colimits/GraphQuotient.v
jdchristensen: For this one, shouldn't `fam` be replaced with `fam_gqdescent`?  And I think it might be better to remove the `_gqdds_sect`, since we only have one transport lemma of the form `transport_fam_gqdescent_gqglue`.

(BTW, can you mark as resolved any comments that you have handled?)

ThomatoTomato: In this case, the type family we want to transport over is the input `Q`, not `fam_gqdescent Pe`. I'm not sure how to represent this, so I went with `fam`. I think I could also just drop `fam`, and write `transport_gqglue_gqdds_sect`. We also already have a version for `transport_fam_gqdescent_gqglue`.

ThomatoTomato:
The partial computation became a lot more troublesome to state without this lemma. Since it is never used anywhere, I don't think it hurts to remove it.

jdchristensen:
> The partial computation became a lot more troublesome to state without this lemma. Since it is never used anywhere, I don't think it hurts to remove it.

Ah, I missed that the lemma was also used for the partial computation rule.  But since that rule isn't needed, I think it's fine to remove it.

theories/Colimits/Coeq.v
```diff
@@ -457,60 +458,295 @@ Proof.
...
+    snrapply (Coeq_ind _ (cddcs_sect s)).
+    intro b.
+    nrapply dpath_arrow.
+    intro pf.
+    lhs nrapply transport_const.
+    rhs nrapply (ap _ (transport_fam_cdescent_cglue Pe b pf)).
+    exact (cddcs_e s b pf).
+  Defined.
+
+  (** In this case, we state a full computation rule. *)
```
jdchristensen, line 627: Rephrase, since the partial computation rule in the other case is gone.  Maybe just "Here is the computation rule on paths." or something like that.

jdchristensen:
@ThomatoTomato @Alizter Is there anything else left to address here?  (Besides my minor suggestion about the comment.)

Alizter:
I would like to do another pass today.

theories/Colimits/Pushout.v
```diff
@@ -538,31 +539,312 @@ Proof.
...
+
+  (** The data for a section into a constant type family. *)
+  Record poDepDescentConstSection (Pe : poDescent f g) (Q : Type) := {
+    poddcs_sectl (b : B) : pod_faml Pe b -> Q;
+    poddcs_sectr (c : C) : pod_famr Pe c -> Q;
+    poddcs_e (a : A) (pf : pod_faml Pe (f a))
+      : poddcs_sectl (f a) pf = poddcs_sectr (g a) (pod_e Pe a pf)
+  }.
+
+  Global Arguments Build_poDepDescentConstSection {Pe Q} poddcs_sectl poddcs_sectr poddcs_e.
```
Alizter, line 702: If you make Pe and Q implicit in the definition of the record, you can replace these with a single arguments command by make the Pe and Q explicit for the record.

Alizter (approved):
This will be a great addition to the library. What's a good example for demonstrating the Kraus-von Raumer induction principle for pushouts? Will it let us construct an induction principle for `loops (psusp A)` for instance?

ThomatoTomato:
Following David Wärn's notes, me and a classmate are working on a PR where we will use this to give another characterization of path spaces of pushouts. This will hopefully allow us to prove various things about pushouts.

> Will it let us construct an induction principle for `loops (psusp A)` for instance?

I'm not sure if this would work, since I believe that one of the endpoints have to be free.

Alizter:
Great. I'll let @jdchristensen merge when he is ready.

jdchristensen (approved):

PR #2151, opened 2024-12-09
double recursion for join

Alizter:
Here is a double recursion principle for `Join`. It is slightly more general than a typical double recursion principle since we allow for the types in each join to be different. This generalisation makes the overall lemma clearer to state and prove.

One place this will be used is in CayleyDickson.v where we define a H-space multiplication operation on `Join (Susp A) (Susp A)` for a Cayley-Dickson imaginaroid `A`. This lemma makes the construction there much clearer.

Presumably there ought to be a dependent variant of this, but I haven't yet thought about how to state or prove it.

### DIFF
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -883,3 +883,36 @@ Section JoinPower.
   Defined.
 
 End JoinPower.
+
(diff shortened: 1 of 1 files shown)


jdchristensen (approved):
Nice!

PR #2152, opened 2024-12-09
cleanup in CayleyDickson and fix #2149

### LINKED ISSUE #2149: Mistake in CayleyDickson.v
In CayleyDickson.v we mistakenly define conjugation on a suspension as:

https://github.com/HoTT/Coq-HoTT/blob/e56b4fa8dd6f4313d52f8c24519e709ef185aef9/theories/Homotopy/CayleyDickson.v#L66-L74

It should be `merid (-a)` rather than `merid a`.

This means that the H-space structure we get out of the Cayley-Dickson construction is incorrect.

Alizter:
We fix the issue in #2149 which didn't affect much of the later proofs in the file. I don't know why this doesn't affect anything later on. It seems like the H-space construction would have gone through still?

We take this opportunity to clean up the argument in this file. For instance, we have a derived negation and conjugation on the suspension, with further properties given properties of the underlying negation. This all works without needing to assume a Cayley-Dickson imaginaroid structure.

We clean up the argument in `cd_op` making it easier to read, prove and use. We could add beta rules for `Join_rec2` in the future, to make computations easier. We currently don't need to do this yet.

### DIFF
--- a/theories/Homotopy/CayleyDickson.v
+++ b/theories/Homotopy/CayleyDickson.v
@@ -82,12 +82,10 @@ End CayleyDicksonSpheroid_Properties.
 
 (** ** Negation and conjugation on suspensions *)
 
-Global Instance conjugate_susp (A : Type) `(Negate A)
-  : Conjugate (Susp A)
+Global Instance conjugate_susp (A : Type) `(Negate A) : Conjugate (Susp A)
   := functor_susp (-).
 
-Global Instance negate_susp (A : Type) `(Negate A)
-  : Negate (Susp A)
+Global Instance negate_susp (A : Type) `(Negate A) : Negate (Susp A)
   := susp_neg _ o conjugate_susp A (-).
 
 (** [conjugate_susp A] and [negate_susp A] commute. *)
--- a/theories/Homotopy/CayleyDickson.v
+++ b/theories/Homotopy/CayleyDickson.v
@@ -84,7 +84,7 @@ End CayleyDicksonSpheroid_Properties.
 
 Global Instance conjugate_susp (A : Type) `(Negate A)
   : Conjugate (Susp A)
-  := Susp_rec North South (fun a => merid (-a)).
+  := functor_susp (-).
 
 Global Instance negate_susp (A : Type) `(Negate A)
   : Negate (Susp A)
@@ -117,8 +117,8 @@ Proof.
   lhs nrapply concat_p1.
   rhs nrapply concat_1p.
   lhs nrapply ap.
-  1: nrapply Susp_rec_beta_merid.
-  lhs nrapply Susp_rec_beta_merid.
+  1: nrapply functor_susp_beta_merid.
+  lhs nrapply functor_susp_beta_merid.
   nrapply ap.
   rapply (involutive x).
 Defined.
@@ -132,13 +132,13 @@ Proof.
   lhs nrapply concat_p1.
   rhs nrapply concat_1p.
   rhs nrapply ap.
-  2: nrapply Susp_rec_beta_merid.
+  2: nrapply functor_susp_beta_merid.
   rhs nrapply Susp_rec_beta_merid.
   lhs nrapply ap.
   1: nrapply Susp_rec_beta_merid.
   lhs nrapply (ap_V _ (merid (-x))).
   apply ap.
-  nrapply Susp_rec_beta_merid.
+  nrapply functor_susp_beta_merid.
 Defined.
 
 (** ** Cayley-Dickson imaginaroids *)
--- a/theories/Homotopy/CayleyDickson.v
+++ b/theories/Homotopy/CayleyDickson.v
@@ -88,57 +88,35 @@ Global Instance conjugate_susp (A : Type) `(Negate A)
 
 Global Instance negate_susp (A : Type) `(Negate A)
   : Negate (Susp A)
-  := Susp_rec South North (fun a => (merid (-a))^).
-
-Global Instance involutive_negate_susp {A} `(Negate A, !Involutive (-))
-  : Involutive (negate_susp A (-)).
-Proof.
-  snrapply Susp_ind_FFlr.
-  1,2: reflexivity.
-  intro x.
-  lhs nrapply concat_p1.
-  rhs nrapply concat_1p.
-  lhs nrapply ap.
-  1: nrapply Susp_rec_beta_merid.
-  lhs nrapply (ap_V (negate_susp A negate) (merid (-x))).
-  lhs nrapply ap.
-  1: nrapply Susp_rec_beta_merid.
-  lhs nrapply inv_V.
-  nrapply ap.
-  rapply (involutive x).
-Defined.
+  := susp_neg _ o conjugate_susp A (-).
 
 Global Instance involutive_conjugate_susp {A} `(Negate A, !Involutive (-))
   : Involutive (conjugate_susp A (-)).
 Proof.
-  snrapply Susp_ind_FFlr.
-  1,2: reflexivity.
-  intro x.
-  lhs nrapply concat_p1.
-  rhs nrapply concat_1p.
-  lhs nrapply ap.
-  1: nrapply functor_susp_beta_merid.
-  lhs nrapply functor_susp_beta_merid.
-  nrapply ap.
-  rapply (involutive x).
+  intros x.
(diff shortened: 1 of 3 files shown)


theories/Homotopy/CayleyDickson.v
```diff
@@ -38,49 +39,87 @@ Section CayleyDicksonSpheroid_Properties.
 
   Context {X : pType} `(CayleyDicksonSpheroid X).
 
+  Local Instance isequiv_cds_conjug : IsEquiv cds_conjug
+    := isequiv_adjointify cds_conjug cds_conjug cds_conjug_inv cds_conjug_inv.
```
jdchristensen, line 43: If everything else is a Global Instance, maybe this should be too?  Also, does `isequiv_involution` work here?

jdchristensen (commented):
I'll look at this closely another day.  My only comment right now is about the comments.  Since you are fixing things up, maybe you can add square brackets around the Coq terms, fix the "multiplciation" typo, and change "a H-space" to "an H-space"?

jdchristensen:
> We fix the issue in #2149 which didn't affect much of the later proofs in the file. I don't know why this doesn't affect anything later on. It seems like the H-space construction would have gone through still?

I haven't looked at this material for a while, so I can't remember exactly how it all goes.  But if that wrong definition of the negation (as the identity map) still made the file go through, does that mean that the assumption that A has a negation can be dropped?  Oh, I guess it's also used in the definition of the conjugation.  Still, I'm a bit surprised that more than one definition of negation makes things go through.

theories/Homotopy/CayleyDickson.v
```diff
@@ -38,49 +39,87 @@ Section CayleyDicksonSpheroid_Properties.
...
 End CayleyDicksonSpheroid_Properties.
 
 Global Instance conjugate_susp (A : Type) `(Negate A)
-  : Conjugate (Susp A).
+  : Conjugate (Susp A)
+  := Susp_rec North South (fun a => merid (-a)).
+
+Global Instance negate_susp (A : Type) `(Negate A)
+  : Negate (Susp A)
+  := Susp_rec South North (fun a => (merid (-a))^).
```
jdchristensen, line 72: Note that `conjugate_susp` is just `functor_susp` applied to the negation map.  And `negate_susp` is the composite of `conjugate_susp` followed by the reflection that every suspension has (swapping north and south, and inverting the meridians).  Moreover, `functor_susp` and this reflection always commute.  So I think the facts that both of these are involutions can be proved using general properties about `functor_susp`, without needing to using any induction directly.

jdchristensen: The attached file contains the lemmas you need about the reflection on a suspension.  (Called negation in the file.)

They you'd need a (easy) lemma saying that a composite of commuting involutions is an involution.

[SuspLemmas.txt](https://github.com/user-attachments/files/18071225/SuspLemmas.txt)

Alizter:
I think the definition of the negation as pole flipping and meridian negating is correct, it was the definition of conjugation that was the identity map before. Given what the construction is supposed to replicate, that seems very wrong so I must be missing something here.

Alizter:
I've added some comments and fixed others like you've suggested. I added a short introduction to the file detailing where the construction comes from.

theories/Homotopy/CayleyDickson.v
Alizter: Thanks, that was very useful.

Alizter:
The entire file goes through if I remove conjugation completely. Is conjugation a red herring here? Is the actual heavy-lifter just `susp_neg` all along?

jdchristensen:
> I think the definition of the negation as pole flipping and meridian negating is correct, it was the definition of conjugation that was the identity map before.

Right, I had them reversed in my comment.

> The entire file goes through if I remove conjugation completely. Is conjugation a red herring here? Is the actual heavy-lifter just `susp_neg` all along? I suspect that something must go wrong for [S^3] where the H-space structure we get might be degenerate in some way.

Can you say more precisely what you mean by removing conjugation completely?  It's currently used to define negation.  In any case, I think what you are saying matches what we already knew before this PR:  conjugation can be taken to be the identity map, and the proof goes through.

I don't think there is any sense in which an H-space structure on S^3 can be degenerate.  But some H-space structure have extra properties, like being left- or right-invertible, associative, etc.

Alizter:
> Can you say more precisely what you mean by removing conjugation completely? It's currently used to define negation. In any case, I think what you are saying matches what we already knew before this PR: conjugation can be taken to be the identity map, and the proof goes through.

Specifically if you replace the definition of `conjugate_susp` with `idmap` and adapt a few proofs.

<details>
<summary>Diff</summary>

```diff
diff --git a/theories/Homotopy/CayleyDickson.v b/theories/Homotopy/CayleyDickson.v
index 6c1077b06..97bf45fb9 100644
--- a/theories/Homotopy/CayleyDickson.v
+++ b/theories/Homotopy/CayleyDickson.v
@@ -83,7 +83,7 @@ End CayleyDicksonSpheroid_Properties.
 (** ** Negation and conjugation on suspensions *)
 
 Global Instance conjugate_susp (A : Type) `(Negate A) : Conjugate (Susp A)
-  := functor_susp (-).
+  := idmap.
 
 Global Instance negate_susp (A : Type) `(Negate A) : Negate (Susp A)
   := susp_neg _ o conjugate_susp A (-).
@@ -92,29 +92,21 @@ Global Instance involutive_conjugate_susp {A} `(Negate A, !Involutive (-))
   : Involutive (conjugate_susp A (-)).
 Proof.
   intros x.
-  lhs_V nrapply functor_susp_compose.
-  rhs_V nrapply functor_susp_idmap.
-  nrapply functor2_susp.
-  rapply involutive.
+  reflexivity.
 Defined.
 
 Global Instance involutive_negate_susp {A} `(Negate A, !Involutive (-))
   : Involutive (negate_susp A (-)).
 Proof.
   intros x.
-  unfold negate_susp.
-  lhs_V nrapply ap.
-  1: nrapply susp_neg_natural. 
-  lhs rapply susp_neg_inv.
-  rapply involutive.
+  rapply susp_neg_inv.
 Defined.
 
 Global Instance swapop_conjugate_susp {A} `(Negate A)
   : SwapOp negate (conjugate_susp A (-)).
 Proof.
   intros x.
-  symmetry.
-  nrapply susp_neg_natural.
+  reflexivity.
 Defined.
 
 (** ** Cayley-Dickson imaginaroids *)
```


</details>

 
> I don't think there is any sense in which an H-space structure on S^3 can be degenerate. But some H-space structure have extra properties, like being left- or right-invertible, associative, etc.

I'm wondering what happens to the Hopf fibration map when you choose the other H-space structures on the 3-sphere. Do you end up picking different generators in the stable homotopy groups?

jdchristensen:
Ok, I re-read the paper by Ulrik and Egbert, and I understand what is going on.  Their construction works for *any* involutive negation map on A, as long as the imaginaroid axioms are satisfied, so of course it also works if you assume that the negation map is the identity.  However, for examples, you may need to choose the negation map to be something other than the identity.  For example, to get the H-space structure on S^3, you need an imaginaroid structure on S^0.  This requires an H-space structure on S^1 with the property that x* x = 1.  But if conjugation is the identity map, then this won't hold, so you won't be able to get the H-space structure on S^3 in this way.

So it's good that you caught this error, but it's not really a mystery any more why the result went through with that error in place.  As a real test, we should use this file to get the H-space structure on S^3.

theories/Homotopy/CayleyDickson.v
```diff
@@ -242,93 +232,70 @@ Section ImaginaroidHSpace.
   Arguments f {_ _}.
   Arguments g {_ _}.
 
-  (** Here is the multiplication map in algebraic form:
-      (a,b) * (c,d) = (a * c - d * b*, a* * d + c * b)
-      the following is the spherical form. *)
-  Global Instance cd_op
-    : SgOp (pjoin (psusp A) (psusp A)).
+  (** Here is the multiplication map in algebraic form: [(a,b) * (c,d) = (a * c - d * b*, a* * d + c * b)] the following is the spherical form. *)
```
jdchristensen, line 235:
```suggestion
  (** Here is the multiplication map in algebraic form: [(a,b) * (c,d) = (a * c - d * b*, a* * d + c * b)].  The following is the spherical form. *)
```

jdchristensen (approved):

PR #2153, opened 2024-12-12
Injective Types

doolster:
In this PR we introduce the notion of injective types and prove a number of results about them following the paper: Injective Types in Univalent Mathematics by Martin Escardo. This includes classifying algebraic injective types as retracts of exponentials of universes, showing injectivity is equivalent to the propositional truncation of algebraic injectivity (up to a universe mis-match), and excluded middle is equivalent to all pointed types being algebraically injective. As a part of this we also formalize a number of results about certain constructions that we can consider as Kan extensions of type families. Finally, we also prove some results about the flabbiness (equivalently injectivity) of sigma types over an injective type which can be found in the type topology library: https://martinescardo.github.io/TypeTopology/InjectiveTypes.MathematicalStructuresMoreGeneral.html. This includes the injectivity of the type of pointed types and the injectivity of reflective subuniverses.


theories/InjectiveTypes/InjectiveTypes.v
```diff
@@ -0,0 +1,422 @@
...
+(** If all pointed types are algebraically flabby, then excluded middle holds. *)
+Definition lem_pointed_types_alg_flab@{w} `{Funext}
+  (ptaf: forall (D : Type@{w}), D -> IsAlgebraicFlabbyType@{w w} D)
+  : ExcludedMiddle_type@{w}.
+Proof.
+  intros P PropP.
+  snrapply decidable_alg_flab_hprop.
+  - apply ptaf.
+    exact (inr tt).
+  - exact PropP. (* Don't know why type class search doesn't find this. *)
```
Alizter, line 421: Because `decidable_alg_flab_hprop` is asking for a `HProp` which Coq tries to build, but doesn't perform typeclass search on. You can fix this by asking for `P : Type@{w}` together with `IsHProp P` instead. Then you will see that `rapply` will do what you wish.

Alizter: Alternatively you can do `rapply (decidable_alg_flab_hprop (Build_HProp P)).`. The technical reason why Coq can build this type is because we use a "canonical structure" in line 411 of `Basics/Trunc.v`. This basically says to Coq "if you need a TruncType but you have only the projection of it (Type), then feel free to insert a Build_TruncType in there without doing typeclass resolutions". This is the only place we use this feature and it doesn't really fit our other paradigms too well.

Alizter (commented):
Very nice! Some of the results are quite technical and I haven't been able to digest all of the details of the paper just yet, so this is a preliminary pass. I hope to do another pass by the end of this weekend.

theories/InjectiveTypes/TypeFamKanExt.v
```diff
@@ -94,16 +94,14 @@ Proof.
   - intros g. by apply path_forall.
 Defined.
 
-(** Here we are taking the perspective that a type family [P : X -> Type] is a sort of oo-presheaf, considering the interpretation of [X] as an oo-groupoid and [Type] as a universe of spaces i.e. an appropriate generalization of the category of sets. It is easy to see that a type family [P] is functorial if we define its action on paths with [transport]. Functoriality then reduces to known lemmas about the [transport] function. *)
+(** Here we are taking the perspective that a type family [P : X -> Type] as an oo-presheaf, considering the interpretation of [X] as an oo-groupoid and [Type] as a universe of spaces i.e. an appropriate generalization of the category of sets. It is easy to see that a type family [P] is functorial if we define its action on paths with [transport]. Functoriality then reduces to known lemmas about the [transport] function. *)
```
jdchristensen, line 97: "is an" not "as an".  (Or rephrase, e.g. "Here we are thinking of a type family [P : X -> Type] as an ...")

theories/InjectiveTypes/InjectiveTypes.v
```diff
@@ -415,8 +413,7 @@ Definition lem_pointed_types_alg_flab@{w} `{Funext}
   : ExcludedMiddle_type@{w}.
 Proof.
   intros P PropP.
-  snrapply decidable_alg_flab_hprop.
-  - apply ptaf.
-    exact (inr tt).
-  - exact PropP. (* Don't know why type class search doesn't find this. *)
+  apply (decidable_alg_flab_hprop (@Build_HProp P PropP)).
```
jdchristensen, line 416: `(Build_HProp P)` works.

theories/InjectiveTypes/InjectiveTypes.v
jdchristensen: Ah, I didn't realize that Coq doesn't do type class search with Canonical Structures.  It seems it doesn't even search the hypotheses in the current context.  That makes it seem unlikely to ever be useful, but I must be misunderstanding something.

theories/InjectiveTypes/TypeFamKanExt.v
Alizter: That's my fault from before https://github.com/HoTT/Coq-HoTT/pull/2153#discussion_r1882997976

I was trying to emphasise that it literally is an oo-presheaf.

jdchristensen:
One question is where this material should go.  A new top-level folder is an option, but I'd prefer to have fewer top-level files and folders, so maybe we should make it a sub-folder of Homotopy?  Or I've been thinking of making a Misc folder for things like BoundedSearch.v, Idempotents.v, Projective.v, etc, and if we do that, it could also hold the InjectiveTypes folder.

Alizter:
> One question is where this material should go. A new top-level folder is an option, but I'd prefer to have fewer top-level files and folders, so maybe we should make it a sub-folder of Homotopy? Or I've been thinking of making a Misc folder for things like BoundedSearch.v, Idempotents.v, Projective.v, etc, and if we do that, it could also hold the InjectiveTypes folder.

That's a good question. I think in the broadest definition of the subject, `Homotopy/` should contain all of these but `BoundedSearch.v`. I'm hesitant to introduce a `Misc` folder since I think it might introduce laziness. However, I can be convinced otherwise.

jdchristensen:
I've created #2155 to discuss the overall organization of the top-level theories/ folder.

theories/Types/Forall.v
```diff
@@ -130,15 +130,31 @@ Defined.
...
+  := ap_lambdaD _ _.
+
+Definition ap_postcompose {B P Q : Type}
+  {f g : Q -> P} (h : f = g) (j : P -> B)
+  : ap (fun (k : Q -> P) => j o k) h
+    = path_forall (j o f) (j o g ) (fun q => ap j (ap10 h q)).
+Proof.
+  destruct h; cbn.
+  symmetry; apply path_forall_1.
+Defined.
```
Alizter, line 156: I think these may belong in `Types/Arrow.v`. There are similar functions there. For instance, ap_postcompose looks like it is just `ap_lambda` from there.

jdchristensen: `ap_precompose` can be generalized to to dependent functions as shown below, and that general version can stay here.  I couldn't figure out a simple way to generalize `ap_postcompose`, as the types get tricky.

For the non-dependent versions already in this PR, they can be adjusted to use `path_arrow` instead of `path_forall`, and so it does make sense to move them to Types/Arrow.v.  I have included the modified versions below.  But I don't see how to prove `ap_postcompose` using `ap_lambda`.

With these changes, the injective types material might need to unfold some path_arrow's.  Not sure.

```coq
(* For Types/Forall.v: *)
(** The action of pre-composition on a path between dependent functions.  See also [ap10_ap_precompose] in PathGroupoids.v and [ap_precompose] in Arrow.v. *)
Definition ap_precomposeD {B Q : Type} {P : Q -> Type}
  {f g : forall q, P q} (h : f = g) (i : B -> Q)
  : ap (fun (k : forall q, P q) => k oD i) h
    = path_forall (f oD i) (g oD i) (apD10 h oD i)
  := ap_lambdaD _ _.

(* For Types/Arrow.v: *)
(** The action of pre/post-composition on a path between functions.  See also [ap10_ap_precompose] and [ap10_ap_postcompose] in PathGroupoids.v and [ap_precomposeD] in Forall.v. *)
Definition ap_precompose {B P Q : Type}
  {f g : Q -> P} (h : f = g) (i : B -> Q)
  : ap (fun (k : Q -> P) => k o i) h
    = path_arrow (f o i) (g o i) (ap10 h o i)
  := ap_lambda _ _.

Definition ap_postcompose {B P Q : Type}
  {f g : Q -> P} (h : f = g) (j : P -> B)
  : ap (fun (k : Q -> P) => j o k) h
    = path_arrow (j o f) (j o g ) (fun q => ap j (ap10 h q)).
Proof.
  destruct h; cbn.
  symmetry; apply path_arrow_1.
Defined.
```

theories/Truncations/Core.v
```diff
@@ -170,6 +170,19 @@ Defined.
 #[export]
 Hint Immediate istruncmap_mapinO_tr : typeclass_instances.
 
+(** ** A few special things about the (-2)-truncation *)
+
+(** The type of contractible types is contractible. *)
+Definition contr_tr_minus_2@{u su | u < su} `{Univalence}
```
Alizter, line 176: I think its better to say `minus_two` since that's the actual name of the constructor for `trunc_index`.

theories/InjectiveTypes/InjectiveSigma.v
```diff
@@ -0,0 +1,179 @@
...
+  (** Here is the condition.  (It would in fact be enough to assume that [s] is section up to homotopies in its domain.) *)
+  Definition alg_flab_sigma_condition
+    := forall P f, {s : _ & alg_flab_map P f o s == idmap}.
+
+  (** A sigma type over an algebraically flabby type is also algebraically flabby, and thus algebraically injective, when the above condition holds. *)
+  Definition alg_flab_sigma (cond : alg_flab_sigma_condition)
+    : IsAlgebraicFlabbyType {x : X & A x}.
+  Proof.
+    snrapply Build_IsAlgebraicFlabbyType; intros P f.
+    all: pose (C := cond P (pr1 o f)).
```
Alizter, line 28: Another way to achieve this is to `intros P f; pose ...`.

Alizter:
Could you also update `STYLE.md` with a small paragraph explaining the `InjectiveTypes/` directory?

theories/Types/Universe.v
```diff
@@ -511,6 +524,20 @@ Proof.
     refine (equiv_induction_inv _ idpath).
 Defined.
 
+(** Any two functions that act like transport along an equivalence, i.e. maps of the type [T : forall X Y, X <~> Y -> S X -> S Y] with a computation rule of type [Trefl : forall X, (T (equiv_idmap X) == idmap)], are homotopic. This can be useful when we want to transport along an equivalence, but [univalent_transport] does not have the computational properties that we want. *)
+Definition homotopic_trequiv (S : Type -> Type) {X Y}
```
Alizter, line 528: I feel like this is in the wrong place. Maybe it should be in `Types/Equiv.v`.

doolster: perhaps it could use a better name to emphasize the connection to `univalent_transport`?

doolster:
> Could you also update `STYLE.md` with a small paragraph explaining the `InjectiveTypes/` directory?

I have "- `InjectiveTypes/*`: Files related to the notion of injective types,
  i.e. the correct notion of an injective object in a universe of types." right now. Would it be good to mention the paper in this description, or is it fine with just the references in the current headers of the files?

theories/Types/Universe.v
Alizter: OK in that case I misread it. I'm not sure about changing the name however, I think its fine as it is.

theories/InjectiveTypes/InjectiveTypes.v
```diff
@@ -0,0 +1,419 @@
...
+
+(** ** Definitions of injectivity and first examples *)
+
+(** Since the universe levels are important to many of the results here, we keep track of them explicitly as much as possible. Due to our inability to use the max and successor operations within proofs, we instead construct these universes and their associated posetal relations in the arguments of the functions. In universe declarations, we use [u], [v], [w], etc. as our typical universe variables. Our convention for the max of two universes [u] and [v] is [uv], and the successor of a universe [u] is [su]. Occasionally we write [T] for a top universe which is strictly larger than all other provided universes. We hope that later versions of Coq will allow us access to the max and successor operations and much of the cumbersome universe handing here can be greatly reduced. *)
+
+Section UniverseStructure.
+  Universes u v w uv uw vw.
+  Constraint u <= uv, v <= uv, u <= uw, w <= uw, v <= vw, w <= vw.
+
+  Definition IsInjectiveType@{uvw | uv <= uvw, uw <= uvw, vw <= uvw} (D : Type@{w})
```
Alizter, line 22: A comment with a prose explanation here would be beneficial for readers. Something like:

> (** A type [D] is said to be an injective type if for every embedding [j : X -> Y], and every function [f : X -> D], there exists a [f' : Y -> D] such that [f' o j == f]. *)

theories/InjectiveTypes/InjectiveTypes.v
```diff
@@ -0,0 +1,419 @@
...
+    := forall (X : Type@{u}) (Y : Type@{v}) (j : X -> Y) (isem : IsEmbedding@{u v uv} j)
+      (f : X -> D), merely@{uvw} (sig@{vw uw} (fun f' => f' o j == f)).
+
+  Class IsAlgebraicInjectiveType@{} (D : Type@{w}) := {
```
Alizter, line 26: Same comment about prose here. This one is especially important since to my eyes it looks very similar to the previous definition and I'm not even sure what the differences are.

Alizter (commented):
I am still going through the proofs in some more detail at the moment, but one comment I can quickly make (along with some that I made so far) is that universe variables can be line broken a bit better. When you have:
```coq
Definition foo@{a b c d .... | ....}
  (a b : A)
  : B.
```
it can make it easier to read and fit on a page if you break it like:
```coq
Definition foo
  @{a b c ....
  | a <= b ....}
  (a b : A)
  : B.
```
Wrapping the long lines of conditions accordingly.

theories/InjectiveTypes/InjectiveTypes.v
```diff
@@ -0,0 +1,419 @@
...
+  Proof.
+    snrapply Build_IsAlgebraicInjectiveType; intros X Y j isem f.
+    - exact (const (center D)).
+    - intros x.
+      exact (contr _).
+  Defined.
+
+  (** [Empty] is not algebraically injective. *)
+  Definition not_alg_inj_empty@{}
+    : IsAlgebraicInjectiveType@{} Empty -> Empty.
```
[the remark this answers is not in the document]
doolster: Writing it this way seems to cause universe problems.

Alizter:
By the way, here is a screenshot of how coqdoc's output render's locally:

image

Unfortunately, we don't have a clean way of embedding external hyperlinks in coqdoc without excessive escaping.

Maybe the heading title in InjectiveSigma could be tweaked. I would just say "Injectivity of sigma types" and then introduce a comment under saying "Here we characterize injectivity of sigma types and give some examples."

You can build the documentation locally by doing `dune build @doc` and then viewing `_build/default/theories/HoTT.html/HoTT.InjectiveTypes.InjectiveSigma.html`. The CSS is in that directory too so the browser should pick it up.

Just have a look through, and make sure the documentation renders as you expect. From a quick skim it all looks fine apart from the minor comments I made above.

theories/InjectiveTypes/InjectiveTypes.v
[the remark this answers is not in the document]
Alizter: That's annoying. I see why this is now. `not` takes a single universe, that of `A : Type` and produces `A -> Empty` which lives in that same universe. Now `IsAlgebraicallyInjectiveType` has universe max of `u v w` but it isn't explicitly mentioned so `not` cannot "find" it. Again another thing that will be fixed once algebraic unvierses land in Coq.

jdchristensen:
> I have "- `InjectiveTypes/*`: Files related to the notion of injective types,

I thought we were going to move this folder into the Homotopy folder (or maybe the Misc folder)?  I don't think every paper we formalize should result in a new top-level folder.

Alizter:
Oh yes, I forgot about that. In that case the changes in style are unneeded.

Alizter (approved):

Alizter:
I've managed to go through everything now, and it all looks good to me. This is some really high quality work, so thank you for this contribution. I'll let Dan decide when to merge.

theories/Homotopy/InjectiveTypes/TypeFamKanExt.v
```diff
@@ -171,7 +171,7 @@ Proof.
     lhs nrapply (ap_compose (fun k : (P <| j) >=> R => k (j x)) (fun ka => ka o i)).
     (* [ap (fun k => k (j x))] is exactly [apD10], so it cancels the first [path_forall]. *)
     lhs nrefine (ap _ (apD10_path_forall _ _ _ _)).
-    lhs rapply (ap_precompose _ i).
+    lhs rapply (ap_precomposeD _ i).
```
jdchristensen, line 174: I think you want to use the version without the `D`.

theories/Types/Forall.v
```diff
@@ -139,22 +139,13 @@ Proof.
...
 Defined.
 
-(** The action of pre/post-composition on a path between functions.  See also [ap10_ap_precompose] and [ap10_ap_postcompose] in PathGroupoids.v. *)
-Definition ap_precompose {B P Q : Type}
+(** The action of pre-composition on a path between dependent functions.  See also [ap10_ap_precompose] in PathGroupoids.v and [ap_precompose] in Arrow.v. *)
+Definition ap_precomposeD {B P Q : Type}
   {f g : Q -> P} (h : f = g) (i : B -> Q)
   : ap (fun (k : Q -> P) => k o i) h
     = path_forall (f o i) (g o i) (ap10 h o i)
   := ap_lambdaD _ _.
```
jdchristensen, line 148: You didn't update the definition to the dependent version.

doolster: Whoops, completely missed that.

jdchristensen (approved):

jdchristensen:
Thanks, Zack!  Nice work!

Issue #2155, opened 2024-12-15
Files in theories/ folder


jdchristensen:
I think we should have fewer files in the theories/ folder.  Let's discuss options for reorganizing things here.

These files correspond to subfolders and import everything in the corresponding subfolder.  I think it is useful to keep them.
```
Basics.v
Categories.v
HoTT.v
Pointed.v
Types.v
Truncations.v
WildCat.v
```

I'm not sure about these two.  Maybe it's fine to keep them.
```
Utf8Minimal.v
Utf8.v
```

This file is empty and should be deleted (even if we make a Misc folder):
```
Misc.v
```

Even though we have a Tactics folder, this file does not index it and is independent.  I suggest we rename it to `Tactics/Core.v`.  I also suggest that we remove the `Require Export Tactics.BinderApply.` line, since that seems independent.  But maybe we should add `Require Export Basics.Tactics.`?
```
Tactics.v
```

I think this file should move to Axioms/.
```
ExcludedMiddle.v
```

This file should move to Homotopy/.
```
NullHomotopy.v
```

Here are the files that require more discussion:
```
BoundedSearch.v
Constant.v
EquivGroupoids.v
Extensions.v
Factorization.v
Functorish.v
HFiber.v
Idempotents.v
Projective.v
```

Functorish.v is unused, but I now see that it overlaps a bit with work in #2153 (the changes to Types/Universe.v).  I think we could just delete this file.  The only question is whether to keep the Class Functorish and use it with the changes in #2153.

EquivGroupoids.v could be appended to Types/Equiv.v.  OTOH, it isn't used anywhere in the whole library.  OTOH, it takes ~0.1s to build, so would not be a big deal to include as part of Equiv.v.

Many of the others could go into the Homotopy/ folder, especially if we regard that as "things developed in HoTT using the foundations in the rest of Coq-HoTT".  The one that fits the least well is BoundedSearch.v.  But there is new work coming on "searchable types" and the related "compact types" that will probably go in Homotopy/, so maybe it's fine to also put it in Homotopy/?  An alternative is a Misc/ folder that could contain a few things.

The work in #2153 on injective types also needs a home, but I think Homotopy/ would be fine for it as well.

Alizter:
> I'm not sure about these two. Maybe it's fine to keep them.
> 
> ```
> Utf8Minimal.v
> Utf8.v
> ```

[...]

PR #2156, opened 2024-12-16
cleanup: remove Utf8.v

Alizter:
It was unused and probably not the best practice. We use UTF8 sparingly throughout the library and as long as we don't over do it, it seems to be fine. I don't believe anybody external is relying on this file, but if they are we can always add it back in on request.

Part of the cleanup suggested in: #2155


Alizter:
OK let's do that later.

PR #2161, opened 2024-12-22
functoriality lemmas for sum

Alizter:
These were missing and are straightforward to prove.

### DIFF
--- a/theories/Types/Sum.v
+++ b/theories/Types/Sum.v
@@ -435,6 +435,19 @@ Proof.
   - exact (ap inr (q b)).
 Defined.
 
+Definition functor_sum_compose {A A' A'' B B' B'' : Type}
+  {f : A -> A'} {f' : A' -> A''} {g : B -> B'} {g' : B' -> B''}
+  : functor_sum (f' o f) (g' o g) == functor_sum f' g' o functor_sum f g.
+Proof.
+  intros [a|b]; reflexivity.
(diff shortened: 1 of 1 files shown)


theories/Types/Sum.v
```diff
@@ -435,6 +435,19 @@ Proof.
...
+  {f : A -> A'} {f' : A' -> A''} {g : B -> B'} {g' : B' -> B''}
+  : functor_sum (f' o f) (g' o g) == functor_sum f' g' o functor_sum f g.
+Proof.
+  intros [a|b]; reflexivity.
+Defined.
+
+Definition functor_sum_idmap {A B : Type}
+  : functor_sum (A:=A) (B:=B) idmap idmap == idmap.
+Proof.
+  intros [|]; reflexivity.
```
jdchristensen, line 448: I think we should use the same style for both `intros` commands, so it doesn't look like something different needs to be done in the two cases.

jdchristensen (approved):

theories/Types/Sum.v
Alizter:
```suggestion
  intros [a|b]; reflexivity.
Defined.

Definition functor_sum_idmap {A B : Type}
  : functor_sum (A:=A) (B:=B) idmap idmap == idmap.
Proof.
  intros [a|b]; reflexivity.
```

PR #2163, opened 2024-12-22
add functoriality lemmas for pushout

Alizter:
I've proven these lemmas many times before, but never managed to push them. Here is a fresh attempt.

### DIFF
--- a/theories/Colimits/Pushout.v
+++ b/theories/Colimits/Pushout.v
@@ -168,6 +168,27 @@ Proof.
     apply ap, q.
 Defined.
 
+Definition functor_pushout_beta_pglue
+  {A B C} {f : A -> B} {g : A -> C}
+  {A' B' C'} {f' : A' -> B'} {g' : A' -> C'}
+  {h : A -> A'} {k : B -> B'} {l : C -> C'}
+  {p : k o f == f' o h} {q : l o g == g' o h}
+  (a : A)
+  : ap (functor_pushout h k l p q) (pglue a)
+    = ap pushl (p a) @ pglue (h a) @ ap pushr (q a)^.
+Proof.
+  lhs nrapply functor_coeq_beta_cglue.
+  symmetry.
+  snrapply ap011.
+  - apply whiskerR.
+    unfold pushl.
+    nrapply ap_compose.
+  - unfold pushr.
+    lhs nrapply ap_compose.
+    apply ap.
+    nrapply ap_V.
+Defined.
+
 Lemma functor_pushout_homotopic
   {A B C : Type} {f : A -> B} {g : A -> C}
   {A' B' C' : Type} {f' : A' -> B'} {g' : A' -> C'}
@@ -189,6 +210,44 @@ Proof.
   exact (j b).
 Defined.
 
+Definition functor_pushout_idmap {A B C : Type} {f : A -> B} {g : A -> C}
+  : functor_pushout (f:=f) (g:=g) idmap idmap idmap (fun _ => 1) (fun _ => 1) == idmap.
+Proof.
+  snrapply Pushout_ind.
+  1,2: reflexivity.
+  simpl.
+  intros a.
+  nrapply transport_paths_Flr'.
+  nrapply moveR_pM.
+  nrapply functor_pushout_beta_pglue.
+Defined.
+
+Definition functor_pushout_compose
+  {A B C f g A' B' C' f' g' A'' B'' C'' f'' g''}
+  (u : A -> A') (v : B -> B') (w : C -> C')
+  (u' : A' -> A'') (v' : B' -> B'') (w' : C' -> C'')
+  (p : v o f == f' o u) (q : w o g == g' o u)
+  (p' : v' o f' == f'' o u') (q' : w' o g' == g'' o u')
+  : functor_pushout (u' o u) (v' o v) (w' o w)
+      (fun x => ap v' (p x) @ p' (u x)) (fun x => ap w' (q x) @ q' (u x))
+    == functor_pushout u' v' w' p' q'
+      o functor_pushout u v w p q.
+Proof.
+  intros a.
+  rhs_V nrapply functor_coeq_compose.
+  snrapply functor_coeq_homotopy.
+  1: reflexivity.
+  1: apply functor_sum_compose.
+  1,2: intros x; simpl.
+  1,2: apply equiv_1p_q1.
+  1,2: lhs_V nrapply whiskerR.
+  1,3: nrapply ap_compose.
+  1,2: simpl.
+  1,2: lhs nrapply whiskerR.
+  1,3: nrapply ap_compose.
+  1,2: symmetry.
+  1,2: apply ap_pp.
+Defined.
 
 (** ** Equivalences *)
 


theories/Colimits/Pushout.v
```diff
@@ -168,6 +168,27 @@ Proof.
...
 Defined.
 
+Definition functor_pushout_beta_pglue
+  {A B C} {f : A -> B} {g : A -> C}
+  {A' B' C'} {f' : A' -> B'} {g' : A' -> C'}
+  {h : A -> A'} {k : B -> B'} {l : C -> C'}
+  {p : k o f == f' o h} {q : l o g == g' o h}
+  (a : A)
+  : ap (functor_pushout h k l p q) (pglue a)
+    = ap pushl (p a) @ (pglue (h a)) @ ap pushr (q a)^.
```
jdchristensen, line 178:
```suggestion
    = ap pushl (p a) @ pglue (h a) @ ap pushr (q a)^.
```

theories/Colimits/Pushout.v
```diff
@@ -189,6 +210,43 @@ Proof.
...
+
+Definition functor_pushout_compose
+  {A B C f g A' B' C' f' g' A'' B'' C'' f'' g''}
+  (u : A -> A') (v : B -> B') (w : C -> C')
+  (u' : A' -> A'') (v' : B' -> B'') (w' : C' -> C'')
+  (p : v o f == f' o u) (q : w o g == g' o u)
+  (p' : v' o f' == f'' o u') (q' : w' o g' == g'' o u')
+  : functor_pushout (u' o u) (v' o v) (w' o w)
+      (fun x => ap v' (p x) @ p' (u x)) (fun x => ap w' (q x) @ q' (u x))
+    == (functor_pushout u' v' w' p' q') o (functor_pushout u v w p q).
```
jdchristensen, line 233:
```suggestion
    == functor_pushout u' v' w' p' q' o functor_pushout u v w p q.
```

jdchristensen (commented):
I wonder if these would be easier to prove if functor_pushout was defined using Pushout_rec?  For example, the proof of functor_coeq_beta_cglue is a one-liner, so maybe the same would be true here?

theories/Colimits/Pushout.v
Alizter: I kept this for clarity, but I can remove if you don't think it is necessary.

Alizter:
> I wonder if these would be easier to prove if functor_pushout was defined using Pushout_rec? For example, the proof of functor_coeq_beta_cglue is a one-liner, so maybe the same would be true here?

They are a little harder to prove using `Pushout_rec` directly. Here is a proof I had laying around from before:

```coq
+  Definition functor_pushout_compose
+    {A B C f g A' B' C' f' g' A'' B'' C'' f'' g''}
+    (u : A -> A') (v : B -> B') (w : C -> C')
+    (u' : A' -> A'') (v' : B' -> B'') (w' : C' -> C'')
+    (p : v o f == f' o u) (q : w o g == g' o u)
+    (p' : v' o f' == f'' o u') (q' : w' o g' == g'' o u')
+    : functor_pushout (u' o u) (v' o v) (w' o w)
+        (comm_square_comp' p p') (comm_square_comp' q q')
+      == (functor_pushout u' v' w' p' q') o (functor_pushout u v w p q).
+  Proof.
+    serapply Pushout_ind_dp.
+    1,2: reflexivity.
+    intro a; cbn.
+    apply sq_dp^-1, sq_1G.
+    symmetry.
+    refine (ap_compose (functor_pushout u v w p q) _ (pglue a) @ _).
+    rewrite Pushout_rec_beta_pglue.
+    rewrite 2 ap_pp.
+    unfold comm_square_comp'.
+    rewrite <- 2 ap_compose.
+    rewrite 2 Pushout_rec_beta_pglue.
+    rewrite ?ap_V.
+    rewrite ?ap_pp.
+    rewrite <- ?ap_compose.
+    unfold functor_pushout.
+    by rewrite ?concat_pp_p, inv_pp.
+  Defined.
```

You can ignore the `dp` stuff, but you can see but there are a lot of `ap_compose` and `ap_pp`.

theories/Colimits/Pushout.v
jdchristensen: When I see the parentheses, I think that there must be some unusual grouping going on, so I prefer it without.  But I can see why some people might prefer the parentheses.

Alizter: I think I will add a line break and remove the parentheses then.

Alizter:
```suggestion
    == functor_pushout u' v' w' p' q'
      o functor_pushout u v w p q.
```

jdchristensen:
Well, that looks similar, since the current proof does things 2 steps at a time.  (Also note that the Coeq result is Qed.)  But I think it would improve if functor_pushout was defined directly using Pushout_rec, as shown below.  And note that functor_pushout_beta_pglue becomes a one-liner:
```coq
Definition functor_pushout
  {A B C} {f : A -> B} {g : A -> C}
  {A' B' C'} {f' : A' -> B'} {g' : A' -> C'}
  (h : A -> A') (k : B -> B') (l : C -> C')
  (p : k o f == f' o h) (q : l o g == g' o h)
  : Pushout f g -> Pushout f' g'.
Proof.
  snrapply Pushout_rec.
  - exact (pushl o k).
  - exact (pushr o l).
  - intro a.
Defined.

Definition functor_pushout_beta_pglue
  {A B C} {f : A -> B} {g : A -> C}
  {A' B' C'} {f' : A' -> B'} {g' : A' -> C'}
  {h : A -> A'} {k : B -> B'} {l : C -> C'}
  {p : k o f == f' o h} {q : l o g == g' o h}
  (a : A)
  : ap (functor_pushout h k l p q) (pglue a)
    = ap pushl (p a) @ pglue (h a) @ ap pushr (q a)^
  := Pushout_rec_beta_pglue _ _ _ _ _.
```
So I think this is another case where trying to reduce to Coeq rather than using the custom recursion and induction principles is going to make things a bit harder.

Alizter:
I've managed to change `functor_pushout` to use `Pushout_rec` instead. This was tricky to do but still possible, I had to change some place in `ReflectiveSubuniverses` that required that specific definition by introducing an appropriate lemma.

Alizter:
Actually, I didn't manage to fix everything whoops.

jdchristensen:
Oh, I had no idea that various things depended on defining functor_pushout in terms of functor_coeq.  Do you think things were better off before you implemented my suggestion or as they are now?  Even the new functoriality results were a fair bit of work.

jdchristensen:
In particular, I'm surprised by how complicated functor_pushout_homotopic is.  I wonder if some lemmas would help with a lot of the repeated things.  For example, in [this PR](https://github.com/jdchristensen/HoTT/pull/33) we added Colimit_rec_homotopy and Colimit_rec_homotopy', which simplified arguments about functor_Colimit.

jdchristensen:
@Alizter Any thoughts on my two previous comments?

Alizter:
Sorry I seem to recall writing a reply before, but it looks like I never sent it. I think we should revert to before the `functor_pushout` change since proving it from `pushout_rec` didn't really have any advantage.

It might be possible to write down some lemmas, but it isn't obvious to me if that would improve anything.

jdchristensen:
Ok, feel free to revert to before my suggestion (maybe by force pushing to keep the history clean).

About the additional lemmas, Colimit_rec_homotopy and Colimit_rec_homotopy' allow you to avoid the idiom where you: "apply *_rec, apply transport_paths_*, apply *beta*glue" by encapsulating them in a lemma.  I see a lot of those steps in the current version that would disappear, but I don't recall if they also happened in the previous version.

(It would be nice if all of our specific colimit files had the same pattern of results and helper lemmas...)

jdchristensen (approved):

PR #2164, opened 2024-12-22
mapping cylinder of map into contr type is contr + require cleanup

Alizter:
Add a lemma about contractibility of mapping cylinder and cleanup requires.

### DIFF
--- a/theories/Colimits/MappingCylinder.v
+++ b/theories/Colimits/MappingCylinder.v
@@ -1,7 +1,8 @@
 (** * Mapping Cylinders *)
 
-Require Import HoTT.Basics Cubical.DPath Cubical.PathSquare.
-Require Import Colimits.Pushout.
+Require Import Basics.Overture Basics.Tactics Basics.Equivalences
+  Basics.PathGroupoids Basics.Contractible Types.Paths
+  Cubical.DPath Cubical.PathSquare Colimits.Pushout.
 Local Open Scope path_scope.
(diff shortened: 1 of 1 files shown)


jdchristensen:
This seems like a specialised result.  Any time we define an equivalence `A <~> B` we could add an instance saying that if `A` is contractible then so is `B`.  Is there some reason this particular case is useful?

Alizter:
I had some work on mapping cones and I needed to show something that was such a cylinder is contractible:
```coq
(** The cofiber object of the cofibration map is equivalent to the suspension of the source map. *)
Definition pcofiber2_psusp {X Y : pType} (f : X ->* Y)
  : pcofiber (pcofib f) <~>* psusp X.
Proof.
  snrapply Build_pEquiv'.
  { nrefine (_ oE pushout_sym).
    nrefine (_ oE equiv_pushout_assoc _ _ _ _).
    snrapply (equiv_pushout equiv_idmap equiv_idmap).
    - change (Pushout idmap (@pconst ?X ?Y)) with (Cyl (@pconst X Y)).
      rapply equiv_contr_unit.
    - reflexivity.
    - reflexivity. }
  exact (merid pt)^.
Defined.
```

I can put this result in a PR when I have finished fleshing out that work on cofibers. I was hoping to finish Puppe sequences at some point. I know you had some work on cofibers but I didn't get round to asking about that yet.

PR #2166, opened 2024-12-22
add transport_paths_FFr and remove unused arg in transport_paths_FFlFFr

Alizter:


theories/Types/Paths.v
```diff
@@ -53,6 +53,13 @@ Proof.
   destruct p. symmetry; apply concat_p1.
 Defined.
 
+Definition transport_paths_FFr {A B C : Type} {f : A -> B} {g : B -> C}
```
jdchristensen, line 56:
```suggestion
Definition transport_paths_FFl {A B C : Type} {f : A -> B} {g : B -> C}
```

jdchristensen (approved):

theories/Types/Paths.v
Alizter: oops fixed

PR #2169, opened 2024-12-30
forall y, Decidable (x = y) is an hprop

jdchristensen:
I noticed that one can strengthen
```coq
IsHProp(forall x y, Decidable (x = y))
```
to
```coq
forall x, IsHProp(forall y, Decidable (x = y))
```
using a cute argument involving restricting to a component.  Not sure if this is useful, but since it's slightly non-trivial, I'm pushing it.

The weaker result is `ishprop_decpaths` in Basics/Decidable.v, but because of dependencies, the new result can't go there.  I originally used `merely A` in the proof, but realized that it can be avoided by using `not (not A)` in this case, which reduces the dependencies.  I put this in `HProp.v`, since that file had similar dependencies and already had one result involving decidable types and being an h-proposition.

I also added in the fact that for decidable `A`, `~~A` has the universal property of propositional truncation.  This isn't needed to prove `ishprop_decpaths'` (the stronger result).  The only thing used in that proof is `not_not_unit : A -> ~~A`, which could be inlined instead.  But having this recorded is probably worthwhile, and makes the proof of `ishprop_decpaths'` more conceptually clear.

While thinking about this, I noticed that some of the comments in `TruncImpliesFunext.v` could be improved.

### DIFF
--- a/theories/Modalities/Notnot.v
+++ b/theories/Modalities/Notnot.v
@@ -3,24 +3,22 @@ Require Import Modality.
 
 Local Open Scope path_scope.
 
-
 (** * The double negation modality *)
 
-(** This is Exercise 7.12 in the book.  Note that it is (apparently) *not* accessible unless we assume propositional resizing. *)
-
+(** The operation sending [X] to [~~X] is a modality.  This is Exercise 7.12 in the book.  Note that it is (apparently) *not* accessible unless we assume propositional resizing. *)
 Definition NotNot `{Funext} : Modality.
 Proof.
   snrapply easy_modality.
-  - intros X; exact (~ (~ X)).
-  - intros T x nx; exact (nx x).
-  - intros A B f z nBz.
+  - intros X; exact (~~X).
+  - exact @not_not_unit.
+  - cbn beta; intros A B f z nBz.
     apply z; intros a.
     exact (f a (transport (fun x => ~ (B x))
                           (path_ishprop _ _)
                           nBz)).
-  - intros A B f a.
+  - cbn beta; intros A B f a.
     apply path_ishprop.
-  - intros A z z'.
+  - cbn beta; intros A z z'.
     refine (isequiv_iff_hprop _ _).
     intros; apply path_ishprop.
 Defined.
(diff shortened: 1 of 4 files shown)


Alizter:
Are you aware of the material in `Modalities/Notnot.v`? It seems there is some overlap with your first commit.

jdchristensen:
> Are you aware of the material in `Modalities/Notnot.v`? It seems there is some overlap with your first commit.

Ah, I'll take a look tomorrow, and will likely just remove that first commit and inline not_not_unit.  Thanks!

jdchristensen:
Thanks for pointing out NotNot.v.  There isn't any direct overlap, since that is about the double-negation modality, which is a reflection onto the stable propositions.  What I added is the (easy) observation that for `A` decidable, `~~A` is the reflection of `A` into *all* propositions, i.e. it is the propositional truncation.  In general, `~~A` and the propositional truncation differ.  This means that when `A` is decidable, we have a way to work with the propositional truncation without HITs or universe issues.

To make this more clear, I decided to characterize the local types for the NotNot modality.  I also realized that my observation about `~~A` being the propositional truncation works more generally for stable types instead of decidable types.

theories/Modalities/Notnot.v
```diff
@@ -22,3 +22,8 @@ Proof.
     refine (isequiv_iff_hprop _ _).
     intros; apply path_ishprop.
 Defined.
+
+(** By definition, the local types for this modality are the types such that [not_not_unit : P -> ~~P] is an equivalence.  These are exactly the stable propositions, i.e. those propositions [P] such that [~~P -> P]. *)
+Definition inO_notnot `{Funext} (P : Type)
```
Alizter, line 27: If you add
```coq
Global Instance ishprop_stable `{Funext} P `{IsHProp P} : IsHProp (Stable P).
Proof.
  unfold Stable; exact _.
Defined.
```
to `Decidable.v` you can upgrade this to an equivalence. Unfortunately you have to give the section explicitly since `IsHProp P` is part of the data.

```coq
Definition inO_notnot `{Funext} (P : Type)
  : In NotNot P <~> (Stable P * IsHProp P).
Proof.
  pose (e := stable_hprop_iff_isequiv_not_not_unit P).
  snrapply (equiv_adjointify (fst e) (snd e)).
  2: intro; rapply path_ishprop.
  intros [s h]; rapply path_ishprop.
Defined.
```

Alizter (approved):
Looks good to me thanks! I left a suggestion but you don't have to act on it. Feel free to merge.

theories/Modalities/Notnot.v
jdchristensen: Thanks, I did something a bit more general, and also cleaned up a few comments.  I'll let you take another quick look before merging.

PR #2170, opened 2024-12-30
Functoriality results for Colimits

ThomatoTomato:
In this PR we add more results pertaining to `functor_colimit`. The library currently supports two approaches to use colimits. One either considers *concrete colimits* that are the HITs defined from the data of the diagrams, or *abstract colimits* that are arbitrary types with the correct universal property. As per the implementation of today, a concrete colimit is an abstract colimit once function extensionality is assumed. 

This PR adds a sort of third approach, where we allow ourselves to use both concrete and abstract colimits at the same time. The map `functor_colimit` has a diagram map `m : D1 => D2` as input, and outputs a map `Colimit D1 -> Colimit D2`. We add `functor_Colimit_half` that acts like the above function, but you can choose an arbitrary cocone `Q` over the target diagram to obtain a map `Colimit D1 -> Q`. Hopefully, this will let us leverage the computational properties of concrete colimits, while keeping some of the flexibility and freedom with abstract colimits.

The main contribution from this PR consists of studying how the recursor for concrete colimits interact with homotopic cocones. This lets us study `functor_Colimit_half`, and prove verious results pertaining to it. We also record naturality for `ap` at the function `colim`, prove some functoriality results for abstract colimits, and perform some cleanup of the existing code.

Hopefully this code will find its use in the zigzag construction for path spaces of pushouts. With some additional luck, we can also simplify some of the contents found in Sequential.v.

### DIFF
--- a/theories/Colimits/Colimit.v
+++ b/theories/Colimits/Colimit.v
@@ -69,24 +69,14 @@ Definition colimp {G : Graph} {D : Diagram G} (i j : G) (f : G i j) (x : D i)
   : colim j (D _f f x) = colim i x
   := (cglue ((i; x); j; f))^.
 
-(** Naturality of [ap] applied to [colim]. *)
-Definition concat_ap_colim {G : Graph} {D : Diagram G} {i j : G} (f : G i j) {x y : D i} (p : x = y)
-  : ap (colim j) (ap (D _f f) p) @ colimp i j f y
-    = colimp i j f x @ ap (colim i) p.
-Proof.
-  lhs_V nrapply (ap_compose _ _ _ @@ 1).
-  exact (concat_Ap (colimp i j f) p).
-Defined.
-
 (** To obtain a path [colim j (D _f f x) = colim j (D _f f y)] from a path [x = y], one can either use [ap (D _f f)] followed by [ap (colim j)], or conjugate [ap (colim i)] by [colimp i j f]. These paths are equivalent. *)
 Definition ap_colim_homotopic {G : Graph} {D : Diagram G} {i j : G} (f : G i j) {x y : D i} (p : x = y)
   : ap (colim j) (ap (D _f f) p)
     = colimp i j f x @ ap (colim i) p @ (colimp i j f y)^.
 Proof.
-  nrapply moveL_pV.
-  nrapply concat_ap_colim.
+  lhs_V nrapply ap_compose.
+  exact (ap_homotopic (colimp _ _ _) p).
 Defined.
-
 
 Definition Colimit_ind {G : Graph} {D : Diagram G} (P : Colimit D -> Type)
 (q : forall i x, P (colim i x))
@@ -299,6 +289,7 @@ Section FunctorialityColimit.
   Context `{Funext} {G : Graph}.
 
   (** Colimits are preserved by composition with a (diagram) equivalence. *)
+
   Definition iscolimit_precompose_equiv {D1 D2 : Diagram G}
     (m : D1 ~d~ D2) {Q : Type}
     : IsColimit D2 Q -> IsColimit D1 Q.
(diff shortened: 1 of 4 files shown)


theories/Colimits/Colimit.v
```diff
@@ -69,26 +69,25 @@ Definition colimp {G : Graph} {D : Diagram G} (i j : G) (f : G i j) (x : D i)
...
-  apply moveR_pM.
-  apply moveR_Vp.
-  symmetry.
-  lhs nrapply concat_p_pp.
-  apply ap_colim.
+  nrapply moveL_pV.
+  nrapply concat_ap_colim.
 Defined.
 
+
```
jdchristensen, line 90: Just one blank line.  As a general principle, we don't have multiple blank lines in a row.

theories/Colimits/Colimit.v
```diff
@@ -300,7 +299,6 @@ Section FunctorialityColimit.
   Context `{Funext} {G : Graph}.
 
   (** Colimits are preserved by composition with a (diagram) equivalence. *)
-
```
jdchristensen, line 192: Since the previous comment applies to the next two results, not just the next result, I hint at this by putting a blank line between the comment and the next result.

theories/Colimits/Colimit.v
```diff
@@ -358,20 +356,19 @@ Section FunctorialityColimit.
...
-    := @equiv_inj _ _ 
-      (cocone_postcompose HQ1) (iscolimit_unicocone HQ1 Q3) _ _ 
+    := @equiv_inj _ _
+      (cocone_postcompose HQ1) (iscolimit_unicocone HQ1 Q3) _ _
       (postcompose_functor_colimit_compose m n HQ1 HQ2 HQ3).
 
   (** ** Colimits of equivalent diagrams *)
 
   (** Now we have that two equivalent diagrams have equivalent colimits. *)
-
```
jdchristensen, line 228: Again, the previous comment is describing multiple results, so a blank line is appropriate.

jdchristensen: My remark is appearing in a strange spot in the "Files changed" tab, so to be clear I mean that there should be a blank line after:
```coq
(** Now we have that two equivalent diagrams have equivalent colimits. *)
```

theories/Colimits/Colimit.v
```diff
@@ -71,6 +69,25 @@ Definition colimp {G : Graph} {D : Diagram G} (i j : G) (f : G i j) (x : D i)
...
+Defined.
+
+(** To obtain a path [colim j (D _f f x) = colim j (D _f f y)] from a path [x = y], one can either use [ap (D _f f)] followed by [ap (colim j)], or conjugate [ap (colim i)] by [colimp i j f]. These paths are equivalent. *)
+Definition ap_colim_homotopic {G : Graph} {D : Diagram G} {i j : G} (f : G i j) {x y : D i} (p : x = y)
+  : ap (colim j) (ap (D _f f) p)
+    = colimp i j f x @ ap (colim i) p @ (colimp i j f y)^.
+Proof.
+  nrapply moveL_pV.
+  nrapply concat_ap_colim.
+Defined.
```
jdchristensen, line 88: I'm still not convinced we need both of these results, as they are essentially the same except for the position of one term.  If both forms were needed in multiple places, I could see the motivation for recording both, but at present we have only one use case (not in this PR).  I'd be inclined to keep the second one, and prove it using `ap_compose` and `ap_homotopic`.

ThomatoTomato: If I'm going to be honest, I'm not really sure if we should be keeping any of them. We might just inline the proof directly whenever we need it. In my cases I have only had the need to use the first one, but that could also be resolved using `concat_Ap`. Does @ben-connors have any other opinions or experiences on what to do with these maps?

theories/Colimits/Colimit.v
ThomatoTomato: I see. I will revert those changes then.

theories/Colimits/Colimit.v
jdchristensen: I think it's reasonable to keep one form of this result, as it might not be obvious to the reader what follows from naturality of ap.

jdchristensen (approved):

Alizter (approved):
This looks good to me, thanks!

PR #2171, opened 2024-12-31
basics on homotopy cofibers

Alizter:
Here is a start on homotopy cofibers. I have more work waiting, but it will be some time before I can clean it up.

@jdchristensen feel free to push any changes as you see fit here. There will probably be some name/style differences with whatever you have.

This builds on
- #2163


jdchristensen:
My work is in the attached file, and isn't polished.  Maybe you can merge the two together?

[Cofiber.txt](https://github.com/user-attachments/files/18303101/Cofiber.txt)

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,92 @@
+Require Import Basics.Overture Basics.Tactics Basics.PathGroupoids.
+Require Import Types.Unit Types.Paths.
+Require Import Colimits.Pushout.
+Require Import Homotopy.NullHomotopy.
+
+Set Universe Minimization ToSet.
```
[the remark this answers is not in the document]
Alizter: I'm not sure, but there are less universes when it is set. Generally, I try to minimise universes as much as possible.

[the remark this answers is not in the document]
Alizter: When it is set, `cofiber` has 3 unvierses which is expected. One for each type and one for their max.

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,92 @@
...
+
+(** * Homotopy Cofibers *)
+
+(** ** Definition *)
+
+(** Homotopy cofibers or mapping cones are spaces constructed from a map [f : X -> Y] by contracting the image of [f] inside [Y] to a point. They can be thought of as a kind of coimage or quotient. *)
+
+Definition cofiber {X Y : Type} (f : X -> Y)
```
Alizter, line 14: Should I make this upper case?

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,92 @@
...
+  := Pushout (const_tt X) f.
+
+(** *** Constructors *)
+
+(** Any element of [Y] can be included in the cofiber. *)
+Definition cofib {X Y : Type} (f : X -> Y) : Y -> cofiber f
```
Alizter, line 20: We call `pFiber`s constructor `pfib` so it seemed appropriate to choose this name.

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,92 @@
...
+  := pushr.
+
+(** We have a distinguised point in the cofiber, where the image of [f] is contracted to. *)
+Definition apex {X Y : Type} (f : X -> Y) : cofiber f
```
Alizter, line 24: Does `cf_apex` seem like a better name?

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,92 @@
...
+  := pushl tt.
+
+(** Given an element [x : X], we have the path [leg] from the [f x] to the [apex]. *)
+Definition leg {X Y : Type} (f : X -> Y) (x : X) : cofib f (f x) = apex f
```
Alizter, line 28: And here `cf_leg`?

theories/Homotopy/Cofiber.v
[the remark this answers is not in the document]
jdchristensen: And when it is unset, it adds a universe to hold `Unit`.  So using it makes sense.

theories/Homotopy/Cofiber.v
jdchristensen: I think that would match many other things, as it is a "type constructor".  But let's keep it lowercase when referred to in lemmas about it.

theories/Homotopy/Cofiber.v
jdchristensen: Seems fine.

jdchristensen: I guess `cf_in` would be an alternative, if we decide to use `cf` as a prefix for various things about cofibers.  Either way seems fine.

theories/Homotopy/Cofiber.v
jdchristensen: I think we should indicate that this has to do with cofibers somehow.  `cf_apex` or `cf_pt` or `cofiber_apex` or `cofiber_pt`.  Actually, we should probably define the cofiber as a pointed type, and then we can just use `pt` to refer to this point...

theories/Homotopy/Cofiber.v
jdchristensen: I think we should call this `cfglue`, like the other paths we have as part of various colimits:  `cglue`, `gcglue`, `pglue`, etc.  (I'm not sure why for general colimits we use `colimp`.)

theories/Homotopy/Cofiber.v
Alizter: The choice of base point is not obvious. When X and Y are pointed, then cf_in would take the base point of Y to `cf_in (point Y)` which is not definitionally `cf_apex`. So I was thinking of doing the unpointed version first, and choosing a different basepoint generally.

Alizter:
@jdchristensen Do you want me to take a stab at applying Blakers-Massey to get that connectivity result?

theories/Homotopy/Cofiber.v
jdchristensen: I'd be inclined to always choose the "apex" as the base point, even when X and Y are pointed, as `cf_in` would still be canonically pointed.  Also, `cf_in o f` has a canonical homotopy to the constant map at the "apex", whereas if you choose `f pt` as the basepoint, the homotopy becomes more complicated.

But if you prefer to leave it unpointed, then we could go with a name like `cf_apex`.

jdchristensen:
> @jdchristensen Do you want me to take a stab at applying Blakers-Massey to get that connectivity result?

Sure, that would be great!

Alizter:
I derived the 8.10.2 HoTT Book Blakers-Massey from the one we had. I'm not yet certain this is enough for the cofiber result, but this is probably something we should cleanup and open a PR for.

theories/Homotopy/Cofiber.v
```diff
@@ -100,3 +105,19 @@ Proof.
   lhs nrapply (ap (cofib f) p^).
   apply leg.
 Defined.
+
+(** The cofiber of a surjective map is [0]-connected. This will be generalized this later. *)
```
jdchristensen, line 109: The second sentence of the comment is grammatically incorrect.  And maybe should just be removed anyways.

theories/Homotopy/BlakersMassey.v
```diff
@@ -539,3 +540,76 @@ Proof.
   1: intros; apply isconnected_join.
   all: exact _.
 Defined.
+
+Definition blakers_massey_po `{Univalence} (m n : trunc_index)
+  {X Y Z : Type} (f : X -> Y) (g : X -> Z)
+  `{H1 : !IsConnMap m.+1 f} `{H2 : !IsConnMap n.+1 g}
+  : IsConnMap (m +2+ n) (pullback_corec (pglue (f:=f) (g:=g))).
```
jdchristensen, line 547: I haven't looked at the proof of this carefully, but I suspect that parts of it should be separate lemmas.

Alizter:
I've done the review suggestions and factored out some parts of the Blakers-Massey proof. It's possible there are more lemmas to factor out. I added comments explaining what is going on in the proof.

Alizter:
I managed to prove the connectivity of cofibers using the same argument as the one for suspensions. It didn't use Blakers-Massey and generalised pretty nicely. I'm not sure what we should do with the BM stuff, I struggled to apply it in this case anyway.

Alizter:
Oh nevermind about my comment about BM I just realised you wanted that to be applied to something else. I'll have a go at that now.
```coq
  (** Blakers-Massey implies that the comparison map is highly connected. *)
  Definition isconnected_fiber_to_cofiber (n m : trunc_index)
    {ac : IsConnected n.+1 A} {fc : IsConnMap m.+1 f} (b : B)
    : IsConnMap (m +2+ n) (@fiber_to_cofiber b).
  (** TODO: is m +2+ n correct? *)
  Admitted.
```

theories/Homotopy/BlakersMassey.v
```diff
@@ -539,3 +540,74 @@ Proof.
...
+    destruct p.
+    snrapply path_sigma.
+    1: exact q.
+    destruct q.
+    cbn -[functor_spushout].
+    lhs nrapply concat_1p.
+    lhs nrapply concat_p1.
+    lhs nrapply functor_coeq_beta_cglue.
+    apply moveR_pM.
+    apply concat_1p_p1. }
```
jdchristensen, line 583: I think this should be more elementary, and maybe it will be more clear once the equivalence mentioned above is in a named lemma.  You are trying to show that a `f1 o f2 o f3 == f4` for some maps, and I suspect that if you change it to `f1 o f2 == f4 o f3^-1`, where `f3` is the named lemma, it might be true definitionally, or at least, more easily.  This also is a special case of a more general fact, that if you have a map of pushout data, you get a map between the pullbacks of the pushouts which commutes with the comparison map we're trying to study.

theories/Homotopy/BlakersMassey.v
```diff
@@ -539,3 +540,74 @@ Proof.
...
+  - intros y'.
+    snrapply isconnected_equiv'.
+    3: by snrapply H1. 
+    nrefine (equiv_sigma_symm _ oE _).
+    snrapply equiv_functor_sigma_id; intros x; cbn beta.
+    nrefine (equiv_functor_sigma_id (fun _ => equiv_prod_symm _ _) oE _).
+    nrefine (equiv_functor_sigma_id (fun _ => equiv_sigma_prod0 _ _) oE _).
+    nrefine ((equiv_sigma_assoc' _ _)^-1 oE _).
+    symmetry.
+    rapply equiv_contr_sigma.
```
jdchristensen, line 612: These are very similar.  Can they be unified or mostly unified?

Also, the second half of this proof should be a separate result:  for the SPushout, the iterated sigma of the spglue map is appropriately connected.  This is an intermediate statement between the existing blakers_massey result and the one you are proving here.  Then the one here is "just" the same claim, with a few of the types changed to equivalent types.

theories/Homotopy/BlakersMassey.v
jdchristensen: Here's an idea that might work.  Suppose we have `f : X -> Y`, `g : X -> Z` and an equivalence `e : X' <~> X`.  Then you can take a pushout involving `X` or a pushout involving `X'`.  Then take the corresponding pullbacks `P` and `P'`, and you get two comparison maps, `X -> P` and `X' -> P'`.  The claim is that if the first is `k`-connected, then so is the second.  Since this is phrased as a general statement, you should be able to prove it using equivalence induction, i.e. by assuming that `e` is the identity equivalence, in which case I think it is trivial.  (It also shouldn't be too hard to prove this by showing that a square involving `X' -> X -> P` and `X' -> P' -> P` commutes.)

Then you apply this to the case where one of them is `X` and the other is the double-sigma type of `Q`, and the result you want should follow.  (You'd still need to factor out the lemma I suggested in a different comment that says that for the SPushout, the iterated sigma of the spglue map is appropriately connected.)

theories/Homotopy/BlakersMassey.v
```diff
@@ -539,3 +540,74 @@ Proof.
...
+  (** Next, we convert our [X] to the equivalent total space of [Q]. We can do this because an equivalence is connected, and we are canceling on the rigght. *)
+  snrapply (cancelR_conn_map _ (equiv_fun _)).
+  1: exact {y : Y & {z : Z & Q y z}}.
+  { unfold Q.
+    symmetry.
+    nrefine ((equiv_sigma_prod' _)^-1 oE _).
+    nrefine (equiv_sigma_symm _ oE _).
+    nrefine (equiv_functor_sigma_id (fun _ => (equiv_sigma_prod_prod _ _)^-1) oE _).
+    symmetry.
+    rapply equiv_sigma_contr. }
```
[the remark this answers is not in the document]
Alizter: I've factored it out and put it in `HFiber`. Not sure if that's the best place, but it seemed to make sense to me.

theories/Homotopy/BlakersMassey.v
Alizter: That simplifies it somewhat, but there is still a `functor_coeq_beta` around.

Alizter: I have some stuff on pushout map data sitting in an unfinished 3x3 branch that might be useful to have.

theories/Homotopy/BlakersMassey.v
Alizter: What is the point of factoring the intermediate statement out?

theories/Homotopy/BlakersMassey.v
jdchristensen: > That simplifies it somewhat, but there is still a `functor_coeq_beta` around.

Hmm, it looks like it doesn't actually simplify it, as the first line can just be replaced with a destructing intros:

```coq
  { intros [y [z [x [[] []]]]].
    snrapply (path_sigma' _ 1 (path_sigma' _ 1 _)); simpl.
    symmetry.
    lhs nrapply concat_1p.
    lhs nrapply concat_p1.
    lhs nrapply functor_coeq_beta_cglue.
    lhs nrapply concat_p1.
    nrapply concat_1p. }
```

(The only other difference is that I moved the `simpl` earlier.)

I also tried replacing your first two lines with `cate_moveL_eV`, but I couldn't get Coq to accept it.

theories/Homotopy/BlakersMassey.v
jdchristensen: Well, the statement about the total map is how Blakers-Massey is usually stated, so it seems worth having it in that form for both SPushouts and Pushouts.  But the other reason is that the proof nicely factors through that with no extra work, and two shorter proofs are easier to understand than one long one.

theories/HFiber.v
```diff
@@ -153,6 +153,18 @@ Proof.
...
+(** This is a useful variant for taking a "double fiber" of two maps. *)
+Definition equiv_double_fibration_replacement
+  {X Y Z : Type} (f : X -> Y) (g : X -> Z)
+  : X <~> {y : Y & {z : Z & {x : X & f x = y /\ g x = z}}}.
+Proof.
+  nrefine ((equiv_sigma_prod' _)^-1 oE _).
+  nrefine (equiv_sigma_symm _ oE _).
+  nrefine (equiv_functor_sigma_id (fun _ => (equiv_sigma_prod_prod _ _)^-1) oE _).
+  symmetry.
+  rapply equiv_sigma_contr.
```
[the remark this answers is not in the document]
Alizter: Nice. I was looking for this and I couldn't find the result before.

theories/Homotopy/BlakersMassey.v
Alizter: I wrote down the natural map from the total space of `Q` to the pushout product of `Q`. Then the connectivity of that is the intermediate you mention.

theories/Homotopy/BlakersMassey.v
jdchristensen: The idea in the previous comment might work, and might save us the one homotopy in the current proof, but that homotopy is pretty straightforward, so I think it's ok to leave it like it is and not try this idea.

Alizter:
I pushed a commit simplifying the proof of Freudenthal. There is no more hard-to-understand SPushout wrangling. It is now more obviously a direct corollary.

Alizter:
I pushed another that uses `make_equiv` but I'm not sure if that is worth it here.

Alizter:
I am having trouble with universe variables and I don't have too much time to debug. Most likely a per-version change is required. I managed to prove the connectedness of the comparison map, which is essentially inbetween BM and Freudenthal.

jdchristensen:
> I am having trouble with universe variables and I don't have too much time to debug. Most likely a per-version change is required.

Since the Freudenthal result is down to ~6-8 universe variables, maybe the best thing to do is to improve the real proof rather than using a wrapper?  Even if we left it with 6-8 universe variables for now, that would be fine, but it might also be possible to reduce that a bit with a few easy annotations.

I haven't looked at the recent commits yet, as I'm just starting a trip, but I like that the pushout version of Blakers-Massey turned out to be useful to simplify Freudenthal.

jdchristensen:
I was able to directly get it down to two universe variables, but I think there are still some improvements that can be made, which I'll leave as comments.

theories/Homotopy/Freudenthal.v
```diff
@@ -1,33 +1,17 @@
...
-  refine (conn_map_homotopic _ _ _ _
-                             (blakers_massey n n (fun (u v:Unit) => X) tt tt)).
-  intros x.
-  refine (_ @ (equiv_pushout_pglue
-                 (equiv_contr_sigma (fun _ : Unit * Unit => X))^-1
-                 (equiv_idmap Unit) (equiv_idmap Unit)
-                 (fun x : X => idpath) (fun x : X => idpath) x)^).
-  exact ((concat_p1 _ @ concat_1p _)^).
+  - exact (Pullback (pushl (f:=const_tt X) (g:=const_tt X)) pushr).
+  - make_equiv.
```
jdchristensen, line 15: I don't like the use of make_equiv here, because it doesn't give us control over which equivalence is produced, and the next line relies on the composition being definitionally the map that `blakers_massey_po` is about.  The goal here is a general description of the pullback of two maps from `Unit` to a type, and I think it should be a lemma in Pullback.v.

theories/Homotopy/Freudenthal.v
```diff
@@ -1,33 +1,17 @@
...
+  - rapply blakers_massey_po@{u u u u v u u u u}.
```
jdchristensen, line 16: This works, but maybe it's also easy to reduce the universe variables needed by `blakers_massey_po`?  (Not worth spending much time on if it isn't trivial.)

jdchristensen: Well, it still doesn't work with all versions, so maybe `blakers_massey_po` requires different numbers of universe variables with different versions?  Sorry, no more time to think about this.

Alizter: That's a bit annoying. I haven't been able to test this myself locally since Coq 9.0 broke the master setup I had with nix, but I don't have the time to go and fix that now. I say we ignore this for now and fix it before we undraft.

theories/Homotopy/Freudenthal.v
Alizter: I switched it back to something more explicit. It did turn out to be slower to step through since the term was big and Coq's printer was choking on it. The proof itself didn't appear to cause a slow down however. I'm not sure if it's worth a separate lemma.

theories/Homotopy/Freudenthal.v
```diff
@@ -1,33 +1,18 @@
...
-                             (blakers_massey n n (fun (u v:Unit) => X) tt tt)).
-  intros x.
-  refine (_ @ (equiv_pushout_pglue
-                 (equiv_contr_sigma (fun _ : Unit * Unit => X))^-1
-                 (equiv_idmap Unit) (equiv_idmap Unit)
-                 (fun x : X => idpath) (fun x : X => idpath) x)^).
-  exact ((concat_p1 _ @ concat_1p _)^).
+  - exact (Pullback (pushl (f:=const_tt X) (g:=const_tt X)) pushr).
+  - symmetry.
+    exact (equiv_contr_sigma _ oE equiv_contr_sigma (fun b : Unit => {c : Unit & pushl b = pushr c})).
```
jdchristensen, line 16: While this is short, it's a fundamental fact about Pullbacks, so I'd still like to have it as a lemma in the right file.  I only had a couple of minutes between flights, so all I did in my last commit is simplify the proof a bit.

theories/Homotopy/Cofiber.v
```diff
@@ -0,0 +1,158 @@
...
+Definition cofib {X Y : Type} (f : X -> Y) : Y -> Cofiber f
+  := pushr.
+
+(** We have a distinguised point in the cofiber, where the image of [f] is contracted to. *)
+Definition cf_apex {X Y : Type} (f : X -> Y) : Cofiber f
+  := pushl tt.
+
+(** Given an element [x : X], we have the path [cfglue f x] from the [f x] to the [cf_apex]. *)
+Definition cfglue {X Y : Type} (f : X -> Y) (x : X) : cofib f (f x) = cf_apex f
+  := (pglue _)^.
```
jdchristensen, line 31: The path inversion here makes later things a bit trickier.  There are two ways to fix it.  One is to change the direction of the path given in cfglue.  The other is to swap f and g in the definition of the Cofiber, so it is `Pushout f (const_tt X)` instead.  I tried the second way, and it works.  Should I push it to this PR?

Alizter: I thought I changed the direction of this already, but it turns out I stashed that change. I've already done this so let me quickly clean it up and push.

Alizter:
@jdchristensen Thanks for writing down the unit lemmas for pullbacks. Anything else about cofibers that comes to mind that you would like to see included here?

I tried to write down the equivalence between the cofiber of the suspension functor and the suspension of the cofiber but it got a bit involved. Doing it directly wasn't easy and it might be possible to do it via yoneda but the proof is getting long. I suspect it is going to be a little shorter than the trijoin proof. It is also a direct application of the 3x3 lemma for pushouts with 7 units and f coming out from the center. Interested to know if you have any ideas how to tackle this, but perhaps its best left for a later time.

jdchristensen:
> Anything else about cofibers that comes to mind that you would like to see included here?

Nothing leaps to mind right now.

> I tried to write down the equivalence between the cofiber of the suspension functor and the suspension of the cofiber but it got a bit involved.  [...] Interested to know if you have any ideas how to tackle this, but perhaps its best left for a later time.

I haven't thought about it and don't have any good ideas.  As you suggest, it should follow from Yoneda and the dual fact about fibre sequences.  But maybe the best approach long term is to get the 3x3 lemma done.  (But of course not in this PR.)

I think this PR could be merged now.  About the universe variables, I think the simplest is to just remove the annotations from the current version, and let Coq generate however many universes it wants.  It won't be too large, since Blakers-Massey just needs around 9.  Then, when we are at a point where the Coq versions we support behave the same way, we could look into simplifying the universe variables in Blakers-Massey.

Alizter:
I've removed the universe annotations from `Freudenthal.v`. Should we just merge this with `BlakersMassey.v` at this point?

jdchristensen:
> Should we just merge this with `BlakersMassey.v` at this point?

Good point.  Yes, that makes sense to me.

jdchristensen (approved):

PR #2172, opened 2025-01-01
category of group presentations

Alizter:
Here is some material on group presentations. We show that group presentations with structure preserving morphisms form a category and that the group presented by a presentation construction becomes a functor. We give some related properties about the standard presentation including all finite groups being finitely presented.

This is in preparation for some WIP on Tietze equivalences which should be a factorisation theorem for presentation equivalences between finitely presented presentations.

Working at the level of morphisms of presentation is much cleaner than having to go through `group_gp` since you only have to deal with free groups.

### DIFF
--- a/theories/Algebra/Groups/Presentation.v
+++ b/theories/Algebra/Groups/Presentation.v
@@ -6,11 +6,10 @@ Require Import Algebra.Groups.GroupCoeq.
 Require Import Spaces.Finite Spaces.List.Core.
 Require Import WildCat.
 
-
 Local Open Scope mc_scope.
 Local Open Scope mc_mult_scope.
 
-(** In this file we develop presentations of groups. *)
+(** * Presentations of groups *)
 
 (** The data of a group presentation *)
 Record GroupPresentation := {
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Presentation.v
```diff
@@ -22,39 +21,327 @@ Record GroupPresentation := {
...
+    rhs nrapply (ap _ (fmap_group_gp_beta_in _ _)).
+    rhs nrapply fmap_group_gp_beta_in_word.
+    nrapply fmap_group_gp_beta_in.
+Defined.
+
+(** ** Standard presentation *)
+
+(** Every group can be given a standard presentation. The set of generators are
+  the elements of the groups and the relations are such that the group operation
+  agrees with word concatenation. *)
```
jdchristensen, line 255:
```suggestion
(** Every group can be given a standard presentation. The generators are the elements of the group and the relations enforce that the group operation agrees with word concatenation. *)
```

theories/Algebra/Groups/Presentation.v
```diff
@@ -22,39 +21,327 @@ Record GroupPresentation := {
...
+Proof.
+  intros G.
+  snrapply Build_GroupPresentation.
+  - exact G.
+  - exact (G * G)%type.
+  - intros [g h].
+    exact ((freegroup_in (g * h))^ * freegroup_in g * freegroup_in h).
+Defined.
+
+(** Given a group, we have a homomorphism to the group presented by the standard presentation. *)
```
jdchristensen, line 266: I think the map more naturally goes in the other direction.

theories/Algebra/Groups/Presentation.v
```diff
@@ -22,39 +21,327 @@ Record GroupPresentation := {
   gp_relators : gp_rel_index -> FreeGroup gp_generators;
 }.
 
-(** Note: A relator is a relation in the form of "w = 1", any relation "w = v" can become a relator "wv^-1 = 1" for words v and w. *)
+(** Note: A relator is a relation in the form of [w = 1], any relation [w = v] can become a relator [wv^-1 = 1] for words [v] and [w]. *)
+
+(** ** Category of group presentations *)
+
+(** Group presentations together with their morphisms form a category. We can later show that this category is equivalent to the category of groups. *)
```
jdchristensen, line 28: Below you say something that suggests that this true in every topos, so I'm confused.

Also, some more comment about the definition of GroupPresentationMorphism would be good.  The first two entries are the same as giving homomorphisms `FreeGroup (gq_generators p) $-> FreeGroup (gq_generators q)` and `FreeGroup (gq_rel_index p) $-> FreeGroup (gq_rel_index q)`, and the third entry says that these commute with the natural maps from latter groups to the former groups.

jdchristensen (commented):
I know this isn't ready for review, but I took a quick look.

Alizter:
There are some things I need to work out with this. I thought they would be easy to resolve but it looks like this PR is not ready. I don't have too much time to work on it any longer. I will come back to it at some point.

PR #2178, opened 2025-01-03
conjugation of group elements

Alizter:
Here are the basics on conjugation in groups.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -1146,3 +1146,69 @@ Proof.
   lhs nrapply grp_homo_op.
   apply ap, grp_homo_inv.
 Defined.
+
+(** ** Conjugation *)
+
+(** Conjugation by a group element is a homomorphism. Often we need to use properties about group homomorphisms in order to prove things about conjugation, so it is helpful to define it directly as a group homomorphism. *)
+Definition grp_conj {G : Group} (x : G) : G $-> G.
+Proof.
+  snrapply Build_GroupHomomorphism.
+  - exact (fun y => x * y * x^).
+  - intros y z.
+    rhs nrapply grp_assoc.
+    apply (ap (.* x^)).
+    rhs nrapply grp_assoc.
+    lhs nrapply grp_assoc.
+    apply (ap (.* z)).
+    symmetry; apply grp_inv_gV_g.
+Defined.
+
+(** Conjugation by the unit element is the identity. *)
+Definition grp_conj_unit {G : Group} : grp_conj (G:=G) 1 $== Id _.
+Proof.
+  intros x.
+  apply grp_moveR_gV.
+  rhs nrapply grp_unit_r.
+  nrapply grp_unit_l.
+Defined.
+
+(** Conjugation commutes with group homomorphisms. *)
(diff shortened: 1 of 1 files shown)


Alizter:
I chose the convention `y * x * y^` since that is how we conjugate elsewhere in the library already. I am however starting to think the other `y^ * x * y` might be better since it is more commonly used in group theory. Not sure.

Alizter:
Perhaps the best way forward is to have both and use them whenever they are needed. Any result about one will become a result about the other in the opposite group.

theories/Algebra/Groups/Group.v
```diff
@@ -1125,3 +1125,67 @@ Proof.
...
+    nrapply ap.
+    lhs_V nrapply grp_assoc.
+    nrapply ap.
+    rhs nrapply grp_assoc.
+    nrapply (ap (.* x^)).
+    rhs nrapply grp_assoc.
+    lhs_V nrapply grp_unit_l.
+    nrapply (ap (.* z)).
+    symmetry.
+    apply grp_inv_l.
```
jdchristensen, line 1148: Here's a shorter proof, using `grp_inv_gV_g` from another PR.
```suggestion
    rhs nrapply grp_assoc.
    apply (ap (.* x^)).
    rhs nrapply grp_assoc.
    lhs nrapply grp_assoc.
    apply (ap (.* z)).
    symmetry; apply grp_inv_gV_g.
```

jdchristensen (approved):

jdchristensen:
> I chose the convention `y * x * y^` since that is how we conjugate elsewhere in the library already. I am however starting to think the other `y^ * x * y` might be better since it is more commonly used in group theory. Not sure.

I think `y * x * y^` is fine.  It has the advantage that `grp_conj` is a homomorphism from the group to the group of self-equivalences rather than an anti-homomorphism.

> Perhaps the best way forward is to have both and use them whenever they are needed. Any result about one will become a result about the other in the opposite group.

In the opposite group, it'll turn out to be `y^ * (x * y)`, which is different from `y^ * x * y`.  So there are really four choices here.

theories/Algebra/Groups/Group.v
Alizter: What about doing these with `rewrite`?

theories/Algebra/Groups/Group.v
```diff
@@ -1146,3 +1146,69 @@ Proof.
...
+  nrapply grp_unit_l.
+Defined.
+
+(** Conjugation commutes with group homomorphisms. *)
+Definition grp_homo_conj {G H : Group} (f : G $-> H) (x : G)
+  : f $o grp_conj x $== grp_conj (f x) $o f.
+Proof.
+  intros z; simpl.
+  by rewrite !grp_homo_op, grp_homo_inv.
+Defined.
```
Alizter, line 1181: I added this lemma when rebasing.

theories/Algebra/Groups/Group.v
jdchristensen: That wouldn't really save much.  The two `ap`s are pretty simple.

jdchristensen:
This LGTM.

PR #2180, opened 2025-01-03
Improve mapping out of a generated subgroup

Alizter:
I noticed that `functor_subgroup_generated` can be proven using a simpler result where the target is any subgroup rather than another generated one.

This isn't the most general way to map out of a generated subgroup however. I have some up coming work on commutators where I need a different eliminator. In that case `f` being a homomoprhism is weakened to a "homomorphism upto conjugation of one factor" and the target subgroup being normal takes care of the conjugation. This doesn't fall into the way of eliminating subgroups as stated here but we can discuss that when its ready.


theories/Algebra/Groups/Subgroup.v
```diff
@@ -557,20 +557,28 @@ Definition subgroup_generated_gen_incl {G : Group} {X : G -> Type} (g : G) (H :
   : subgroup_generated X
   := (g; tr (sgt_in H)).
 
+Definition subgroup_generated_rec {G H : Group} (X : G -> Type) (S : Subgroup H)
+  (f : G $-> H) (i : forall x, X x -> S (f x))
+  : forall x, subgroup_generated X x -> S (f x).
+Proof.
+  intros x; rapply Trunc_rec; intros p.
```
jdchristensen, line 564: I think you should change `x` to `g` in these three lines.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -557,20 +557,28 @@ Definition subgroup_generated_gen_incl {G : Group} {X : G -> Type} (g : G) (H :
...
-  apply Trunc_functor.
-  intro p.
-  induction p as [g i | | g h p1 IHp1 p2 IHp2].
-  - apply sgt_in, preserves, i.
-  - rewrite grp_homo_unit.
-    apply sgt_unit.
-  - rewrite grp_homo_op, grp_homo_inv.
-    by apply sgt_op.
+  apply subgroup_generated_rec.
+  intros g x.
```
jdchristensen, line 579: Now this `g` will match what was `x` above.  The `x` could be changed to `p`, to match the previous result.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -557,20 +557,28 @@ Definition subgroup_generated_gen_incl {G : Group} {X : G -> Type} (g : G) (H :
   : subgroup_generated X
   := (g; tr (sgt_in H)).
 
+Definition subgroup_generated_rec {G H : Group} (X : G -> Type) (S : Subgroup H)
+  (f : G $-> H) (i : forall x, X x -> S (f x))
+  : forall x, subgroup_generated X x -> S (f x).
```
jdchristensen, line 562: This result is really a combination of two things.  First, consider the special case when `f` is the identity map.  Then it says that `subgroup_generated X` is the smallest subgroup containing the generating set.  I think this is something worth recording on its own.

And then this result follows from that special case, since `S o f` is again a subgroup of `G` (by `subgroup_preimage`).

I think separating this out into those two pieces would be useful and a bit shorter (since the current approach redoes `subgroup_preimage`).

Alizter: I simplified the lemma like you suggested and added a comment. I'll add the preimage version back in if it becomes useful.

Alizter:
This worked out much nicer this way. My funny elimination principle I alluded to earlier became a special case by choosing a certain subgroup. It turns out you can precompose commutators with normal subgroups to get another subgroup. I have no idea what this is called, but it allowed for my later arguments to go through.

jdchristensen (approved):

theories/Algebra/Groups/Subgroup.v
jdchristensen: Sounds good.  Do you think `subgroup_generated_rec` should be renamed to something like `subgroup_generated_minimal`?

jdchristensen: Did you see the previous comment?

Alizter: I missed that. Whats wrong with `subgroup_generated_rec`?

jdchristensen: What the result is saying is that every subgroup containing the generators contains the generated subgroup, so the generated subgroup is the minimal subgroup containing the generators.  So I thought a name in that direction might make more sense.  `subgroup_generated_rec` would seem more appropriate for something that defines a function from the generated subgroup to an arbitrary group (or even an arbitrary type).

Alizter: That's a good point. I think I'll hold off on that for now. At some point I should go around and start using inclusions like we do in Ideal.v for subgroups, that way it would make sense to call this a `sg_incl` or something to that effect.

PR #2183, opened 2025-01-04
some helpful lemmas for FreeGroup

Alizter:
I forgot to PR this before #2172.

We state some useful results that come up often when working with maps out of free groups.

### DIFF
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -396,6 +396,7 @@ Section Reduction.
 End Reduction.
 
 Arguments FreeGroup_rec {A G}.
+Arguments FreeGroup_ind_homotopy {A G f f'}.
 Arguments freegroup_eta {A}.
 Arguments freegroup_in {A}.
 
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/FreeGroup.v
```diff
@@ -536,6 +537,27 @@ Section FreeGroupGenerated.
...
+
+Definition freegroup_rec_in {A : Type}
+  : FreeGroup_rec freegroup_in $== Id (FreeGroup A).
+Proof.
+  by snrapply FreeGroup_ind_homotopy.
+Defined.
+
+Definition freegroup_rec_compose {A B : Type}
+  (f : A -> FreeGroup B) {G : Group} (i : FreeGroup B $-> G)
+  : FreeGroup_rec (i o f) $== i $o FreeGroup_rec f.
```
jdchristensen, line 550: Here `FreeGroup B` can be replaced by any group `H`, right? (But maybe swap the names of `G` and `H` so the map `i` goes from `G` to `H`.  I would also change `i` to a different letter, maybe `k`.

Alizter: The letters are poor I agree. The free group is needed since i is post composed with free group rec.

jdchristensen: It works for me with `FreeGroup B` replaced with a Group `H`.

jdchristensen:
```coq
Definition freegroup_rec_compose {A : Type} {G H : Group}
  (f : A -> G) (k : G $-> H)
  : FreeGroup_rec (k o f) $== k $o FreeGroup_rec f.
Proof.
  by snrapply FreeGroup_ind_homotopy; intros x.
Defined.
```

jdchristensen (approved):

PR #2184, opened 2025-01-04
Elimination principles for GroupCoeq

Alizter:
Needed for #2172

### DIFF
--- a/theories/Algebra/Groups/FreeProduct.v
+++ b/theories/Algebra/Groups/FreeProduct.v
@@ -33,9 +33,9 @@ Local Open Scope mc_mult_scope.
 
 (** We will call M [amal_type] and prefix all the constructors with [amal_] (for amalgamated free product). *)
 
-Section FreeProduct.
+Section AmalgamatedFreeProduct.
 
-  Context (G H K : Group)
+  Context {G H K : Group}
     (f : GroupHomomorphism G H) (g : GroupHomomorphism G K).
 
   Local Definition Words : Type := list (H + K).
@@ -58,7 +58,6 @@ Section FreeProduct.
     simpl.
     destruct x; rhs napply app_assoc; f_ap.
   Defined.
-
 
   (** There are five source types for the path constructors. We will construct this HIT as the colimit of five forks going into [Words]. We can bundle up this colimit as a single coequalizer. *)
 
@@ -528,11 +527,25 @@ Section FreeProduct.
     snapply (Build_Group amal_type); repeat split; exact _.
   Defined.
 
+End AmalgamatedFreeProduct.
+
+Arguments amal_eta {G H K f g} x.
+Arguments amal_mu_H {G H K f g} x y.
+Arguments amal_mu_K {G H K f g} x y.
+Arguments amal_tau {G H K f g} x y z.
+Arguments amal_omega_H {G H K f g} x y.
+Arguments amal_omega_K {G H K f g} x y.
+
+Section RecInd.
+
+  Context {G H K : Group}
+    {f : GroupHomomorphism G H} {g : GroupHomomorphism G K}.
+
   (** Using foldr. It's important that we use foldr as foldl is near impossible to reason about. *)
   Definition AmalgamatedFreeProduct_rec' (X : Group)
     (h : GroupHomomorphism H X) (k : GroupHomomorphism K X)
     (p : h o f == k o g)
-    : AmalgamatedFreeProduct -> X.
+    : AmalgamatedFreeProduct f g -> X.
   Proof.
     srapply amal_type_rec.
     { intro w.
@@ -593,14 +606,14 @@ Section FreeProduct.
   Definition AmalgamatedFreeProduct_rec (X : Group)
     (h : H $-> X) (k : K $-> X)
     (p : h $o f $== k $o g)
-    : AmalgamatedFreeProduct $-> X.
+    : AmalgamatedFreeProduct f g $-> X.
   Proof.
     snapply Build_GroupHomomorphism.
     1: srapply (AmalgamatedFreeProduct_rec' X h k p).
     exact _.
   Defined.
 
-  Definition amal_inl : H $-> AmalgamatedFreeProduct.
+  Definition amal_inl : H $-> AmalgamatedFreeProduct f g.
   Proof.
     snapply Build_GroupHomomorphism.
     { intro x.
@@ -612,7 +625,7 @@ Section FreeProduct.
     reflexivity.
   Defined.
 
-  Definition amal_inr : K $-> AmalgamatedFreeProduct.
+  Definition amal_inr : K $-> AmalgamatedFreeProduct f g.
   Proof.
     snapply Build_GroupHomomorphism.
     { intro x.
@@ -631,7 +644,7 @@ Section FreeProduct.
 
   Theorem equiv_amalgamatedfreeproduct_rec `{Funext} (X : Group)
     : {h : H $-> X & {k : K $-> X & h $o f $== k $o g }}
-      <~> (AmalgamatedFreeProduct $-> X).
+      <~> (AmalgamatedFreeProduct f g $-> X).
   Proof.
     snapply equiv_adjointify.
     1: intros [h [k p]]; exact (AmalgamatedFreeProduct_rec X h k p).
@@ -666,7 +679,7 @@ Section FreeProduct.
     intro; simpl; rapply right_identity.
   Defined.
   
-  Definition amalgamatedfreeproduct_ind_hprop (P : AmalgamatedFreeProduct -> Type)
+  Definition amalgamatedfreeproduct_ind_hprop (P : AmalgamatedFreeProduct f g -> Type)
     `{forall x, IsHProp (P x)}
     (l : forall a, P (amal_inl a)) (r : forall b, P (amal_inr b))
     (Hop : forall x y, P x -> P y -> P (x * y))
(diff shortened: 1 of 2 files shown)


theories/Algebra/Groups/GroupCoeq.v
```diff
@@ -4,17 +4,47 @@ Require Import Truncations.Core.
...
+  rewrite <- (right_identity (f x)).
+  rewrite <- (right_identity (g x)).
+  rhs_V nrapply (amal_glue (freeproduct_inr x)).
+  symmetry.
+  nrapply (amal_glue (freeproduct_inl x)).
+Defined.
+
+Definition groupcoeq_rec {A B C : Group} (f g : A $-> B)
+  (h : B $-> C) (p : h $o f $== h $o g)
+  : GroupCoeq f g $-> C.
```
jdchristensen, line 38: It's a bit odd to define this one way, and to have a later result `equiv_groupcoeq_rec` that doesn't use this definition.  I realize that this one is useful as a separate result because it doesn't use Funext, but is it possible to use it in `equiv_groupcoeq_rec` somehow?

theories/Algebra/Groups/GroupCoeq.v
```diff
@@ -4,17 +4,47 @@ Require Import Truncations.Core.
...
+Proof.
+  rapply (AmalgamatedFreeProduct_rec C (h $o f) h).
+  snrapply freeproduct_ind_homotopy.
+  - refine (cat_assoc _ _ _ $@ _ $@ cat_assoc_opp _ _ _).
+    exact ((_ $@L freeproduct_rec_beta_inl _ _) $@ cat_idr _
+      $@ (_ $@L freeproduct_rec_beta_inl _ _)^$).
+  - refine (cat_assoc _ _ _ $@ _ $@ cat_assoc_opp _ _ _).
+    exact ((_ $@L freeproduct_rec_beta_inr _ _) $@ (cat_idr _ $@ p)
+      $@ (_ $@L freeproduct_rec_beta_inr _ _)^$).
```
jdchristensen, line 47: This is pretty hard to read, and the goals turn out to be simple things like `h (f (x * 1)) = h (f x * 1)`.  Also, since this is about homotopies, composition is definitionally associative (once you `intro x`).  That said, this gets the job done in just a few lines and is "abstract nonsense", so it has some good aspects.

theories/Algebra/Groups/GroupCoeq.v
```diff
@@ -4,17 +4,47 @@ Require Import Truncations.Core.
...
 Defined.
 
 Definition equiv_groupcoeq_rec `{Funext} {A B C : Group} (f g : GroupHomomorphism A B)
```
jdchristensen, line 50: I know it's not part of this PR, but the proof of `equiv_groupcoeq_rec` is long and involved.  It's a general fact about pushouts.  The key point is obscured, which is that given `f, g : FreeProduct A B $-> C`, homotopies `f $== g` are equivalent to pairs of homotopies `(f $o freeproduct_inl $== g $o freeproduct_inl) * (f $o freeproduct_inr $== g $o freeproduct_inr)`.  Then it's formally the kind of thing that `make_equiv_basedpaths` would solve, except that it all involves homotopies instead of paths.  But while this is easy to say abstractly, it's probably not that easy to implement...

theories/Algebra/Groups/GroupCoeq.v
```diff
@@ -55,6 +85,46 @@ Proof.
...
+Defined.
+
+Definition groupcoeq_ind_homotopy {G H K : Group} {f g : G $-> H}
+  {h h' : GroupCoeq f g $-> K}
+  (r : h $o groupcoeq_in $== h' $o groupcoeq_in)
+  : h $== h'.
+Proof.
+  rapply (groupcoeq_ind_hprop _ r).
+  intros x y p q; by nrapply grp_homo_op_agree.
+Defined.
```
jdchristensen, line 117: These lines are a bit slow.  Can they be sped up?

theories/Algebra/Groups/GroupCoeq.v
Alizter: Yes, the associativity is definitional, but doing so causes Coq to unfold a load of terms to unify stuff causing a performance penalty. Maybe I should add a comment explaining this however.

theories/Algebra/Groups/GroupCoeq.v
Alizter: This is because we are reusing `equiv_amalgamatedfreeproduct_rec` for the equivalence. Ideally, that proof should have an `isequiv_` version that we can reuse here. I am pretty sure the underlying maps are the same however.

theories/Algebra/Groups/GroupCoeq.v
jdchristensen: I think the slow down here as well as the other slow downs we discussed recently in FreeProduct.v come down to the fact that the SgOp instance for these groups involves the underlying amalgamated free product.  For example, if you Set Printing All, then just before the slow line (116 above) the goal looks like:
```coq
  forall (x y : group_type (@GroupCoeq G H f g))
    (_ : @paths (group_type K) (@grp_homo_map (@GroupCoeq G H f g) K h x)
           (@grp_homo_map (@GroupCoeq G H f g) K h' x))
    (_ : @paths (group_type K) (@grp_homo_map (@GroupCoeq G H f g) K h y)
           (@grp_homo_map (@GroupCoeq G H f g) K h' y)),
  @paths (group_type K)
    (@grp_homo_map (@GroupCoeq G H f g) K h
       (@sg_op (group_type (@GroupCoeq G H f g))
          (sgop_amal_type (FreeProduct G G) G H
             (@FreeProduct_rec G G G (@Id Group isgraph_group is01cat_group G)
                (@Id Group isgraph_group is01cat_group G)) (@FreeProduct_rec G G H f g)) x y))
    (@grp_homo_map (@GroupCoeq G H f g) K h'
       (@sg_op (group_type (@GroupCoeq G H f g))
          (sgop_amal_type (FreeProduct G G) G H
             (@FreeProduct_rec G G G (@Id Group isgraph_group is01cat_group G)
                (@Id Group isgraph_group is01cat_group G)) (@FreeProduct_rec G G H f g)) x y))
```
Since we defined GroupCoeq to have type Group, I'm not sure why the SgOp always ends up unfolded like this.  It would be great to figure this out, as I think we could solve multiple slowdowns.

Ah, I figured it out.  It's because of the Global Instance for SgOp for the amalgamated type.  Making it Local solves the slowdown here and the worst slowdowns in FreeProduct.v.  There's one other slightly slow bit in FreeProduct.v that might be solved by closing and reopening the section.

I push a commit that makes all of those Instances Local.

theories/Algebra/Groups/GroupCoeq.v
Alizter: I've rephrased the argument like you've suggested. It is more concise now.

theories/Algebra/Groups/FreeProduct.v
```diff
@@ -722,16 +739,21 @@ Definition freeproduct_ind_hprop {G H} (P : FreeProduct G H -> Type)
   (Hop : forall x y, P x -> P y -> P (x * y))
   : forall x, P x
   := amalgamatedfreeproduct_ind_hprop P l r Hop.
+(* The above is slow, ~0.15s. *)
```
jdchristensen, line 742: I'm surprised by how slow the above is.  I tried various things, and couldn't improve it.  Even providing the term with `exact_no_check` and all implicit arguments is slow:
```coq
Time Definition freeproduct_ind_hprop {G H} (P : FreeProduct G H -> Type)
  `{forall x, IsHProp (P x)}
  (l : forall g, P (freeproduct_inl g))
  (r : forall h, P (freeproduct_inr h))
  (Hop : forall x y, P x -> P y -> P (x * y))
  : forall x, P x.
Proof.
  unfold FreeProduct in *.
  unfold freeproduct_inl, freeproduct_inr in *.
  Time exact_no_check (@amalgamatedfreeproduct_ind_hprop grp_trivial G H (grp_trivial_rec G) 
  (grp_trivial_rec H) P H0 l r Hop).  (* 0.075s *)
Time Defined. (* 0.075s *)
```

jdchristensen:
I cleaned up and sped up a few things, but there are also some slow lines, some of which I marked with comments.

But here's a bigger question about GroupCoeq:  would it be better to define the coequalizer of `f g : G $-> H` as the quotient of `H` by the normal subgroup generated by `f(x) g(x)^` for `x : G`?  The current definition involves *two* amalgamated products, which makes it quite complicated.  With my proposal, the recursion and induction principles would be easier to see.  It would also be obvious that the map from `H` to the coequalizer is surjective, making the `_ind_hprop` result trivial (and with fewer hypotheses).

My proposal would require us to define the normal subgroup generated by a subset, but that shouldn't be too bad, and is important to have in any case.

theories/Algebra/Groups/GroupCoeq.v
```diff
@@ -4,57 +4,112 @@ Require Import Truncations.Core.
...
+    by rewrite 2 grp_unit_r.
+  - snrapply equiv_functor_forall_id; simpl; intros a.
+    by rewrite 2 grp_unit_r.
+Defined.
+
+Definition groupcoeq_ind_hprop {G H : Group} {f g : G $-> H}
+  (P : GroupCoeq f g -> Type) `{forall x, IsHProp (P x)}
+  (c : forall h, P (groupcoeq_in h))
+  (Hop : forall x y, P x -> P y -> P (x * y))
+  : forall x, P x.
```
jdchristensen, line 81: Since the map from `H` to the coequalizer is surjective, the `Hop` assumption shouldn't be needed.  But I'm not sure how to show that this map is surjective using the current definition of coequalizer.

Alizter:
> I cleaned up and sped up a few things, but there are also some slow lines, some of which I marked with comments.

Thanks, they were annoying me too.

> But here's a bigger question about GroupCoeq: would it be better to define the coequalizer of `f g : G $-> H` as the quotient of `H` by the normal subgroup generated by `f(x) g(x)^` for `x : G`? The current definition involves _two_ amalgamated products, which makes it quite complicated. With my proposal, the recursion and induction principles would be easier to see. It would also be obvious that the map from `H` to the coequalizer is surjective, making the `_ind_hprop` result trivial (and with fewer hypotheses).

Yes, there is no real reason to define the coequalizer as an amalgamated free product. In fact, the reason I probably chose to do it this way was a blind application of coproducts + pushout gives coequalizers. Now thinking about it, this seems kind of backwards as pushouts are quite complicated compared to coequalizers. I would tehrefore 

> My proposal would require us to define the normal subgroup generated by a subset, but that shouldn't be too bad, and is important to have in any case.

jdchristensen:
I would be ok with merging this as-is and leaving the job of redefining the coequalizer to another PR, if you want.

Alizter:
I will need to rebase this before however.

I'm travelling at the moment, but I think I know how to formalize normal closures. We need to take the conjugates of all elements of a set S and then take the subgroup closure.

There will be a bit of work on various properties we expect normal closures to have, but it should be enough to replace the definition of the coequalizer. Not sure if it will fit in this PR however, so maybe it's a good idea to merge after rebasing.

jdchristensen:
Yeah, I think it's probably best to merge this after rebasing, and then make a new PR with the new implementation.

jdchristensen:
I just remembered that two of the slow lines could be sped up by closing and reopening the section, getting rid of the local instances that caused the slowdown there.  It doesn't help `freeproduct_ind_hprop`, which is still a bit of a mystery to me.

While I was at it, I also improved various implicit arguments.

Alizter:
@jdchristensen I rebased this without pulling in the latest changes. Could you force push your latest commit again, and I will try to rebase properly. Sorry about that.

Alizter:
I was not able to rebase this and ran out of time today. @jdchristensen Could you have a go?

jdchristensen:
@Alizter I was able to rebase it, but it was a bit of a pain.  I guess we should have worked through the backlog of PRs before doing library-wide changes.  In any case, things work now.  Can you give it a quick check before merging, to make sure I didn't mess up the rebase?  (I just force-pushed a second time, moving my build-fix into the commit that broke the build, so now I think all commits should build without error.  I'm done now.)

Alizter:
Thanks @jdchristensen. I also had a lot of trouble rebasing, but it looks like we went down the same route. I'll have a go at redefining `GroupCoeq` as we discussed now.

PR #2186, opened 2025-01-04
improvements to trivial group

Alizter:
First, we rename `IsTrivialSubgroup` to `IsTrivialGroup`. `maximal_subgroup` is now a coercion, which means we can write `IsTrivialGroup G` for a subgroup `G`, therefore dropping the `sub` seemed appropriate.

Next we simplify the definition by only requiring a map into the trivial group, which corresponds to a subgroup inclusion. The reverse inclusion is trivial, and is stated for completeness. It didn't need to be part of the data.

Finally we show that a group is trivial iff it is isomorphic to the trivial group. Seems negligible on paper, but some details need to be provided.

### DIFF
--- a/theories/Algebra/Groups/Subgroup.v
+++ b/theories/Algebra/Groups/Subgroup.v
@@ -496,9 +496,8 @@ Proof.
 Defined.
 
 (** The property of being the trivial group is useful. Note that any group can be automatically coerced to its maximal subgroup, so it makes sense for this predicate to be applied to groups in general. *)
-Definition IsTrivialGroup@{i} {G : Group@{i}} (H : Subgroup@{i i} G) :=
-  forall x, H x -> trivial_subgroup G x.
-Existing Class IsTrivialGroup.
+Class IsTrivialGroup@{i} {G : Group@{i}} (H : Subgroup@{i i} G) :=
+  istrivialgroup : forall x, H x -> trivial_subgroup G x.
 
 Global Instance istrivial_trivial_subgroup {G : Group}
   : IsTrivialGroup (trivial_subgroup G)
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Subgroup.v
```diff
@@ -488,14 +495,38 @@ Proof.
...
-  forall x, H x <-> trivial_subgroup G x.
-Existing Class IsTrivialSubgroup.
-
-Global Instance istrivialsubgroup_trivial_subgroup {G : Group}
-  : IsTrivialSubgroup (@trivial_subgroup G)
-  := ltac:(hnf; reflexivity).
+(** The property of being the trivial group is useful. Note that any group can be automatically coerced to its maximal subgroup, so it makes sense for this predicate to be applied to groups in general. *)
+Definition IsTrivialGroup@{i} {G : Group@{i}} (H : Subgroup@{i i} G) :=
+  forall x, H x -> trivial_subgroup G x.
+Existing Class IsTrivialGroup.
```
jdchristensen, line 501: Why not make it a Class directly instead of a Definition?

theories/Algebra/Groups/Subgroup.v
```diff
@@ -488,14 +495,38 @@ Proof.
...
+
+Global Instance istrivial_trivial_subgroup {G : Group}
+  : IsTrivialGroup (trivial_subgroup G)
+  := fun x => idmap.
+
+(** Trivial groups are isomorphic to the trivial group. *)
+Definition istrivial_iff_grp_iso_trivial_subgroup {G : Group} (H : Subgroup G)
+  : IsTrivialGroup H <-> (H : Group) $<~> trivial_subgroup G.
```
jdchristensen, line 509:
```suggestion
  : IsTrivialGroup H <-> ((H : Group) $<~> trivial_subgroup G).
```
In general I'm against redundant parentheses, but it's really not clear which of `<->` and `$<~>` have higher priority, so they help here.  I think it would also be better to explictly write the coercions here and then drop the `: Group` annotation.

jdchristensen (approved):

PR #2187, opened 2025-01-04
group of maximal subgroup of a group is iso to said group

Alizter:
This isn't true definitionally due to the way we define subgroups. This lemma at least lets us pass along the obvious equivalence when it appears in arguments.

Relies on (because of Require)
- #2186

### DIFF
--- a/theories/Algebra/Groups/Subgroup.v
+++ b/theories/Algebra/Groups/Subgroup.v
@@ -222,7 +222,7 @@ Defined.
 Coercion maximal_subgroup : Group >-> Subgroup.
 Add Printing Coercion maximal_subgroup.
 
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Subgroup.v
```diff
@@ -222,6 +222,15 @@ Defined.
 Coercion maximal_subgroup : Group >-> Subgroup.
 Add Printing Coercion maximal_subgroup.
 
+(** The group inferred from the maximal subgroup is the original group. *)
```
jdchristensen, line 225: "inferred"?  Maybe "associated to"?

jdchristensen (approved):

PR #2188, opened 2025-01-04
quotient by a trivial group

Alizter:
Quotienting by a trival group does nothing.

### DIFF
--- a/theories/Algebra/Groups/QuotientGroup.v
+++ b/theories/Algebra/Groups/QuotientGroup.v
@@ -304,3 +304,16 @@ Proof.
     + srapply path_sigma_hprop.
       reflexivity.
 Defined.
+
+(** The group quotient by a trivial group is isomorphic to the original group. *)
+Definition grp_quotient_trivial (G : Group) (N : NormalSubgroup G)
+  : IsTrivialGroup N -> G / N ≅ G.
+Proof.
+  intros T.
+  snrapply cate_adjointify.
+  - srapply (grp_quotient_rec _ _ (Id _)).
+    apply T.
+  - srapply grp_quotient_map.
+  - reflexivity.
+  - by srapply grp_quotient_ind_hprop.
+Defined.


theories/Algebra/Groups/QuotientGroup.v
```diff
@@ -304,3 +304,16 @@ Proof.
...
+  : IsTrivialGroup N -> G / N ≅ G.
+Proof.
+  intros T.
+  snrapply cate_adjointify.
+  - srapply (grp_quotient_rec _ _ (Id _)).
+    apply T.
+  - srapply grp_quotient_map.
+  - reflexivity.
+  - by srapply grp_quotient_ind_hprop.
+Defined.
```
jdchristensen, line 319: This is fine, but there's also a conceptual way:  By the previous result, the kernel of `grp_quotient_map` is the trivial group, so it is an embedding, by `equiv_kernel_isembedding`.  And `grp_quotient_map` is always surjective.  So it's iso.  Not sure how this approach would work out in practice, but it would be good if all the tools were in place to make arguments like that easy.

theories/Algebra/Groups/QuotientGroup.v
```diff
@@ -304,3 +304,16 @@ Proof.
     + srapply path_sigma_hprop.
       reflexivity.
 Defined.
+
+(** The group quotient by a trivial group is isomorphic to the original group. *)
+Definition grp_quotient_trivial (G : Group) (N : NormalSubgroup G)
+  : IsTrivialGroup N -> G / N ≅ G.
```
jdchristensen, line 310: I would move `IsTrivialGroup N` to the left of the colon, reverse the direction of the equivalence (since the map in the other direction always exists), and replace the isomorphism symbol with `$<~>`.  Also, it could be sharped to an IsCatEquiv statement about `grp_quotient_map`.

theories/Algebra/Groups/QuotientGroup.v
Alizter: It seems however univalence is required for showing that `grp_quotient_map` has a trivial kernel. This is because the inclusion of the kernel in N means we have to turn an equality `grp_quotient_map g = 1` into something about `N` and the only way to do that is with `related_quotient_paths` which in turn gets univalence from `in_class` which in turn requires it because a relation on `A / R` will need to elminate into a set, therefore has to be HProp valued.

Alizter: Correction: Univalence is required to show the kernel is N, even if we weaken to subgroup iff. The kernel is then trivial because N is.

Alizter: I pushed some stuff on trivial kernels being embeddings. But I didn't use it yet since univalence is required. It seems in this case giving the equivalence directly is simpler if a bit more verbose.

theories/Algebra/Groups/Kernel.v
```diff
@@ -55,17 +56,34 @@ Proof.
     apply path_sigma_hprop; reflexivity.
 Defined.
 
-(** ** Characterisation of group embeddings *)
-Proposition equiv_kernel_isembedding `{Univalence} {A B : Group} (f : A $-> B)
-  : (grp_kernel f = trivial_subgroup A :> Subgroup A) <~> IsEmbedding f.
+(** The underlying map of a group homomorphism with a trivial kernel is an embedding. *)
+Global Instance isembedding_istrivial_kernel {G H : Group} (f : G $-> H)
+  (triv : IsTrivialGroup (grp_kernel f))
+  : IsEmbedding f.
```
jdchristensen, line 62: It turns out that this is definitionally the same as the first implication of `isembedding_grouphomomorphism`, so one can use
```coq
:= fst (isembedding_grouphomomorphism f) triv.
```
This raises the question of whether we really need this result at all, or if the two can be merged somehow.

theories/Algebra/Groups/Kernel.v
```diff
@@ -55,17 +56,34 @@ Proof.
...
+  apply isembedding_grouphomomorphism.
+  intros g p.
+  apply triv.
+  exact p.
+Defined.
+
+(** If the underlying map of a group homomorphism is an embedding then the kernel is trivial. *)
+Definition istrivial_kernel_isembedding {G H : Group} (f : G $-> H)
+  (emb : IsEmbedding f)
+  : IsTrivialGroup (grp_kernel f).
```
jdchristensen, line 73: And this is just the other direction, so this works:
```coq
  := snd (isembedding_grouphomomorphism f) emb.
```
If we split `isembedding_grouphomomorphism` into its two implications, then maybe these two results wouldn't be needed?

theories/Algebra/Groups/Kernel.v
```diff
@@ -55,17 +56,34 @@ Proof.
...
-  refine (_ oE (equiv_path_subgroup' _ _)^-1%equiv).
   apply equiv_iff_hprop_uncurried.
-  refine (iff_compose _ (isembedding_grouphomomorphism f)); split.
-  - intros E ? ?.
-    by apply E.
-  - intros e a; split.
-    + apply e.
-    + intro p.
-      exact (ap _ p @ grp_homo_unit f).
+  split; exact _.
```
jdchristensen, line 88: Here `apply isembedding_grouphomomorphism.` works instead of the second line.

theories/Algebra/Groups/QuotientGroup.v
jdchristensen: Ok, it seems like identifying the kernel of the map `G -> G / N` in general requires univalence, so passing via this is not the best approach.  So I'm ok with leaving this in the current style.  But my other comments about the direction of the map and using `CatIsEquiv` still apply, I think.

theories/Algebra/Groups/Kernel.v
Alizter: I've removed the iff and just kept both directions separately.

jdchristensen (approved):

PR #2190, opened 2025-01-08
some remaining exercises in Chapter 2

Theongck:


contrib/HoTTBookExercises.v
```diff
@@ -680,7 +680,8 @@ Definition Book_2_13 := @HoTT.Types.Bool.equiv_bool_aut_bool.
 (* ================================================== ex:equality-reflection *)
 (** Exercise 2.14 *)
 
-
+(** For any q : x = y, q and refl_x has the same type x = x, so we can form the type Pi_{x,y} Pi_q (q = refl_x). *)
+(** A path induction produces an element r, with (r x x p) : p = refl_x, which is also definitional by assumption *)
```
jdchristensen, line 684: Can you state Exercise 2.14 before giving your answer?  Otherwise someone reading this will find what you wrote very confusing, as `q` doesn't have type `x = x` in the theory used in the library.

About formatting:  make all of your answer a single comment, and put no line breaks in it.  Also, put type-theoretic notation in square brackets, e.g. "For any [q : x = y], [q] and [refl_x] have ..."   (And note that it is "have" not "has".)  End the last sentence with a period.

contrib/HoTTBookExercises.v
```diff
@@ -700,7 +703,16 @@ Definition Book_2_13 := @HoTT.Types.Bool.equiv_bool_aut_bool.
 (* ================================================== ex:dep-htpy-natural *)
 (** Exercise 2.18 *)
 
+(** This exercise is solved in the library as HoTT.Basics.PathGroupoids.apD_homotopic *)
```
jdchristensen, line 706:
```suggestion
(** This exercise is solved in the library as HoTT.Basics.PathGroupoids.apD_homotopic, except that the equation is rearranged slightly. *)
```
Or maybe this version should be added to the library, to be analogous to `concat_Ap`?  And then `apD_homotopic` could be proved using this new result?  @Alizter , what do you think?

jdchristensen (commented):
Thanks for this!  Just a couple of comments.

contrib/HoTTBookExercises.v
Alizter: This sounds sensible. It may be a bit tricky to do due to dependent paths however.

Alizter (approved):
Thanks, this is great! I'm happy with this once you've addressed Dan's suggestion of adding a variant of `apD_homotopic`. Not sure what to name that though. It ought to be something like `concat_Ap` but that doesn't quite work since we are doing some sort of dependent concatenation.

Alizter (approved):
Thanks, this is great! I'm happy with this once you've addressed Dan's suggestion of adding a variant of `apD_homotopic`. Not sure what to name that though. It ought to be something like `concat_Ap` but that doesn't quite work since we are doing some sort of dependent concatenation.

contrib/HoTTBookExercises.v
```diff
@@ -680,17 +680,19 @@ Definition Book_2_13 := @HoTT.Types.Bool.equiv_bool_aut_bool.
 (* ================================================== ex:equality-reflection *)
 (** Exercise 2.14 *)
 
-
+(** Assuming the equality reflection rule, given any [q : x = y], [x] and [y] are definitionally equal, so [q] and [refl_x] have the same type [x = x]. We can form the type [forall x y, forall q, q = refl_x]. A path induction produces an element r, with [(r x x p) : p = refl_x], which is also definitional by the equality reflection rule. *)
```
jdchristensen, line 683:
```suggestion
(** Assuming the equality reflection rule, given any [q : x = y], [x] and [y] are definitionally equal, so [q] and [refl_x] have the same type [x = x]. We can form the type [forall x y, forall q, q = refl_x]. A path induction produces an element [r], with [r x x p : p = refl_x], which is also definitional by the equality reflection rule. *)
```

jdchristensen:
> Thanks, this is great! I'm happy with this once you've addressed Dan's suggestion of adding a variant of `apD_homotopic`. Not sure what to name that though. It ought to be something like `concat_Ap` but that doesn't quite work since we are doing some sort of dependent concatenation.

What if we just call it `apD_natural` and put it right before `apD_homotopic`?  (I'm also tempted to rename `concat_Ap` to `ap_natural`, but then there would be many other similar results of the form "concat_...A..." that would not fit into the pattern.)

Alizter:
Looks great, thanks! @jdchristensen feel free to merge when you've taken a look.

jdchristensen:
Thanks, @Theongck!  If you'd like to contribute more to the library, that would be great!

I'll merge once most of the CI passes.

PR #2191, opened 2025-01-09
symmetric monoidal categories satisfy the other hexagon

Alizter:
This is the other hexagon axiom you get when considering braided monoidal categories. I noticed that this is the hexagon axiom we get for opposite symmetric monoidal categories upto some associativity.

When we define braided monoidal categories, the opposite braided monoidal category will just have its hexagons swapped and reassociated which is quite nice.

### DIFF
--- a/theories/WildCat/Monoidal.v
+++ b/theories/WildCat/Monoidal.v
@@ -113,7 +113,20 @@ Class HexagonIdentity {A : Type} `{HasEquivs A}
 Coercion hexagon_identity : HexagonIdentity >-> Funclass.
 Arguments hexagon_identity {A _ _ _ _ _} F {_ _}.
 
+(** *** Hexagon identity with inverse associators *)
+
 (** ** Monoidal Categories *)
+
+Class HexagonIdentityInverseAssoc {A : Type} `{HasEquivs A}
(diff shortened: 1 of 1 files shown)


Alizter:
ping @jdchristensen

theories/WildCat/Monoidal.v
```diff
@@ -97,7 +97,6 @@ Class SymmetricBraiding {A : Type} `{Is1Cat A}
   braiding_symmetricbraiding :: Braiding F;
   braid_braid : forall a b, braid a b $o braid b a $== Id (F b a);
 }.
-(** We could have used [::>] in [braiding_symmetricbraiding] instead however due to bug https://github.com/coq/coq/issues/18971 the coercion isn't registered, so we have to register it manually instead. *)
```
jdchristensen, line 100: I removed this since our current style is not to use :> since it introduces a reversible coercion, IIRC.

jdchristensen (approved):

jdchristensen:
Sorry again for the delay.  This LGTM.  I just pushed a few minor clean-ups.  BTW, `lhs'`, `rhs'` and setoid rewriting would probably make things like this easier now, but I know they weren't available while you were working on this.

PR #2192, opened 2025-01-11
Defining properties of maxima and minima, lemmas for the take operation in lists, lists from sequences.

thchatzidiamantis:
Added various small results and improvements that will be used in future PRs.

**Nat.Core:**

- Properties of `nat_max` and `nat_min`.

**List.Theory:**

-  Results for `take` (two takes are the same as a take with the minimum, commutativity, invariance of `take n` with respect to concatenation if n is less than or equal to the length of the first list).
- Adapted the `Build_Vector` constructor to get a `Build_list`  constructor for a list given a function from `nat`, changed the results for Vectors so that they are based on this.
- Simplification of definitions by @jdchristensen

### DIFF
--- a/theories/Spaces/List/Theory.v
+++ b/theories/Spaces/List/Theory.v
@@ -542,32 +542,22 @@ Defined.
 
 (** [drop n l] removes the first [n] elements of [l]. *)
 Fixpoint drop {A : Type} (n : nat) (l : list A) : list A :=
-  match l, n with
-  | _ :: l, n.+1 => drop n l
-  | _, _ => l
+  match n with
+  | 0 => l
+  | n.+1 => drop n (tail l)
   end.
 
-(** A [drop] of zero elements is the identity. *)
-Definition drop_0 {A : Type} (l : list A)
-  : drop 0 l = l.
-Proof.
-  by destruct l.
-Defined.
+(** A [drop] of zero elements is the identity, by definition. *)
+Definition drop_0 {A : Type} (l : list A) : drop 0 l = l := idpath.
 
-(** A [drop] of one element is the tail of the list. *)
-Definition drop_1 {A : Type} (l : list A)
-  : drop 1 l = tail l.
-Proof.
-  induction l.
-  1: reflexivity.
-  by destruct l.
-Defined.
+(** A [drop] of one element is the tail of the list, by definition. *)
+Definition drop_1 {A : Type} (l : list A) : drop 1 l = tail l := idpath.
 
 (** A [drop] of the empty list is the empty list. *)
 Definition drop_nil {A : Type} (n : nat)
   : drop n (@nil A) = nil.
 Proof.
-  by destruct n.
+  by induction n.
 Defined.
 
 (** A [drop] of [n] elements with [length l <= n] is the empty list. *)
@@ -603,7 +593,7 @@ Proof.
   induction l as [|a l IHl] in n, H, x |- * using list_ind@{i i}.
   1: rewrite drop_nil in H; contradiction.
   destruct n.
-  1: rewrite drop_0 in H; assumption.
+  1: exact H.
   right; nrapply (IHl _ _ H).
 Defined.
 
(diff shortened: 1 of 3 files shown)


jdchristensen:
The commits here are best read independently, as the %nat changes clutter the diff.  The library builds after each commit.

theories/Spaces/Nat/Core.v
```diff
@@ -803,6 +831,33 @@ Proof.
   exact (ap S IH).
 Defined.
 
+(** The defining properties of [nat_min]: [nat_min n m] is less than or equal to [m] and [n] and any number less than or equal to [m] and [n] is less than or equal to [nat_min n m]. *)
+Definition nat_min_leq_l@{} n m : nat_min n m <= n.
```
Alizter, line 835: The names here are inconsistent with the max versions. İt should be leq_nat_min. Same for the other min lemmas.

theories/Spaces/List/Theory.v
```diff
@@ -653,28 +655,56 @@ Proof.
...
+(** [take] is commutative in [n]. *)
+Definition take_comm {A : Type} {m n : nat} (l : list A)
+  : take n (take m l) = take m (take n l).
+Proof.
+  by rewrite !take_take_min, nat_min_comm.
+Defined.
+
+(** A [take n] does not change under concatenation if [n] is less than or equal to the length of the first list. *)
+Definition take_app {A : Type} {n : nat} (l1 l2 : list A) (hn : n <= length l1)
+  : take n l1 = take n (l1++l2).
```
Alizter, line 686:
```suggestion
  : take n l1 = take n (l1 ++ l2).
```

theories/Spaces/Nat/Core.v
thchatzidiamantis: Do you have a suggestion for a better name for `nat_max_leq` and `nat_min_leq`? I had originally named them based on the side of the `<=` that `nat_max` or `nat_min` appeared, but I realise that this was confusing and inconsistent.

Alizter: I think `leq_nat_min_l` is the best name. It says to me it is a `leq` about `nat_min` focused on the left argument. That is unambiguous for `nat_min`. You had the right idea with `leq_nat_max_l`.

theories/Spaces/List/Theory.v
```diff
@@ -925,6 +955,50 @@ Proof.
...
+  := Build_list n (fun m _ => s m).
+
+Definition length_list_restrict {A : Type} (s : nat -> A) (n : nat)
+  : length (list_restrict s n) = n
+  := length_Build_list _ _.
+
+(** [nth'] of the restriction of a sequence is the corresponding term of the sequence.  *)
+Definition nth'_list_restrict {A : Type} (s : nat -> A) (n : nat)
+  {i : nat} (Hi : i < n)
+  : nth' (list_restrict s n) i ((length_list_restrict s n)^ # Hi) = s i.
```
Alizter, line 995: You can make this slightly more general by accepting any proof that `i < length (list_restrict s n)` rather than this specific one.

theories/Spaces/List/Theory.v
```diff
@@ -620,7 +617,7 @@ Defined.
 
 (** A [take] of [n] elements with [length l <= n] is the original list. *)
 Definition take_length_leq@{i|} {A : Type@{i}} (n : nat) (l : list A)
-  (H : (length l <= n)%nat)
+  (H : (length l <= n))
```
Alizter, line 620:
```suggestion
  (H : length l <= n)
```

theories/Spaces/Nat/Core.v
```diff
@@ -740,6 +740,34 @@ Proof.
...
+Definition leq_nat_max_r@{} n m : m <= nat_max n m.
+Proof.
+  induction n as [|n IHn] in m |- *.
+  1: cbn; reflexivity.
+  destruct m; cbn.
+  1: exact (leq_zero_l _).
+  exact (leq_succ (IHn m)).
+Defined.
+
+Definition nat_max_leq@{} {n m k : nat} (Hn : n <= k) (Hm : m <= k)
```
Alizter, line 762:
```suggestion
Definition leq_nat_max@{} {n m k : nat} (Hn : n <= k) (Hm : m <= k)
```
This won't be ambiguous with the other lemmas. Same for `nat_min_leq` please.

Alizter:
We can also consider making those `leq` lemmas for `nat_max` and `nat_min` instances, but I think that's something we should do once we start using it. Presumably you have a development that relies on those lemmas, and if it makes the proofs more slick there then its worth turning these into instances in that later PR.

thchatzidiamantis:
> Presumably you have a development that relies on those lemmas, and if it makes the proofs more slick there then its worth turning these into instances in that later PR.

I have only used these lemmas once to prove `length_take_leq` for lists, the results for lists are the important ones to what I am working on.

Alizter (approved):
LGTM, I'll let @jdchristensen merge.

PR #2194, opened 2025-01-15
maximal subgroups

Alizter:
Similarly to `IsTrivialGroup` we define what it means for a subgroup to be maximal (with respect to the tautological predicate). This means that the predicate of the subgroup always has a section.

We also add a comment mentioning that being isomorphic to the maximal subgroup is not enough to be maximal. Take `2Z` and `Z` in `Z`.

### DIFF
--- a/theories/Algebra/Groups/Subgroup.v
+++ b/theories/Algebra/Groups/Subgroup.v
@@ -547,7 +547,7 @@ Class IsMaximalSubgroup {G : Group} (H : Subgroup G) :=
 
 Global Instance ishprop_ismaximalsubgroup `{Funext}
   {G : Group} (H : Subgroup G)
-  : IsHProp (IsMaximalSubgroup G)
+  : IsHProp (IsMaximalSubgroup H)
   := istrunc_forall.
 
 Global Instance ismaximalsubgroup_maximalsubgroup {G : Group}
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Subgroup.v
```diff
@@ -541,6 +541,21 @@ Proof.
     apply path_contr.
 Defined.
 
+(** A maximal subgroup of a group [G] is a subgroup where every element of [G] is included. *)
+Class IsMaximalSubgroup {G : Group} (H : Subgroup G) :=
+  ismaximalsubgroup : forall (x : G), H x.
+
+Global Instance ishprop_ismaximalsubgroup {F : Funext}
```
jdchristensen, line 548:
```suggestion
Global Instance ishprop_ismaximalsubgroup `{Funext}
```

theories/Algebra/Groups/Subgroup.v
```diff
@@ -541,6 +541,21 @@ Proof.
...
+  {G : Group} (H : Subgroup G)
+  : IsHProp (IsMaximalSubgroup G)
```
jdchristensen, line 550:
```suggestion
  : IsHProp (IsMaximalSubgroup H)
```

jdchristensen (approved):
LGTM once the suggestions are done.

PR #2198, opened 2025-01-15
Misc cleanups to PathGroupoids, Equiv, HProp and BoundedSearch

jdchristensen:
The commits are independent and self-explanatory.

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -16,7 +16,7 @@ Local Open Scope path_scope.
 
    Theorems about concatenation of paths are called [concat_XXX] where [XXX] tells us what is on the left-hand side of the equation. You have to guess the right-hand side. We use the following symbols in [XXX]:
 
-   - [1] means the identity path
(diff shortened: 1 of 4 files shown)


theories/Basics/PathGroupoids.v
```diff
@@ -16,7 +16,7 @@ Local Open Scope path_scope.
 
    Theorems about concatenation of paths are called [concat_XXX] where [XXX] tells us what is on the left-hand side of the equation. You have to guess the right-hand side. We use the following symbols in [XXX]:
 
-   - [1] means the identity path
+   - [1] means the identity path, or sometimes the identity function
```
Alizter, line 19: Where does it mean the identity function? Typically we write `id` or `idmap` in those cases.

Alizter (approved):

theories/Basics/PathGroupoids.v
jdchristensen: `concat_A1p' and some others.  Should we rename them?

PR #2200, opened 2025-01-16
minor cleanups in Groups/

Alizter:


theories/Algebra/Groups/Group.v
```diff
@@ -995,7 +995,7 @@ Defined.
 
 (** Pairs in direct products can be decomposed *)
 Definition grp_prod_decompose {G H : Group} (g : G) (h : H)
-  : (g, h) = ((g, group_unit) : grp_prod G H) * (group_unit, h).
+  : (g, h) = ((g, 1) : grp_prod G H) * (1, h).
```
jdchristensen, line 998: Does this work?  If so, I think it's cleaner.
```suggestion
  : (g, h) = (g, 1) * (1, h) :> grp_prod G H.
```

jdchristensen (approved):
Feel free to merge with or without my suggested change.

theories/Algebra/Groups/Group.v
Alizter: It does not since it gets stuck looking for sgop of a prod. I had tried this before ideally we would have a  pair "constructor" for groups.

PR #2201, opened 2025-01-17
image of a subgroup

Alizter:
Here we introduce the image of a subgroup under a group homomorphism. This is obviously a generalisation of what exists in `Image.v` however since `subgroup_group (maximal_subgroup G)` isn't definitionally `G`, it makes working with the "standard image" a little bit more finicky, so I deemed it was not worth merging them.

I therefore opted to introduce this version specifically for subgroups and proved some standard properties about it.

We show that images are adjoint to preimages and images of normal subgroups under a surjective homomorphism are normal.


theories/Algebra/Groups/Subgroup.v
```diff
@@ -782,3 +782,87 @@ Defined.
...
+
+(** Converting the subgroups to groups, we get the expected surjective (epi) restriction homomorphism. *)
+Definition grp_homo_subgroup_image {G H : Group} (f : G $-> H) (J : Subgroup G)
+  : subgroup_group J $-> subgroup_image f J
+  := functor_subgroup_group f (subgroup_image_in _ _).
+
+(** The restriction map from the subgroup to the image is surjective as expected. *)
+Global Instance issurj_grp_homo_subgroup_image {G H : Group}
+  (f : G $-> H) (J : Subgroup G)
+  : IsSurjection (grp_homo_subgroup_image f J).
```
jdchristensen: This follows from `conn_map_factor1_image`.  I'll push the change.  I wonder if we should more explicitly use `himage` or `image` for some of these results?  Unfortunately, `himage` and `image` produce the whole sigma type, not just the predicate needed in `subgroup_image`.

jdchristensen: BTW, the file compiles with this result removed.  I think it's good to have as documentation, but maybe it shouldn't be an instance since it is redundant?

Alizter: Yes, I didn't have that import which made me miss that it was inferred. Perhaps we should export that instance in this file also to make sure.

jdchristensen: Yes, that makes sense.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -782,3 +782,87 @@ Defined.
...
+
+(** By definition, values of [f x] where [x] is in a subgroup [J] are in the image of [J] under [f]. *)
+Definition subgroup_image_in {G H : Group} (f : G $-> H) (J : Subgroup G)
+  : forall x, J x -> subgroup_image f J (f x)
+  := fun x Jx => tr ((x; Jx); idpath).
+
+(** Converting the subgroups to groups, we get the expected surjective (epi) restriction homomorphism. *)
+Definition grp_homo_subgroup_image {G H : Group} (f : G $-> H) (J : Subgroup G)
+  : subgroup_group J $-> subgroup_image f J
+  := functor_subgroup_group f (subgroup_image_in _ _).
```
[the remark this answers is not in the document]
Alizter: I changed the name of `grp_image_in` to `grp_homo_image_in` and added `grp_image_in` as an intermediate lemma.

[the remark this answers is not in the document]
Alizter: I've also added `_in` as you suggested.

Alizter:
I've also absorbed `Kernel` since it didn't make much sense for this to be a separate file. It is a special case of `subgroup_preimage` so I've redefined it as that with no issues.

Alizter:
I also added some more comments and reorganised some things in `Subgroup.v`.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -803,6 +805,78 @@ Proof.
...
+Definition grp_homo_image_in {G H : Group} (f : G $-> H) : G $-> grp_image f.
+Proof.
+  snrapply Build_GroupHomomorphism.
+  - intros x.
+    exists (f x).
+    apply grp_image_in.
+  - intros x y.
+    rapply path_sigma_hprop; simpl.
+    apply grp_homo_op.
+Defined.
```
jdchristensen: I wasn't happy that this couldn't be replaced with the short version using functor_subgroup, so I generalized functor_subgroup to subgroup_corec, which could be used here.  Then it turned out that it could be used in a few other places as well.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -778,7 +820,212 @@ Proof.
...
+(** Corecursion principle for group kernels *)
+Definition grp_kernel_corec {A B G : Group} {f : A $-> B}
+  (g : G $-> A) (h : f $o g == grp_homo_const)
+  : G $-> grp_kernel f.
+Proof.
+  snrapply (subgroup_corec g); exact h.
+Defined.
+
+Definition equiv_grp_kernel_corec `{Funext} {A B G : Group} {f : A $-> B}
+  : (G $-> grp_kernel f) <~> (exists g : G $-> A, f $o g == grp_homo_const).
```
jdchristensen, line 977: This is a special case of a similar result that one could prove for `subgroup_corec`.  But that could be left for a future PR.  If so, maybe leave a TODO comment here?

jdchristensen (approved):
I left one idea for an optional improvement, but feel free to insert a TODO instead.

theories/Algebra/Groups/Subgroup.v
Alizter: Nice!

theories/Algebra/Groups/Subgroup.v
```diff
@@ -220,6 +220,12 @@ Global Instance isembedding_subgroup_incl {G : Group} (H : Subgroup G)
   : IsEmbedding (subgroup_incl H)
   := fun _ => istrunc_equiv_istrunc _ (hfiber_fibration _ _).
 
+(** Restriction of a group homomorphism to a subgroup. *)
+Definition grp_homo_restr {G H : Group} (f : G $-> H) (K : Subgroup G)
+  : subgroup_group K $-> H
+  := f $o subgroup_incl _.
+
+(** Corestriction of a group homomorphism to a subgroup. *)
```
Alizter, line 228: This is the first time I've come across the word "corestriction". Do you know where it commonly used?

theories/Algebra/Groups/Subgroup.v
Alizter: I've done what you've asked. It turns out we don't actually use this lemma anywhere. I don't know if its worth keeping the kernel version around if we have the general corec version.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -237,6 +237,24 @@ Proof.
     snrapply grp_homo_op.
 Defined.
 
+(** Corestriction is an equivalence on group homomorphisms. *)
+Definition equiv_subgroup_corec {F : Funext}
+  (G : Group) {H : Group} (K : Subgroup H)
+  : (exists (f : G $-> H), forall x, K (f x)) <~> (G $-> subgroup_group K).
```
jdchristensen, line 243: Maybe `{ f : G $-> H & ... }` would be a bit clearer for the domain?

jdchristensen (approved):
Just one optional comment.

theories/Algebra/Groups/Subgroup.v
jdchristensen: It's not that common, but a quick search confirms that it is used, e.g. https://en.wikipedia.org/wiki/Corestriction

theories/Algebra/Groups/Subgroup.v
jdchristensen: I could go either way.  That's a pretty fundamental property of the kernel that people might look for, so maybe we should keep it?  (We could reverse one of the two, to get rid of the need for the inverse.)

Alizter: I think reversing the `grp_kernel` result looks better since `grp_corec` is the "forward direction" of the equivalence anyway.

Alizter:
I've just force pushed to resolve the conflicts. The latest commit was new which flipped the direction of the kernel equivalence and made the sigma types easier to see.

PR #2205, opened 2025-01-22
more on commutator subgroups

Alizter:
This PR was originally about some extra results on commutator subgroups but ended up introducing a definitionally involutive opposite groups and various improvements to groups.


theories/Algebra/Groups/Commutator.v
```diff
@@ -187,11 +188,36 @@ Proof.
     by apply subgroup_in_inv.
 Defined.
 
-Definition subgroup_precomp_commutator {G : Group} (H : Subgroup G) (y : G)
+Global Instance issubgroup_precomp_commutator_r {G : Group} (H : Subgroup G)
+  `{!IsNormalSubgroup H} (x : G)
+  : IsSubgroup (fun y => H [x, y]).
```
jdchristensen, line 193: Note that `[y, x] = [x, y]^`, so this also follows from the fact that precomposing `H` with inversion is again a subgroup.  Maybe it would be better to prove that fact about `H o grp_inv`, and use it here?  Alternatively, can we use that subgroup are invariant under taking the "opposite" group, and just using `issubgroup_precomp_commutator_l` here?

theories/Algebra/Groups/Commutator.v
```diff
@@ -233,28 +281,101 @@ Proof.
...
+  intros x.
+  snrapply functor_sigma.
+  - snrapply functor_sigma.
+    + exact (grp_conj y).
+    + exact (fun _ => isnormal_conj _).
+  - intros h.
+    snrapply functor_sigma.
+    + snrapply functor_sigma.
+      * exact (grp_conj y).
+      * exact (fun _ => isnormal_conj _).
```
jdchristensen, line 301: It looks like the same argument is being used twice here.  Is this worth abstracting as a lemma?  I haven't looked to see what it would say.

theories/Algebra/Groups/Commutator.v
```diff
@@ -233,28 +281,101 @@ Proof.
...
+    + intros j.
+      intros p; simpl.
+      lhs_V nrapply (grp_homo_commutator (grp_conj y)).
+      exact (ap (grp_conj y) p).
+Defined.
+
+(** Commutator subgroups distribute over products of normal subgroups on the left. *)
+Definition subgroup_commutator_normal_prod_l {G : Group}
+  (H K L : NormalSubgroup G)
+  : forall x, [subgroup_product H K, L] x <-> subgroup_product [H, L] [K, L] x.
```
jdchristensen, line 311: Does this really require all three to be normal?  Where is normality used in the proof?

theories/Algebra/Groups/Commutator.v
```diff
@@ -233,28 +281,101 @@ Proof.
...
+      apply subgroup_product_incl_l.
+    + revert x KLx.
+      apply functor_subgroup_commutator; trivial.
+      apply subgroup_product_incl_r.
 Defined.
 
+(** Commutator subgroups distribute over products of normal subgroups on the right. *)
+Definition subgroup_commutator_normal_prod_r {G : Group}
+  (H K L : NormalSubgroup G)
+  : forall x, [H, subgroup_product K L] x <-> subgroup_product [H, K] [H, L] x.
```
jdchristensen, line 337: As usual, I really don't like these duplicated proofs.  Can we use the opposite group for this?

theories/Algebra/Groups/Commutator.v
Alizter: When we precompose with a commutator the subgroup needs to be normal for the whole predicate to be a subgroup. We need the target to be a subgroup for the `subgroup_commutator_rec` lemma to be applied.

I got this statement as exercise 5.1.5 in J.S. Robinson's "A Course In the Theory of Groups". There it is stated as all subgroups being normal, but it is not clear if that is necessary.

I can easily come up with counter examples where none of the subgroups are normal, but I haven't been able to come up with one where only one is non-normal.

theories/Algebra/Groups/Commutator.v
Alizter: I went the route of introducing the opposite group since this was something we should have done before.

theories/Algebra/Groups/Commutator.v
Alizter: That was easier than I thought. The key fact is that the commutator of subgroups is (confusingly) commutative.

jdchristensen (commented):
I reviewed the latest commits, but I'll want to make one more pass over the whole thing another day.

theories/Algebra/Groups/Commutator.v
jdchristensen: No need to worry about this, I was just checking if you knew already.  Feel free to Resolve.

jdchristensen:
(The build error seems to be a glitch.)

Alizter:
Coq is undergoing renaming (to rocq) and the docker images have some teething problems it seems. I think this is known upstream, but I'm unsure if it will be fixed or if we have to do something.

theories/Algebra/Groups/Commutator.v
Alizter: I'm genuinely curious and couldn't find an answer so I asked a question on mathoverflow: https://mathoverflow.net/questions/487598/when-does-a-commutator-subgroup-distribute-over-a-product-of-subgroups

theories/Algebra/Groups/Group.v
```diff
@@ -560,6 +560,37 @@ Section GroupMovement.
...
+  - exact (flip (.*.)).
+  - exact 1.
+  - exact (^).
+  - intros x y z.
+    symmetry.
+    exact (grp_assoc z y x).
+  - intro x.
+    exact (grp_unit_r x).
+  - intro x.
+    exact (grp_inv_r x).
```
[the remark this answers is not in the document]
Alizter: Done.

Alizter:
I pushed a little fast with my last commit. I hadn't built the rest of the library by mistake. I will fix this.

Alizter:
I've made groups definitionally involutive. I was in a bit of a rush so some of the naming and choices could be improved. The `equiv_path_group` lemma is very sensitive to changes, and it caused a lot of perf problems. Hopefully these are fixed now, but needs further more careful review.

theories/Algebra/Groups/Group.v
```diff
@@ -165,11 +198,25 @@ Definition Build_Group' (G : Type) `{IsHSet G}
     !LeftInverse (.*.) (^) mon_unit)
```
jdchristensen, line 198: The comment above should also mention that we are dropping the redundant associativity axiom.

jdchristensen:
> I've made groups definitionally involutive.

Great!  I'm about to push a commit that takes advantage of this in one spot.  Not sure if there are other places as well.

> I was in a bit of a rush so some of the naming and choices could be improved.

We now have five constructors for groups:  `Build_Group`, `Build_Group'`, `Build_Group_internal`, `Build_Group_internal_opp`, and `Build_Group_with_opp`.  I can't figure out any pattern to the naming or what `_opp` is supposed to indicate.  And now one or both of the "internal" ones are used in many places.  Can we reduce this to two or three variants with clearer names?

Alizter:
I just pushed a commit clearing up the confusing naming for the constructors. The summary is as follows:

There are 3 constructors
- `Build_Group_internal`, this is the actual record constructor and contains the extra redundant info. This shouldn't really be used outside of `Groups/Group.v`.
- `Build_Group` is the main constructor, it doesn't have the redundant associativity axiom, and still uses `IsGroup`. There is a technical reason for this, which is that a completely unbundled version becomes extremely slow since axioms (with huge proofs) become arguments.
- `Build_Group'` is the "smart" constructor where all the redundancy has been removed. It is used sparingly throughout the library for the reason mentioned above. However it can be convenient for building groups that you don't wish to unfold.

Alizter:
I've pruned some more changes that turned out to be unnecessary. I think this is in a good spot for a more thorough review. I'm interested if the build time has been affected in some files for instance. `AbExt` is a good place to check.

I had some previous experiments with changing around group constructors and they negatively impacted performance which is why I have chosen this setup. Hopefully, we have a balance of clarity, ease-of-use and utility (like with `grp_op`).

Alizter:
Another trick I did that is worth mentioning was to encode the opposite associativity axiom as `Associative (flip op)` and add an instance so that this can be inferred from `Associative op`. Previously, we didn't use typeclasses which meant we couldn't just do `exact _`. This won't affect general use since it is hidden behind `Build_Group`, but it is something we could use for rings also. I'll take a look.

jdchristensen:
Nice!

Can you squash 7 commits into one:  starting from "improve group constructors" and including the 6 "revert ..." commits?  You could also include the "clarify comment for Build_Group'" change if you want.

I'll check things over today or tomorrow.

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -512,18 +512,6 @@ Proof.
...
-  set (a := equiv_sigma_prod0 _ _ a').
-  set (b := equiv_sigma_prod0 _ _ b').
-  clearbody a b; clear a' b'.
-  srapply (contr_equiv _ (ap (equiv_sigma_prod0 _ _)^-1)).
-  srapply (contr_equiv _ (equiv_path_prod _ _)).
-  srapply contr_prod.
-  destruct a as [a' a], b as [b' b].
-  do 3 (nrefine (contr_equiv' _ (@equiv_path_forall H _ _ _ _));
-  nrefine (@contr_forall H _ _ _); intro).
-  exact _.
```
jdchristensen, line 526: I'm curious why this disappeared.  (It's great that it did.)

jdchristensen:
> Can you squash 7 commits into one

Actually, if you can squash "definitionally involutive grp_op" as well, that would mean that files like FreeProduct.v don't get changed at all by this PR, which makes their history cleaner.

theories/Classes/interfaces/abstract_algebra.v
Alizter: Nice surprise from associativity being more obviously a hprop.

Alizter:
@jdchristensen I've squashed the commits related to definitional grp_op into one. THat should be clearer now.

theories/Classes/interfaces/abstract_algebra.v
Alizter: I had a quick look. All of these proofs were written by me 5 years ago and are frankly terrible. They are all quick corollaries of truncation lemmas of their parts. I have pushed a commit that does this cleanup getting rid of whatever this argument was.

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -465,9 +465,14 @@ Class CutMinusSpec A (cm : CutMinus A) `{Zero A} `{Plus A} `{Le A} := {
   cut_minus_0 : forall x y, x ≤ y -> x ∸ y = 0
 }.
 
+Global Instance istrunc_isunitpreserving `{Funext} {n A B} unitA unitB f
+  : IsTrunc n.+1 B -> IsTrunc n (@IsUnitPreserving A B unitA unitB f).
+Proof.
+  unfold IsUnitPreserving; exact _.
+Defined.
```
jdchristensen, line 472: Since the proof here is found by typeclass search, can we avoid needing this lemma at all by making IsUnitPreserving transparent for typeclass search?  Or some similar idea?  (And the same for the many results that use the same idiom.)

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -482,106 +487,29 @@ Definition issig_IsMonoidPreserving {A B : Type} `{SgOp A} `{SgOp B}
...
-  rewrite <- (eisretr (issig_issemigroup x y) a).
-  rewrite <- (eisretr (issig_issemigroup x y) b).
-  set (a' := (issig_issemigroup x y)^-1 a).
-  set (b' := (issig_issemigroup x y)^-1 b).
-  clearbody a' b'; clear a b.
-  srapply (contr_equiv _ (ap (issig_issemigroup x y))).
-Defined.
+Global Instance ishprop_issemigroup `{Funext} x y
+  : IsHProp (@IsSemiGroup x y)
+  := istrunc_equiv_istrunc _ (issig_issemigroup _ _).
```
jdchristensen, line 503: I was just about to do something similar!  Not sure why that proof was so roundabout...

theories/Classes/interfaces/abstract_algebra.v
jdchristensen: Yes, the last commit is great!  BTW, the reason the particular chunk above disappeared is because you added `istrunc_associative`.

theories/Classes/interfaces/abstract_algebra.v
Alizter: I would rather not mark classes as being transparent for typeclass search or else Coq will spend time unfolding classes and trying to search each class it finds. I agree that this is a bit verbose, but I'm not sure if I can make the proof any slicker.

jdchristensen: Is there something like Hint Unfold that will only do it in limited cases?

jdchristensen (commented):
Ok, I've finished a full pass through and it looks good to me.  But I haven't done any timing tests.  Are you able to do some?  If not, I'll try tomorrow.

jdchristensen:
The build error is a glitch.

Alizter:
> Ok, I've finished a full pass through and it looks good to me. But I haven't done any timing tests. Are you able to do some? If not, I'll try tomorrow.

I won't have time until tomorrow evening.

jdchristensen:
I did some timing tests of -j8 builds of four commits:  master, the HEAD of this PR, and two in between.  All took the same time within the noise.

I also did a -j1 file-by-file comparison of the HEAD of this PR vs master and got the following:
```
   After |  Peak Mem | File Name                                     |   Before |  Peak Mem ||    Change || Change (mem) | % Change | % Change (mem)
0m50.20s | 600588 ko | Total Time / Peak Mem                         | 0m48.40s | 600752 ko || +0m01.80s ||      -164 ko |   +3.73% |         -0.02%
0m00.20s | 386344 ko | Classes/interfaces/abstract_algebra.vo        | 0m00.59s | 390148 ko || -0m00.38s ||     -3804 ko |  -66.10% |         -0.97%
0m01.60s | 468968 ko | Algebra/AbSES/Pushout.vo                      | 0m01.43s | 469736 ko || +0m00.17s ||      -768 ko |  +11.88% |         -0.16%
0m00.63s | 450480 ko | Algebra/AbGroups/AbPushout.vo                 | 0m00.46s | 450056 ko || +0m00.17s ||       424 ko |  +36.95% |         +0.09%
0m01.01s | 470772 ko | Homotopy/HomotopyGroup.vo                     | 0m00.84s | 469752 ko || +0m00.17s ||      1020 ko |  +20.23% |         +0.21%
0m01.37s | 463256 ko | Algebra/AbGroups/TensorProduct.vo             | 0m01.22s | 462648 ko || +0m00.15s ||       608 ko |  +12.29% |         +0.13%
0m00.96s | 482056 ko | Homotopy/PinSn.vo                             | 0m00.82s | 482188 ko || +0m00.14s ||      -132 ko |  +17.07% |         -0.02%
0m00.86s | 440892 ko | Algebra/Groups/Group.vo                       | 0m00.74s | 441348 ko || +0m00.12s ||      -456 ko |  +16.21% |         -0.10%
0m03.12s | 464720 ko | Classes/implementations/natpair_integers.vo   | 0m03.00s | 462544 ko || +0m00.12s ||      2176 ko |   +4.00% |         +0.47%
0m01.43s | 465968 ko | Algebra/AbSES/Core.vo                         | 0m01.32s | 466608 ko || +0m00.10s ||      -640 ko |   +8.33% |         -0.13%
0m01.24s | 466256 ko | Algebra/AbSES/Pullback.vo                     | 0m01.16s | 464852 ko || +0m00.08s ||      1404 ko |   +6.89% |         +0.30%
0m00.07s | 240056 ko | test/Algebra/Groups/Group.vo                  |   N/A    |    N/A    || +0m00.07s ||    240056 ko |        ∞ |              ∞
0m01.82s | 469096 ko | Algebra/Groups/FreeProduct.vo                 | 0m01.76s | 469004 ko || +0m00.06s ||        92 ko |   +3.40% |         +0.01%
0m00.75s | 445940 ko | Algebra/Groups/Subgroup.vo                    | 0m00.68s | 446096 ko || +0m00.06s ||      -156 ko |  +10.29% |         -0.03%
0m00.55s | 446304 ko | Algebra/Groups/QuotientGroup.vo               | 0m00.49s | 446404 ko || +0m00.06s ||      -100 ko |  +12.24% |         -0.02%
0m00.62s | 478516 ko | Algebra/AbSES/SixTerm.vo                      | 0m00.56s | 477876 ko || +0m00.05s ||       640 ko |  +10.71% |         +0.13%
0m01.11s | 483644 ko | Homotopy/ClassifyingSpace.vo                  | 0m01.06s | 484028 ko || +0m00.05s ||      -384 ko |   +4.71% |         -0.07%
0m00.42s | 453940 ko | Algebra/Groups/Commutator.vo                  | 0m00.37s | 454032 ko || +0m00.04s ||       -92 ko |  +13.51% |         -0.02%
0m00.41s | 439256 ko | Algebra/Groups/GrpPullback.vo                 | 0m00.36s | 439420 ko || +0m00.04s ||      -164 ko |  +13.88% |         -0.03%
0m01.40s | 472820 ko | Algebra/AbSES/PullbackFiberSequence.vo        | 0m01.35s | 472696 ko || +0m00.04s ||       124 ko |   +3.70% |         +0.02%
```
That indicates a general trend to be slightly slower with this PR, but nothing major.  So I'm ok with merging this.

jdchristensen:
Actually, taking a quick look at the timing output for AbPushout.v showed two lines that got slower.  I'll indicate them and any others I find.

jdchristensen:
Besides maybe fixing a few slow lines, there's nothing else I can think of adding now.

jdchristensen:
(In case it isn't clear, I edited a previous comment to give details of slow lines.)

Alizter:
I'm not fully certain why those lines in AbPushout are slow. I tried rewriting some of the proofs but it seemed to make things worse. I'll see if I can have a better look soon.

jdchristensen:
> * Line 136 (`refine (_ o (path_ab_pushout f g (grp_prod_inl c0) (grp_prod_inl c1))^-1).`) went from 0.08 to 0.12s.

My last commit gets this down to ~0.09s.  But if I make the same change in master, that line only takes 0.04s, so there is still a discrepancy.  Not a big deal, but it must be small slowdowns like this that are showing up in the -j1 results.

jdchristensen:
The slowest line in `freeproduct_ind_homotopy` is just as slow in master, so it isn't because of this PR.  I can't figure out why it is slow, as it looks like a trivial step.  I made a minor improvement, but it's still slow, and nothing else I tried helped.  In any case, I think this PR is good and can be merged when you are ready.

Alizter:
As you said before, the slowdown is within the realm of acceptability so its probably not worth the effort of splitting hairs.

PR #2209, opened 2025-01-29
Adapt to coq/coq#20150 (more aggressive universe minimization)

SkySkimmer:


Alizter:
I don't understand the reason for the changes in Idempotents.v and also they are not backwards compatible. Where does the failure occur? Maybe we can do some backwards-compatible workaround there.

jdchristensen:
coq/coq#20150 was just closed, so I'll close this too.  Feel free to reopen if I am misunderstanding.

PR #2210, opened 2025-01-31
Sequences, uniform structures

thchatzidiamantis:
Work on types of sequences `[nat -> X]`, will be used for formalising searchable types. Some definitions are stated in the generality of `nat`-graded equivalence relations, which we call `UStructures` here. This could be further generalised to directed systems or uniform spaces in the topological sense later.

### DIFF
--- a/theories/Misc/UStructures.v
+++ b/theories/Misc/UStructures.v
@@ -1,0 +1,81 @@
+(** * Uniform Structures *)
+
+Require Import Basics Types.
+Require Import Truncations.Core.
+Require Import Spaces.Nat.Core.
+
+Open Scope nat_scope.
+
+(** ** [nat]-graded uniform structures *)
+
+(** A uniform structure on a type consists of an equivalence relation for every natural number, each one being stronger than its predecessor. *)
+Class UStructure (us_type : Type) := {
+  us_rel : nat -> Relation us_type;
+  us_reflexive : forall n : nat, Reflexive (us_rel n);
+  us_symmetric : forall n : nat, Symmetric (us_rel n);
+  us_transitive : forall n : nat, Transitive (us_rel n);
+  us_pred : forall (n : nat) (x y : us_type), us_rel n.+1 x y -> us_rel n x y
+}.
+
+Existing Instances us_reflexive us_symmetric us_transitive.
+
+Notation "u =[ n ] v" := (us_rel n u v) (at level 70).
+
+Definition us_rel_leq {X : Type} {struct : UStructure X}
+  {m n : nat} (hm : m <= n) {u v : X} (h : u =[n] v)
+  : u =[m] v.
+Proof.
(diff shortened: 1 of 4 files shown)


theories/Misc/UStructures.v
```diff
@@ -0,0 +1,81 @@
...
+  us_rel : nat -> Relation us_type;
+  us_reflexive : forall n : nat, Reflexive (us_rel n);
+  us_symmetric : forall n : nat, Symmetric (us_rel n);
+  us_transitive : forall n : nat, Transitive (us_rel n);
+  us_pred : forall (n : nat) (x y : us_type), us_rel n.+1 x y -> us_rel n x y
+}.
+
+Existing Instances us_reflexive us_symmetric us_transitive.
+
+Notation "u =[ n ] v" := (us_rel n u v) (at level 70).
```
Alizter, line 22: Try this so that the spacing is as you expect it:
```coq
Notation "u =[ n ] v" := (us_rel n u v)
  (at level 70, format "u  =[ n ]  v").
```

theories/Spaces/NatSeq/Core.v
```diff
@@ -0,0 +1,75 @@
+(** * Types of Sequences [nat -> X] *)
+
+Require Import Basics Types.
+Require Import Spaces.Nat.Core.
+
+Open Scope nat_scope.
+Open Scope type_scope.
```
Alizter, line 7: Local

theories/Spaces/NatSeq/UStructure.v
```diff
@@ -0,0 +1,81 @@
+(** * Uniform structure on types of sequences *)
+
+Require Import Basics Types.
+Require Import Spaces.Nat.Core.
+Require Import Misc.UStructures.
+Require Import List.Core List.Theory.
+Require Import Spaces.NatSeq.Core.
+
+Open Scope nat_scope.
+Open Scope type_scope.
```
Alizter, line 10: Local

Alizter (commented):
I'm not certain if `Misc/` is the best place to put this, but I can't think of a better place. I've left some comments. I don't know anything about uniform spaces, but from some quick reading I see that they aim to axiomatize uniform convergence. I trust that the theory here is correct, but I can only really comment on formalisation style.

theories/Misc/UStructures.v
Alizter: In the `format` modifier: 1 space is flush and 2 spaces is a space.

jdchristensen:
I've already reviewed this in a PR to my repo, so I approve of this version.  I think using a Misc folder is fine, and that some of the things mentioned in #2155 should be moved there as well (in a separate PR).

Alizter (approved):

Alizter:
Feel free to merge when ready.

PR #2211, opened 2025-02-09
connectivity of wedge inclusion

Alizter:
In this PR we show that the wedge connectivity lemma actually says something about wedges.

We show that the inclusion map from a wedge to a product has the expected connectivity. The wedge connectivity lemma could be adjusted to make it easier to apply, since it was a bit awkward.


Alizter:
I will rebase this after #2253.

theories/Homotopy/Wedge.v
```diff
@@ -370,3 +420,63 @@ Proof.
   - reflexivity.
 Defined.
 
+(** ** Connectivity of wedge inclusion *)
+
+Definition conn_map_wedge_incl `{Univalence} {m n : trunc_index} (X Y : pType)
```
Alizter, line 425: Looking at this again, it's a little long. I'll see if I can try to break up the proof.

theories/Homotopy/Wedge.v
```diff
@@ -370,3 +412,50 @@ Proof.
...
+Defined.
+
+Definition conn_map_wedge_incl `{Univalence} {m n : trunc_index} (X Y : pType)
+  `{!IsConnected m.+1 X, !IsConnected n.+1 Y}
+  : IsConnMap (m +2+ n) (wedge_incl X Y).
+Proof.
+  apply conn_map_from_extension_elim.
+  intros P h d.
+  snapply extension_wedge_incl.
+  - rapply (wedge_incl_elim _ _ (fun x y => P (x, y)) (d o wedge_inr) (d o wedge_inl)).
```
jdchristensen, line 455: Now the x and y are in the same order, so we don't need to commute the addition.

theories/Homotopy/Wedge.v
```diff
@@ -370,3 +412,50 @@ Proof.
   - reflexivity.
 Defined.
 
+(** ** Connectivity of wedge inclusion *)
+
+Definition extension_wedge_incl {X Y : pType} (P : X * Y -> Type)
+  (d : forall a : X \/ Y, P (wedge_incl X Y a))
+  (f : forall (x : X) (y : Y), P (x, y))
+  (p : forall (x : X), f x pt = d (wedge_inl x))
+  (q : forall (y : Y), f pt y = d (wedge_inr y))
```
jdchristensen, line 421: These two lines were definitionally simplified.

theories/Homotopy/Wedge.v
```diff
@@ -370,3 +412,50 @@ Proof.
...
+                 @ transport_compose_path_ap P (wedge_incl X Y) (wedge_incl_beta_wglue) _))
+  : ExtensionAlong (wedge_incl X Y) P d.
+Proof.
+  exists (prod_ind _ f).
+  napply (wedge_ind _ p q).
+  lhs napply (transport_paths_FlFr_D wglue).
+  apply moveR_pM.
+  apply moveR_Vp.
+  unshelve lhs napply ap_homotopic_id.
+  1: exact (transport_compose_path_ap P (wedge_incl X Y) wedge_incl_beta_wglue).
```
jdchristensen, line 434: This new lemma just makes things a bit easier to read, which led me to a slightly shorter proof.

jdchristensen (approved):
@Alizter I'll let you review my changes and merge/adjust as needed.

theories/Homotopy/Wedge.v
```diff
@@ -3,6 +3,8 @@ Require Import Pointed.Core Pointed.pSusp.
 Require Import Colimits.Pushout.
 Require Import WildCat.
 Require Import Homotopy.Suspension.
+Require Import Truncations.Core Truncations.Connectedness.
+Require Import Extensions Modalities.ReflectiveSubuniverse.
 
 Local Set Universe Minimization ToSet.
```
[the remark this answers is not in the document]
jdchristensen: While I approved, this needs to be dealt with before merging.  Either the minimization should be removed, or universe annotations need to be added to at least the last result in the file.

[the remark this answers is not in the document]
Alizter: Without this, `Wedge` has 4 universe levels rather than 3 and the number of universe levels in the file quickly gets out of hand.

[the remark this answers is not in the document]
Alizter: I removed the line and annotated the definition manually. It seems to have sensible universe variables in later definitions, though not always. For now I think this is good enough. If we annotate too heavily we are sure to run into version differences with Coq.

Alizter:
I'll let you merge after you've had another look. I am happy with it.

jdchristensen:
I noticed that the use of wildcat notations like `$->` was (as usual) causing there to be an extra universe variable for the wildcat type and was also constraining many of the types to land in the same universe.  And I didn't see anywhere that it was being used in a meaningful way.  So I removed it, and the universe variables improved.  I also made a few other minor changes.  So I'll hand it back to you to review and merge if you agree with these changes.

PR #2212, opened 2025-02-09
ci: fix permission issues in dev

Alizter:
This is the recommended fix in https://github.com/coq-community/docker-coq-action/releases/tag/v1.5.1


Alizter:
Nice, that seems to have worked as intended.

PR #2217, opened 2025-02-18
surjective_factor: cleanup

jdchristensen:
I happened to glance at this file the other day and cleaned up some of it.  But probably this material should be moved elsewhere.  Not sure why it is in the HIT folder.  It's not used anywhere, but if we want to keep it, I suggest we add it into Truncations/Core.v in the section about surjections.


Alizter:
I would suggest moving it to Truncations/Core like ypu mention.

theories/HIT/surjective_factor.v
```diff
@@ -1,36 +1,32 @@
...
-        HoTT.Basics
-        HoTT.Truncations.Core Modalities.Modality.
+Require Import Basics Types.Sigma Truncations.Core Modalities.Modality.
 
-(** Definition by factoring through a surjection. *)
+(** * Definition of a map by factoring through a surjection. *)
 
 Section surjective_factor.
-  Context `{Funext}.
+  Context `{Funext}. (* Can [Funext] be avoided? *)
```
Alizter, line 6: Last I looked at this file, my feeling was yes. I think we have to tweak a few definitions however. At some point it is required, but not for most things here.

theories/Truncations/Core.v
```diff
@@ -265,6 +265,36 @@ Proof.
   exact (tr (s y; h y)).
 Defined.
 
+(** Definition of a map by factoring through a surjection. *)
+Section surjective_factor.
+  Context `{Funext}. (* Can [Funext] be avoided? *)
```
Alizter, line 270: I can avoid funext if `C` is a hprop rather than a hset.
```coq
    unshelve epose ((fst (extendable_conn_map_inO (-1) 1 g (fun _ => C))) f).
    { simpl.
      admit. }
    destruct e as [r p].
    intros b.
    exists (r b).
    intros a q.
    rhs_V nrapply p.
    apply ap.
    exact q.
```
I feel like there is something we can do with exentable maps here but I haven't been able to work out how.

jdchristensen: I found a way, using the material in Constant.v

jdchristensen: Because of the dependencies, I put this in Constant.v.  Let me know if you know a better spot.

Alizter:
That argument cleaned up nicely! I think its fine for this to be in `Constant.v` since it is not used anywhere.

I have some minor style tweaks that I can push if you are okay with that.

jdchristensen:
> I have some minor style tweaks that I can push if you are okay with that.

Yes, of course!

jdchristensen:
@Alizter I have some style cleanups of my own.  Can I push them, or would that conflict with what you are doing?

Alizter:
@jdchristensen I've already pushed, so we will see :)

jdchristensen:
@Alizter Oh, you already pushed!  I didn't get the notification.  I'll add my changes in.

jdchristensen:
Mine were all different from yours (and yours were good too), so I think this is ready to go.

Alizter:
Great, feel free to merge when you are ready.

Alizter (approved):

PR #2219, opened 2025-02-18
HoTT.v: add (some) missing Exports

jdchristensen:
I noticed a few missing Exports in HoTT.v.  I didn't systematically check for others.


Alizter (approved):

Alizter:
I wonder if its time for a script to generate this list together with a list of excluded items.

jdchristensen:
> I wonder if its time for a script to generate this list together with a list of excluded items.

Sounds like a good idea.  The same script could generate the other index files we have (like Basics.v), since the knowledge of which of those to generate is essentially the same as the knowledge of what to exclude from HoTT.v.

PR #2221, opened 2025-02-20
fix: change Warnings Append -> Warnings

Alizter:
This should fix some of the recent warnings we have been seeing.

If I've understood correctly, there should be no change in semantics.


jdchristensen (approved):
LGTM if it works.

PR #2222, opened 2025-02-20
move Constant.v to Truncations/

Alizter:
I believe that this makes the most logical sense as a home. The main utility of `Constant.v` is to eliminate a hprop to a hset, therefore I think it makes sense to live with the rest of `Truncations/`.


jdchristensen (approved):
Seems like a reasonable choice!

PR #2225, opened 2025-02-23
Strong_cat into own submodule

patrick-nicodemus:
Currently in the codebase there is a known typeclass loop caused by two hints, `Is1Cat_Strong -> Is1Cat` and `Is1Cat + HasMorExt -> Is1Cat_Strong`. 

The typeclass loop is currently ameliorated by setting one of the hints to a low priority, so that the loop is never activated if there are other valid solutions.

Not all of the development environments for Coq seem able to easily interrupt such a loop. For VsCoq I have resorted to keeping a terminal window open so I can send the `kill` command when this happens. Typeclass resolution failures are very common so this really impedes my workflow.

In this PR I move all the Strong 1Cat functionality into its own submodule `Strong` within `Core`. In order for the associated typeclass hints to be used in a file, one must write `Import WildCat.Core.Strong`. If one only wants to use a lemma or a definition in the file, one can write `Strong.lemma_name` rather than importing it globally and writing `lemma_name`.

In some cases where there is a file `A.v` downstream which only has one or two lemmas involving `Strong`, I have put those lemmas into a subsection or submodule and imported the `Strong` module only within that submodule so that the loop is not imported globally into the file.

### DIFF
--- a/theories/Algebra/AbSES/Pullback.v
+++ b/theories/Algebra/AbSES/Pullback.v
@@ -444,7 +444,7 @@ Proof.
   nrefine (_ @ (ap (fun x => x $@ _)) _).
   2: { refine (_ @ ap _ equiv_path_absesV_1^).
        exact (fmap_id_strong _ _)^. }
-  exact (cat_idr_strong _)^.
+  exact (Strong.cat_idr_strong _)^.
 Defined.
 
 Definition equiv_hfiber_abses_pullback `{Univalence} {A B B' : AbGroup} {f : B' $-> B}
--- a/theories/Algebra/AbSES/Pullback.v
+++ b/theories/Algebra/AbSES/Pullback.v
@@ -1,6 +1,7 @@
 Require Import Basics Types.
 Require Import HSet Limits.Pullback.
 Require Import WildCat Pointed.Core Homotopy.ExactSequence.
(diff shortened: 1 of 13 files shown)


jdchristensen:
> Currently in the codebase there is a known typeclass loop caused by two hints, `Is1Cat_Strong -> Is1Cat` and `Is1Cat + HasMorExt -> Is1Cat_Strong`.

Our policy is to not have loops at all, so your solution of isolating some instances into modules isn't good.  As soon as the module is imported, the loop would exist.  Instead, when we have a potential loop, we have to decide on a place to break it, often using Hint Immediate for that so it can still automatically be applied in some cases.  See #2228 for what I mean.  Does it fix the problem?

(Note that I mentioned this already in #2208:

> And thanks for pointing out that loop in typeclass search. We try to avoid any possibilities of loops. The library builds if we use Hint Immediate is1cat_is1cat_strong : typeclass_instances. instead of making it an instance, so we should do that I think. This will probably fix the loop you found, although I didn't check.

)

patrick-nicodemus:
Thanks, I just forgot about your comment. I think that the previous PR I suggested involving Hint Mode should highlight the changes, so I'll switch to that PR and see if this breaks those cycles.

PR #2228, opened 2025-02-23
WildCat/Core: reduce is1cat_is1cat_strong to a hint to avoid loops

jdchristensen:
@patrick-nicodemus Can you check whether this gets rids of the loop in typeclass search that you mentioned?

### DIFF
--- a/theories/WildCat/Core.v
+++ b/theories/WildCat/Core.v
@@ -197,8 +197,8 @@ Arguments cat_assoc_opp_strong {_ _ _ _ _ _ _ _ _} f g h.
 Arguments cat_idl_strong {_ _ _ _ _ _ _} f.
 Arguments cat_idr_strong {_ _ _ _ _ _ _} f.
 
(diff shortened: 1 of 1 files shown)


Alizter:
Do we even use this instance anywhere? I am not able to check myself at the moment.

jdchristensen:
> Do we even use this instance anywhere? I am not able to check myself at the moment.

If I remove the Hint Immediate, the build fails in WildCat/Opposite.v.  But Hint Immediate is enough to make everything build.

Alizter:
This seems sensible to me. @patrick-nicodemus Can you check this fixes your issue.

patrick-nicodemus:
To be honest this didn't solve my problem, so either there's something else going on or I misdiagnosed the problem. Still, I think it is a loop so I would recommend merging it.

jdchristensen:
Ok, I'll go ahead and merge.

PR #2232, opened 2025-02-25
more transport_paths lemmas

Alizter:


theories/Types/Paths.v
```diff
@@ -162,6 +180,30 @@ Proof.
...
+  {f : A -> B} {g : B -> C} {h : C -> A} {x1 x2 : A}
+  (p : x1 = x2) (q : h (g (f x1)) = x1) (r : h (g (f x2)) = x2)
+  (h' : ap h (ap g (ap f p)) @ r = q @ p)
+  : transport (fun x => h (g (f x)) = x) p q = r.
+Proof.
+  refine (transport_paths_FFFlr (h:=h) _ _ @ _).
+  refine (concat_pp_p _ _ _ @ _).
+  apply moveR_Vp.
+  exact h'^.
+Defined.
```
jdchristensen, line 193: I think we should replace all of the primed versions with a tactic that takes which lemma to apply as an argument.  Also, if we added parentheses to the statements of the unprimed lemmas, then we wouldn't need to apply `concat_pp_p` in the primed versions.  Not sure if that would complicated other uses of the unprimed lemmas.

Alizter: I've gone ahead and done that. We now have tactics `transport_forall FlFr` for every combination.

Note sure about the `concat_pp_p` comment yet however.

theories/Types/Paths.v
```diff
@@ -128,74 +154,36 @@ Proof.
...
-  refine (transport_paths_FlFr _ _ @ _).
-  refine (concat_pp_p _ _ _ @ _).
-  apply moveR_Vp.
-  exact h^.
-Defined.
+Tactic Notation "transport_paths" uconstr(lemma) :=
+  lhs nrapply lemma;
+  lhs nrapply concat_pp_p;
+  nrapply moveR_Vp;
+  nrefine (_^).
```
jdchristensen, line 163: Should the last line just be `symmetry.`?

theories/Types/Paths.v
```diff
@@ -128,74 +154,36 @@ Proof.
...
-  (p : x1 = x2) (q : g (f x1) = x1) (r : g (f x2) = x2)
-  (h : (ap g (ap f p)) @ r = q @ p)
-  : transport (fun x => g (f x) = x) p q = r.
-Proof.
-  refine (transport_paths_FFlr _ _ @ _).
-  refine (concat_pp_p _ _ _ @ _).
-  apply moveR_Vp.
-  exact h^.
-Defined.
+Tactic Notation "transport_paths" "Flr" := transport_paths transport_paths_Flr.
```
jdchristensen, line 165: In Ltac, can we take a string `s` as an argument and build an identifier by doing something like `concat "transport_paths_" s`?

theories/Types/Paths.v
```diff
@@ -128,74 +154,36 @@ Proof.
...
+Tactic Notation "transport_paths" "FFlFFr" := transport_paths transport_paths_FFlFFr.
+
+(** Sometimes we are required to reason about the term produced by [transport_paths FlFr] so we give it a name. It is local so to discourage its use. *)
+Local Definition transport_paths_FlFr' {A B : Type} {f g : A -> B} {x1 x2 : A}
+  (p : x1 = x2) (q : f x1 = g x1) (r : (f x2) = (g x2))
+  (h : (ap f p) @ r = q @ (ap g p))
+  : transport (fun x => f x = g x) p q = r.
+Proof.
+  by transport_paths transport_paths_FlFr.
 Defined.
```
jdchristensen, line 187: I think this should be moved to TriJoin.v.

theories/Types/Paths.v
Alizter: Maybe in ltac2, but I wasn't able to work out how to do it in ltac1.

theories/Types/Paths.v
jdchristensen: I see that the `transport_paths_*` lemmas are still used a ton in the library, so it would be way too much work to change their definition now.  So I think it's best to ignore my comment about `concat_pp_p`.

theories/Types/Paths.v
jdchristensen: Actually, to keep the proof terms smaller, the last three steps of the tactic can be made into a lemma:
```coq
Definition moveR_Vpp_inv {A : Type} {w x y z : A}
  (p : x = w) (q : x = y) (r : y = z) (s : w = z)
  (h : p @ s = q @ r)
  : p^ @ q @ r = s.
Proof.
  lhs nrapply concat_pp_p.
  apply moveR_Vp, h^.
Defined.

Tactic Notation "transport_paths" uconstr(lemma) :=
  lhs nrapply lemma; apply moveR_Vpp_inv.
```

At this point, it's not clear if the tactic is actually worth it.  We could just inline its usage, which would then be more clear to the reader.  I could go either way.

theories/Types/Paths.v
Alizter: It's not supported in ltac1. Not a big deal IMO.

theories/Types/Paths.v
Alizter: That looks like a useful lemma to introduce. It doesn't seem to quite fit into the naming scheme in PathGroupoids however. Any suggestions?

Alizter: I think having these custom tactics around will only confuse new readers, so I think its best to replace it.

jdchristensen: About the name, the `Vpp` part should be `Vp_p` I think, to indicate the parentheses.  I'm not sure about `_inv`.  Would `_symm` be better?  It's a bit special to this situation that we reverse the equality, but it's because in practice our computation rules go from left to right, so most of the time after applying `transport_paths_*` we want to reverse the equality.  About moving it to PathGroupoids, I could go either way.  It's pretty special, so it probably ok to leave it near where it is used, but it might also have other uses, in which case PathGroupoids.v would be better.

BTW, I'm travelling soon, so I'll mostly be out of the loop for around 48 hours or more.

theories/Types/Paths.v
```diff
@@ -128,75 +154,36 @@ Proof.
...
-  apply moveR_Vp.
-  exact h^.
-Defined.
+(** The above lemmas have some common rearrangements that are useful. Since these all follow the same pattern, we introduce a tactic to apply it. *)
 
-Definition transport_paths_FlFr' {A B : Type} {f g : A -> B} {x1 x2 : A}
-  (p : x1 = x2) (q : f x1 = g x1) (r : (f x2) = (g x2))
-  (h : (ap f p) @ r = q @ (ap g p))
-  : transport (fun x => f x = g x) p q = r.
+(** TODO: move to PathGroupoids and rename? *)
```
jdchristensen, line 159: About this comment, let's just decide: keep the lemma here or move it?  I'd be inclined to just leave it here and remove the comment.

Alizter:
```suggestion
```

jdchristensen (approved):

Issue #2202, opened 2025-01-17
introduce subgroup inclusions and equality


Alizter:
We should introduce subgroup inclusions and subgroup equality (based on iff) like we have done for ideals. In fact, we could probably just move the ideal versions to `Subgroup.v`. It is used in quite a few places implicitly already, so we will have to go through and change them where it makes sense to do so.

PR #2233, opened 2025-02-25
subgroups of direct products

Alizter:
Here was some miscellaneous stuff about subgroups of products that I didn't end up using.

### DIFF
--- a/theories/Algebra/Groups/Subgroup.v
+++ b/theories/Algebra/Groups/Subgroup.v
@@ -1183,3 +1183,46 @@ Proof.
   apply equiv_iff_hprop_uncurried.
   split; exact _.
 Defined.
+
+(** ** Subgroups of direct products *)
+
+Definition subgroup_grp_prod_l {G H : Group} (K : Subgroup G)
+  : Subgroup (grp_prod G H)
+  := subgroup_image grp_prod_inl K.
+
+Definition subgroup_grp_prod_r {G H : Group} (K : Subgroup H)
+  : Subgroup (grp_prod G H)
+  := subgroup_image grp_prod_inr K.
+
+(** These two subgroups have trivial intersection. *)
+Instance istrivial_intersection_subgroup_grp_prod {G H : Group}
+  (J : Subgroup G) (K : Subgroup H)
+  : IsTrivialGroup
+    (subgroup_intersection (subgroup_grp_prod_l J) (subgroup_grp_prod_r K)).
+Proof.
+  intros [x y] [Jx Ky].
+  strip_truncations.
+  destruct Jx as [j p], Ky as [k q].
+  apply path_prod.
+  - exact (ap fst q^).
+  - exact (ap snd p^).
+Defined.
+
+(** TODO: use [pred_eq] *)
+(** The maximal subgroup of a direct product may be decomposed as the subgroup product of the factor subgroups. *)
+Definition subgroup_eq_grp_prod_maximal {G H : Group}
+  : forall x, maximal_subgroup (grp_prod G H) x
+    <-> subgroup_product
+          (subgroup_grp_prod_l (maximal_subgroup G))
+          (subgroup_grp_prod_r (maximal_subgroup H)) x.
+Proof.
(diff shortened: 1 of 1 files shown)


theories/Algebra/Groups/Subgroup.v
```diff
@@ -1183,3 +1183,46 @@ Proof.
   apply equiv_iff_hprop_uncurried.
   split; exact _.
 Defined.
+
+(** ** Subgroups of direct products *)
+
+Definition subgroup_grp_prod_l {G H : Group} (K : Subgroup G)
+  : Subgroup (grp_prod G H)
+  := subgroup_image grp_prod_inl K.
```
jdchristensen, line 1191: This definition of the induced subgroup of G x H is concise, but in some ways is more complicated than necessary.  First, `subgroup_image` is defined in terms of `grp_image`, which adds a truncation, but because `grp_prod_inl` is an embedding, the truncation is not needed.  So maybe we should have a `subgroup_image_embedding` that uses `grp_image_embedding`, and then use that here?

Second, `grp_image_embedding` adds a sigma type (via `hfiber`), but in this case even that is not necessary.  The predicate for the induced subgroup can be defined to send `(g, h)` to `K g * (h = 1)`.  But then we'd have to prove that this does define a subgroup, which probably makes it not worthwhile.

Neither of these are very important, but I thought I'd mention them.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -1183,3 +1183,46 @@ Proof.
...
+Proof.
+  intros [x y] [Jx Ky].
+  strip_truncations.
+  destruct Jx as [j p], Ky as [k q].
+  apply path_prod.
+  - exact (ap fst q^).
+  - exact (ap snd p^).
+Defined.
+
+(** TODO: use [pred_eq] *)
```
jdchristensen, line 1211: What is `pred_eq`?  Is this referring to something we discussed earlier?

theories/Algebra/Groups/Subgroup.v
```diff
@@ -1183,3 +1183,46 @@ Proof.
...
+(** The maximal subgroup of a direct product may be decomposed as the subgroup product of the factor subgroups. *)
+Definition subgroup_eq_grp_prod_maximal {G H : Group}
+  : forall x, maximal_subgroup (grp_prod G H) x
+    <-> subgroup_product
+          (subgroup_grp_prod_l (maximal_subgroup G))
+          (subgroup_grp_prod_r (maximal_subgroup H)) x.
```
jdchristensen, line 1217: Another way to state the conclusion is
```coq
forall x, subgroup_product
            (subgroup_grp_prod_l (maximal_subgroup G))
            (subgroup_grp_prod_r (maximal_subgroup H)) x.
```
but maybe your way is better?

Independently, the two groups on the right can be simplified to `(grp_image_embedding grp_prod_inl)` and `(grp_image_embedding grp_prod_inr)`, which avoids mentioning maximal groups, avoids `subgroup_grp_prod_l` and `_r`, and avoids the truncations.  If your goal was this result, then using these versions of these groups would be a shorter way to get here.

theories/Algebra/Groups/Subgroup.v
Alizter: Predicate equality which is introduced in #2121 addressing #2202.

jdchristensen: Ah, great.  I'll try to get to 2121 soon.

theories/Algebra/Groups/Subgroup.v
jdchristensen: @Alizter Did you have an opinion about the two possible changes to the statement of this result?  The first is that `maximal_subgroup ...` just simplifies to `Unit`, so it can be avoided.  The second is that the two subgroups on the RHS can be simplified a bit using `grp_image_embedding`.  In both cases, what is the best choice depends a bit on how this might be used, so I'm flexible.

Issue #2230, opened 2025-02-24
Generate timing reports with dune


Alizter:
Dune doesn't know about the .timing files generated by Coq with `-time`.

We do however have a makefile, so I could imagine a sandboxed rule that copies all the .v files and runs `make timings` on them to generate the timing files and move them to a special directory target. It would be a bit hacky, but cleaner than running make yourself.

I'm creating this issue as a reminder for trying this.

PR #2234, opened 2025-02-25
dune: enable directory targets

Alizter:
This PR adds support for building timing-html files as mentioned in #2230.

### TODO
- [ ] Also put .timing files somewhere, for now only .timing.html is built
- [ ] Generate report of slowest lines
- [ ] Generate comparison report from last report

### DIFF
--- a/dune
+++ b/dune
@@ -129,3 +129,35 @@
   (file file_to_bench))
  (action
   (run echo "Starting bench. This may take a while.")))
+
+; Timing report
+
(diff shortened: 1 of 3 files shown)


jdchristensen:
This looks like a nice idea.

Alizter:
I've added a way for dune to generate the .timing files. I'll need to now work out a way to compile this into a report about slowest lines.

PR #2235, opened 2025-02-25
style md  add section on bullets

Alizter:


jdchristensen (commented):
I think this looks good.  I'm not in favour of using braces to restart the bullets, though, and I don't remember seeing that anywhere.  I'd prefer doubled bullets if the nesting gets that deep.  I would keep the part about using lemmas in such situations, but delete the part about braces and simplify the example.  I would also give an example of the `1: tactic` method to avoid bullets when there are, say, two subgoals and one has a short proof.

PR #2238, opened 2025-02-28
Bundled categories

patrick-nicodemus:
Introduces a bundled type of Graphs and a bundled type of Categories. 
Proves that Graph is a (0,1)-category and a 2-graph.
Proves that Category is a (0,1)-category and a 3-graph with "wild modifications" as 3-cells.

### DIFF
--- a/theories/Colimits/CoeqUnivProp.v
+++ b/theories/Colimits/CoeqUnivProp.v
@@ -15,6 +15,8 @@ Require Import WildCat.Paths.
 Require Import WildCat.FunctorCat.
 Require Import WildCat.ZeroGroupoid.
 
+(* Successful typeclass inference in this file is sensitive to the order of imports; in particular FunctorCat should not be the last import. *)
+
 (** Using wild 0-groupoids, the universal property can be proven without funext. A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
 Section UnivProp.
   Context {B A : Type} (f g : B -> A) (P : Coeq f g -> Type).
--- a/theories/Colimits/CoeqUnivProp.v
+++ b/theories/Colimits/CoeqUnivProp.v
@@ -107,7 +107,7 @@ Section UnivProp.
       $<~> Build_ZeroGpd (sig Coeq_ind_data) _ _ _.
   Proof.
     snrapply Build_CatEquiv.
-    1: rapply Build_Morphism_0Gpd.
+    1: rapply FunctorCat.Build_Fun01.
     rapply isequiv_0gpd_issurjinj.
   Defined.
 
--- a/theories/Colimits/CoeqUnivProp.v
+++ b/theories/Colimits/CoeqUnivProp.v
@@ -12,6 +12,7 @@ Require Import WildCat.EquivGpd.
 Require Import WildCat.Forall.
 Require Import WildCat.NatTrans.
 Require Import WildCat.Paths.
+Require Import WildCat.FunctorCat.
 Require Import WildCat.ZeroGroupoid.
 
 (** Using wild 0-groupoids, the universal property can be proven without funext. A simple equivalence of 0-groupoids between [Coeq f g -> P] and [{ h : A -> P & h o f == h o g }] would not carry all the higher-dimensional information, but if we generalize it to dependent functions, then it does suffice. *)
@@ -107,7 +108,7 @@ Section UnivProp.
       $<~> Build_ZeroGpd (sig Coeq_ind_data) _ _ _.
   Proof.
     snrapply Build_CatEquiv.
-    1: rapply FunctorCat.Build_Fun01.
+    1: rapply Build_Fun01.
     rapply isequiv_0gpd_issurjinj.
   Defined.
 
--- a/theories/WildCat/Category.v
+++ b/theories/WildCat/Category.v
@@ -9,12 +9,12 @@ Require Import WildCat.Core WildCat.NatTrans WildCat.FunctorCat.
 
 (** A category is a [Type] together with a proof that it is a 1-category. *)
 Record Category := {
-    category_carrier :> Type;
-    isgraph_category_carrier :: IsGraph category_carrier;
-    is01cat_category_carrier :: Is01Cat category_carrier;
-    is2graph_category_carrier :: Is2Graph category_carrier;
-    is1cat_category_carrier :: Is1Cat category_carrier
-  }.
+  category_carrier :> Type;
+  isgraph_category_carrier :: IsGraph category_carrier;
+  is01cat_category_carrier :: Is01Cat category_carrier;
+  is2graph_category_carrier :: Is2Graph category_carrier;
+  is1cat_category_carrier :: Is1Cat category_carrier
+}.
 
 (** Morphisms of categories are given by functors. *)
 Instance isgraph_cat : IsGraph Category
--- a/theories/WildCat/Category.v
+++ b/theories/WildCat/Category.v
@@ -1,0 +1,29 @@
+Require Import Basics.Overture.
+Require Import WildCat.Core WildCat.NatTrans WildCat.FunctorCat.
+
+(** * Category of categories *)
+
+(** TODO: make into a bicategory *)
+
+(** ** Definition *)
+
+(** A category is a [Type] together with a proof that it is a 1-category. *)
+Record Category := {
+    category_carrier :> Type;
+    isgraph_category_carrier :: IsGraph category_carrier;
+    is01cat_category_carrier :: Is01Cat category_carrier;
+    is2graph_category_carrier :: Is2Graph category_carrier;
+    is1cat_category_carrier :: Is1Cat category_carrier
+  }.
+
+(** Morphisms of categories are given by functors. *)
+Instance isgraph_cat : IsGraph Category
+  := { Hom A B := Fun11 A B }.
+
+(** 2-morphisms of categories are given by natural transformations. *)
+Instance is2graph_cat : Is2Graph Category
+  := fun A B => _.
+
+(** 3-morphisms of categories are given by "wild modifications" - the underlying data of a modification, subject to no coherence conditions. *)
+Instance is3graph_cat : Is3Graph Category
+  := fun A B => _.
--- a/theories/WildCat/FunctorCat.v
+++ b/theories/WildCat/FunctorCat.v
@@ -18,6 +18,7 @@ Coercion fun01_F : Fun01 >-> Funclass.
 Existing Instance fun01_is0functor.
 
 Arguments Build_Fun01 A B {isgraph_A isgraph_B} F {fun01_is0functor} : rename.
+Arguments fun01_F {A B isgraph_A isgraph_B} : rename.
 
 Definition issig_Fun01 (A B : Type) `{IsGraph A} `{IsGraph B}
   : _  <~> Fun01 A B := ltac:(issig).
--- a/theories/WildCat/Graph.v
+++ b/theories/WildCat/Graph.v
@@ -17,7 +17,7 @@ Record Graph := {
 Instance isgraph_graph : IsGraph Graph
   := { Hom A B := Fun01 A B }.
 
-  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
+  (** A 2-cell between graph homomorphisms [F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
 Instance is2graph_graph : Is2Graph Graph
   := fun A B => {| Hom F G := Transformation (fun01_F F) (fun01_F G) |}.
 
--- a/theories/WildCat/Graph.v
+++ b/theories/WildCat/Graph.v
@@ -1,0 +1,28 @@
+Require Import Basics.Overture Basics.Tactics.
+Require Import WildCat.Core WildCat.NatTrans WildCat.FunctorCat.
+
+(** * Category of graphs *)
+
+(** ** Definition *)
+
+(** A graph is a [Type] together with a relation [$->]. *)
+Record Graph := {
+  graph_carrier :> Type;
+  isgraph_graph_carrier :: IsGraph graph_carrier
+}.
+
+(** ** Category of graphs *)
+
+(** Morphisms of graphs are 0-coherent 1-functors, i.e., maps between the underlying types of the graphs together with an [fmap] functorial map taking arrows to appropriate arrows. This is known as a graph homomorphism. *)
+Instance isgraph_graph : IsGraph Graph
+  := { Hom A B := Fun01 A B }.
+
+  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) transformation.*)
+Instance is2graph_graph : Is2Graph Graph
+  := fun A B => {| Hom F G := Transformation (fun01_F F) (fun01_F G) |}.
+
+(** We have identity graph homomorphisms together with compositions of graph homomorphisms. *)
+Instance is01cat_graph : Is01Cat Graph := {
+  Id A := fun01_id;
+  cat_comp A B C G F := fun01_compose G F
+}.
--- a/theories/WildCat/Graph.v
+++ b/theories/WildCat/Graph.v
@@ -17,7 +17,7 @@ Record Graph := {
 Instance isgraph_graph : IsGraph Graph
   := { Hom A B := Fun01 A B }.
 
-  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) transformation.*)
+  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
 Instance is2graph_graph : Is2Graph Graph
   := fun A B => {| Hom F G := Transformation (fun01_F F) (fun01_F G) |}.
 
--- a/theories/Diagrams/ConstantDiagram.v
+++ b/theories/Diagrams/ConstantDiagram.v
@@ -2,7 +2,7 @@ Require Import Basics Types.Paths.
 Require Import Cone.
 Require Import Cocone.
 Require Import Diagram.
-Require Import Graph.
+Require Import Diagrams.Graph.
 
 (** * Constant diagram *)
 
--- a/theories/Homotopy/Join/Core.v
+++ b/theories/Homotopy/Join/Core.v
@@ -323,7 +323,7 @@ Global Instance is0functor_joinrecdata_0gpd (A B : Type) : Is0Functor (joinrecda
 Proof.
   apply Build_Is0Functor.
   intros P Q g.
-  snrapply Build_Morphism_0Gpd.
+  snrapply Build_Fun01.
   - exact (joinrecdata_fun g).
   - apply is0functor_joinrecdata_fun.
 Defined.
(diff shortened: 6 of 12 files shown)


theories/WildCat/Core.v
```diff
@@ -9,6 +9,18 @@ Class IsGraph (A : Type) :=
   Hom : A -> A -> Type
 }.
 
+Module Graph.
```
Alizter, line 12: Quick comments:

If you want to do "module driven development" (a la OCaml), typically its better to write `t`, so the type becomes `Graph.t`. Also you can do `Export` after the module for the instances and coercions using import/export filters. 

However, we typically avoid uses of modules in the library, as it can make understanding things a lot harder. Coq's module system also differs from OCaml's module system in some painful ways which makes using them a lot less ergonomic.

Here the goal is to introduce a namespace. I would instead prefer that the bundled `Graph` lives in its own file in `WildCat/Graph.v`. The name of the type should be `Graph` and the fields should be `graph_carrier` and `isgraph_graph_carrier`.

theories/WildCat/FunctorCat.v
```diff
@@ -14,11 +14,26 @@ Record Fun01 (A B : Type) `{IsGraph A} `{IsGraph B} := {
   fun01_is0functor : Is0Functor fun01_F;
 }.
 
+Instance is0Graph_Graph : IsGraph Graph := {
```
Alizter, line 17: This should be in `Graph.v` I mentioned above.

theories/WildCat/FunctorCat.v
```diff
@@ -14,11 +14,26 @@ Record Fun01 (A B : Type) `{IsGraph A} `{IsGraph B} := {
...
+Arguments fun01_F {A B H H0}.
 Global Existing Instance fun01_is0functor.
 
 Arguments Build_Fun01 A B {isgraph_A isgraph_B} F {fun01_is0functor} : rename.
 
+Instance Is2GraphGraph : Is2Graph Graph :=
+  fun A B => {| Hom F G := Transformation (fun01_F F) (fun01_F G)|}.
+
+(** There is a (0,1)-category of graphs under composition. *)
+Instance is01cat_Graph : Is01Cat Graph := {
```
Alizter, line 31: And here

theories/WildCat/Core.v
```diff
@@ -115,6 +127,27 @@ Arguments cat_assoc_opp {_ _ _ _ _ _ _ _ _} f g h.
 Arguments cat_idl {_ _ _ _ _ _ _} f.
 Arguments cat_idr {_ _ _ _ _ _ _} f.
 
+Module Category.
+  Class class_of (A : Type) := Class {
```
Alizter, line 131: Why have this separate? It should be the fields of the type.

theories/WildCat/Core.v
```diff
@@ -115,6 +127,27 @@ Arguments cat_assoc_opp {_ _ _ _ _ _ _ _ _} f g h.
 Arguments cat_idl {_ _ _ _ _ _ _} f.
 Arguments cat_idr {_ _ _ _ _ _ _} f.
 
+Module Category.
```
Alizter, line 130: Same situation here about modules. I would rather this lives in `Category.v`, where all related things show up. This is similar to functors living in `Core.v` and bundled functors getting their own file.

theories/WildCat/FunctorCat.v
```diff
@@ -132,10 +147,16 @@ Proof.
   exists F; exact _.
 Defined.
 
+Instance isgraph_Cat : IsGraph Category := { Hom A B := Fun11 A B }.
```
Alizter, line 150: This and the others should be moved to `Category.v`.

theories/WildCat/Paths.v
```diff
@@ -24,11 +24,11 @@ Local Existing Instances isgraph_paths is2graph_paths is3graph_paths | 10.
 
 (** Any type has composition and identity morphisms given by path concatenation and reflexivity. *)
 Global Instance is01cat_paths (A : Type) : Is01Cat A
-  := {| Id := @idpath _ ; cat_comp := fun _ _ _ x y => concat y x |}.
+  := { Id := @idpath _ ; cat_comp := fun _ _ _ x y => concat y x }.
```
Alizter, line 27: What happened here?

patrick-nicodemus: I created some new typeclass instances and it messed up the order of hints and caused this to not elaborate properly. The bidirectionality hint helps it find the right instances.

theories/WildCat/Core.v
patrick-nicodemus: I think that there will be a necessity to later bundle other structures as well for similar reasons as to here, so I wanted to do this in a prudent way. I know that the math-comp library has used this style of "packed classes" which is organized around the module system, so I looked in the book "Mathematical Components" textbook by Mahboubi and Tassi https://math-comp.github.io/mcb/, Section 8.3 describes this style. I am not experienced enough with tall hierarchies to have a feel for why this style is important but I thought of it as a kind of "defensive" approach.

> If you want to do "module driven development" (a la OCaml), typically its better to write t, so the type becomes Graph.t.

I am familiar with this OCaml convention. The name `type` appears to be the math-comp convention, from the book cited.

> we typically avoid uses of modules in the library, as it can make understanding things a lot harder. Coq's module system also differs from OCaml's module system in some painful ways which makes using them a lot less ergonomic.

Can you elaborate on this? I think moving things to their own file in `Graph.v` and `Category.v` is fine but I am wondering about what the pitfalls are with the use of modules (both difficulty of understanding and inconvenience). You can link to previous discussions or Zulip threads.

theories/WildCat/Core.v
```diff
@@ -129,6 +162,17 @@ Definition Build_Is1Cat' (A : Type) `{!IsGraph A, !Is2Graph A, !Is01Cat A}
   := Build_Is1Cat A _ _ _ is01cat_hom is0gpd_hom is0functor_postcomp is0functor_precomp
       cat_assoc (fun a b c d f g h => (cat_assoc a b c d f g h)^$) cat_idl cat_idr.
 
+Record AreInverse {A} `{Is1Cat A} {a b : A} (f : a $->b) (g : b $->a) := {
```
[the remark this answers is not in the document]
patrick-nicodemus: I intend to use it in a later commit that builds on this one introducing bicategories, but I can see that it appears unrelated on its own. I can remove it if you want, and put it into the later commit.

theories/WildCat/Core.v
patrick-nicodemus: I can collapse `Category.type` and `Category.class_of` into one layer if you want, I am just trying to follow the packed-classes discipline as closely as possible.

patrick-nicodemus:
@Alizter While following up on your commits, I noticed that there is already defined a type of bundled Graphs in Diagrams.Graph. I was not aware of this, and I assume we should not duplicate it? I imagine it is it better to have `Diagrams` refactored to depend on `WildCat.Graph`, but this will take some work and so perhaps we should discuss what reorganization is appropriate before I dive into it.

theories/WildCat/Core.v
patrick-nicodemus: I have collapsed this in the current PR to get things to go forward but I think we should discuss some bundling methodology or standard approach to use. As I said before, bundled versions of structures will often be necessary so there has to be some style to use, it will be nice to have a discussion about the kinds of issues that may arise with different styles.

patrick-nicodemus:
@Alizter I made the changes you suggested. I tried to match the indentation and line break style to other code in the WildCat library but it's not super consistent through the library.

theories/WildCat/Category.v
```diff
@@ -0,0 +1,21 @@
...
+    isgraph_category_carrier :: IsGraph category_carrier;
+    is01cat_category_carrier :: Is01Cat category_carrier;
+    is2graph_category_carrier :: Is2Graph category_carrier;
+    is1cat_category_carrier :: Is1Cat category_carrier (* This can be seen as the "mixin" in the sense of packed classes. *)
+  }.
+
+Instance isgraph_Cat : IsGraph Category
+  := { Hom A B := Fun11 A B }.
+
+Instance Is2GraphCategory : Is2Graph Category
```
Alizter, line 15: This isn't needed. `Fun11` is already a 1-category so it should be inferred. `:= _.`. Also the name should be `is2graph_cat` and a similar naming scheme for the others. I would prefer lower case especially in the middle of the name. We can distinguish it from category by using `cat`.

Alizter: Specifically `:= fun A B => _.`.

Alizter: So basically:

```coq
Instance isgraph_cat : IsGraph Category := { Hom A B := Fun11 A B }.
Instance is2graph_cat : Is2Graph Category := fun A B => _.
Instance is3graph_cat : Is3Graph Category := fun A B => _.
```

Alizter:
Actually, I have a few minor changes. If you are alright with it, I can push directly here and save some back-and-forth.

theories/WildCat/Core.v
Alizter: Yes, we should definitely discuss some of the tradeoffs. When I last looked into this problem, it wasn't obvious which solution would be better.

theories/WildCat/Core.v
Alizter: My main gripe with Coq modules is that they are difficult for newcomers to understand. Have everything laid out flatly can greatly benefit readability, and I am not yet convinced modules buy us any advantage when it comes to "modular code generation".

A classic example of the pitfall of modules is the Coq stdlib material on natural numbers and integers. All such lemmas are sparsely populated between many different files, and some definitions are generated with module functor applications. Implementations of these definitions become difficult to source or follow, which makes proving properties about them all the more difficult.

The library has always been hesitant to use modules in the past and we had one big use that was difficult to work with. There were various universe bugs that stopped us from having a good definition of modality a few years ago. Because of this there was some module magic whose sole purpose was to share the work between reflective subuniverses and modalities. It made understanding the code very difficult, which wasn't helped by Coq's lack of good documentation at the time. Eventually, it was fixed so that a modality was a special case of a reflective subuniverse, and this made working with it easier. That's one example from memory.

Alizter:
It might be interesting to replace `Diagrams.Graph` now that we have this. I don't think we should touch that just yet however, as others are working on colimits at the moment so we don't want to give them a hard time rebasing.

theories/WildCat/Graph.v
```diff
@@ -0,0 +1,27 @@
...
+
+(** A graph is a [Type] together with a relation [$->]. *)
+Record Graph := {
+  graph_carrier :> Type;
+  isgraph_graph_carrier :: IsGraph graph_carrier
+}.
+
+(** ** Category of graphs *)
+
+(** Morphsims of graphs are 0-coherent 1-functors. I.e. maps between the underlying types of the graphs together wtih an [fmap] functional map taking arrows to appropriate arrows. This is known as a graph homomophism. *)
```
jdchristensen, line 16:
```suggestion
(** Morphisms of graphs are 0-coherent 1-functors. I.e. maps between the underlying types of the graphs together wtih an [fmap] functional map taking arrows to appropriate arrows. This is known as a graph homomorphism. *)
```

theories/WildCat/Graph.v
```diff
@@ -0,0 +1,27 @@
...
+Instance isgraph_graph : IsGraph Graph := { Hom A B := Fun01 A B }.
+
+(** The 2-graph structure on [Fun01 A B] only appears when [B] is a 1-category. We can give a different notion of "graph homotopy" betwen graph homomorphisms by supplying a transformation between the two homomorphisms. *)
```
jdchristensen, line 19:
```suggestion
(** The 2-graph structure on [Fun01 A B] only appears when [B] is a 1-category. We can give a different notion of "graph homotopy" between graph homomorphisms by supplying a transformation between the two homomorphisms. *)
```

jdchristensen: This comment is really confusing to me.  First, instead of "2-graph structure on [Fun01 A B]", shouldn't it say "1-graph structure on ..."?  Second, it sounds to me like this comment is saying that we will assume that `B` is a 1-category.  How about switching things around, e.g.
```coq
(** The 2-cells are given by transformations.  (When [B] is a 1-category, [Fun01 A B] has a different graph structure, namely the *natural* transformations, but that is not what we use here.) *)
```

theories/WildCat/Category.v
```diff
@@ -0,0 +1,26 @@
+Require Import Basics.Overture.
+Require Import WildCat.Core WildCat.NatTrans WildCat.FunctorCat.
+
+(** * Category of categories *)
+
+(** TODO: make into bicategory *)
```
jdchristensen, line 6:
```suggestion
(** TODO: make into a bicategory *) 
```

theories/WildCat/Category.v
```diff
@@ -0,0 +1,26 @@
...
+
+(** ** Definition *)
+
+(** A category is a [Type] together with a proof that it is a 1-category. *)
+Record Category := {
+    category_carrier :> Type;
+    isgraph_category_carrier :: IsGraph category_carrier;
+    is01cat_category_carrier :: Is01Cat category_carrier;
+    is2graph_category_carrier :: Is2Graph category_carrier;
+    is1cat_category_carrier :: Is1Cat category_carrier (* This can be seen as the "mixin" in the sense of packed classes. *)
```
jdchristensen, line 16: The comment on the last line is confusing to me.  Drop it?

theories/WildCat/Category.v
```diff
@@ -0,0 +1,26 @@
...
+  }.
+
+(** Morphsisms of categories are given by functors. *)
+Instance isgraph_cat : IsGraph Category := { Hom A B := Fun11 A B }.
+
+(** 2-morphsisms of categories are given by natural transformations. *)
+Instance is2graph_cat : Is2Graph Category := fun A B => _.
+
+(** 3-morphsisms of categories are given by the underlying transformation of the expected modification. *)
```
jdchristensen, line 25: Typos in all three comments.  Not sure who to blame, but please proofread comments.

jdchristensen: I'm also not sure what "the underlying transformation of the expected modification" means.  Can you be more explicit?

jdchristensen (commented):
BTW, is there any overlap with things in ZeroGroupoid.v?  Or at least, are the styles compatible?

theories/WildCat/Graph.v
patrick-nicodemus: Two other typos: "wtih" and "homomophism". Is there a way to use standard Emacs spell checking tools like ispell in comments?

patrick-nicodemus: I disagree with the choice of the term "functional" here, as this definition will also be used for graphs whose edges are not functions.

Alizter:
I am the one to blame about the comments. I wrote them a bit too quickly my bad. Feel free to fix, I will check them again later today.

theories/WildCat/Graph.v
Alizter: Here is what I am trying to say:

`Fun01 A B` has its own 1-category structure, but this in turn relies on the 1-category structure of `B`. Therefore in order to give a 2-morphisms in the category of graphs we use the weaker notion of a transformation between the two functors.

Since graphs don't have composition, natural transformations don't make sense anyway.

Dan's comment should be fine.

patrick-nicodemus:
I was not familiar with the ZeroGroupoid file before you pointed it out. They are written in the same style so it's consistent. On the other hand some things are deliberately redundant in that file. I created a post in the Coq Zulip to get feedback on whether it would be appropriate to relate bundled Graphs to bundled ZeroGroupoids more directly

theories/WildCat/Graph.v
```diff
@@ -0,0 +1,28 @@
...
+  isgraph_graph_carrier :: IsGraph graph_carrier
+}.
+
+(** ** Category of graphs *)
+
+(** Morphisms of graphs are 0-coherent 1-functors, i.e., maps between the underlying types of the graphs together with an [fmap] functorial map taking arrows to appropriate arrows. This is known as a graph homomorphism. *)
+Instance isgraph_graph : IsGraph Graph
+  := { Hom A B := Fun01 A B }.
+
+  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) transformation.*)
```
jdchristensen, line 20:
```suggestion
  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
```

patrick-nicodemus:
I just added a second commit in response to @jdchristensen 's question about whether ZeroGroupoid should be refactored to depend on Graph. We can use the `isgraph_induced`, `is01cat_induced` and `is2graph_induced` to show that these structures on ZeroGroupoid are inherited from the type Graph. However, the author of the ZeroGroupoid file deliberately duplicated an isomorphic copy of the `Fun01` type because "we put a different graph and 01-category structure on it, so we give this a custom name", so they might have good reasons for not wanting this functionality to be merged. (The fix was easy, just a lot of renaming.)

theories/Colimits/CoeqUnivProp.v
```diff
@@ -107,7 +107,7 @@ Section UnivProp.
       $<~> Build_ZeroGpd (sig Coeq_ind_data) _ _ _.
   Proof.
     snrapply Build_CatEquiv.
-    1: rapply Build_Morphism_0Gpd.
+    1: rapply FunctorCat.Build_Fun01.
```
jdchristensen, line 110: Can you import FunctorCat so this can be shorter?

jdchristensen (commented):
I wrote the ZeroGroupoid.v file, and I guess I thought that Coq would get confused if `Fun01` had more than one graph (etc) structure.  But since these instances come from the underlying type of objects rather than the type of morphisms, I guess there is no confusion.  So these latest changes look good to me.

theories/Colimits/CoeqUnivProp.v
patrick-nicodemus: I did but importing those typeclass instances causes `rapply` to spin in some places. I can try reorganizing the imports to avoid this or debugging the typeclass search.

jdchristensen (approved):
The overall PR looks good to me.

One procedural comment is that I would reduce the use of forced pushes.  They are appropriate in some cases, e.g. to remove a commit that we decide should be omitted, but in general it's easier to follow the progress in a PR when fixes are pushed as new commits.  If by the end there are too many such minor fixes, some of the commits can be squashed before merging.

theories/Colimits/CoeqUnivProp.v
jdchristensen: Ah, maybe that's the kind of thing I was thinking of when I chose to make a duplicate copy of Fun01?  In any case, if you don't quickly find an easy solution, it's to leave line 110 as it is.  But I would then add `Require FunctorCat` (without an `Import`) at the top, preceded by a comment:  We depend on this but do not import it because doing so adds typeclass instances that we do not want.

patrick-nicodemus:
> The overall PR looks good to me.
> 
> One procedural comment is that I would reduce the use of forced pushes. They are appropriate in some cases, e.g. to remove a commit that we decide should be omitted, but in general it's easier to follow the progress in a PR when fixes are pushed as new commits. If by the end there are too many such minor fixes, some of the commits can be squashed before merging.

Definitely! Understood.

jdchristensen:
Your last point is a good one, and I'm not sure how to deal with it, except to use more `(A:=...)` annotations.

theories/Colimits/CoeqUnivProp.v
jdchristensen: Did it turn out that the result was sensitive to the ordering of the imports?  If so, maybe add a comment like: `(** FunctorCat must be imported before XXX so that the instances in the latter are tried first. *)`

jdchristensen:
@Alizter Do you want to review this?  Modulo my one comment request, I think this is ready to merge.

theories/WildCat/Category.v
```diff
@@ -0,0 +1,29 @@
...
+(** ** Definition *)
+
+(** A category is a [Type] together with a proof that it is a 1-category. *)
+Record Category := {
+    category_carrier :> Type;
+    isgraph_category_carrier :: IsGraph category_carrier;
+    is01cat_category_carrier :: Is01Cat category_carrier;
+    is2graph_category_carrier :: Is2Graph category_carrier;
+    is1cat_category_carrier :: Is1Cat category_carrier
+  }.
```
Alizter, line 17:
```suggestion
  category_carrier :> Type;
  isgraph_category_carrier :: IsGraph category_carrier;
  is01cat_category_carrier :: Is01Cat category_carrier;
  is2graph_category_carrier :: Is2Graph category_carrier;
  is1cat_category_carrier :: Is1Cat category_carrier
}.
```

theories/WildCat/Graph.v
```diff
@@ -0,0 +1,28 @@
...
+  isgraph_graph_carrier :: IsGraph graph_carrier
+}.
+
+(** ** Category of graphs *)
+
+(** Morphisms of graphs are 0-coherent 1-functors, i.e., maps between the underlying types of the graphs together with an [fmap] functorial map taking arrows to appropriate arrows. This is known as a graph homomorphism. *)
+Instance isgraph_graph : IsGraph Graph
+  := { Hom A B := Fun01 A B }.
+
+  (** A 2-cell between graph homomorphisms $[F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
```
Alizter, line 20:
```suggestion
  (** A 2-cell between graph homomorphisms [F G : A $-> B] is a wild [Transformation], i.e., a family of 1-cells [forall x, F x $-> G x] subject to no coherence conditions. When [A] and [B] have more structure (e.g., they are 1-categories or bicategories) a stronger notion of 2-cell is appropriate, such as a (possibly lax or oplax) natural transformation.*)
```

Alizter (approved):
Modulo my comments, this LGTM.

patrick-nicodemus:
Both comments have been addressed

PR #2239, opened 2025-02-28
Change Global instances to #[export]

patrick-nicodemus:
Suggests a different approach to managing typeclass instances now that the "export" locality attribute is available


STYLE.md
```diff
@@ -430,15 +430,22 @@ Hint Immediate foo : typeclass_instances.
...
+in the instance database for typeclass resolution outside its containing section,
+module or file). The `Global` attribute makes the instance available in
+any file that "loads" this file via the `Require` command,
+which is a transitive relation including all direct and indirect dependencies of the current file.
+The `#[export]` locality uses the module system to manage the scope of the typeclass hint, 
+so that the hint is available wherever the module is opened using the `Import` keyword
+(analogous to "open" in ML languages).
+
+When declaring an `Instance` you should prefer the `export` locality attribute
+to the `global` locality attribute. Inside a section, you should
```
jdchristensen, line 445:
```suggestion
The `#[export]` locality uses the module system to manage the scope of the typeclass hint, 
so that the hint is available wherever the module is opened using the `Import` keyword,
but not transitively (analogous to "open" in ML languages).

When declaring an `Instance` you should prefer the `export` locality attribute
to the `Global` locality attribute. Inside a section, you should
```

patrick-nicodemus:
In this series of commits I remove all the global instances from some of the folders to see how much maintenance is required.

I think the documentation you're citing is either incorrect or we are misinterpreting it. For me on Coq 8.20 I have to add the `#[export]` annotation to hints inside a section to get it to compile. So one can do a global search and replace of `Instance` by `#[export] Instance` but it will look a bit nicer if it's not annotated.

The kind of maintenance I had to do was:
1. Adding new explicit import statements
2. Fixing breakages which I believe are due to changes in the order of hints in the typeclass database.

To elaborate on (2.), if `A` imports `B`  at the top of the file, `A` declares a hint `Hint_A` and `B` declares a global hint `Hint_B`, then if `C` imports `A` and then `B`, then `Hint_B` will occur in the hint database before `Hint_A` because `Hint_B` was added first (at the time that `B` was first loaded, at the beginning of `A`). But if you now make `B` a module-scoped hint, then `Hint_A` gets added before `Hint_B`, because `C` imports `A` before `B`.

So you have typeclass searches which were quick before but now take a very long time, or go down a dead end. Most of these can be fixed by setting the typeclass search depth to something lower.

patrick-nicodemus:
The library builds after each commit, each one is self contained. Perhaps I should have put the "fixes" in separate commits for readability.

patrick-nicodemus:
> I guess you could try to do it based on indentation. Replace   Global Instance with   #[export] Instance and then replace the remaining occurences of Global Instance with Instance

I have been doing the manual equivalent of this but indenting style is not consistent throughout the library, so this is a very noisy signal for guessing whether you are in a file.

theories/Algebra/Universal/Homomorphism.v
```diff
@@ -66,9 +66,10 @@ Proof.
   issig.
 Defined.
 
-Global Instance hset_homomorphism `{Funext} {σ} (A B : Algebra σ)
+Instance hset_homomorphism `{Funext} {σ} (A B : Algebra σ)
   : IsHSet (A $-> B).
 Proof.
+  Search (IsHSet (SigHomomorphism _ _)).
```
jdchristensen, line 72: Remove.

theories/Classes/categories/ua_category.v
```diff
@@ -5,6 +5,7 @@ Require Import
   HoTT.Classes.theory.ua_isomorphic.
 
 
+
```
jdchristensen, line 8: Remove added line.

jdchristensen:
I have just skimmed through the commits.  Here's my summary of the changes besides removing `Global`:

- `Basics`:  only two lines, in V.v, getting rid of `simpl` in both cases.  Fine.
- `Wildcat`:  only five files changed outside of WildCat/*, and I think there were no changes within WildCat.  Not bad, but the changes needed were a bit ugly, so hopefully by reordering imports, changing priorities, or using `Export`, this can be improved.
- `Types`:  none!
- `Algebra`: This commit included removing `Global` from two files in Basics.  There are no other changes outside of the Algebra folder.  Within Algebra, there are two or three additional imports, a few type annotations `: AbGroup`, a couple of changes of `GroupHomomorphism` to `Hom (A:=AbGroup)`, and the coercion from `GroupIsomorphim` to `Hom`.  (Those last few seem to be the same as what was done in the Hint Mode IsGraph PR, so I'm not sure if it's intentional that they are here too.  Maybe the same issue arises, so the same changes help here?)  This is all fine.
- rest of library (except Classes):  none!
- `Classes`: none!

(There's also a merge commit, but github seems to just show me the combination of the previous commits, so I'm assuming nothing substantial changed there.  I also might have missed one or two things in my skimming, but if so, the diff was small enough to not catch my eye.)

So the upshot is that this change affected almost nothing.  It's pretty amazing to me that we didn't seem to be relying on the transitive behaviour of `Global` except in a couple of spots.

The only commit I'm not completely happy with is the WildCat commit, but since only five files were affected, it shouldn't be hard to find a way to get rid of the changes made to those files.  We should also check that the build time has not gone up.

theories/Algebra/AbSES/SixTerm.v
```diff
@@ -38,7 +38,8 @@ Proof.
   hnf. intros [f q]; rapply contr_inhabited_hprop.
   srefine (tr (_; _)).
   { refine (grp_homo_compose _
-              (abses_cokernel_iso (inclusion E) (projection E))^-1$).
+              (abses_cokernel_iso (inclusion E) (projection E)
```
patrick-nicodemus: This was apparently unnecessary after all the dust settled, and I have removed it.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -208,12 +208,12 @@ Section MonoidEnriched.
     `{!HasMorExt A} `{forall x y, IsHSet (x $-> y)}.
 
   Section Monoid.
-    Context `{!IsMonoidObject _ _ y}.
+    Context `{!IsMonoidObject (fun a b : A => cat_binprod a b) unit y}.
```
patrick-nicodemus: I looked into this to see if it could be fixed by rearranging the typeclass database and I would prefer to just be more explicit in some parts.

I have rewritten a lot of this section on the philosophy that more changes to keep things pretty is better than a few ugly changes.

patrick-nicodemus:
I have done a timing test and on my machine the time increased by ~1% on this PR. I assume this is considered negligible/within the realm of noise.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -201,19 +201,22 @@ Defined.
 (** ** Monoid enrichment *)
 
 (** A hom [x $-> y] in a cartesian category where [y] is a monoid object has the structure of a monoid. Equivalently, a hom [x $-> y] in a cartesian category where [x] is a comonoid object has the structure of a monoid. *)
-
```
jdchristensen, line 204: When a comment applies to many following things, it is often followed by a blank line.  When it applies to just the next thing, it usually is not followed by a blank line.  In any case, you shouldn't change things not relevant to the PR.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -278,22 +280,21 @@ End MonoidEnriched.
 
 (** ** Preservation of monoid objects by lax monoidal functors *)
 
-Local Set Typeclasses Depth 2.
 Definition mo_preserved {A B : Type}
-  {tensorA : A -> A -> A} {tensorB : B -> B -> B} {IA : A} {IB : B} 
+  { tensorA : A -> A -> A} {tensorB : B -> B -> B} {IA : A} {IB : B}
```
jdchristensen, line 284: Undo.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -201,19 +201,22 @@ Defined.
...
 
 (** A hom [x $-> y] in a cartesian category where [y] is a monoid object has the structure of a monoid. Equivalently, a hom [x $-> y] in a cartesian category where [x] is a comonoid object has the structure of a monoid. *)
-
 Section MonoidEnriched.
   Context {A : Type} `{HasEquivs A} `{!HasBinaryProducts A}
     (unit : A) `{!IsTerminal unit} {x y : A}
     `{!HasMorExt A} `{forall x y, IsHSet (x $-> y)}.
+  
+  Notation "*" := (fun x y : A => cat_binprod (A:=A) x y).
+  Notation I := unit.
```
jdchristensen, line 210: Here I would prefer to just change `unit` to `I` in the `Context` line.

theories/Pointed/pSusp.v
```diff
@@ -294,6 +294,9 @@ Global Instance is1natural_loop_susp_adjoint_r `{Funext} (A : pType)
 Proof.
   snrapply Build_Is1Natural.
   intros B B' g f.
+  match goal with 
+  | [|- _ = ?b ] => change b with  (fmap loops g $o (fmap loops f $o (loop_susp_unit A)))
+  end.
```
jdchristensen, line 299: You can revert all of the changes in `pSusp` (two hunks) if you just move the `Require Import WildCat.` earlier, e.g. to right after `Types` is imported.  This makes sense, since the imports of `Pointed.*` are more specialized and so should come later.

theories/Pointed/Loops.v
```diff
@@ -106,7 +106,7 @@ Defined.
 
 Lemma fmap_loops_pconst {A B : pType} : fmap loops (@pconst A B) ==* pconst.
 Proof.
-  rapply fmap_zero_morphism.
+  exact (fmap_zero_morphism loops).
```
jdchristensen, line 109: It doesn't matter much, but this can be reverted by moving the `Require Import WildCat` earlier.  I think the move should be done, so that if future changes are made to Loops.v, they will benefit from the more appropriate typeclasses being found.

jdchristensen:
> I have done a timing test and on my machine the time increased by ~1% on this PR. I assume this is considered negligible/within the realm of noise.

Yes.

> I was hoping this PR would improve performance somewhat, but mostly I was hoping it would help to debug long loops of `rapply` because I don't have to worry about transitively imported hints that are difficult to find. I think that the PR can be justified on grounds of modularity even though it doesn't appear to improve performance.

I was hoping for an improvement as well, but like you, I still agree with this PR on principle.

theories/Pointed/pSusp.v
jdchristensen: Or add `Existing Instance isgraph_ptype | 0.` after the imports are done.  This also means that we try the `pType` instance before the `pForall` instance.

theories/Algebra/Categorical/MonoidObject.v
patrick-nicodemus: Sorry, accident.

jdchristensen:
To me, this is essentially done, except for the minor issue about the `*` notation.  With the recent commits, all of the awkward workarounds are gone, so there are only minor changes besides all of the removals of `Global`.  So I have marked this as Ready for review.

theories/Algebra/Categorical/MonoidObject.v
```diff
@@ -204,16 +204,19 @@ Defined.
 
 Section MonoidEnriched.
   Context {A : Type} `{HasEquivs A} `{!HasBinaryProducts A}
-    (unit : A) `{!IsTerminal unit} {x y : A}
+    (I : A) `{!IsTerminal I} {x y : A}
     `{!HasMorExt A} `{forall x y, IsHSet (x $-> y)}.
+  
+  Notation "*" := (fun x y : A => cat_binprod (A:=A) x y).
```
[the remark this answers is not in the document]
patrick-nicodemus: Frankly I would have preferred to use the Unicode multiplication symbol but I guess this is part of the core library.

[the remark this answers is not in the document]
patrick-nicodemus: Oh, I was thinking of the `Let` keyword, not `Local Definition`.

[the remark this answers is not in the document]
Alizter: I like the idea with `cat_binprod` and `cat_binprod'` but it should be the other way round. I agree with avoiding the `*` notation.

theories/Algebra/AbGroups/AbPullback.v
```diff
@@ -12,7 +12,7 @@ Section AbPullback.
   (* Variables are named to correspond with Limits.Pullback. *)
   Context {A B C : AbGroup} (f : B $-> A) (g : C $-> A).
 
-  Global Instance ab_pullback_commutative
+  Instance ab_pullback_commutative
```
Alizter, line 15: Other times you've replaced such instances with `#[export]`, but here it doesn't look like this will survive the end of the section. It isn't used anywhere, but it might be confusing if it is.

theories/Classes/orders/orders.v
```diff
@@ -558,6 +558,6 @@ Section pseudo.
     - destruct (pseudo_order_antisym y z (ltyz , ltzy)).
   Qed.
 
-  Global Existing Instance lt_transitive.
+  Existing Instance lt_transitive.
```
Alizter, line 561: `#[export]`?

Alizter:
Can you hardwrap the paragraph in STYLE.md at 70 chars? I had a look through all of the changes and they look good. Thanks for tacking this! This is something we should have done a while back.

I have the same comment as Dan on the changes in `MonoidObject`. Let me know if you need any help.

theories/WildCat/Products.v
```diff
@@ -777,15 +779,16 @@ Section Associativity.
...
+      2: rapply fmap22.
+      1: rapply fmap11_comp.
       1: exact (cat_idl _ $@ (cat_idr _)^$).
       nrapply cat_binprod_beta_pr2.
   Defined.
 
   Local Existing Instance symmetricbraiding_binprod.
 
-  Local Instance associator_cat_binprod : Associator (fun x y => cat_binprod x y).
+  Local Instance associator_cat_binprod : Associator (cat_binprod).
```
jdchristensen, line 791: You can omit the parentheses here.

jdchristensen:
```suggestion
  Local Instance associator_cat_binprod : Associator cat_binprod.
```

jdchristensen:
This looks good.  I'll merge when the CI is done, unless someone speaks up before then.

Alizter (approved):

Alizter:
Again @patrick-nicodemus, thanks a lot for tackling this. It's really appreciated.

patrick-nicodemus:
@jdchristensen 
The commit you merged does not remove all Global Instances, only the Global Instances from the handful of sub-libraries which contain the majority of them (I followed your suggestion up above: remove everything from one library, change what needs to be changed, and go onto another.)

The merged commit removed some 2500 Global Instances. I have gone through and removed the remaining ~1000. There are only two changes:
1. isO_contr was changed to have priority 2 rather than 1
2. An import of the Pointed library in an AbSES file was moved to the top of the file so that group instances take priority over ptype instances in that file

jdchristensen:
Ah, I misread this commit message: "Removed all 'Global Existing Instances' from the entire library."

patrick-nicodemus:
@jdchristensen Yes, that is a silly distinction to make but since I was going by textual search and replace, I removed `Global Existing Instances` in a separate commit from `Global Instances`

jdchristensen:
Oh, no worries, the commit message was fine and I just misread it.

PR #2240, opened 2025-03-01
rapply definition variation

patrick-nicodemus:
Experiment with a weaker definition of rapply


patrick-nicodemus:
This pull request has two commits which suggest weaker definitions of refine.
The definition in the first commit should have behavior which is equivalent to refine where the tactic succeeds.
The definition in the second commit has different shelving behavior with regards to subgoals.

PR #2241, opened 2025-03-01
Alternate definitions of rapply

patrick-nicodemus:
This PR introduces two commits which define slightly different and weaker versions of the `rapply` tactic.

The first commit is intended to behave the same way as the original `rapply` wherever it succeeds, but it only succeeds if `nrapply` succeeds.

The second commit has a simpler definition but behaves slightly differently. In particular it has different shelving/unshelving behavior for subgoals, so most changes in the second commit are goal renumberings. It is also apparently slightly weaker due to the fact that `typeclasses eauto` is called on each typeclass goal individually, not all goals together.

The "unify" tactic in Ltac apparently fails if it does not instantiate all existential variables, so it is not applicable here.


theories/Algebra/AbGroups/AbHom.v
```diff
@@ -150,10 +150,10 @@ Proof.
   - exact (functor_ab_coeq a^-1$ b^-1$ (hinverse _ _ p) (hinverse _ _ q)).
   - nrefine (functor_ab_coeq_compose _ _ _ _ _ _ _ _
       $@ functor2_ab_coeq _ _ _ _ _ $@ functor_ab_coeq_id _ _).
-    rapply cate_isretr.
+      exact (cate_isretr _).
   - nrefine (functor_ab_coeq_compose _ _ _ _ _ _ _ _
       $@ functor2_ab_coeq _ _ _ _ _ $@ functor_ab_coeq_id _ _).
-    rapply cate_issect.
+      exact (cate_issect _).
```
jdchristensen, line 156: Minor point: the indentation is off here.

theories/Basics/Tactics.v
```diff
@@ -320,7 +320,7 @@ Tactic Notation "snrefine" uconstr(term) := simple notypeclasses refine term; gl
 
 (** Note that the Coq standard library has a [rapply], but it is like our [rapply'] with many-holes first.  We prefer fewer-holes first, for instance so that a theorem producing an equivalence will by preference be used to produce an equivalence rather than to apply the coercion of that equivalence to a function. *)
 Tactic Notation "rapply" uconstr(term)
-  := do_with_holes ltac:(fun x => refine x) term.
+  := do_with_holes (ltac:(fun x => unshelve nrefine x; try typeclasses eauto)) term.
```
jdchristensen, line 323: Using `unshelve` should make this behave like `srapply`, not `rapply`, so this is not the right way to go.  We definitely want both the `simple refine` and the non-simple `refine` behaviours available to us.  If you remove `unshelve`, does it cause problems?

jdchristensen (commented):
The first commit is pretty good.  But I think we should introduce `tapply` defined exactly like the old `rapply`, so that the code changes can be just a single letter change without extra parentheses and underscores.  (However, there are a few cases where the changed code in the first commit is better than it was before.  E.g. if `rapply` can be changed to `exact` without any other changes to the term, then that's better.)

theories/Homotopy/Bouquet.v
```diff
@@ -82,7 +82,7 @@ Section AssumeUnivalence.
     intros f.
     refine (_ @ @is1natural_equiv_pi1bouquet_rec S _ _ f grp_homo_id).
     simpl; f_ap; symmetry.
-    rapply (cat_idr_strong f).
+    exact (cat_idr_strong f).
```
jdchristensen, line 85: This is an example of a change that should be kept independent of this PR.

jdchristensen: I'm now preparing a large PR that uses `exact` whenever possible, but will wait until this PR is settled to avoid conflicts.

theories/Basics/Tactics.v
```diff
@@ -320,7 +320,7 @@ Tactic Notation "snrefine" uconstr(term) := simple notypeclasses refine term; gl
 
 (** Note that the Coq standard library has a [rapply], but it is like our [rapply'] with many-holes first.  We prefer fewer-holes first, for instance so that a theorem producing an equivalence will by preference be used to produce an equivalence rather than to apply the coercion of that equivalence to a function. *)
 Tactic Notation "rapply" uconstr(term)
-  := do_with_holes ltac:(fun x => refine x) term.
+  := do_with_holes ltac:(fun x => try (try (nrefine x || fail 2); fail); refine x) term.
```
jdchristensen: @JasonGross This change to `rapply` make it try to match the term `x` to the goal *without* typeclass search, and then only if that succeeds, uses `refine` to do typeclass search to fill in as many missing arguments as possible.  This avoids some slow, futile typeclass searches that can happen when `x` doesn't yet have enough underscores, and most uses of `rapply` work fine with this change.  So I'm in favour of it.  My question for you is whether you can see a simpler implementation that is easier to understand.  Note that the successful `nrefine` has to be undone so it can be redone with `refine`.

I had thought that something like `:= nrapply term; try typeclasses eauto.` might work, but the problem is that the typeclass search isn't applied to the shelved goals in that case.  If you do `:= unshelve nrapply term; try typeclasses eauto.`, then that does do typeclass search on all goals, but it doesn't shelve the unsolved ones that would have been shelved by `rapply`.   In other words, it ends up being like `srapply`, but sometimes we want the `rapply` behaviour.

If there's nothing obvious, no worries, as I think the above implementation can stay, with a comment explaining what it is doing.

theories/Basics/Tactics.v
jdchristensen: > If you remove `unshelve`, does it cause problems?

I think it does cause problems, as `typeclasses eauto` will only be applied to the unshelved goals.  So I now understand the dilemma and why you created the more complicated definition with `fail` and `fail 2`.  For now, I think it's best if you remove the second commit from this PR, so we can focus on the first commit.

jdchristensen:
BTW, if we go ahead with this, we need to adjust `rapply'`, `srapply` and `srapply'` as well.  (But note that the primed versions are not used anywhere, so to test the changes, you'd need to temporarily add primes in a few places.)

theories/Homotopy/Bouquet.v
patrick-nicodemus: I tried only to make necessary changes, so unless I made a mistake here, this change is necessary. `exact` is like `refine` in that it invokes typeclass search to further elaborate the terms if unification fails, so it's more powerful than the new `rapply`

theories/Basics/Tactics.v
patrick-nicodemus: @jdchristensen Currently an annoying bug in Coq prevents the widespread of other tactic languages like Elpi and Ltac2 in this library. Coq 9.0 should resolve this. The successor languages for Ltac give more fine-grained control, it should be possible to implement something like this in a more direct way without the silly try/fail logic.

Of course, there are the maintainability issues that come with introducing a new language into the codebase that maintainers now are expected to be able to read, so care has to be taken to document the use of the language. But I also think the model that these languages present of Coq terms and evars is easier to understand and debug than for Ltac.

theories/Homotopy/Bouquet.v
jdchristensen: Oh, I see that my phrasing was unclear.  I meant "This is an example of a change that we should do, even if we don't change `rapply` as in this PR."  In other words, I like this change independent of the fact that it is needed with the new definition of `rapply`.  And I like it so much that I have prepared a commit that does this in around 500 other places...

theories/Basics/Tactics.v
patrick-nicodemus: I don't have strong feelings that this should be merged, I just created it as an experiment/proof-of-concept to test how much work has to be done to change things if `rapply` was made weaker in this way, there's no rush to merge it. So I will make some minor tweaks on this and wait a week or two to see if Jason can propose an alternative definition or an alternative approach to the overall problem.

patrick-nicodemus:
As suggested, I have added primes to a half dozen instances of these tactics to check that the new primed versions work correctly where the original primed versions work correctly.

theories/Basics/Tactics.v
```diff
@@ -319,15 +319,18 @@ Tactic Notation "nrefine" uconstr(term) := notypeclasses refine term; global_axi
 Tactic Notation "snrefine" uconstr(term) := simple notypeclasses refine term; global_axiom.
 
 (** Note that the Coq standard library has a [rapply], but it is like our [rapply'] with many-holes first.  We prefer fewer-holes first, for instance so that a theorem producing an equivalence will by preference be used to produce an equivalence rather than to apply the coercion of that equivalence to a function. *)
+(** This tactic is weaker than tapply, and equivalent in strength to nrapply, i.e., it should succeed iff nrapply succeeds, but it solves all possible typeclasses afterward. The implementation is: try nrefine t, t _, t _ _, ... until success; upon success, revert the last (successful) application of nrefine and call refine (t _ _ _). *)
```
jdchristensen, line 322: Here and in the other new comments below we need to have square brackets around `nrapply`, `t`, etc., so the generated documentation looks good.

theories/Homotopy/CayleyDickson.v
```diff
@@ -149,13 +149,13 @@ Global Instance cds_susp_cdi {A} `(CayleyDicksonImaginaroid A)
...
 Proof.
-  srapply cds_conjug_left_inv.
+  exact cds_conjug_left_inv.
 Defined.
 
 Global Instance cdi_conjugate_susp_right_inverse {A} `(CayleyDicksonImaginaroid A)
   : RightInverse hspace_op (conjugate_susp A cdi_negate) mon_unit.
 Proof.
-  srapply cds_conjug_right_inv.
+  exact (cds_conjug_right_inv _).
```
jdchristensen, line 158: Can we use `tapply` here and in all of the other places in both commits that introduced underscores or other annotations?  (I don't think the `s` version is needed, since no goals are left.)  The point of `tapply` is supposed to be that we can have the diff be a single character change in the places where `rapply` no longer solves a goal.  Similarly, when `srapply` no longer works, we should use `stapply`.

theories/Basics/Tactics.v
```diff
@@ -339,6 +342,12 @@ Tactic Notation "snrapply" uconstr(term)
 Tactic Notation "snrapply'" uconstr(term)
   := do_with_holes' ltac:(fun x => snrefine x) term.
 
+(** This tactic is strictly stronger than rapply, because if the type of the argument term t (with holes) cannot be unified with the goal type, it calls typeclass search on all typeclass holes within t (independently of the goal) and then tries to unify with the goal again. Our implementation of rapply requires that the type (with holes) of the argument term unifies with the goal directly, without any help from typeclass search to fill in the holes in the type. The typeclass search after unification is more robust than the typeclass search before unification, because there is more information available to guide the typeclass search. *)
+Tactic Notation "tapply" uconstr(term)
+  := do_with_holes ltac:(fun x => refine x) term.
+Tactic Notation "tapply'" uconstr(term)
+  := do_with_holes' ltac:(fun x => refine x) term.
```
jdchristensen, line 349: Can we also have `stapply` and `stapply'`?  And can you update the part of STYLE.md that discusses these tactics?

STYLE.md
```diff
@@ -1197,27 +1197,54 @@ They are described more fully, usually with examples, in the files
...
-    have, try `nrapply`.
-  - If you want to use type class resolution as well, try `rapply`.
+    have, try `nrapply`, and then `rapply` if `nrapply` generates
+	typeclass goals.
+  - If you want to use type class resolution as well, try `tapply`.
     But it's better to use `nrapply` if it works.
   - You could add a prime to the tactic, to try with many arguments
     first, decreasing the number on each try.
-  - If you don't want Coq to create evars for certain subgoals,
-    add an `s` to the tactic name to make it use `simple refine`.
```
patrick-nicodemus, line 1220: This is somewhat outside the scope of this PR but I think that this wording is problematic. `Goal { x & x = 0 }. srapply exist.` does create an evar, it's just part of the second goal, not the first one: the evar in the second goal is itself the first goal. The question is whether you want that evar to be an an explicit goal or not.

theories/Basics/Tactics.v
JasonGross: It's not exactly the same, but ` := snrapply term; try typeclasses eauto; shelve_unifiable` should be pretty close to the behavior you want

JasonGross: Otherwise you can clone the goal and use tactics in terms, match on all the evars in the body and run typeclasses eauto in them.  That is, something like `let v := open_constr:(ltac:(nrapply term) : goal_type) in repeat match v with context[?e] => is_evar e; let t := type of e in let e' := open_constr:(ltac:(typeclasses eauto) : t) in unify e e' end; refine v`.  You can strip casts with `Ltac strip_casts v := match v with ?x => x end`, I believe.  Or maybe it's `Notation strip_casts v := (match v with x => x end) (only parsing).`

JasonGross: If you're just looking for easy to understand, there's also `assert_succeds` from Coq's standard library

STYLE.md
Alizter: I agree. Rather than creating subgoals this is about shelving subgoals. I would say

```
If you don't want Coq to automatically shelve certain subgoals....
```

theories/Basics/Tactics.v
jdchristensen: Jason's first and third suggestions look promising to me.  I think the second one looks more complicated than the current implementation.  Thanks, Jason!

jdchristensen:
This is looking very clean.  Besides optionally seeing if Jason's suggestions work for us, my only other thought is that we might rename `nrapply` to `napply` (and `snrapply` to `snapply`).  Then the pattern is a bit more clear:  `napply`, `rapply` and `tapply`, each with an optional `s`.  If we do this, it should be a separate commit, because it will be big.  I could go either way.

patrick-nicodemus:
I have force pushed the branch.
1. The first commit is identical to what it was before except that we use Jason's 3rd definition. Jason's 3rd definition is identical in behavior to the one I originally proposed, and it is more readable. There are no other changes to the first commit other than redefining `rapply` and the other tactics.
2. The second commit uses Jason's first definition and fixes all breakages over the first commit. The first definition has slightly different shelving behavior, it is less powerful in some places but more powerful in others because of the call to `typeclasses eauto`, I suppose. As a result, some proofs in this commit become longer and others become shorter (a few occurrences of `exact _` are removed.)

patrick-nicodemus:
I don't know.
I like that calling `typeclasses eauto` makes some of the proofs shorter; I assume that `typeclasses eauto` takes very little time on holes that are not typeclasses, so I'm not concerned about performance.
Regarding the meaning of the "s" prefix, I find this annoying, because IMO it would be nice to have a hierarchy of three tactics `napply < rapply < tapply` which differ only in the typeclass behavior and not with respect to unrelated things like whether beta reduction is employed in unification or shelving behavior. So, I will lean towards the original definition (Jason's third definition) because I think that that linear hierarchy is easy to understand, and the "simple" prefix is orthogonal, which makes it easier to incorporate.

jdchristensen:
Going with Jason's third definition (using `assert_succeeds`) seems like a good choice.

@Alizter What do you think about renaming `nrapply` to `napply`?  Then a single letter `n`, `r` or `t` selects the strength of the typeclass search, and an optional `s` affects the shelving behaviour.

patrick-nicodemus:
I have reorganized and rewritten the comments in `Basic/Tactics` to present the tactics in the order: `napply, rapply, tapply, snapply, srapply, stapply`, where we first explain what `napply` does, then how `rapply` builds on it, then how `tapply` builds on `rapply`.

jdchristensen (approved):
LGTM modulo one change in STYLE.md

Alizter:
I added a deprecation warning to `nrapply` and `snrapply` since they might be used outside of the library. The message makes it clear how to fix.

I also added a test for the new tactics. I also showed that there is still doing `napply` and `exact _` afterwards, which isn't covered by the new tactics.

theories/Basics/Tactics.v
```diff
@@ -324,6 +324,11 @@ Tactic Notation "napply" uconstr(term)
 Tactic Notation "napply'" uconstr(term)
   := do_with_holes' ltac:(fun x => nrefine x) term.
 
+(** TODO: remove when min Coq/Rocq version is 9.0 *)
+#[deprecated(note="nrapply was renamed to napply", since="9.0")]
+Tactic Notation "nrapply" uconstr(term)
+  := fail "nrapply was renamed to napply".
```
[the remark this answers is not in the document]
jdchristensen: (The meaning of `rapply` also changed, but it's harder to see how to gracefully handle this.  I guess we could have left `rapply` with its old meaning, and added a new definition with the new intermediate behaviour, but we have a lot of muscle memory for `rapply`, and we should be preferring it to `tapply` when possible, I think.)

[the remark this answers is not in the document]
Alizter: I've made them aliases for now. I don't see a good way to handle `rapply`. We could perhaps introduce a compatability module that defines `rapply := tapply` and import that in works where it isn't feasible to switch.

theories/Basics/Tactics.v
```diff
@@ -318,27 +318,52 @@ Tactic Notation "nrefine" uconstr(term) := notypeclasses refine term; global_axi
...
 Tactic Notation "snrefine" uconstr(term) := simple notypeclasses refine term; global_axiom.
 
-(** Note that the Coq standard library has a [rapply], but it is like our [rapply'] with many-holes first.  We prefer fewer-holes first, for instance so that a theorem producing an equivalence will by preference be used to produce an equivalence rather than to apply the coercion of that equivalence to a function. *)
+(** The tactics [napply], [rapply] and [tapply] are similar but they differ in their reliance on typeclass search. [napply t] tries [nrefine t], [nrefine (t _)], [nrefine (t _ _)], so on until it succeeds in unifying with the goal. At each iteration, [nrefine] computes the type of the given term and tries to unify it with the goal; if it succeeds, all holes remaining in the term after unification are new goals. *)
+Tactic Notation "napply" uconstr(term)
+  := do_with_holes ltac:(fun x => nrefine x) term.
+Tactic Notation "napply'" uconstr(term)
+  := do_with_holes' ltac:(fun x => nrefine x) term.
+
+(** TODO: remove when min Coq/Rocq version is 9.0 *)
```
jdchristensen, line 327: Or maybe 9.1?  Or maybe "remove in 2026"?
```suggestion
(** TODO: remove when min Coq/Rocq version is 9.1 *)
```

Alizter:
I'm in favour of merging #2245 before this. There are a few things left in this PR that I would like to change, and it makes more sense for #2245 to be merged before.

jdchristensen:
@Alizter What else are you thinking of changing in this PR?

theories/Basics/Tactics.v
```diff
@@ -324,9 +324,9 @@ Tactic Notation "napply" uconstr(term)
 Tactic Notation "napply'" uconstr(term)
   := do_with_holes' ltac:(fun x => nrefine x) term.
 
-(** TODO: remove when min Coq/Rocq version is 9.0 *)
+(** TODO: remove when min Coq/Rocq version is 9.1 (~ end of 2026). *)
 #[deprecated(note="nrapply was renamed to napply and will be removed soon",
-  since="Coq-HoTT v9.0")]
+  since="Coq-HoTT v9.1")]
```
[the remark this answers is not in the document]
Alizter: These changes will appear in a v9.1 release of the library rather than the v9.0 which is already done.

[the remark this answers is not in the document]
Alizter: But I think switching to dates will make this all clearer. I'll do that in a bit.

Alizter:
@jdchristensen There were a few lines where `tapply` was introduced but `napply` could be used with a bit of tweaking. Some of them were finishing `rapply`s which is why the other PR will solve some of them.

Alizter:
FYI I will attempt to rebase this now.

Alizter:
My rebase has scrubbed your ownership of the commits @patrick-nicodemus. I will rebase again and add you as the main author.

Alizter:
I'm not sure what I am doing wrong. @jdchristensen do you remember how to give ownership of a commit?

jdchristensen:
My tool (magit) provides a keystroke to set the commit author during a rebase, so I can try to fix these if you want.

Alizter:
@jdchristensen That would be great, thanks!

Alizter:
Actually, it looks like the ownership information is correct. Just that the GitHub interface is only showing my picture. Weird.

jdchristensen:
I agree that the `git log` info looks correct, so I won't do a rebase.

Alizter:
I will rebase this again.

Alizter:
That's all rebased, and all comments addressed. Merging now.

Alizter (approved):

Alizter:
Thanks @patrick-nicodemus, great work as always.

PR #2244, opened 2025-03-08
Remove all remaining Global Instances.

### LINKED ISSUE #2227: Heavy use of global instances
I noticed widespread usage of Global Instances through the codebase.

I can understand the reasoning for this, I think. For example, we might have a chain of four dependent files A.v -> B.v -> C.v -> D.v, where A.v treats graphs and graph homomorphisms, B.v treats 2-graphs, C treats categories and D treats 2-categories.
When the user imports D.v, they maybe want typeclass inference to "just work", and so perhaps it should be assumed that they automatically want to import the instances for A, B, and C as well, so that they do not get errors about "typeclass IsGraph is not found" while they are trying to define a 2-category.

Over the past few days of getting used to Coq after not having touched it in a while, I have spent an unpleasant amount of time debugging typeclass issues and unification failures. My instinct is that having a very large typeclass database with a large search space, potentially containing loops, makes it harder to debug than a smaller database which carries only the hints that are relevant for the problems in your file. It seems easier to solve a problem of a missing hint to the database by importing the file, than to solve a problem with typeclass search going off in the wrong direction entirely, because there's so much debug information/logs you have to wade through. To back up my intuition I can appeal to the Coq documentation which deprecates global typeclass hints - https://coq.inria.fr/doc/master/refman/language/core/modules.html#coq:attr.global.

I am wondering if global instances are a design decision you have openly discussed, and if so I'd be interested in the reasoning. I don't want to do a big PR where I systematically remove all the Global annotations but when I see them in the code my impulse is to remove them.

patrick-nicodemus:
The previous PR did not remove *all* global instances, it removed them from 5-6 of the most important sublibraries. This PR finishes the job by removing all global instances.

The only breakages that needed to be fixed are in `AbSES/Core.v` (at the top of the file) and in `Modalities/ReflectiveSubuniverse.v` (increase cost/decrease priority of inO_contr)

Closes #2227


theories/Modalities/ReflectiveSubuniverse.v
```diff
@@ -762,27 +762,27 @@ Section Reflective_Subuniverse.
...
-    Global Instance inO_unit : In O Unit.
+    #[export] Instance inO_unit : In O Unit.
     Proof.
       apply inO_to_O_retract@{Set} with (mu := fun x => tt).
       exact (@contr@{Set} Unit _).
     Defined.
 
     (** It follows that any contractible type is in [O]. *)
-    Global Instance inO_contr {A : Type} `{Contr A} : In O A.
+    #[export] Instance inO_contr {A : Type} `{Contr A} : In O A | 2.
```
jdchristensen, line 772: Here's the only additional change to this file.

jdchristensen (approved):
Thanks.  I'll let @Alizter take a quick look and merge.

jdchristensen:
Since this was completely straightforward and analogous to the previous PR removing the "Global" keyword, I'll go ahead and merge.

PR #2245, opened 2025-03-08
Replace *rapply, *refine and apply with exact when possible

jdchristensen:
I don't like it when code uses `*rapply term` or `*refine term` when `exact term` also works, so I fixed those.  The code that does this makes sure that each individual change doesn't affect the timing substantially, and I also did overall checks at the end to make sure that the timing difference is within the noise.

Note that I also replaced `nrapply term` and `nrefine term` with `exact term`, which potentially does typeclass search, but since the timing didn't go up, I still think that this is a worthwhile change.

While writing this, I realized that I should do the same for ordinary `apply`, so the last commit handles that.  ...  There were way more changes than I expected here, so I kind of regret tackling this.  The difficulty was in undoing changes made to comments and tactics, and while I checked things twice, there's a chance I missed some.

Note:  I only handled the dependencies of Categories.v and HoTT.v, as well as HoTTBookExercises.v, so not all of Classes got done.


Alizter:
I tried looking into removing further finishing `apply`s. Take for example in `WildCat/Core.v`:
```coq
Definition is1cat_is1cat_strong (A : Type) `{Is1Cat_Strong A}
  : Is1Cat A.
Proof.
  srapply Build_Is1Cat.
  all: intros a b.
  - apply is01cat_hom_strong.
  - apply is0gpd_hom_strong.
  - apply is0functor_postcomp_strong.
  - apply is0functor_precomp_strong.
  - intros; apply GpdHom_path, cat_assoc_strong.
  - intros; apply GpdHom_path, cat_assoc_opp_strong.
  - intros; apply GpdHom_path, cat_idl_strong.
  - intros; apply GpdHom_path, cat_idr_strong.
Defined.
```
Would we deem these "bad style"?

jdchristensen:
> ```coq
>   - apply is01cat_hom_strong.
>   - apply is0gpd_hom_strong.
>   - apply is0functor_postcomp_strong.
>   - apply is0functor_precomp_strong.
>   - intros; apply GpdHom_path, cat_assoc_strong.
>   - intros; apply GpdHom_path, cat_assoc_opp_strong.
>   - intros; apply GpdHom_path, cat_idl_strong.
>   - intros; apply GpdHom_path, cat_idr_strong.
> ```
I don't see how those are related.  As far as I can see, all of those lemmas need to be given arguments, so `apply` is needed.  E.g. if I change the first `apply` to `exact`, it doesn't work.  And if I change the last `apply ...` to `apply GpdHom_path; exact cat_idr_strong.` it doesn't work.

And regarding the idea of enforcing the use of `exact` when the result being applied needs no arguments, I don't think it's worth forcing it on people.  Lots of lemmas do in fact need arguments, but `exact` works because those arguments are implicit, and it's hard for the user to remember that the arguments are implicit.

jdchristensen:
This PR isn't about avoiding using `apply foo` to finish a proof.  It's about avoiding using `apply foo` when `exact foo` works.  The point is that the use of `exact` gives the reader more information, so it's nice to use it when it works.  But I don't think it should be mandatory.

Alizter (approved):

PR #2249, opened 2025-03-11
Spelling and style changes

patrick-nicodemus:
This is a pretty boring and mundane PR which entirely deals with comments. We:
- fix some typos
- make some stylistic changes to comments
- change some comments to Coqdoc comments
- add [-] around some identifiers in comments so that Coqdoc renders them properly
- adds a few new comments

My apologies if reviewing a bunch of typos seems more work than it's worth.

There are a few issues which I didn't want to put in this PR because it's already very large and these issues are more subjective and require some discussion. However if they are requested I can do these things too.
- In some files (like Truncations/Core.v), paper abbreviations (like CORS) are used without explanation. Maybe there should be a comment at the top of each file introducing the acronyms used? I did this in one file.
- I feel that this comment is unclear and needs to be rewritten. I don't think the term "homwise" is helpful for me. https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/WildCat/Core.v#L245
- Should funext be capitalized everywhere?
- Spelling: hprop, hProp or HProp? Is it a big deal if this is consistent? Similarly with hset/hSet
- Stylistically, should we capitalize objects named after people? Abelian group vs abelian group, Cartesian product vs cartesian product. In both of these cases I prefer the capital style.
- A few references in the code to "Homomorphic properties", "homomorphic image", etc. I think the word "homomorphic" is uncommon. Does it mean something other than "relating to homomorphisms"? I think we could use a better word here.
- The file on ring ideals refers to "coprimeness" and "coprimality". It would be better to pick one term and stick to it, maybe?
- I have no idea what this comment means. https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/HIT/Flattening.v#L149
- The terms "coherentification, coherentified, coherentifiable" appear in the comments in a few places and I think this is vague enough that it needs some explication.
- Is "[typally](https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/Pointed/pMap.v#L50)" a common term, as in "typal computation"? It's not a bad coinage but maybe we should use a more common word.

Some American / British spelling inconsistencies I saw in the code: should these be standardized?
- cancelling vs canceling (American English)
- axiomatisation / axiomatization
- specialised / specialized
- hypothesise / hypothesize
- characterize / characterise
- generaliz* / generalis* (generalize, generalization)
- fiber / fibre
- flavour / flavor
- among / amongst
- behavior / behaviour
- normalised / normalized
- judgemental equality / judgmental equality
- uniqueness / unicity (I'm not even sure this is British, unicity is a very obscure word to me)
- parameterize / parametrize

Redundancy issues: Some comment blocks are copypasted several times across many files. I can understand why you would want the documentation to be right there when you're reading the file, but block copypasting the same comment across several files or lines in the same file doesn't seem good. Could we centralize some of this documentation in a README or other file and just point back to it? Just a couple quotes I noticed appearing more than once, if you search for these you'll see a few files that contain the same paragraph:
- We don't want to make UTF-8 notations the default, so we don't export them.
- that's Unicode.
- Quoting David Spivak
- work around bug #5567
- Dwyer and Spalinski
- Coq produces phantom universes

I think some comments are Coqdoc comments which should not be Coqdoc comments. I think TODO's or complaints like 'This is broken because of Coq bug so-and-so, so I had to do this weird hack to get around it.' are better off as inline comments than doc comments because they're not relevant to you unless you're actively working on the code and trying to understand it or maintain it.

Last, there is a good amount of dead, commented-out-code in the library which could be deleted if it is no longer being used as an illustrative example or planning to be uncommented soon.


Alizter:
> My apologies if reviewing a bunch of typos seems more work than it's worth.

Not at all. Such improvements are always welcome.

> There are a few issues which I didn't want to put in this PR because it's already very large and these issues are more subjective and require some discussion. However if they are requested I can do these things too.
> 
> * In some files (like Truncations/Core.v), paper abbreviations (like CORS) are used without explanation. Maybe there should be a comment at the top of each file introducing the acronyms used? I did this in one file.

Yes, this is something we can sort out in perhaps another PR since there are a few ways to go about this. One option is to write a description in the header and yet another is do collect a bibliography. As time goes on, we will formalise more and more papers and not just the HoTT book, so it might make sense to keep track of extra reading material.

> * I feel that this comment is unclear and needs to be rewritten. I don't think the term "homwise" is helpful for me. https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/WildCat/Core.v#L245

Pointwise = between points so homwise is between homs. Feel free to suggest an improvement however.

> * Should funext be capitalized everywhere?

I think in HoTT vernacular funext is pretty standard. The reason we might capitalize it is to literally mean the term.

> * Spelling: hprop, hProp or HProp? Is it a big deal if this is consistent? Similarly with hset/hSet

I prefer hprop in vernacular. I don't know if we use `hProp` anywhere. The type of hprops is called `HProp`.

> * Stylistically, should we capitalize objects named after people? Abelian group vs abelian group, Cartesian product vs cartesian product. In both of these cases I prefer the capital style.

Abel is a funny one in mathematics since it is one of those names that is commonly not capitlised. I think maybe the same for "cartesian". I think for types we have a clear captilisation convention in STYLE.md, but in everyday mathematical vernacular, especially in comments, I would prefer being closer to source material. So if a paper writes Abelian or abelian, then its fine to keep convention with that.

> * A few references in the code to "Homomorphic properties", "homomorphic image", etc. I think the word "homomorphic" is uncommon. Does it mean something other than "relating to homomorphisms"? I think we could use a better word here.

I think "homomorphic image" is really "image of a homomorphism". Not sure about the other usage however. I will need to look up some literature on that usage, I think it is a bit old-fashioned.

> * The file on ring ideals refers to "coprimeness" and "coprimality". It would be better to pick one term and stick to it, maybe?

I think the latter is the correct one to pick.

> * I have no idea what this comment means. https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/HIT/Flattening.v#L149

This file is essentially in a cemetery now. The results on flattening have all been moved to their appropriate files. I wouldn't spend time improving it.

As for the meaning, I can't grok what it was meant to be. I would guess "non-dependently" fits better.

> * The terms "coherentification, coherentified, coherentifiable" appear in the comments in a few places and I think this is vague enough that it needs some explication.

I think it would depend on the context. If you can point to some uses, maybe we can give some better comments.

> * Is "[typally](https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/Pointed/pMap.v#L50)" a common term, as in "typal computation"? It's not a bad coinage but maybe we should use a more common word.

I've seen it used in type theoretic literature. It's not super common. Something like "computes at the type level" or "computation at the type level" is perhaps clearer.

> Some American / British spelling inconsistencies I saw in the code: should these be standardized?
> 
> * cancelling vs canceling (American English)
> * axiomatisation / axiomatization
> * specialised / specialized
> * hypothesise / hypothesize
> * characterize / characterise
> * generaliz* / generalis* (generalize, generalization)
> * fiber / fibre
> * flavour / flavor
> * among / amongst
> * behavior / behaviour
> * normalised / normalized
> * judgemental equality / judgmental equality
> * uniqueness / unicity (I'm not even sure this is British, unicity is a very obscure word to me)
> * parameterize / parametrize

We always prefer American English spellings, but due to being British, some of my Britishisms (and others) may have slipped in. Feel free to fix as you see them. Unicity is an uncommon word, but I think it might be old-fashioned. I don't mind using "uniqueness" instead.

> Redundancy issues: Some comment blocks are copypasted several times across many files. I can understand why you would want the documentation to be right there when you're reading the file, but block copypasting the same comment across several files or lines in the same file doesn't seem good. Could we centralize some of this documentation in a README or other file and just point back to it? Just a couple quotes I noticed appearing more than once, if you search for these you'll see a few files that contain the same paragraph:

We need to tackle these on a case-by-case basis.

> * We don't want to make UTF-8 notations the default, so we don't export them.

These are specific to the category theory library. I think its fine to keep them as it is to remind the reader why a certain choice was made. Having this consolidated elsewhere would be difficult to see or find.

> * that's Unicode.

This seems to be a different comment in both places.

> * work around bug #5567
I think this is fine to keep around, especially if it appears in multiple places. When we check bugs are fixed, find can find all the places we included a workaround.

> * Quoting David Spivak
> * Dwyer and Spalinski

I think this is fine, since it is paraphrasing a specific book. Again with the bibliography idea above, it might be worth including the book in a bibliography too. Not a top priority.

> * Coq produces phantom universes

I prefer to keep this comment, as it is really a TODO for cleaning up these workarounds. At some point we need to remove these universe workarounds as they should have long been fixed.

> I think some comments are Coqdoc comments which should not be Coqdoc comments. I think TODO's or complaints like 'This is broken because of Coq bug so-and-so, so I had to do this weird hack to get around it.' are better off as inline comments than doc comments because they're not relevant to you unless you're actively working on the code and trying to understand it or maintain it.

I may be wrong, but I think at the moment, bot inline and coqdoc comments appear in the documentation. It may be better for TODOs to be inline, but then again sometimes its important for the reader of the docs too.

> Last, there is a good amount of dead, commented-out-code in the library which could be deleted if it is no longer being used as an illustrative example or planning to be uncommented soon.

This needs to be handled on a case-by-case basis. Some code was ported from elsewhere and is dead, that can be removed. Some code is unfinished and is really a TODO with rich information. Others are just workaround due to some Coq bug that need to be addressed and readded. It really depends.

theories/WildCat/Yoneda.v
```diff
@@ -376,7 +376,7 @@ Proof.
   set (fa := (cate_fun f a) (Id a)).     (* Equivalently, [un_opyoneda_0gpd a _ f]. *)
   set (gb := (cate_fun f^-1$ b) (Id b)). (* Equivalently, [un_opyoneda_0gpd b _ f^-1$]. *)
   srapply (cate_adjointify fa gb).
-  (* [opyoneda_0gpd] is defined by postcomposition, so [opyoneda_isretr_0gpd] simplifies both LHSs.*)
+  (* [opyoneda_0gpd] is defined by postcomposition, so [opyoneda_isretr_0gpd] simplifies both LHS's.*)
```
Alizter, line 379: I think it was correct before? Perhaps we can write "left-hand sides" instead.

theories/Homotopy/Hopf.v
```diff
@@ -118,7 +118,7 @@ Proof.
   exact (freudenthal_hspace' (m:=m.+1) X).
 Defined.
 
-(** Here we give a generalization of a result from Eilenberg-MacLane Spaces in Homotopy Type Theory, Dan Licata and Eric Finster.  Their version corresponds to [m = -2] in our version.  Their encode-decode proof was formalized in this library in EMSpace.v until this shorter and more general approach was found. *)
+(** Here we give a generalization of a result from Eilenberg-Mac Lane Spaces in Homotopy Type Theory, Dan Licata and Eric Finster.  Their version corresponds to [m = -2] in our version.  Their encode-decode proof was formalized in this library in EMSpace.v until this shorter and more general approach was found. *)
```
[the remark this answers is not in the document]
jdchristensen: I've always used a space "Mac Lane", based on what I learned a long time ago, but that doesn't mean that it is correct.  Since both are commonly used, I don't think it matters too much what we do.

theories/Spaces/Nat/Core.v
```diff
@@ -1054,7 +1054,7 @@ Definition lt_iff_not_geq {n m} : ~(n >= m) <-> n < m
 
 (** *** Dichotomy of [<>] *)
 
-(** The inequality of natural numbers is equivalent to [n < m] or [n > m]. This could be an equivalence however one of the sections requires funext since we are comparing two inequality proofs. It is therefore more useful to keep it as a biimplication. Note that this is a negated version of antisymmetry of [<=]. *)
+(** The inequality of natural numbers is equivalent to [n < m] or [n > m]. This could be an equivalence; however, one of the sections requires [Funext] since we are comparing two inequality proofs. It is therefore more useful to keep it as a bi-implication. Note that this is a negated version of antisymmetry of [<=]. *)
```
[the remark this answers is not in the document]
jdchristensen: Or "keep is as an iff" would be fine.  While informal, it is clear.  But I have no strong feelings about these.

theories/WildCat/Yoneda.v
jdchristensen: I used to use apostrophes for such things, but lately I do not.  I don't think it matters too much.

jdchristensen:
> yet another is to collect a bibliography.

I like this idea (in another PR).

> > * Stylistically, should we capitalize objects named after people? Abelian group vs abelian group, Cartesian product vs cartesian product. In both of these cases I prefer the capital style.

Both forms are commonly used, and I don't think it's necessary to enforce a standard form.

> > * A few references in the code to "Homomorphic properties", "homomorphic image", etc. I think the word "homomorphic" is uncommon. Does it mean something other than "relating to homomorphisms"? I think we could use a better word here.

I agree that it's uncommon and old fashioned, and can be removed.  "image" is fine.  "Properties of homomorphisms".  Etc.  But not a big deal.

> > * Is "[typally](https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/Pointed/pMap.v#L50)" a common term, as in "typal computation"? It's not a bad coinage but maybe we should use a more common word.
> 
> I've seen it used in type theoretic literature. It's not super common. Something like "computes at the type level" or "computation at the type level" is perhaps clearer.

"Typal" was introduced to replace "propositional equality" with "typal equality", since the equality types aren't necessarily propositions.  I don't think it really caught on, but that's probably the source of this work in the library.  It's fine to replace it with something more clear.

> > Some American / British spelling inconsistencies I saw in the code: should these be standardized?

To make it worse, Canadian spelling is a mix of the two...  I don't think we need to go through and standardize.  That's just work without little benefit.  But I think we should standardize on American English for identifiers.

> > * uniqueness / unicity (I'm not even sure this is British, unicity is a very obscure word to me)

I've seen "unicity" used more and more lately, but I don't like it.

> > Redundancy issues: Some comment blocks are copypasted several times across many files.

I agree with Ali that this would depend on the situation.  I don't might duplicated comments now and then, as they can be very helpful to the reader.

> > I think some comments are Coqdoc comments which should not be Coqdoc comments. I think TODO's or complaints like 'This is broken because of Coq bug so-and-so, so I had to do this weird hack to get around it.' are better off as inline comments than doc comments because they're not relevant to you unless you're actively working on the code and trying to understand it or maintain it.
> 
> I may be wrong, but I think at the moment, bot inline and coqdoc comments appear in the documentation.

Ali is right, all comments appear in the documentation.  And I think that's fine.  But it's probably best to use `(*` for TODO items, so they are less highlighted.

For things I didn't comment on, I agree completely with Ali.

I haven't looked at the PR changes, but I think it's most important to fix actual errors, and not worry about being too pedantic about consistency.  The library has had many authors, and will hopefully have many more.

jdchristensen:
> * I have no idea what this comment means. https://github.com/HoTT/Coq-HoTT/blob/239bcf4ef0f8e8ddbda2ce795214bdb43d8e94c9/theories/HIT/Flattening.v#L149

"but if we define it diindly ..." should be "but if we define it directly ..."  Someone did a search and replace "rect" -> "ind" that went awry.

theories/Algebra/AbSES/BaerSum.v
```diff
@@ -159,15 +159,15 @@ Defined.
 
 (** Our next goal is to prove that the Baer sum is associative.  Rather than showing this directly, we first prove [baer_sum_twist], which says that [abses_baer_sum (abses_baer_sum E F) G = abses_baer_sum (abses_baer_sum G F) E].  The proof of this mimics the proof of commutativity above.  Then we prove associativity by combining this with commutativity. *)
 
-(** The trinary Baer sum of three short exact sequences. *)
+(** The ternary Baer sum of three short exact sequences. *)
```
jdchristensen, line 162: Maybe "ternary" would have been better, but "trinary" is a word, and since the identifies use "trinary" I think we should stick with that in the comments.  And there are examples of "trinary function" in the wild. (This applies here and below as well.)

theories/Categories.v
```diff
@@ -70,7 +70,7 @@ Require HoTT.Categories.Yoneda.
 Require HoTT.Categories.Structure.
 (** ** Fundamental Pregroupoids *)
 Require HoTT.Categories.FundamentalPreGroupoidCategory.
-(** ** Homotopy PreCategory *)
+(** ** Homotopy Precategory *)
```
jdchristensen, line 73: I think "precategory" would be better.   While we aren't consistent, I don't think we want to capitalize all words in headings.

theories/Categories/HomotopyPreCategory.v
```diff
@@ -1,4 +1,4 @@
-(** * Homotopy PreCategory of Types *)
+(** * Homotopy Precategory of Types *)
```
jdchristensen, line 1:
```suggestion
(** * Homotopy precategory of types *)
```

theories/Categories.v
jdchristensen, line 73:
```suggestion
(** ** Homotopy precategory *)
```

theories/Categories/HomotopyPreCategory.v
```diff
@@ -41,7 +41,7 @@ Module HomotopyPreCategoryInternals.
   End homotopy_precategory.
 End HomotopyPreCategoryInternals.
 
-(** ** The Homotopy PreCategory of Types *)
+(** ** The Homotopy Precategory of Types *)
```
jdchristensen, line 44:
```suggestion
(** ** The homotopy precategory of types *)
```

theories/Homotopy/BlakersMassey.v
```diff
@@ -94,7 +96,7 @@ Section GBM.
 
     (** *** Left-hand codes *)
 
-    (** We enhance the HFLL and ABFJ theorems by defining a version of code-left that doesn't depend on one map being surjective. *)
+    (** We enhance the the HFLL and ABFJ theorems by defining a version of code-left that doesn't depend on one map being surjective. *)
```
jdchristensen, line 99:
```suggestion
    (** We enhance the HFLL and ABFJ theorems by defining a version of code-left that doesn't depend on one map being surjective. *)
```

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -583,7 +583,7 @@ Proof.
   exact (is1natural_prewhisker (G:=opyon X) B (opyoneda _ _ _)).
 Defined.
 
-(** The classifying space functor and the fundamental group functor form an adjunction (pType needs to be restricted to the subcategory of 0-connected pTypes). Note that the full adjunction should also be natural in X, but this was not needed yet. *)
+(** The classifying space functor and the fundamental group functor form an adjunction ([pType] needs to be restricted to the subcategory of 0-connected [pType]s). Note that the full adjunction should also be natural in [X], but this was not needed yet. *)
```
jdchristensen, line 586:
```suggestion
(** The classifying space functor and the fundamental group functor form an adjunction ([pType] needs to be restricted to the subcategory of 0-connected pointed types). Note that the full adjunction should also be natural in [X], but this was not needed yet. *)
```

theories/Homotopy/ExactSequence.v
```diff
@@ -126,7 +126,7 @@ Proof.
     exact (concat_p1 _ @ concat_1p _).
 Defined.
 
-(** If Y is a set, then IsComplex is an HProp. *)
+(** If Y is a set, then [IsComplex] is an HProp. *)
```
jdchristensen, line 129:
```suggestion
(** If [Y] is a set, then [IsComplex] is an [HProp]. *)
```

theories/Homotopy/ExactSequence.v
```diff
@@ -145,7 +145,7 @@ Existing Instance conn_map_isexact.
 Definition issig_isexact (n : Modality) {F X Y : pType} (i : F ->* X) (f : X ->* Y)
   : _ <~> IsExact n i f := ltac:(issig).
 
-(** If Y is a set, then IsExact is an HProp. *)
+(** If Y is a set, then [IsExact] is an HProp. *)
```
jdchristensen, line 148:
```suggestion
(** If [Y] is a set, then [IsExact] is an [HProp]. *)
```

theories/Sets/GCHtoAC.v
```diff
@@ -246,7 +246,7 @@ Hypothesis HN_ninject : forall X, ~ InjectsInto (HN X) X.
 Variable HN_bound : nat.
 Hypothesis HN_inject : forall X, InjectsInto (HN X) (power_iterated X HN_bound).
 
-(* This section then concludes the intermediate result that abstractly, any function HN behaving like the Hartogs number is tamed in the presence of GCH.  Morally we show that X <= HN(X) for all X, we just ensure that X is large enough by considering P(N + X). *)
+(* This section then concludes the intermediate result that abstractly, any function [HN] behaving like the Hartogs number is tamed in the presence of GCH.  Morally we show that [X <= HN X] for all [X], we just ensure that X is large enough by considering P(N + X). *)
```
jdchristensen, line 249:
```suggestion
(* This section then concludes the intermediate result that abstractly, any function [HN] behaving like the Hartogs number is tamed in the presence of GCH.  Morally we show that [X <= HN X] for all [X], we just ensure that [X] is large enough by considering P(N + X). *)
```
Not sure whether `P(N + X)` should be in [...] too.

theories/Spaces/Nat/Division.v
```diff
@@ -109,7 +109,7 @@ Proof.
   nrapply nat_dist_sub_r.
 Defined.
 
-(** The divisor is greater than zero when the divident is greater than zero. *)
+(** The divisor is greater than zero when the dividend is greater than zero. *)
```
jdchristensen, line 112: Does "dividend" mean "the number being divided"?  Maybe better to say "the number being divided".

jdchristensen (commented):
Thanks for this!  I approve, once the minor suggestions are taken in to account.

theories/Homotopy/Hopf.v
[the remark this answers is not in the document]
patrick-nicodemus: > I heard a story that MacLane's wife did some of the typesetting and preferred not to put a space due to quirks with the typewriter.

Interesting. I've never heard this story, but it's in the first page of the Wikipedia page about him. You got it backwards, though: he was born MacLane, but his wife found it *more* convenient to insert the space, and that's what stuck around. I would really prefer to say "Mac Lane" because this is the name that appears in all his publications and books.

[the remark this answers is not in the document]
jdchristensen: > I would really prefer to say "Mac Lane" because this is the name that appears in all his publications and books.

He was really inconsistent.  See the attached screenshot for 13 papers where he used "MacLane".
Screenshot from 2025-03-11 15-16-23

(I see I had the alignment wrong in the French one, but it was also "MacLane" there.)

jdchristensen (approved):

PR #2254, opened 2025-03-12
Typeclass hints

patrick-nicodemus:
This PR collects minor fixes relating to the big Global -> export rename. I tried to do this somewhat manually by using a regex to strip universe variables out of the typeclass instances and then diffing the related files but this is too time intensive so I will have to write a proper script to sort and compare them later.

The file HoTT.v relies on the Classes library but does not export it, so after recent changes
hints from the Classes library are no longer added to the typeclass db by HoTT.v. I view this an intentional change, so this is good, but it makes this PR harder because I have to sort out the hints that were supposed to be removed (because they were not exported by HoTT.v) from the ones that were not. These include:

- @orders.lt_transitive
- hints relating to canonical_names.(CutMinus, Compare, SwapOp)
- equality and <= are stable under double negation for types equipped with an apartness relation
- hints about natpair_integers
- hints about integers.IntAbs
- hints about peano_naturals
- hints about orders.naturals
- hints about theory.rationals
- hints about canonical_names.DecRecip
- hints about canonical_names.lattices
- hints about canonical_names.Biinduction

Note that even if a typeclass is defined in the Classes library, it can still have a hint defined in the Homotopy library, so we should still have some of the typeclass hints for these classes.

The original typeclass hint db apparently contained duplicated hints for some hints relating to surreal numbers. Not sure how/why - maybe a file gets exported twice somewhere.

Spaces.Nat.Division doesn't get exported by HoTT, so hints were missing for Division.NatBezout.


jdchristensen:
> The original typeclass hint db apparently contained duplicated hints for some hints relating to surreal numbers. Not sure how/why - maybe a file gets exported twice somewhere.

Spaces/No.v contains:

```coq
Require HoTT.Spaces.No.Core.
Include HoTT.Spaces.No.Core.
```

I think these two lines should be merged into a `Require Export`, like we use everywhere else, apart from the Categories library.

The Categories library uses separate `Require` and `Include` lines a lot in the meta-files, but I don't know why.  @JasonGross , can you explain why this is done instead of just doing `Require Export`?

theories/Algebra/ooGroup.v
```diff
@@ -236,7 +236,7 @@ Section Subgroups.
     exact (grouphom_V incl h @ inverse2 p @ inv_pp _ _ @ whiskerR (inv_V _) _).
   Defined.
 
-  Instance transitive_coset : Transitive in_coset.
+  #[export] Instance transitive_coset : Transitive in_coset.
```
jdchristensen, line 239: Looking at the PR that removed `Global`, I see that all of the Instances in this section were previously marked Global.  I'm curious what process led you to fixing this one but not the others.

JasonGross:
> The Categories library uses separate `Require` and `Include` lines a lot in the meta-files, but I don't know why. @JasonGross , can you explain why this is done instead of just doing `Require Export`?

This is a poor man's version of namespacing, so that everything can be `Functor.*` instead of `Functor.Core.*` or w/e.  This is not important if names are only every used after `Import`ing, but is quite important if you want to use qualified names for disambiguation.

There's an alternative pattern that avoids `Include` but requires a bit of extra bookkeeping, where, e.g., `No.Core` would start with `Module No.` and end with `End No.`

jdchristensen:
@JasonGross Thanks for the explanation.  I don't think we need those shorter names, so we should remove that trick from No.v.  @patrick-nicodemus , can you add that change to this PR?

theories/Algebra/ooGroup.v
patrick-nicodemus: The hints in the typeclass db are sorted by class first, so `Symmetric` and `Reflexive` hints would appear later in the alphabet. As I said, doing this the "lazy way" with a mix of regexes and iff was time consuming and I will need to write a small script to parse the text output and sort the hints. I was planning to use OCaml, but if @JasonGross is interested I can try writing it in Python and submitting it to the coq-scripts repository.

patrick-nicodemus:
Ah, thanks. I wasn't familiar with the command `Print Typeclasses`. I was using `Print HintDb typeclass_instances`, which might have slightly different syntax.

jdchristensen:
> Ah, thanks. I wasn't familiar with the command `Print Typeclasses`. I was using `Print HintDb typeclass_instances`, which might have slightly different syntax.

No, I'm the one who is confused. `Print Typeclasses` prints the typeclasses (not surprisingly!), and not the instances...

jdchristensen:
Maybe this is good?
```sh
coqc ... | grep -o '\w*(level' | sed 's/(.*//' | sort -u
```

jdchristensen:
My suggestion prints all of the words before "(level", so I think it matches what you suggest.

patrick-nicodemus:
That sed one-liner appears to work well.

`theories/Algebra/Groups/Group.v` exports several modules in the `Classes` library:
```
Require Export (hints) Classes.interfaces.abstract_algebra.
Require Export (hints) Classes.interfaces.canonical_names.
(** We only export the parts of these that will be most useful to users of this file. *)
Require Export Classes.interfaces.canonical_names (SgOp, sg_op, MonUnit, mon_unit,
    LeftIdentity, left_identity, RightIdentity, right_identity,
    Inverse, inv, Associative, simple_associativity, associativity, associative_flip,
    LeftInverse, left_inverse, RightInverse, right_inverse, Commutative, commutativity).
Export canonical_names.BinOpNotations.
Require Export Classes.interfaces.abstract_algebra (IsGroup(..), group_monoid, inverse_l, inverse_r,
    IsSemiGroup(..), sg_set, sg_ass,
    IsMonoid(..), monoid_left_id, monoid_right_id, monoid_semigroup,
    IsMonoidPreserving(..), monmor_unitmor, monmor_sgmor,
    IsSemiGroupPreserving, preserves_sg_op, IsUnitPreserving, preserves_mon_unit).
Require Export Classes.theory.groups.
```

There are some hints here that don't have much to do with group theory, such as theorems about apartness.
This is in turn exported by `AbelianGroups.v`. Is this good?

patrick-nicodemus:
I think this is done. There are several *new* hints in master since the original commit where I started removing `Global` annotations, my assumption is that these are due to later changes where new "Export module" lines were added to various files in the library.

patrick-nicodemus:
Okay. I think this is done, though it's possible I overlooked some instances while looking through the diff lists.

jdchristensen:
> There are some hints here that don't have much to do with group theory, such as theorems about apartness.
> This is in turn exported by `AbelianGroups.v`. Is this good?

I'm not sure if this is good.  Is it possible to selectively export hints?  (IIUC, this question is independent of this PR, and the hints that are exported are the same before and after the "Global" changes, right?  But it's still a good question.)

> There are several _new_ hints in master since the original commit where I started removing `Global` annotations, my assumption is that these are due to later changes where new "Export module" lines were added to various files in the library.

Can you give an example of new hints in master and new "Export module" lines?

theories/Classes/orders/lattices.v
```diff
@@ -279,15 +279,15 @@ Section meet_semilattice_order.
   - rewrite <-E. apply meet_lb_r.
   Qed.
 
-  #[export] Instance: forall z, OrderPreserving (z ⊓).
+  #[export] Instance OrderPreserving_left_meet : forall z, OrderPreserving (z ⊓).
```
jdchristensen, line 282: Oh, I didn't know you could have an unnamed instance.  Does Coq just generate a name?

In any case, can you make the new names here and in other places all lowercase?  While we often capitalize type formers and class names, we are supposed to use lowercase when referring to those types in other identifiers (but are not consistent about this).

theories/Classes/theory/apartness.v
```diff
@@ -12,7 +12,7 @@ intros ap e;revert ap.
 apply tight_apart. assumption.
 Qed.
 
-#[export] Instance: forall x y : A, Stable (x = y).
+#[export] Instance apartness_eq_stable: forall x y : A, Stable (x = y).
```
jdchristensen, line 15: `stable_apartness_eq` would fit our pattern better.

theories/Classes/theory/fields.v
```diff
@@ -52,7 +52,7 @@ Proof.
 intros ? E. rewrite <-E. trivial.
 Qed.
 
-#[export] Instance: IsStrongInjective (-).
+#[export] Instance IsStrongInjective_FieldNegation: IsStrongInjective (-).
```
jdchristensen, line 55: Below you used `plus_field` to name the operation, and I think we should follow that pattern: `isstronginjective_negation_field`.

theories/Classes/theory/fields.v
```diff
@@ -67,7 +67,7 @@ repeat (split; try exact _); intros x y E.
   apply symmetry;trivial.
 Qed.
 
-#[export] Instance: IsStrongInjective (//).
+#[export] Instance IsStrongInjective_FieldDivision: IsStrongInjective (//).
```
jdchristensen, line 70: `isstronginjective_recip_field`?  I think `//` denotes the unary reciprocal operation.

theories/Classes/theory/fields.v
```diff
@@ -296,7 +296,7 @@ Section morphisms.
 
   (* We have the following for morphisms to non-trivial strong rings as well.
     However, since we do not have an interface for strong rings, we ignore it. *)
-  #[export] Instance: IsStrongInjective f.
+  #[export] Instance IsStrongInjective_FieldHomomorphism : IsStrongInjective f.
```
jdchristensen, line 299: In addition to lowercase changes, insert underscore: "field_homomorphism", here and elsewhere.

theories/Classes/theory/naturals.v
```diff
@@ -158,20 +158,20 @@ Section borrowed_from_nat.
...
-  #[export] Instance: forall z : N, PropHolds (z <> 0) -> LeftCancellation (.*.) z.
+  #[export] Instance LeftCancellation_plus_nat : forall z : N, PropHolds (z <> 0) -> LeftCancellation (.*.) z.
   Proof.
   refine (from_nat_stmt nat (fun s =>
     forall z : s, PropHolds (z <> 0) -> LeftCancellation mult z) _).
   simpl. exact nat_mult_cancel_l.
   Qed.
 
-  #[export] Instance: forall z : N, PropHolds (z <> 0) -> RightCancellation (.*.) z.
+  #[export] Instance RightCancellation_times_nat : forall z : N, PropHolds (z <> 0) -> RightCancellation (.*.) z.
```
jdchristensen, line 171: `mult` instead of `times`.  (`mul` is also used, but `mult` seems more common in the Classes library.)

theories/Classes/theory/naturals.v
```diff
@@ -202,7 +202,7 @@ Section borrowed_from_nat.
     apply S_inj in E. destruct (S_neq_0 _ E).
   Qed.
 
-  #[export] Instance: ZeroProduct N.
+  #[export] Instance ZeroProduct_N : ZeroProduct N.
```
jdchristensen, line 205: I guess the `N` should remain capitalized here?

theories/Spaces/Nat.v
```diff
@@ -2,4 +2,6 @@
 
 Require Export Nat.Core.
 Require Export Nat.Arithmetic.
+Require Export Nat.Division.
 Require Export Nat.Paths.
+Require Export Nat.Factorial.
```
jdchristensen, line 7: Please add newline, and if possible, tell your editor to always end files with a newline.

jdchristensen (commented):
That's a surprisingly small list of `export` annotations you needed to add!

theories/Classes/orders/lattices.v
patrick-nicodemus: I was not aware of this feature either, I found these while going through the typeclass hints and investigating them. I suppose that it makes sense, if you only want to use it as a typeclass instance without ever writing code to refer to it explicitly, but the auto generated names only refer to the class itself, i.e. `StronglyOrderPreserving_instance_0` where for debugging typeclass resolution failures I would prefer a name that describes also the subject of the predicate, `StronglyOrderPreserving_ring_homomorphism`

patrick-nicodemus:
> > There are several _new_ hints in master since the original commit where I started removing `Global` annotations, my assumption is that these are due to later changes where new "Export module" lines were added to various files in the library.
> 
> Can you give an example of new hints in master and new "Export module" lines?

This comment was incorrect and based on me not understanding how dune works. Dune has a size limited internal buffer on how much it will output on stderr. The reader must disable this with --no-buffer (optionally -j1 if it's important to compile files one at a time so that output from different files is not interleaved). Something like

```
dune build --no-buffer file.vo 2> output.txt
```

should work to get all output from a Coq file (and all dependencies that it needs to rebuild). The output text file contains terminal color control characters for pretty printing which have to be stripped away.

jdchristensen:
Oh, and don't forget about the fix to Spaces/No.v

theories/Classes/orders/lattices.v
Alizter: Yes, you can have unnamed hints, but it is a really fragile "feature" and best if we don't rely on it.

JasonGross:
> Is it possible to selectively export hints?

`Export (hints) mod_name` I believe (it's possible the import categories come after the module name)

jdchristensen:
> `Export (hints) mod_name` I believe

@JasonGross What I meant to ask was whether you can export only *some* of the hints.  The context of the question was that we are exporting all of the hints from Classes, but many of them are irrelevant for group theory, so it would be nice if we could export just some of them.

JasonGross:
As far as I know, you need to put them in separate modules to export only some of them

jdchristensen (approved):
Thanks, this looks good!  @Alizter , do you want to take a look, or should I go ahead and merge this?

Alizter:
@jdchristensen I would like to give it one last look.

Alizter (approved):

PR #2259, opened 2025-03-18
Add things related to suspensions and joins

jdchristensen:
Some miscellaneous things that are useful in some out-of-tree work, and which also simplify some of the technical proofs in the library.  I could have limited the changes to the long proofs to be just a line or two, but while I was looking at them, I simplified them a bit and removed some rewrites.  A couple of slow `Defined` or `Qed` lines became slightly slower, but the difference was fairly small, so I think it's worth it to have proofs without rewrites.

The commits can be read independently or together.


Alizter (approved):

Alizter:
I like the new `ap_pV` lemma and I used it in a few places I recalled. Once you've had a look, feel free to merge.

PR #2263, opened 2025-03-21
Using new syntax for projection instance declarations

patrick-nicodemus:
This commit replaces all `Existing Instance proj1` occurrences in the library with the new `::` syntax.

I have opted not to replace the occurrences of `Coercion` as I didn't realize the `:>` makes it reversible.

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -509,22 +509,33 @@ IsTrunc n (P a)`, and similarly for other data types. For instance,
 
 ### 3.7. Coercions and Existing Instances
 
-A "coercion" from `A` to `B` is a function that Coq will insert
-silently if given an `A` when it expects a `B`, and which it doesn't
-display. For example, we have declared `equiv_fun` as a coercion from
-`A <~> B` to `A -> B`, so that we can use an equivalence as a function
-without needing to manually apply the projection `equiv_fun`.
-Coercions can make code easier to read and write, but when used
-carelessly they can have the opposite effect.
+A ["coercion" from `A` to
+`B`](https://rocq-prover.org/refman/addendum/implicit-coercions.html)
+is a function that Coq will insert silently if given an `A` when it
+expects a `B`, and which it doesn't display. For example, we have
+declared `equiv_fun` as a coercion from `A <~> B` to `A -> B`, so that
+we can use an equivalence as a function without needing to manually
+apply the projection `equiv_fun`. Coercions can make code easier to
+read and write, but when used carelessly they can have the opposite
+effect.
 
-When defining a `Record`, Coq allows you to declare a field as a
-coercion by writing its type with `:>` instead of `:`.
-When defining a `Class`, the `:>` notation has a
-different meaning: it declares a field as an `Existing Instance`.
+When making a record field `proj1` into a coercion, we prefer that the
+coercion should not be reversible, because of the additional
+complexity that reversible coercions add to unification problems, and
+we prefer that coercions be conspicuous when reading the source code.
+Therefore, define the coercion on its own line after the record
+declaration as `Coercion proj1 : MyRec >-> FieldType` rather than
+using the reversible coercion syntax `proj1 :> FieldType`.
 
-(In addition, the `:>` notation is used when needed for path types
-to indicate the type in which the paths are taken: `x = y :> A`
-for `x, y : A`.)
+(This library also uses the `:>` notation when needed for path types
+to indicate the type in which the paths are taken: `x = y :> A` for
+`x, y : A`.)
+
+If `proj1` is a record field whose type is a typeclass, then we prefer
+the concise ["substructure"
+notation](https://rocq-prover.org/refman/addendum/type-classes.html#substructures)
+`proj1 :: ClassType` over the separate vernacular command
+`Existing Instance proj1.` after the record declaration.
 
 ## 4. Axioms
 
--- a/STYLE.md
+++ b/STYLE.md
@@ -531,7 +531,8 @@ using the reversible coercion syntax `proj1 :> FieldType`.
 to indicate the type in which the paths are taken: `x = y :> A` for
 `x, y : A`.)
 
-If `proj1` is a record field whose type is a typeclass and we'd like to add this field as a typeclass instance, then we prefer
+If `proj1` is a record field whose type is a typeclass and we'd
+like to add this field as a typeclass instance, then we prefer
 the concise ["substructure"
 notation](https://rocq-prover.org/refman/addendum/type-classes.html#substructures)
 `proj1 :: ClassType` over the separate vernacular command
--- a/STYLE.md
+++ b/STYLE.md
@@ -531,7 +531,7 @@ using the reversible coercion syntax `proj1 :> FieldType`.
 to indicate the type in which the paths are taken: `x = y :> A` for
 `x, y : A`.)
 
-If `proj1` is a record field whose type is a typeclass, then we prefer
+If `proj1` is a record field whose type is a typeclass and we'd like to add this field as a typeclass instance, then we prefer
 the concise ["substructure"
 notation](https://rocq-prover.org/refman/addendum/type-classes.html#substructures)
 `proj1 :: ClassType` over the separate vernacular command
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -18,11 +18,10 @@ Local Open Scope mc_add_scope.
 
 Record AbGroup := {
   abgroup_group : Group;
-  abgroup_commutative : @Commutative abgroup_group _ (+);
-}.
+  abgroup_commutative :: @Commutative abgroup_group _ (+);
+  }.
 
 Coercion abgroup_group : AbGroup >-> Group.
-Existing Instance abgroup_commutative.
 
 Definition zero_abgroup (A : AbGroup) : Zero A := mon_unit.
 Definition negate_abgroup (A : AbGroup) : Negate A := inv.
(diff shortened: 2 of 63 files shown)


theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -18,11 +18,10 @@ Local Open Scope mc_add_scope.
 
 Record AbGroup := {
   abgroup_group : Group;
-  abgroup_commutative : @Commutative abgroup_group _ (+);
-}.
+  abgroup_commutative :: @Commutative abgroup_group _ (+);
+  }.
```
Alizter, line 22: I generally prefer that the closing curly brace matches the indentation of `Record` since it makes it easier to "match".

I don't think we have a style guide about this however.

Here this whitespace change is unwarranted however.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -129,13 +127,10 @@ Defined.
 
 (** The type (set) of subgroups of a group G. *)
 Record Subgroup (G : Group) := {
-  subgroup_pred : G -> Type ;
-  subgroup_issubgroup : IsSubgroup subgroup_pred ;
+  subgroup_pred :> G -> Type ;
```
Alizter, line 130: This one slipped in it seems.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -441,15 +436,12 @@ Class IsNormalSubgroup {G : Group} (N : Subgroup G)
   := isnormal : forall {x y}, N (x * y) -> N (y * x).
 
 Record NormalSubgroup (G : Group) := {
-  normalsubgroup_subgroup : Subgroup G ;
-  normalsubgroup_isnormal : IsNormalSubgroup normalsubgroup_subgroup ;
+  normalsubgroup_subgroup :> Subgroup G ;
```
patrick-nicodemus, line 439: fixed in next commit.

theories/Algebra/AbGroups/AbelianGroup.v
patrick-nicodemus: This is the Proof General default indentation style, it automatically fires after a line break, this was accidental.

theories/Basics/Overture.v
```diff
@@ -92,11 +92,8 @@ Class Transitive {A} (R : Relation A) :=
 
 (** A [PreOrder] is both Reflexive and Transitive. *)
 Class PreOrder {A} (R : Relation A) :=
-  { PreOrder_Reflexive : Reflexive R | 2 ;
```
Alizter, line 95: This is funny. Back when the syntax was `:>` for an existing instance in a `Class` we had these priorities. When they were removed the parser still accepted and ignored them. Now that we are adding them back in, the priority is back to the original one.

theories/Classes/interfaces/abstract_algebra.v
```diff
@@ -238,48 +205,35 @@ Section lattice.
   Context A {Ajoin: Join A} {Ameet: Meet A} {Abottom : Bottom A} {Atop : Top A}.
 
   Class IsJoinSemiLattice :=
-    join_semilattice : @IsSemiLattice A join_is_sg_op.
-  #[export] Existing Instance join_semilattice.
+    join_semilattice :: @IsSemiLattice A join_is_sg_op.
```
Alizter, line 208: We should really check that `::` has the same intended meaning here. The parser accepts it, but we should inspect the typeclass database to see if it really is being added. Definitional typeclasses may not have the extra same implementation here.

theories/Homotopy/CayleyDickson.v
```diff
@@ -123,20 +112,13 @@ Defined.
...
-  cdi_susp_hspace : IsHSpace (psusp A);
-  cdi_susp_factorneg_r : FactorNegRight (negate_susp A cdi_negate) hspace_op;
-  cdi_susp_conjug_left_inv : LeftInverse hspace_op (conjugate_susp A cdi_negate) mon_unit;
-  cdi_susp_conjug_distr : DistrOpp hspace_op (conjugate_susp A cdi_negate);
+  cdi_negate :: Negate A;
+  cdi_negate_involutive :: Involutive cdi_negate;
+  cdi_susp_hspace :: IsHSpace (psusp A);
+  cdi_susp_factorneg_r :: FactorNegRight (negate_susp A cdi_negate) hspace_op;
+  cdi_susp_conjug_left_inv :: LeftInverse hspace_op (conjugate_susp A cdi_negate) mon_unit;
+  cdi_susp_conjug_distr :: DistrOpp hspace_op (conjugate_susp A cdi_negate)
```
Alizter, line 120: This is also another preference. I like to always end each line of a record with `;` since it makes it easier to reorder lines.

theories/Homotopy/HSpace/Coherent.v
```diff
@@ -11,10 +11,10 @@ Class IsCoherent (X : pType) `{IsHSpace X} :=
   iscoherent : left_identity pt = right_identity pt.
 
 Record IsCohHSpace (A : pType) := {
-    ishspace_cohhspace : IsHSpace A;
-    iscoherent_cohhspace : IsCoherent A;
+    ishspace_cohhspace :: IsHSpace A;
+    iscoherent_cohhspace :: IsCoherent A;
   }.
-#[export] Existing Instances ishspace_cohhspace iscoherent_cohhspace.
+
```
Alizter, line 17: extra space

theories/Classes/interfaces/abstract_algebra.v
patrick-nicodemus: Yep, good call. It works.

```
   Class IsJoinSemiLattice :=
    join_semilattice :: @IsSemiLattice A join_is_sg_op.

  Print Instances IsSemiLattice.
  (* join_semilattice@{} : IsJoinSemiLattice -> IsSemiLattice A *)
```

theories/Basics/Overture.v
patrick-nicodemus: I was pleasantly surprised that this notation was accepted by the parser.

Alizter (commented):
I had an initial look and it mostly looks good. Thanks for taking this on. I will do a more thorough review once my previous comments have been addressed.

theories/Basics/Overture.v
jdchristensen: Does that mean that this change is also a semantic one?  Why should we set the cost of this instance to something other than the default?  (And since this wasn't just a rote change, it should have been highlighted in the PR.)

theories/Homotopy/CayleyDickson.v
jdchristensen: And I'll emphasize once again that you should not change things unrelated to the PR.

theories/Basics/Overture.v
Alizter: The original intention is for the reflexivity instance coming from here to be lower priority than the rest. This was broken some time when we removed the legacy `:>` syntax.

We have two choices: Either we remove the `| 2`s or we keep them the way they are. The secondary option seems better as I think I see the reason for choosing a lower priority. During typeclass search, you don't want to try the reflexivity proof from the preorder class of `R` rather than just choosing a direct reflexivity proof if one is available.

jdchristensen (commented):
Thanks for doing this.  It looks good to me.

Can you also update Section 3.7 of STYLE.md?  It should say that we prefer "::" to "Existing Instance" and that we do *not* want to use ":>" for coercions, in part because it makes them reversible, and in part because we like being explicit about Coercions.  It also has out-of-date info about what ":>" means.

(And another task would be to replace the current uses of ":>" with an explicit Coercion command. They were done without realizing that it made the coercions reversible.)

theories/Basics/Overture.v
patrick-nicodemus: I thought that the original hint had a custom priority of 2. If there was no custom priority in the original hint than this is a mistake and I'll look at it again to see what happened. Maybe I mixed it up with another hint

STYLE.md
```diff
@@ -513,22 +513,33 @@ The conventions for the typeclass `IsTrunc` are:
...
+we prefer that coercions be conspicuous when reading the source code.
+Therefore, define the coercion on its own line after the record
+declaration as `Coercion proj1 : MyRec >-> FieldType` rather than
+using the reversible coercion syntax `proj1 :> FieldType`.
+
+(This library also uses the `:>` notation when needed for path types
+to indicate the type in which the paths are taken: `x = y :> A` for
+`x, y : A`.)
+
+If `proj1` is a record field whose type is a typeclass, then we prefer
```
jdchristensen, line 538:
```suggestion
If `proj1` is a record field whose type is a typeclass and we'd like to add this field as a typeclass instance, then we prefer
```

theories/Basics/Overture.v
jdchristensen: As Ali explained, I think historically it had a custom priority of 2, but when the old `:>` notation for an instance was replaced with an explicit `Existing Instance` line, that priority was lost (but the `| 2` was still in the record field, doing nothing).  Then your change caused the priority to go back to 2.  Ali suggests that this is probably a good thing, so there is no need to change this PR.  I was just pointing out that this is a semantic change, which should have been mentioned, but of course it was the kind of thing that would be easy to miss.

jdchristensen:
This looks good to me, modulo one tweak to STYLE.md.

(And, just to be clear, my comment about removing `:>` was meant as a task for another PR.)

jdchristensen (approved):

jdchristensen:
@patrick-nicodemus Can you rebase or merge with master?

jdchristensen:
@Alizter Did you want to give this another look?

PR #2266, opened 2025-03-23
unboxed HasMorExt and IsUnivalent1Cat

patrick-nicodemus:
This commit "unboxes" `HasMorExt` and `IsUnivalent1Cat`, removing the constructor term necessary to build an instance of these, which will make terms slightly smaller. I guess the performance gains are slightly offset because the definition still has to be unfolded.

Of course the proofs are shorter by one line because one no longer has to apply the constructor.

What do you think, is this a good idea?

@Alizter suggested I use "Typeclasses Opaque" after this, but the documentation says:
> Definitional typeclasses are not wrapped inside records, and the trivial projection of an instance of such a typeclass is convertible to the instance itself. This can be useful to make instances of existing objects easily and to reduce proof size by not inserting useless trivial projections. **The typeclass [constant](https://rocq-prover.org/doc/V9.0.0/refman/language/core/definitions.html#term-constant) itself is declared rigid during resolution so that the typeclass abstraction is maintained.**

And this makes me wonder if the extra "Typeclasses Opaque" command is really necessary.

(There are other singleton classes in the library which could be unboxed, so based on this discussion we could decide to unbox other classes as well.)

### DIFF
--- a/theories/WildCat/Equiv.v
+++ b/theories/WildCat/Equiv.v
@@ -496,8 +496,8 @@ Proof.
 Defined.
 
 Class IsUnivalent1Cat (A : Type) `{HasEquivs A}
-  := { isequiv_cat_equiv_path : forall a b, IsEquiv (@cat_equiv_path A _ _ _ _ _ a b) }.
-Existing Instance isequiv_cat_equiv_path.
+  := isequiv_cat_equiv_path :: forall a b, IsEquiv (@cat_equiv_path A _ _ _ _ _ a b).
+#[export] Typeclasses Opaque IsUnivalent1Cat.
 
 Definition cat_path_equiv {A : Type} `{IsUnivalent1Cat A} (a b : A)
   : (a $<~> b) -> (a = b)
@@ -623,7 +623,6 @@ Instance hasmorext_core {A : Type} `{HasEquivs A, !HasMorExt A}
   `{forall (x y : core A) (f g : uncore x $<~> uncore y), IsEquiv (ap (x := f) (y := g) cate_fun)}
   : HasMorExt (core A).
 Proof.
-  snapply Build_HasMorExt.
   intros X Y f g; cbn in *.
   snapply isequiv_homotopic.
   - exact (GpdHom_path o (ap (x:=f) (y:=g) cate_fun)).
--- a/theories/WildCat/Equiv.v
+++ b/theories/WildCat/Equiv.v
@@ -497,7 +497,6 @@ Defined.
 
 Class IsUnivalent1Cat (A : Type) `{HasEquivs A}
   := isequiv_cat_equiv_path :: forall a b, IsEquiv (@cat_equiv_path A _ _ _ _ _ a b).
-#[export] Typeclasses Opaque IsUnivalent1Cat.
 
 Definition cat_path_equiv {A : Type} `{IsUnivalent1Cat A} (a b : A)
   : (a $<~> b) -> (a = b)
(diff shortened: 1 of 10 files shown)


jdchristensen:
I tried to fix the conflict.  For some reason, git lost a line in the first attempt, so I force pushed a second attempt.  Please check.

I'll let @Alizter comment on the PR in general.  Is there any reason to box a single-constructor class/record?  The only downside I'm aware of is that you can't declare an unboxed class/record as cumulative.

I'm also not sure about whether the opaque setting is needed.

BTW, can you link to where we discussed this unboxing before, just to give context?

theories/WildCat/Equiv.v
```diff
@@ -496,7 +496,8 @@ Proof.
 Defined.
 
 Class IsUnivalent1Cat (A : Type) `{HasEquivs A}
-  := { isequiv_cat_equiv_path : forall a b, IsEquiv (@cat_equiv_path A _ _ _ _ _ a b) }.
+  := isequiv_cat_equiv_path : forall a b, IsEquiv (@cat_equiv_path A _ _ _ _ _ a b).
+#[export] Typeclasses Opaque IsUnivalent1Cat.
 Existing Instance isequiv_cat_equiv_path.
```
jdchristensen, line 501: Switch to `::` notation here?

patrick-nicodemus:
> BTW, can you link to where we discussed this unboxing before, just to give context?

We didn't really discuss it. I made this change alongside other changes I was doing things in this PR https://github.com/HoTT/Coq-HoTT/pull/2262 but we agreed that it was a separate issue and could be left for a separate PR. Ali suggests here: https://github.com/HoTT/Coq-HoTT/pull/2262#issuecomment-2739428877 that the "Set Typeclasses Opaque` might be necessary, but as I said above it's not clear to me.

>  Is there any reason to box a single-constructor class/record? The only downside I'm aware of is that you can't declare an unboxed class/record as cumulative.

There are no downsides highlighted in the refman, outside that I have no idea.

theories/WildCat/Equiv.v
patrick-nicodemus: This was already done in the other big `::` PR, so this is taken care of by rebasing onto master.

patrick-nicodemus:
> I tried to fix the conflict. For some reason, git lost a line in the first attempt, so I force pushed a second attempt. Please check.

Oh, whoops, I didn't see this comment. I'll check now.

patrick-nicodemus:
Alright, I've looked through this and I think it's in good shape, so we just have to discuss whether the `Typeclasses Opaque` command is necessary. I am not really familiar with how this works, but if Ali doesn't know either I'm willing to do another dive into the refman and see if I can make sense of it. :)

jdchristensen:
Is it possible to display the typeclasses database (including transparency info) before and after applying this setting to see if it changes anything?

If we can't figure it out and @Alizter doesn't know for sure whether the typeclasses opaque setting is needed, then I'd say we should just leave it out, and add it later if we notice an issue.

Alizter:
I have vague memories of proofs becoming faster when Is1Natural was changed to be boxed. I don't have time to test that or find where we discussed it however.

jdchristensen:
@Alizter I guess you are talking about #1968?  I didn't see any mention there of speed, and the boxing was done for other reasons.  In any case, if @patrick-nicodemus can do a careful timing test before and after the proposed commit, that would tell us whether there is a difference in this case.

patrick-nicodemus:
Alright. I have inspected the typeclass database and `HasMorExt` and `IsUnivalent1Cat` are both automatically declared as opaque for typeclass resolution, which is consistent with my interpretation of the refman (that this is done for all unboxed singleton classes.) So there is no need for the `Typeclasses Opaque` annotation, I have removed it.

jdchristensen (approved):
LGTM.  @Alizter , do you want to look at it?

Alizter (approved):

PR #2267, opened 2025-03-23
Clean annotations

patrick-nicodemus:
This collects a few changes from the `Hint Mode` rewrite which I deemed sufficiently unobtrusive that I think they are worth submitting again in a separate PR.

Each of these changes is motivated by identifying breakages in the code due to the `Hint Mode +` changes in the WildCat library, I have previously argued that this elaboration behavior is not robust, so my motivation for this PR is that I think these fix potential future bugs due to refactoring (e.g., the current PR I have outstanding to refactor (2,1)-cats so that they depend on the definition of a bicategory)

Some of the changes are clear improvements to the existing code as well and are unlikely to be controversial.

I believe the only change in this PR which could affect downstream users is that I defined one or two new typeclass instances in AbelianGroups.v. There is also what I would call a "naming bug" in this file where some things are called "group" where they should be called "abgroup"

@jdchristensen I tried using Magit this time to organize the changes. :smile:

### DIFF
--- a/theories/Algebra/AbGroups/AbHom.v
+++ b/theories/Algebra/AbGroups/AbHom.v
@@ -2,6 +2,10 @@ Require Import Basics Types.
 Require Import WildCat HSet Truncations.Core Modalities.ReflectiveSubuniverse.
 Require Import Groups.Group Groups.QuotientGroup AbelianGroup Biproduct.
 
+(* If Hint A about groups appears before Hint B about Abelian groups appears in the typeclass database,
+   Coq may coerce an Abelian group to a group to apply A, so we position some hints carefully. *)
+Existing Instance is01cat_hom | 0.
+
 (** * Homomorphisms from a group to an abelian group form an abelian group. *)
 
 (** In this file, we use additive notation for the group operation, even though some of the groups we deal with are not assumed to be abelian. *)
@@ -40,7 +44,7 @@ Defined.
 (** ** Coequalizers *)
 
 (** Using the cokernel and addition and negation for the homs of abelian groups, we can define the coequalizer of two group homomorphisms as the cokernel of their difference. *)
-Definition ab_coeq {A B : AbGroup} (f g : GroupHomomorphism A B)
+Definition ab_coeq {A B : AbGroup} (f g : A $-> B)
   := ab_cokernel ((-f) + g).
 
 Definition ab_coeq_in {A B : AbGroup} {f g : A $-> B} : B $-> ab_coeq f g.
@@ -135,7 +139,7 @@ Proof.
 Defined.
 
 Definition functor_ab_coeq_id {A B : AbGroup} (f g : A $-> B)
-  : functor_ab_coeq (f:=f) (g:=g) (Id _) (Id _) (hrefl _) (hrefl _) $== Id _.
+  : functor_ab_coeq (Id _) (Id _) (hrefl f) (hrefl g) $== Id _.
 Proof.
   snapply ab_coeq_ind_homotopy.
   reflexivity.
@@ -148,13 +152,13 @@ Definition grp_iso_ab_coeq {A B : AbGroup} {f g : A $-> B}
 Proof.
   snapply cate_adjointify.
   - exact (functor_ab_coeq a b p q).
-  - exact (functor_ab_coeq a^-1$ b^-1$ (hinverse _ _ p) (hinverse _ _ q)).
+  - exact (functor_ab_coeq a^-1$ b^-1$ (hinverse a _ p) (hinverse a _ q)).
   - nrefine (functor_ab_coeq_compose _ _ _ _ _ _ _ _
       $@ functor2_ab_coeq _ _ _ _ _ $@ functor_ab_coeq_id _ _).
-    tapply cate_isretr.
+    exact (cate_isretr b).
   - nrefine (functor_ab_coeq_compose _ _ _ _ _ _ _ _
       $@ functor2_ab_coeq _ _ _ _ _ $@ functor_ab_coeq_id _ _).
-    tapply cate_issect.
+    exact (cate_issect b).
 Defined.
 
 (** ** The bifunctor [ab_hom] *)
--- a/theories/Algebra/AbGroups/AbHom.v
+++ b/theories/Algebra/AbGroups/AbHom.v
@@ -207,10 +207,8 @@ Proof.
 Defined.
 
 Instance is0bifunctor_ab_hom `{Funext}
-  : Is0Bifunctor (ab_hom : Group^op -> AbGroup -> AbGroup).
-Proof.
-  rapply Build_Is0Bifunctor''.
-Defined.
+  : Is0Bifunctor (ab_hom : Group^op -> AbGroup -> AbGroup)
+  := Build_Is0Bifunctor'' _.
 
 Instance is1bifunctor_ab_hom `{Funext}
   : Is1Bifunctor (ab_hom : Group^op -> AbGroup -> AbGroup).
--- a/theories/Algebra/AbGroups/AbHom.v
+++ b/theories/Algebra/AbGroups/AbHom.v
@@ -2,10 +2,6 @@ Require Import Basics Types.
 Require Import WildCat HSet Truncations.Core Modalities.ReflectiveSubuniverse.
 Require Import Groups.Group Groups.QuotientGroup AbelianGroup Biproduct.
 
-(* If Hint A about groups appears before Hint B about Abelian groups appears in the typeclass database,
-   Coq may coerce an Abelian group to a group to apply A, so we position some hints carefully. *)
-Existing Instance is01cat_hom | 0.
-
 (** * Homomorphisms from a group to an abelian group form an abelian group. *)
 
 (** In this file, we use additive notation for the group operation, even though some of the groups we deal with are not assumed to be abelian. *)
--- a/theories/Algebra/AbGroups/AbelianGroup.v
+++ b/theories/Algebra/AbGroups/AbelianGroup.v
@@ -145,14 +145,17 @@ Instance isgraph_abgroup : IsGraph AbGroup
 Instance is01cat_abgroup : Is01Cat AbGroup
   := is01cat_induced abgroup_group.
 
-Instance is01cat_grouphomomorphism {A B : AbGroup} : Is01Cat (A $-> B)
-  := is01cat_induced (@grp_homo_map A B).
-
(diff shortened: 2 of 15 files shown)


jdchristensen:
@Alizter Any idea what's up with the build errors?
```
make: coq_makefile: No such file or directory
```

theories/WildCat/DisplayedEquiv.v
```diff
@@ -30,7 +30,7 @@ Class DHasEquivs {A : Type} `{HasEquivs A}
     DGpdHom (cate_issect' _ _ f) (dcate_inv' f' $o' dcate_fun f') (DId a');
   dcate_isretr' : forall {a b} {f : a $<~> b} {a'} {b'} (f' : DCatEquiv f a' b'),
     DGpdHom (cate_isretr' _ _ f) (dcate_fun f' $o' dcate_inv' f') (DId b');
-  dcatie_adjointify : forall {a b} {f : a $-> b} {g : b $-> a}
+  dcatie_adjointify : forall {a b : A} {f : a $-> b} {g : b $-> a}
```
jdchristensen, line 33: I see some earlier lines with unannotated `a, b, a', b'`, and maybe it's worth being consistent?

theories/WildCat/Monoidal.v
```diff
@@ -255,7 +255,7 @@ Section SymmetricBraid.
     : F a b $<~> F b a
     := Build_CatEquiv (braid a b).
 
-  Definition moveL_braidL a b c f (g : c $-> _)
+  Definition moveL_braidL (a b c : A) f (g : c $-> _)
```
jdchristensen, line 258: It's unrelated to this change, but I wonder if `a, b, c` should be implicit, here and below?

theories/Algebra/AbGroups/AbelianGroup.v
```diff
@@ -145,15 +145,18 @@ Instance isgraph_abgroup : IsGraph AbGroup
 Instance is01cat_abgroup : Is01Cat AbGroup
   := is01cat_induced abgroup_group.
 
-Instance is01cat_grouphomomorphism {A B : AbGroup} : Is01Cat (A $-> B)
+Instance is2graph_abgroup : Is2Graph AbGroup
+  := is2graph_induced abgroup_group.
+
+Instance isgraph_abgrouphomomorphism {A B : AbGroup} : IsGraph (A $-> B)
+  := is2graph_induced abgroup_group A B.
```
jdchristensen, line 152: These two are expressing the same thing.  Do we usually make instances for both?  @Alizter ?

theories/WildCat/Core.v
```diff
@@ -23,6 +23,7 @@ Class Is01Cat (A : Type) `{IsGraph A} :=
   cat_comp : forall (a b c : A), (b $-> c) -> (a $-> b) -> (a $-> c);
 }.
 
+Arguments Build_Is01Cat A &.
```
jdchristensen, line 26: A comment about both the meaning and the motivation for this would be good.  I only found two other places we use the `&` feature, and one of them says:
```
(** The [&] tells Coq to use the context to infer the later arguments. *)
```
(with an irrelevant part stripped out).  You could use the same, but it would also be nice to briefly say why this is helpful.  And maybe in response to this comment you can give a slightly more detailed reason?  (Sorry if we discussed this before.)

theories/Algebra/AbGroups/AbHom.v
```diff
@@ -2,6 +2,10 @@ Require Import Basics Types.
 Require Import WildCat HSet Truncations.Core Modalities.ReflectiveSubuniverse.
 Require Import Groups.Group Groups.QuotientGroup AbelianGroup Biproduct.
 
+(* If Hint A about groups appears before Hint B about Abelian groups appears in the typeclass database,
+   Coq may coerce an Abelian group to a group to apply A, so we position some hints carefully. *)
```
jdchristensen, line 6: - No line break.
- Remove second "appears".
- "before" is ambiguous.  I believe the most recently declared hints are used first (when of the same priority).

So how about:
```
(** If Hint A about groups is added to the typeclass database after Hint B about Abelian groups, Coq may coerce an Abelian group to a group to apply A, so we re-add the following hint to give it priority. *)
```
But I don't understand how the comment relates to the next line, which isn't about groups or abelian groups.  And looking at the order of imports, material about groups is imported before material about abelian groups, so I would expect the hints to be in a good order already.

jdchristensen (commented):
Thanks for remembering these cleanups, and for organizing them so well.  This was very easy to review.

Alizter:
> @Alizter Any idea what's up with the build errors?
> 
> ```
> make: coq_makefile: No such file or directory
> ```

Rocq 9.0 has been released and the github action bumped `latest` to point to there. For now lets ignore this error. I will try fix the CI in the coming days.

theories/WildCat/Core.v
patrick-nicodemus: I think that comment might have it backwards, although I'm not 100% clear on this. I think we are using context to infer the arguments occurring to the left of `&`, rather than to the right. In short, if you write 
`Definition My01Cat : Is01Cat X := Build_Is01Cat _ _ id comp` it will fill in the first hole using the *context* of the application (here, the expected type is `Is01Cat X`, so it will plug in `A :=X`) rather than trying to solve for `X` from the record fields `id` and `comp`.

I did some experiments with this a while ago and this doesn't seem always beneficial because coercions make unification "path-dependent" - changing the order of operations can cause a coercion to be inserted that leads to a different answer and a unification failure. However, for the case of record constructors, it generally seems like it's beneficial to put the `&` before the arguments and after the parameters, because when you're constructing an element of a record type the parameters can usually be inferred from context (as opposed to the fields of the record.) I think @JasonGross made a pull request asking for a way to make this a default behavior, so maybe he will be willing to explain why this seems to work well for record types in particular.

~~~~

Here is a longer answer if you are interested. This link is the most thorough description I could find, but I'll try to explain in my own words
https://proofassistants.stackexchange.com/questions/1343/how-do-coqs-bidirectionality-hints-affect-type-checking.

Type checking contains the following components:
1. Compute the type of a term ("type synthesis")
2. Given a term and a type, validate that the term is of the given type ("type checking")
3. Given two terms, unify them (possibly instantiating metavariables)

These operations all call each other recursively. The Stack Exchange comment considers the example of typechecking a definition `Definition abc : T := f s t`, where `f` is a constant whose type is declared to be 
`forall (x:A), B x -> C x`. At the highest level, to *type check* this definition, we to compute the type of the term `f s t` (synthesis) and unify it with `T`. To compute the type of a function application, we *check* that `s` is of type `A` and *check* that `t` is of type `B s`; if this is successful, then we return the type of `f s t` as `C s`, and finally try to unify `C s` with `T`.  (So type synthesis invokes type checking, and vice versa.)

The bidirectionality hint & changes the order of operations in this algorithm. If we do
```
Arguments f x & y.
```
then we check that `s` is of type `A`, *then* unify `C s` with `T`, and finally check that `t` is of type `B s`.

Because the unification of `C s` with `T` is happening earlier in the process, you are potentially instantiating metavariables in `s` now (using the expected type `T`) that are helpful in checking that `t` is of type `B s`.

theories/Algebra/AbGroups/AbelianGroup.v
Alizter: Sometimes it can make sense to have both, but usually in the other order and the 2-graph instance being inferred from the graph on the hom. Typically we do that when the hom can be generalised a bit more as a "forall".

theories/WildCat/Core.v
jdchristensen: According to the [documentation](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#bidirectionality-hints), the comment is correct.  For example, here is a quote from the documentation:

> An [Arguments](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#coq:cmd.Arguments) command containing [arg_specs1](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#grammar-token-arg_specs) & [arg_specs2](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#grammar-token-arg_specs) provides bidirectionality hints. It tells the typechecking algorithm, when type checking applications of [qualid](https://rocq-prover.org/doc/master/refman/language/core/modules.html#grammar-token-qualid), to first type check the arguments in [arg_specs1](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#grammar-token-arg_specs) and then propagate information from the typing context to type check the remaining arguments (in [arg_specs2](https://rocq-prover.org/doc/master/refman/language/extensions/arguments-command.html#grammar-token-arg_specs)).

jdchristensen: I did some experimenting with `Build_pMap` in Pointed/Core.v.  A while ago, I added this line:
```coq
Arguments Build_pMap & _ _ _ _.
```
My motivation was to let Coq deduce the 1st and 2nd arguments from the context, and this works, justifying the comment  "The [&] tells Coq to use the context to infer the later arguments."  But in contrast to what that comment says, *all* of the following work:
```coq
Arguments Build_pMap & _ _ _ _.
Arguments Build_pMap _ & _ _ _.
Arguments Build_pMap _ _ & _ _.
```
while all of the following fail:
```coq
Arguments Build_pMap _ _ _ & _.
Arguments Build_pMap _ _ _ _ &.
Arguments Build_pMap _ _ _ _.
```
When I say "work", I mean that the first three allow `Build_pMap _ _ f p` to work in many situations.  Just try building the library with that one line changed and you can see the failures.

So I think any summary of the form "infer the earlier/later arguments from the context" is misleading.  We should say "unify with the typing context after type checking the arguments before the [&]", or something like that.  To be honest, I'm not really sure why the first two work.  Maybe it would be more "correct" to use the third one, even though all three work?

jdchristensen: The failure happens in pModality.v.  But if I replace args 1 and 2 with underscores in other places, I can make many failures occur with the last three options, while all of the first three options work.  I think I'll make a PR where I move the & to after the second arg, update the comment, and change some args to underscores.  But I'll wait to hear ideas of what the comment should say and what the "correct" location of the & is.

jdchristensen:
I think the changes required here are pretty minor, so maybe we can get this PR wrapped up?

jdchristensen (approved):
This LGTM!  @Alizter , do you want to look at it, or should I go ahead and merge?

Alizter (approved):
LGTM. I just didn't have enough time to think about the bidirectionality hints but it looks like you've both investigated their behaviour. I'll trust that things are working as expected.

PR #2268, opened 2025-03-23
ci: fix docker-coq-action to 1.5.1

Alizter:


Alizter:
So the explanation for the breakage is as follows:

1. docker-coq-action 1.5.2 switched to version 9.0 for the latest coq/rocq.
2. This means that coq_makefile has been renamed and is no longer available.
3. We would have expected to have seen this issue caught by the `dev` image, however there are extra compat shims installed there allowing for `coq_makefile` to be available as a wrapper.

The solution is either that we fix the version as we have done now or find a way for the compat shims to be installed. There is also upstream discussion so perhaps a new version of the action would become available.

Alizter:
Here is the Zulip thread:

https://coq.zulipchat.com/#narrow/channel/237656-Coq-devs-.26-plugin-devs/topic/docker-coq-action.201.2E5.2E2.20latest

jdchristensen (approved):
I'm happy with merging this for now, and then waiting to see how things develop upstream.

Alizter:
OK let's merge this for now, we can always revert this commit when we work out a better fix.

PR #2270, opened 2025-03-24
Make A and B implicit arguments in Build_{pMap,pEquiv,pEquiv'}; add/adjust bidirectionality hint &

jdchristensen:
The bidirectionality hint for `Build_pMap` and `Build_pEquiv'` is changed from `& _ _ _ _` to `_ _ & _ _`.  Both work, but the new way expresses the intention more clearly.  The same hint is added to `Build_pEquiv`.

The comments near two of these are updated to be more accurate.  (The third is in the same file as one of the others, so I didn't add a comment there.)

Because these hints work so well, almost every use in the library doesn't need arguments one and two, so they have been made implicit.  (In the case of `Build_pEquiv'`, they had already been declared implicit, but the `Arguments` command accidentally undid this.)

### DIFF
--- a/theories/Algebra/AbSES/Core.v
+++ b/theories/Algebra/AbSES/Core.v
@@ -378,7 +378,7 @@ Definition pmap_abses_const {B' A' B A : AbGroup@{u}} : AbSES B A -->* AbSES B' 
 
 Definition to_pointed `{Univalence} {B' A' B A : AbGroup@{u}}
   : (AbSES B A -->* AbSES B' A') -> (AbSES B A ->* AbSES B' A')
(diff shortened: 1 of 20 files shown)


jdchristensen:
This comes from discussion with @patrick-nicodemus in #2267.

jdchristensen:
I'm guessing there are other Records and probably other definitions where bidirectional type checking would help us, so we should keep an eye open for such things.

jdchristensen:
@patrick-nicodemus Let me know if you have any comments on this.

jdchristensen:
@Alizter Any comments on this?  It's a pretty straightforward change, so I'll go ahead and merge if I don't hear anything.

Alizter:
Seems fine to me. Could you rebase for a fresh CI before meging?

PR #2276, opened 2025-04-09
Make f and g explicit arguments to isequiv_compose

jdchristensen:
`isequiv_compose` had `f` and `g` as implicit arguments, just because of the way it was defined with implicit generalization of arguments.  That meant that it was often used in an unwieldy `@isequiv_compose` form with lots of underscores.  This PR makes `f` and `g` into explicit arguments, which makes many places cleaner, although it means that `apply` has to change to `rapply` in a few other places.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -30,7 +30,9 @@ Instance reflexive_equiv : Reflexive Equiv | 0 := @equiv_idmap.
 Arguments reflexive_equiv /.
 
 (** The composition of equivalences is an equivalence. *)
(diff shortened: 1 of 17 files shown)


Alizter (approved):
This seems like a sensible change. I think its always best to reduce the number of ways we can do a single thing.

jdchristensen:
Right, I should have mentioned in the comment that I removed `isequiv_compose'`, which makes `f`, `g` and the proofs that they are equivalences all explicit.  This seems rarely to be helpful, so wasn't worth complicating the code for.

PR #2277, opened 2025-04-10
Add and use concat_l, concat_r and concat_lr

jdchristensen:
I noticed that a lot of proofs in Paths.v were essentially reproving isequiv_concat_lr using isequiv_compose.  This PR directly uses isequiv_concat_lr for those results.  While doing this, I thought it would be cleaner to explicitly define concat_l, concat_r and concat_lr (corresponding to equiv_concat_l, equiv_concat_r and equiv_concat_lr).  We already had a notation concatR, but in all of the places it was used, rhs_V worked better.  So the first commit removes concatR, and the next two commits add the new definitions and then use them.

The fourth commit is independent, but similar in spirit.

The library builds after each commit.

### DIFF
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -62,8 +62,6 @@ Local Open Scope path_scope.
 
 (** ** The 1-dimensional groupoid structure. *)
 
-(** [concat], with arguments flipped. Useful mainly in the idiom [apply (concatR (expression))]. Given as a notation not a definition so that the resultant terms are literally instances of [concat], with no unfolding required. *)
-Notation concatR := (fun p q => concat q p).
 
 (** The identity path is a right unit. *)
 Definition concat_p1 {A : Type} {x y : A} (p : x = y) :
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -76,6 +76,11 @@ Definition concat_lr {A : Type} {w x y z : A} (p : w = x) (r : y = z)
   : (x = y) -> (w = z)
   := fun q => p @ q @ r.
 
+(** The "/" indicates that these should be unfolded by [cbn] and [simpl] when the specified arguments are given. *)
+Arguments concat_l {A x y z} p q /.
+Arguments concat_r {A x y z} q p /.
+Arguments concat_lr {A w x y z} p r q /.
+
 (** The identity path is a right unit. *)
 Definition concat_p1 {A : Type} {x y : A} (p : x = y) :
   p @ 1 = p
--- a/theories/Basics/PathGroupoids.v
+++ b/theories/Basics/PathGroupoids.v
@@ -62,6 +62,19 @@ Local Open Scope path_scope.
 
 (** ** The 1-dimensional groupoid structure. *)
 
+(** Partially applied versions of [concat].  The first is a synonym for [concat p], but we include it to parallel the second one. *)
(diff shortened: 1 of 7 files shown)


Alizter (approved):

theories/Basics/PathGroupoids.v
```diff
@@ -62,8 +62,24 @@ Local Open Scope path_scope.
...
+  := fun p => p @ q.
+
+(** The operation of composing a path on two sides. *)
+Definition concat_lr {A : Type} {w x y z : A} (p : w = x) (r : y = z)
+  : (x = y) -> (w = z)
+  := fun q => p @ q @ r.
+
+(** The "/" indicates that these should be unfolded by [cbn] and [simpl] when the specified arguments are given. *)
+Arguments concat_l {A x y z} p q /.
+Arguments concat_r {A x y z} q p /.
```
jdchristensen, line 81: Hmm, this doesn't work like I hoped.  `simpl` unfolds `concat_r q p` as hoped, but unfortunately unfolds it all the way down to a match.  I thought it would unfold it to `p @ q` (notation for `concat p q`).  `concat` is marked as `simpl nomatch` and `simpl` doesn't unfold it in this case.  Any idea what's going on?  Why does `simpl` do both steps?
```coq
Require Import Basics.

Definition test {A : Type} {x y : A} (p : x = y)
  : concat_r p 1 = p.
Proof.
  simpl.
  (* bad: match p in ... = p *)
  Undo.
  cbn.
  (* good: 1 @ p = p *)
  Undo.
  unfold concat_r.
  (* good: 1 @ p = p *)
  simpl.
  (* still good: 1 @ p = p *)
Abort.
```

Alizter: I think this is where simpl and cbn differ. Simpl has some heuristics that may cause bigger unfolds.

jdchristensen: Just to clarify, the issue is not about `simpl` vs `cbn`, but about how `simpl` isn't consistent with itself.  It unfolds `concat_r -> concat -> match`, but doesn't unfold `concat -> match`.

jdchristensen: And the issue is that if I don't include the above Arguments commands, then I need to adjust several places in the library to make it build.  But if I do, then when I use `simpl` I get `match` appearing in some goals and interfering with some proofs.  (And this is in a context where `cbn` creates a terrible mess, so I can't use it.)

Alizter: Does marking these as `simpl nomatch` work?

jdchristensen: No, then `simpl` doesn't unfold it at all:
```coq
Arguments concat_r {A x y z} q p / : simpl nomatch.
  simpl.
  (* bad in a different way:  concat_r p 1 *)
  Undo.
  cbn.
  (* good: 1 @ p = p *)
  Undo.
```

Alizter: I think this is a bug. `cbn` appears to treat `/` and `simpl nomatch` together fine, but `simpl` is missing the `nomatch` in its heuristic somewhere.

jdchristensen: I've asked about this on the [Rocq zulip](https://rocq-prover.zulipchat.com/#narrow/channel/237977-Rocq-users/topic/simpl.20unfolds.20more.20than.20expected), using a minimal example involving nat instead of paths.

PR #2274, opened 2025-04-09
Formalisation of bar induction axioms, first steps towards formalising the fan theorem and its consequences

thchatzidiamantis:

PR #2278, opened 2025-04-11
Improvements to Truncations/Core

jdchristensen:
The makes some changes that were originally in https://github.com/jdchristensen/HoTT/pull/39 but didn't make it into #2274.

The main thing is that `decidable_trunc_decidable` is added (as an instance), which abstracts out something done in BarInduction.v.  One other result is generalized, and a few proofs are simplified.

Since some things get moved around, it's best to read this one commit at a time.  The commits that move things don't make any changes.

@thchatzidiamantis Can you look this over as well?

### DIFF
--- a/theories/Truncations/Core.v
+++ b/theories/Truncations/Core.v
@@ -200,11 +200,8 @@ Notation "A \/ B" := (hor A B) : hprop_scope.
 
 Definition himage {X Y} (f : X -> Y) := image (Tr (-1)) f.
 
-Definition contr_inhab_prop {A} `{IsHProp A} (ma : merely A) : Contr A.
-Proof.
-  refine (@contr_trunc_conn (Tr (-1)) A _ _); try assumption.
-  exact (contr_inhabited_hprop _ ma).
-Defined.
+Definition contr_merely_inhabited_hprop {A} `{IsHProp A} (ma : merely A) : Contr A
+  := contr_inhabited_hprop _ (tr^-1 ma).
 
 (** A stable type is logically equivalent to its (-1)-truncation. (It follows that this is true for decidable types as well.) *)
 Definition merely_inhabited_iff_inhabited_stable {A} {A_stable : Stable A}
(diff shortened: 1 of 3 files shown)


thchatzidiamantis:
Looks good to me, the other places I use this are in proofs I haven't made a PR for yet, so I don't think anything else needs to be changed.

theories/Truncations/Core.v
```diff
@@ -347,16 +347,14 @@ Defined.
 (** ** Tactic to remove truncations in hypotheses if possible *)
 
 Ltac strip_truncations :=
-  (** search for truncated hypotheses *)
   progress repeat
-    match goal with
-    | [ T : _ |- _ ]
+    match goal with | [ T : _ |- _ ]
```
Alizter: Very minor, but I think I prefer the bar to start on a new line.

Alizter (approved):
Great! Just one minor comment, but feel free to merge.

theories/Truncations/Core.v
jdchristensen: Sure, fixed.

theories/Truncations/Core.v
```diff
@@ -200,21 +240,8 @@ Notation "A \/ B" := (hor A B) : hprop_scope.
...
-Definition merely_inhabited_iff_inhabited_stable {A} {A_stable : Stable A}
-  : Tr (-1) A <-> A.
-Proof.
-  refine (_, tr).
-  intro ma.
-  apply stable; intro na.
-  revert ma; rapply Trunc_ind; exact na.
-Defined.
+Definition contr_merely_inhabited_hprop {A} `{IsHProp A} (ma : merely A) : Contr A
+  := contr_inhabited_hprop _ (tr^-1 ma).
```
jdchristensen, line 244: I'm now thinking that we should remove this result.  When `A` is an hprop, `A` and `merely A` are equivalent, so it doesn't seem like this is likely to be useful.  The one place it had been used was using it in a wasteful way, giving an element of `merely A` of the form `tr a`; so it was better of using `contr_inhabited_hprop` directly.

jdchristensen: I see a thumbs up, so I will do this.  I'll also squash a few commits, and then will merge.

PR #2280, opened 2025-04-28
Cofiber: add isconnmap_isconnected_cofiber, converse of isconnected_cofiber

jdchristensen:
The main goal here is in the last commit, which adds a result saying that if `f : X -> Y` is a map with 1-connected domain whose cofibre is n.+1 connected, then f is n-connected.  To make this go through smoothly, I tweaked many things in the library.  The two biggest changes are in BlakersMassey:

- In the results about pushouts, instead of assuming that both maps are surjective, I only assume that `g` is.  I needed this to handle the inductive step uniformly.

- Unfortunately, the way it worked out for cofibres, I needed the opposite case, where only `f` is assumed to be surjective.  I could have reversed the pushout that defines the cofibre to make this work out, but it does seem slightly more natural in the current form, so instead I also added the result about the flipped pushout, using symmetry of pushouts and pullbacks.

Since some of the commits to BlakersMassey.v overlap, it might be easiest to read them separately.  The library builds after each commit.

This all builds on Ali's nice work in #2171.

### DIFF
--- a/theories/Homotopy/BlakersMassey.v
+++ b/theories/Homotopy/BlakersMassey.v
@@ -45,7 +45,7 @@ Defined.
 Section GBM.
     Context {X Y : Type} (Q : X -> Y -> Type).
 
-    (** Here's the hypothesis of ABFJ's generalized Blakers-Massey.  It works for any reflective subuniverse, not only modalities! *)
+    (** Here's the hypothesis of ABFJ's generalized Blakers-Massey theorem.  It works for any reflective subuniverse, not only modalities!  In the application, [O] will be [Tr (n +2+ m)] and we'll use [isconnected_join] to verify [isconnected_cogap]. *)
     Context (O : ReflectiveSubuniverse).
     Context
       (isconnected_cogap :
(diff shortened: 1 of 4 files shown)


theories/Homotopy/BlakersMassey.v
```diff
@@ -45,7 +45,7 @@ Defined.
 Section GBM.
     Context {X Y : Type} (Q : X -> Y -> Type).
 
-    (** Here's the hypothesis of ABFJ's generalized Blakers-Massey.  It works for any reflective subuniverse, not only modalities! *)
+    (** Here's the hypothesis of ABFJ's generalized Blakers-Massey theorem.  It works for any reflective subuniverse, not only modalities!  In the application, [O] will be [Tr (n +2+ m)] and we'll use [isconnected_join] to verify [isconnected_cogap]. *)
```
Alizter, line 48: One thing I never quite groked: Is ABFJ's definition of a modality the same as a reflective universe? Is this really a more general statement that we have here?

Alizter (approved):

theories/Homotopy/BlakersMassey.v
jdchristensen: I'm pretty sure that the ABFJ definition of a modality is the same as our definition of modality.  RSS has four equivalent descriptions of a modality, and one is as a stable orthogonal factorization system, which is what ABFJ use, IIRC.  The theorem goes through more generally for a reflective subuniverse.  But reflective subuniverses don't have a nice characterization in terms of factorization systems, which means that they are less natural from ABFJ's point of view.

PR #2281, opened 2025-05-06
Generalize lhs and rhs tactics to other relations

jdchristensen:
I created `lhs'`, `rhs'`, `lhs_V'` and `rhs_V'` which are like the unprimed versions, but work for any transitive relation.  (Symmetry is also needed in two cases.)  The main use case is for reasoning in a wild category or with homotopies of (pointed) functions, as demonstrated in the second and third commits.

Since `paths' is a transitive and symmetric relation, these also work for paths.  If the existing tactics are replaced with these, then almost all of the library builds fine, with just a few lines failing.  I tried very hard to tweak these so that they could completely replace the existing tactics, but I couldn't find a way to make them as robust, so I decided to create these primed versions.

The core issue is that we start with `@transitivity A R _`, which is different from `@concat A`.  We apply `cbn` to what we have, which does reduce it to `@concat A'`, but in some cases `A'` is different from `A`, which causes issues.  I tried many things but had no luck.

### DIFF
--- a/theories/Basics/Tactics.v
+++ b/theories/Basics/Tactics.v
@@ -365,11 +365,16 @@ Tactic Notation "stapply'" uconstr(term)
   := do_with_holes' ltac:(fun x => srefine x) term.
 
 (** Apply a tactic to one side of an equation.  For example, [lhs rapply lemma].  [tac] should produce a path. *)
-
 Tactic Notation "lhs" tactic3(tac) := nrefine (ltac:(tac) @ _).
 Tactic Notation "lhs_V" tactic3(tac) := nrefine (ltac:(tac)^ @ _).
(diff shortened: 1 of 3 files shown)


jdchristensen:
@ThomatoTomato These should help with what you are working on.  Can you review this?

ThomatoTomato:
This looks amazing! I guess it is good to have two different versions if they behave differently, but I find it very strange that they are different to begin with...

jdchristensen:
> Very nice! It seems you managed to avoid the pitfall I fell into by introducing a new version. When I attempted it I had generalised `lhs` directly which had caused universe issues.

IIUC, `etransitivity` uses `cbn` precisely to get rid of the extra universe variable that appears in the transitivity type class.  In any case, no extra universe variables arose when I used these in place of the unprimed versions.  The (rare) errors had to do with different amounts of unfolding.

PR #2282, opened 2025-05-22
Miscellaneous improvements

jdchristensen:
Each commit is independent, and the library builds after each commit.

The fifth commit, changing Connectedness.v, is the only one with any substance.  I noticed that `is0connected_merely_allpaths` can be generalized to any truncation level.  I also added an `isconnmap` version of this and also of the converse statement.

The first commit moves a result that was in a strange place, interrupting the flow of other related results.

The others are pretty self-explanatory; see the commit messages.


jdchristensen:
@Alizter These are all pretty straightforward, but if you think it would be better to separate one or two of these commits into its own PR, let me know.

Alizter:
Sorry it took a while for me to take a look at this. LGTM.

PR #2292, opened 2025-06-25
Category/Objects: use path_contr

jdchristensen:
I should have noticed when reviewing #2291 that the `unique` tactic is essentially reproving `path_contr` each time.

@CharlesCNorton I think you'll find `path_contr` useful in other places.  Can you give this PR a quick review?


CharlesCNorton:
Tested locally - changes work correctly. LGTM.

PR #2288, opened 2025-06-16
Add stable category theory formalization

CharlesCNorton:
This formalization includes:
- Pre-stable and proper stable categories with suspension (Σ) and loop (Ω) functors
- Natural transformations η : 1 → Ω∘Σ and ε : Σ∘Ω → 1 with adjunction structure
- Distinguished triangles and the axioms of triangulated categories (TR1, TR2)
- The fundamental duality principle: opposite of stable is stable with Σ and Ω swapped
- Basic cone/cofiber construction for completing morphisms to triangles
- Extensive theory of semi-stable categories and triangle identities

Known limitations:
- No shift/translation functor [1] that shifts degrees
- No cone sequences beyond basic cone construction  
- No proof that rotation of distinguished triangles preserves the distinguished property in all cases
- No completion of morphisms to distinguished triangles beyond the basic framework
- No concrete examples (stable homotopy category, derived categories, etc.)
- Uses traditional Categories library instead of newer WildCat framework

Technical details:
- Currently compiles with JsCoq 0.10.0~beta1 and Coq 8.10+beta2 (July 2019)
- Requires refactoring for compatibility with current Coq and HoTT versions
- ~6000 lines covering foundations through advanced duality theorems

Opening as draft PR per discussion with @Alizter for collaborative updating to latest versions.

Rest added by @jdchristensen :  Files to review.  A checkmark indicates that at least one person has *reviewed* it, not that the comments have been taken into account.  That could be indicated by a strikethrough, or something like that.

- [x] ~~ZeroObjects.v~~
- [x] ~~TransportMorphisms.v~~
- [x] ~~Biproducts.v~~
- [x] ~~SemiAdditive.v~~
- [x] ~~AdditiveCategories.v~~
- [ ] OppositeCategories.v
- [ ] PreStableCategories.v
- [ ] SemiStableCategories.v
- [ ] Triangles.v
- [ ] OppositePreStable.v
- [ ] PreStableCofiber.v
- [ ] TriangleMorphisms.v
- [ ] TriangleRotation.v
- [ ] ZeroTransformations.v
- [ ] OctahedralLemmas.v
- [ ] ProperStableCategories.v
- [ ] SuspensionFixedPoints.v
- [ ] AdvancedStructures.v
- [ ] OctahedralAxiom.v
- [ ] TriangulatedAxiomsTR123.v
- [ ] StableTriangulated.v
- [ ] DualityPrinciple.v
- [ ] DualityApplications.v

PR #2293, opened 2025-06-26
Add zero objects infrastructure for stable categories

CharlesCNorton:
Per @jdchristensen's suggestion in #2288, this PR extracts just ZeroObjects.v for focused review and merging.

This provides the foundational infrastructure for zero objects and zero morphisms needed for stable category theory.

Changes from original PR based on review feedback:
- Uses `IsInitialObject`/`IsTerminalObject` with `::` for instances  
- Uses new `map_from_initial`/`map_to_terminal` helpers from recently merged PR #2291
- Follows HoTT style conventions (`rapply`, single-line tactics with `1:`, etc.)
- Removed lemmas that now belong in Objects.v
- Fixed all import and style issues

Some lemmas are retained for backwards compatibility.


theories/Categories/Stable/ZeroObjects.v
```diff
@@ -0,0 +1,102 @@
...
+  (X Y : object C) 
+  (f : morphism C X (zero Z))
+  : (map_from_initial Y o f)%morphism = zero_morphism Z X Y
+  := morphism_through_zero_is_zero _ _ _.
+
+(** The zero morphism from zero is the initial morphism. *)
+Definition zero_morphism_from_zero {C : PreCategory} (Z : ZeroObject C)
+  (Y : object C)
+  : zero_morphism Z (zero Z) Y = map_from_initial Y
+  := initial_morphism_unique _ _ _.
```
jdchristensen: I'm still skeptical that it's worth having the above four results.  You should be able to just use the indicated more general functions instead of these special cases.  (The two lemmas below require two steps, so they make sense to have.)  Can you point to specific places (back in the original PR) where the general lemmas won't work?

CharlesCNorton:
> We also need to discuss the location of the file before considering merging this. There's a good argument for having it beside Category/Objects.v, since it is very closely related. In fact, if the four results are removed it becomes short enough that it could just be added to that file. A third option is to put it in a folder Additive, for things about Additive categories. (I think it's likely we're going to want a folder like that for some of the other files.)

My apologies it appears 

> It looks like you didn't test this before creating the PR? It wouldn't build with any version of Coq the CI uses. I fixed that in the first new commit.
> 
> In the second commit, I made many of the changes that I had suggested in #2288. I thought you would have addressed those before creating this PR?

I'm really sorry about this! There was a mixup on my end - I've only been working with GitHub for about two years and clearly made an error with which version got pushed.

I sincerely apologize for the extra work this created for you, especially after you'd already taken the time to provide detailed feedback. Thank you for fixing the build issues and implementing those simplifications.

theories/Categories/Stable/ZeroObjects.v
CharlesCNorton: > I'm still skeptical that it's worth having the above four results. You should be able to just use the indicated more general functions instead of these special cases. (The two lemmas below require two steps, so they make sense to have.) Can you point to specific places (back in the original PR) where the general lemmas won't work?

Two are used in Triangles.v (terminal_comp_is_zero and zero_morphism_from_zero). The general lemmas would work fine there - I'll remove all four if you prefer.

CharlesCNorton:
> We also need to discuss the location of the file before considering merging this. There's a good argument for having it beside Category/Objects.v, since it is very closely related. In fact, if the four results are removed it becomes short enough that it could just be added to that file. A third option is to put it in a folder Additive, for things about Additive categories. (I think it's likely we're going to want a folder like that for some of the other files.)

Sounds good👍Happy to go with whichever location you think works best for the structure. We have Biproducts.v, AdditiveCategories.v coming up, and so it's clear that Additive makes the most sense.

theories/Categories/Stable/ZeroObjects.v
jdchristensen: If the general lemmas work, then please remove these four.  (Another argument for this is that while two of the four are duals of each other, the other two aren't duals (or self-dual), to to complete the pattern we'd need to add two more.)

jdchristensen:
Let's start the Additive folder and put it there.  Thanks for explaining about the git mix-up;  it's easy for things like that to happen.

theories/Categories/Stable/ZeroObjects.v
```diff
@@ -0,0 +1,102 @@
...
+    Such objects play a fundamental role in additive and abelian categories,
+    where they enable the definition of zero morphisms and kernels.
+*)
+
+Record ZeroObject (C : PreCategory) := {
+  zero : object C;
+  is_initial :: IsInitialObject C zero;
+  is_terminal :: IsTerminalObject C zero
+}.
+
```
jdchristensen: I think we should add the following coercion:
```suggestion
Record ZeroObject (C : PreCategory) := {
  zero : object C;
  is_initial :: IsInitialObject C zero;
  is_terminal :: IsTerminalObject C zero
}.

Coercion zero : ZeroObject >-> object.

```
This would then mean that `(zero Z)` and `zero Z` can be replaced with `Z` everywhere, which I think will look better.

theories/Categories/Stable/ZeroObjects.v
```diff
@@ -0,0 +1,102 @@
...
+From HoTT.Categories Require Import InitialTerminalCategory.
+
+(** * Zero objects
+    
+    A zero object in a category is an object that is both initial and terminal.
+    Such objects play a fundamental role in additive and abelian categories,
+    where they enable the definition of zero morphisms and kernels.
+*)
+
+Record ZeroObject (C : PreCategory) := {
```
jdchristensen: If we make this a `Class` instead of a `Record`, then Rocq will automatically use a zero object found in the context.  That would mean that we could make all `Z : ZeroObject C` arguments implicit and remove them from the caller.

We could also do the same for `InitialObject` and `TerminalObject`, make `map_from_initial` and `map_to_terminal` take these types instead of an object and a proof, and add coercions from `ZeroObject` to these two classes.  Then we could write `map_from_initial Y` with `Z` in the context and it would just work.  I can push this if you think it's ok.

jdchristensen: Oh, I just realized that the Category library uses "morphism" not "map", so we should rename those two functions to `morphism_from_initial` and `morphism_to_terminal`.  Sorry for suggesting the wrong names initially!

CharlesCNorton: > I can push this if you think it's ok.

Good idea! Should be fairly contained since most of the remaining files use add_zero anyway. Please feel free to push as offered.

theories/Categories/Stable/ZeroObjects.v
CharlesCNorton: 👍

theories/Categories/Stable/ZeroObjects.v
jdchristensen: Ok, I did a bunch of things I mentioned.

jdchristensen:
Let me know if there's anything that still needs to be handled.  I'll probably merge tomorrow.

CharlesCNorton:
> Let me know if there's anything that still needs to be handled. I'll probably merge tomorrow.

LGTM! Thank you!

PR #2294, opened 2025-06-26
Adapt to rocq-prover/rocq#20813 (Update default hypothesese names in generated schemes)

dhalilov:


jdchristensen:
This change is not backwards compatible.  I guess you could add `: rename` to make it backwards compatible, along with something like:
```
(* TODO: ": rename" is needed because the default names changed in Rocq x.y.z.  When the minimum supported version is >= x.y.z, the ": rename" can be removed. *)
```

jdchristensen:
@dhalilov Thanks, this looks fine.  I'll wait to see what happens with the Rocq PR before merging.

ppedrot:
Please merge now.

PR #2295, opened 2025-07-07
The wedge of n-connected types is n-connected

jdchristensen:


jdchristensen:
Does this generalize to some sorts of pushouts?  Is it already in the library in some form?

@CharlesCNorton , pinging you.

Alizter (approved):

CharlesCNorton:
> Does this generalize to some sorts of pushouts? Is it already in the library in some form?
> 
> @CharlesCNorton , pinging you.

Sorry for the late reply, yes I believe this generalizes to certain pushouts through Blakers-Massey. Thank you very much for pinging.

CharlesCNorton:
> > Sorry for the late reply, yes I believe this generalizes to certain pushouts through Blakers-Massey.
> 
> Do you know which pushouts?
> 
> I'm going to merge this as is, but we can keep a more general result in mind for future work. (My force push was just a rebase on master.)

So, I found wedge_incl_elim in Connectedness.v but no general pushout connectivity preservation.

jdchristensen:
> So, I found wedge_incl_elim in Connectedness.v but no general pushout connectivity preservation.

I think one could use `wedge_incl_elim` to prove the connectedness result, but I'm pretty sure we'd have to do several steps with that approach, so I decided the direct approach was shorter.

PHart3:
I think the following answers the question posed here.

All colimits (in the universe) over contractible graphs (i.e., trees) preserve n-connectedness. (The wedge can be computed as the colimit over a tree.) Indeed, all colimits preserve the left class of an OFS (where the OFS on the diagram wild category is taken to be levelwise), and the colimit of the constant diagram at the unit type is contractible exactly when the graph is a tree.

jdchristensen:
@PHart3 Thanks, Perry!  In particular, we should be able to prove that all pushouts of n-connected types are n-connected.  Nice!

Issue #2290, opened 2025-06-16
Build error with CI / opam-build (dev, ubuntu-latest) (pull_request)


jdchristensen:
The build error is:
```
 Error: - Error: Required library Basics matches several files in path (found
  - /home/rocq/.opam/4.14.2+flambda/.opam-switch/build/coq-hott.dev/_build/default/theories/Basics.vo and
  - /home/rocq/.opam/4.14.2+flambda/lib/coq/theories/Program/Basics.vo).
```

This must be related to the changes to Coq about not accepting ambiguous path names.  We could probably work around it by changing `Basics` to `HoTT.Basics` everywhere, but that's annoying.  Can we instead remove `Program` or maybe all of `coq/theories` from the search path somehow?

@Alizter

Alizter:
@jdchristensen I'll try to take a look this weekend.

jdchristensen:
I tried adding
```
-arg -exclude-dir -arg Program
```
to _CoqProject, but this caused a problem because SetoidRewrite.v indirectly uses this part of the standard library.

Is there a way to only exclude Program from the search path in the core library?  Or is there another quick solution?

If not, I guess we'll need to put `From HoTT` everywhere?

Any thoughts, @Alizter ?

SkySkimmer:
(the makefile build doesn't use -boot and so doesn't get the issue)

PR #2296, opened 2025-07-10
generate_coqproject: avoid Program.Basics and HoTT.Basics conflict

jdchristensen:
This is a temporary solution to #2290.

Feel free to suggest other solutions.  I will merge this if I don't hear about any other options soon, as it is annoying that our CI is always giving errors.


jdchristensen:
This solution works for the core library, but it doesn't work for contrib, because the file SetoidRewrite.v requires Coq.Setoid.Setoid, which indirectly requires Coq.Program.

PR #2298, opened 2025-07-13
Add: transport lemmas for morphisms in categories

CharlesCNorton:
Extract transport lemmas from stable categories formalization (PR #2288).

Changes from original formalization:
- Rename file from ZeroMorphismLemmas.v to TransportMorphisms.v to better reflect content
- Remove ~150 lines of unused/redundant lemmas, keeping only the 4 transport lemmas actually used
- Remove zero-specific sections that weren't actually zero-specific  
- Place in Categories/Additive/ folder alongside ZeroObjects.v

The retained lemmas are:
- transport_compose_morphism
- transport_compose_both_inverse
- transport_inverse_eq
- morphism_eq_transport_inverse

These lemmas handle how morphisms behave under transport along equality proofs and are  essential for proving preservation properties of functors in AdditiveCategories.v and beyond. I verified these are the only lemmas from the original file actually used in the formalization.

The file has been tested to compile successfully using:
```
make -f Makefile.coq theories/Categories/Additive/TransportMorphisms.vo
```

Follows lessons learned and style conventions established in PR #2293 review.

@jdchristensen Please let me know if you see any issues or have suggestions for improvement. 

Sincere apologies if I've overlooked anything from your previous feedback.

### DIFF
--- a/theories/Categories/Additive/TransportMorphisms.v
+++ b/theories/Categories/Additive/TransportMorphisms.v
@@ -1,0 +1,56 @@
+(** * Transport lemmas for morphisms in categories
+
+    Lemmas for transporting morphisms along equality proofs, essential for
+    proving preservation properties of functors.
+
+    These lemmas are primarily used in AdditiveCategories.v to prove that
+    additive functors preserve zero morphisms.
+*)
+
+From HoTT Require Import Basics Types.
+From HoTT.Categories Require Import Category.
+
+Section TransportMorphisms.
+  Context {C : PreCategory}.
+
+  (** Transport distributes over composition. *)
+  Lemma transport_compose_morphism {X Y Z W : object C} (p : X = W)
+    (f : morphism C X Y) (g : morphism C Y Z)
+    : transport (fun U => morphism C U Z) p (g o f)%morphism =
+      (g o transport (fun U => morphism C U Y) p f)%morphism.
+  Proof.
+    destruct p; reflexivity.
+  Qed.
+
+  (** Composing transported morphisms along inverse paths. *)
+  Lemma transport_compose_both_inverse {W X Y Z : object C} (p : W = X)
+    (f : morphism C W Z) (g : morphism C Y W)
+    : (transport (fun U : object C => morphism C U Z) p f o
+       transport (fun U : object C => morphism C Y U) p g)%morphism =
+      (f o g)%morphism.
+  Proof.
+    destruct p; reflexivity.
+  Qed.
+
+  (** Convert equations with transport to equations with inverse transport. *)
+  Lemma transport_inverse_eq {A : Type} {P : A -> Type}
+    {x y : A} (p : x = y) (u : P x) (v : P y)
+    : transport P p u = v -> u = transport P p^ v.
+  Proof.
+    intro H.
+    rewrite <- H.
+    destruct p; reflexivity.
+  Qed.
+
+  (** Specialized version for morphisms. *)
+  Lemma morphism_eq_transport_inverse {W X Y : object C} (p : W = X)
+    (f : morphism C W Y) (g : morphism C X Y)
+    : transport (fun Z => morphism C Z Y) p f = g ->
+      f = transport (fun Z => morphism C Z Y) p^ g.
+  Proof.
+    intro H.
+    rewrite <- H.
+    destruct p; reflexivity.
+  Qed.
+
+End TransportMorphisms.


theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
+(** * Transport lemmas for morphisms in categories
+
+    Lemmas for transporting morphisms along equality proofs, essential for
+    proving preservation properties of functors.
+
+    These lemmas are primarily used in AdditiveCategories.v to prove that
+    additive functors preserve zero morphisms.
```
jdchristensen, line 7: I think you should remove these lines.  Predictions of future use are likely to become stale.

theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
...
+*)
+
+From HoTT Require Import Basics Types.
+From HoTT.Categories Require Import Category.
```
jdchristensen, line 11: You can greatly reduce the dependencies.  For the second line, you just need Category.Core, I think.  And for the first line, note that Category.Core only needs Basics.Overture and Basics.Notations, so you might get by with something similar (plus Basics.PathGroupoids).

(Reducing dependencies makes it possible to build more files in parallel.)

theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
...
+  Context {C : PreCategory}.
+
+  (** Transport distributes over composition. *)
+  Lemma transport_compose_morphism {X Y Z W : object C} (p : X = W)
+    (f : morphism C X Y) (g : morphism C Y Z)
+    : transport (fun U => morphism C U Z) p (g o f)%morphism =
+      (g o transport (fun U => morphism C U Y) p f)%morphism.
+  Proof.
+    destruct p; reflexivity.
+  Qed.
```
jdchristensen, line 23: This is a special case of `ap_transport`, but I think it is ok to leave it, since when using `ap_transport`, Coq can't figure out the function to use.

(For the record, this works:
```
    exact (ap_transport p (fun w k => (g o k)%morphism) f)^.
```
and `p` and `f` can be inferred.)

theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
...
+
+  (** Convert equations with transport to equations with inverse transport. *)
+  Lemma transport_inverse_eq {A : Type} {P : A -> Type}
+    {x y : A} (p : x = y) (u : P x) (v : P y)
+    : transport P p u = v -> u = transport P p^ v.
+  Proof.
+    intro H.
+    rewrite <- H.
+    destruct p; reflexivity.
+  Qed.
```
jdchristensen, line 43: This is `moveL_transport_V`, so should be removed.

theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
...
+  (** Specialized version for morphisms. *)
+  Lemma morphism_eq_transport_inverse {W X Y : object C} (p : W = X)
+    (f : morphism C W Y) (g : morphism C X Y)
+    : transport (fun Z => morphism C Z Y) p f = g ->
+      f = transport (fun Z => morphism C Z Y) p^ g.
+  Proof.
+    intro H.
+    rewrite <- H.
+    destruct p; reflexivity.
+  Qed.
```
jdchristensen, line 54: Instead of reproving it, we can reuse the general case. And since it's a one-liner, maybe this result isn't worth recording?
```suggestion
  Definition morphism_eq_transport_inverse {W X Y : object C} (p : W = X)
    (f : morphism C W Y) (g : morphism C X Y)
    : transport (fun Z => morphism C Z Y) p f = g ->
      f = transport (fun Z => morphism C Z Y) p^ g
    := moveL_transport_V _ p _ _.
```

theories/Categories/Additive/TransportMorphisms.v
```diff
@@ -0,0 +1,56 @@
+(** * Transport lemmas for morphisms in categories
```
jdchristensen, line 1: These results don't belong in the Additive folder, since they are general results about transporting morphisms.  And since only two results are left, each having a one line proof, I think you should just add those two to the end of Category/Core.v.

theories/Categories/Additive/TransportMorphisms.v
jdchristensen: BTW, one of the reasons we tend to use `Definition` instead of `Lemma`, `Theorem`, etc, is that it is then easy to switch to `:=` notation, which only works for definitions.

PR #2300, opened 2025-07-26
[WIP] Add biproduct infrastructure for additive categories

CharlesCNorton:
Extracting Biproducts.v from stable categories formalization (#2288).

**Status: WIP - not ready for merge**

Changes from original formalization:
- Fixed imports (removed non-existent ZeroMorphismLemmas.v)
- Made `ZeroObject` implicit throughout using `` `{Z : ZeroObject C} ``
- Updated `zero_morphism` calls to use implicit Z parameter
- Removed unused definitions (bi_inl, bi_inr, bi_outl, bi_outr, biproduct)
- Removed unused lemma (biproduct_prod_unique)
- Applied HoTT style conventions

The file provides:
- `BiproductData`: injection/projection morphisms
- `IsBiproduct`: axioms (beta_l, beta_r, mixed_l, mixed_r)
- `HasBiproductUniversal`: universal properties as product and coproduct
- `Biproduct`: complete structure combining the above
- Operations: biproduct_coprod_mor, biproduct_prod_mor with beta lemmas
- Uniqueness lemma: biproduct_coprod_unique

@jdchristensen This is work in progress, naturally - submitting early for initial feedback. I will be continuously re-reviewing for anything I missed. Thank you very much for your time

### DIFF
--- a/theories/Categories/Additive/Biproducts.v
+++ b/theories/Categories/Additive/Biproducts.v
@@ -133,6 +133,14 @@ Section BiproductOperations.
     : (biproduct_coprod_mor W f g o inr B = g)%morphism
     := snd_type (biproduct_coprod_universal W f g).2.
 
+  Definition biproduct_coprod_unique (W : object C)
+    (f : morphism C X W) (g : morphism C Y W)
+    (h : morphism C B W)
+    (Hl : (h o inl B = f)%morphism)
+    (Hr : (h o inr B = g)%morphism)
+    : h = biproduct_coprod_mor W f g
+    := ap pr1 (contr (h; (Hl, Hr)))^.
+
   (** The product universal morphism and its properties. *)
   Definition biproduct_prod_universal (W : object C) 
     (f : morphism C W X) (g : morphism C W Y)
@@ -155,16 +163,6 @@ Section BiproductOperations.
     : (outr B o biproduct_prod_mor W f g = g)%morphism
     := snd_type (biproduct_prod_universal W f g).2.
   
-  (** ** Uniqueness properties *)
-
-  Definition biproduct_coprod_unique (W : object C)
-    (f : morphism C X W) (g : morphism C Y W)
-    (h : morphism C B W)
-    (Hl : (h o inl B = f)%morphism)
-    (Hr : (h o inr B = g)%morphism)
-    : h = biproduct_coprod_mor W f g
-    := ap pr1 (contr (h; (Hl, Hr)))^.
-
   Definition biproduct_prod_unique (W : object C)
     (f : morphism C W X) (g : morphism C W Y)
     (h : morphism C W B)
--- a/theories/Categories/Additive/Biproducts.v
+++ b/theories/Categories/Additive/Biproducts.v
@@ -8,6 +8,9 @@ From HoTT Require Import Basics Types.
 From HoTT.Categories Require Import Category Functor.
 From HoTT.Categories.Additive Require Import ZeroObjects.
 
+Local Notation fst_type := Basics.Overture.fst.
+Local Notation snd_type := Basics.Overture.snd.
+
 (** * Biproduct structures *)
 
 (** ** Biproduct data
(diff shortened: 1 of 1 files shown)


theories/Categories/Additive/Biproducts.v
```diff
@@ -0,0 +1,199 @@
...
+(** * Biproduct structures *)
+
+(** ** Biproduct data
+    
+    The data of a biproduct consists of an object together with injection
+    and projection morphisms.
+*)
+
+Record BiproductData {C : PreCategory} (X Y : object C)
+  : Type
```
jdchristensen: I would remove the `: Type` annotation.  It doesn't add anything (Records are always of type Type), but in some cases it causes an extra universe variable (although not in this case).  Same with other Records and Classes below.  (We often also omit it for Definitions, for the same reason.)

theories/Categories/Additive/Biproducts.v
```diff
@@ -0,0 +1,199 @@
...
+  Context {C : PreCategory} `{Z : ZeroObject C} {X Y : object C}.
+  Variable (B : Biproduct X Y).
+
+  (** ** Uniqueness from universal property *)
+
+  (** Extract the unique morphism from the coproduct universal property. *)
+  Definition biproduct_coprod_mor (W : object C) 
+    (f : morphism C X W) (g : morphism C Y W)
+    : morphism C (biproduct_data B) W
+    := pr1 (@center _ (coprod_universal (biproduct_universal B) W f g)).
```
jdchristensen: Since `(@center _ (coprod_universal (biproduct_universal B) W f g)` appears four times (and similarly for the dual results), it's probably best to make a Definition that names this term.  Then all three results could be one-liners (using `.1`, `.2.1` and `.2.2`, for example).

theories/Categories/Additive/Biproducts.v
```diff
@@ -0,0 +1,199 @@
...
+  Proof.
+    unfold biproduct_prod_mor.
+    set (c := @center _ (prod_universal (biproduct_universal B) W f g)).
+    destruct c as [h [Hl Hr]].
+    exact Hr.
+  Qed.
+  
+  (** ** Uniqueness properties *)
+
+  Lemma biproduct_coprod_unique (W : object C)
```
jdchristensen: Can you add the dual result as well?

theories/Categories/Additive/Biproducts.v
```diff
@@ -0,0 +1,199 @@
...
+  (** ** Uniqueness properties *)
+
+  Lemma biproduct_coprod_unique (W : object C) 
+    (f : morphism C X W) (g : morphism C Y W)
+    (h : morphism C (biproduct_data B) W)
+    : (h o inl (biproduct_data B) = f)%morphism -> 
+      (h o inr (biproduct_data B) = g)%morphism ->
+      h = biproduct_coprod_mor W f g.
+  Proof.
+    intros Hl Hr.
```
jdchristensen: May as well name the hypotheses and use the coercion:
```suggestion
    (h : morphism C B W)
    (Hl : (h o inl B = f)%morphism)
    (Hr : (h o inr B = g)%morphism)
    : h = biproduct_coprod_mor W f g.
  Proof.
```

jdchristensen:
@Alizter @JasonGross Do either of you know why the doc-dep-graphs build is failing?  It happened on two PRs today.  The error is
```
+ (script @ line 12) $ cabal update
Config file path source is default config file.
Config file /home/coq/.cabal/config not found.
Writing default configuration to /home/coq/.cabal/config
<repo>/root.json does not have enough signatures signed with the appropriate keys
```

JasonGross:
The internet suggests this is an issue with cabal-install metadata being outdated, and that adding `ghcup install cabal` before `cabal update` might fix the issue

theories/Categories/Additive/Biproducts.v
CharlesCNorton: @jdchristensen Sorry if this is obvious, but I can't get the `.2.1` and `.2.2` projection syntax to work with the factored definition. The sigma type contains a product type, so `(biproduct_coprod_universal W f g).2` gives us a `*` type, not another sigma. What's the correct projection syntax for making these one-liners?

jdchristensen: Sorry, I was confused.  I don't think there is a notation for projections from a `prod` type, so you'd have to use `fst (...).2`, for example.  Still avoids a `destruct` and so fits on one line.

CharlesCNorton: @jdchristensen Ok I tried to make them one-liners with fst/snd but kept getting type errors. I did factor out the repeated center expressions into `biproduct_coprod_universal` and `biproduct_prod_universal` definitions as you suggested, which avoids the duplication. The beta lemmas still need set/destruct but at least the code is cleaner now - I think.

jdchristensen: It turns out that the Categories library redefines fst, snd and some other things.  I added a workaround used elsewhere in the library for accessing the original meanings.  I also found that other proofs were overly verbose and compressed them.  (BTW, I don't think one-liners are something we should strive for.  But I feel that in these cases they express the proofs more clearly than the long-winded approaches.)

Let me know if you see anything else that needs changing before merging.

CharlesCNorton: @jdchristensen Thank you for tracking down the fst/snd issue and implementing the fix. The compressed proofs look great. 👍 Ready to merge on my end - much appreciated!

PR #2303, opened 2025-07-29
Add "ghcup install cabal" to work around error

jdchristensen:
This is an attempt to fix the doc-dep-graphs build, which is failing with:
```
+ (script @ line 12) $ cabal update
Config file path source is default config file.
Config file /home/coq/.cabal/config not found.
Writing default configuration to /home/coq/.cabal/config
<repo>/root.json does not have enough signatures signed with the appropriate keys
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -256,8 +256,9 @@ jobs:
           libxml2-dev libxslt1-dev time lua5.1 unzip npm
           # libnode-dev node-gyp
 
-          # The next line will hopefully fix the error:
+          # The next two lines will hopefully fix the error:
           # <repo>/root.json does not have enough signatures signed with the appropriate keys
+          curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
           ghcup install cabal
           cabal update
           cabal install --lib graphviz text fgl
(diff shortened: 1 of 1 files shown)


jdchristensen:
The line
```
curl --proto '=https' --tlsv1.2 -sSf https://get-ghcup.haskell.org | sh
```
seemed to work, but is intended to be used in an interactive session, so it is extremely verbose, generates errors about there being no tty, and installs a bunch of other things that we don't want (since we aren't there to say no).  And then after succeeding, `ghcup` is still not found, probably because the `PATH` variable needs to be amended.  Yuck.

Would something like [ghcup-setup](https://github.com/haskell/ghcup-setup) be better?

@JasonGross @Alizter Can one of you take this over?  I really don't know anything about github CI or haskell installation...

.github/workflows/ci.yml
```diff
@@ -256,6 +256,11 @@ jobs:
           libxml2-dev libxslt1-dev time lua5.1 unzip npm
           # libnode-dev node-gyp
 
+          # The next three lines will hopefully fix the error:
```
jdchristensen:
```suggestion
          # The next three lines fix the error:
```

JasonGross:
> Should we now remove `cabal-install` from the list of Ubuntu packages a few lines earlier?

Seems fine either way

PR #2304, opened 2025-08-02
Add additive categories infrastructure

CharlesCNorton:
Defines additive categories on top of the semi-additive structure from #2305, following the design suggested in review.

Provides:
- `AdditiveCategory`: a `SemiAdditiveCategory` together with an `Inverse` operation on each hom-set and the left inverse law
- An `IsAbGroup` instance and a bundled `abgroup_hom : AbGroup` for each hom-set
- Uniqueness of additive inverses, and compatibility of negation with composition on both sides
- `AdditiveFunctor`: a functor whose action on each hom-set is monoid-preserving, with preservation of zero morphisms and negation derived
- Identity and composition of additive functors

### DIFF
--- a/theories/Categories/Additive/Additive.v
+++ b/theories/Categories/Additive/Additive.v
@@ -5,10 +5,9 @@
 
 From HoTT Require Import Basics.Overture Basics.Tactics.
 From HoTT.Classes.interfaces Require Import canonical_names abstract_algebra.
-From HoTT.Classes.theory Require Import groups.
 From HoTT.Categories Require Import Category.Core Functor.Core.
 From HoTT.Categories.Functor Require Import Identity Composition.Core.
-From HoTT.Categories.Additive Require Import ZeroObjects Biproducts SemiAdditive.
+From HoTT.Categories.Additive Require Import ZeroObjects SemiAdditive.
 From HoTT.Algebra.AbGroups Require Import AbelianGroup.
 
 Local Open Scope morphism_scope.
@@ -52,8 +51,8 @@ End HomAbGroup.
 
 (** ** Negation and composition *)
 
-(** Additive inverses are unique. *)
-Definition inverse_morphism_unique {A : AdditiveCategory} {X Y : object A}
+(** A left additive inverse equals the negation. *)
+Definition hom_moveL_1V {A : AdditiveCategory} {X Y : object A}
   {f g : morphism A X Y} (H : g + f = 0)
   : g = - f
   := grp_moveL_1V (G:=abgroup_hom X Y) H.
@@ -63,7 +62,7 @@ Definition inverse_precompose {A : AdditiveCategory} {X Y W : object A}
   (f : morphism A X Y) (a : morphism A W X)
   : (- f) o a = - (f o a).
 Proof.
-  apply inverse_morphism_unique.
+  apply hom_moveL_1V.
   lhs_V napply addition_precompose.
   lhs napply (ap (fun g => g o a) (additive_inverse_l f)).
   napply zero_morphism_left.
@@ -74,7 +73,7 @@ Definition inverse_postcompose {A : AdditiveCategory} {X Y W : object A}
   (a : morphism A Y W) (f : morphism A X Y)
   : a o (- f) = - (a o f).
 Proof.
-  apply inverse_morphism_unique.
+  apply hom_moveL_1V.
   lhs_V napply addition_postcompose.
   lhs napply (ap (fun g => a o g) (additive_inverse_l f)).
   napply zero_morphism_right.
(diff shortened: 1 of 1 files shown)


jdchristensen:
Thanks @CharlesCNorton !  I just pushed a commit that shows that `add_zero_morphism` isn't needed.  I suspect that many of your other wrappers can probably be avoided (maybe with the addition of some appropriate instances).

theories/Categories/Additive/AdditiveCategories.v
```diff
@@ -0,0 +1,356 @@
...
+(** ** Definition
+    
+    An additive category has a zero object and biproducts for all pairs
+    of objects. This is a key step toward abelian categories.
+*)
+
+Class AdditiveCategory := {
+  cat : PreCategory;
+  add_zero :: ZeroObject cat;
+  add_biproduct : forall (X Y : object cat), Biproduct X Y;
```
jdchristensen: I'm confused by this class.  The three fields above are already enough to show that `cat` has a unique enrichment over commutative monoids.  That means that eight of the remaining nine fields are redundant.  You only need to assume that inverses exist.

Here's an idea:  Have a file, maybe called "SemiAdditive.v", which has a class with just the above three fields.  In that file, prove that in every SemiAdditive category, the hom sets are commutative monoids in a way compatible with composition.  It's probably best to use `IsCommutativeMonoid` from Classes/interfaces/abstract_algebra.v, which should then give you access to lots of lemmas (and notations) that hold for commutative monoids.  Then, in Additive.v, you add the axiom that the commutative monoid structure defined in SemiAdditive.v has inverses.  Then you can in fact say that the hom sets are abelian groups, using Algebra/AbGroups/AbelianGroup.v.

CharlesCNorton: > I'm confused by this class. The three fields above are already enough to show that `cat` has a unique enrichment over commutative monoids. That means that eight of the remaining nine fields are redundant. You only need to assume that inverses exist.
> 
> Here's an idea: Have a file, maybe called "SemiAdditive.v", which has a class with just the above three fields. In that file, prove that in every SemiAdditive category, the hom sets are commutative monoids in a way compatible with composition. It's probably best to use `IsCommutativeMonoid` from Classes/interfaces/abstract_algebra.v, which should then give you access to lots of lemmas (and notations) that hold for commutative monoids. Then, in Additive.v, you add the axiom that the commutative monoid structure defined in SemiAdditive.v has inverses. Then you can in fact say that the hom sets are abelian groups, using Algebra/AbGroups/AbelianGroup.v.

Just checking in - working on this, but it's taking some sweat. Will put a PR in soon.

CharlesCNorton:
I think this is good to go. I've been wrong in the past though....

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+    Semi-additive categories in which morphism addition admits inverses,
+    so that hom-sets are abelian groups and composition is bilinear. *)
+
+From HoTT Require Import Basics.Overture Basics.Tactics.
+From HoTT.Classes.interfaces Require Import canonical_names abstract_algebra.
+From HoTT.Classes.theory Require Import groups.
+From HoTT.Categories Require Import Category.Core Functor.Core.
+From HoTT.Categories.Functor Require Import Identity Composition.Core.
+From HoTT.Categories.Additive Require Import ZeroObjects Biproducts SemiAdditive.
+From HoTT.Algebra.AbGroups Require Import AbelianGroup.
```
jdchristensen, line 11: Can you check that all of the Required modules are used?  (Note that sometimes removing a Require line doesn't prevent the file from building, even if it is used.  But then various things get displayed in a long-winded way.  So only remove things that aren't used.)

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+
+  #[export] Instance isabgroup_morphisms : IsAbGroup (morphism A X Y).
+  Proof.
+    split.
+    - split.
+      + exact _.
+      + exact additive_inverse_l.
+      + exact additive_inverse_r.
+    - exact _.
+  Defined.
```
jdchristensen: You can use `Build_AbGroup'`, which avoids the need for `additive_inverse_r`.  I'll push a fix.

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+
+Arguments abgroup_hom {A} X Y.
+
+(** ** Negation and composition *)
+
+(** Additive inverses are unique. *)
+Definition inverse_morphism_unique {A : AdditiveCategory} {X Y : object A}
+  {f g : morphism A X Y} (H : g + f = 0)
+  : g = - f
+  := grp_moveL_1V (G:=abgroup_hom X Y) H.
```
jdchristensen: I don't like the name of this.  It's just a wrapper, so it's really just the movement lemma.   If you are willing to replace `morphism A X Y` with `abgroup_hom X Y` everywhere, then you can just use `grp_moveL_1V` directly, with no wrapper.  But if you want to be able to work with morphisms that aren't given the type `abgroup_hom  X Y`, then wrappers like this are fine, but I think the name should parallel the underlying name, e.g. `hom_moveL_1V`, or something like that.

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+
+Record AdditiveFunctor (A B : AdditiveCategory) := {
+  add_functor :> Functor A B;
+
+  addfunctor_monoid :: forall (X Y : object A),
+    IsMonoidPreserving (@morphism_of _ _ add_functor X Y);
+}.
+
+Arguments add_functor {A B} F : rename.
+Arguments addfunctor_monoid {A B} F X Y : rename.
```
jdchristensen: Be consistent:  either `add_functor` or `addfunctor` in both cases.  Also, maybe the second should be `ismonoidpreserving_addfunctor`?  (Since it's an instance, it won't be typed much.)

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+
+Section AdditiveFunctorLaws.
```
jdchristensen: Are the three laws in this section found by typeclass search?  Or can the user just use `preserves_sg_op`, etc, directly, so these wrappers won't be needed?

theories/Categories/Additive/Additive.v
```diff
@@ -0,0 +1,161 @@
...
+  Context {A B : AdditiveCategory} (F : AdditiveFunctor A B)
+    {X Y : object A}.
+
+  (** Additive functors preserve addition of morphisms. *)
+  Definition additive_functor_preserves_addition (f g : morphism A X Y)
+    : morphism_of F (f + g) = morphism_of F f + morphism_of F g
+    := preserves_sg_op f g.
+
+  (** Additive functors preserve zero morphisms. *)
+  Definition additive_functor_preserves_zero_morphisms
```
jdchristensen: If these are kept, the name should involve `addfunctor` or `add_functor`.  Having three different forms means that user has to remember which to use in which case.

jdchristensen:
LGTM.  Thanks!

PR #2305, opened 2025-08-10
Add semi-additive categories infrastructure

CharlesCNorton:
Extracting SemiAdditive.v from additive categories work per PR #2304 review as part of general formalization for #2288

**Provides:**
- `SemiAdditiveCategory`: Class with just cat, zero object, and biproducts
- Derives `IsCommutativeMonoid` structure on morphisms using biproducts
- Addition defined as: f + g = ∇ ∘ (f ⊕ g) ∘ Δ  
- Proves identity laws, commutativity, associativity from biproduct axioms
- Bilinearity of composition (distributivity)

**Note:** The file is almost 1000 lines and requires refactoring, especially the associativity proof. @jdchristensen this was the only way I could get it to work - would very much appreciate help simplifying and identifying which helper lemmas can be removed. The associativity proof in particular requires many intermediate lemmas about how biproducts interact. It took me an entire week of 16 hour days to get here, but I assume there is a much more elegant method.

### DIFF
--- a/theories/Categories/Additive/Biproducts.v
+++ b/theories/Categories/Additive/Biproducts.v
@@ -4,28 +4,30 @@
     to additive category theory.
 *)
 
-From HoTT Require Import Basics Types.
-From HoTT.Categories Require Import Category Functor.
+From HoTT Require Import Basics.Overture Basics.Tactics.
+From HoTT.Categories Require Import Category.Core.
 From HoTT.Categories.Additive Require Import ZeroObjects.
 
 Local Notation fst_type := Basics.Overture.fst.
 Local Notation snd_type := Basics.Overture.snd.
 
+Local Open Scope morphism_scope.
+
 (** * Biproduct structures *)
 
 (** ** Biproduct data
-    
+
     The data of a biproduct consists of an object together with injection
     and projection morphisms.
 *)
 
 Record BiproductData {C : PreCategory} (X Y : object C) := {
   biproduct_obj : object C;
-  
+
   (* Coproduct structure: injections *)
   inl : morphism C X biproduct_obj;
   inr : morphism C Y biproduct_obj;
-  
+
   (* Product structure: projections *)
   outl : morphism C biproduct_obj X;
   outr : morphism C biproduct_obj Y
@@ -40,20 +42,20 @@ Arguments outl {C X Y} b : rename.
 Arguments outr {C X Y} b : rename.
 
 (** ** Biproduct axioms
-    
+
     The axioms that make biproduct data into an actual biproduct.
     These ensure the projection-injection pairs behave correctly.
 *)
 
-Record IsBiproduct {C : PreCategory} `{Z : ZeroObject C} {X Y : object C} 
+Record IsBiproduct {C : PreCategory} `{Z : ZeroObject C} {X Y : object C}
                    (B : BiproductData X Y) := {
   (* Projection-injection identities *)
-  beta_l : (outl B o inl B = 1)%morphism;
-  beta_r : (outr B o inr B = 1)%morphism;
-  
+  beta_l : outl B o inl B = 1;
+  beta_r : outr B o inr B = 1;
+
   (* Mixed terms are zero *)
-  mixed_l : (outl B o inr B)%morphism = zero_morphism Y X;
-  mixed_r : (outr B o inl B)%morphism = zero_morphism X Y
+  mixed_l : outl B o inr B = zero_morphism Y X;
+  mixed_r : outr B o inl B = zero_morphism X Y
 }.
 
 Arguments beta_l {C Z X Y B} i : rename.
@@ -62,7 +64,7 @@ Arguments mixed_l {C Z X Y B} i : rename.
 Arguments mixed_r {C Z X Y B} i : rename.
 
 (** ** Universal property of biproducts
-    
+
     The universal property states that morphisms into/out of the biproduct
     are uniquely determined by their components.
 *)
@@ -70,28 +72,29 @@ Arguments mixed_r {C Z X Y B} i : rename.
 Record HasBiproductUniversal {C : PreCategory} {X Y : object C}
   (B : BiproductData X Y) := {
   (* Universal property as a coproduct *)
-  coprod_universal : forall (W : object C) 
+  coprod_universal : forall (W : object C)
     (f : morphism C X W) (g : morphism C Y W),
-    Contr {h : morphism C B W & 
-           ((h o inl B = f)%morphism * (h o inr B = g)%morphism)};
-  
+    Contr {h : morphism C B W &
+           ((h o inl B = f) * (h o inr B = g))};
+
   (* Universal property as a product *)
-  prod_universal : forall (W : object C) 
+  prod_universal : forall (W : object C)
     (f : morphism C W X) (g : morphism C W Y),
-    Contr {h : morphism C W B & 
-           ((outl B o h = f)%morphism * (outr B o h = g)%morphism)}
+    Contr {h : morphism C W B &
+           ((outl B o h = f) * (outr B o h = g))}
 }.
 
 Arguments coprod_universal {C X Y B} u W f g : rename.
 Arguments prod_universal {C X Y B} u W f g : rename.
 
 (** ** Complete biproduct structure
-    
+
     A biproduct is biproduct data together with the proof that it satisfies
-    the biproduct axioms and universal property.
+    the biproduct axioms and universal property.  The zero object is an
+    explicit argument so that instances can be written [Biproduct Z X Y].
 *)
 
-Class Biproduct {C : PreCategory} `{Z : ZeroObject C} (X Y : object C) := {
+Class Biproduct {C : PreCategory} (Z : ZeroObject C) (X Y : object C) := {
   biproduct_data : BiproductData X Y;
   biproduct_is : IsBiproduct biproduct_data;
   biproduct_universal : HasBiproductUniversal biproduct_data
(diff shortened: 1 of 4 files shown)


CharlesCNorton:
Also apologies if there are any nits I missed - I'll give it another lookover soon.

CharlesCNorton:
Refactored. File is about 20% shorter than before at 800 lines instead of 1000. Going through it again...

CharlesCNorton:
We're now down to about half the length of the original PR. Continuing.

jdchristensen:
@CharlesCNorton I haven't looked at this yet and will wait until you ok it.  But I'll point out a couple of things that may help.

First, there's a trick for showing the a binary operation `*` is associative and commutative.  You prove commutativity as expected, but instead of directly proving associativity, you prove the "twist" law:  `(x * y) * z <~> (z * y) * x` (or some other variant).  This is easier than proving associativity, because you can use the same function in both directions, and you only have to prove one composite is the identity (since that covers both composites, just as when you are proving symmetry).  Associativity follows from these two.  See `baer_sum_twist` and `equiv_trijoin_twist'` in the library.

The second thing that sometimes is helpful is to study the triple product `x * y * z` as its own operation, which is done for both the Baer sum and the join.

The Baer sum formalization should be particularly relevant, since it uses a lot about biproducts of abelian groups which should generalize to biproducts in any category.

CharlesCNorton:
> @CharlesCNorton I haven't looked at this yet and will wait until you ok it. But I'll point out a couple of things that may help.
> 
> First, there's a trick for showing the a binary operation `*` is associative and commutative. You prove commutativity as expected, but instead of directly proving associativity, you prove the "twist" law: `(x * y) * z <~> (z * y) * x` (or some other variant). This is easier than proving associativity, because you can use the same function in both directions, and you only have to prove one composite is the identity (since that covers both composites, just as when you are proving symmetry). Associativity follows from these two. See `baer_sum_twist` and `equiv_trijoin_twist'` in the library.
> 
> The second thing that sometimes is helpful is to study the triple product `x * y * z` as its own operation, which is done for both the Baer sum and the join.
> 
> The Baer sum formalization should be particularly relevant, since it uses a lot about biproducts of abelian groups which should generalize to biproducts in any category.

Thank you very much!

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,510 @@
...
+End MorphismAddition.
+
+(** Make the operations instances for typeclass search. *)
+#[export] Instance morphism_sgop (C : SemiAdditiveCategory) 
+  (X Y : object C) : SgOp (morphism C X Y) 
+  := morphism_addition C X Y.
+
+#[export] Instance morphism_monunit (C : SemiAdditiveCategory) 
+  (X Y : object C) : MonUnit (morphism C X Y) 
+  := morphism_zero C X Y.
```
jdchristensen: Instead of making synonyms for these two results, you can make these `Instance`s instead of `Definition`s when they are first defined.  As for the names, we have all sorts of styles in the library, but what would follow our general conventions would be `sgop_morphism` and `monunit_morphism`, I think.  [I haven't looked at the rest of this file, but just happened to see this while switching tabs.  I still plan to wait until you tell me it's ready for review.]

theories/Categories/Additive/Biproducts.v
```diff
@@ -4,8 +4,8 @@
     to additive category theory.
 *)
 
-From HoTT Require Import Basics Types.
-From HoTT.Categories Require Import Category Functor.
+From HoTT Require Import Basics.Overture.
+From HoTT.Categories Require Import Category.Core.
```
jdchristensen: It's important to keep imports lean.  This means that fewer files need to be rebuilt when changes are made, and that the builds parallelize better.  (It doesn't need to be strictly the minimum.  E.g. if you depend on many files in Types, you can just import Types.  Or if you depend on things that depend on Types, there is no harm if depending on Types.  But this basic material about Categories is fairly independent of the rest of the library, so it's worth keeping the dependencies small.)

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+Section MorphismAddition.
+  Context (C : SemiAdditiveCategory) (X Y : object C).
+  
+  (** Codiagonal composed with pairing. *)
+  #[export] Instance sgop_morphism : SgOp (morphism C X Y) :=
+    fun f g =>
+      (biproduct_coprod_mor (semiadditive_biproduct Y Y) Y 
+         1%morphism 1%morphism
+       o biproduct_prod_mor (semiadditive_biproduct Y Y) X 
+         f g)%morphism.
```
jdchristensen: This is verbose.  Can you come up with a better way to express these morphisms, with some arguments implicit and provided via typeclass search?  This comes up a lot in the file.  (A similar thing appears in the ⊕ Notation.)

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+  
+  (** The zero morphism is the unit for addition. *)
+  #[export] Instance monunit_morphism : MonUnit (morphism C X Y)
+    := @zero_morphism C semiadditive_zero X Y.
+    
+End MorphismAddition.
+
+(** ** Biproduct characterization lemmas *)
+
+Section BiproductCharacterization.
```
jdchristensen: It seems to me that the lemmas in the section only require a single Biproduct to exist, rather than a SemiAdditiveCategory structure, so they could be proved in Biproducts.v.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+       (outl (biproduct_data (semiadditive_biproduct A A))).
+
+  (** Swapping components of a biproduct morphism. *)
+  Lemma biproduct_prod_swap (A B : object C) (f g : morphism C A B)
+    : biproduct_prod_mor (semiadditive_biproduct B B) A g f 
+      = (biproduct_swap B o biproduct_prod_mor (semiadditive_biproduct B B) A f g)%morphism.
+  Proof.
+    symmetry; rapply biproduct_prod_unique; rewrite <- Category.Core.associativity; unfold biproduct_swap;
+    [rewrite biproduct_prod_beta_l; rapply biproduct_prod_beta_r | rewrite biproduct_prod_beta_r; rapply biproduct_prod_beta_l].
+  Qed.
```
jdchristensen: Can you rewrite all proofs so that it's possible to step through them and see what is going on?  The semicolons and brackets don't actually save any characters in most cases, so I see no benefit to this style.  (There are cases where such tricks are useful, e.g. if the form of the goal changes cleanly when two tactics are used together, making it more clear what is going on.)

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+
+(** ** Commutativity of morphism addition *)
+
+  Theorem morphism_addition_commutative (C : SemiAdditiveCategory) 
+    (X Y : object C) : Commutative (@sgop_morphism C X Y).
+  Proof.
+    intros f g; unfold sgop_morphism;
+    rewrite (biproduct_prod_swap C X Y f g), <- Category.Core.associativity, codiagonal_swap_invariant;
+    reflexivity.
+  Qed.
```
jdchristensen: Indentation issues here and elsewhere.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+    rewrite <- Category.Core.associativity.
+    rewrite biproduct_coprod_beta_l.
+    rapply Category.Core.left_identity.
+  Qed.
+
+End IdentityLaws.
+
+(** ** Swap morphism for biproducts *)
+
+Section BiproductSwap.
```
jdchristensen: Again, some or all of this section probably belongs in Biproducts.v, since these results could be useful for a single biproduct object, even if you don't know you are in a semiadditive category.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+  Context (C : SemiAdditiveCategory).
+
+  (** The swap morphism for biproducts. *)
+  Definition biproduct_swap (A : object C) 
+  : morphism C (A ⊕ A) (A ⊕ A)
```
jdchristensen: Why not [A ⊕ B -> B ⊕ A]?  Then you are probably close to proving that biproducts are commutative.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+
+(** ** Biproduct characterization lemmas *)
+
+Section BiproductCharacterization.
+  Context (C : SemiAdditiveCategory).
+
+  (** Special cases: biproduct morphisms with zero components. *)
+  Lemma biproduct_zero_right_is_inl (Y Z : object C) (h : morphism C Z Y)
+    : biproduct_prod_mor (semiadditive_biproduct Y Y) Z h (zero_morphism Z Y)
+      = (Biproducts.inl (biproduct_data (semiadditive_biproduct Y Y)) o h)%morphism.
```
jdchristensen: This and the next lemma are ok, but I think the proof could be built out of a few more conceptual tools.  In Biproducts.v, you could prove that `Biproducts.inl = biproduct_prod_mor id zero` (not typed precisely), which is a useful general fact.  Then the third lemma in this section is a naturality result, which tells you that `biproduct_prod_mor id zero o h = biproduct_prod_mor (id o h) (zero o h)`.  Then you just use the properties of id and zero.  Seems cleaner, and that result about inl is independently useful.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,354 @@
...
+      rapply zero_morphism_left.
+    - rewrite (beta_r (biproduct_is (semiadditive_biproduct Y Y))).
+      rapply Category.Core.left_identity.
+  Qed.
+
+  (** Composition of biproduct morphisms. *)
+  Lemma biproduct_comp_general (W X Y Z : object C)
+    (f : morphism C W X) (g : morphism C W Y) (h : morphism C Z W)
+    : (biproduct_prod_mor (semiadditive_biproduct X Y) W f g o h)%morphism
+      = biproduct_prod_mor (semiadditive_biproduct X Y) Z (f o h) (g o h).
```
jdchristensen: This is a naturality result, so that should be mentioned in the comment and in the name.  (Also, this is a result that is just a fact about products, not biproducts.  Really products (and coproducts) should have been studied on their own first, and then biproducts could make use of those results, but I think that's a refactoring for another time.)

theories/Categories/Additive/ZeroObjects.v
```diff
@@ -4,9 +4,8 @@
     concepts for additive and stable category theory.
 *)
 
-From HoTT Require Import Basics Types.
-From HoTT.Categories Require Import Category Functor NaturalTransformation.
-From HoTT.Categories Require Import InitialTerminalCategory.
```
jdchristensen, line 9: Tons of dependencies were unnecessary here.

jdchristensen (commented):
I've only looked over the first 20% or so.

jdchristensen:
> Was having a tough time implementing the twist, and so kept a more direct proof for associativity.

The examples where I've seen it used are for proving equivalences `A * (B * C) <~> (A * B) * C`, but here you are proving equalities, so maybe it's not as useful.  For equalities, you just have to give a path in one direction and you are done.  For equivalences, you have to give maps in both directions and then prove that both composites are the identity, and the twist saves you a factor of 2 of work, as the same map works in both directions and the same proof handles both composites.

That said, the proof of associativity still looks really long, and conceptually I don't see why it should be long.  I wonder if what you are missing is the associativity of the biproduct itself (for which the twist approach would help).  #2016 does this for wild categories using the twist approach (but doesn't get to addition of morphisms).  Not sure if it would be worth trying this.  It would certainly be a lot of work, but the idea is that instead of pushing through proofs, one tries to develop the basic tools that make later proofs easy.  And knowing that biproducts are associative is one of the basic facts  one would expect to have.  Probably not worth it here, so just take this as an aside.

(This also suggests the idea of switching to wild categories.  We have more about biproducts there, and they are also defined in terms of coproducts and products, so results are stated at the correct level of generality.  But #2016 is still a work in progress, with @ThomatoTomato planning to reorganize things a bit.)

theories/Categories/Additive/SemiAdditive.v
jdchristensen: Also, I think one of the `Y`s can be changed to a fresh object `Y'`, giving a slightly more general result.

theories/Categories/Additive/SemiAdditive.v
CharlesCNorton: @jdchristensen Understood - I was prioritizing keeping the line count minimal. I'll keep things steppable from now on.

jdchristensen:
> @jdchristensen Thanks for the review. Do you think I should switch to WildCat now or stick with the current approach? I suspect I know the answer, but wanted your opinion in case there's some utility in continuing.

I don't think there's an easy answer.  If your current proofs just needed mild touch-ups, I'd lean towards just sticking with Categories, since the code has already been written and it might be quite a burden to rewrite it.  But so far it seems that each file needs fairly substantial revisions, which makes me wonder if it wouldn't be that much extra work to switch to WildCat.  The advantage of switching is that multiple people will be interested in using some of these results and may be able to contribute to review or even proofs.  One disadvantage of switching (besides the fact that you already have code for Categories) is that dealing with the plethora of typeclasses used by WildCat can be a bit confusing.

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+
+Section SelfBiproductOperations.
+  Context {C : PreCategory} `{Z : ZeroObject C}.
+
+  (** Pairing into a self-biproduct. *)
+  Definition biproduct_sum_pair {X Y : object C}
+    `{BYY : @Biproduct C Z Y Y}
+    (f g : morphism C X Y)
+    : morphism C X BYY
+    := biproduct_prod_mor BYY X f g.
```
jdchristensen: Is it useful to have this definition?  Except for the different choices of implicit arguments, one can just use `biproduct_prod_mor`.

(And it would seem reasonable to make the `W` argument of `biproduct_prod_mor` implicit, since it can be deduced from the provided maps.  Not sure how well it would work to make the biproduct argument implicit, but you could try.)

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+
+  (** A morphism induced on self-biproducts by maps on the two summands. *)
+  Definition biproduct_sum_map {Y Y' : object C}
+    `{BY : @Biproduct C Z Y Y} `{BY' : @Biproduct C Z Y' Y'}
+    (a b : morphism C Y Y')
+    : morphism C BY BY'
+    := biproduct_prod_mor BY' BY (a o outl BY) (b o outr BY).
```
jdchristensen: This works equally well for biproducts of non-equal objects, and is just the functoriality of biproducts, so it should be done in that generality.

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+  (** The left injection is pairing with zero on the right. *)
+  Lemma biproduct_inl_sum_pair {Y : object C}
+    `{BYY : @Biproduct C Z Y Y}
+    : inl BYY = biproduct_sum_pair 1%morphism (zero_morphism Y Y).
+  Proof.
+    symmetry.
+    unfold biproduct_sum_pair.
+    rewrite biproduct_prod_zero_r.
+    apply right_identity.
+  Qed.
```
jdchristensen: This and the next result don't need the objects the same.  (`biproduct_prod_zero_l` is the dual result, generalized even further.)

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+    - rewrite <- associativity.
+      rewrite (mixed_l (biproduct_is B)).
+      apply zero_morphism_left.
+    - rewrite <- associativity.
+      rewrite (beta_r (biproduct_is B)).
+      apply left_identity.
+  Qed.
+
+  (** Product pairing is natural in the domain. *)
+  Lemma biproduct_prod_comp (V W : object C)
```
jdchristensen: The name of this and the next result should mention `nat` for naturality.  Not sure how to indicate that it involves the domain.  Maybe there are other examples to compare to?

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+    repeat rewrite right_identity.
+    reflexivity.
+  Qed.
+
+  (** Pairing is natural in the codomain. *)
+  Lemma biproduct_sum_pair_natural {X Y Y' : object C}
+    `{BYY : @Biproduct C Z Y Y} `{BYY' : @Biproduct C Z Y' Y'}
+    (a : morphism C Y Y') (f g : morphism C X Y)
+    : biproduct_sum_pair (a o f) (a o g)
+      = (biproduct_sum_map a a o biproduct_sum_pair f g)%morphism.
```
jdchristensen: This also works when the objects are not the same.  Same with most or all later results.

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+    - unfold biproduct_sum_pair.
+      apply biproduct_prod_zero_l.
+  Qed.
+
+  (** The codiagonal of a self-biproduct map is the corresponding copairing. *)
+  Lemma biproduct_codiagonal_factor_through_sum_map {Y Y' : object C}
+    `{BYY : @Biproduct C Z Y Y} `{BYY' : @Biproduct C Z Y' Y'}
+    (a b : morphism C Y Y')
+    : (biproduct_codiagonal Y' o biproduct_sum_map a b)%morphism
+      = biproduct_coprod_mor BYY Y' a b.
```
jdchristensen: This is a special case of naturality of `biproduct_coprod_mor` in the domain, so maybe it should be proved in that generality first?

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+    apply left_identity.
+  Qed.
+
+  (** ** Symmetry of self-biproducts *)
+
+  (** The swap morphism exchanges the two summands of a self-biproduct. *)
+  Definition biproduct_swap (Y : object C)
+    `{BYY : @Biproduct C Z Y Y}
+    : morphism C BYY BYY
+    := biproduct_sum_pair (outr BYY) (outl BYY).
```
jdchristensen: As I mentioned a few months ago, there's no reason to only define the swap when the two objects are the same.

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,320 @@ Section BiproductOperations.
...
+
+  (** Swapping components of a self-biproduct pairing. *)
+  Lemma biproduct_prod_swap {X Y : object C}
+    `{BYY : @Biproduct C Z Y Y}
+    (f g : morphism C X Y)
+    : biproduct_sum_pair g f
+      = (biproduct_swap Y o biproduct_sum_pair f g)%morphism.
```
jdchristensen: This also generalizes to different objects.

jdchristensen (commented):
So far I only reviewed the changes to Biproducts.v.  After fixing things, can you re-read your other changes to make sure you are not missing generalizations?

CharlesCNorton:
Re-read SemiAdditive.v; nothing else generalizes. Also removed biproduct_sum_pair, proved copairing natural in the domain and derived the codiagonal factorization from it, and proved uniqueness of the enrichment.

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,305 @@ Section BiproductOperations.
...
+
+Section SelfBiproductOperations.
+  Context {C : PreCategory} `{Z : ZeroObject C}.
+
+  (** A morphism induced on self-biproducts by maps on the two summands. *)
+  Definition biproduct_sum_map {X Y X' Y' : object C}
+    `{BXY : @Biproduct C Z X Y} `{BXY' : @Biproduct C Z X' Y'}
+    (a : morphism C X X') (b : morphism C Y Y')
+    : morphism C BXY BXY'
+    := biproduct_prod_mor BXY' (a o outl BXY) (b o outr BXY).
```
jdchristensen: This should move elsewhere, and the comment should be updated, as it's not about self-biproducts.  (Please check the other results too.)

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,305 @@ Section BiproductOperations.
...
+
+  (** The codiagonal of a self-biproduct. *)
+  Definition biproduct_codiagonal (Y : object C)
+    `{BYY : @Biproduct C Z Y Y}
+    : morphism C BYY Y
+    := biproduct_coprod_mor BYY Y 1%morphism 1%morphism.
+
+  (** The codiagonal is natural in the codomain. *)
+  Lemma biproduct_codiagonal_natural {Y Y' : object C}
+    `{BYY : @Biproduct C Z Y Y}
```
jdchristensen: It looks like you always have to write `@Biproduct C Z Y Y`, probably because `Z` can't be inferred.  That suggests that `Z` should be made an explicit argument, so you could write `Biproduct Z Y Y`.  Can you look for other places where the notation is awkward and can be improved?

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,305 @@ Section BiproductOperations.
...
+    (a : morphism C Y Y')
+    : (a o biproduct_codiagonal Y)%morphism
```
jdchristensen: Can these `%morphism` annotations be avoided?  Does the whole Categories library need to always annotate composition?  Is that because `o` is overloaded?

theories/Categories/Additive/Biproducts.v
```diff
@@ -171,8 +171,305 @@ Section BiproductOperations.
...
+      = biproduct_coprod_mor BYY Y' a a.
+  Proof.
+    unfold biproduct_codiagonal.
+    rewrite biproduct_coprod_mor_nat.
+    repeat rewrite right_identity.
+    reflexivity.
+  Qed.
+
+  (** The functorial biproduct map commutes with pairing. *)
+  Lemma biproduct_sum_map_prod {X Y X' Y' W : object C}
```
jdchristensen: Not about self-biproducts.

jdchristensen (commented):
So far I just skimmed Biproducts.v and there are some obvious mistakes in organization.  Please re-read SemiAdditive.v and look for similar issues.  Can you also check whether all of my previous comments have been addressed?  (Even if github flags them as outdated, they are often still relevant.)

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,278 @@
+(** * Semi-additive categories
+
+    Categories with zero objects and biproducts, which automatically
+    have commutative monoid structures on hom-sets. *)
+
+From HoTT Require Import Basics.Overture.
+From HoTT.Categories Require Import Category.Core.
+From HoTT.Categories.Additive Require Import ZeroObjects Biproducts.
+From HoTT.Classes.interfaces Require Import abstract_algebra.
```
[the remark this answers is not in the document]
jdchristensen: The reason the build succeeds is because those dependencies must be coming via one of the other imports.  But I agree with copilot that we should explicit require the things we use.  (I don't remember for sure, but I may have been the one to suggest trimming the imports too aggressively.)

CharlesCNorton:
Fixed in b32c4cf6 and 17de95f2: the zero object is now explicit (`Biproduct Z X Y`), the misplaced lemmas moved into functoriality/symmetry/self-biproduct sections with corrected comments, imports are explicit, and swapping twice is proven to be the identity. On `%morphism`: `o` is also function composition, and `Bind Scope` only covers arguments of type `morphism`, not equations — opening `morphism_scope` locally per file removes the annotations entirely. I went back through all the earlier comments; the one thing I left alone is making the biproduct argument of the operations implicit, since typeclass search picks an arbitrary instance when two biproducts of the same pair are in scope, as in the swap lemmas.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+  semiadditive_biproduct :: forall (X Y : object cat),
+    Biproduct semiadditive_zero X Y
+}.
+
+Coercion cat : SemiAdditiveCategory >-> PreCategory.
+
+(** Notation for biproduct objects *)
+Local Notation "X ⊕ Y" :=
+  (semiadditive_biproduct X Y)
+  (at level 40, left associativity).
```
jdchristensen: We usually reserve notations in Basics/Notations.v.  (I'm not exactly sure why, but I guess we should do so here too.)

theories/Categories/Additive/Biproducts.v
```diff
@@ -106,80 +109,408 @@ Arguments biproduct_universal {C Z X Y} b : rename.
...
+    := biproduct_prod_mor BX'Y' (a o outl BXY) (b o outr BXY).
+
+  (** The biproduct map composed with the left injection is the left
+      summand map. *)
+  Lemma biproduct_sum_map_inl
+    (a : morphism C X X') (b : morphism C Y Y')
+    : biproduct_sum_map a b o inl BXY
+      = inl BX'Y' o a.
+  Proof.
+    etransitivity (biproduct_prod_mor BX'Y' a (zero_morphism X Y')).
```
jdchristensen: I'll push a commit showing a slightly cleaner way to do this using `rhs_V`.  More generally, `lhs`, `lhs_V`, `rhs`, and `rhs_V` can be used to avoid using `rewrite` when the thing being rewritten is at the top level of one side of an equality.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+(** ** Commutativity of morphism addition *)
+
+Theorem morphism_addition_commutative (C : SemiAdditiveCategory)
+  (X Y : object C)
+  : Commutative (@sgop_morphism C X Y).
+Proof.
+  intros f g.
+  unfold sgop_morphism.
+  rewrite (biproduct_prod_swap f g).
+  rewrite <- Category.Core.associativity.
```
jdchristensen: I think `Category.Core.` can be removed here and elsewhere.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+    reflexivity.
+  Qed.
+
+  Lemma addition_of_pairs (X Y : object C) (f1 f2 g1 g2 : morphism C X Y)
+    : sgop_morphism C X (Y ⊕ Y)
+        (biproduct_prod_mor (Y ⊕ Y) f1 g1)
+        (biproduct_prod_mor (Y ⊕ Y) f2 g2)
+      = biproduct_prod_mor (Y ⊕ Y)
+          (sgop_morphism C X Y f1 f2)
+          (sgop_morphism C X Y g1 g2).
```
jdchristensen: I think this generalizes to `Y ⊕ Y'`.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+  Proof.
+    rapply biproduct_prod_unique.
+    - rewrite addition_postcompose.
+      rewrite biproduct_prod_beta_l.
+      rewrite biproduct_prod_beta_l.
```
jdchristensen: `rewrite 2` here and below?

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+    unfold sgop_morphism.
+    rewrite <- Category.Core.associativity.
+    rewrite (biproduct_codiagonal_nat a).
+    rewrite <- (biproduct_codiagonal_factor_through_sum_map a a).
+    rewrite Category.Core.associativity.
+    rewrite <- (biproduct_sum_map_prod a a f g).
+    reflexivity.
+  Qed.
+
+  Lemma addition_of_pairs (X Y : object C) (f1 f2 g1 g2 : morphism C X Y)
```
jdchristensen: Make X and Y implicit?  I think this applies to other results in this file too.  If the objects can be inferred from the morphisms, they should usually be implicit.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+  (** The zero morphism is the unit for addition. *)
+  #[export] Instance monunit_morphism : MonUnit (morphism C X Y)
+    := zero_morphism X Y.
+
+End MorphismAddition.
+
+(** ** Identity laws for morphism addition *)
+
+Section IdentityLaws.
+  Context (C : SemiAdditiveCategory).
```
jdchristensen: I think C should be implicit.  Elsewhere too.  And X Y below.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+Local Notation "X ⊕ Y" :=
+  (semiadditive_biproduct X Y)
+  (at level 40, left associativity).
+
+(** ** Morphism addition via biproducts
+
+    Addition is the codiagonal composed with the pairing morphism. *)
+
+Section MorphismAddition.
+  Context (C : SemiAdditiveCategory) (X Y : object C).
```
jdchristensen: All implicit?

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+      rewrite biproduct_prod_beta_r.
+      reflexivity.
+  Qed.
+
+  (** Associativity follows by functoriality of pairing and codiagonal. *)
+  Theorem morphism_addition_associative
+    (X Y : object C) (f g h : morphism C X Y)
+    : (f + g) + h = f + (g + h).
+  Proof.
+    unfold sgop_morphism at 1.
```
jdchristensen: This unfold doesn't seem to do anything.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+    etransitivity
+      (biproduct_codiagonal Y
+        o biproduct_prod_mor (Y ⊕ Y) (sgop_morphism C X Y f g)
+            (sgop_morphism C X Y (zero_morphism X Y) h)).
```
jdchristensen: This seems overly complicated.  Can this be expressed using `+` notation?  Can we avoid the etransitivity by adjusting the lhs with `zero_left_identity` directly?

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,279 @@
...
+  : IsCommutativeMonoid (morphism C X Y).
+Proof.
+  split.
+  - split.
+    + split.
+      * exact _.
+      * intros f g h.
+        unfold sg_op, sgop_morphism.
+        symmetry.
+        rapply (morphism_addition_associative C X Y).
```
jdchristensen: Maybe the lemma stating associativity should be in the same direction as the typeclass?  Then this can become as simple as the other cases are in the version I'm about to push.

jdchristensen (commented):
I made a quick review through everything.  I still want to make another pass through SemiAdditive.v after you respond to my feedback.

CharlesCNorton:
Arguments are implicit wherever inferable from the morphisms, `⊕` is reserved in Basics/Notations.v, `addition_of_pairs` is generalized to `Y ⊕ Y'`, the remaining `Category.Core.` qualifiers are gone, and associativity is restated in the typeclass direction. With the zero-identity adjustments done directly on the lhs, its proof is now four lines and the monoid instance fields are all bare exacts. Thanks for the commits and your patience.

theories/Categories/Additive/SemiAdditive.v
```diff
@@ -0,0 +1,248 @@
...
+End MorphismAddition.
+
+(** ** Identity laws for morphism addition *)
+
+Section IdentityLaws.
+  Context {C : SemiAdditiveCategory}.
+
+  (** Zero is a left identity for morphism addition. *)
+  Definition zero_left_identity {X Y : object C} (f : morphism C X Y)
+    : sgop_morphism (zero_morphism X Y) f = f
```
jdchristensen: Is there any reason to not use `+` and `0`, so this would read `: 0 + f = f`?  I think this could clarify lots of things in this file.

jdchristensen:
LGTM!  Can you squash to one commit?

CharlesCNorton:
done

jdchristensen:
Thanks!

PR #2306, opened 2025-08-11
Extend EncodeDecode with framework for homotopy groups

CharlesCNorton:
Adds systematic machinery for computing fundamental and higher homotopy groups, compositional structures for products, and type class automation. 

Enables computation of π₁ for Klein bottle and other surfaces.

### DIFF
--- a/theories/Homotopy/EncodeDecode.v
+++ b/theories/Homotopy/EncodeDecode.v
@@ -1,10 +1,10 @@
 Require Import Basics Pointed.
 Require Import Truncations.Core.
+Require Import Truncations.Connectedness.
+Require Import Types.Prod.
 
 (** ** Encode-decode method of characterizing identity types *)
-
 (** See Homotopy/IdentitySystems.v for a related characterization of identity types. *)
-
 Definition encode_decode
   (A : Type) (a0 : A) (code : A -> Type)
   (c0 : code a0) (decode : forall x, code x -> a0 = x)
@@ -21,7 +21,6 @@ Proof.
   - intros [].
     exact r.
 Defined.
-
 (** Encode-decode for truncated identity-types *)
 Definition encode_decode_trunc n
   (A : Type) (a0 : A) (code : A -> Type) `{forall a, IsTrunc n (code a)}
@@ -44,7 +43,6 @@ Proof.
     destruct p.
     exact r.
 Defined.
-
 (** Encode-decode for loop spaces *)
 Definition encode_decode_loops
   (A : pType) (code : pFam A)
@@ -52,8 +50,7 @@ Definition encode_decode_loops
   (s : forall (c : code (point A)), decode _ c # (dpoint code) = c)
   (r : decode _ (dpoint code) = idpath)
   : loops A <~> code (point A)
-  := encode_decode _ _ code (dpoint code) decode s r _.
-
+  := encode_decode A (point A) code (dpoint code) decode s r (point A).
 (** Encode-decode for truncated loop spaces *)
 Definition encode_decode_trunc_loops n
   (A : pType) (code : pFam A) `{forall a, IsTrunc n (code a)}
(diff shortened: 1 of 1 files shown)


theories/Homotopy/EncodeDecode.v
```diff
@@ -1,10 +1,10 @@
 Require Import Basics Pointed.
 Require Import Truncations.Core.
+Require Import Truncations.Connectedness.
+Require Import Types.Prod.
 
 (** ** Encode-decode method of characterizing identity types *)
-
```
jdchristensen, line 5: Please don't change unrelated white space.

theories/Homotopy/EncodeDecode.v
```diff
@@ -44,16 +43,14 @@ Proof.
...
 (** Encode-decode for loop spaces *)
 Definition encode_decode_loops
   (A : pType) (code : pFam A)
   (decode : forall x, code x -> point A = x)
   (s : forall (c : code (point A)), decode _ c # (dpoint code) = c)
   (r : decode _ (dpoint code) = idpath)
   : loops A <~> code (point A)
-  := encode_decode _ _ code (dpoint code) decode s r _.
-
+  := encode_decode A (point A) code (dpoint code) decode s r (point A).
```
jdchristensen, line 53: I think this is fine either way, but it's easier to review things if you don't make unrelated changes.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+    := fun x => isequiv_adjointify (decode x) (encode x) 
+                  (decode_encode x) (encode_decode x).
+
+  (** Fundamental group computation via encode-decode *)
+  Definition fundamental_group_via_encode_decode
+    `{IsHSet (Code a)}
+    `{IsConnected 0%trunc A}
+    (decode_encode : forall x (p : a = x), decode x (encode x p) = p)
+    (encode_decode : forall x (c : Code x), encode x (decode x c) = c)
+    : Tr 0 (a = a) <~> Code a
```
jdchristensen, line 87: This seems to be essentially the same as `encode_decode_trunc_loops` above, except that the `encode` function isn't assumed to be `transport`, and the condition `decode_encode` is required for all `p` rather than just for `idpath`.  Note that the `encode` function will always be homotopic to transport, so this isn't really any more general.  So I'm not sure what it adds.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+    := equiv_adjointify 
+         (Trunc_rec (encode a))
+         (fun c => tr (decode a c))
+         (encode_decode a)
+         (Trunc_ind (fun _ => _) (fun p => ap tr (decode_encode a p))).
+
+  (** Generalization to higher homotopy groups *)
+  Definition higher_group_via_encode_decode
```
jdchristensen, line 95: The same comments apply here.  But also, if we were to keep this, then there is no need for the special case when `n` is `0` to be stated separately.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+         (Trunc_rec (encode a))
+         (fun c => tr (decode a c))
+         (encode_decode a)
+         (Trunc_ind (fun _ => _) (fun p => ap tr (decode_encode a p))).
+
+  (** Computational extraction of group elements *)
+  Definition compute_fundamental_element
+    (loop : a = a)
+    : Code a
+    := encode a loop.
```
jdchristensen, line 112: I don't see what purpose this serves.  The user passes in an `encode` function and gets it back again.  (I'm also not sure in what sense this is "computational".  It's exactly as computational as the `encode` function provided by the user.)

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+
+  Fixpoint loop_power (n : nat) (loop : a = a) : a = a :=
+    match n with
+    | O => idpath
+    | S n' => loop @ (loop_power n' loop)
+    end.
```
jdchristensen, line 118: We already have `loopexp` is Spaces/Int.v, which handles integer powers, so not sure if this is needed.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+  Definition compute_iterated_fundamental_element
+    (n : nat)
+    (loop : a = a)
+    : Code a
+    := encode a (loop_power n loop).
+
+  Definition compute_inverse_element
+    (c : Code a)
+    : a = a
+    := decode a c.
```
jdchristensen, line 129: These two don't seem to be adding anything.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+
+(** ** Encode-decode from covering space data *)
+Section CoveringSpace.
+  Context {A : Type} {a : A}
+          (P : A -> Type)
+          `{forall x, IsHSet (P x)}
+          (p0 : P a).
+
+  Definition encode_from_covering : forall x, (a = x) -> P x
+    := fun x p => transport P p p0.
```
jdchristensen, line 141: I'm not sure what the point of this section is.  This is just renaming the `transport` function, which is already used heavily in the original part of the file.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+
+  Definition covering_to_code : A -> Type := P.
```
jdchristensen, line 143: This is literally `idmap`, sending `P` to `P`, so is not needed.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+
+  Context (path_lift : forall x, P x -> (a = x)).
+
+  Definition decode_from_covering : forall x, P x -> (a = x)
+    := path_lift.
```
jdchristensen, line 148: This is also `idmap`.  Moreover, there's no reason for a `Context` line when it's only used for one function.

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+
+End CoveringSpace.
+
+(** ** Compositional encode-decode for product spaces *)
+Section Compositional.
+
```
jdchristensen, line 154: I can't figure out what this Section is trying to achieve.  It seems to start something, but not finish it.  I also don't know in what sense this is "compositional".

theories/Homotopy/EncodeDecode.v
```diff
@@ -62,5 +59,142 @@ Definition encode_decode_trunc_loops n
...
+    (decodeA : forall x, CodeA x -> (a = x))
+    (decodeB : forall y, CodeB y -> (b = y))
+    : forall xy, (CodeA (fst xy) * CodeB (snd xy)) -> ((a,b) = xy)
+    := fun xy cc => path_prod (a,b) xy (decodeA (fst xy) (fst cc)) (decodeB (snd xy) (snd cc)).
+
+End Compositional.
+
+(** ** Type class for encode-decode structure *)
+Module Synthesis.
+
```
jdchristensen, line 179: I'm not sure how this would be useful.  Until there are convincing examples, I don't think it's worth having.

jdchristensen (commented):
Sorry, I don't see how this will be useful.  I know you are using AI to help, but this seems like an example of AI-generated code that typechecks but doesn't do things in a sensible way.  Maybe it's part of a bigger picture that gets somewhere, but that should be possible using the existing encode-decode functions.  If they don't suffice, they could be generalized, rather than being repeated with very minor tweaks.

CharlesCNorton:
closed and submitted typo correction PRs as penance

PR #2307, opened 2025-08-12
Various typo fixes

CharlesCNorton:
First commit - the comment in theories/Algebra/Rings/Matrix.v uses “analagous” instead of “analogous. Further commits below.


jdchristensen:
Thanks!  Can you squash it down to one commit?

PR #2309, opened 2025-08-24
Add results on compact and searchable types (mostly following Escardo), lemmas for HProps

thchatzidiamantis:
We add definitions and basic facts about compact and searchable types, including some examples and characterisations. We prove that the type `nat->X` is uniformly searchable if `X` is searchable. Useful lemmas about propositions and Σ-types are added to `TruncType` and `Sigma` respectively.


jdchristensen:
I'll just go ahead and merge, but if anyone has comments in the future, just let @thchatzidiamantis and me know.

PR #2314, opened 2025-09-19
chore: update nix flake

Alizter:


Alizter:
Used in CI and local development. Works for me so merging.

jdchristensen:
Currently we have opam-build, quick-build and build, and for each of those we build against three versions of Coq.  And in addition we have nix building against one version of Coq.  I wonder if we should only do some of those builds.  E.g. opam-build could build against v1, v2 and v4 of Coq, quick-build could build against v3 and v4, while build could be against v2 and v4, while nix is against v3.  (I'm just making this up, but the idea is that we mix and match, rather than doing the full cartesian product of the sets.)  OTOH, the builds run in parallel, so it doesn't waste much wall time, just github resources.

I also wondered about doing one build really fast, with more cores, before doing any others.  Then when you pushed something faulty, you'd find out very quickly (ideally in < 1 minute), and we wouldn't waste cpu time trying all of the other build variants.

Just some random thoughts, since you are adjusting things.

PR #2315, opened 2025-10-09
Adapt to Rocq#21180

Tragicus:
There are a few `repeat split` in definitions of groups, which causes `split` to try and solve equalities using reflexivity. This used to fail, but with https://github.com/rocq-prover/rocq/pull/21180 the unification algorithm becomes powerful enough to find a solution, albeit not the required one.

### DIFF
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -241,7 +241,14 @@ Section Reduction.
   (** Finally we have defined the free group on [A] *)
   Definition FreeGroup : Group.
   Proof.
-    snapply (Build_Group freegroup_type); repeat split; exact _.
+    snapply (Build_Group freegroup_type).
+    - exact _.
+    - exact _.
+    - exact _.
+    - split.
+      + repeat split; exact _.
+      + exact _.
+      + exact _.
   Defined.
   
   Definition word_rec (G : Group) (s : A -> G) : A + A -> G.
--- a/theories/Algebra/Groups/FreeGroup.v
+++ b/theories/Algebra/Groups/FreeGroup.v
@@ -242,13 +242,8 @@ Section Reduction.
   Definition FreeGroup : Group.
   Proof.
     snapply (Build_Group freegroup_type).
-    - exact _.
-    - exact _.
-    - exact _.
-    - split.
-      + repeat split; exact _.
-      + exact _.
-      + exact _.
+    4: split. 4: repeat split.
+    all: exact _.
   Defined.
   
   Definition word_rec (G : Group) (s : A -> G) : A + A -> G.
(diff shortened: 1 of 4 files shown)


theories/Algebra/Groups/FreeGroup.v
```diff
@@ -241,7 +241,14 @@ Section Reduction.
...
   Proof.
-    snapply (Build_Group freegroup_type); repeat split; exact _.
+    snapply (Build_Group freegroup_type).
+    - exact _.
+    - exact _.
+    - exact _.
+    - split.
+      + repeat split; exact _.
+      + exact _.
+      + exact _.
```
jdchristensen: I'm curious how reflexivity could possibly solve any of the goals here.  That doesn't look possible to me.  And it's a bit annoying to be unable to use `repeat split'.

Do either of the following work?
```
    snapply (Build_Group freegroup_type).
    4: repeat split.
    all: exact _.
```
or with the second line replaced with
```
    4: split.  4: split.  4: split.
```

theories/Algebra/Groups/FreeProduct.v
```diff
@@ -524,7 +524,14 @@ Section AmalgamatedFreeProduct.
...
   Proof.
-    snapply (Build_Group amal_type); repeat split; exact _.
+    snapply (Build_Group amal_type).
+    - exact _.
+    - exact _.
+    - exact _.
+    - split.
+      + repeat split; exact _.
+      + exact _.
+      + exact _.
```
jdchristensen: I'd also prefer something more concise here.

jdchristensen (commented):
I guess an alternative is a custom `justsplit' tactic that does nothing on equality goals and does split on all other goals...

theories/Algebra/Groups/FreeGroup.v
Tragicus: One of the goals is the fact that `inv x` is the left-inverse of `x`. Since the unit and the operation are not yet instantiated, there is a solution that makes `op` the constant operation returning the unit.
Oh, good catch, `4: repeat split` causes the issue, but `4: split. 4: repeat split` does work, thank you.

jdchristensen (approved):
I think this looks fine now, without needing a custom tactic.  I think we should only merge it if the Coq change is definitely going to be merged.  Do we know that yet or should I wait to find out?

jdchristensen (approved):
I like the latest version the most.  It's more robust, as once you've specified the operation, the remaining things are properties.

Alizter (approved):

Alizter:
@Tragicus What's the status of the upstream PR? It looks to have been inactive for a bit.

Tragicus:
It is stuck at an issue I do not know how to solve.

Alizter:
@Tragicus I'm happy to merge this if you would like. One more thing off your plate.

Tragicus:
Yes, thank you, the UniMath overlay has just been merged, so this is the only remaining one, which means we should be ready to merge the Rocq PR soon.

PR #2318, opened 2025-10-20
Changes for Rocq PR#21164

jdchristensen:
The changes here make Coq-HoTT build with Rocq PR https://github.com/rocq-prover/rocq/pull/21164

Each commit is independent, and in all but the last case, these are simplifications that are good to have anyways.  (The last case is a minor change, so not a big deal.)

@ppedrot @tabareau

### DIFF
--- a/contrib/HoTTBookExercises.v
+++ b/contrib/HoTTBookExercises.v
@@ -1109,7 +1109,7 @@ Defined.
 Definition Book_3_10_LEM@{j} := forall A : HProp@{j}, A + ~ A.
 
 Definition Book_3_10_Lift@{i j | i < j} (A : HProp@{i}) : HProp@{j}
-  := Build_HProp A.
+  := @Build_HProp A (trunctype_istrunc A).
 
 Definition Book_3_10_impred@{i j k | i < j, j < k} `{Univalence}
   (LEM : Book_3_10_LEM@{j})


theories/Truncations/Core.v
```diff
@@ -28,9 +28,9 @@ Module Export Trunc.
     tr : A -> Trunc n A.
   Arguments tr {n A} a.
 
-  (** Without explicit universe parameters, this instance is insufficiently polymorphic. *)
-  #[export] Instance istrunc_truncation (n : trunc_index) (A : Type@{i})
-    : IsTrunc@{j} n (Trunc@{i} n A).
+  (** Because [IsTrunc] is cumulative, we can use only one universe variable here. *)
+  #[export] Instance istrunc_truncation@{i} (n : trunc_index) (A : Type@{i})
+    : IsTrunc@{i} n (Trunc@{i} n A).
```
jdchristensen, line 33: We decided to make `IsTrunc` (really, `IsTrunc_internal`) cumulative, but this extra flexibility makes it harder for Rocq to choose universe variables optimally.  E.g. with the original version of `istrunc_truncation`, there is not a unique choice of `j`, since anything smaller than the desired `j` will work.  With this change, there is no longer the flexibility to choose `j`, making it easier for Rocq, but then Rocq can fail to find a solution in other situations.  It's a bit of a dilemma, and we might end up making different design decisions once we have algebraic universes.

tabareau:
@jdchristensen  I have updated the Rocq PR https://github.com/rocq-prover/rocq/pull/21164 to handle cumulative inductive type properly, so you can revert last commit  [HoTTBookExercises: make build with Rocq PR#21164](https://github.com/HoTT/Coq-HoTT/pull/2318/commits/b2dfaef636a9d173377fe2e860c9864fe0fa9d81)

PR #2319, opened 2025-10-29
Miscellaneous small changes

jdchristensen:
I've accumulated a bunch of small things.  Each commit is independent and affects only one file, so they can be reviewed separately or together.

### DIFF
--- a/theories/Spaces/Nat/Core.v
+++ b/theories/Spaces/Nat/Core.v
@@ -135,7 +135,7 @@ Scheme leq_ind := Induction for leq Sort Type.
 Scheme leq_rect := Induction for leq Sort Type.
 Scheme leq_rec := Induction for leq Sort Type.
 
-Notation "n <= m" := (leq n m) : nat_scope.
+Infix "<=" := leq : nat_scope.
 
 Existing Class leq.
 Existing Instances leq_refl leq_succ_r.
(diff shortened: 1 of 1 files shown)


theories/Homotopy/Join/Core.v
```diff
@@ -956,6 +964,13 @@ Section JoinPower.
     - exact (equiv_functor_join equiv_idmap IHn).
   Defined.
 
+(*
+  #[export] Instance isconnected_join_power `{Funext} {m : trunc_index}
+    (A : Type) `{IsConnected m A} (n : nat)
+    : IsConnected (??) (join_power A n).
+  -2 m m+2+m m+2+m+2+m  : So ?? should be -2 + n*(2+m).  But we don't have multiplication.
+*)
+
```
ThomatoTomato: What is the reason for adding in this commented block of code? If it is something we want to add in the future, is it better to make an issue and leave it out of the PR?

theories/Categories/Category/Morphisms.v
```diff
@@ -44,9 +44,6 @@ Local Infix "<~=~>" := Isomorphic : category_scope.
 (** ** Theorems about isomorphisms *)
 Section iso_contr.
   Variable C : PreCategory.
-
-
-
   Variables s d : C.
```
ThomatoTomato, line 47: Perhaps we should combine these statements into a single line: `Variables (C : PreCategory) (s d : C)`?

theories/Homotopy/Join/Core.v
jdchristensen: Thanks, this was an error!

theories/Categories/Category/Morphisms.v
jdchristensen: The style in this part of the library is to have only one type per line in most cases, so I'll leave it like that.  Thanks for the suggestion.

Alizter:
I would like to take a look tomorrow at the weekend if that's no problem.

Alizter:
I'm not certain. Since it's only dev that is failing it could be an upstream issue so that would have to be checked. I can take a look tomorrow if it persists.

jdchristensen:
I re-ran the jobs, and there was no error this time, so it was just a temporary glitch.

theories/Basics/Settings.v
```diff
@@ -2,8 +2,6 @@
 
 (** This file contains all the tweaks and settings we make to Coq. *)
 
-(** ** Warnings *)
```
Alizter, line 5: I think I kept this as the natural place to include warnings. I don't feel strongly about keeping it but just FYI.

Alizter (approved):

theories/Basics/Settings.v
jdchristensen: Yeah, that's what I thought, but we may as well keep things simple until we need them.

Issue #2316, opened 2025-10-10
SetoidRewrite.v


jdchristensen:
I'm looking again at the file contrib/SetoidRewrite.v that @patrick-nicodemus kindly contributed to the library in late 2022.  See #1684 and #1695 .

I think this is going to be quite handy for some work in progress with @ThomatoTomato .

A couple of updates to the discussion from back then:

At least as of Rocq 9.0, SetoidRewrite.v builds fine without having the Rocq standard library installed.  You just need the core part of Rocq.  So this is promising, and makes me think we should consider moving it back to the main library.

However, adding `Require Coq.Init.Tactics` at the top of Basics/Settings.v does cause at least one build failure, in Spaces/Nat/Core.v.  The line `inversion leq_m_Sn as [p | k H p]` produces a `p` of type `Logic.eq` instead of a path.  This isn't the end of the world, as we could just import SetoidRewrite.v when needed, but it would be better if we had full compatibility with the library.  Does anyone know a (robust) fix for this particular issue?

Even better, can someone see if they can make the whole library build with SetoidRewrite loaded?  I'd estimate that around half of the library built fine (with no warnings) with Coq.Init.Tactics loaded, so it's probably not too hard to get the whole library to build.  I'm still on Coq 8.20.1, so that's what I tested with.  So another question is whether things are already better with Rocq 9.0 or 9.1.

@Alizter @patrick-nicodemus @ThomatoTomato

patrick-nicodemus:
You can use `ocamldebug` to step through Rocq interactively to debug something, so I spent some time today configuring that hoping to step through the execution of `rewrite` to see what happens. I haven't quite mastered the use of it.

jdchristensen:
Here's a further data point that I think can justify moving SetoidRewrite into the main library as soon as 9.0 is our minimum.  What I found is that we no longer need to ensure that Coq.Init.Tactics is the first thing imported into a file.  Importing it used to clear the typeclass database (e.g. this happens in 8.20.1), but with 9.0 and 9.1 I tested that it doesn't do that anymore.  (With 9.0 I did still find one or two glitches when it is imported later, but they are rare.)

So the upshot is that with 9.0 and up, it is ok to selectively use SetoidRewrite in the library, as it doesn't then require importing Coq.Init.Tactics in all of the dependencies.

It would still be good to figure out a way to ensure that importing Coq.Init.Tactics doesn't break existing proofs, but this is less of a blocking issue now.

[...]

PR #2323, opened 2025-11-09
get the library working with Coq.Init.Tactics loaded

patrick-nicodemus:
Branch for trying to get the library working with `Coq.Init.Tactics` loaded. Dan reported an error where `inversion` uses `Logic.eq` instead of `paths`, so we try to fix it by registering appropriate terms for `core.eq.*`, but this causes breakages in the `rewrite` tactic, how can we fix this?

### DIFF
--- a/theories/Basics/Overture.v
+++ b/theories/Basics/Overture.v
@@ -282,6 +282,10 @@ Definition paths_rect := paths_ind.
 Register paths as core.identity.type.
 Register idpath as core.identity.refl.
 Register paths_rect as core.identity.ind.
+Register paths as core.eq.type.
+Register idpath as core.eq.refl.
+Register paths_ind as core.eq.ind.
+Register paths_rect as core.eq.rect.
 
 Notation "x = y :> A" := (@paths A x y) : type_scope.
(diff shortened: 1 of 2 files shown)


theories/Basics/Overture.v
```diff
@@ -282,6 +282,10 @@ Definition paths_rect := paths_ind.
 Register paths as core.identity.type.
 Register idpath as core.identity.refl.
 Register paths_rect as core.identity.ind.
+Register paths as core.eq.type.
+Register idpath as core.eq.refl.
+Register paths_ind as core.eq.ind.
```
jdchristensen, line 287: Is it possible that `core.eq.ind` is expecting an induction principle for sort `Prop` and providing one of sort `Type` is an issue?  (Unlikely, but I thought I'd check.)

jdchristensen (commented):
Thanks for working on this!  I'm not sure how to proceed, but I made comments in #2316.

jdchristensen:
Based on other developments, the issues with Coq.Init.Tactics are gone, so I'm going to close this.

Issue #2324, opened 2025-11-12
Confusion of orientation in Cubical/PathSquare.v


marcbezem:
It seems there is a typo in line 13 of Cubical/PathSquare.v, as two points are labelled x01:
```
(* 
        x00 ----pi0---- x01
         |               |
         |               |
        p0i     ==>     p1i
         |               |
         |               |
        x01-----pi1-----x11
 *)
```
Since the top path is labelled pi0, it seems logical that the top right point is x10.
(This is also compatible with the convention of an xy-plane with horizontal x-axis.)
However, the first argument of PathSquare has the type a00 = a10, and the order
of the arguments would then become top, bottom, left, right. All later references
to this order, as well as to 'horizontal' and 'vertical', would be confusing, or even wrong
I happen to work on this file (adding an interchange lemma), and would propose
to change the diagram into:
```
(* 
        x00 ----p0i---- x01
         |               |
         |               |
        pi0     ==>     pi1
         |               |
         |               |
        x10-----p1i-----x11
 *)
```
together with a cautioning remark about the yx-plane reading.
This would be a much smaller change than changing the rest of the file.

jdchristensen:
That sounds reasonable to me.  I think you should also change the `x`s into `a`s to match the formalization.  @Alizter do you agree with both suggestions?

PR #2325, opened 2025-11-14
Cubical/PathSquare.v: edit diagram in comment + formulate interchange lemma

marcbezem:
1. Please take a look at the edited diagram at the beginning of PathSquare.v, fixing #2324.
2. I added a formulation of an interchange lemma at the end.
2a. (minor) I didn't get the infix operators `@@h, @@v` to work.
2b. Just before `Admitted`, the proof state is:
```
1 goal
A : Type
x : A
h2i, vi2 : x = x
sq10 : PathSquare 1 1 1 h2i
sq01 : PathSquare 1 vi2 1 1
sq_concat_h (sq_concat_v 1%square sq10) (sq_concat_v sq01 1%square) =
sq_concat_v (sq_concat_h 1%square sq01) (sq_concat_h sq10 1%square)
```
 My plan would be the following. Through `equiv_sq_path` and some easy path algebra, the `sq`-variables correspond to 2-paths between `1 : x=x` and `h2i, vi2`, respectively. The latter are also variables, so we can do path induction on the 2-paths and assume they are reflexivity paths. In that case the `sq`-variables correspond  both to a `1%square` and the goal can be solved by reflexivity.

I'm new to this library, so I have difficulties getting this idea to work. Any help will be appreciated, both regarding the formulation of the interchange lemma and the feasibility of my idea to prove it.

### DIFF
--- a/theories/Cubical/PathSquare.v
+++ b/theories/Cubical/PathSquare.v
@@ -614,6 +614,26 @@ Defined.
 
 (* Interchange playground *)
 
+Definition experimental {A}
+(a : A)
+(P : A -> Type)
+(f : forall (x : A), (a = x) -> P(x))
+(g : forall (x : A), P(x) -> (a = x))
+(fxgxid: forall (x : A) (p : P x), (p = f x (g x p)))
+:
+forall (x : A)(p : P x), p = transport P (g x p) (f a 1).
+
+Proof.
+  intros.
+  remember (g x p) as gxp eqn:H.
+  destruct gxp. (* Syntax error: destruct (g x p) as gxp eqn:H.  *)
+  simpl. (* Here you see that p = f a 1 if g a p = 1 *)
+  destruct H.
+  apply fxgxid.
+Defined.
+
+
+
 Definition interchange {A}
 {a00 a10 a20 a01 a11 a21 a02 a12 a22 :A} (** 9 points in big square of 2x2 squares *)
 (** sq00, top left, 4 paths *)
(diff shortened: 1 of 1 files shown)


jdchristensen:
The following gets you a bit further:
```coq
Proof.
  destruct sq00, sq11.
  destruct vi2, h2i.
  simpl.
  rewrite (sq_concat_h_1s (p0y:=1) (p1y:=1) sq01).
  simpl.
  (* We're missing the analogous [sq_concat_v_1s].  Even with it, it's not clear to me how to proceed. *)
Admitted.
```
The reason I choose to destruct `vi2` and `h2i` on the second line is that the definition of the horizontal and vertical compositions uses `concat_p1`, so we want to destruct the paths that appear there, so that `concat_p1` computes to `refl`.  Then we have a computation law for horizontal composition of `sq_refl` on the left.  We're missing the corresponding result for vertical composition.  Even if we had it, I'm not sure what I would do next.

I've basically never used this part of the library.  Maybe @Alizter has more ideas?

jdchristensen:
> 2a. (minor) I didn't get the infix operators `@@h, @@v` to work.

Those notations are defined within a Section, and notations don't persist outside of the section they are defined in.  And it seems that Notation and Infix don't support `#[export]` or `Global`.   So those notations should be moved to right after the Section closes.  The `hr` notation is maybe too short to export globally, so it should remain in its section.

marcbezem:
See #2327. If that one is OK I didn't need help with `git`, but thanks anyway!

Alizter:
@marcbezem Generally PRs focusing on one particular issue or addition are easier to review and merge since my time budget is limited. If you have a bunch of loosely related changes that's fine too.

Here we have a GitHub bug. The other PR had the same commits and made this PR be considered merged. That wasn't my intention with merging #2327. What you will have to do is re-introduce the experiment you had the end, which is evidently now in the commit history. If you make a PR for that we can continue the discussion of how best to finish the goal there.

PR #2328, opened 2025-11-19
An interchange law for squares and cleanups to Cubical/PathSquare.v

marcbezem:
As asked by @Alizter .

One simplification added, using the library (`experimental` is not used, but reminds me of the plan). I'm still thinking what should be `remember`-ed to get from `h0j` back to `sq01`.

### DIFF
--- a/theories/Cubical/PathSquare.v
+++ b/theories/Cubical/PathSquare.v
@@ -9,7 +9,9 @@ Local Unset Elimination Schemes.
 
 (** * Homogeneous squares *)
 
-(** 
+(** Squares are fillers in diagrams of the following form:
+
+<<
         a00 ----p0i---- a01
          |               |
          |         >     |
@@ -17,7 +19,8 @@ Local Unset Elimination Schemes.
          |     =         |
          |               |
         a10-----p1i-----a11
- 
+>>
+
 Indexing of points in a square follows the convention for matrices, first a row index, then a column index. Unless stated otherwise, paths are oriented in the direction of increasing index i (or x, etc). The stylized 2-path on the antidiagonal goes from [pi0 @ p1i] to [p0i @ pi1], complying with the definition of [equiv_sq_path] below. *)
 
 (** Contents:
@@ -601,7 +604,25 @@ Section KanUnique.
 
 End KanUnique.
 
-(** Interchange law. *)
+(** An interchange law between horizontal and vertical composites in the following square:
+<<
+            h0i          h0j
+       a00 ━━━━━━━━ a01 ━━━━━━━━ a02
+        ┃            ┃            ┃
+        ┃  sq00      ┃  sq01      ┃
+    vi0 ┃        vi1 ┃            ┃ vi2
+        ┃            ┃            ┃
+        ┃   h1i      ┃   h1j      ┃
+       a10 ━━━━━━━━ a11 ━━━━━━━━ a12
+        ┃            ┃            ┃
+        ┃  sq10      ┃  sq11      ┃
+    vj0 ┃        vj1 ┃            ┃ vj2
+        ┃            ┃            ┃
+        ┃   h2i      ┃   h2j      ┃
+       a20 ━━━━━━━━ a21 ━━━━━━━━ a22
+>>
+*)
+
 Definition sq_interchange {A}
   {a00 a10 a20 a01 a11 a21 a02 a12 a22 : A} (* 9 points in big square of 2x2 squares *)
   (* sq00, top left, 4 paths *)


jdchristensen:
I figured out that you can write induction principles for square that only require *one* of the edges to be free.  They make this proof trivial, and probably make other proofs trivial too.  So the file can probably be rearranged so that these are earlier and they are used more often.  They depend on the Kan fillers, but those have no dependencies, so could appear near the start of the file.  I have to run and will be busy for a day or two, so feel free to keep working on these things.

marcbezem:
The new `pathsquare_ind_?` functions are very powerful. Surprisingly, only one suffices!

Alizter:
FTR we had a discussion before about using the equivalent definition of a square using paths directly and the transition mainly went smoothly. The only proof I recall getting stuck at was the T=S1xS1 proof in Homotopy/Torus. This was my original motivation for introducing these kinds of squares originally. I'm thinking that these single edge induction principles might help out with the Torus proof actually, but I won't dwell too much on it here.

jdchristensen:
> FTR we had a discussion before about using the equivalent definition of a square using paths directly and the transition mainly went smoothly.

What do you mean by "the transition"?  PathSquare doesn't yet use the path definition, so you must be talking about somewhere else in the library?

Alizter:
Sorry I wasn't clear in my message. I meant that when I tried to redefine PathSquare using the path definition.

marcbezem:
I did some clean up and feel we could update from draft PR to full PR. Feel free to clean up a bit more.

I'm planning a follow up, but have still to think about it. Also, I'm unsure where to put it.
It's about formalizing a sketch by Ulrik Buchholtz of an alternative proof of Lemma 2.7(3) from his joint paper with Dan et al. https://arxiv.org/pdf/2301.02636, using squares.

theories/Cubical/PathSquare.v
```diff
@@ -497,19 +508,10 @@ Section Kan.
...
-    - intros px1; exact (sq_fill_l px1 p0x p1x).1.
-    - intros px1.
-      exact (sq_fill_r_uniq (sq_fill_r _ p0x p1x).2
-                            (sq_fill_l px1 p0x p1x).2).
-    - intros px0.
-      exact (sq_fill_l_uniq (sq_fill_l _ p0x p1x).2
-                            (sq_fill_r px0 p0x p1x).2).
-  Defined.
+    : (a00 = a10) <~> (a01 = a11)
+    := equiv_concat_r p1x _ oE equiv_concat_l p0x^ a10.
```
jdchristensen: @Alizter Are you ok with me removing this?

theories/Cubical/PathSquare.v
```diff
@@ -527,6 +529,104 @@ Section Kan.
...
+  Definition sq_fill_b_uniq'
+             {px0 : a00 = a10} {px1 : a01 = a11} {p0x : a00 = a01}
+    : Contr {p1x : a10 = a11 & PathSquare px0 px1 p0x p1x}.
+  Proof.
+    apply (Build_Contr _ (sq_fill_b px0 px1 p0x)).
+    intros [p1x' s'].
+    by destruct s'.
+  Defined.
+
+  (* TODO: Should we remove the earlier uniqueness results which are weaker? *)
```
jdchristensen: @Alizter I suggest that we remove the earlier `*_uniq` results, since they just assert that the first components are equal.  Then we can rename these to have no prime at the end.  Sound good?

jdchristensen: Actually, it's a bit fiddly to deduce the previous uniqueness results using the new ones, and the previous ones were used in DPathSquare, so I added them back in with one-line proofs using the new contractibility results.  I'll be pushing this once I've finished my cleanup.

theories/Cubical/PathSquare.v
```diff
@@ -616,6 +616,37 @@ Section KanUnique.
 
 End KanUnique.
 
+(** Interchange law. *)
```
Alizter: Here is a diagram that may aid readers:
```
        h0i          h0j
   a00 ━━━━━━━━ a01 ━━━━━━━━ a02
    ┃            ┃            ┃
    ┃  sq00      ┃  sq01      ┃
vi0 ┃        vi1 ┃            ┃ vi2
    ┃            ┃            ┃
    ┃   h1i      ┃   h1j      ┃
   a10 ━━━━━━━━ a11 ━━━━━━━━ a12
    ┃            ┃            ┃
    ┃  sq10      ┃  sq11      ┃
vj0 ┃        vj1 ┃            ┃ vj2
    ┃            ┃            ┃
    ┃   h2i      ┃   h2j      ┃
   a20 ━━━━━━━━ a21 ━━━━━━━━ a22
```

jdchristensen: Nice!  I've added that, including the unicode, with << >> markers.  Let's see how it looks in the generated documentation.  @Alizter Can you remind me where I can find the coqdoc/alectryon documentation that the CI generates?

jdchristensen:
Since I changed the title, I'll say explicitly that this finishes the work started in #2325, which was not actually merged.

theories/Cubical/PathSquare.v
Alizter: It's not possible to view it in the CI for a PR (online), but the main CI will upload and host the artefacts. Links to the hosted docs are on the wiki https://github.com/HoTT/Coq-HoTT/wiki

Alizter: If you go to the specific CI job there is an upload artifact step which gives you a link to download the produced docs. That's technically one way you can view it.

Alizter: I don't see the diagram in the recent commit btw, did you push it?

jdchristensen: Whoops, I only commited part of the changes I made.  I just force pushed to fix the last commit.

jdchristensen: We can just merge this once @marcbezem takes a look, and then check the documentation at that point.

PR #2331, opened 2025-11-27
Adapt with respect to rocq-prover/rocq#21098

tabareau:


tabareau:
Hi @jdchristensen, you can revert the PR, I will do a new one that is backward compatible with a modification of PR#21098. However, the new mechanism implemented in PR#21098 for rewriting cannot be used in a backward compatible way for Coq-HoTT as the current code explicitly exposes the existence of internal_rew_... that no longer exists with the new mechanism. Maybe you should revisit your development process as having the same branch for benefiting from new Rocq features and being at the same time backward compatible with older version is fundamentally not possible.  Such a process forces you to benefit from new features of the oldest supported version only.

jdchristensen:
@tabareau It looks like you might be doing the revert now?  If so, I'll wait.

PR #2333, opened 2025-12-20
Wildcat pullbacks

jdchristensen:
This is the first of what will be a series of pull requests to merge in work done with @ThomatoTomato on diagram chasing.  I'm breaking it up into smaller chunks to make it easier to review in stages.  A snapshot (not completely up-to-date) of the full set of changes can be seen in the diagram-chasing branch here:  https://github.com/jdchristensen/HoTT-diagram-chasing

The first five commits in this PR are some minor cleanups that were made along the way.  These are probably best viewed one commit at a time.

The remaining five commits are the new material on pullbacks in wild categories.  These could be viewed one commit at a time or together.

I'll wait until this is merged before doing the next PR.

### DIFF
--- a/theories/WildCat/Adjoint.v
+++ b/theories/WildCat/Adjoint.v
@@ -32,13 +32,13 @@ Record Adjunction {C D : Type} (F : C -> D) (G : D -> C)
   `{Is1Cat C, Is1Cat D, !Is0Functor F, !Is0Functor G} :=
 {
   equiv_adjunction (x : C) (y : D) : (F x $-> y) <~> (x $-> G y) ;
-  (** Naturality condition in both variable separately *)
+  (** Naturality condition in both variables separately. *)
   (** The left variable is a bit trickier to state since we have opposite categories involved. *)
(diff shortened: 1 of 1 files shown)


theories/WildCat/Adjoint.v
```diff
@@ -67,13 +67,9 @@ Lemma fun01_profunctor {A B C D : Type} (F : A -> B) (G : C -> D)
 Proof.
   snapply Build_Fun01.
   1: exact (functor_prod F G).
-  rapply is0functor_prod_functor.
+  rapply is0functor_prod_functor.  (* Why not found by typeclass search? *)
```
Alizter: Is `functor_prod` blocking the search? Does unfolding functor_prod work and then exact work?

jdchristensen: I investigated a bit, and it has to do with the opposite categories.  I see that a similar issue happens in other places, so I just replaced the comment with an explanation.  (I also adjusted a couple of other things I saw.)

PR #2334, opened 2025-12-22
move SetoidRewrite to main library and other changes in preparation for diagram chasing

jdchristensen:
This is a fairly big PR that contains most of the work @ThomatoTomato and I have done on diagram chasing.  (But there's at least one more PR still, which will show that AbGroup satisfies the assumptions of our set-up.)

The commits mostly touch disjoint files, and so can be read individually or sequentially.  But since it's large, it's probably best to read them sequentially, which gets the dependencies correct.  The library builds after each commit, even with 8.20.

The first commit just fixes a missing #[export] on an instance.

The second commit reduces the use of the meta-files WildCat.v, AbGroups.v and Pointed.v.  WildCat.v is now never used internally.  This has been done so that the next commit builds with 8.20, but even after we move to >= 9.0, this commit is valuable, as it means that a large part of the library doesn't need to be rebuilt when a leaf file like HomologicalAlgebra.v is changed.  Also, the parallelization of the build is better, which saves some time.

The third commit moves SetoidRewrite.v from contrib to theories/WildCat/SetoidRewrite.v, with the tests moving to tests/WildCat/SetoidRewrite.v.  No changes were made besides imports.  This file is very useful for the work in commits 5, 6 and 7.

The fourth commit just has some minor changes to ZeroGroupoid.v.

The fifth commit (AbEnriched.v) introduces wild categories enriched in abelian groups, where the abelian group structure laws only hold up to 2-cells.

The sixth commit (Regular.v) introduces IsRegular and IsAbRegular wild categories, the latter being a nice place in which to do diagram chasing.  Regular.v has the key tactic which makes diagram chasing go through smoothly.

The seventh commit (HomologicalAlgebra.v) proves many diagram lemmas using this method.

Edit: two commits added, see below for descriptions.

Edit: reduced this PR to the first four commits, plus some later things related to those commits.  The rest will come as separate PRs.

### DIFF
--- a/test/WildCat/SetoidRewrite.v
+++ b/test/WildCat/SetoidRewrite.v
@@ -54,7 +54,7 @@ Section SetoidRewriteTests.
   Proof.
     intros A ? ? ? a b c eq_ab eq_bc.
     by rewrite eq_ab, <- eq_bc.
-  Abort.
+  Defined.
 
   Goal forall (A : Type) `(H : Is0Gpd A) (a b c : A),
       a $== b -> b $== c -> a $== c.
@@ -64,14 +64,14 @@ Section SetoidRewriteTests.
     rewrite eq_ab, <- eq_bc.
     rewrite eq_bc.
     by rewrite <- eq_bc.
-  Abort.
+  Defined.
 
   Goal forall (A B : Type) (F : A -> B) `{Is1Functor _ _ F} (a b : A) (f g : a $-> b), f $== g -> fmap F f $== fmap F g.
   Proof.
     do 17 intro.
     intro eq_fg.
     by rewrite eq_fg.
-  Abort.
+  Defined.
 
   Goal forall (A : Type) `{Is1Cat A} (a b c : A) (f1 f2 : a $-> b) (g : b $-> c), f1 $== f2 -> g $o f1 $== g $o f2.
   Proof.
@@ -79,7 +79,7 @@ Section SetoidRewriteTests.
     intro eq.
     rewrite <- eq.
     by rewrite eq.
-  Abort.
+  Defined.
 
   Goal forall (A : Type) `{Is1Cat A} (a b c : A) (f : a $-> b) (g1 g2 : b $-> c), g1 $== g2 -> g1 $o f $== g2 $o f.
   Proof.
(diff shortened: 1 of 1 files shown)


test/WildCat/SetoidRewrite.v
```diff
@@ -0,0 +1,98 @@
...
+From Coq Require Init.Tactics. (* TODO: Can remove this line and previous once Rocq 9.0 is our minimum. *)
+From HoTT Require Import Basics.Overture Basics.Tactics.
+From HoTT.WildCat Require Import Core NatTrans Equiv SetoidRewrite.
+
+(** * Examples and tests of setoid rewriting *)
+
+(** ** Examples of setoid rewriting *)
+
+(** Rewriting works in hypotheses. *)
+Proposition IsEpic_HasSection {A} `{Is1Cat A}
```
Alizter: Should we adapt this to the usual naming convention?

jdchristensen: Good catch.  Fixed, along with a few other things I noticed, in commit ten.

jdchristensen:
@Alizter If you want to just review an initial segment of this sequence of commits, that's fine, and I can split the rest off.  I included a lot, because the later commits show what the earlier work is used for, but I know it's a lot to digest, so I'm happy to split it up.  (We also have noticed that some things in HomologicalAlgebra.v can be cleaned up, so you might choose to skip that commit if you want to wait for us to clean it up first.)  Cc: @ThomatoTomato

test/WildCat/SetoidRewrite.v
```diff
@@ -0,0 +1,103 @@
...
+  intros faithful_F x y f g eq_Gf_Gg.
+  apply faithful_F.
+  apply (cate_monic_equiv (tau y)).
+  rewrite 2 (isnat tau).
+  by apply cat_prewhisker.
+Defined.
+
+(** ** Tests of setoid rewriting *)
+
+Section SetoidRewriteTests.
```
Alizter, line 50: We should move these to the tests and remove the aborts.

jdchristensen: This is already in the test folder, but I have gone ahead and changed Abort to Defined in all places.

PR #2335, opened 2026-01-02
AbEnriched: wild categories enriched in abelian groups

jdchristensen:
This is one of the commits extracted from #2334, which works towards doing diagram chasing.  It has been extended with a few more tests and a couple of clean ups.

It sets up the theory of wild categories enriched in abelian groups.  Within AbEnriched.v, the theory of a 0-groupoid abelian group is set up.  If it's needed elsewhere in the future, it could be factored out.  Also, if we ever need non-abelian 0-groupoid groups, we could easily remove commutativity, and add it in a second layer.  But for now, I kept things to what we need for diagram chasing.

### DIFF
--- a/theories/WildCat/AbEnriched.v
+++ b/theories/WildCat/AbEnriched.v
@@ -9,13 +9,13 @@ Require Export Classes.interfaces.canonical_names (SgOp, sg_op,
     MonUnit, mon_unit, Inverse, inv).
 Export canonical_names.BinOpNotations.
 
+Local Open Scope mc_add_scope.
+
 (** * Wild categories enriched in abelian groups *)
 
 (** ** 0-groupoid abelian groups *)
 
-(** Hom sets in wild categories are 0-groupoids, and we'd like to put an abelian group structure on these hom sets that satisfies the axioms up to 1-cells, to avoid needing function extensionality.  So we define the abstract idea of a 0-groupoid with an abelian group structure.  Because we use 1-cells, not paths, we can't reuse any of the work done in Algebra/Groups or Algebra/AbGroups. *)
-
-Local Open Scope mc_add_scope.
+(** Hom sets in wild categories are 0-groupoids, and we'd like to put an abelian group structure on these hom sets that satisfies the axioms up to 1-cells, to avoid needing function extensionality.  So we define the abstract idea of a 0-groupoid with an abelian group structure.  Because we use 1-cells, not paths, we can't reuse any of the work done in Algebra/Groups or Algebra/AbGroups. Note that our definition has redundant fields, which could be filled in using a "smart constructor":  [is0functor_inverse_0gpd], [mon_unit_r_0gpd], [inverse_r_0gpd] and the fact that [sgop_0gpd] is a 0-functor in the second variable. *)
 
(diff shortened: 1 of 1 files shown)


theories/WildCat/AbEnriched.v
```diff
@@ -0,0 +1,344 @@
...
+(** See Groups/Group.v and AbGroups/AbelianGroup.v for additional things that it might be useful to export in the future. *)
+Require Export Classes.interfaces.canonical_names (SgOp, sg_op,
+    MonUnit, mon_unit, Inverse, inv).
+Export canonical_names.BinOpNotations.
+
+(** * Wild categories enriched in abelian groups *)
+
+(** ** 0-groupoid abelian groups *)
+
+(** Hom sets in wild categories are 0-groupoids, and we'd like to put an abelian group structure on these hom sets that satisfies the axioms up to 1-cells, to avoid needing function extensionality.  So we define the abstract idea of a 0-groupoid with an abelian group structure.  Because we use 1-cells, not paths, we can't reuse any of the work done in Algebra/Groups or Algebra/AbGroups. *)
```
Alizter: Presumably if we generalised that work to groups internal to a given wild category we could reuse it. But that doesn't seem like a good payoff in this situation.

theories/WildCat/AbEnriched.v
```diff
@@ -0,0 +1,344 @@
...
+
+Class IsAbGroup_0gpd (A : Type) `{Is0Gpd A} := {
+    (* Data: *)
+    sgop_0gpd :: SgOp A;
+    mon_unit_0gpd :: MonUnit A;
+    inverse_0gpd :: Inverse A;
+
+    (* 0-functoriality: *)
+    is0bifunctor_sgop_0gpd :: Is0Bifunctor sgop_0gpd;
+    is0functor_inverse_0gpd :: Is0Functor inverse_0gpd;
```
Alizter, line 28: This is redundant data right? I cannot recall the details but there are various things that follow from eachother. Might be worth adding a comment so that we can introduce the "smart constructor" versions at some point. The definition however is fine.

theories/WildCat/AbEnriched.v
```diff
@@ -0,0 +1,344 @@
...
+
+(** We will try to parallel the naming in Group.v, when applicable. *)
+
+Section Lemmas.
+
+  Context {A : Type} `{IsAbGroup_0gpd A}.
+  Context (x y : A).
+
+  Definition inv_V_gg_0gpd : (-x) + (x + y) $== y
+    := (assoc_0gpd _ _ _)^$ $@ fmap10 sg_op (inverse_l_0gpd x) y $@ mon_unit_l_0gpd _.
```
Alizter, line 65: Is there a reason to use `sg_op` over `(+)` here? Might be the `sg_op` vs `add` thing, but I forget the details.

jdchristensen: Both work, as does `sgop_0gpd`.  But I think `sg_op` is probably the most clear.

theories/WildCat/AbEnriched.v
jdchristensen: I updated the comments and will merge when the CI is green.

PR #2336, opened 2026-01-04
EpiStable.v: categories in which one can do diagram chasing

jdchristensen:
This is a continuation of the work in #2334 and #2335.

We set up a theory of "regular" categories.  These aren't the same as the usual notion of regular category.  We chose a definition that is much more elementary to work with, with no need for coequalizers, but there wasn't an obvious alternate name.  Moreover, we plan to generalize this in the future to allow the user of this file to *specify* a class of epimorphisms they are interested in, which could then be chosen to be the effective epimorphisms, in which case this would be almost the same as a regular category.  (We still wouldn't be assuming the existence of coequalizers of kernel pairs.  But that hasn't come up so far, so I don't want to assume it.)  So maybe the name "regular" is ok?

### DIFF
--- a/theories/WildCat/EpiStable.v
+++ b/theories/WildCat/EpiStable.v
@@ -4,25 +4,25 @@ Require Import WildCat.Core WildCat.Equiv WildCat.EquivGpd WildCat.PointedCat
   WildCat.Yoneda WildCat.Graph WildCat.ZeroGroupoid WildCat.Pullbacks
   WildCat.AbEnriched WildCat.FunctorCat.
 
-(** * A variant of regular categories in which one can do diagram chasing *)
+(** * Epi-stable categories *)
+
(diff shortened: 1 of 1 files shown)


jdchristensen:
> Do we have some examples, perhaps in some tests of the tactics in action?

The next PR will contain the examples.  (See HomologicalAlgebra.v in early commits in #2334, if they are still cached.)  It may be a few days until we make the next PR, as we've noticed some cleanups we need to do.

> Regarding nomenclature, I could not find anything about this particular concept, but I feel uneasy with the name regular since it already means something specific here. What about something like "epi-stable category"? Otherwise, it's not a hugely big deal and I can live with it.

EpiStable.v, "epi-stable" (in prose) and `IsEpiStable` might be good choices.  And `IsAbEpiStable` for the version enriched over abelian groups.  @ThomatoTomato , what do you think?

jdchristensen:
@Alizter The renaming is done.  Let me know if this looks good.

PR #2338, opened 2026-01-10
ci: bump minimal rocq version and improve ci

Alizter:
We update the CI to use a minimum Rocq version of 9.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -462,7 +462,7 @@ jobs:
           echo "::remove-matcher owner=coq-problem-matcher::" # remove problem matcher installed by Coq docker action, so we don't get duplicate warning annotations
           ## Test install target
           make install
-          echo 'Require Import HoTT.HoTT.' | rocq -q
+          echo 'Require Import HoTT.HoTT.' | rocq repl -q
 
     - name: Revert permissions
       # to avoid a warning at cleanup time - https://github.com/coq-community/docker-coq-action#permissions


.github/workflows/ci.yml
```diff
@@ -6,9 +6,12 @@ concurrency:
...
   cancel-in-progress: true
 
-# We set the supported coq-version from here. In order to use this environment variable correctly, look at how they are used in the following jobs.
+# Coq version configuration
+# - coq-version-supported: The main supported version used for documentation
+# - Other versions are tested but docs are only built with supported version
 env:
-  coq-version-supported: '8.19'
+  coq-version-supported: '9.0'
+  coq-version-9-1: '9.1'
```
jdchristensen, line 14: I'm not sure it matter much, but I think it would be better if the docs were generated using the latest released version (9.1 in this case).

.github/workflows/ci.yml
```diff
@@ -34,28 +37,21 @@ jobs:
...
         os:
           - ubuntu-latest
-    env:
-      coq-version: ${{ matrix.coq-version-dummy }}
     runs-on: ${{ matrix.os }}
     steps:
-    # Github actions doesn't let us set workflow level environment variables inside the strategy of a job. Therefore we use the dummy variable coq-version in the matrix to set an environment variable env.coq-version, which uses the globally set coq-version-supported when running the 'supported' case.
-    - name: Set supported coq-version
-      if: matrix.coq-version-dummy == 'supported'
-      run: echo "coq-version=${{ env.coq-version-supported }}" >> $GITHUB_ENV
```
jdchristensen, line 50: I see you were able to get rid of some hacks.  Did something change to make this possible, or did you just figure out  a better way?

.github/workflows/ci.yml
Alizter: I've created #2343

.github/workflows/ci.yml
Alizter: This is essentially removing the placeholder hack and hardcoding the versions everywhere. Although this duplicates work, it is much easier to follow and understand. I have some WIP reorganising the entire CI to be much more declarative and easy to understand but I didn't want to mix concerns here.

PR #2339, opened 2026-01-11
chore: bump minimal rocq version

Alizter:
We bump the minimal rocq version needed in the dune-project file.


Alizter:
@jdchristensen I've forgotten about this. The .opam file is generated from the tempalte file. It should regenerate when you do `dune build coq-hott.opam`.

Alizter:
Sure that's fine if you edit it by hand for now, you can get rid of the template and also the `(generate_opam_files)` in the `dune-project` file. We can edit it by hand from now on.

PR #2340, opened 2026-01-11
fix and update alectryon

Alizter:
I broke this job in an earlier PR on purpose. We fix the CI job for it here and change the backend that alectryon was using. It used to use serapi, but that has now been subsumed by rocq-lsp. We update the submodule alectryon to a version that understands it.

### DIFF
--- a/coq-hott.opam
+++ b/coq-hott.opam
@@ -31,6 +31,6 @@ build: [
 ]
 dev-repo: "git+https://github.com/HoTT/HoTT.git"
 depends: [
-  "dune" {>= "3.13"}
+  "dune" {>= "3.14"}
   ("rocq-core" {>= "9.0"} & "coq-core" {>= "9.0"})
 ]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,51 +163,26 @@ jobs:
 # - coqchk: Runs coqchk
 # - install: Tests install target
 
-   # alectryon job
+  # alectryon doc job (using Nix + dune for reproducible builds)
   doc-alectryon:
-    needs: build
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
       with:
         submodules: recursive
-    # Download artifact
-    - uses: actions/download-artifact@v4
+    - uses: nixbuild/nix-quick-install-action@v30
+    - uses: nix-community/cache-nix-action@v6
       with:
-        name: workspace-${{ env.coq-version-supported }}
-    # Unpack Tar
-    - run: tar -xf workspace.tar
+        primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.lock') }}
+        restore-prefixes-first-match: nix-${{ runner.os }}-
     - name: add problem matchers
       run: |
-        #echo "::add-matcher::etc/coq-scripts/github/coq-oneline-error.json" # now via a script
-        #echo "::add-matcher::etc/coq-scripts/github/coqdoc.json" # disabled for now, since they don't have file names
         echo "::add-matcher::etc/coq-scripts/github/alectryon-error.json"
-        #echo "::add-matcher::etc/coq-scripts/github/alectryon-warning.json" # too noisy right now, cf https://github.com/cpitclaudel/alectryon/issues/34 and https://github.com/cpitclaudel/alectryon/issues/33
-    - uses: coq-community/docker-coq-action@v1.5.1
-      with:
-        coq_version: ${{ env.coq-version-supported }}
-        ocaml_version: ${{ env.ocaml-version }}
-        custom_script: |
-          opam install -y coq-serapi
-          sudo apt-get -o Acquire::Retries=30 update -q
-          sudo apt-get -o Acquire::Retries=30 install python3-pip python3-venv autoconf -y --allow-unauthenticated
-          startGroup "Workaround permission issue" # https://github.com/coq-community/docker-coq-action#permissions
-            sudo chown -R 1000:1000 .
-          endGroup
-          # Create and activate a virtual environment
-          python3 -m venv myenv
-          source myenv/bin/activate
-          # Install the required Python packages in the virtual environment
-          python -m pip install --upgrade pip
-          python -m pip install pygments dominate beautifulsoup4 docutils==0.17.1
-          echo "::remove-matcher owner=coq-problem-matcher::" # remove problem matcher installed by Coq docker action, so we don't get duplicate warning annotations
-          make alectryon ALECTRYON_EXTRAFLAGS=--traceback
-    - name: Revert permissions
-      # to avoid a warning at cleanup time - https://github.com/coq-community/docker-coq-action#permissions
-      if: ${{ always() }}
-      run: sudo chown -R 1001:116 .
+    - name: Build alectryon docs with dune
+      run: |
+        nix develop .#coq_9_1 -c dune build @alectryon
     - name: tar alectryon artifact
-      run: tar -cf alectryon-html.tar alectryon-html
+      run: tar -cf alectryon-html.tar -C _build/default alectryon-html
     - name: upload alectryon artifact
       uses: actions/upload-artifact@v4
       with:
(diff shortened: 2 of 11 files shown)


jdchristensen:
Looks like there is still a build issue.  BTW, do you know if alectryon is using more than one core when building?  It would be great to speed it up if possible.

Alizter:
Alectryon was starting coq-lsp and querying it for every goal one at a time. That is slow. I've instead used coq-lsp's Coq compiler `fcc` to dump the goals of each .v file. After that I have a script that can pass this to alectryon in a format it will understand. I've written a generator for dune rules for this, so as long as you have coq-lsp installed in your opam switch or if you use the nix flake, you can build `dune build alectryon-html/` incrementally.

jdchristensen:
Wow, that sounds great!  It's much faster than before.  Does it also use multiple cores?

Alizter:
I did a self-review. There are some issues:

1. Fix JSON parser to handle strings properly (use json.loads() or streaming parser)
2. Add error handling and validation to Python converter
3. Fix exit code handling
4. Remove unused directory-targets
5. Restore problem matcher

I think we should hold off merging for now until I can address this points.

jdchristensen:
Thanks for pushing this forward.  How should I review this?  Have you checked the html output?  Can I check it?  Did you address the five points you listed earlier?

Alizter:
Sorry, I should have been clearer earlier. The 5 points from the earlier self-review were all addressed. The HTML files look to be correct.

Alizter:
To try it yourself you can initialise the alectryon submodule and afterwards run `dune build @alectryon`. The html files will appear in `_build/default/alectryon-html/`.

jdchristensen:
I had trouble checking out the submodule.  Here's what Claude said:

The `.gitmodules` file points to the upstream `cpitclaudel/alectryon`, but the pinned commit lives in @Alizter's fork and is not reachable from any branch tip there — it must be fetched by exact hash.

```sh
# From the repo root:
git submodule update --init etc/alectryon

cd etc/alectryon
git remote add alizter https://github.com/alizter/alectryon.git
git fetch alizter 8ff5afaa4f747400dd1b41b59782dcba400bb544
git checkout 8ff5afaa4f747400dd1b41b59782dcba400bb544
```

jdchristensen:
Then I noticed various errors scroll by in the build, e.g.

[Deleted.  I just needed to install the python3-dominate package.]

Alizter:
I'll fix the submodule as soon as I get another opportunity. In the meantime are you able to fix it?

jdchristensen:
I'll let you fix the submodule as I don't really know how that works.

As for testing the generated html, when I open the pages, I just get static html, without the interactive features.  I guess I have to serve them through a web server?  Is this how you tested them?

jdchristensen:
To clarify, I did get the submodule working by using the instructions Claude gave.  But I don't know how to "fix" it in the sense of making it work without special instructions.  So I have successfully built the html, but not yet tested it properly.

Alizter:
Two issues:
1. The css is missing, which is strange because I thought it was being copied before
2. The webpage is no longer interactive. This is a bit more serious and I'll have to look into it.

Alizter:
I stupidly tweaked the json formatting code from earlier and that's probably the issue with the python script.

Alizter:
Actually the issue for far simpler. The `--root` argument I was giving `fcc` (the alterantive Coq compiler shipped with coq-lsp) was getting confused with the relative paths and instead needed them to be rooted from the project root.

Alizter:
In my own testing, it appears to work now. The other issue earlier with the gitsubmodule was because I didn't specify the branch from upstream alectryon. I've fixed these issues now and it should be working.

jdchristensen:
Now it seems to be working for me as well.  Is there a table of contents page too?

Alizter:
Added one now:
```
google-chrome _build/default/alectryon-html/index.html
```

jdchristensen:
The index.html helps.  Is this what ours used to look like for alectryon:  https://hott.github.io/Coq-HoTT/alectryon-html/toc.html ?  That has the advantage that it shows the headings, but it's too bright for my taste, so maybe the simpler form is better?  Whatever you prefer is fine, and feel free to merge when you've decided.

BTW, I did a fresh checkout, and am still getting the issue that the commit referenced in the submodule is not found.  But this may just be a quirk of the way I have it checked out.

Alizter:
@jdchristensen That's not ideal. I'll try to see what the issue with the submodule is today.

The index can be improved, I'll take a stab at that too.

Alizter:
@jdchristensen One thing I am confused about: How is the CI able to use the submodule if it doesn't have the hash?

Alizter:
The gitsubmodule should be pointing at alectryon main, at some point it was pointing at an experimental branch but I got rid of this because I don't want to maintain that. The new goaldump workflow should work with standard alectryon.

jdchristensen:
The submodule seems to be fine now.  I'm not sure how the CI managed, but I guess it had access to the commit that was specified.  I wouldn't spend too much time on the index.html file.  I think yours is better looking than what we had.  (But should it be renamed to toc.html, so the wiki link still works?)

Alizter:
@jdchristensen I've reverted to the index file we had before for now. We can however improve it if we know what we would like in the end. I don't think this is the place to do it since we just want to reach feature-parity.

Alizter:
@jdchristensen I've done a complete rewrite of the rule generation to make it easier to read and maintain. You don't have to know exactly how it works, claude will explain it if you want. The generated rules are immediately read by dune anyway and you can incrementally build the alectryon web pages that you wish. The submodule is not pointing at alectryon main.

jdchristensen:
I don't really understand how this all works, but it built fine for me.

The CI gave an error, though.

I also wonder about needing so much custom code.  Shouldn't all of this code be part of alectryon?  Wouldn't it be useful for any project building documentation with alectryon?  This shouldn't prevent merging this, but it's at least worth pointing out to upstream what your approach is to see if (some of) it might get incorporated.

Alizter:
@jdchristensen I've fixed the CI. It was something silly not serious.

The question about custom code is a good one. Rocq-lsp as a project has been discontinued as of last year, but it still receives intermittent maintenance. It provides an alternative compiler to rocq called `fcc` (Fleche Coq Compiler). The feature we are using is its "goal dump plugin" which lets you dump all the goals of each statement of a given file. The python script then rewires that into something alectryon can understand. cc @cpitclaudel in case he is interested.

There was a rocq-lsp branch for alectryon at some point but that was calling rocq-lsp directly, so alectryon was sending lsp requests to rocq-lsp. This was extremely slow so I quickly abandoned this approach. Getting fcc to dump its goals is much faster and presumably in the future rocq could do the same.

The OCaml code is just generating Dune rules so we don't have to write a file by hand. The rules can be observed in `_build/default/alectryon_rules.sexp` once you've built it. It's not particularly interesting but it does allow us to incrementally build the alectryon docs.

jdchristensen:
As I mentioned, I think you can merge when you are ready.

cpitclaudel:
>  It provides an alternative compiler to rocq called fcc (Fleche Coq Compiler). The feature we are using is its "goal dump plugin" which lets you dump all the goals of each statement of a given file. The python script then rewires that into something alectryon can understand. cc @cpitclaudel in case he is interested.

I am interested :) Could we merge that rewiring code directly into Alectryon?  Then it could do the full sequence (call fcc, parse output, render) all on its own.

coq-hott.opam
```diff
@@ -31,6 +31,6 @@ build: [
 ]
 dev-repo: "git+https://github.com/HoTT/HoTT.git"
 depends: [
-  "dune" {>= "3.13"}
+  "dune" {>= "3.14"}
```
jdchristensen, line 34: @Alizter This automatically generated file was changed, but the source for it (coq-hott.opam.template) was left unchanged.  So now this file keeps getting rebuild to say 3.13, and then git shows it as a change in the working tree.  Should we revert this to 3.13, or change coq-hott.opam.template to 3.14?  dune-project says 3.14 as well.

(This is a comment on a PR that was merged back in April.)

Alizter: Yes, I don't mind bumping dune to say 3.20.

jdchristensen: I'm at 3.21.1, so that's fine with me, but if we don't use any newer features, maybe it's best to be generous?  Most important is that the two (or three?) files are in sync so we don't get the spurious changed file in git.

PR #2341, opened 2026-01-11
fix dune-project version

Alizter:


Alizter:
Looks like this will have to wait until we update dune to 3.21 so that we can get rid of the coq compat shims.

Issue #2313, opened 2025-09-17
Name conflict between theories/* and test/* in HoTT namespace


jdchristensen:
`From HoTT Require Import Group.` gives an error in recent Rocq versions because both HoTT and HoTT.Tests contain a file Group.v.  We should separate these namespaces.

jdchristensen:
This is fixed by #2342 , so closing it.

Issue #2285, opened 2025-06-06
Renaming HoTT.Tests and maybe HoTT.Contrib


jdchristensen:
Currently we often use the same path names in the `test` folder as in the `theories` folder, which means that `From HoTT Require Import Foo.Bar.` is ambiguous.  As long as `Foo.Bar` is the full name, Coq prefers to get this from `HoTT` (i.e. `theories`) than from `HoTT.Tests` (i.e. `test`), but this is not clear to the reader.

So maybe it would be a good idea to rename `HoTT.Tests` to something like `HoTT_Tests`?

(We could consider something similar for `contrib`, but I don't think we have conflicts there.)

See #2283 for related discussion.

jdchristensen:
Here are some examples:
```
test/Algebra/Groups/Group.v   theories/Algebra/Groups/Group.v
test/Pointed/Core.v           theories/Pointed/Core.v
test/WildCat/Opposite.v       theories/WildCat/Opposite.v
```

jdchristensen:
This is fixed by #2342 .

PR #2342, opened 2026-01-11
chore: HoTT.Tests -> HoTT_Tests, HoTT.Contrib -> HoTT_Contrib

Alizter:
This fixes issues #2313 and #2285 by avoiding ambiguous module paths like HoTT.Contrib.X which could conflict with HoTT.X.

### DIFF
--- a/Makefile.coq.local
+++ b/Makefile.coq.local
@@ -57,7 +57,7 @@ CORE_DPDFILES=$(patsubst theories.%,file-dep-graphs/HoTT.%,$(subst /,.,$(CORE_VF
 # The list of files from contrib
 CONTRIB_VOFILES=$(CONTRIB_VFILES:.v=.$(VO))
 CONTRIB_GLOBFILES=$(CONTRIB_VFILES:.v=.glob)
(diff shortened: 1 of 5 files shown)


jdchristensen (commented):
This definitely makes sense for HoTT.tests, since we use the same paths there.  I'm not sure about HoTT.Contrib, but it's fine with me to change that.

BTW, I've always found it odd that the folder is called `test` but the module is called `Tests`. Should we rename the folder to `tests`?  (There's also the issue of the leading capital in the names Tests and Contrib; should the folders be capitalized as well?)

Alizter:
I don't mind the name change if that's what you think is better. I personally prefer shorter names since it means less typing. I can take a better look during the weekend.

Alizter:
Since we install the theory Contrib, its not a good idea to have such a generic name. We could fix this by making it a private theory, i.e. one that is not installed. To make a theory private, we just have to remove the `(package)` field of the `rocq.theory` stanza.

jdchristensen:
If you rebase, I still think this is good to merge.

PR #2344, opened 2026-01-14
Homological Algebra: Diagram chasing in practice

ThomatoTomato:
This PR continues the work of #2336, #2335, #2334, and #2333, and is joint work with @jdchristensen.

We are finally able to start reaping the fruits of our labor. In this PR we prove many lemmas using the diagram chasing tactic developed so far. In EpiStable.v we prove some cancellation properties of monics and epics in epi-stable categories. HomologicalAlgebra.v contains a lot of diagram lemmas one can prove in ab-enriched epi-stable categories, among these we find the familiar 4-lemmas and 3⨉3-lemmas.

This PR does not contain any diagram lemmas that would require us to create a map, as we don't have a tactic to achieve this. Therefore no kernel-cokernel sequence or snake lemma is present.

### DIFF
--- a/theories/WildCat/EpiStable.v
+++ b/theories/WildCat/EpiStable.v
@@ -302,10 +302,8 @@ Section EpiStableLemmas.
     rhs_V' napply lx'.
     napply cat_postwhisker.
     apply (ismonic (g $o f)).
-
     lhs' napply cat_assoc.
     rhs' napply cat_assoc.
-
     lhs' napply (g $@L lx).
     rhs' napply (g $@L lx').
     exact p.
--- a/theories/WildCat/HomologicalAlgebra.v
+++ b/theories/WildCat/HomologicalAlgebra.v
@@ -53,8 +53,7 @@ Section FourLemma.
       rewrite ld.
       rewrite cat_assoc_opp.
       rewrite complex_cde'.
-      apply precomp_zero.
-    }
+      apply precomp_zero. }
 
     (* The element [c] is not necessarily the correct lift of [c'].  We correct this by lifting the element [c' - h $o c] using exactness of the bottom row. This yields an element [b' : elt _ B']. *)
     elt_lift_exact bc' cd' (c' - h $o c) b' lb' using clear sq3 lc ld d.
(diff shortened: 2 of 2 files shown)


Alizter:
I plan to take a look at this next weekend.

jdchristensen:
> You should add HomologicalAlgebra.v to Wildcat.v, after EpiStable.v.

I meant to tag @ThomatoTomato

theories/WildCat/EpiStable.v
```diff
@@ -273,3 +272,55 @@ Ltac rewrite_with_assoc h :=
...
+    apply monic_via_elt; intros P y y' p.
+
+    elt_lift_epic f y x lx; elt_lift_epic f y' x' lx'.
+
+    lhs_V' napply lx.
+    rhs_V' napply lx'.
+    napply cat_postwhisker.
+    apply (ismonic (g $o f)).
+
+    lhs' napply cat_assoc.
```
Alizter, line 305: @jdchristensen I should have been more clear, sorry. I meant these kinds of blank lines. If you think it helps reading the proof then I am fine with leaving it.

theories/WildCat/HomologicalAlgebra.v
```diff
@@ -0,0 +1,840 @@
...
+    (* We lift [b'] to [b : elt _ B]. *)
+    elt_lift_epic g b' b lb.
+
+    (* Since [h] is monic, it follows that [b] is a lift of [c]. *)
+    assert (p' : bc $o b $== c).
+    { apply (ismonic h).
+      rewrite_with_assoc sq2.
+      rewrite lb.
+      exact lb'.
+    }
```
jdchristensen: @ThomatoTomato Can you move the `}` to the end of the previous line (preceded by a single space character)?  And do the same other places it happens?  (`baby_dragon_aux_left` is an example where it is done how I like it.)

theories/WildCat/EpiStable.v
jdchristensen: I agree that the last two empty lines in this proof should be removed.  @ThomatoTomato, can you do that?  I do like the first two blank lines here and in many of the proofs in HomologicalAlgebra.v.  Because we are doing forward reasoning, there aren't as many bullet points or other things to structure the proofs, so I think it helps to break them up into logical pieces visually.

PR #1628, opened 2022-02-06
Revert to Ltac1 tactic in Syllepsis; add Syllepsis+EncodeDecode to HoTT.v

jdchristensen:

PR #2352, opened 2026-04-08
Syllepsis: use Ltac2 for slight speed up; reverts #1628

jdchristensen:
It was easy to go back to the Ltac2 version, since I just had to revert #1628.  (I also made a couple of minor style fixes, and improved the comment slightly.)

On a slow-ish machine, this reduces the build time for Syllepsis.v from 5.25s to 4.55s.  Not a huge gain, but not nothing.


Alizter (approved):
Cool :)

PR #2353, opened 2026-04-23
Adapt to https://github.com/rocq-prover/rocq/pull/21947

proux01:
Adapt to https://github.com/rocq-prover/rocq/pull/21947

Should be backward compatible


Alizter (approved):

Alizter:
@proux01 Thanks!

PR #2356, opened 2026-04-24
chore: put asymmetric patterns commands on separate lines

Alizter:
Cleanup what was started in #2355 and remove redundant settings from Categories/.

### DIFF
--- a/theories/Basics/Settings.v
+++ b/theories/Basics/Settings.v
@@ -43,7 +43,7 @@ Global Set Printing Primitive Projection Parameters.
 
 (** This flag removes parameters from constructors in patterns that appear in a match statement. *)
 Global Set Asymmetric Patterns.
-(** The warning clause here and in other parts of the library can be removed once our minimum Rocq version is 9.3. *)
+(** The warning clause here can be removed once our minimum Rocq version is 9.3. *)
 #[warning="-unknown-option"]
 Global Set Asymmetric Patterns No Implicits.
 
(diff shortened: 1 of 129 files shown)


theories/Basics/Settings.v
```diff
@@ -42,7 +42,10 @@ Global Set Printing Primitive Projection Parameters.
 (** ** Pattern Matching *)
 
 (** This flag removes parameters from constructors in patterns that appear in a match statement. *)
-Global Set Asymmetric Patterns. #[warning="-unknown-option"] Global Set Asymmetric Patterns No Implicits.
+Global Set Asymmetric Patterns.
+(** The warning clause here and in other parts of the library can be removed once our minimum Rocq version is 9.3. *)
```
jdchristensen:
```suggestion
(** The warning clause here can be removed once our minimum Rocq version is 9.3. *)
```

jdchristensen (approved):
I quickly reviewed it and it looks better.  I noticed a few places where no blank lines were left between the last settings line and the next line.  I fixed two in the github GUI, but then found more and stopped trying.  (Github wouldn't let me create suggested edits for these.)  So this looks fine, or the blank lines in a few more places could be restored if you feel like it.

My edit to the first comment assumes that this is now the only place with this warning clause, but I didn't check that.

Alizter:
Yes, this should be the only place now. The others turned out to be unnecessary like you said.

PR #2358, opened 2026-04-24
build: migrate Dune integration to Rocq

Alizter:
Migrate from Dune's deprecated Coq integration to `(using rocq 0.11)` and `rocq.*` stanzas.

3.21 is the version of dune pinned ( :cry: ) in the docker images, so it's worth keeping as the lower bound for now. I could unpin it but that messes with the opam CI too much.

### DIFF
--- a/coq-hott.opam
+++ b/coq-hott.opam
@@ -30,6 +30,7 @@ build: [
   ]
 ]
 dev-repo: "git+https://github.com/HoTT/HoTT.git"
+x-maintenance-intent: ["(latest)"]
 depends: [
   "dune" {>= "3.13"}
   ("rocq-core" {>= "9.0"} & "coq-core" {>= "9.0"})
--- a/contrib/dune
+++ b/contrib/dune
@@ -1,4 +1,4 @@
-(coq.theory
+(rocq.theory
  (name HoTT.Contrib)
  (package coq-hott)
  (theories HoTT))
--- a/dune
+++ b/dune
@@ -26,7 +26,8 @@
   (glob_files_rec ./*.vo))
  (action
   (run
-   coqchk
+   rocq
+   check
    -R
    ./theories
    HoTT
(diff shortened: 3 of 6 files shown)


coq-hott.opam
```diff
@@ -15,6 +15,11 @@ authors: ["The HoTT Library Development Team"]
 license: "BSD-2-Clause"
 homepage: "http://homotopytypetheory.org/"
 bug-reports: "https://github.com/HoTT/HoTT/issues"
+depends: [
+  "dune" {>= "3.21"}
+  "rocq-core" {>= "9.0.0"}
+  "odoc" {with-doc}
```
Alizter, line 21: This gets automatically added but is harmless so I opt to leave it in.

Alizter:
@jdchristensen This is ready now, I've got a workaround in place that allows for compatibility between dune 3.21 and Rocq 9.0. I'll only proceed when you give the go ahead so I don't cause anguish in your opam switches. (Assuming you need to bump dune)

etc/rocq-compat/rocq
```diff
@@ -0,0 +1,27 @@
+#!/bin/sh
+
+# Dune 3.21 and 3.22 call [rocq --config], but Rocq 9.0 only supports
+# [rocq c --config].  Find the next [rocq] on PATH to avoid calling this
+# wrapper recursively.
```
jdchristensen, line 5: Maybe this is a reason to require a minimum of 9.1.0?  That would be fine with me.  We've usually tried to support roughly two versions, so that would be 9.1 and 9.2 at this point.  Then the whole wrapper could be avoided.

jdchristensen:
I don't understand the implications of this change, but it builds fine on my system with Rocq 9.1.0 and dune 3.21.1.  I believe dune 3.21.0 was released in January, around 8 months ago, so I think it's ok to require it as a minimum.

etc/rocq-compat/rocq
Alizter: I've created a PR doing this #2393.

Alizter:
Newer versions of dune have completely dropped the old `coq.*` stanzas in accordance with the renaming of rocq. There have been a few bug fixes with the rocq support since as well.

This should let us use newer versions of dune with better rocq support with less hassle.

jdchristensen (approved):
I tested building with dune and make on one of my systems, and everything looked fine, using Rocq 9.1.0 and dune 3.21.1.

PR #2362, opened 2026-04-29
Updated BiInv.v and related files

maxipetro:
We update BiInv.v and all related files to make the types BiInv and IsBiInv follow the same pattern as Equiv and IsEquiv, the former being a record and the latter being a typeclass. This is an improvement of the current version where BiInv is a product type which should be according to the naming conventions of the library in fact be called 'IsBiInv'. We also added some useful lemmas to the file that can be used in future projects within the library in which bi-invertible maps are used.

This is jww @jdchristensen.

### DIFF
--- a/theories/Basics/Equivalences.v
+++ b/theories/Basics/Equivalences.v
@@ -310,6 +310,14 @@ Definition equiv_inj `(f : A -> B) `{IsEquiv A B f} {x y : A}
   : (f x = f y) -> (x = y)
   := (ap f)^-1.
 
+Definition equiv_inj_comp `(f : A -> B) `{IsEquiv A B f} {x y : A}
+  (p : f x = f y)
+  : ap f (equiv_inj f p) = p.
+Proof.
+  unfold equiv_inj.
+  apply eisretr.
+Defined.
+
 (** Assuming function extensionality, composing with an equivalence is itself an equivalence *)
 
 Instance isequiv_precompose `{Funext} {A B C : Type}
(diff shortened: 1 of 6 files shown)


theories/Basics/Equivalences.v
```diff
@@ -310,6 +310,14 @@ Definition equiv_inj `(f : A -> B) `{IsEquiv A B f} {x y : A}
   : (f x = f y) -> (x = y)
   := (ap f)^-1.
 
+Definition equiv_inj_comp `(f : A -> B) `{IsEquiv A B f} {x y : A}
```
Alizter: Maybe `ap_equiv_inj` is a better name? The current name wasn't obvious to me what it was.

jdchristensen:
I've reviewed this already in a PR to my branch, so I won't review it again here.

theories/Equiv/BiInv.v
```diff
@@ -7,47 +7,205 @@ Generalizable Variables A B f.
...
+  := Build_BiInv A B e (isbiinv_isequiv e (equiv_isequiv e)).
+
+Definition equiv_biinv_equiv `{Funext} A B
+  : BiInv A B <~> (A <~> B).
+Proof.
+  nrefine ((issig_equiv A B) oE _ oE (issig_biinv A B)^-1).
+  napply (equiv_functor_sigma_id equiv_isbiinv_isequiv).
+Defined.
+
+Definition idmap_biinv (A : Type) : BiInv A A.
```
Alizter: `biinv_idmap` is probably more inline with the rest.

Alizter (approved):

jdchristensen:
@maxipetro If you make the two changes as a second commit, I'll then merge.  Thanks for this!

maxipetro:
@jdchristensen Done. Thanks!

Alizter:
Thanks!

PR #2363, opened 2026-05-04
Adapt to rocq-prover/rocq#21987 (destruct clears non section variable even when reusing a section variable's name)

SkySkimmer:


jdchristensen:
This is a file that is slow to type check.  Can you see if these changes affect the build time for this file in a non-trivial way?

Also curious what changed in Rocq to make these changes necessary?  Does it now fail to use `clear` on a Section variable?

Alizter:
@SkySkimmer I see that you ran the bench in that PR but it failed on `coq-hott`, can you rerun with the overlay?

SkySkimmer:
ran a bench, seems fine
[coq-hott.BOTH.perfile_timings.1.log](https://github.com/user-attachments/files/28011903/coq-hott.BOTH.perfile_timings.1.log)

theories/Homotopy/Syllepsis.v
```diff
@@ -403,7 +403,7 @@ Section eh_p_pp.
     revert z1 urnat_z1.
     snapply equiv_path_ind_rlucancel.
     destruct wry0, wry1, wrz0, wrz1.
-    clear wry0 wry1 wrz0 wrz1.
+    clear wry0 wry1.
```
jdchristensen, line 406: I'm curious why it's ok to clear two of these but not the other two.  All of them are Context variables.  I don't understand what the Rocq change is and why it affects these variables.

jdchristensen:
@SkySkimmer Thanks for the timing check, that looks fine.

I'd still like to understand what is going on before merging, if you have a minute to explain.

(As an aside, I can't see any reason for the use of Context in this file in most or all places, since each Section generally has a single result, so all of those Context variables could become hypotheses of the single result, and then destruct would automatically clear them.  I may make a future PR to clean this up.)

jdchristensen:
(Ah, already merged, which is fine, but a comment would still be helpful.)

theories/Homotopy/Syllepsis.v
SkySkimmer: IIRC some of them have been modified by previous destruct or `apply in` or similar, so they are not the actual section variable but only share a name with it.
With rocq-prover/rocq#21987 this is now detected and the destruct on the previous line cleared them (destruct does not clear section variables)

Alizter:
@jdchristensen Sorry about that, I was a bit quick. Would you like me to revert?

theories/Homotopy/Syllepsis.v
SkySkimmer: typically
~~~coq
Section S.
  Variable x : unit.

  Goal False.
    apply id in x.
    (* proof term is now "let x := id x in ?Goal@{x:=x}" so the x in the context is not the section variable x *)
    destruct x.
~~~
before rocq-prover/rocq#21987 x is not cleared by destruct, after it is cleared (and so explicit `clear x` would fail)

jdchristensen:
@SkySkimmer Thanks!  That's very clear now.

@Alizter No worries, there was no need to avoid merging, I just wanted to understand.

PR #2367, opened 2026-05-15
Minor cleanups to homotopy groups, etc

jdchristensen:
The first commit is very minor stuff.  The second simplifies one proof slightly.


theories/Homotopy/EMSpace.v
```diff
@@ -130,7 +130,7 @@ Section EilenbergMacLane.
       K( (Pi n.+2 X) n.+2)
    <~>* K( (Pi n.+1 (loops X)), n.+2)
    = pTr n.+2 (psusp K( (Pi n.+1 (loops X)), n.+1))  [by definition]
-   <~>* pTr n.+2 (psusp (loops X))
+   <~>* pTr n.+2 (psusp (loops X))                   [by induction]
```
Alizter, line 133: It would be cool if we could write our proofs like this.

Alizter (approved):

PR #2374, opened 2026-07-02
Homotopy groups of types of functions into a classifying space

thchatzidiamantis:
Work on classifying spaces is split into ClassifyingSpace/Core and ClassifyingSpace/Maps.
- In ClassifyingSpace/Core, we introduce new eliminators and generalize the map that gives the H-space structure. We also change the content on the adjunction between `B` and `Pi 1` for better computational properties. 
- In ClassifyingSpace/Maps, we compute `Pi 0` and `Pi 1` of types of the form `X -> B G`, where `X` is a pointed connected type.

We also add necessary background work on group theory (centralizers, images, conjugate homomorphisms) and lemmas/tactics for paths.

This is joint work with @jdchristensen.


jdchristensen:
I co-authored parts of this and already reviewed it internally, so I'll ask @Alizter to review it for Coq-HoTT.

theories/Homotopy/ClassifyingSpace.v
```diff
@@ -1,607 +1,2 @@
...
-  : Is1Natural (opyon (Pi1 X)) (opyon X o B)
-      (equiv_bg_pi1_adjoint X).
-Proof.
-  rapply (is1natural_natequiv (natequiv_bg_pi1_adjoint X)).
-  (** Why so slow? Fixed by making this opaque. *)
-  Opaque equiv_bg_pi1_adjoint.
-Defined.
-Transparent equiv_bg_pi1_adjoint.
+Require Export ClassifyingSpace.Core.
+Require Export ClassifyingSpace.Maps.
```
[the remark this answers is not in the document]
jdchristensen: I don't think this comment was correct (see the metafile Join.v), but there is no harm in being more precise.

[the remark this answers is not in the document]
thchatzidiamantis: @jdchristensen I just noticed the same thing going through the library, does this still export properly or should I revert the commit?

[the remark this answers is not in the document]
jdchristensen: Both ways are fine.  The way with the `Homotopy` prefix is slightly more stable, as if another file in another folder gets called `ClassifyingSpace.v`, a prefix would be needed.  But that's not likely to happen.

theories/Homotopy/ClassifyingSpace/Core.v
```diff
@@ -536,6 +535,34 @@ Proof.
...
+  (X : pType) `{IsConnected 0 X} (G : Group)
+  : (B (Pi 1 X) -> B G) <~> (X -> B G).
+Proof.
+  rapply natequiv_map_bg.
+Defined.
+
+Definition equiv_map_bg_pointed `{Univalence}
+  (X : pType) `{IsConnected 0 X} (G : Group) (f : Pi 1 X $-> G)
+  : equiv_map_bg X G (fmap B f) = equiv_bg_pi1_adjoint X G f
+  := @idpath (X -> B G) (equiv_bg_pi1_adjoint X G f).
```
Alizter: What's the purpose of this idpath?

theories/Types/Paths.v
```diff
@@ -98,6 +98,14 @@ Proof.
   exact ((concat_1p q)^ @ (concat_p1 (1 @ q))^).
 Defined.
 
+(** When the same function appears twice, there is a special computation rule for transporting [idpath]. *)
+Definition transport_paths_FlFr_1 {A B : Type} {f : A -> B} {x1 x2 : A}
```
Alizter: This is slightly more specific than just having `p = 1` as you are also forcing `f = g` from above. If you generalise to `f` and `g` then you should just get `(ap f p)^ @ ap g p`, which then for identical `f` and `g` should `concat_Vp` to this.

@jdchristensen wdyt?

theories/Homotopy/ClassifyingSpace/Core.v
```diff
@@ -348,6 +450,10 @@ Proof.
     apply bloop_pp.
 Defined.
 
+Definition ap_fmap_b {G H : Group} (u : G $-> H) (g : G)
```
Alizter: `ap_fmap_b_beta_bloop`?

jdchristensen:
@Alizter Are you ok if I merge this?  I could ask @ThomatoTomato to give it a quick look-over if you are too busy.

Alizter (commented):
I had some comments pending which I forgot to submit

Alizter (approved):
This is good to go from my perspective, modulo the comments I've made.

theories/Homotopy/ClassifyingSpace/Core.v
thchatzidiamantis: I added this to highlight a difference from the previous proof of `equiv_bg_pi1_adjoint`, for which this equality was not definitional. It can be removed.

jdchristensen: Yeah, probably best to remove it.  If you think it's important for the reader, you could just add a comment that "[equiv_bg_pi1_adjoint X G f] is definitionally equal to [equiv_map_bg X G (fmap B f)]" near where the former is defined.

theories/Types/Paths.v
jdchristensen: If we keep it, we could consider adjusting the name.  OTOH, the `1` strongly implies that the functions are the same as well, as otherwise it would be unlikely to typecheck.  But maybe it's not worth keeping?  It was used twice in ClassifyingSpace/Core.v, but while looking there, I realized that one of the uses (and the nearby code) could be replaced with `reflexivity`, so now there is just one use.  (My first commit.)  And then I saw that the one use could be replaced with two lines, so I went ahead and did that and removed `transport_paths_FlFr_1`.

theories/Homotopy/ClassifyingSpace/Core.v
jdchristensen: That would be fine.  (If we parallel the naming of ClassifyingSpace_rec_beta_bloop, we wouldn't include the `ap_` in the name, but I think it's good to include it.)  I'll let @thchatzidiamantis make this and the other change, and then I'll merge.

Alizter:
Thank you all very much!

PR #2376, opened 2026-07-21
TIMING.md: rough documentation about various timing methods

jdchristensen:
I just put this in the top-level folder, since that's where INSTALL.md and STYLE.md are, and this timing info is of a similar nature.  But I'm open to other locations.  This is slightly cleaned up compared to what I posted in a comment elsewhere a few weeks ago.


TIMING.md
```diff
@@ -0,0 +1,117 @@
...
+  to see a nice summary of the user time for each file, sorted.  Saved in time-of-build.log and time-of-build-pretty.log.  Add `TIMING_REAL=1` to see real time instead.  Can also compare revisions, etc.  See
+
+    https://coq.inria.fr/refman/practical-tools/utilities.html#timing-targets-and-performance-testing
+
+- To find the slowest lines in the whole library:
+
+    ```
+    make [-jJ]
+    cd theories [or subdir to limit search]
+    find . -name '*.v.timing' -exec awk '{printf "%-55s %s\n", "{}:", $0}' {} \; | sort -nr -k7 | head -20 | sed 's/.v.timing//'
```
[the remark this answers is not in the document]
jdchristensen: Copilot is wrong here, at least with Gnu find, but it is correct that this is fragile, and moreover there is a faster way.  I'll push it.

Alizter:
What about a doc/ directory? The files other than README.md could also be moved there (in a followup PR).

doc/TIMING.md
```diff
@@ -0,0 +1,117 @@
...
+- It may help to be on A/C power, depending on the machine.
+
+Methods for timing overall builds.
+Unless said otherwise, all of the commands should be run from the top-level repo dir, above the theories dir.
+
+- To get overall build time, can use:
+
+    ```
+    dune clean; make clean > /dev/null 2>&1; sync; sleep 8
+    time nice -n -10 dune build --cache=disabled --display=quiet
```
Alizter, line 20: You can also inspect `dune trace cat | jq | less` and see some rich information with dune's internal events. `dune trace cat` will dump `_build/trace.csexp` into json, and you can use `jq` to rummage through the data. It contains important stuff like the actual invocation time for the rocq calls and the overall build time.

You can also use `dune trace cat --chrome-trace` and open it with perfetto to see a flamegraph of the build. Newer unreleased versions of dune also will now display parallelism metrics which can help understand if a build is particularly constrained.

They can be downloaded here if you are interested https://nightly.dune.build/

jdchristensen:
> What about a doc/ directory? The files other than README.md could also be moved there (in a followup PR).

Ok, I moved this to a new doc/ folder.  Should I just move the others as part of this PR for simplicity?

doc/TIMING.md
jdchristensen: Do you want to push an update to this PR mentioning dune trace?  (My dune doesn't have this command, so I can't test and can't tell how helpful it is.)  Alternatively, we could add this when it is more widely available?

Alizter:
@jdchristensen Sounds good, but cross-links may have to be updated.

jdchristensen:
There's also a link to STYLE.md on the wiki that I will update when this is merged.  Maybe we should skip the dune trace addition for now?

Alizter (approved):

jdchristensen:
> There's also a link to STYLE.md on the wiki that I will update when this is merged.

wiki updated.

Alizter:
`doc/` was used as a staging area for the deployment action so there was a conflict. I will adjust the deployment action to use a different name.

PR #2379, opened 2026-08-05
Add repository agent guidance

Alizter:
## Summary

- add concise repository-wide guidance for coding agents
- expose the same guidance to Claude and GitHub Copilot
- prepare Copilot's Nix development environment before agent sessions
- ignore local state created by common coding agents

## Motivation

Give coding agents one project-specific source of truth for Rocq development,
build validation, and repository conventions while keeping environment-specific
Copilot setup separate.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,13 @@
+# GitHub Copilot Instructions
+
+Follow [AGENTS.md](../AGENTS.md).
+
+## Environment Setup
+
+The Nix development shell is not activated automatically, so prefix project
+commands with `nix develop -c`:
+
+```sh
+nix develop -c dune build
+nix develop -c dune test
+```
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -1,0 +1,38 @@
+name: Copilot Setup Steps
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+  pull_request:
+    paths:
+      - .github/workflows/copilot-setup-steps.yml
+
+jobs:
+  # This job name is required by Copilot coding agent.
+  copilot-setup-steps:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+        with:
+          submodules: recursive
+
+      - name: Install Nix
+        uses: nixbuild/nix-quick-install-action@v35
+
+      - name: Restore Nix cache
+        uses: nix-community/cache-nix-action@v7
+        with:
+          primary-key: nix-${{ runner.os }}-${{ hashFiles('flake.lock') }}
+          restore-prefixes-first-match: nix-${{ runner.os }}-
+          gc-max-store-size-linux: 2G
+          save: false
+
+      - name: Enter Nix development shell
+        run: nix develop -c true
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,9 @@ nix/profiles/
 
 # Ignore the file to bench
 file_to_bench
+
+# Local AI assistant state
+/.agents/
+/.claude/
+/.codex/
+/.pi/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,67 @@
+# AGENTS.md
+
+Repository-wide guidance for coding agents. A more specific `AGENTS.md` in a
+subdirectory takes precedence for files below it.
+
+## Project
+
+HoTT is a Rocq (formerly Coq) library for homotopy type theory. Library sources
+live in `theories/`, additional developments in `contrib/`, tests in `test/`,
+and developer tooling in `etc/`.
+
+- Preserve the required `-noinit` and `-indices-matter` flags by using the
+  project build rather than invoking Rocq with ad-hoc arguments.
+- Treat the matrices in `.github/workflows/ci.yml` as the source of truth for
+  supported Rocq versions; avoid unnecessary version-specific features.
+
+## Proof Development
+
+- Read `STYLE.md` before changing `.v` files and match the surrounding style.
+  In particular, do not hard-wrap prose in Rocq comments.
+- Prove results at their natural level of generality: consider arbitrary
+  truncation levels, modalities, or reflective subuniverses when appropriate.
+- Prefer reusable lemmas over long proofs, and keep intermediate goals readable
+  enough for a human to follow. Useful project tactics include `napply`,
+  `rapply`, `snapply`, and `srapply`, plus the `lhs`, `lhs_V`, `rhs`, and
+  `rhs_V` tacticals (and primed variants) described in `STYLE.md`.
+- Preserve intended universe polymorphism while avoiding unnecessary universe
+  parameters and constraints. Inspect universe-sensitive definitions with
+  `About` or `Print` and `Set Printing Universes.`; add regression checks under
+  `test/` when appropriate.
+
+## Working Rules
+
+- Keep changes focused and avoid unrelated formatting or refactoring.
+- Do not introduce `Admitted`, `admit`, or new axioms to finish a proof. Follow
+  the axiom conventions in `STYLE.md` when axioms are genuinely in scope.
+- Do not edit generated files or build products such as `_CoqProject`,
+  `Makefile.coq`, `*.vo`, or `_build/`.
+- Wire new library files into the existing exports and build structure, usually
+  `theories/HoTT.v` and the relevant `dune` stanza.
+- Keep commits atomic and self-contained, with tests and implementation grouped
+  so that each commit is independently valid.
+- Do not stage, commit, or rewrite repository history unless explicitly asked.
+
+## Build and Test
+
+Prefer Dune. Use `dune build` for fast feedback while iterating:
+
+```sh
+dune build
+```
+
+Before each commit, run the complete validation target; it subsumes the build:
+
+```sh
+dune test
+```
+
+Use `make -j2` only when a task specifically concerns the Make build or needs
+to reproduce that CI path. Finish with `git diff --check` and report what was
+and was not validated.
+
+## Rocq MCP
+
+When a Rocq MCP server is available, use it for quick proof inspection and
+iteration. Treat it as optional fast feedback and validate final edits with the
+normal Dune build and tests.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -14,7 +14,7 @@ and developer tooling in `etc/`.
 - Note that `Basics` (especially `Basics.Overture`) provides many of the
   things usually provided by `Init` in standard Rocq developments.
 - Look up the supported Rocq versions in the CI matrices
-  (`.github/workflows/ci.yml`) rather than hard-coding them here; avoid
+  (`.github/workflows/ci.yml`); avoid
   unnecessary version-specific features.
 
 ## Proof Development
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -11,36 +11,51 @@ and developer tooling in `etc/`.
 
 - Preserve the required `-noinit` and `-indices-matter` flags by using the
   project build rather than invoking Rocq with ad-hoc arguments.
-- Treat the matrices in `.github/workflows/ci.yml` as the source of truth for
-  supported Rocq versions; avoid unnecessary version-specific features.
+- Note that `Basics` (especially `Basics.Overture`) provides many of the
+  things usually provided by `Init` in standard Rocq developments.
+- Look up the supported Rocq versions in the CI matrices
+  (`.github/workflows/ci.yml`) rather than hard-coding them here; avoid
+  unnecessary version-specific features.
 
 ## Proof Development
 
-- Read `STYLE.md` before changing `.v` files and match the surrounding style.
+- Read `doc/STYLE.md` before changing `.v` files and match the surrounding style.
   In particular, do not hard-wrap prose in Rocq comments.
 - Prove results at their natural level of generality: consider arbitrary
   truncation levels, modalities, or reflective subuniverses when appropriate.
 - Prefer reusable lemmas over long proofs, and keep intermediate goals readable
-  enough for a human to follow. Useful project tactics include `napply`,
-  `rapply`, `snapply`, and `srapply`, plus the `lhs`, `lhs_V`, `rhs`, and
-  `rhs_V` tacticals (and primed variants) described in `STYLE.md`.
+  enough for a human to follow.
+- Useful project tactics include `napply`, `snapply`, `rapply`, `srapply`,
+  `tapply`, `stapply`, and the `lhs`, `lhs_V`, `rhs`, and `rhs_V` tacticals
+  (and primed variants) described in `doc/STYLE.md`.
 - Preserve intended universe polymorphism while avoiding unnecessary universe
   parameters and constraints. Inspect universe-sensitive definitions with
-  `About` or `Print` and `Set Printing Universes.`; add regression checks under
-  `test/` when appropriate.
+  `About` or `Print` and `Set Printing Universes.`
+- Add regression checks under `test/` when appropriate.
+- Before stating a lemma, search `theories/` for an existing or more general
+  form; if a close variant exists, generalize it in place rather than adding
+  a new one.
+- Do not add definitions that merely rename, specialize, or compose existing
+  lemmas in one step; use the existing lemmas directly.
+- Make arguments implicit when inferable from later arguments (objects from
+  morphisms, groups from homomorphisms), and declare instances with
+  `Instance` at the definition rather than wrapping them later.
 
 ## Working Rules
 
 - Keep changes focused and avoid unrelated formatting or refactoring.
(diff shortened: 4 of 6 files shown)


Alizter:
@jdchristensen Can you take a look at this PR? I'm interested to know if it improve's claudes handling on your end. This PR should also make it possible to use with other coding agents and also make GitHub copilot able to interact with proofs.

Alizter:
Generally we should try to keep the knowledge in AGENTS.md as minimal and project specific as possible. Anything that we suspect the agents might already know (such as using dune) is not worth putting in.

Alizter:
One thing that I would like to try is getting agents to reduce universe level explosions around the library. In my own testing it appears to work pretty well, when you point them at a particular definition that you are unhappy with (i.e. has hundreds of levels) you can task it with minimizing as much as possible. There is obviously some implicit knowledge that we have with trying to keep as much polymorphism as possible when appropriate and keeping small things small, we ought to write these down somewhere.

CLAUDE.md
```diff
@@ -0,0 +1 @@
+AGENTS.md
```
jdchristensen:
```suggestion
@AGENTS.md
```
The @ will inline AGENTS.md into this file, ensuring it is read.

AGENTS.md
```diff
@@ -0,0 +1,50 @@
+# AGENTS.md
+
+Repository-wide guidance for coding agents. A more specific `AGENTS.md` in a
+subdirectory takes precedence for files below it.
+
+## Project
+
+HoTT is a Rocq (formerly Coq) library for homotopy type theory. Library sources
+live in `theories/`, tests in `test/`, and developer tooling in `etc/`.
```
jdchristensen: Should mention contrib as well.

AGENTS.md
```diff
@@ -0,0 +1,50 @@
...
+
+- Read `STYLE.md` before changing `.v` files and follow the surrounding style.
+- Preserve the required `-noinit` and `-indices-matter` flags by using the
+  project build rather than invoking Rocq with ad-hoc arguments.
+- CI supports Rocq 9.0, 9.1, 9.2, and development versions; avoid unnecessary
+  version-specific features.
```
jdchristensen: Can we point the AI to the appropriate file so it can look up the current versions we support?  Then we won't need to edit this file each time we update versions.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Environment Setup
+
+The Nix development shell is not activated automatically, so prefix project
+commands with `nix develop -c`:
+
+```sh
+nix develop -c dune build
+nix develop -c dune runtest
+```
```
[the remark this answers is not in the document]
jdchristensen: This Copilot feedback seems to be sensible.

jdchristensen (commented):
My CLAUDE.md contains the following, some of which might be good to include in the one being created here:
```
## Style

- Please follow the guidance from ./doc/STYLE.md, but also try to match the style in any given file.
  Note in particular that comments never contain line breaks, except to separate paragraphs.

- The `(s)rapply` and `(s)napply` family of tactics are quite useful.  Also `lhs`, `lhs_V`, `rhs` and `rhs_V` are handy for reasoning on one side of an equality, and the primed versions for reasoning on one side of any symmetric, transitive relation (such as `==*` or `<~>`).

- Try to prove results in as much generality as possible.  For example, instead of proving something about 0-connectedness, you could try proving it about n-connectedness or even about O-connectedness for any modality or reflective subuniverse O.

- Rather than producing long proofs, try to make general purpose lemmas that can handle intermediate steps.

- Make proofs easy for a human to step through and understand.  Try to make the goals look readable at each step.
```

CLAUDE.md
Alizter: This is a symlink but Git renders it like this.

jdchristensen: Oh, a symlink is fine too, although maybe less portable than the @ method.

Alizter: That's true, I'll change it to the syntax you describe.

Alizter:
@jdchristensen I've addressed your comments. Let me know how it works and feels on your end.

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+and developer tooling in `etc/`.
+
+- Preserve the required `-noinit` and `-indices-matter` flags by using the
+  project build rather than invoking Rocq with ad-hoc arguments.
+- Treat the matrices in `.github/workflows/ci.yml` as the source of truth for
+  supported Rocq versions; avoid unnecessary version-specific features.
+
+## Proof Development
+
+- Read `STYLE.md` before changing `.v` files and match the surrounding style.
```
jdchristensen: Replace STYLE.md with doc/STYLE.md in all places.

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Working Rules
+
+- Keep changes focused and avoid unrelated formatting or refactoring.
+- Do not introduce `Admitted`, `admit`, or new axioms to finish a proof. Follow
+  the axiom conventions in `STYLE.md` when axioms are genuinely in scope.
+- Do not edit generated files or build products such as `_CoqProject`,
+  `Makefile.coq`, `*.vo`, or `_build/`.
+- Wire new library files into the existing exports and build structure, usually
+  `theories/HoTT.v` and the relevant `dune` stanza.
```
jdchristensen: I don't think any dune stanza needs changing when a new file is added?  Maybe the whole line should be replaced with:
```
- Add new library files to the closest meta-file above them in the folder structure, such as `theories/WildCat.v` or `theories/HoTT.v`.
```

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+- Keep commits atomic and self-contained, with tests and implementation grouped
+  so that each commit is independently valid.
+- Do not stage, commit, or rewrite repository history unless explicitly asked.
```
jdchristensen, line 56: The last bullet somewhat conflicts with the previous one, which mentions how to commit.  Maybe reverse the order, and change the "Keep commits" one to "When asked to commit, keep commits ..."?

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+
+Before each commit, run the complete validation target; it subsumes the build:
+
+```sh
+dune test
+```
+
+Use `make -j2` only when a task specifically concerns the Make build or needs
+to reproduce that CI path. Finish with `git diff --check` and report what was
+and was not validated.
```
jdchristensen: Add a note that `make` will only build a file if `git add` has been run on it, and give permission to do so.

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+Prefer Dune. Use `dune build` for fast feedback while iterating:
+
+```sh
+dune build
+```
+
+Before each commit, run the complete validation target; it subsumes the build:
+
+```sh
+dune test
```
jdchristensen, line 73: Claude points out that `dune test` runs coqchk, and on my laptop that takes about 19s, so it's probably not needed for every commit.  Should we recommend running the tests with `dune build test/` before every commit, and running `dune test` (which additionally runs coqchk) when the whole task is complete?

AGENTS.md
jdchristensen: Ah, it looks like this PR might be based on a version before we created the doc/ folder?

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+## Proof Development
+
+- Read `STYLE.md` before changing `.v` files and match the surrounding style.
+  In particular, do not hard-wrap prose in Rocq comments.
+- Prove results at their natural level of generality: consider arbitrary
+  truncation levels, modalities, or reflective subuniverses when appropriate.
+- Prefer reusable lemmas over long proofs, and keep intermediate goals readable
+  enough for a human to follow. Useful project tactics include `napply`,
+  `rapply`, `snapply`, and `srapply`, plus the `lhs`, `lhs_V`, `rhs`, and
+  `rhs_V` tacticals (and primed variants) described in `STYLE.md`.
```
jdchristensen: We should add the following sentence to the discussion of `lhs`, etc, in STYLE.md:
```
  Primed variants (`lhs'`, etc) exist which apply a specified tactic
  to one side of any symmetric, transitive relation (such as `==*` or `<~>`).
```

.gitignore
```diff
@@ -116,3 +113,9 @@ nix/profiles/
 
 # Ignore the file to bench
 file_to_bench
+
+# Local AI assistant state
+/.agents/
+/.claude/
+/.codex/
+/.pi/
```
jdchristensen, line 124: I'm not sure I agree, but Claude says:

/.claude/ in .gitignore is probably too broad. That directory is where a project would keep shared config — slash commands, subagents, hooks, project settings. Ignoring the whole thing means the repo can never ship any of that. The usual split is to track .claude/ and ignore .claude/settings.local.json (plus CLAUDE.local.md). More generally, per-developer tool state is a decent candidate for a global gitignore or .git/info/exclude rather than the project's; if you do want it in-repo, the list is also arbitrary (.pi/, .codex/, but not .cursor/, .aider*, .gemini/, .opencode/).

.github/workflows/copilot-setup-steps.yml
jdchristensen, line 1: I don't understand this, so I'll just quote Claude:

### Copilot workflow: version drift and an implicit cache dependency

The pins in `copilot-setup-steps.yml` are all real and current (`actions/checkout@v7.0.1`, `nixbuild/nix-quick-install-action@v35`, `nix-community/cache-nix-action@v7`), but they're well ahead of `ci.yml`, which is on `checkout@v4`, `nix-quick-install@v30`, `cache-nix-action@v6`. Two consequences:

- The new job passes `save: false`, so it relies on some *other* job having populated `nix-${{ runner.os }}-${{ hashFiles('flake.lock') }}`. Today that's `doc-alectryon`, which uses the same primary key — so it works, but it's an implicit cross-job coupling worth a comment in the YAML, since deleting or re-keying that job would silently turn this step into a no-op.
- That saving job is on `cache-nix-action@v6` while this one restores with `v7`. Worth confirming the cache formats are interchangeable; if they aren't, the restore misses every time and the caching is dead weight. `gc-max-store-size-linux: 2G` is also inert under `save: false`.

Simplest fix is to pin the new file to the versions `ci.yml` already uses, or bump both together. Two `checkout` majors in one repo tends to rot.

*Drafted by Claude (Opus 5).*

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+Repository-wide guidance for coding agents. A more specific `AGENTS.md` in a
+subdirectory takes precedence for files below it.
+
+## Project
+
+HoTT is a Rocq (formerly Coq) library for homotopy type theory. Library sources
+live in `theories/`, additional developments in `contrib/`, tests in `test/`,
+and developer tooling in `etc/`.
+
+- Preserve the required `-noinit` and `-indices-matter` flags by using the
```
jdchristensen, line 12: Mention that `Basics` (especially `Basics.Overture`) provides many of the things usually provided by Init?

AGENTS.md
```diff
@@ -0,0 +1,67 @@
...
+- Prove results at their natural level of generality: consider arbitrary
+  truncation levels, modalities, or reflective subuniverses when appropriate.
+- Prefer reusable lemmas over long proofs, and keep intermediate goals readable
+  enough for a human to follow. Useful project tactics include `napply`,
+  `rapply`, `snapply`, and `srapply`, plus the `lhs`, `lhs_V`, `rhs`, and
+  `rhs_V` tacticals (and primed variants) described in `STYLE.md`.
+- Preserve intended universe polymorphism while avoiding unnecessary universe
+  parameters and constraints. Inspect universe-sensitive definitions with
+  `About` or `Print` and `Set Printing Universes.`; add regression checks under
+  `test/` when appropriate.
```
jdchristensen: `add regression checks...` could be a separate bullet point, since it applies to things besides universe variables.

AGENTS.md
jdchristensen: Split into two bullets, and add (s)tapply:
```suggestion
- Prefer reusable lemmas over long proofs, and keep intermediate goals readable
  enough for a human to follow.
- Useful project tactics include `napply`, `snapply`, `rapply`, srapply`,
  `tapply`, `stapply` and the `lhs`, `lhs_V`, `rhs`, and `rhs_V` tacticals
  (and primed variants) described in `STYLE.md`.
```

jdchristensen (commented):
I asked Claude to review, and filtered out about 80% of its feedback.  I paraphrased the most sensible comments, and cut-and-paste a couple that I don't feel qualified to paraphrase.

AGENTS.md
Alizter: Typically you don't have to give full paths to LLMs as long as it can grep for the file. Since we only have one style file it should be good.

AGENTS.md
Alizter: Yeah, that only matters in the rare case we have two stanzas in the same directory and we need to partition the modules by hand.

.gitignore
Alizter: If we want to explicitly track these files with git, `.gitignore` will not prevent us from doing so. This will just make it stay untracked by default.

.github/workflows/copilot-setup-steps.yml
Alizter: Not particularly important IMO

AGENTS.md
Alizter: I think `STYLE.md` already makes this point, so probably not worth repeating.

AGENTS.md
Alizter: Yeah, I think the more correct thing to say would be to add the premise of "If asked to write commits, ..." for the second to last bullet. I won't modify this unless it causes confusion for the agent however, I think it can infer what we mean.

AGENTS.md
Alizter: Yes, that is probably the better place to do this. However, I am actually unsure how successful we are at telling the agent which tactics to use. If we think its worth mentioning these tactics in style, let's do that in a follow up and remove this here.

AGENTS.md
jdchristensen: It's only four characters, so I think it's better to give the exact path.  I frequently see Claude having to do an extra search step when it doesn't know a path, and it is a distraction.

AGENTS.md
jdchristensen: Why not just be more precise?  And I also think reversing the order makes sense.

AGENTS.md
jdchristensen: I think it's important to mention them in AGENTS.md.  Claude was frequently not using these tactics until I highlighted it.  STYLE.md is large and the agent doesn't read it completely or see all of the details.

jdchristensen:
@CharlesCNorton You've used an LLM to formalize work involving Coq-HoTT more than anyone else.  Do you have any input about the proposed AGENT.md file?

CharlesCNorton:
> @CharlesCNorton You've used an LLM to formalize work involving Coq-HoTT more than anyone else. Do you have any input about the proposed AGENT.md file?

Will need until Tuesday or so to take a closer look. Probably one or two things.

CharlesCNorton:
Suggested additions:

- Before stating a lemma, search `theories/` for an existing or more general form; if a close variant exists, generalize it in place rather than adding a new one.
- Do not add definitions that merely rename, specialize, or compose existing lemmas in one step; use the existing lemmas directly.
- Require exactly what the file uses: no reliance on transitive imports, and no meta-file imports (`Categories`, `Basics`) from within their own subtrees.
- Time changed files against master (see `doc/TIMING.md`) and avoid slow `Defined`/`Qed` lines. End a proof with `Defined` when later proofs compute through its content (equivalences, homotopies), and with `Qed` otherwise.
- Make arguments implicit when inferable from later arguments (objects from morphisms, groups from homomorphisms), and declare instances with `Instance` at the definition rather than wrapping them later.

jdchristensen:
> Suggested additions:

Thanks, @CharlesCNorton , those are helpful!

The only one I would modify is:

> * Time changed files against master (see `doc/TIMING.md`) and avoid slow `Defined`/`Qed` lines. End a proof with `Defined` when later proofs compute through its content (equivalences, homotopies), and with `Qed` otherwise.

I would omit mention of timing in general, as it's only important for large developments and it's a bit tricky to do well, depending on the hardware.

And about Defined/Qed, our default is to end most things with `Defined` unless there is a specific performance reason to use `Qed`.  Our STYLE.md leans a bit more towards `Qed.`, so it should probably be revised to reflect our actual practice, and then nothing needs to be said about this in AGENTS.md.

AGENTS.md
```diff
@@ -0,0 +1,85 @@
...
+live in `theories/`, additional developments in `contrib/`, tests in `test/`,
+and developer tooling in `etc/`.
+
+- Preserve the required `-noinit` and `-indices-matter` flags by using the
+  project build rather than invoking Rocq with ad-hoc arguments.
+- Note that `Basics` (especially `Basics.Overture`) provides many of the
+  things usually provided by `Init` in standard Rocq developments.
+- Look up the supported Rocq versions in the CI matrices
+  (`.github/workflows/ci.yml`) rather than hard-coding them here; avoid
+  unnecessary version-specific features.
```
jdchristensen, line 18:
```suggestion
- Look up the supported Rocq versions in the CI matrices
  (`.github/workflows/ci.yml`); avoid
  unnecessary version-specific features.
```

jdchristensen (approved):
LGTM!

PR #2381, opened 2026-08-08
π₃(S²) = ℤ and supporting material

jdchristensen:
Using the usual argument involving the Hopf fibration, we prove that pi_n(S^3) = pi_n(S^2) and in particular that pi_3(S^2) is the integers.  A direct proof only required about 50 lines of new code, but some of the constructions were awkward because we were missing the group-valued LES in homotopy groups.  (We had the LES as pointed-sets, but the objects were only definitionally equal to `Pi n (-)` after destructing `n`, and the morphisms have different pointedness proofs, so it was awkward to apply that in a situation where you want to have groups.

I also added a couple of general lemmas about exact sequences of groups to ShortExactSequence.v, and a few more equivalences to JoinSusp.v.  (Only one is used, but the other two will likely be handy to have.)

HomotopyGroup.v now depends on ExactSequence.v, which increases its transitive dependencies from 121 to 125, which doesn't seem like a big deal.

This was co-written with Claude.  Fable 5 did the first proof entirely autonomously in about 30 minutes for US$20 in usage credits.  Then Opus 5 and I went through many rounds of revisions before settling on the current structure.  I used the draft AGENTS.md file from #2379, along with some custom advice, and it worked well.


Alizter (approved):
I'm very happy to see this!

PR #2369, opened 2026-06-11
EM-space classification infrastructure

CharlesCNorton:
Short exact sequences of abelian groups are classified by pointed maps of Eilenberg-Mac Lane spaces:

```
AbSES B A  <~>  (K(B,2) ->* K(A,3))
```

This is Theorem 2.2.2 of Christensen and Flaten, "Ext groups in homotopy type theory". It follows that `AbSES B A` and `Ext B A` are essentially small.

### New file
- `Algebra/AbSES/Classification.v`: the classifying map of a short exact sequence, namely the connecting map of `K(-,3)` applied to it; its inverse, built from the fibre of a pointed map; the two round trips; the resulting equivalence; smallness of `AbSES` and `Ext`; and naturality in both variables.

### Supporting changes
- `Homotopy/EMSpace.v`: `K'`, the functor `K(-,n)` on groups, with its `Is0Functor`, `Is1Functor` and `IsPointedFunctor` instances, and its naturality with respect to the loop identifications; connectivity of `fmap (K' n.+1)` of a surjection; `equiv_em_fmap` and `path_em_pmap_pi`; that every `(n-1)`-connected `n`-type is `K(Pi n, n)`, with the resulting `path_pmap_pi_connected`; and the delooping equivalence on mapping types.
- `Homotopy/ExactSequence.v`: naturality and rotation of the connecting map.
- `Pointed/pFiber.v`: `pfiber2_loops_beta`, and naturality of `pfiber2_loops`.
- `Pointed/Loops.v`: `equiv_istrunc_istrunc_iterated_loops`, generalizing 7.2.9 to any truncation level, with the previous result and `istrunc_iterated_loops` as instances.
- `Homotopy/ClassifyingSpace/Core.v`, `Homotopy/HomotopyGroup.v`, `Pointed/Core.v`, `Pointed/pEquiv.v`, `Pointed/pMap.v`, `Basics/Trunc.v`: supporting lemmas and generalizations.

### Provenance
This PR was generated by an LLM: `gemma-4-31B-it-qat-q4_0`, a quantized model running locally on an RTX 6000 Ada Generation with ECC enabled, driven through Claude Code as the agent harness.

PR #2382, opened 2026-08-11
Classify short exact sequences by maps of Eilenberg-Mac Lane spaces

CharlesCNorton:
Short exact sequences of abelian groups are classified by pointed maps of Eilenberg-Mac Lane spaces:

```
AbSES B A  <~>  (K(B,2) ->* K(A,3))
```

This is Theorem 2.2.2 of Christensen and Flaten, "Ext groups in homotopy type theory". It follows that `AbSES B A` and `Ext B A` are essentially small. This is the file deferred from #2369, with the review feedback from that PR applied in advance: goals kept readable, large posed terms and asserts removed where they could arise naturally, `lhs'`-style proofs, and no redundant wrappers or annotations.

### New file
- `Algebra/AbSES/Classification.v`: the classifying map of a short exact sequence, namely the connecting map of `K(-,3)` applied to it; its inverse, built from the fibre of a pointed map; the two round trips; the resulting equivalence; smallness of `AbSES` and `Ext`; and naturality in both variables. Exported through `AbSES.v` and hence `HoTT.v`.

### Supporting changes
- `Homotopy/EMSpace.v`: instances `isconnected_em_succ` and `istrunc_em_succ`, so that typeclass search finds connectivity and truncatedness of `K(G, n.+2)` at successor-shaped indices; this resolves the search failures noted in the review of #2369.

### Provenance
This PR was generated by an LLM: `Muse-Glimmer-30B` at `Q5_K_XL` quantization, running locally, driven through Claude Code as the agent harness.


test/Homotopy/EMSpace.v
```diff
@@ -0,0 +1,15 @@
...
+From HoTT Require Import Homotopy.EMSpace Algebra.AbGroups.AbelianGroup.
+From HoTT Require Import Truncations.Core Truncations.Connectedness.
+
+(** Test that typeclass search finds [isconnected_em_succ] and [istrunc_em_succ]. *)
+
+Local Open Scope trunc_scope.
+
+Definition test_isconnected_em_succ `{Univalence} (G : AbGroup) (n := 1%nat)
+  : IsConnected n.+1 (K' 3 G)
+  := _.
```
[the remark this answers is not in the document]
Alizter: @CharlesCNorton Is there a point to these tests that I've missed?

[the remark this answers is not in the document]
CharlesCNorton: @Alizter Dropping `isconnected_em_succ` makes `test/Homotopy/EMSpace.v` fail, since resolution matches on the index and will not invert `nat_to_trunc_index` to reach `isconnected_em`. The truncation one was not needed, since unifying the space already determines the index there, so I removed it with its test in b010ccd3.

jdchristensen:
@CharlesCNorton Thanks for this!  I'm a bit behind on things, but plan to look closely at this next week.

theories/Homotopy/EMSpace.v
```diff
@@ -57,6 +57,11 @@ Section EilenbergMacLane.
     rapply (is0connected_isconnected n.-2).
   Defined.
 
+  (** Typeclass search cannot invert the [nat] to [trunc_index] coercion, so it does not find [isconnected_em] when the index is written [n.+1] in [trunc_scope]. *)
+  #[export] Instance isconnected_em_succ {G : Group} (n : nat)
+    : IsConnected ((nat_to_trunc_index n).+1)%trunc K(G, n.+2)
+    := isconnected_em n.+1.
```
jdchristensen: This instance is very specific to this situation.  E.g. it doesn't handle the goal
```
   : IsConnected ((nat_to_trunc_index n).+2)%trunc K(G, n.+3)
```
But I guess there's no easy way to handle all such goals with one new instance, or to tell instance search to unfold things so that the original instance works?

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+
+(** * Classification of short exact sequences
+
+Short exact sequences [A -> E -> B] of abelian groups are classified by pointed maps [K(B,2) ->* K(A,3)] (Christensen and Flaten, "Ext groups in homotopy type theory", Theorem 2.2.2). *)
+
+Local Open Scope pointed_scope.
+
+(** TODO: The main results of this file, such as [equiv_abses_classifying_map] and [issmall_abses], have a large number of universe variables, inherited from the delooping layer in EMSpace.v.  See the TODO there. *)
+
+(** ** Vanishing of homotopy groups *)
```
jdchristensen: A lot of things near the beginning of this file aren't specific at all to AbSES, so should be moved to appropriate files.  (And maybe this is true of things deeper in the file too?  I haven't looked yet.)

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+(** ** Injectivity of homotopy groups of fiber inclusions
+
+If the [n.+1]-st homotopy group of the double fiber of [g] vanishes, then [Pi n.+1] of the fiber inclusion of [g] is an embedding. *)
+Instance isembedding_pi_pfib `{Univalence} {X Y : pType} (g : X ->* Y)
+  (n : nat) `{Contr (Pi n.+1 (pfiber (pfib g)))}
+  : IsEmbedding (fmap (Pi n.+1) (pfib g)).
+Proof.
+  exact (isembedding_isexact (i := fmap (pPi n.+1) (pfib (pfib g)))
+           (isexact_pi_total (pfib (pfib g)) (pfib g) (n.+1)%nat)).
+Defined.
```
jdchristensen: This is much too specific.  In *any* exact sequence of groups, if one of the groups is trivial, the next map is an embedding.  This is part of #2381 , and the material there can probably make many things in this file simpler.  Please try to make use of that material.  (Also, why a heading for a single minor result?  And the scope of a heading continues until the next heading at the same or higher level, so that's also incorrect here.)

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+         _ issurjection_abses_pfiber_proj _.
+
+End AbSESPfiber.
+
+Section PfiberDeloop.
+  Context `{Univalence} {B A : AbGroup@{u}} (psi : K(B, 3) ->* K(A, 4)).
+
+  (** The fiber of a map [K(B,3) ->* K(A,4)] is 3-truncated. *)
+  Local Instance istrunc_pfiber_em : IsTrunc 3 (pfiber psi)
+    := _.
```
jdchristensen: Since this is found by typeclass search, why is it made an instance?  Sometimes this is done for speed, and if that's the reason, there should be a comment indicating it.  But otherwise, this should be removed.

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+  exact (isembedding_isexact (i := fmap (pPi n.+1) (pfib (pfib g)))
+           (isexact_pi_total (pfib (pfib g)) (pfib g) (n.+1)%nat)).
+Defined.
+
+Section EMFiberSequence.
+  Context `{Univalence} {B A : AbGroup@{u}} (E : AbSES B A) (n : nat).
+
+  (** Applying [K(-,n.+1)] to a short exact sequence yields a complex. *)
+  Definition iscomplex_em_abses
+    : IsComplex (fmap (K' n.+1) (inclusion E)) (fmap (K' n.+1) (projection E)).
```
jdchristensen: This should be true for any pointed functor, so should probably be stated in that generality.  Even better, a pointed functor takes *any* complex to a complex.

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+    refine ((fmap_comp (Pi n.+1) (cxfib iscomplex_em_abses)
+              (pfib (fmap (K' n.+1) (projection E))) _)^ @ _).
+    refine (fmap2 (Pi n.+1) (pfib_cxfib _) _ @ _).
+    refine (pi_em_fmap (inclusion E) n a @ _).
+    exact (ap (equiv_g_pi_n_em E n) (ap pr1 q) @ eisretr _ x).
+  Defined.
+
+  (** Therefore the comparison map is an equivalence: it is an isomorphism on [Pi n.+1], and both sides are [n]-connected and [n.+1]-truncated. *)
+  Local Instance isequiv_pi_cxfib
+    : IsEquiv (fmap (Pi n.+1) (cxfib iscomplex_em_abses)).
```
jdchristensen: This result and the previous result that it depends on have long, complicated proofs, but this result is conceptually trivial.  We know that the two homotopy groups we are comparing are both kernels of the *same* map, so they have to be equivalent.  This is a simple lemma about abelian groups that should be proved and then applied here.  (Also note that the comment before this result is not correct.)

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+    apply (equiv_inj (equiv_g_pi_n_em E n)).
+    refine ((pi_em_fmap (inclusion E) n a)^ @ _).
+    refine (ap (fmap (Pi n.+1) (fmap (K' n.+1) (inclusion E)))
+              (eisretr (equiv_g_pi_n_em A n) z) @ _).
+    refine (wi @ _).
+    exact (grp_homo_unit (equiv_g_pi_n_em E n))^.
+  Defined.
+
+  (** Both sides of the comparison map are [n]-connected and [n.+1]-truncated, so it is an equivalence. *)
+  Local Instance isequiv_cxfib_em : IsEquiv (cxfib iscomplex_em_abses).
```
jdchristensen: This also has a long, indirect proof, but once you have the equivalence on pi_{n+1}, it's a simple consequence of Whitehead's theorem, and shouldn't involve `A` at all.

theories/Algebra/AbSES/Classification.v
jdchristensen: Also note that here and probably in many places, there is no reason to use `n.+1` instead of `n`.  (Another example is `contr_pi_succ_istrunc` from the last PR.)

theories/Algebra/AbSES/Classification.v
```diff
@@ -0,0 +1,1035 @@
...
+Local Open Scope pointed_scope.
+
+(** TODO: The main results of this file, such as [equiv_abses_classifying_map] and [issmall_abses], have a large number of universe variables, inherited from the delooping layer in EMSpace.v.  See the TODO there. *)
+
+(** ** Vanishing of homotopy groups *)
+
+(** Homotopy groups below the connectivity vanish. *)
+Definition contr_pi_isconnected `{Univalence} (n : nat) (X : pType)
+  `{IsConnected n X}
+  : Contr (Pi n X).
```
jdchristensen: Here's a more conceptual proof, using helper that I'm surprised we don't have, but definitely want.
```
(** This homotopy groups of a contractible space vanish. *)
Instance contr_pi_contr (n : nat) (X : pType) `{Contr X}
  : Contr (Pi n X).
Proof.
  generalize dependent X; induction n; intros.
  - exact _.
  - exact (contr_equiv' _ (pi_loops n X)^-1%equiv).
Defined.

(** Homotopy groups below the connectivity vanish. *)
Definition contr_pi_isconnected `{Univalence} (n : nat) (X : pType)
  `{IsConnected n X}
  : Contr (Pi n X).
Proof.
  rapply (contr_equiv' (Pi n (pTr n X))).
  symmetry; rapply pequiv_pi_Tr.
Defined.
```

jdchristensen (commented):
I just did a few spot checks, and every place I looked had awkward arguments, often not stated in full generality.  I think this file is long and slow to build because conceptual ideas aren't being used, and things are being brute-forced inside proofs.  As I mention in one comment, I think that the results in #2381 will help a lot.  But I'm not sure I'm going to have the stamina to guide this to a proper execution, and am considering the idea of putting it in contrib.  This also has the advantage that it won't slow down ordinary library builds.  But let's at least try a round or two of fixes and see how it goes.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Added both to HomotopyGroup.v, as written.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Moved out: `contr_pi_contr`, `contr_pi_isconnected` and `isconnected_succ_contr_pi` to HomotopyGroup.v; `abgroup_pi`, `pi_em_fmap'` and `pequiv_em_pi` to EMSpace.v; the fiber of the connecting map to ExactSequence.v.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Deleted, with the heading. Uses go through `isembedding_isexact` against `Pi_les`, which now also gives all three conditions defining `abses_pfiber`.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Added `fmap_iscomplex` to PointedCat.v; `iscomplex_em_abses` is that applied to `iscomplex_abses`, stated at `n`. I also lowered the index of `contr_pi_succ_istrunc`.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Added `isequiv_isexact_factor` to ShortExactSequence.v: two maps exact at the same `f`, one factoring through the other, agree. Applied to the exactness of `E`, transported along `equiv_g_pi_n_em`. The comment is fixed.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Added `isequiv_pi_connected_truncated` to WhiteheadsPrinciple.v, by induction through `isequiv_is0connected_isequiv_loops`. `isequiv_cxfib_em` is now four lines and does not mention `A`.

theories/Algebra/AbSES/Classification.v
CharlesCNorton: Removed.

theories/Homotopy/EMSpace.v
CharlesCNorton: The obstruction is the scope of the index: `IsConnected (n.+1)%nat X` is found without any such instance, while `IsConnected (nat_to_trunc_index n).+1 X` is not. So those hypotheses are now in nat scope, and both successor instances and their test are deleted.

CharlesCNorton:
@jdchristensen Pushed a few rounds addressing your comments. The general results now live in PointedCat, ShortExactSequence, WhiteheadsPrinciple, HomotopyGroup, EMSpace and ExactSequence; what remains in Classification.v is specific to `AbSES`.

CharlesCNorton:
Timed against master `b52bd4e9`:

```
                     Classification.vo    Peak Mem
before (ba5bee3f)             0m07.74s   736300 ko
after                         0m03.19s   485824 ko
```

The branch is +5.20s over master.

jdchristensen:
> Timed against master `b52bd4e9`:
> 
> ```
>                      Classification.vo    Peak Mem
> before (ba5bee3f)             0m07.74s   736300 ko
> after                         0m03.19s   485824 ko
> ```
> 
> The branch is +5.20s over master.

So the new file is down to 3.2s, but other changes add 2s?  The 3.2s is a great improvement, but what other additions or changes are adding 2s?  Or maybe I'm misunderstanding.

I pushed a few things that generalize some of your work, but still need to look through things further.  I'm on vacation, so I may not get a chance for a little while.

theories/Algebra/Groups/ShortExactSequence.v
```diff
@@ -66,6 +67,26 @@ Definition equiv_grp_isexact_kernel `{Univalence} {A B : Group} (f : A $-> B)
   := (equiv_istrivial_kernel_isembedding f)^-1%equiv
        oE equiv_iff_hprop_uncurried (iff_grp_isexact_isembedding f).
 
+(** If [i] factors through [j] as [j o k], and [i] and [j] are both exact at [B] with respect to the same [f], then [k] is an equivalence.  Both [A] and [C] are the kernel of [f], so they agree.  Only [cx] is needed for [j], and only [(-1)]-exactness for [i]. *)
+Definition isequiv_isexact_factor {A B C D : Group}
+  {i : A $-> B} {j : C $-> B} (k : A $-> C) {f : B $-> D}
+  (p : j $o k $== i) `{IsEmbedding i} `{IsEmbedding j}
+  (ex : IsExact (Tr (-1)) i f) (cx : IsComplex j f)
+  : IsEquiv k.
```
jdchristensen: `isequiv_isexact_factor` has assymmetrical hypotheses.  It's probably better to assume that both i and j are part of exact sequences with the same map f, so both are kernels.  (In other words, get rid of `k` and make the assumption on `j` the same as the assumption on `i`.)  In this form, it will follow easily from a lemma that says that two kernels of the same map are isomorphic.  The latter lemma would probably fit best in Groups/Subgroup.v.  The argument for it is symmetrical, so the proof of the two composites being the identity could be factored out and used twice if it's more than, say, two lines.

jdchristensen: (Actually, the lemma about kernels is true in any wild category, and Thomas was thinking of adding a file on kernels there, so once that is done, the lemma in Groups/Subgroup.v could be replaced by the general result.)

CharlesCNorton:
About 3.9s is attributable: `Classification.vo` +2.99s, `Subgroup.vo` +0.62s (the kernel lemmas), `EMSpace.vo` +0.28s, everything else under 0.1s. The remaining 1.4s is `-j8` jitter -- the biggest movers are files nothing here touches, in both directions (`premetric.vo` -0.81s, `Abelianization.vo` -0.36s against `Hartogs.vo` +0.39s, `BAut/Bool.vo` +0.29s), plus 265 untouched files each under 0.1s summing to +0.98s.

theories/Algebra/Groups/Subgroup.v
```diff
@@ -1372,6 +1372,53 @@ Proof.
...
 Global Hint Immediate istrivial_kernel_isembedding : typeclass_instances.
 
+(** ** Two embeddings with the same image *)
+
+(** An embedding is determined by its image: if [i] and [j] are embeddings into [B] and everything in the image of [i] is in the image of [j], then [i] factors through [j].  The factorization is unique, since [j] is an embedding. *)
+Definition grp_homo_embedding_image {A C B : Group} (i : A $-> B) (j : C $-> B)
+  `{IsEmbedding j} (h : forall b, merely (hfiber i b) -> merely (hfiber j b))
+  : A $-> C.
+Proof.
+  (** Since [j] is an embedding, its fibers are propositions, so the mere factorization gives an actual one. *)
```
jdchristensen: Comments within proofs should use `(*` not `(**`.

theories/Algebra/Groups/ShortExactSequence.v
jdchristensen: The latest commit makes things more complicated not less complicated.  When I get a chance, I'll see if I can improve this.

jdchristensen:
> About 3.9s is attributable: `Classification.vo` +2.99s, `Subgroup.vo` +0.62s (the kernel lemmas), `EMSpace.vo` +0.28s, everything else under 0.1s. The remaining 1.4s is `-j8` jitter -- the biggest movers are files nothing here touches, in both directions (`premetric.vo` -0.81s, `Abelianization.vo` -0.36s against `Hartogs.vo` +0.39s, `BAut/Bool.vo` +0.29s), plus 265 untouched files each under 0.1s summing to +0.98s.

You'll get more reliable data with -j1 or -j2.  And if you use the "only-diff" version, it only compares files changed and files that depend on changed files.  Note that even if a file is not changed, changes to its dependencies can affect its timing so timing differences can't be assumed to be noise.

theories/Algebra/Groups/ShortExactSequence.v
CharlesCNorton: `grp_iso_cxfib` already identifies any embedding exact at `f` with `grp_kernel f`, so the symmetric lemma is that composite:

```coq
Definition grp_iso_isexact_isexact {A B C D : Group}
  {i : A $-> B} {j : C $-> B} {f : B $-> D}
  `{IsEmbedding i} `{IsEmbedding j}
  (exi : IsExact (Tr (-1)) i f) (exj : IsExact (Tr (-1)) j f)
  : GroupIsomorphism A C
  := grp_iso_compose (grp_iso_inverse (grp_iso_cxfib exj)) (grp_iso_cxfib exi).
```

and `isequiv_isexact_factor` follows from it in four lines through `grp_iso_cxfib_beta`. That drops `merely_hfiber_isexact` and the three `grp_homo_embedding_image` definitions, so Subgroup.v is back to master and the addition here is 22 lines in one file. Since the proof is `grp_iso_cxfib`, the lemma sits directly beneath it rather than in Subgroup.v.

theories/Algebra/Groups/Subgroup.v
CharlesCNorton: Fixed. Those two were inside the definitions that are now gone, so Subgroup.v is unchanged from master. I also checked the rest of the branch and corrected four more that it had introduced, in Classification.v and WhiteheadsPrinciple.v.

CharlesCNorton:
Redone at `-j1` with `only-diff`, against master `a2a6ced1`:

```
   After | File Name                        |   Before ||    Change
------------------------------------------------------------------
1m19.37s | Total Time                       | 1m16.00s || +0m03.37s
0m03.12s | Algebra/AbSES/Classification.vo  |   N/A    || +0m03.12s
0m00.97s | Homotopy/EMSpace.vo              | 0m00.76s || +0m00.20s
```

Nothing else contributes: across the other 299 files rebuilt, the changes sum to -0.01s, with +1.91s of increases against -1.92s of decreases and no single file past 0.18s. So the extra time in the `-j8` figure was noise. `WildCat/Core.v` is what pulls 299 files into the comparison. The `Subgroup.vo` entry from the previous table is gone with the change above.

The branch also rebases onto `a2a6ced1` without conflicts and builds there.

theories/Algebra/Groups/ShortExactSequence.v
jdchristensen: Great, this is much better.  While I was looking it over, I found some things I could improve in code that is used by this PR but wasn't added by this PR, so I pushed them.

theories/Homotopy/EMSpace.v
```diff
@@ -352,10 +363,45 @@ Section EilenbergMacLane.
...
 
+  Section NormalizedEM.
+    Context (n : nat) (X : pType)
+      `{IsConnected (n.+1)%nat X} `{IsTrunc (n.+2)%nat X}.
+
+    (** The automorphism of [Pi n.+2 X] by which [pequiv_em_connected_truncated] differs from [equiv_g_pi_n_em]. *)
+    Local Definition eta_em_pi
+      : abgroup_pi n X $<~> abgroup_pi n X
+      := emap (Pi n.+2) (pequiv_em_connected_truncated X n.+1)
+             $oE equiv_g_pi_n_em (abgroup_pi n X) n.+1.
```
jdchristensen: I strongly suspect that `eta_em_pi` is homotopic to the identity map, so that the corrected equivalence isn't needed, but I don't immediately see a way to prove this.  Can you see if your LLM can prove this, and therefore remove the need to correct the equivalence?  Maybe earlier results in the file, such as `loops_em_em_ptr_unit`, `em_fmap_loops_natural`, `equiv_g_pi_n_em_succ`, and `pi_em_fmap`, might help?

It would be fine if this was easier using different proofs of `pequiv_em_connected_truncated` or `equiv_g_pi_n_em`, with the former being the one that seems most likely to be fixable to make this work.

Looking at the bigger picture, what we are really missing is the fact that for `X` and `Y` n-connected and n.+1-truncated, `fmap (Pi n.+1) : (X ->* Y) -> (Pi n.+1 X $-> Pi n.+1 Y)` is an equivalence.  In fact, for what we need, it's enough to have it when one of the two is an EM space.  From this, we could prove `pequiv_em_connected_truncated` so that it computed on homotopy groups like we want.  (This missing result in fact follows from `pequiv_em_connected_truncated`, so we could fall back to proving it using that result, but it would be better if we could prove it more directly, so that `isequiv_em_fmap` was a consequence, rather than needing a separate proof.)

Summarizing:  (1) First try to just prove that `eta_em_pi` is homotopic to the identity, with no changes.  (2) Then try with changes to the proofs of `pequiv_em_connected_truncated` or `equiv_g_pi_n_em`.  (3) Optionally, see if you can prove the missing result about maps between connected, truncated types in a way that lets other results follow easily.  (3) is probably a big job, so is certainly not required.

In all cases, this is only worth it if it ends up being fairly clean.  If it ends up being more complicated than what we have now, then we can keep the current approach.

jdchristensen (commented):
I've now reviewed all of the changes outside of Classification.v.  It may be a few days before I have time to look at this again, but I'll keep at it when I am able to.

theories/Homotopy/EMSpace.v
jdchristensen: @CharlesCNorton Actually, (3) is pretty easy, and I'm most of the way done it, so hold off on any changes for now.

jdchristensen: It took longer than I expected to get (3) into a PR quality state, but it's now done in #2389, which has been merged.  If you rebase or merge against it, it should make NormalizedEM go away.

CharlesCNorton:
back to working on this later today. thank you for the review.

CharlesCNorton:
@jdchristensen Merged with master: `NormalizedEM` is gone, replaced by `pequiv_em_connected_truncated` and its computation rule `fmap_pi_pequiv_em_connected_truncated`, and the five uses of `path_pmap_pi_connected` now sit one index lower without the connectivity and truncatedness arguments.  The second commit names the round-trip lemmas after their statements, replaces the local loop-inversion naturality lemma by `is1natural_loops_inv`, moves `loops_inv_inv` next to `loops_inv` in `Pointed/Loops.v`, and drops the `Opaque` bracket around the naturality section, which measured as noise.

jdchristensen:
I worked on this all day and just pushed a bunch of commits which clean up and simplify a lot of things.  I first finished reviewing the files outside of Classification.v and am happy with them now.  Then I asked Claude Fable to review Classification.v.  Opus and I handled around 15-20 of the items, but I am now out of time, so I'm posting the remaining items here.  (As usual, I ask that you look at the changes I've made to see improvements that you could have made and as style advice for future wo.  For example, I had asked that the `lhs` family of tacticals be used, but they were not consistently used.)

@CharlesCNorton , if you want to tackle some of these, please go ahead.  Keep in mind that they may not all be accurate, so each one should be evaluated and should only be committed if it leads to a genuine improvement.  I have not looked closely at these remaining items.  (But so far almost all of the suggestions did result in shorter and faster code.)  If you do some of them, each should be a separate commit, and should reference which item it is address using the numbering below.

I am also willing to work on them at some point, but it will likely be more than a week before I can devote the time to this again, as I've more than used up my time budget for now.

I should also clarify that I haven't reviewed Classification.v carefully myself, and when I do, I may find additional things that need fixing.

# Review of `theories/Algebra/AbSES/Classification.v` (PR #2382)

Scope: `Classification.v` (596 lines), with a few neighbouring files where a lemma would better belong.  Line numbers refer to the current file at commit 5cd2cb9f.  Items are numbered for reference; the letters group them by kind, roughly from "ask the author" to "just fix and push".  The numbering has gaps because many of the original items have been completed and removed.

Items marked "verified" were checked by compiling the code shown.  The lemmas proposed in section B were compiled standalone against a minimal import set rather than spliced into the library, so their homes and names are still open.

Build/timing: `coqc -time` takes about 2.9s for 247 sentences.

(When this review was first written, at commit 8d9c57b1b, Classification.v was 701 lines with 373 sentences and took 3.6s to build.)

Tactic note: when an argument is dropped and elaboration has to unify a wildcat shape (`?a $-> ?b`, `?f $== ?g`, `Hom Group` against `Hom AbGroup`), `napply`, `rapply` and `exact` all fail where `tapply` succeeds.

## A. Larger issues (for the author)

**A1. Naturality of connecting maps is proved twice, in special cases, and belongs in `ExactSequence.v`.** `PfiberDeloop` (`phomotopy_cxfib_connect` + `connecting_map_em_loops`, lines 239-295) and `Naturality` (`em_cxfib_square` + `cm_natural`, lines 539-582) both chain `connecting_map_cxfib`, `connecting_map_natural(_functor)`, `moveR_pequiv_Vf`/`moveL_pequiv_Vf` and associativity to transport a connecting map along a square, 20-25 lines each.

This is not specific to Eilenberg-Mac Lane spaces.  For purely exact `(i, f)` and `(i', f')`, maps `g, h, k`, squares `p : h o* i' ==* i o* g`, `q : k o* f' ==* f o* h`, and a coherence `c : functor_pfiber q o* pequiv_cxfib ==* pequiv_cxfib o* g` (the datum that makes `(g, h, k)` a map of fibre sequences), one gets `g o* connecting_map i' f' ==* connecting_map i f o* fmap loops k` by exactly the argument of `cm_natural`.  This is the TODO at `ExactSequence.v:521` ("It would be nice to extend them to any `FiberSeq` or equivalently any purely-exact sequence").  With such a lemma, `cm_natural` disappears, and `connecting_map_em_loops` reduces to that lemma applied to the square into the rotated sequence `(connecting_map (pfib psi) psi, pfib psi)` (purely exact by the instance `isexact_purely_fiberseq (connect_fiberseq (pfib psi) psi)`) followed by `connecting_map_pfib2`.  Not verified; the amount of work is moderate.

**A2. `Section EMFiberSequence` proves a general Whitehead-style fact in a special case.**  Apart from `iscomplex_em_abses`, `isexact_pi_em_abses` and the embedding bullet at lines 67-75, nothing in lines 25-95 is about Eilenberg-Mac Lane spaces.  The general statement is: a complex `i : F ->* X`, `f : X ->* Y` with `F` `n`-connected and `n.+1`-truncated, `X, Y` `n.+1`-truncated, `f` `n`-connected, such that `fmap (Pi n.+1) i` is an embedding and `Pi n.+1` of the complex is exact, is purely exact.  That would be a natural addition to `WhiteheadsPrinciple.v` (it uses `isequiv_isconnected_istrunc_isequiv_pi`, `isequiv_isexact_factor` and the `Pi_les` lemmas), and the section here would shrink to the three EM-specific inputs.  Lower priority than A1.

**A4. The loop-space detour in `grp_iso_a_pi_loops`.** `equiv_g_pi_n_em A n.+2` is *defined* as `groupiso_pi_loops^-1 o groupiso_pi_functor (pequiv_loops_em_em A n.+2) o equiv_g_pi_n_em A n.+1` (`EMSpace.v:181-189`).  So `grp_iso_a_pi_loops` (line 115), which is `groupiso_pi_loops o equiv_g_pi_n_em A n.+2`, goes to `Pi n.+3 K(A, n.+3)` and back to `Pi n.+2 (loops K(A, n.+3))`, and later proofs have to undo it: the `eisretr (groupiso_pi_loops 2 K(A, 4))` at line 219, and lines 433-441 of `grp_iso_pi_pfiber_classifying_map_projection`.

That second one shows the cost most clearly.  The first step there makes the left side deliberately *more* complicated, inserting a `groupiso_pi_loops o groupiso_pi_loops^-1` round trip so that `equiv_g_pi_n_em E 2 (grp_iso_pi_pfiber_classifying_map x)` appears as a subterm; the proof then pushes `pi_loops` out by naturality, peels it off with `ap (groupiso_pi_loops 1 K(B, 3))`, and closes with `eisretr`, which works only because `equiv_g_pi_n_em B 2` is defined through `groupiso_pi_loops`.  Both the inserted round trip and the closing `eisretr` exist solely to undo that definition.

Defining `grp_iso_a_pi_loops := groupiso_pi_functor n.+1 (pequiv_loops_em_em A n.+2) $oE equiv_g_pi_n_em A n.+1`, or equivalently defining `abses_pfiber_incl` as `fmap (Pi n.+2) (connecting_map (pfib f) f o* pequiv_loops_em_em A n.+2) $o equiv_g_pi_n_em A n.+1` (mirroring `abses_classifying_map`, which is `connecting_map ... o* pequiv_loops_em_em B 2`), should remove those steps.  Not verified; the round-trip proofs would need to be redone.

**A5. `abses_pfiber` indexing.**  `abses_pfiber n f` takes `f : K(B, n.+2) ->* K(A, n.+3)`, so the sequence classified by a map `K(B,2) ->* K(A,3)` is `abses_pfiber 0 f` and the delooped one is `abses_pfiber 1 psi`.  Indexing by the level of the source Eilenberg-Mac Lane space would read more naturally, if easy to do.

## B. General lemmas to factor out (medium)

**B1. `fmap (Pi n.+1) i` is an embedding when `Pi n.+2 Y` is trivial.** Used three times with the same three-line proof: `isembedding_pi_pfib_em` (49-56), the `emb` assert in `isembedding_abses_pfiber_incl` (141-144), and `isembedding_pi_pfib_pfib` (230-236).  Verified to compile (with only `HomotopyGroup`, `Groups.ShortExactSequence`, `ExactSequence` imported):

```coq
Definition isembedding_fmap_pi_isexact `{Univalence} {F X Y : pType}
  (i : F ->* X) (f : X ->* Y) `{IsExact purely F X Y i f}
  (n : nat) `{Contr (Pi n.+2 Y)}
  : IsEmbedding (fmap (pPi n.+1) i)
  := isembedding_isexact (A := pPi n.+2 Y) (isexact_pi_fiber i f n.+1).
```

It belongs in `HomotopyGroup.v`'s `PiLES` section (that file would need to import `Groups.ShortExactSequence`; there is no cycle, since `ShortExactSequence.v` does not depend on `HomotopyGroup.v`).

With B1, `isembedding_abses_pfiber_incl` becomes the composite of an equivalence and an embedding (`mapinO_compose` plus B1), so the `isembedding_isinj_hset` / `isinj_embedding` detour at lines 145-148 goes away.

**B2. `fmap (Pi n) i` is surjective when `Pi n Y` is trivial.**  Used once, in `issurjection_abses_pfiber_proj` (151-158), where the exactness input is currently written out by hand.  Verified:

```coq
Definition issurj_fmap_pi_isexact `{Univalence} {F X Y : pType}
  (i : F ->* X) (f : X ->* Y) `{IsExact purely F X Y i f}
  (n : nat) `{Contr (Pi n Y)}
  : IsSurjection (fmap (pPi n) i)
  := isconnmap_O_isexact_base_contr (Tr (-1)) _ (fmap (pPi n) f)
       (H := isexact_pi_total i f n).
```

**B3. `fmap (Pi n.+1) (fmap (K' n.+1) f)` is an embedding when `f` is.** The bullet at lines 67-75 proves this inline (via injectivity and `pi_em_fmap'`).  It belongs next to `pi_em_fmap'` in `EMSpace.v`, and it is more general than needed here (any embedding `f : G $-> G'`).  The proof is `mapinO_homotopic (Tr (-1)) _ (fun x => (pi_em_fmap' f n x)^)` followed by `mapinO_compose` for `e' o f o e^-1`.  Verified in place of the bullet (with `napply` the two `MapIn` side conditions end up shelved and need an `Unshelve. all: exact _.`; `rapply` should avoid that).

B1 and B3 together remove both uses of `HSet` in the file — the detour in `isembedding_abses_pfiber_incl` (B1) and the one in the bullet above (B3) — after which `Require Import HSet.` can go.  B2 touches neither.

**B4. `pfiber2_loops_connect` (259-268) is a general fact about the tautological fibre sequence.**  Verified (scratch session importing only `Pointed` and `Homotopy.ExactSequence`; same proof):

```coq
Definition connect_fiberseq_pfib {X Y : pType} (f : X ->* Y)
  : pfiber2_loops f o* (connect_fiberseq (pfib f) f).2 ==* pmap_idmap.
```

It belongs in `ExactSequence.v` next to `pequiv_pfiber_cxfib_taut`; the proofs of `connecting_map_natural_functor`, `connecting_map_cxfib` and `connecting_map_pfib2` there all do the same `unfold connecting_map ...; ...; pequiv_pfiber_cxfib_taut` dance and could use it too.

## D. Style, naming, comments

**D1. Names.**  `cm_natural` (557) is an abbreviation; something like `connecting_map_em_natural`.  `contr_pi_em'` (134) uses a prime where `contr_pi_em_above`/`_below` (or similar) would say what differs.

**D4. `connecting_map_classifying_map` (368-394)** is the hardest proof to follow: four nested `pmap_postwhisker`s, and at lines 378-386 the same `connecting_map_natural` + `fmap_id loops` + `pmap_precompose_idmap` step appears twice (once for `square_classifying_map`, once for the square inside `pequiv_pfiber_connecting_map`).  A naturality lemma for squares whose base map is `pequiv_pmap_idmap` would remove the repeated `fmap_id loops` step.

## G. Future work

**G1. Universe variables.**  Many results have large numbers of universe variables, stemming from results in `EMSpace.v`.  Every *use* of an unannotated lemma contributes a fresh copy of its universes, so the file amplifies the EMSpace numbers: `isequiv_pi_cxfib` uses `pi_em_fmap'` twice, `isexact_pi_em_abses` uses it twice more, and so on.  Counts of universe parameters (`Set Printing Universes`, `About`):

| constant | universes |
|---|---|
| `pequiv_loops_em_em`, `equiv_g_pi_n_em` | 33 each |
| `pi_em_fmap'`, `em_fmap_loops_natural` | 67 each |
| `isconnmap_em_fmap` | 77 |
| `equiv_deloop_em_pmap` / `_unfold` | 79 / 110 |
| `pequiv_em_connected_truncated` | 199 |
| `iscomplex_em_abses` | 28 |
| `isequiv_pi_cxfib` | 202 |
| `isexact_em_abses` | 294 |
| `abses_classifying_map` | 325 |
| `abses_pfiber` | 100 |
| `equiv_abses_classifying_map` / `issmall_abses` | 621 / 622 |
| `abses_classifying_map_natural` | 646 |

The root cause is in `EMSpace.v` (`pequiv_loops_em_em` and `equiv_g_pi_n_em` should have about one universe each), so fixing it is out of scope for this file.

**G2. Build time.**  The slowest lines are currently all `Qed` lines:

| lemma | `Qed` | universes |
|---|---|---|
| `connecting_map_em_loops` (271-295) | 0.46s | 403 |
| `phomotopy_cxfib_connect` (239-256) | 0.30s | 609 |
| `phomotopy_em_proj_pfib` (193-203) | 0.13s | 239 |

They are among the most universe-heavy local definitions in the file, so G1 bears on them, but universe count does not explain the ordering: `connecting_map_em_loops` carries fewer universes than `phomotopy_cxfib_connect` and still costs half again as much, so the size of the proof term built by the long `lhs'`/`rhs'` chains is a separate contributor.  Fixing G1 should help but will not by itself make these fast.

A1 already attacks the worst of them for an unrelated reason: the chain it proposes replacing by a general naturality lemma is exactly the body of `connecting_map_em_loops`.

CharlesCNorton:
@jdchristensen Went through and pushed all of the A, B and D items, one commit each with the item in the subject, where for A5 the index of `abses_pfiber` is now implicit.  One further commit has no item: it factors the two `cxfib` coherence proofs through a general lemma in EMSpace.v, and Classification.v now builds in 1.9s here against 3.4s at 5cd2cb9f.

jdchristensen:
Great!  I'll take another look when I have a block of free time, probably in around a week or so.

PR #2384, opened 2026-08-17
Integers as Higher Inductive Type

maxipetro:
## Summary

This PR replaces the current inductive definition of `Int` in `Int.v` with a higher inductive type, following the work of Thorsten Altenkirch and Luis Scoccola in [*The Integers as a Higher Inductive Type*](https://arxiv.org/abs/2007.00167) (2020).

The PR is a joint work with Dan Christensen (@jdchristensen), with some additional polishing done by **Claude Opus 5**.

The main motivation is that the usual signed-integer representation is rather inconvenient to work with: definitions and proofs often have to deal with the normal-form representation of integers, resulting in many case distinctions. The HIT representation instead treats the integers morally as the free pointed type equipped with a biinvertible self-map. This gives a much more natural interface for defining and proving properties of integer operations.

## Advantages of the HIT approach

The new implementation has several practical advantages:

- Definitions of integer operations can work directly with `0`, successor, and predecessor, rather than having to perform case distinctions on the signed-integer representation.
- Proofs of algebraic properties are shorter and better behaved, with more results holding definitionally.  For example, `(x.+1) + y = (x + y).+1` and `(x.-1) + y = (x + y).-1` both hold definitionally for `x y : Int`, as do many other laws.
- There are several induction and recursion principles as well as a convenient lemma `int_homotopic` for proving that two functions `Int -> P` are homotopic.  Part of what makes it easy to use is that the functions being compared often compute definitionally on `0` and `int_succ`.
- The implementation follows the higher-inductive construction described by Altenkirch and Scoccola, while remaining equivalent to the existing signed-integer representation.

There are, however, some trade-offs.

## Downsides

The main disadvantages are:

- The representation of integers is no longer unique definitionally. For example, `2 - 3` is not definitionally equal to `-1`.
- The construction depends on HIT axioms, and is therefore less foundational/basic than the ordinary inductive definition of signed integers.

## Changes to the interface

There are a few important differences from the previous implementation.

- There is no default induction principle registered for `Int`. Instead, the new implementation provides several more targeted principles, including: `int_homotopic`, `int_ind_sint` and `int_ind_equiv`.  This makes it possible to choose an induction principle appropriate to the particular proof, rather than having a single registered principle that is often awkward to use.
- Some other definitions and lemmas have also been renamed to reflect the new organization.

## The role of `SInt`

The signed integers are still needed, and remain in `SInt.v`.

In particular, `SInt` is used for:

- printing and parsing integers;
- showing that `Int` has decidable paths and is therefore a set;
- proving the `int_ind_sint` induction principle;
- reducing integer expressions to signed-integer normal forms via `int_reduce`.

Thus `SInt` is not being removed. Rather, the two representations serve different purposes: `Int` provides the convenient abstract/HIT representation used by most mathematical developments, while `SInt` provides a canonical computational representation when one is needed.  However, this PR reduces `SInt.v` to the minimal skeleton required for the above uses, to avoid duplication.

## Commit structure

The first commit concerns `BiInv.v` and makes some changes that are necessary for our implementation of the integers regarding the satisfaction of the adjoint laws of bi-invertible maps. 

The second commit is a cleanup of `BiInv.v`. The material concerning `biinv_compat*` turns out not to be needed for our final implementation.

The third commit contains the main contributions of this PR which is the implementation of the integers as HIT in `Int.v` and all necessary changes and simplifications that this yields for the library. It also adds the file `SInt.v` for the old signed integers.

The library builds after each commit, so it should be possible to review the history one commit at a time. We recommend doing so, since the individual commits are substantially easier to understand than the complete diff.

For the large commit replacing the implementation of `Int.v`, the diff between the old and new versions of `Int.v` is not particularly useful because the representations are so different. For that commit, it is better to read the new `Int.v` directly. The diffs for the other changed files are much more informative and should be useful for reviewing the effects on the rest of the library.

## Size and performance

The old implementation of `Int.v` was 785 lines, while the new implementation is 751 lines. However, the new implementation also relies on `SInt.v`, which is 129 lines, so the overall implementation is longer when both files are considered.

On the other hand, the diffstat for the rest of the library (excluding tests) is:

```text
+54 -89
```

The main benefit is therefore that users of the library can write **shorter proofs and better-behaved definitions** when working with integers.

Dan tested the build timing of the large commit. It adds approximately **0.6 seconds** to the total build time, with no individual file increasing by more than **0.08 seconds**.

## Review

Dan has reviewed the changes extensively, and Claude Opus 5 has also given the development a careful review.

Overall, this PR replaces the low-level signed-integer representation used as `Int` with the more convenient higher-inductive representation, while retaining `SInt` as the canonical computational representation. The result should provide a substantially nicer interface for developments that use the integers algebraically, at the cost of relying on the HIT axioms and giving up definitional uniqueness of integer representations.


Alizter (approved):
Very nice!

PR #2386, opened 2026-08-21
Decidable paths updates, speed improvements

jdchristensen:
These are just things that were in my HIT integers branch but were stripped out of that PR to make it cleaner.

The first commit just makes some naming consistent, using `decidablepaths` everywhere instead of sometimes using `decidable_paths`.  It also slightly generalizes one result.

The second adds that `list A` has decidable paths when `A` does.  This means that equality vectors and matrices over a ring with decidable equality can be decided.  This can be used for their tests, but in the end using `int_reduce` was faster, so this wasn't needed.  Still, it's worth having.

The last commit is completely independent, but it speeds up the build of the library by 1.0s.

### DIFF
--- a/theories/Algebra/Groups/Group.v
+++ b/theories/Algebra/Groups/Group.v
@@ -55,6 +55,9 @@ Arguments group_assoc_opp {_}.
 (** We should never need to unfold the proof that something is a group. *)
 Global Opaque group_isgroup.
 
+(** While this can be found by typeclass search, it is sometimes slow, so we add it as an instance. *)
+Instance ishset_group (G : Group) : IsHSet G := _.
+
 Definition issig_group : _ <~> Group
   := ltac:(issig).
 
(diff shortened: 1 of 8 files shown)


theories/Algebra/Groups/Group.v
```diff
@@ -55,6 +55,9 @@ Arguments group_assoc_opp {_}.
 (** We should never need to unfold the proof that something is a group. *)
 Global Opaque group_isgroup.
 
+(** While this can be found by typeclass search, it is sometimes slow, so we add it as an instance. *)
+Instance ishset_group (G : Group) : IsHSet G := _.
```
jdchristensen, line 59: I'll highlight this change.  For some reason in various AbSES proofs, it takes Rocq a while to search for the chain of instances that show that an Abelian group is a set.  It tries semi-groups and other algebraic structures before finding the right path.  So this gives  a short-cut.  It's redundant, which we usually try to avoid, but I think it's a net win.

Alizter (approved):
LGTM I don't mind redundant instances like this if they reduce the search tree

PR #2392, opened 2026-09-09
No double negation

cunha-francisco:
Added not_forall_dneg in Types.Universe, which corresponds to theorem 3.2.2 in the book.

### DIFF
--- a/theories/Types/Universe.v
+++ b/theories/Types/Universe.v
@@ -558,8 +558,10 @@ Proof.
   symmetry; apply transport_path_universe.
 Defined.
 
-(** And we can prove that it is not the case that for all types A, the double negation A implies A, which corresponds to theorem 3.2.2 in the book*)
-Definition not_forall_dneg: ~(forall (A : Type), (~~A) -> A).
+(** ** Not all types are stable *)
+
+(** Similar methods let us prove that it is not the case that every type is stable.  That is, is it not the case that for every type [A], the double negation of [A] implies [A].  This is Theorem 3.2.2 in the book. *)
+Definition not_forall_stable : ~(forall (A : Type), (~~A) -> A).
 Proof.
   intro f.
   pose (u := fun (g : Bool -> Empty) => g false).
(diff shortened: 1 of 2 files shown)


theories/Types/Universe.v
```diff
@@ -558,4 +558,21 @@ Proof.
...
 
+(** ** Not all types are stable *)
+
+(** Similar methods let us prove that it is not the case that every type is stable.  That is, is it not the case that for every type [A], the double negation of [A] implies [A].  This is Theorem 3.2.2 in the book. *)
+Definition not_forall_stable : ~(forall (A : Type), (~~A) -> A).
+Proof.
+  intro f.
+  pose (u := fun (g : Bool -> Empty) => g false).
+  refine (not_fixed_negb (f Bool u) _).
+  pose (p := (path_universe equiv_negb)).
```
jdchristensen: On this line you can remove a pair of parentheses.

theories/Types/Universe.v
```diff
@@ -558,4 +558,21 @@ Proof.
...
+  pose (f_eq_negf := happly ((apD f p)^) u @ transport_arrow p (f Bool) u @ (happly (transport_idmap_path_universe equiv_negb) (f Bool (transport (fun A => ~~A) p^ u)))).
+  refine ((f_eq_negf @ (ap (negb o (f Bool)) _ ))^).
```
jdchristensen: These two lines are very hard to follow.  Can you instead manipulate the LHS or RHS one step at a time?  Given your term, it might work best to manipulate the RHS.  E.g. the first two steps would be:
```coq
  rhs_V exact (ap10 (apD f p) u).
  rhs napply transport_arrow.
```
`rhs_V` applies the inverse of the term to the RHS, while `rhs` applies the term.  Use `exact` if you give the full term, `napply` if you give a partial term, `rapply` if typeclass search is needed.  `lhs` and `lhs_V` work similarly on the LHS.   Also note that in the library we use `ap10`, not `happly`, and that I got rid of unneeded parentheses in the first line.  There are other parts of your terms with extra parentheses as well, so check for that too.

theories/Types/Universe.v
```diff
@@ -558,4 +558,21 @@ Proof.
...
+  (* Our goal is to show that two functions to [Empty] are equal, which is trivial. *)
+  funext v.
+  contradiction (u v).
+Defined.
+
+
```
jdchristensen: Only one blank line between items.

jdchristensen (commented):
@cunha-francisco Thanks for this contribution!  I pushed a commit that cleaned up a couple of things, and added comments below for you to address.  Please let me know if you have any questions.

theories/Types/Universe.v
jdchristensen: Note that one of the advantages of the `lhs`/`rhs` approach is shown in `rhs napply transport_arrow.`, which requires to punctuation or underscores.

cunha-francisco: Thank you so much for mentioning those commands, I wasn't familiar with them before. I've pushed the changes.

jdchristensen (approved):
Thanks for this!  Feel free to contribute more, and skim through the STYLE.md file for more tips about our conventions.  @Alizter , do you want to take a quick look, or should I go ahead and merge?

PR #2395, opened 2026-09-17
Enable the long-path GpdHom notation

Alizter:


jdchristensen:
Is there a place where I can see how this looks?

Alizter:
@jdchristensen Yes, have a look at the proof of `pentagon_twist`. The scope is already opened there.

theories/WildCat/MonoidalTwistConstruction.v
```diff
@@ -63,10 +63,9 @@ Section TwistConstruction.
   Local Infix "⊗R" := (fmap01 cat_tensor) (at level 40) : monoidal_scope.
   Local Infix "⊗L" := (fmap10 cat_tensor) (at level 40) : monoidal_scope.
   Local Notation "f ∘ g" := (f $o g) (at level 61, left associativity, format "f  '/' '∘'  g") : monoidal_scope.
-  (** TODO: in 8.19 this notation causes issues, when updating to 8.20 we should uncomment this. *)
-  (* Local Notation "f $== g :> A" := (GpdHom (A := A) f g)
+  Local Notation "f $== g :> A" := (GpdHom (A := A) f g)
     (at level 70, format "'[v' '[v' f ']' '/'  $== '/'  '[v' g ']'  '/' :>  '[' A ']' ']'")
-    : long_path_scope. *)
+    : long_path_scope.
```
jdchristensen, line 68: I don't think the nested `'[v' ... ']'` around `f` and `g` do anything, do they? Doesn't the outer `'[v'` put the whole thing into vertical mode?

jdchristensen (approved):
LGTM, but maybe can be simplified as I indicated.  (But I'm probably missing a nuance.)

theories/WildCat/MonoidalTwistConstruction.v
Alizter: Here they do not. I think my intention at the time was to force `f` and `g` to be vertically formatted if they happened to be unboxed, but that doesn't seem to be the case here. They are redundant and therefore should be removed. The outer vertical box doesn't put the "whole" thing into vertical mode, but the outermost layer of boxes if that makes sense.

jdchristensen:
LGTM.

PR #2399, opened 2026-09-17
Spaces/Int: bind the integer notation scope

Alizter:


test/Spaces/Int.v
```diff
@@ -11,9 +11,9 @@ Definition test3 : int_to_sint (int_succ (int_pred2 zero)) = sint_zero := idpath
 
 Definition test4 : int_to_sint (int_neg (int_succ (int_succ zero))) = sint_NegS 1 := idpath.
 
-Definition test5 : int_to_sint (-4)%int = sint_NegS 3 := idpath.
+Definition test5 : int_to_sint (-4) = sint_NegS 3 := idpath.
 
-Definition test6 : int_to_sint ((0.+1)%int) = sint_PosS 0 := idpath.
+Definition test6 : int_to_sint (0.+1) = sint_PosS 0 := idpath.
 
 Definition test7 : int_to_sint ((0.+1)-1)%int = sint_zero := idpath.
```
jdchristensen: Just curious why some `%int` annotations are still needed.

jdchristensen (approved):
Nice!

test/Spaces/Int.v
Alizter: Seems to be because we define subtraction as an anonymous function. If instead we defined:

```
Definition int_sub (x y : Int) : Int := x + -y.
Infix "-" := int_sub : int_scope.
```

Then the bind would work. Should we do this here?

jdchristensen: Oh, I see.  I think it's probably worth having the function, but I'm fine with leaving it as is.

Alizter: I've pushed a follow up commit introducing `int_sub`. I've needed to adjust one lemma since `x + (-y)` is not immediately `x - y` any longer without unfolding further.

PR #2404, opened 2026-09-22
Minor backward compatible fixes in preparation for Rocq PR introducing Univ@{s;l}

mattam82:
`Univ` will become a keyword. We fix the `dummy` universe in EvalIn to avoid lowering it to Prop.
rocq-prover/rocq#22513


jdchristensen (approved):
Looks fine to me.

